PR #7514, opened 2021-05-11
Add strict null checking

zephraph:
I wouldn't feel right leaving without doing _something_ sizable.

Here's a little write-up: https://just-be.dev/notes/Projects/Enabling+Strict+Type+Checking+in+Force
 
This PR enables strict null checks on our codebase. It does _not_ enable complete strict checking. 

I've went through and added comments like
```
// @ts-ignore-error STRICT_NULL_CHECK
```
anywhere this new error appeared to get us to the point where all new code going forward can have strict null checks.

There's a CI step, **strict-migration-check**, which will check the number of those comments against master. If the number of comments on a PR is higher than that of master, the build will fail. This is essentially the mechanism we have for ensuring the incremental migration proceeds. 

I'd recommend doing a bug bash at some point to remove a bunch of these.


.eslintrc.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     "react-hooks",
     "@typescript-eslint",
     "styled-components-a11y",
-    "jest",
+    // "jest",
```
zephraph, line 8: This is just me trying to quiet things down. We should obviously re-enable this.

ds300 (approved):
🙇🏼 Yay strict mode!!! Thanks so much for taking the time to do this 💜 

My only non-blocking suggestion would be to replace betterer with the tooling I built for eigen because:

- Betterer adds 3-5s to the pre-commit hook.
- Betterer's snapshot file is liable to get out of sync during rebases/merges/conflict resolution.
- Betterer's snapshot file is liable to cause merge conflicts on open PRs. We have this problem with eigen's changelog and it's really annoying.

My script avoids all that by not using a snapshot file and only running on CI.

https://github.com/artsy/eigen/blob/master/.circleci/config.yml#L223-L225

https://github.com/artsy/eigen/blob/master/scripts/strictness-migration.js

It has been very robust. Can recommend 👍🏼

zephraph:
Thanks! Yeah, if I'm completely honest with myself the reason I used betterer in this case is mostly because I've wanted to use it on something for a while, but I haven't had the chance to. 

That, of course, is a pretty emotional decision and isn't necessarily aligned with what the team needs are. If I'm thinking objectively here, there are dependencies and complexities not needed. 

Thanks for the suggestion @ds300.

Let me dig into the script a little and figure out how to integrate it. In my mind, ideally, migration checks wouldn't be opt-in but rather opt-out.

damassi:
Thanks for the PR @zephraph 💯

PR #7518, opened 2021-05-11
feat: implements full bleed header

dzucconi:
This extracts the pattern we use for those full-bleed image headers; which are a little complicated to set up optimally.

I've aligned with Jonathan re: dimensions as far as the design system goes. It also supports a `caption` and can accept arbitrary children, for instance if you need to position some kind of banner text over this.

I'll follow up with some stories later in the Storybook.


src/v2/Components/FullBleedHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  const xs = cropped(src, { width: 767, height: 320 })
+  const sm = cropped(src, { width: 1024, height: 600 })
+  const md = cropped(src, { width: 1440, height: 600 })
+  const lg = cropped(src, { width: 2000, height: 600 })
+
+  return (
+    <Container bg="black10" height={600} position="absolute" {...rest}>
+      <picture>
+        <source srcSet={lg.srcSet} media="(min-width: 1720px)" />
```
dzucconi, line 27: These media values are the middle points between each size

src/v2/Components/FullBleedHeader.tsx
```diff
@@ -0,0 +1,72 @@
+import { BoxProps, Flex, HTML } from "@artsy/palette"
+import React from "react"
+import styled from "styled-components"
+import { FullBleed } from "v2/Components/FullBleed"
```
damassi, line 4: Don't you think we should use the `FullBleed` component from Palette?

src/v2/Components/FullBleedHeader.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Image src={md.src} alt="" loading="lazy" />
+      </picture>
+
+      {caption && (
+        <Overlay>
+          <HTML html={caption} color="white100" variant="xs" px={2} py={1} />
+        </Overlay>
+      )}
+
+      {children}
```
erikdstock, line 41: how would one position the children within this component? I'm not clear how the picture and overlay components work together- does picture not impact the flow of the dom?

erikdstock (commented):
No blockers here, but if this is mergeable i will hold off on #7516 . If not I may merge and revisit.

dzucconi:
Will merge when green

PR #7519, opened 2021-05-11
chore: updates Image urls to Vanity generated ones

ashleyjelks:
Additional QA fixes on Buyer Guarantee page:
- uses Vanity's static urls for page images
- Fixes some text misalignment

https://artsyproduct.atlassian.net/browse/PURCHASE-2620
https://artsyproduct.atlassian.net/browse/PURCHASE-2637

cc @artsy/purchase-devs

### DIFF
--- a/src/v2/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx
+++ b/src/v2/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx
@@ -59,6 +59,14 @@ export const BuyerGuaranteeIndex: React.FC<BuyerGuaranteeIndexProps> = ({
   const supportArticleURL =
     "https://support.artsy.net/hc/en-us/articles/360048946973"
 
+  const authenticityGuaranteeImageURL =
+    "http://files.artsy.net/authenticityguaranteeartwork.jpg"
+  const securePaymentImageURL =
+    "http://files.artsy.net/securepaymentartwork.jpg"
+  const moneyBackGuaranteeImageURL =
+    "http://files.artsy.net/moneybackguaranteeartwork.jpg"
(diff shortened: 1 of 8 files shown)


src/v2/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx
```diff
@@ -59,6 +59,14 @@ export const BuyerGuaranteeIndex: React.FC<BuyerGuaranteeIndexProps> = ({
   const supportArticleURL =
     "https://support.artsy.net/hc/en-us/articles/360048946973"
 
+  const authenticityGuaranteeImageURL =
+    "http://files.artsy.net/authenticityguaranteeartwork.jpg"
+  const securePaymentImageURL =
+    "http://files.artsy.net/securepaymentartwork.jpg"
+  const moneyBackGuaranteeImageURL =
+    "http://files.artsy.net/moneybackguaranteeartwork.jpg"
+  const heroImageURL = "http://files.artsy.net/buyerGuaranteeHeroImage.jpg"
```
dzucconi, line 68: You're going to want to pass these through the resized function and pull out optimized versions along with the srcSets.

ashleyjelks: The designer requested these images be used because they are resized according to the specs.  Is there a reason to resize them again?

dzucconi: According to the specs with the expectation that they'll be run through our resize function maybe? Currently they are at  ~1600x1200 @1x and non-optimized. Resizing them to exactly the 1x display dimensions and utilizing the returned `srcSet` ensures they are correct .

PR #7199, opened 2021-03-23
feat: add partner header image

kiryl-zubarau:
This PR adds a fullwidth partner image to the profile page

JIRA - https://artsyproduct.atlassian.net/browse/PX-3823

image

image

PR #7526, opened 2021-05-12
fix(PX-4080): partner page layout

kiryl-zubarau:
This PR fixes the current partner page layout. The partner profile page has special scroll behavior for the header image, the solution described here: https://github.com/artsy/force/pull/7199#discussion_r600512091. In short, we use a fixed image and foreground which scrolls over it. As far as I see this PR https://github.com/artsy/force/pull/7497 broke partner page layout. To fix it I moved back `AppContainer` and `HorizontalPadding`. 

JIRA - https://artsyproduct.atlassian.net/browse/PX-4080

Before: 
image

Now:
image

### DIFF
--- a/src/v2/Apps/Partner/PartnerApp.tsx
+++ b/src/v2/Apps/Partner/PartnerApp.tsx
@@ -8,6 +8,8 @@ import { PartnerHeaderImageFragmentContainer as PartnerHeaderImage } from "./Com
 import styled from "styled-components"
 import { PartnerMetaFragmentContainer } from "./Components/PartnerMeta"
 import { StickyProvider } from "v2/Components/Sticky"
+import { AppContainer } from "../Components/AppContainer"
+import { HorizontalPadding } from "../Components/HorizontalPadding"
 
 export interface PartnerAppProps {
   partner: PartnerApp_partner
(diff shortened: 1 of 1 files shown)


dzucconi:
Wondering if we're going to use this fixed style image header elsewhere — in which case I'm not opposed to over-engineering a solution to avoid all this trouble. 🤔

kiryl-zubarau:
AFAIK only the home page has the same scroll behavior. https://staging.artsy.net/

dzucconi:
> AFAIK only the home page has the same scroll behavior

So far at least 😬

kiryl-zubarau:
> So far at least 😬

So, do you want me to add a special component for this? Maybe it'll be better to make the changes in separate PR because it's blocking QA?

dzucconi:
Oh no no, definitely not! Just thinking outloud!

damassi:
Hmm, @kiryl-zubarau  - I'm a bit torn because pretty soon this page will need to be migrated to Palette v3, whch will then extend things full screen automatically. Is there anything we can do to maybe encapsulate this behavior without adding back the `AppContainer` and `HorizontalPadding` components? By adding these back this page now becomes an outlier compared to the rest of our application stack (post #7497).

kiryl-zubarau:
@damassi that is a good question. As far as I understood the main reason why we got rid of using `AppContainer `and `HorizontalPadding` it's because we want to control page width at the one place - `AppShell`, am I right?
I have a few ideas, but I will only be able to test them tomorrow. I understand that it's not good, but could you please merge this PR because our QA engeneer  blocked and is waiting for this issue to be resolved?

damassi:
Oh for sure! Didn't realize this was blocking. Will merge it 👍

PR #7530, opened 2021-05-12
feat(v3): converts /collection/:id to v3

dzucconi:
https://user-images.githubusercontent.com/112297/118010568-2c161400-b31d-11eb-85e9-5b13190b52e6.mov


dzucconi:
Will probably be pending a few more specs

src/v2/Components/RelatedCollectionsRail/RelatedCollectionsRail.tsx
```diff
@@ -38,14 +37,15 @@ export const RelatedCollectionsRail: React.FC<RelatedCollectionsRailProps> = pro
 
   if (collectionsWithArtworks.length > 3) {
     return (
-      <Box>
+      <>
+        {/* FIXME: Must remove this dependency */}
         <Waypoint onEnter={once(trackImpression)} />
```
dzucconi, line 42: Assuming this is what's actually crashing the page. We should probably just add an intersection observer hook at some point.

damassi (approved):
Oof 😅 this was more than expected. No blockers, lets get it out and iterate 👍

PR #7535, opened 2021-05-12
feat: converts /collections to v3

dzucconi:
Not a particularly important page; I was just using it to check a bunch of collections and so now I think that covers the whole /collect app.


src/v2/Apps/Collect/Routes/Collections/Components/CollectionsCategory.tsx
```diff
@@ -0,0 +1,88 @@
...
+export const CollectionsCategory: React.FC<CollectionsCategoryProps> = ({
+  category,
+  category: { collections },
+}) => {
+  const [showAll, toggleShowAll] = useState(false)
+
+  const sorted = [...collections].sort((a, b) => a.title.localeCompare(b.title))
+
+  const displayableCollections =
+    showAll || isServer ? sorted : sorted.slice(0, TRUNCATED_AMOUNT)
```
dzucconi, line 27: I copied this logic over; I think this is trying to render all of them for Google but truncate them for the real user? It's kinda weird... leads to a flash if you do a full page load. I guess we could just hide them with CSS instead?

damassi: I vaguely remember this, yeah it was for SEO reasons, a css solution would def be better but not a blocker

damassi (approved):

src/v2/Apps/Collect/Routes/Collections/Components/CollectionsCategory.tsx
dzucconi: 71cf58143669271c1fa6c9df038212dac637a097 👍

PR #7536, opened 2021-05-12
smoke-tests: Bump retries to 5

damassi:
We're already on 5, so lets bump retries up a bit.


cypress.json
```diff
@@ -1,4 +1,4 @@
 {
   "baseUrl": "http://localhost:5000",
-  "retries": 3
+  "retries": 5
```
dzucconi, line 3: 😬

PR #7538, opened 2021-05-12
feat: converts /artists & /artists/:letter to v3

dzucconi:
https://user-images.githubusercontent.com/112297/118045179-993d9f80-b345-11eb-8570-21da118f5740.mov

### DIFF
--- a/src/v2/Apps/Artists/Components/ArtistsArtistCard.tsx
+++ b/src/v2/Apps/Artists/Components/ArtistsArtistCard.tsx
@@ -1,37 +1,31 @@
 import React from "react"
-import { Box, BoxProps, Image, ResponsiveBox, Text } from "@artsy/palette"
+import { Box, Flex, Image, ResponsiveBox, Text } from "@artsy/palette"
 import { RouterLink } from "v2/Artsy/Router/RouterLink"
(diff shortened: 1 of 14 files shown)


src/v2/Apps/Artists/Components/ArtistsArtistCard.tsx
```diff
@@ -1,37 +1,31 @@
...
-      {...rest}
-    >
-      <RouterLink to={artist.href} noUnderline style={{ display: "block" }}>
+    <>
+      <RouterLink
+        to={artist.href}
+        noUnderline
+        style={{ display: "block" }}
+        // Avoid the redundant tab-stop
+        tabIndex={-1}
```
damassi, line 27: What do you mean? Can you paste in an example? (wondering if i should be doing this too)

damassi (approved):

src/v2/Apps/Artists/Components/ArtistsArtistCard.tsx
dzucconi: This is just because in this case we've got the follow button in there so we can have one big link; so it's broken into two links and I'm just skipping the first one when tabbing via keyboard.

PR #7540, opened 2021-05-12
Remove references to unmaintained URLs

joeyAghion:
There are several references to unmaintained URLs in the source. In all cases, our deployed applications already override these defaults with valid, maintained endpoints. Even so, the `*.herokuapp.com` URLs are catnip to "security researchers" and the others might cause confusion or mislead developers.

The `EUROPA_URL` led me down a small detour to clean up unreferenced backbone classes.


dzucconi:
🙏

PR #7542, opened 2021-05-13
fix(partner): fixes _ is not a function error

dzucconi:
OK let this be a lesson to myself to actually look into errors that crop up in the smoke tests instead of dismissing them as spurious. I think this was sometimes passing because it was in the callback after an async action. I don't know _why_ this was actually undefined because this should be OK, but I'm not motivated enough to track down the root cause of an issue in a Backbone view that's going to get replaced in a week or two anyway.


damassi:
😄

PR #7543, opened 2021-05-13
fix(PX-4080): remove AppContainer and HorizontalPadding from partner app component

kiryl-zubarau:
Removed AppContainer and HorizontalPadding from the partner app component.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4080

ezgif com-gif-maker (27)


damassi:
Thanks for following up with this @kiryl-zubarau 👍

PR #7548, opened 2021-05-13
fix: [fair/:id] remove related articles header to remove whitespace

damassi:
Addresses https://artsyproduct.atlassian.net/browse/FX-2921 

This removes the related articles header section on the page, as it was leading to strange whitespace issues.

### DIFF
--- a/src/v2/Apps/Fair/FairApp.tsx
+++ b/src/v2/Apps/Fair/FairApp.tsx
@@ -73,12 +73,8 @@ const FairApp: React.FC<FairAppProps> = ({ children, fair }) => {
       <FairHeaderFragmentContainer fair={fair} />
 
       {hasArticles && (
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -72,25 +66,6 @@ const FairApp: React.FC<FairAppProps> = ({ children, fair }) => {
...
-            {fair.articlesConnection.totalCount > FAIR_EDITORIAL_AMOUNT && (
-              <RouterLink to={`${fair.href}/articles`} noUnderline>
-                <Text variant="sm">View all</Text>
-              </RouterLink>
-            )}
-          </Box>
-
-          <FairEditorialFragmentContainer fair={fair} />
-        </Box>
-      )}
```
damassi, line 92: Note, I didn't delete the actual components because perhaps we might want to add this back in a more visually elegant way later?

PR #7554, opened 2021-05-14
feat: create top banner for non subscriber

ivan-badyulya:
JIRA -> [Update Top Banner on Overview page for non-subscribing partners](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4022&quickFilter=228)

**Attention**
I used `hasFairPartnership ` field so far to distinguish `fair` partners from `churned`. I am not completely sure if it is the right field.

This is a link to discussion with Sarah and Oksana about that -> [Should I use `hasFairPartnership`? ](https://artsy.slack.com/archives/C01NW1YSDPF/p1620919437034500)

### DIFF
--- a/src/v2/Apps/Partner/Components/Overview/SubscriberBanner.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/SubscriberBanner.tsx
@@ -1,0 +1,50 @@
+import React from "react"
+import { Box, Text } from "@artsy/palette"
+import { RouterLink } from "v2/Artsy/Router/RouterLink"
+import { createFragmentContainer, graphql } from "react-relay"
+import { SubscriberBanner_partner } from "v2/__generated__/SubscriberBanner_partner.graphql"
+
+export interface SubscriberBannerProps {
+  partner: SubscriberBanner_partner
+}
+
+export const SubscriberBanner: React.FC<SubscriberBannerProps> = ({
+  partner: { hasFairPartnership, name },
+}) => {
+  return (
+    <Box p={2} bg="black5">
+      {hasFairPartnership ? (
+        <Text
+          mb={1}
+          variant="title"
+        >{`${name} participates in Artsy’s art fair coverage but does not have a full profile.`}</Text>
+      ) : (
+        <Text
+          mb={1}
+          variant="title"
+        >{`${name} is not currently an Artsy partner and does not have a full profile.`}</Text>
+      )}
+      <Text
+        display="inline"
+        variant="subtitle"
+      >{`Do you represent ${name}?`}</Text>
+      <RouterLink to="https://partners.artsy.net/gallery-partnerships/">
+        <Text display="inline" variant="subtitle">
+          &nbsp;Learn about Artsy gallery partnerships.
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Partner/Components/Overview/SubscriberBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Box, Text } from "@artsy/palette"
+import { RouterLink } from "v2/Artsy/Router/RouterLink"
+import { createFragmentContainer, graphql } from "react-relay"
+import { SubscriberBanner_partner } from "v2/__generated__/SubscriberBanner_partner.graphql"
+
+export interface SubscriberBannerProps {
+  partner: SubscriberBanner_partner
+}
+
+export const SubscriberBanner: React.FC<SubscriberBannerProps> = ({
```
dzucconi, line 11: Should this be a [Message](https://palette-storybook.artsy.net/?path=/story/components-message--default) component?

ivan-badyulya: Do you mean use `Message` component instead of `SubscriberBanner` or  wrapping the layout inside into `Message` would be enough?

I mean something like that

Screen Shot 2021-05-14 at 5 38 30 PM

dzucconi: Yeah; in this case you'd pass it the `title` prop for the title; and children for the description.

ivan-badyulya: @dzucconi Fixed. Butin this case I will be unable to set the font size for title. It is 24px in figma.

dzucconi: Should just use the design system component. The Figma is likely mistaken or incorrect.

oxaudo: #neat-picky - but I would name this something like `FullProfileBanner` or something to indicate that this is full page vs limited page. I think we should not tie this to subscription. Some partners like institutions might get it too. And we might change the rules who gets what view to not depend on subscription. This subscription logic should be incapsulated in Gravity/MP. Ui should not encapsulate this rule I believe.

oxaudo: Also  - is not this `LimitedPageBanner` - that is what people that are not subscribers get?

ivan-badyulya: @oxaudo yes, this component displays on limited overview page so it might be a good idea to name it like `LimitedPageBanner`. On the other hand If I named it LimitedPageBanner then it would mean that we also should have FullPageBanner/FullProfileBanner but we don't :)

 I named it `SubscriberBanner` to emphasize that the banner has some info related to subscribe status.

PR #7555, opened 2021-05-14
feat: update about section non subscribe

ivan-badyulya:
JIRA -> [Update ‘About’ section on Overview page for non-subscribing partners](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4023&quickFilter=228)


src/v2/Apps/Partner/Components/Overview/AboutPartner.tsx
```diff
@@ -25,15 +28,16 @@ export const AboutPartner: React.FC<AboutPartnerProps> = ({
 
       <Column span={6}>
         {profile?.fullBio && <Text variant="text">{profile.fullBio}</Text>}
-        {website && (
+        {canRenderWebsite && (
+          /* @ts-expect-error STRICT_NULL_CHECK */
```
dzucconi, line 32: Shouldn't be introducing new ignores here — I don't see why the guard he wouldn't be sufficient though.

ivan-badyulya: Screen Shot 2021-05-14 at 5 27 30 PM

dzucconi: If we know for sure it'll never be null I'd rather use an assertion `{website!}` vs the ts-ignore

ivan-badyulya: Fixed

PR #7556, opened 2021-05-14
feat: update shows for non subscriber

ivan-badyulya:
JIRA -> [Update Show rail on Overview page for non-subscribing partners](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4024&quickFilter=228)


src/v2/Apps/Partner/Components/Overview/ShowsRail.tsx
```diff
@@ -17,44 +17,47 @@ const ShowsRail: React.FC<ShowsRailProps> = ({ partner, ...rest }) => {
...
+    showsConnection: { edges },
+    fullProfileEligible,
   } = partner
 
+  const canShowAll = isSeeAllAvaliable && fullProfileEligible
+  const shows = [...edges]
+
+  if (fullProfileEligible && shows.length === 20) {
+    shows.pop()
+  }
```
dzucconi, line 37: This is fine, just wondering the "why" of it?

ivan-badyulya: well, if we have a non-subscriber then we have to display 20 cards without `view all` card. But if we have a subscriber we have to display 19 cards + `view all` card. I decided that it would be more efficient to replace the last card with `view all` card if a user is subscriber.

The issue was that we have to fetch first 20 for subs and first 19 for non subs.

PR #7561, opened 2021-05-14
fix: implements sticky footer

dzucconi:
I think there was a ticket associated with this but I can't find it now.

### DIFF
--- a/src/v2/Apps/Components/AppContainer.tsx
+++ b/src/v2/Apps/Components/AppContainer.tsx
@@ -4,6 +4,9 @@ import React from "react"
 
 interface AppContainerProps extends BoxProps {}
 
+/**
+ * Handles the max-width of the application. That's it!
+ */
 export const AppContainer: React.FC<AppContainerProps> = ({
   children,
   maxWidth: defaultMaxWidth,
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Components/HorizontalPadding.tsx
```diff
@@ -1,36 +1,14 @@
...
-    `};
-  ${p =>
-    // @ts-expect-error STRICT_NULL_CHECK
-    p.px[1] &&
-    css`
-      padding-right: ${// @ts-expect-error STRICT_NULL_CHECK
-      space(p.px[1])}px;
-      padding-left: ${// @ts-expect-error STRICT_NULL_CHECK
-      space(p.px[1])}px;
-    `};
```
dzucconi, line 29: All this was doing was just manually allowing padding overwrites — which doesn't need to be done if you use `defaultProps`

src/v2/Apps/Components/AppContainer.tsx
```diff
@@ -4,6 +4,9 @@ import React from "react"
 
 interface AppContainerProps extends BoxProps {}
 
+/**
+ * Handles the max-width of the application. That's it!
+ */
```
damassi, line 9: ty 👍

src/v2/Apps/Components/HorizontalPadding.tsx
damassi: heh, nice refactor

damassi:
🎉

PR #7566, opened 2021-05-17
fix: added checking for article title existence and filling in the em…

anastasiapyzhik:
…pty title

This is supposed to fix https://artsyproduct.atlassian.net/browse/GRO-324

Articles that had property `title` equal to `" "` didn't render `h1` tag so I check for `title` property existence and if it's missing set `thumbnail_title` value to it.

But I'm not sure whether it's got valid for SEO because, after all, `h1` tag was visible in elements tab, though I haven't found any other differences between articles which had `h1` tag and those hadn't
Screen Shot 2021-05-17 at 5 05 51 PM

Before
before

After
after


jonallured:
We decided to not do this after all - more chatter here: https://artsy.slack.com/archives/C01ADJNCS5D/p1621347556007800

PR #7569, opened 2021-05-17
chore: Remove Nightmare.js-based acceptance tests

damassi:
This removes our Nightmare.js-based acceptance tests in favor of our more modern Cypress based smoke tests.

### Some history: 

Back in 2017 we wanted to find a way to better ensure our end-to-end flows worked, so Craig spun up a library wrapper around Nightmare.js which would inject fixture data into a mock service layer that would be served to a running local instance of force. As end-to-end testing has always been something of a delicate matter, this work was largely an experiment to see how it would go. And it worked well! There was no flake, and issues were occasionally caught. 

The problem, however, centered around updates to pages, and fixing errors when they did arise. Stack traces from errors are indecipherable, and the complexity of capturing fixture data to use for new pages led to very few new pages being added to the system, which further contributed to the confusion around errors since the team didn't know how to use it.

A couple years pass and Cypress is released, which has become the standard for end-to-end testing in the JS ecosystem. First we added it in the form of Integrity, our end-to-end testing service, and then later we added it to force, for smoke-test reasons. 

Devs have been able to contribute both to Integrity and to Force's new smoke tests relatively easily, since the new Cypress-based system relies on staging data, rather than fixture data, and has a modern and intuitive UX/DX. 

Since these Cypress-based smoke tests have been added to force we've experienced very little flake, which has been great. And errors due to changes to staging data and / or app code have been fixed quickly and efficiently, and we've been able to add new pages without issue. 

And so after our long trial, it's time to finally delete Nightmare from our stack and rely exclusively on Cypress. Nightmare issues still often pop up and consistently cause confusion, and this PR will further simplify our tech stack.

Assigning to @anandaroop as he was the first to add this to Force.


anandaroop (approved):
Yep, makes sense to me 👍🏽 

These tests were flake-free and confidence-inspiring _to the extent_ that the accompanying fixtures accurately reflected both the upstream data and the local Relay queries. It was the latter that in practice would get out of sync, and were very onerous to maintain, so I'm happy to see these go.

Two caveats:

- The only potential drawback I can think of is the issue of staging flakiness. I think that's the reason why we kept both suites to begin with. As far as I can remember the Cypress suite was added at a time of _extra_ staging flakiness, which is what made us hesitant to rip out the old one. While it may be better now, I do think staging flakiness remains an issue. Then again if staging issues are transient and a simple `Rebuild ⟲` on Circle takes care of it, then maybe that's fine. (The cypress step will continue to fail a PR build, correct?)

- Some of the Nightmare specs seems to cover areas that weren't yet ported over to cypress — auth and bidding. Should we  prioritize new smoke tests for those?

damassi:
> Some of the Nightmare specs seems to cover areas that weren't yet ported over to cypress — auth and bidding. Should we prioritize new smoke tests for those?

We could, although these areas are thoroughly covered in Integrity. Feel free to file a ticket on the [Web Practice](https://artsyproduct.atlassian.net/jira/software/projects/WP/boards/97) board.

PR #7575, opened 2021-05-18
fix: Deleted extra slash in href string

anastasiapyzhik:
This fixes https://artsyproduct.atlassian.net/browse/GRO-332


jonallured (approved):
LGTM!! 🎉

PR #7576, opened 2021-05-18
fix: buyer guarantee icon fixes

jo-rs:
Resolves [PURCHASE-2635]

Also, pulls icons into its own file, since the buyer guarantee page file is quite large, and standardises them with Palette components in case we want to move them later.

comparison Screenshot 2021-05-17 at 21 16 53

[PURCHASE-2635]: https://artsyproduct.atlassian.net/browse/PURCHASE-2635


xtina-starr (approved):
This looks good!

ashleyjelks (approved):

PR #7580, opened 2021-05-18
fix: deleted hiding of Museums tab in nav bar

anastasiapyzhik:
This fixes https://artsyproduct.atlassian.net/browse/GRO-247

**Before**
nTab_before

**After**
nTab_after

In order to avoid wrapping of "Download App" button text I added to nav items negative x-margin that decreases space between them. Without this change it looks like:
wrap

### DIFF
--- a/src/v2/Components/NavBar/NavBar.tsx
+++ b/src/v2/Components/NavBar/NavBar.tsx
@@ -84,6 +84,8 @@ export const NavBar: React.FC = track(
     }
   }
 
+  const navItemsMargins = ["-2px", "-2px", "0", "0"]
+
   return (
     <>
       <NavBarSkipLink />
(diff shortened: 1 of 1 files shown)


src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -235,17 +237,22 @@ export const NavBar: React.FC = track(
...
 
-                <NavItem href="/auctions">Auctions</NavItem>
-                <NavItem href="/viewing-rooms">Viewing&nbsp;Rooms</NavItem>
-                <NavItem href="/galleries">Galleries</NavItem>
-                <NavItem href="/fairs">Fairs</NavItem>
-                <NavItem href="/Shows">Shows</NavItem>
-                <NavItem
-                  // Hide link at smaller viewports — corresponding display inside of `MoreNavMenu`
-                  // If we need to do this again, consider a more abstract solution
-                  display={["none", "none", "flex", "flex"]}
```
damassi, line 246: I think we can reduce the changes to this PR to just removing this `display` line. 

(If not familiar with this pattern, our design system Palette uses ["responsive props" via styled-system](https://styled-system.com/responsive-styles)

damassi:
@anastasiapyzhik If you're done with this PR be sure to use the `assignee` field to assign it to a specific grow dev.

jonallured:
Oh I wanted to call out on this one that I think we'll want to get some design feedback from @deliciousnachos when she's back next week!

deliciousnachos:
Ah nice, this works! Thanks @anastasiapyzhik

damassi:
Ah sorry @deliciousnachos - we should really avoid breaking out of our design system, per my comments above.

jonallured:
For posterity: we're going to close this for now and return to it after some design conversations settle.

PR #7581, opened 2021-05-18
feat(PX-4027): add nearby galleries rail

kiryl-zubarau:
This PR adds nearby galleries rail.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4027, https://artsyproduct.atlassian.net/browse/PX-4026

image

### DIFF
--- a/src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
@@ -8,6 +8,7 @@ import { useSystemContext } from "v2/Artsy"
 import { NearbyGalleryCardFragmentContainer } from "./NearbyGalleryCard"
 import { NearbyGalleriesRailPlaceholder } from "./NearbyGalleriesRailPlaceholder"
 import { SystemQueryRenderer as QueryRenderer } from "v2/Artsy/Relay/SystemQueryRenderer"
+import { compact } from "underscore"
 
 interface NearbyGalleriesRailProps extends BoxProps {
   partners: NearbyGalleriesRail_partners
@@ -23,6 +24,8 @@ const NearbyGalleriesRail: React.FC<NearbyGalleriesRailProps> = ({
     return null
   }
 
+  const partnerList = compact(partners.map(c => c.node))
+
   return (
     <Box {...rest}>
       <Flex mb={4} justifyContent="space-between" alignItems="center">
@@ -30,12 +33,12 @@ const NearbyGalleriesRail: React.FC<NearbyGalleriesRailProps> = ({
       </Flex>
 
       <Carousel itemsPerViewport={[2, 2, 3]}>
-        {partners.map(({ node }) => {
+        {partnerList.map(node => {
           return (
             <NearbyGalleryCardFragmentContainer
-              key={node?.id}
+              key={node.id}
               width={[300, "100%"]}
-              partner={node!}
+              partner={node}
               city={city}
             />
           )
--- a/src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
@@ -1,0 +1,105 @@
+import React from "react"
+import { Box, BoxProps, Flex, Text } from "@artsy/palette"
+import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import { NearbyGalleriesRail_partners } from "v2/__generated__/NearbyGalleriesRail_partners.graphql"
+import { NearbyGalleriesRailRendererQuery } from "v2/__generated__/NearbyGalleriesRailRendererQuery.graphql"
+import { Carousel } from "../Carousel"
+import { useSystemContext } from "v2/Artsy"
+import { NearbyGalleryCardFragmentContainer } from "./NearbyGalleryCard"
+import { compact } from "lodash"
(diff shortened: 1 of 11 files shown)


src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
```diff
@@ -0,0 +1,105 @@
...
+export const NearbyGalleriesRailRenderer: React.FC<
+  {
+    near: string
+  } & Omit<NearbyGalleriesRailProps, "partners">
+> = ({ near, ...rest }) => {
+  const { relayEnvironment } = useSystemContext()
+
+  return (
+    <QueryRenderer<NearbyGalleriesRailRendererQuery>
+      //  @ts-expect-error STRICT_NULL_CHECK
```
sweir27, line 73: I'm surprised we need this `//  @ts-expect-error STRICT_NULL_CHECK` here! What error do you get if you remove?

src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
```diff
@@ -0,0 +1,105 @@
...
+    <Box {...rest}>
+      <Flex mb={4} justifyContent="space-between" alignItems="center">
+        <Text variant="title">Nearby Galleries</Text>
+      </Flex>
+
+      <Carousel itemsPerViewport={[2, 2, 3]}>
+        {partners.map(({ node }) => {
+          return (
+            <NearbyGalleryCardFragmentContainer
+              key={node?.id}
```
sweir27, line 36: If `node` can be null here, can we just remove those cases from the `partners` array before iterating through?

src/v2/Apps/Partner/Components/Overview/NearbyGalleryCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+        profile {
+          image {
+            cropped(height: 300, width: 400, version: "wide") {
+              src
+              srcSet
+            }
+          }
+          ...FollowProfileButton_profile
+        }
+        locationsConnection(first: 20) {
```
sweir27, line 131: 20 seems like a lot to pull for this little card-- can it be like `5` instead?

src/v2/Apps/Partner/Components/Overview/NearbyGalleryCard.tsx
```diff
@@ -0,0 +1,142 @@
...
+interface NearbyGalleryCardProps extends BoxProps {
+  partner: NearbyGalleryCard_partner
+  city: string | null
+}
+
+function normalizeCityName(city: string) {
+  return capitalize(city.trim())
+}
+
+function getLocation(cities: Array<string>, preferredCity: string | null) {
```
sweir27, line 20: Would you mind adding some tests for this?

src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
kiryl-zubarau: Did you mean something like this?
image

In this case, we still have the same error
image

src/v2/Apps/Partner/Components/Overview/NearbyGalleryCard.tsx
kiryl-zubarau: locationsConnection using here to show this line
image

The city at the card should be equal to the city from the first partner location, so if I load a small number of locations there may not be a preferred city and I'll show the wrong city.

What do you think if we change the line from ‘CityName & # other locations’ to '# locations'? I think in this case we can use just `totalCount` from the `locationsConnection`.

src/v2/Apps/Partner/Components/Overview/NearbyGalleryCard.tsx
kiryl-zubarau: Sure, I will add it.

src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
sweir27: Yeah that's what I meant. Surprised we're still getting that-- @dzucconi (or others), is this expected?

dzucconi: Filter doesn't necessarily return the correct type I believe. Just `if (!node) return null` should be fine though.

src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
kiryl-zubarau: Replaced `QueryRenderer` with `SystemQueryRenderer` to fix a problem described here: https://github.com/artsy/force/pull/7605. It also solve the problem with STRICT_NULL_CHECK

src/v2/Apps/Partner/Components/Overview/NearbyGalleryCard.tsx
kiryl-zubarau: Done

src/v2/Apps/Partner/Components/Overview/NearbyGalleriesRail.tsx
kiryl-zubarau: `if (!node) return null` also results in a type error because `Carousel` doesn't work with empty children.
image

I used `compact(partners.map(c => c.node))`. Let me know if you are okay with it.

PR #7582, opened 2021-05-18
GRO-336: Implement IP address tracking solution in Force 

The-Beez-Kneez:
This is to complete the work associated with the ask from Legal that required an auto-enrollment in Email Subscriptions during onboarding sign up for US users only.

We updated Metaphysics to supply us with country code based on IP address of the users during sign up, and altered the `SignUpForm` to display these conditional checkboxes.

GDPR Country Experience:
Screen Shot 2021-05-18 at 2 29 21 PM

USA Experience:
Screen Shot 2021-05-18 at 2 29 27 PM

### DIFF
--- a/src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
+++ b/src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
@@ -60,6 +60,8 @@ export function bootstrapSharifyAndContextLocalsMiddleware(
 
   updateSharifyAndContext(res, "IS_GOOGLEBOT", Boolean(ua.match(/Googlebot/i)))
 
+  updateSharify(res, "IP_ADDRESS", req.ip)
+
   // Required to know the hashed dll name.try
   updateSharifyAndContext(
     res,
--- a/src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
+++ b/src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
@@ -60,7 +60,8 @@ export function bootstrapSharifyAndContextLocalsMiddleware(
 
   updateSharifyAndContext(res, "IS_GOOGLEBOT", Boolean(ua.match(/Googlebot/i)))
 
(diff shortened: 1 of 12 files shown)


src/typings/sharify.d.ts
```diff
@@ -39,6 +39,7 @@ declare module "sharify" {
       readonly IMAGE_LAZY_LOADING: boolean
       IS_GOOGLEBOT: boolean
       IS_MOBILE: boolean
+      IP_ADDRESS: string
```
damassi, line 42: we can also add `readonly` here

The-Beez-Kneez: Added.

src/v2/Components/Authentication/Desktop/SignUpForm.tsx
```diff
@@ -164,3 +211,48 @@ export class SignUpForm extends Component<FormProps, SignUpFormState> {
...
+      variables={variables}
+      query={graphql`
+        query SignUpFormLocationQuery($ip: String!) {
+          requestLocation(ip: $ip) {
+            ...SignUpForm_requestLocation
+          }
+        }
+      `}
+      render={({ error, props }) => {
+        // what's the cooler way to do this??
```
dzucconi, line 240: 😎

src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
```diff
@@ -60,6 +60,8 @@ export function bootstrapSharifyAndContextLocalsMiddleware(
 
   updateSharifyAndContext(res, "IS_GOOGLEBOT", Boolean(ua.match(/Googlebot/i)))
 
+  updateSharify(res, "IP_ADDRESS", req.ip)
```
damassi, line 63: Ah, it looks like @jonallured might have this commit

src/v2/Components/Authentication/Desktop/SignUpForm.tsx
jonallured: What are you saying the code that's there is _already_ cool?? 😝

PR #7583, opened 2021-05-18
feat: converts recently viewed to v3

dzucconi:
Closes: [DSWGW-58](https://artsyproduct.atlassian.net/browse/DSWGW-58)

Implements a V3 version of `RecentlyViewed` — also changes how we do the placeholder for these.

Previously we would have a weirdly placed loading spinner, then the component would pop in. Now there's an appropriately sized placeholder. One thing to note is that if you *haven't* recently viewed anything then you'll see the placeholder which will then pop _out_. Since we only show these to logged in users and a logged in viewer is pretty likely to have looked at at least one artwork, I figure that it's more appropriate to include the placeholder than not.

I also refactored the lazy-loading into the component itself since we lazy-load it in all cases that it appears.

### DIFF
--- a/src/v2/Utils/Hooks/useLazyLoadComponent.tsx
+++ b/src/v2/Utils/Hooks/useLazyLoadComponent.tsx
@@ -20,13 +20,22 @@ export const useLazyLoadComponent = ({
   const [isEnteredView, setIsEnteredView] = useState(false)
 
   const Waypoint = () => (
-    <LazyLoadComponent threshold={threshold} style={{ display: "inline" }}>
-      <Ready
-        onMount={() => {
-          setIsEnteredView(true)
-        }}
-      />
-    </LazyLoadComponent>
+    <>
+      {/*
+       * While although I have no reason to believe this client-side only check is necessary,
+       * it was wrapping every instance of this library, so it's been copied here.
+       * In the future we should replace this library with a simple intersection observer hook.
+       */}
+      {typeof window !== "undefined" && (
+        <LazyLoadComponent threshold={threshold} style={{ display: "inline" }}>
+          <Ready
+            onMount={() => {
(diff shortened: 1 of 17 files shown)


src/v2/Utils/Hooks/useLazyLoadComponent.tsx
```diff
@@ -20,13 +20,22 @@ export const useLazyLoadComponent = ({
...
-          setIsEnteredView(true)
-        }}
-      />
-    </LazyLoadComponent>
+    <>
+      {/*
+       * While although I have no reason to believe this client-side only check is necessary,
+       * it was wrapping every instance of this library, so it's been copied here.
+       * In the future we should replace this library with a simple intersection observer hook.
+       */}
```
dzucconi, line 28: 🤷

src/v2/Components/RecentlyViewed/V3RecentlyViewed.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import { extractNodes } from "v2/Utils/extractNodes"
+import { ShelfArtworkFragmentContainer } from "../Artwork/ShelfArtwork"
+import { V3RecentlyViewedPlaceholder } from "./V3RecentlyViewedPlaceholder"
+
+export interface V3RecentlyViewedProps {
+  me: V3RecentlyViewed_me
+}
+
+export const V3RecentlyViewed: React.FC<V3RecentlyViewedProps> = ({ me }) => {
```
damassi, line 19: Can we update this name? `RecentlyViewedV3` follows other idioms

dzucconi: Ah, yeah sure

PR #7584, opened 2021-05-18
dont-merge: Refactor sign up

jonallured:
The purpose of this PR is to demonstrate how we could use Formik to conditionally set email preferences via the terms checkbox. There's some refactoring here that's coincidental so the important bits are:

* create a boolean called `collapseCheckboxes` that decides if we should collapse the checkboxes (I spoofed this!)
* suppress rendering `EmailSubscriptionCheckbox` with that boolean
* send that boolean to the `TermsOfServiceCheckbox` along with [`setFieldValue`][sfv] from Formik
* inside `TermsOfServiceCheckbox` intercept change events and use `setFieldValue` to imperatively update `agreed_to_receive_emails`

@The-Beez-Kneez next steps would be to blend these changes with the work we did with @damassi yesterday - happy to work together on this!

[sfv]: https://formik.org/docs/api/formik#setfieldvalue-field-string-value-any-shouldvalidate-boolean--void

### DIFF
--- a/src/lib/middleware/sharifyLocals.ts
+++ b/src/lib/middleware/sharifyLocals.ts
@@ -60,6 +60,8 @@ export function sharifyLocalsMiddleware(
 
   updateSharify(res, "IS_GOOGLEBOT", Boolean(ua.match(/Googlebot/i)))
 
+  updateSharify(res, "IP_ADDRESS", req.ip)
+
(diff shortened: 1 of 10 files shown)


jonallured:
Closing this branch, it served it's purpose.

PR #7594, opened 2021-05-19
fix: display works tab fix

ivan-badyulya:
JIRA -> [PPP: Works tab remains visible after turning it off in CMS](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4098&quickFilter=228)

Added hidden property to Works tab route. From now the works tab won't display if it is turned off via CMS

### DIFF
--- a/src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
+++ b/src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
@@ -46,7 +46,7 @@ export const NavigationTabs: React.FC<NavigationTabsProps> = ({ partner }) => {
         name: "Works",
         href: route("/works"),
         exact: true,
-        hidden: !displayWorksSection || !filteredWorks.counts.total,
+        hidden: !displayWorksSection || !filteredWorks?.counts?.total,
       },
       {
         name: "Artists",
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
```diff
@@ -114,6 +117,12 @@ export const NavigationTabsFragmentContainer = createFragmentContainer(
       fragment NavigationTabs_partner on Partner {
         slug
         displayArtistsSection
+        displayWorksSection
+        filteredWorks: filterArtworksConnection(first: 0) {
```
sweir27, line 121: Instead of querying for filtered artworks here, you can instead use [this field](https://github.com/artsy/metaphysics/blob/master/src/schema/v2/partner.ts#L380), which will return the number of published artworks that the partner has.

(This is preferred because it will avoid making an extra call to our API).

ivan-badyulya: Fixed

sweir27:
Thanks!!

PR #7603, opened 2021-05-19
docs: Update relay test docs with note about QueryRenderer

damassi:
Add some light docs about how we can't test `QueryRenderer`s with our Relay test tools, but rather need to extract the contents of the `QueryRenderer` into a Relay fragment-like container and then test that. 

See the `SoldRecently` [component](https://github.com/artsy/force/blob/daad34183723be649e6031859842d65f4d902c21/src/v2/Apps/Consign/Routes/MarketingLanding/Components/SoldRecently.tsx#L143) and [test](https://github.com/artsy/force/blob/daad34183723be649e6031859842d65f4d902c21/src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/SoldRecently.jest.tsx) for an example of this pattern.

### DIFF
--- a/docs/best_practices.md
+++ b/docs/best_practices.md
@@ -110,10 +110,10 @@ Avoid aliasing imports:
 
 ```js
 // avoid
-import { FooFragmentContainer as Foo } from './Foo'
+import { FooFragmentContainer as Foo } from "./Foo"
 
 // good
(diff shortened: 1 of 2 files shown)


jonallured (approved):
This is wonderful - Future Jon thanks you!! 🤗

PR #7605, opened 2021-05-20
fix: use SystemQueryRenderer instead of QueryRenderer

kiryl-zubarau:
This PR replaces all QueryRenderer components with SystemQueryRenderer for the Partner Profile page. From time to time we had some render issues that I think were related to `QueryRenderer` that runs on the server side.

Pay attention to progress dots:
ezgif com-gif-maker (30)

### DIFF
--- a/src/v2/Apps/Partner/Components/Overview/ArtworksRail.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/ArtworksRail.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from "react"
 import { Box, BoxProps, Flex, Text } from "@artsy/palette"
-import { createFragmentContainer, graphql, QueryRenderer } from "react-relay"
+import { createFragmentContainer, graphql } from "react-relay"
 import { ArtworksRail_partner } from "v2/__generated__/ArtworksRail_partner.graphql"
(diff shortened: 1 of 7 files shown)


sweir27:
Confirming: I believe that `SystemQueryRenderer` is what we use if we want something to _only_ show on the client (and not the server).

That's the desired behavior in these cases, right? If so then I think this is 👍 with me! @damassi -- pinging you to make sure we're doing this right.

damassi:
This is indeed correct 👍 

I'll add a ticket for Danger rule today and look into an audit since this has pretty critical performance consequences if left unchecked.

damassi (approved):

kiryl-zubarau:
@damassi @sweir27 what do you think if we add a `placeholder?` field that will contain skeleton spinner to `SystemQueryRenderer` and return it instead of null if it is not empty? ([code](https://github.com/artsy/force/blob/4e7ef2ffdea27a672755cf235d9552818d01ca99/src/v2/Artsy/Relay/SystemQueryRenderer.tsx#L29))

I ask because for now `SystemQueryRenderer` returns nothing from the server, so the page is blinking like in the gif below.
ezgif com-gif-maker (31)

PR #7606, opened 2021-05-20
fix: Adjust margins on order flow screens

lilyfromseattle:
[PURCHASE-2682]

Recently the margins have changed in the web view of checkout. The content is far more squared and would like to use as much as the width of the phone screen. Confirmed with Arvind that the order stepper margins should be the same as the page content.

Pretty sure this regression was caused by [this PR](https://github.com/artsy/force/commit/4bea40764b1a9387a5a6514c8bfa8fd76015d72b)

_Before:_
Screen Shot 2021-05-20 at 12 17 32 PM

_After:_

Screen Shot 2021-05-20 at 10 36 38 AM

[PURCHASE-2682]: https://artsyproduct.atlassian.net/browse/PURCHASE-2682

### DIFF
--- a/src/v2/Apps/Order/Components/OrderStepper.tsx
+++ b/src/v2/Apps/Order/Components/OrderStepper.tsx
@@ -26,7 +26,7 @@ export function OrderStepper<Steps extends string[]>({
   return (
     <>
       <Media between={["xs", "md"]}>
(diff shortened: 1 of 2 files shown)


jo-rs (approved):
Super weird diff display. WTH Github? 🤔 

Nice! Merging 🚀

PR #7608, opened 2021-05-20
feat: [Artist2] Rebuild /artist page 

damassi:
The artist page was the first framework app we built with relay / palette / etc, and it's built up a lot of crust over the years. Now that we're migrating to the new design sys (which also has a simpler page layout) we can clean an enormous amount of code up. 

(Though there is gonna be some copy 🍝 from old app, notably `ArtistMeta` and the auction results filter.) 

This PR is the first pass at the rebuild and forgive the big pr 😓 

Follow-up:
- QA on Friday
- Test page with more artists
- Tests 

artist


src/v2/Apps/Artist2/Components/Artist2Meta.tsx
```diff
@@ -0,0 +1,319 @@
...
+import { data as sd } from "sharify"
+import { get } from "v2/Utils/get"
+import { Artist2MetaCanonicalLinkFragmentContainer as Artist2MetaCanonicalLink } from "./Artist2MetaCanonicalLink"
+
+interface Props {
+  artist: Artist2Meta_artist
+}
+
+// @ts-expect-error STRICT_NULL_CHECK
+type ArtworkNode = Artist2Meta_artist["artworks_connection"]["edges"][0]["node"]
```
dzucconi, line 16: `non-blocking/comment:` I realize kind of typing is (and should be) frowned upon but was wondering if there's a solution for doing an existential access of an array like this in types. (I think this probably shakes out to an `any` because of that)

damassi: (Yah this meta wa sall copy 🍝 -- stuff that seems crazy / old is likely crazy / old)

src/v2/Apps/Artist2/Routes/AuctionResults/Components/Artist2AuctionResultsCount.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface Artist2AuctionResultsCountProps {
+  results: Artist2AuctionResultsCount_results
+}
+
+export const Artist2AuctionResultsCount = ({
+  results,
+}: Artist2AuctionResultsCountProps) => {
+  return (
+    <Sans size="2" weight="medium">
```
dzucconi, line 14: 👋  `Sans`

src/v2/Apps/Artist2/Routes/AuctionResults/Components/AuctionFilterMobileActionSheet.tsx
```diff
@@ -0,0 +1,50 @@
...
+    <Container mt={6}>
+      <MobileTopBar>
+        <Button
+          variant="noOutline"
+          size="small"
+          onClick={() => filterContext.resetFilters()}
+        >
+          Reset
+        </Button>
+        <Sans size="3" weight="medium">
```
dzucconi, line 23: Dunno how much of this you wanna tackle now or after the fact but more `Sans`

src/v2/Apps/Artist2/Routes/AuctionResults/Components/AuctionFilters/MediumFilter.tsx
```diff
@@ -0,0 +1,51 @@
...
+              // @ts-expect-error STRICT_NULL_CHECK
+              selected: filterContext.filters.categories.includes(name),
+            }
+            return <Checkbox {...props}>{displayName}</Checkbox>
+          })}
+        </ShowMore>
+      </Flex>
+    </FilterExpandable>
+  )
+}
```
dzucconi, line 51: Why does this have it's own filter anyway?

src/v2/Components/ArtistInsight.tsx
```diff
@@ -52,20 +53,37 @@ export class ArtistInsight extends React.Component<ArtistInsightProps> {
     this.setState({ expanded: true })
   }
 
+  getTextByTheme() {
```
dzucconi, line 56: Never occurred to me we'd want to do these kind of switches in class components 😱

dzucconi:
This looks good, while acknowledging it's broadly un-reviewable :) — lets merge and follow up. At some point I'd recommend scoping a search for any deprecated components to uncover anything that's overlooked.

damassi:
Re your comments above -- all of that stuff was just copied over. I didn't rebuild everything, and specifically the auctions results (and other items in the description up above). I didn't want to mess with that in this pass since its more complex. In the future I think that can be rebuilt quickly, and kinda needs to as it uses a 3-generations-old state managment solution, etc.

PR #7614, opened 2021-05-24
fix: add redirection logic for works page

ivan-badyulya:
JIRA -> [Add redirection logic for works tab. If no artworks or CMS field for works tab display is false it should redirect to the overview page](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=36&projectKey=PX&modal=detail&selectedIssue=PX-4098&quickFilter=228)

### DIFF
--- a/src/v2/Apps/Partner/partnerRoutes.tsx
+++ b/src/v2/Apps/Partner/partnerRoutes.tsx
@@ -180,9 +180,38 @@ export const partnerRoutes: AppRouteConfig[] = [
           ) {
             partner(id: $partnerId) @principalField {
               ...Works_partner @arguments(input: $input)
+              displayWorksSection
+              counts {
+                eligibleArtworks
+              }
             }
           }
         `,
+        render: ({ Component, props, match }) => {
+          if (!(Component && props)) {
+            return
+          }
+
+          const { partner } = props as any
+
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Partner/partnerRoutes.tsx
```diff
@@ -180,9 +180,38 @@ export const partnerRoutes: AppRouteConfig[] = [
...
+              }
             }
           }
         `,
+        render: ({ Component, props, match }) => {
+          if (!(Component && props)) {
+            return
+          }
+
+          const { partner } = props as any
```
sweir27, line 195: Is it possible to get rid of this `any`?

Looks like you could also shorten this to `if (!props.partner) ...` which might avoid this as well.

ivan-badyulya: Screen Shot 2021-05-24 at 4 33 44 PM

This way wouldn't avoid the problem. The render props can't find somehow the partner field, that's why we decided avoiding the issue with `as any`

dzucconi: @damassi do you know if there's a way to provide types for the `render` function?

PR #7620, opened 2021-05-24
chore: Lock Node >=12.20.2

damassi:
Noticed some local failures as the node version installed was less than what was required from https://github.com/artsy/force/pull/7612. 

Bump to 12.20.2


dzucconi (approved):
🙏

PR #7622, opened 2021-05-25
feat: add placeholder for system query renderer

kiryl-zubarau:
This PR adds a placeholder for the system query renderer. The placeholder field uses for rendering skeleton spinners at the server time to prevent some content blinking. 

Before:
119349967-739a8980-bca7-11eb-9dc8-d1a6fe9212db

Now:
ezgif com-gif-maker (32)

### DIFF
--- a/src/v2/Apps/Partner/Components/Overview/ArtworksRail.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/ArtworksRail.tsx
@@ -125,6 +125,7 @@ export const ArtworksRailRenderer: React.FC<
           }
         }
       `}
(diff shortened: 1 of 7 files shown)


dzucconi (approved):

damassi:
The loading state (which is fixed in place) and then the animation in afterwards is a bit odd. Might be better to just kill the animation and fade in from the loading state? 

(Not a blocker however)

kiryl-zubarau:
@damassi Updated!

ezgif com-gif-maker (33)

damassi:
Much better! Thanks for the quick fix 👍

PR #7626, opened 2021-05-25
feat: 2FA respect after_update param.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3340

We are going to redirect partners from CMS to Force to setup their 2FA (if they are required to but haven't yet) like so:

`https://www.artsy.net/user/edit?after_update=https%3A%2F%2Fcms.artsy.net%3A%2Fauth%2Fartsy#two-factor-authentication`

After they finish setting up 2FA, we would like them redirected to the URL encoded in `after_update` param (back to `https://cms.artsy.net/auth/artsy`).

This PR adds a 2FA setup completion modal specifically for these users. On completion, they are informed that they will be redirected to `after_update` URL.

The URL is sanitized to prevent rogue redirects (such as to www.evilsite.com, ...)

Changes to `hokusai/development.yml` are to get `hokusai dev` working.

Screenshots of new modal:

Screenshot from 2021-05-25 16-42-09

Screenshot from 2021-05-25 16-44-06

### DIFF
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -5,6 +5,7 @@ services:
 {% include 'templates/docker-compose-service.yml.j2' %}
     environment:
       - NODE_ENV=development
+      - REDIS_URL=redis://force-redis:6379/0
       - OPENREDIS_URL=redis://force-redis:6379/0
       - PORT=5000
       - S3_KEY
--- a/hokusai/development.yml
+++ b/hokusai/development.yml
@@ -5,14 +5,21 @@ services:
 {% include 'templates/docker-compose-service.yml.j2' %}
     environment:
       - NODE_ENV=development
-      - REDIS_URL=redis://force-redis:6379/0
+      - OPENREDIS_URL=redis://force-redis:6379/0
       - PORT=5000
       - S3_KEY
       - S3_SECRET
       - DEPLOY_ENV
     env_file: ../.env
+    command: ["yarn", "start:dev"]
     ports:
       - 5000:5000
+    volumes:
+      - ../:/app
+      # prevents bind mount from overwriting node_modules in docker image.
+      - /app/node_modules
+    depends_on:
+      - force-redis
   force-redis:
     image: redis:3.2-alpine
     ports:
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
@@ -249,6 +249,7 @@ export const OnCompleteRedirModal: React.FC<OnCompleteRedirModalProps> = props =
       title="Set up with app"
       onClose={onClick}
       show={show}
+      hideCloseButton={true}
       FixedButton={
         <Button onClick={onClick} width="100%">
           OK
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
@@ -235,13 +235,13 @@ const InnerForm: React.FC<InnerFormProps> = ({
   )
 }
 
-interface OnCompleteRedirModalProps {
+interface OnCompleteRedirectModalProps {
   onClick: () => void
   redirectTo: string
   show: boolean
 }
 
-export const OnCompleteRedirModal: React.FC<OnCompleteRedirModalProps> = props => {
+export const OnCompleteRedirectModal: React.FC<OnCompleteRedirectModalProps> = props => {
   const { onClick, redirectTo, show } = props
 
   return (
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Flex, Input, Modal, Sans } from "@artsy/palette"
+import { Box, Button, Flex, Input, Modal, Sans, Serif } from "@artsy/palette"
 import { CreateAppSecondFactorMutationResponse } from "v2/__generated__/CreateAppSecondFactorMutation.graphql"
 import { useSystemContext } from "v2/Artsy"
 import { Formik, FormikHelpers as FormikActions, FormikProps } from "formik"
@@ -232,5 +232,35 @@ const InnerForm: React.FC<InnerFormProps> = ({
         </Button>
       </Flex>
     </Box>
+  )
+}
+
+interface OnCompleteRedirModalProps {
+  onClick: () => void
+  redirectTo: string
+  show: boolean
+}
+
+export const OnCompleteRedirModal: React.FC<OnCompleteRedirModalProps> = props => {
+  const { onClick, redirectTo, show } = props
+
+  return (
+    <Modal
+      title="Set up with app"
+      onClose={onClick}
+      show={show}
+      FixedButton={
+        <Button onClick={onClick} width="100%">
+          OK
+        </Button>
+      }
+    >
+      <Serif size="3t" color="black60">
+        You’ve successfully set up two-factor authentication!
+      </Serif>
+      <Serif mt={2} size="3t" color="black60">
+        You will be redirected to: {redirectTo}
+      </Serif>
+    </Modal>
   )
 }
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
@@ -1,4 +1,5 @@
 import { Box, Button, Flex, Input, Modal, Sans, Serif } from "@artsy/palette"
+import { getURLHost } from "v2/Utils/url"
 import { CreateAppSecondFactorMutationResponse } from "v2/__generated__/CreateAppSecondFactorMutation.graphql"
 import { useSystemContext } from "v2/Artsy"
 import { Formik, FormikHelpers as FormikActions, FormikProps } from "formik"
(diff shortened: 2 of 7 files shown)


src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
```diff
@@ -234,3 +234,34 @@ const InnerForm: React.FC<InnerFormProps> = ({
...
+        <Button onClick={onClick} width="100%">
+          OK
+        </Button>
+      }
+    >
+      <Serif size="3t" color="black60">
+        You’ve successfully set up two-factor authentication!
+      </Serif>
+      <Serif mt={2} size="3t" color="black60">
+        You will be redirected to: {redirectTo}
```
artsyjian, line 263: url gets wrapped which could be a hassle for users to read. how do we handle long lines in modals?

src/v2/Components/UserSettings/TwoFactorAuthentication/helpers.ts
```diff
@@ -0,0 +1,17 @@
+import { getClientParam } from "v2/Utils/getClientParam"
+import sanitizeRedirect from "@artsy/passport/sanitize-redirect"
```
artsyjian, line 2: borrowed from https://github.com/artsy/force/blob/652750420e1d698a86ba16fcdfde29d6a857e19d/src/desktop/apps/unsupported_browser/routes.coffee#L1.

artsyjian:
I would like to add tests to confirm the redirect works. Are there existing tests I can borrow from?

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
```diff
@@ -234,3 +234,34 @@ const InnerForm: React.FC<InnerFormProps> = ({
...
+      FixedButton={
+        <Button onClick={onClick} width="100%">
+          OK
+        </Button>
+      }
+    >
+      <Serif size="3t" color="black60">
+        You’ve successfully set up two-factor authentication!
+      </Serif>
+      <Serif mt={2} size="3t" color="black60">
```
damassi, line 262: We're fazing out `Sans / Serif` in favor of just `Text`, but since all of these sorts of items will eventually need some new designs I think this is fine for now.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
```diff
@@ -234,3 +234,34 @@ const InnerForm: React.FC<InnerFormProps> = ({
...
   )
 }
+
+interface OnCompleteRedirModalProps {
+  onClick: () => void
+  redirectTo: string
+  show: boolean
+}
+
+export const OnCompleteRedirModal: React.FC<OnCompleteRedirModalProps> = props => {
```
damassi, line 244: Redir -> Redirect

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
```diff
@@ -234,3 +234,34 @@ const InnerForm: React.FC<InnerFormProps> = ({
     </Box>
   )
 }
+
+interface OnCompleteRedirModalProps {
```
damassi, line 238: ditto

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
```diff
@@ -66,6 +80,14 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = props => {
     }
   }
 
+  async function onCompleteRedir() {
```
damassi, line 83: Since we're not doing anything with promises / asyncronously / `async / await` no need for `async` here

damassi (commented):
This is really nice work @artsyjian 💯 

Since there's so much new logic thoughts on adding some tests to cover it?

hokusai/development.yml
```diff
@@ -5,14 +5,21 @@ services:
 {% include 'templates/docker-compose-service.yml.j2' %}
     environment:
       - NODE_ENV=development
-      - REDIS_URL=redis://force-redis:6379/0
+      - OPENREDIS_URL=redis://force-redis:6379/0
```
joeyAghion, line 8: Weird... I actually don't see any references to the removed `REDIS_URL` in the codebase, but both staging and production provide it. Is there a chance libraries depend on it directly?

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
joeyAghion: I suggest rendering only the host here (e.g., `cms.artsy.net`). That's more meaningful to users than the `.../auth/artsy` path anyway, and avoids the wrapping concern. (Same for SMS.)

src/v2/Components/UserSettings/TwoFactorAuthentication/helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const afterUpdateRedir = () => {
+  const afterUpdateURL = getClientParam("after_update")
+
+  if (afterUpdateURL) {
+    const sanitizedURL = sanitizeRedirect(afterUpdateURL)
+
+    // a '/' means URL is bad
```
joeyAghion, line 10: Surprising... what if a caller was trying to redirect to the root page?

src/v2/Components/UserSettings/TwoFactorAuthentication/helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (sanitizedURL !== '/') {
+      return sanitizedURL
+    }
+  }
+
+  return ''
```
joeyAghion, line 16: If I'm following the logic correctly, returning blank here means the `if(redirectTo) ...` checks [above](https://github.com/artsy/force/pull/7626/files#diff-bd0d1fdf420315bc176a724649061dc306c49c7f18d9c2c2feee55904b53b7efR64) will be false, so the non-redirecting modal will display, right?

joeyAghion (approved):
Great!

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
artsyjian: done. thanks!

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
artsyjian: done.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
artsyjian: okay.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
artsyjian: fixed. thanks.

hokusai/development.yml
artsyjian: not sure. added it back just to be safe.

artsyjian:
> Since there's so much new logic, thoughts on adding some tests to cover it?
Thanks @damassi. Definitely should include tests. I will holler if I need help.

src/v2/Components/UserSettings/TwoFactorAuthentication/helpers.ts
artsyjian: yes, that's the intent, but..yeah, it's kinda confusing.

src/v2/Components/UserSettings/TwoFactorAuthentication/helpers.ts
artsyjian: not sure. shall test.

artsyjian: tested. if `after_update` points to `/`:

`http://localhost:5000/user/edit?after_update=%2F#two-factor-authentication`

[sanitize_redirect](https://github.com/artsy/force/blob/master/src/mobile/components/sanitize_redirect/index.coffee) returns `/` which causes `afterUpdateRedirect` to return empty string, so the old `Log back in` modal is displayed. I think this is desired behavior?

if `after_update` is a relative path other than `/`, say `/foo`:

`http://localhost:5000/user/edit?after_update=%2Ffoo#two-factor-authentication`

sanitize_redirect returns `/foo` which does cause the redirect modal to show which displays no target hostname (b/c `/foo` is relative), and clicking OK takes one to `localhost:5000/foo` and 404's. Ooo, that's bad.

i guess we should show the redirect modal only when `after_update` points to an absolute URL?

joeyAghion: This case is totally fine with the special `/` meaning. I was just surprised it operated that way generally for artsy-passport's other cases.

Ideally the new `after_update` param would accept either absolute or relative URLs, but you're right that the modal's copy would need to change in that case. In the case of relative redirects, I'd be fine if it just didn't say the "you'll be redirected..." bit.

artsyjian: revised. now, the presence of `after_update` param always cause the redirect modal to show. if the redir url is absolute, the modal shows `You will be redirected to: <host>`. if the url is relative (`/`, `/foo`, ...), the modal omits `You will be redirected...` and it will only say `You’ve successfully set up two-factor authentication!` and clicking OK will still redirect to that relative url.

PR #7084, opened 2021-02-25
feat(fx-2629): implements custom price range filter

dzucconi:
Closes: [FX-2629](https://artsyproduct.atlassian.net/browse/FX-2629)

* The currently selected price range is always reflected in the inputs (so if selecting a radio or coming in from a link)
* You can only input positive integers into the inputs, left hand side has to be at a minimum of 0, right hand side can be empty which is `"*"`
* If the input changes to one not reflected in the radios; then no radios will be highlighted

There might be some subtleties to this interaction here that can be tweaked; so this needs some eyes on it; but this should work as a first-pass.

PR #7629, opened 2021-05-26
test: introduce React Testing Library

anandaroop:
### Problem

In the 2021-05-25 Frontend Practice meeting I raised an (old) issue about some [difficulties I had](https://github.com/artsy/force/pull/7352/files#r612656495) with getting Jest/Enzyme tests written when working on updates to the artwork filters

I shared that when I have these kinds of difficulties, I sometimes write throwaway Cypress specs in order to get regression safety and test coverage around my changes. As with our other Cypress specs, these are straightforward to write and confidence-inspiring bc of their high fidelity — but also slower than unit tests and occasionally flaky due to relying on the staging environment

Weird but true: I write these specs, and then I **_toss_** them since we don't have a precedent for Cypress feature specs in Force. (We have basic [smoke tests](https://github.com/artsy/force/tree/652750420e1d698a86ba16fcdfde29d6a857e19d/cypress/integration) — that's another story).

So in the end I get confidence around the feature I'm writing 👍🏽 but the change gets committed without a test 👎🏽. This seems **not great**.

bob

### A solution?

@dzucconi mentioned that some people prefer **React Testing Library** (https://github.com/testing-library/react-testing-library)

I recall this coming up [here](https://artsy.slack.com/archives/C2TQ4PT8R/p1572971172013600?thread_ts=1572970740.013100&cid=C2TQ4PT8R) and [there](https://artsy.slack.com/archives/C2MEQ2HA6/p1615386986004200) and [everywhere](https://github.com/artsy/force/pull/7084#discussion_r583715560), and even being [discussed to death at one point](https://artsy.slack.com/archives/C2TQ4PT8R/p1615909369001100?thread_ts=1615909047.000500&cid=C2TQ4PT8R
). The RN version is also [in Eigen now](https://github.com/artsy/eigen/pull/3897). But I wasn't aware of it being trialled in Force, and I hadn't ever used it myself.

So that's what I'm up to here. I did a 1:1 rewrite of one test suite from Enzyme to RTL (https://github.com/artsy/force/commit/1ee059dcbc1382007bd48ef1b6b571f5d91c40a7), and then I retroactively added the test I struggled to write a few weeks ago (https://github.com/artsy/force/commit/1b67c5df3e0a448c85359d29aea628a1455d2d5e).

### Observations

tldr; I think the hype is justified — after an afternoon getting familiar with RTL, I found the tests easy to reason about and easy to write, for the most part.

- RTL shares with Cypress an orientation towards interacting with your UI as a user would — by referring to visible aspects of the interface and making it difficult-to-impossible for you to fiddle with component internals & implementation details

  - In the case of the artwork grid filters, this is actually a serious limitation, since a "unit test" on a filter does reasonably require peeking into the global artwork filter state — my solution to that was to create a new helper/inspector component that just blasts the global filter state to the UI as json text, so that we _can_ make ordinary UI assertions against it.

- RTL encourages an accessibility mindset by nudging you to find components by their [ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) — these are usually implicit in whatever DOM elements you are using, and it only takes a handful to get rolling (e.g. `checkbox` and `spinbutton` for the artwork filters).

- RTL wraps all renders and user events in `act()`, so your components are reliably ready for assertions — no faffing with `wrapper.update()` and `flushPromiseQueue()` 😌

- The resulting specs are slightly shorter, and very readable imo. If you squint it's often easy to make out the arrange-act-assert structure of a spec because _arrange_ is usually `render`, _act_ is usually an easily spotted series of `userEvent.foo` calls, and _assert_ is just the usual `expects`

- My experience, while good, was not _perfect_. I had some leaky state between tests, for which I came up with a goofy workaround. I'd hope that with some more eyes we could figure out the root cause of this: it involves our complex artwork filter state not being re-initialized between successive tests. Maybe an easy fix that I'm not seeing? 

- I didn't have to deal with Relay at all in this spec. That would be a good candidate for continuing this spike.

### Conclusion 

My feeling is that RTL might give us a good compromise between the speed of Jest unit tests, and the ease and confidence of user-centric Cypress e2e tests. 

The fact that my small rewrite already uncovered one serious gremlin  is a good reminder that there is no panacea when it comes to testing real-world apps, but I still think it's worth pursuing.  (I recognize that it would be a whole thing to figure out how to adopt pragmatically.) 

Finally I'd encourage anyone to pick their favorite (or anti-favorite) spec file in Force and go through the exercise of rewriting it as I did here. Steal commit https://github.com/artsy/force/commit/d92f44c8fcd292defe3c25f3118cf90cfe178ce3 to get a jumpstart.


src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const Wrapper: React.FC = ({ children }) => {
+  return (
+    <ArtworkFilterContextProvider>
+      <ClearAllButton />
+      {children}
+      <ArtworkFilterContextInspector />
+    </ArtworkFilterContextProvider>
+  )
+}
```
anandaroop, line 27: This pattern of overriding render with any custom setup/wrappers/providers is [recommended](https://testing-library.com/docs/react-testing-library/setup#custom-render).

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const ArtworkFilterContextInspector: React.FC = () => {
+  const artworkFilterContext = useArtworkFilterContext()
+
+  return (
+    <aside data-testid="artwork-filter-context-inspector">
+      {JSON.stringify(artworkFilterContext, null, 2)}
+    </aside>
+  )
+}
```
anandaroop, line 47: For a UI focused testing framework, how _do_ you make assertions about invisible application state?

My solution was to serialize it right into the UI of the test harness. If you use RTL's `screen.debug()` you see this:

debug

This allows the test to "see" the same state that the user indirectly would.

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+  try {
+    contextInspector = screen.getByTestId("artwork-filter-context-inspector")
+  } catch (error) {
+    if (error.name === "TestingLibraryElementError")
+      throw new Error(
+        `The currentContext() helper function requires an <ArtworkFilterContextInspector /> to be mounted in the current DOM.`
+      )
+  }
+  return JSON.parse(contextInspector!.textContent!)
+}
```
anandaroop, line 60: Once it's serialized to the DOM you can de-serialize it and make assertions against it, as seen below.

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+describe("SizeFilter2", () => {
+  it("toggles custom input render", () => {
+    render(<SizeFilter2 expanded />)
+
+    // note: an <input type="number" /> has a role of 'spinbutton'
+    expect(screen.queryByRole("spinbutton")).not.toBeInTheDocument()
```
anandaroop, line 67: The inputs in this UI use `<input type="number" />`:



TIL these have an aria role of `spinbutton`, so that's the RTL way of finding these elements.

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+    expect(currentContext().filters?.sizes).toEqual([])
+    expect(currentContext().filters?.height).toEqual("4.72-6.3")
+    expect(currentContext().filters?.width).toEqual("4.72-6.3")
+  })
+
+  it("updates the filter values when only one dimension is added", () => {
+    render(<SizeFilter2 expanded />)
+    userEvent.click(screen.getByText("Show custom size"))
+    screen.getAllByRole("spinbutton").map(field => userEvent.clear(field)) // 😭
```
anandaroop, line 104: Here and L127 you can see the one goofy workaround I needed in order to get around the problem of state leaking from one test to another — somehow the input fields retain the values that are "typed" in from earlier specs, so I manually cleared them 😕

pepopowitz:
This is **amazing** @anandaroop. Not to give you more work but it'd make a great blog post 😄

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
damassi: @dzucconi - didn't you do something similar to this in a hook test at some point? Context around that? I recall that it was an unpleasant work around...

dzucconi: Yeah, [I've been doing this exact thing](https://github.com/artsy/eigen/blob/1cf8cebc3e2ec232197e5f1c512c42cc7bc4d2f5/src/lib/Components/ArtworkFilter/Filters/__tests__/SizeOptionsNew-tests.tsx#L29-L63). Feels kinda wrong but works well 👍

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter2.rtl.jest.tsx
```diff
@@ -0,0 +1,171 @@
...
+  it("toggles custom input render", () => {
+    render(<SizeFilter2 expanded />)
+
+    // note: an <input type="number" /> has a role of 'spinbutton'
+    expect(screen.queryByRole("spinbutton")).not.toBeInTheDocument()
+
+    userEvent.click(screen.getByText("Show custom size"))
+
+    expect(screen.getByText("Hide custom size")).toBeInTheDocument()
+    expect(screen.getAllByRole("spinbutton")).toHaveLength(4)
```
damassi, line 72: Can one search by component name? Like this `role` search feels a little more indirect versus `getByRole("SomeButton")`

dzucconi: Check the query functions here: https://testing-library.com/docs/queries/about

Finding by component type is really brittle — I've broken a ton of specs due to people reaching into Palette components and others this way. `getByTestId` is what we'd probably reach for most of the time.

damassi: Ah i see, that pattern also seems to break the spirit of the lib

anandaroop: Yeah exactly about the sprit of of the lib — have your tests view your app the way a user would, in terms of affordances like `checkbox` and `textbox`

I had a similar reaction at first, like why can't I just query by `div` or `SomeButton `, but here is why I think this is a better approach:

- `SomeButton` (and for that matter `div`) is an implementation detail, and requires you and other readers of your test to know something about your implementation 

- It exploits knowledge (i.e. component names) that is very specific to your app. (Contrast this with learning what `spinbutton` means, a little bit of jargon that you can now generalize to **all** tests)

- It inhibits pure refactors because of the coupling between your tests and your implementation

There are times when you do legit need to know more about the implementation, even in this PR, and that's what the `getByTestId` escape hatch is for, as Damon mentioned.

damassi (commented):
This looks great @anandaroop, and thank your for applying some investigation to a non-trivial example 👍 

We should add a test for a Relay component by updating / duplicating [this testing lib](https://github.com/artsy/force/blob/master/src/v2/DevTools/setupTestWrapper.tsx#L85), see how it feels, and then we should make a call. Since this subject keeps coming up I'm certainly down to give it a try but we'll want to be sure that everything is all good with our Relay layer first. 

Also note, that should we want to go forward with this there's this [line from Eigen](https://github.com/artsy/eigen/blob/master/dangerfile.ts#L34) we can steal for our danger file to help guide folks in the right direction.

jest.config.v2.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
       },
       reporters: ["default", "jest-junit"],
       setupFilesAfterEnv: ["<rootDir>/src/v2/jest.envSetup.ts"],
+      testEnvironment: "jest-environment-jsdom-thirteen",
```
anandaroop, line 17: TODO: add context for this change

anandaroop: [Added](https://github.com/artsy/force/pull/7629/commits/ec911988e865f22a0f6bb03729e1bde61f59d496#diff-8aa60dd701d62e5679b59339faf9418a8b5764591cca2779bdcf5998884303e3) to Hacks.md

damassi: Thanks for adding to hacks 👍 Since its easy, we still might want to add a `// comment` here in the file too

anandaroop:
Think I've tackled everything we mentioned in the Web Practice meeting:

- documented the JSDom hack & fixed accompanying regressions
- extracted the `findMetaTagBySelector` helper
- promoted the two new RTL specs to replace their Enzyme counterparts

This can be followed up with 
- updates to the example app to share best practices (See [WP-67](https://artsyproduct.atlassian.net/browse/WP-67))
- an announcement that React Testing Library is available in Force
- some more education / L&L type stuff
- (eventually) a new [linter warning](https://github.com/artsy/force/blob/ad3b729ed37df213a5fd95eb59a80792249a1ab4/.eslintrc.js#L84-L95) about Enzyme

cc folks from last practice mtg @laurabeth @pepopowitz @annacarey @dzucconi

damassi:
Awesome work @anandaroop and thanks for taking this over the line. Mergin' 💯

PR #7642, opened 2021-05-27
fix(PX-4015): use partnerArtist artwork connection for loading artist's artworks in the right order

kiryl-zubarau:
This PR uses artworkConnection from the partnerArtist type for loading the artist's artworks in the right order.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4015

ezgif com-gif-maker (34)


sweir27:
😍 Nice!!

PR #7644, opened 2021-05-27
Created Banner component with Sign up and Log in buttons for m…

anastasiapyzhik:
…obile

**Before**
Screen Shot 2021-05-27 at 5 41 34 PM

**After**
Screen Shot 2021-05-27 at 5 42 15 PM


src/v2/Components/LoginSignUpBanner/Banner.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface Props {}
+
+const Container = styled(Flex)`
+  width: 100%;
+  padding: 12px;
+  background-color: ${color("black100")};
+`
+
+export class Banner extends React.Component<Props> {
```
damassi, line 14: Can we use React funcction components instead of classes? (All class use should be considered depreciated, since we can't combine classes and react hooks)

damassi (commented):
This looks good 👍  One request about swapping a class component for a react function component however

src/v2/Components/LoginSignUpBanner/Button.tsx
```diff
@@ -0,0 +1,34 @@
...
+})`
+  width: 50%;
+  height: 40px;
+  text-align: center;
+  background-color: ${color("black80")};
+  color: ${color("white100")};
+  font-family: "ll-unica77", "Helvetica Neue", Helvetica, Arial, sans-serif;
+  font-weight: bold;
+  font-size: 15px;
+`
```
dzucconi, line 22: It seems to me that we should just adapt our `Button` component to support `as` — we should never be setting font-family/font-size etc directly.

anastasiapyzhik: What do you mean by that? Could you give some example?

dzucconi: https://github.com/artsy/palette/pull/946

So we can pass `as` with either a string name of an element or an actual component and the component will internally use that as the basis while keeping all the styling. So once this is released we can just do `<Button as={RouterLink} to="/log_in">` for example.

```tsx
<Flex p={1}>
  <Button flex={1} as={RouterLink} to="/sign>Sign up</Button>

  <Spacer ml={1} />

  <Button flex={1}>Log in</Button>
</Flex>

<Separator />
```

Chatting about the color scheme inconsistencies here: https://artsy.slack.com/archives/C9XJKPY9W/p1623078390072600

dzucconi: I've got a PR coming for this as I need to move it anyway for an unrelated reason!

anastasiapyzhik: Thank you for clarifying! So as I understand you are fixed this? Or I should do it?

dzucconi: All good now!

PR #7645, opened 2021-05-27
feature: Webpack Chunking

icirellik:
This PR modifies our Webpack bundling strategy to follow several common patterns used elsewhere in the industry.

Creates a minimal chunk for each App
Creates a minimal commons chunk for each App
Create a framework chunk for `@artsy` framework code to ensure that it will not be cache-invalidated
Create a framework chunk for core framework code to ensure that it will not be cache-invalidated
Creates library chunks for any large node_module dependencies
Creates shard chunks for any code used by multiple pages

Additional Optimizations

* Reduces the number of low KiB bundles by using naming common preloaded chunks
* Restricts bundles to 240 KiB unzipped
* Cleans up legacy code that was accidentally imported by the orders page

Notable Impacts:

Reduces the Artist page JS bundle payload by ~275 KiB compressed

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "test:ci:mocha": "NODE_ENV=test yarn mocha src/**/*.test.js src/**/*.test.ts",
     "type-check": "tsc",
     "unlink-all": "yalc remove --all && yarn --check-files",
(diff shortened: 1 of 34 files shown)


damassi:
Nice! 👍

PR #7656, opened 2021-05-28
Add android download mobile footer

anastasiapyzhik:
I united 2 subtasks in 1 PR because they use the same logic of device detecting.

https://artsyproduct.atlassian.net/browse/GRO-341

**Before**
Screen Shot 2021-05-28 at 3 03 10 PM

**After**
Screen Shot 2021-05-28 at 3 04 44 PM

Screen Shot 2021-05-28 at 3 05 16 PM

https://artsyproduct.atlassian.net/browse/GRO-343

**Before**
Screen Shot 2021-05-28 at 5 35 23 PM

**After**
Screen Shot 2021-05-28 at 5 36 00 PM


damassi (approved):
👌 👌

src/v2/Utils/Hooks/useDeviceDetection.tsx
```diff
@@ -0,0 +1,29 @@
...
+    } else if (window.navigator.userAgent.match(/iPhone/)) {
+      setDevice(Device.iPhone)
+      setDownloadAppUrl(
+        "https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080"
+      )
+    }
+  }, [])
+
+  return { device, downloadAppUrl }
+}
```
dzucconi, line 29: We should probably introduce a default state here — this introduces a page shift since it only returns something useful on client-load

PR #7633, opened 2021-05-26
fix: padding and line break on buyer guarantee page

mc-jones:
This PR: 

1. Adds padding to The Artsy Guarantee table on Mobile and Desktop
Screen Shot 2021-05-26 at 11 39 53 AM

2. Creates a line break for the Money-Back Guarantee Passage
Screen Shot 2021-05-26 at 11 40 27 AM

Jira Ticket: [PURCHASE-2699]

[PURCHASE-2699]: https://artsyproduct.atlassian.net/browse/PURCHASE-2699

PR #7660, opened 2021-05-28
fix: add double bang to conditional render

mc-jones:
This PR adds a double bang in the condition render as a correction to this PR https://github.com/artsy/force/pull/7633.


sepans (approved):
Thank you 🙏🏽 🚀  
Sorry I missed this. Good to merge.

PR #7611, opened 2021-05-21
feat(PX-4058): partner profile page transition

ivan-badyulya:
### This PR is just a migrating plan from partner2 to partner. We'll start to implement it after its approval. Please take a look into.

- [X] **1. New partner profile page** https://github.com/artsy/force/pull/7682
  - replace /partner2 with /partner

- [X] **2. Add redirects for the old routes** https://github.com/artsy/force/pull/7677

  We have to implement a middleware to manage an overlapping redirection logic to the new routes in these files: [partner2 routes](https://github.com/artsy/force/blob/master/src/desktop/apps/partner2/index.coffee#L8) and [partner](https://github.com/artsy/force/blob/master/src/desktop/apps/partner/index.coffee#L8) routes.

  The current routes desktop should be redirected in accordance with the next schema:

/:id                              ->  /partner/:id
/:id/overview              ->  /partner/:id/overview -> /partner/:id/       
/:id/contact                ->  /partner/:id/contact
/:id/works                   ->  /partner/:id/works
/:id/shows                  ->  /partner/:id/shows
/:id/artists                  ->  /partner/:id/artists
/:id/artist/:artistId      ->  /partner/:id/artists/:artistId
/:id/articles                ->  /partner/:id/articles
/:id/about                   ->  /partner/:id
/:id/collection            ->  /partner/:id/works
/:id/shop                    ->  /partner/:id/works
/:id/article/:articleId  ->  _It is shouldn't be redirected. Should be used the current version of the page_

  [Mobile routes](https://github.com/artsy/force/blob/master/src/mobile/apps/partner_profile/index.coffee#L12):

/:profileId                            ->  /partner/:id/
/:profileId/overview            ->  /partner/:id/overview -> /partner/:id/ 
/:profileId/artists                 ->  /partner/:id/artists
/:profileId/artist/:artistId     ->  /partner/:id/artists/:artistId
/:profileId/contact               ->  /partner/:id/contact
/:profileId/articles               ->  /partner/:id/articles
/:profileId/shows                 ->  /partner/:id/shows
/:profileId/collection            ->  /partner/:id/works
/:profileId/shop                    ->  /partner/:id/works
/:profileId/article/:articleId -> _It is shouldn't be redirected. Should be used the current version of the page_
 
- [ ] **3. Metaphisics** https://github.com/artsy/metaphysics/pull/3193
   - update href field for Partner, Profile and maybe https://github.com/artsy/metaphysics/blob/master/src/lib/routing.ts

[...]

PR #7677, opened 2021-06-02
feat(PX-4058): partner transition part 2

kiryl-zubarau:
**Do not merge this PR until the first part will be merged https://github.com/artsy/force/pull/7682**

The second step of the partner profile page transition. https://github.com/artsy/force/pull/7611
JIRA - https://artsyproduct.atlassian.net/browse/PX-4058


ivan-badyulya (approved):

cypress/integration/partnerProfile.spec.js
```diff
@@ -20,48 +20,31 @@ describe("/:partner_id", () => {
...
-    it("shows the list of shows", () => {
-      cy.visit("gagosian-gallery/shows")
-      cy.contains("Current Shows")
-    })
-    it("shows partner articles", () => {
-      cy.visit("gagosian-gallery/articles")
-      cy.wait(2000)
-      cy.contains("Articles")
-    })
+  it.skip("does not show contact information for non-active partner", () => {
```
sweir27, line 38: Why is this one `skip`ped?

src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts
```diff
@@ -0,0 +1,34 @@
...
+  next: NextFunction
+) => {
+  if (!res.locals?.profile?.isPartner() || !res.redirect) return next()
+
+  const href: string | null = `/partner${req.path}`
+
+  // /:id/overview -> /partner/:id
+  // /:id/about -> /partner/:id
+  if (req.route.path === "/:id/overview" || req.route.path === "/:id/about") {
+    return res.redirect(301, `/partner/${req.params.id}`)
```
sweir27, line 16: I think that `req.params.id` here will be the _profile_ id, right?

If that's the case, this won't necessarily be correct. We'll have to use the partner ID/slug instead.

src/desktop/apps/partner/index.coffee
```diff
@@ -5,15 +5,15 @@ app = module.exports = express()
...
-app.get '/:id/contact', routes.requirePartner, routes.contact
-app.get '/:id/about', routes.requirePartner, routes.about
-app.get '/:id/collection', routes.requirePartner, routes.collection
-app.get '/:id/shop', routes.requirePartner, routes.shop
-app.get '/:id/works', routes.requirePartner, routes.works
-app.get '/:id/shows', routes.requirePartner, routes.shows
-app.get '/:id/artists', routes.requirePartner, routes.artists
-app.get '/:id/artist/:artistId', routes.requirePartner, routes.artist
-app.get '/:id/articles', routes.requirePartner, routes.articles
+# app.get '/:id', routes.requirePartner, routes.overview
```
sweir27, line 8: We should add a comment to explain why all of these are commented out

src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts
sweir27: We'll need to do something [similar to this](https://github.com/artsy/force/blob/master/src/desktop/apps/fair_redirect/fairRedirectionMiddleware.ts#L13).

sweir27 (changes requested):
Looking good but we'll need to handle the fact that profile IDs don't always equal partner IDs! Let me know if I can clarify further.

cypress/integration/partnerProfile.spec.js
kiryl-zubarau: To tell the truth, I don't know, this test was already skipped when I started working on this file.

src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts
kiryl-zubarau: Updated

src/desktop/apps/partner/index.coffee
kiryl-zubarau: What if I remove them completely? I mean I don't see the sense to comment them out, because it wouldn't impact the code.

src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts
kiryl-zubarau: Should we update the links that we updated at part 4 https://github.com/artsy/force/pull/7685 to use partner `id` instead of `default_profile_id`?

kiryl-zubarau: I also think that we should update metaphysics to also use `slug` instead of `default_profile_id` at `href` field https://github.com/artsy/metaphysics/pull/3193

cypress/integration/partnerProfile.spec.js
sweir27: Ah-- I see below. Looks like we never chose a partner to use for this case.

src/desktop/apps/partner/index.coffee
sweir27: I think it's fine to remove! I'd still add a little comment (maybe a `# TODO`?) to explain why we just have the articles path left. 

Something like `# TODO: Remove once we've rebuilt the partner article page in the new framework`

kiryl-zubarau: Done

src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts
kiryl-zubarau: @sweir27 ?

PR #7684, opened 2021-06-02
fix: `Set price` and `Set size` are made as links instead of buttons

anastasiapyzhik:
…ore' and 'Set price'

https://artsyproduct.atlassian.net/browse/FX-2846

**Before**
Screen Shot 2021-06-02 at 4 38 25 PM

**After**
Screen Shot 2021-06-02 at 4 38 44 PM


dblandin:
As discussed [in Jira](https://artsyproduct.atlassian.net/browse/FX-2846), we're going to keep the existing design of the buttons for now instead of swapping for links.

Sorry for the change of direction @anastasiapyzhik!

PR #7691, opened 2021-06-03
fix:Remove login/signup gate on viewing rooms

alexj105:
Remove  auth modal window in the Viewing room


damassi:
Hi @alexj105, is there a JIRA ticket associated with this ticket?

damassi (commented):
@alexj105, is there a JIRA ticket associated with this PR?

joeyAghion:
I believe this change will require matching updates to [the integration tests](https://github.com/artsy/integrity) that assert about the previous modals.

PR #6969, opened 2021-02-03
feat: Redirect some auth paths

jonallured:
This PR alters force to redirect like so:

* `/log_in` => `/login`
* `/sign_up` => `/signup`

This change is motivated by some SEO guidance but it's also been an opportunity to do some refactoring which is why this PR is 12 commits, haha.

## Forgetting What We Already Knew

While reading through the auth code to make this change, I found a familiar anti-pattern that I refer to as "forgetting what we already knew". Check out this routing code:

```ts
app.get("/forgot", authHandler)
app.get("/login", authHandler)
app.get("/signup", authHandler)

const authHandler = (req, res, next) = > {
  switch (req.path) {
    case "/forgot":
      type = ModalType.forgot
      break
    case "/login":
      type = ModalType.login
      break
    case "/signup":
      type = ModalType.signup
      break
  }

  // lots more code
}
```

This code starts by telling Express that we want to handle three paths and sends them each to a single handler. That handler then examines the path on the request to know which type we're working with. This is bananas! We already knew what type of auth we were dealing with when Express picked up the route and called the handler but then we forgot it by routing them all through a single callback!

The refactor here is to extract a handler for each path and DECLARE the values that are specific to that path and then call shared code. This allows us to remove conditionals that are specific to a particular path. Removing ifs and case statements is a great way to reduce the complexity of a piece of code.

## Finding Seams When SRP is Violated

There's another, bigger problem with the code I found and it's a bit harder to see at first. When I started reading the code that handled the three requests I was overwhelmed. What was signal and what was noise? This function was violating the Single Responsibility Principle, that much was clear, but how could I improve it?

I turned to the tests hoping that they could help unravel what was going on. What I saw there were a bunch of tests asserting about the arguments being sent to a particular function call. We used jest to mock calls to [stitch][] and then assert about the various intricacies of the object being sent to it.

[stitch]: https://github.com/artsy/stitch

Ah ha! The true responsibility of the handler function is to call `stitch` and we're very concerned with the shape of the arguments. This allowed me to see that there was a missing seam - we could extract a function that computes the options being sent to `stitch` and then reduce the noise in our handlers by encapsulating all the complexity of request parsing, etc in this compute function.

## When to Stop Refactoring

I could have done so much more refactoring but I felt this was a good place to stop because it was a coherent piece of work. I wanted to make a change so I refactored to make that change easier and then I made that change. Beck would be so proud of me!

[...]

PR #7695, opened 2021-06-03
Preserve querystrings when redirecting to /login or /signup paths (PLATFORM-3341)

joeyAghion:
Back in https://github.com/artsy/force/pull/6969, the `/log_in` path started being redirected to `/login` for consistency. I believe this broke the intended behavior where logging in would redirect the user back to a specified URL.

In this PR, I try to update all the locations I could find that direct a visitor to `/log_in` to instead send them to `/login` directly, saving a redirect step. For good measure (and in case there are links to `/log_in?...` out there in our apps or the wild), I also update that internal redirect to pass the querystring through. When I updated links that used legacy querystring parameters like `redirect_uri`, I also switched them to `redirectTo` which seems to be our [current preference](https://github.com/artsy/force/blob/62646da8d4b62943566bba6e4fcc4313606a61f2/src/desktop/apps/authentication/routeHelpers.ts#L114).

https://artsyproduct.atlassian.net/browse/PLATFORM-3341

### DIFF
--- a/src/desktop/apps/authentication/index.ts
+++ b/src/desktop/apps/authentication/index.ts
@@ -1,4 +1,5 @@
 import express from "express"
+import { parse } from "url"
 import {
   forgotPassword,
   login,
@@ -17,9 +18,11 @@ app.get("/signup", redirectLoggedInHome, signup)
 app.get("/forgot", forgotPassword)
 app.get("/reset_password", resetPassword)
 
-app.get("/log_in", (_req, res) => {
-  res.redirect(301, "/login")
+app.get("/log_in", (req, res) => {
+  const parsedUrl = parse(req.url)
+  res.redirect(301, `/login?${parsedUrl.query}`)
 })
-app.get("/sign_up", (_req, res) => {
-  res.redirect(301, "/signup")
+app.get("/sign_up", (req, res) => {
+  const parsedUrl = parse(req.url)
+  res.redirect(301, `/signup?${parsedUrl.query}`)
 })
--- a/src/desktop/apps/auction_support/routes.coffee
+++ b/src/desktop/apps/auction_support/routes.coffee
@@ -38,7 +38,7 @@ registerAndRedirect = (sale, req, res, next) ->
 
 @modalAuctionRegistration = (req, res, next) ->
   unless req.user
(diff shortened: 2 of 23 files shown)


src/desktop/apps/authentication/index.ts
```diff
@@ -17,9 +18,11 @@ app.get("/signup", redirectLoggedInHome, signup)
...
-  res.redirect(301, "/login")
+app.get("/log_in", (req, res) => {
+  const parsedUrl = parse(req.url)
+  res.redirect(301, `/login?${parsedUrl.query}`)
 })
-app.get("/sign_up", (_req, res) => {
-  res.redirect(301, "/signup")
+app.get("/sign_up", (req, res) => {
+  const parsedUrl = parse(req.url)
+  res.redirect(301, `/signup?${parsedUrl.query}`)
```
joeyAghion, line 27: Is there a better way to accomplish this?

joeyAghion:
Another bug this PR attempts to fix: some redirect URLs were not properly encoded when passed in querystrings. E.g.:
* visiting `/user/edit?a=b&c=d`
* would redirect to `/login?redirectTo=/user/edit?a=b&c=d`
* which would eventually redirect to `/user/edit?a=b` rather than the correct `/user/edit?a=b&c=d`.

src/desktop/apps/authentication/index.ts
damassi: A better way to redirect? Or a better way to get query params? If the latter, should be able to do [`req.query.someParam`](https://coderrocketfuel.com/article/handle-get-request-query-string-parameters-in-express-js), but this also works fine since your translating it back into a string.

damassi (approved):
This LGTM 👍 with the sole caveat that it might be worth digging into [artsy-passport](https://github.com/artsy/artsy-passport/search?q=login) as well.

src/desktop/apps/authentication/index.ts
joeyAghion: I meant a better way to get the full querystring.

joeyAghion:
Force appears to configure artsy-passport with `loginPagePath` and `signupPagePath` _without_ underscores already.

Also did a quick search through Integrity and didn't find assertions about `log_in` or `sign_up` paths, though that surprises me a bit.

PR #7697, opened 2021-06-03
fix: Lower auction count on first load to 10

damassi:
We were noticing some performance problems around there being too many items on the dom. This lowers things to 10, and the user can click load more.


src/v2/Apps/Auctions/Routes/UpcomingAuctions.tsx
```diff
@@ -19,6 +19,7 @@ const UpcomingAuctions: React.FC<UpcomingAuctionsProps> = ({
   viewer,
   relay,
 }) => {
+  console.log("hiik")
```
dzucconi, line 22: 👋

PR #7707, opened 2021-06-04
feat(mWeb): scroll filters into view when expanded

anastasiapyzhik:
…k height for calculating scrolling offset

https://artsyproduct.atlassian.net/browse/FX-2832

**Before**
before

**After**
after

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
@@ -108,7 +108,7 @@ export const ArtworkFilterMobileActionSheet: React.FC<{
       </Header>
 
       <Content ref={contentRef as any}>
-        <Box width="100%" p={2}>
+        <Box width="100%" p={2} pb={0}>
           {children}
         </Box>
       </Content>
--- a/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
@@ -108,7 +108,7 @@ export const ArtworkFilterMobileActionSheet: React.FC<{
       </Header>
 
       <Content ref={contentRef as any}>
-        <Box width="100%" p={2} pb={0}>
+        <Box width="100%" px={2} pt={2}>
           {children}
         </Box>
       </Content>
--- a/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
@@ -88,9 +88,9 @@ export const ArtworkFilterMobileActionSheet: React.FC<{
         {/* TODO: This extraneous Flex is not necessary, Clickable (and Box) should have Flex props*/}
         <Flex flex="1">
           <Clickable width="100%" onClick={handleScrollToTop}>
-            <Title size="3" weight="medium" textAlign="center">
+            <FilterTitle size="3" weight="medium" textAlign="center">
               Filter
-            </Title>
+            </FilterTitle>
           </Clickable>
         </Flex>
 
@@ -167,6 +167,8 @@ const Content = styled(Flex)`
   -webkit-overflow-scrolling: touch;
 `
 
-const Title = styled(Sans)`
+const FilterTitle = styled(Sans)`
   flex: 1;
 `
+
+FilterTitle.displayName = "FilterTitle"
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
@@ -2,16 +2,22 @@ import {
   Box,
   Expandable,
   ExpandableProps,
+  themeProps,
   useThemeConfig,
 } from "@artsy/palette"
 import React, { useEffect, useRef, useState } from "react"
-import { Media } from "v2/Utils/Responsive"
+import { useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
 import { getElementParams } from "./helpers"
 import { useScrollRefContext } from "./useScrollContext"
 
-const FilterExpandableWithScrollIntoView: React.FC<
-  ExpandableProps & { tokens: { mb: number } }
-> = ({ tokens, ...props }) => {
+export const FilterExpandable: React.FC<ExpandableProps> = props => {
+  const tokens = useThemeConfig({
+    v2: { mb: 1 },
+    v3: { mb: 6 },
+  })
+
+  const isScrollIntoViewEnabled = useMatchMedia(themeProps.mediaQueries.xs)
+
   const ctx = useScrollRefContext()
   const scrollRef = ctx?.scrollRef
 
@@ -34,10 +40,10 @@ const FilterExpandableWithScrollIntoView: React.FC<
         anchorRef.current?.scrollIntoView({ block: "end" })
       }
     }
-  }, [open, filterRef])
+  }, [open])
 
   const onClick = () => {
-    setOpen(open => !open)
+    isScrollIntoViewEnabled && setOpen(open => !open)
   }
 
   return (
@@ -45,23 +51,5 @@ const FilterExpandableWithScrollIntoView: React.FC<
       <Expandable mb={tokens.mb} onClick={onClick} {...props} />
       <div ref={anchorRef}></div>
     </Box>
-  )
-}
-
-export const FilterExpandable: React.FC<ExpandableProps> = props => {
-  const tokens = useThemeConfig({
-    v2: { mb: 1 },
-    v3: { mb: 6 },
-  })
-
-  return (
-    <>
-      <Media at="xs">
-        <FilterExpandableWithScrollIntoView tokens={tokens} {...props} />
-      </Media>
-      <Media greaterThan="xs">
-        <Expandable mb={tokens.mb} {...props} />
-      </Media>
-    </>
   )
 }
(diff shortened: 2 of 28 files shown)


src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
```diff
@@ -14,30 +14,36 @@ import { ArtworkLocationFilter } from "./ArtworkLocationFilter"
...
-      {showNewFilters ? (
-        <PartnersFilter />
-      ) : (
-        <>
-          <GalleryFilter />
-          <InstitutionFilter />
-        </>
-      )}
-    </>
+    <Box ref={scrollRef} overflowY="scroll" height="100%">
```
anastasiapyzhik, line 26: This `Box` is not supposed to be on desktop, so far I'm just trying to make it work correctly on mobile

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,42 @@
...
+
+  const [open, setOpen] = useState(!!props.expanded)
+
+  useEffect(() => {
+    if (open) {
+      // @ts-expect-error STRICT_NULL_CHECK
+      const height = ref.current.clientHeight
+
+      scrollRef.current.scrollTo({
+        top: height,
```
anastasiapyzhik, line 28: That's not gonna work properly, I only made it scroll on click and show the height of expanded block. I have to figure out which equation is correct. Any hint would be highly appreciated

dzucconi: Do you want its `top` via `getClientRect`?

dzucconi:
My first question would be what about `scrollIntoView` is "clumsy and inconsistent"? — we could always offset it in various ways. Maybe posting a screen capture would help me understand? I only bring this up again because it would allow us to avoid all of this quite nicely.

That said — since this really just a one-off I'm not opposed to just adding a clearly named `.js-whatever` selector and selecting the DOM node that way. My hope being that we wouldn't need to do this more than once, since the hook only accommodates a single ref anyway.

(Previously I had used this pattern in a few places so abstracted it with a map: https://github.com/auspices/gaea/blob/master/src/hooks/useContextualRef.tsx so that multiple refs could be accessed this way.)

damassi:
Here's an example of adding a dom node `id` and scrolling to it that way: https://github.com/artsy/force/blob/master/src/v2/Apps/Artist2/Artist2App.tsx#L43

```js
scrollIntoView(id, offset)
``` 

It's not the most intelligent implementation, but it gets the job really simply.

anastasiapyzhik:
> Here's an example of adding a dom node `id` and scrolling to it that way: https://github.com/artsy/force/blob/master/src/v2/Apps/Artist2/Artist2App.tsx#L43
> 
> ```js
> scrollIntoView(selector, offset)
> ```
> 
> It's not the most intelligent implementation, but it gets the job done really simply.

I tried to use custom `scrollIntoView` but it didn't work for me. You can look through this discussion:
https://artsy.slack.com/archives/C9SATFLUU/p1622737727149200

anastasiapyzhik:
> My first question would be what about `scrollIntoView` is "clumsy and inconsistent"? — we could always offset it in various ways. Maybe posting a screen capture would help me understand? I only bring this up again because it would allow us to avoid all of this quite nicely.
> 
> That said — since this really just a one-off I'm not opposed to just adding a clearly named `.js-whatever` selector and selecting the DOM node that way. My hope being that we wouldn't need to do this more than once, since the hook only accommodates a single ref anyway.
> 
> (Previously I had used this pattern in a few places so abstracted it with a map: https://github.com/auspices/gaea/blob/master/src/hooks/useContextualRef.tsx so that multiple refs could be accessed this way.)

"clumsy and inconsistent" - this was about native `scrollIntoView`. As I know we can't pass offset to it, we can only pass values `start/center/end/nearest` to `block` prop

dzucconi:
> As I know we can't pass offset to it

This is true but we could position a dummy div which is offset and use that as the basis for scrolling. We do this in the sticky implementation, for example, to position a sentinel to offset the nav height.

dzucconi:
And right, I was also talking about the native `element.scrollIntoView` as well — we should just deprecate [this thing.](https://github.com/artsy/force/blob/10bd7fcde3d07ce062c3a767e85a9f69399a3350/src/v2/Utils/ScrollIntoView.tsx) I'm not really sure why it exists in the first place.

anastasiapyzhik:
> > As I know we can't pass offset to it
> 
> This is true but we could position a dummy div which is offset and use that as the basis for scrolling. We do this in the sticky implementation, for example, to position a sentinel to offset the nav height.

You're right! Scrolling to anchor is easier and more reliable, thank you. I wish I'd seen your comment a few hours earlier

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,51 @@
...
+import { ScrollRefContext } from "./useScrollContext"
 
 export const FilterExpandable: React.FC<ExpandableProps> = props => {
   const tokens = useThemeConfig({
     v2: { mb: 1 },
     v3: { mb: 6 },
   })
 
-  return <Expandable mb={tokens.mb} {...props} />
+  // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 17: This will probably error out in our CI as we can remove these, but we can't add any additional one's. 

(This is part of some tooling we added to migrate to strick type checking in force.)

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,51 @@
...
+      const { height: containerHeight, top: containerTop } = getElementParams(
+        scrollRef.current
+      )
+      const { height: filterHeight, top: filterTop } = getElementParams(
+        filterRef.current
+      )
+      const visiblePartHeight = containerHeight - filterTop + containerTop
+
+      if (visiblePartHeight < filterHeight) {
+        // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 35: Ditto here

dzucconi: In general it makes sense to just do existential checks on all refs at the top of any effects.

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,51 @@
...
   const tokens = useThemeConfig({
     v2: { mb: 1 },
     v3: { mb: 6 },
   })
 
-  return <Expandable mb={tokens.mb} {...props} />
+  // @ts-expect-error STRICT_NULL_CHECK
+  const { scrollRef } = useContext(ScrollRefContext)
+  const filterRef = useRef(null)
+  const anchorRef = useRef(null)
```
dzucconi, line 20: You'd wanna type these: `useRef<HTMLDivElement | null>(null)`

anastasiapyzhik:
@artsy/fx-devs I'm gonna start updating tests once you accept these changes

src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
```diff
@@ -108,7 +108,7 @@ export const ArtworkFilterMobileActionSheet: React.FC<{
       </Header>
 
       <Content ref={contentRef as any}>
-        <Box width="100%" p={2}>
+        <Box width="100%" p={2} pb={0}>
```
dblandin, line 111: I'm surprised that Box would have inherent padding without this explicit 0 value.

dblandin (approved):
This looks good to me!

**One minor naming suggestion**: Perhaps instead of `WithScroll` which sounds like you can scroll within the filter, maybe a more verbose `WithScrollIntoView` to communicate the added behavior?

src/v2/Components/ArtworkFilter/ArtworkFilterMobileActionSheet.tsx
dzucconi: It doesn't — but there's a p={2} here. It'd be better to just do `px={2} pt={2}`

dblandin: Ahh, that's right. Eyes skipped right over that first attribute.

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,67 @@
...
+import React, { useEffect, useRef, useState } from "react"
+import { Media } from "v2/Utils/Responsive"
+import { getElementParams } from "./helpers"
+import { ScrollRefContextValue, useScrollRefContext } from "./useScrollContext"
+
+const FilterExpandableWithScroll: React.FC<
+  ExpandableProps & { tokens: { mb: number } }
+> = ({ tokens, ...props }) => {
+  const ctx = useScrollRefContext()
+  const { scrollRef } = ctx as ScrollRefContextValue
```
dzucconi, line 16: This should return it's own type without having to cast

src/v2/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -1,11 +1,67 @@
...
+  return (
+    <>
+      <Media at="xs">
+        <FilterExpandableWithScrollIntoView tokens={tokens} {...props} />
+      </Media>
+      <Media greaterThan="xs">
+        <Expandable mb={tokens.mb} {...props} />
+      </Media>
+    </>
+  )
```
dzucconi, line 66: I'd consider just using `FilterExpandableWithScrollIntoView` and have it accepting a prop to enable/disable scrolling. The reason being that: there's a lot of DOM in these filters, so if possible we should avoid duplicating them. In this case we don't need to worry about the server-side rendering because any click interaction is going to happen on the client-side so we can just detect if it's mobile or not there.

dzucconi (approved):
Thanks for bearing with — looks good!

anastasiapyzhik:
> Thanks for bearing with — looks good!

Thank you for review!

PR #7715, opened 2021-06-07
fix(PX-4191): Fixes logged in sticky position

dzucconi:
You can see a thread about the styling of the buttons here: https://artsy.slack.com/archives/C9XJKPY9W/p1623078390072600 — (They are currently V2 because the whole nav is V2 and doesn't change with route)


src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -87,7 +89,10 @@ export const NavBar: React.FC = track(
   return (
     <>
       <NavBarSkipLink />
```
dzucconi, line 91: We want the skip link to be the first focusable thing. This is a little hard to enforce unfortunately.

src/v2/Components/NavBar/constants.ts
```diff
@@ -1,9 +1,7 @@
-import { space } from "@artsy/palette"
-
 export const NAV_BAR_BORDER_OFFSET = 1
-export const NAV_BAR_TOP_TIER_HEIGHT = space(6) - NAV_BAR_BORDER_OFFSET
+export const NAV_BAR_TOP_TIER_HEIGHT = 60 - NAV_BAR_BORDER_OFFSET
 export const NAV_BAR_BOTTOM_TIER_HEIGHT = 44
-export const MOBILE_BANNER_HEIGHT = 64
+export const MOBILE_BANNER_HEIGHT = 62
```
dzucconi, line 4: This value will change when we switch it to V3 — V3 buttons are `50px` high, V2 buttons are `41px`

PR #7279, opened 2021-04-01
fix: bring back ask a question modal to BNMO checkout flow

olerichter00:
Addresses [PURCHASE-2175]

[PURCHASE-2175]: https://artsyproduct.atlassian.net/browse/PURCHASE-2175

PR #7716, opened 2021-06-07
chore: render Zendesk chat in checkout flow regardless of artwork price

ashleyjelks:
Zendesk collector chat within Checkout flow regardless of price. Previously we only rendered the chat option when the artwork price was over $10,000 USD or its equivalent.

Screen Shot 2021-06-07 at 12 32 56 PM

https://artsyproduct.atlassian.net/browse/PURCHASE-2693

### DIFF
--- a/src/v2/Apps/Order/OrderApp.tsx
+++ b/src/v2/Apps/Order/OrderApp.tsx
@@ -109,30 +109,6 @@ class OrderApp extends React.Component<OrderAppProps, {}> {
   }
 
   renderZendeskScript() {
-    const artwork = get(
-      this.props,
-      // @ts-expect-error STRICT_NULL_CHECK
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Order/OrderApp.tsx
```diff
@@ -183,6 +159,7 @@ class OrderApp extends React.Component<OrderAppProps, {}> {
                     content="width=device-width, initial-scale=1, maximum-scale=5 viewport-fit=cover"
                   />
                 )}
+                {!isEigen && this.renderZendeskScript()}
```
ashleyjelks, line 162: This was accidentally removed in a previous PR: https://github.com/artsy/force/pull/7279

erikdstock (approved):
lookin good here

PR #7718, opened 2021-06-07
fix(navbar): prevents doubled auth buttons

dzucconi:
So — since I moved this to the NavBar it was appearing on legacy pages too. Moved it back and will address some other issues tomorrow. This just prevents the doubling so that we can get a deploy in tonight.

One thing to note that is for sure a problem: on legacy pages we actually don't have a position fixed nav — this is fine but I'm assuming we would want to collapse this too as you scroll. The logged out view on a small phone is going to have a *tiny* screen area if we don't.


damassi:
Thanks for the quick fix 👍

PR #7731, opened 2021-06-08
fix: Legacy Dev Mode

icirellik:
During the recent asset renaming logacy development mode broke due to
how asset lookup is configured.


damassi (approved):
This looks good once the runtime chunk issue is fixed 👍

PR #7732, opened 2021-06-08
fix(unsubscribe): implements unsubscribe app; deletes old apps

dzucconi:
So at the moment [the unsubscribe API endpoint is a little weird](https://github.com/artsy/gravity/blob/master/app/api/v1/me_unsubscribe_endpoint.rb). Effectively, you can only really opt out of emails completely. We _could_ update that endpoint to accept a frequency explicitly (and clear up some of the weirdness with what it's returning, and wrap it in Metaphysics while we're at it...). Not sure if we need or want that. Lemme know how you wanna proceed.

Deletes the old desktop/mobile apps as well.


src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx
```diff
@@ -0,0 +1,58 @@
...
+            Email Preferences
+          </Text>
+
+          <Spacer mt={6} />
+
+          {!me ? (
+            <UnsubscribeLoggedOut authenticationToken={authenticationToken} />
+          ) : (
+            <UnsubscribeLoggedInFragmentContainer me={me} />
+          )}
```
jonallured, line 42: I have a slight preference for writing ternary expressions in the affirmative. No biggie but noticed it here.

```suggestion
          {me ? (
            <UnsubscribeLoggedInFragmentContainer me={me} />
          ) : (
            <UnsubscribeLoggedOut authenticationToken={authenticationToken} />
          )}
```

src/v2/Apps/Unsubscribe/Components/UnsubscribeLoggedOut.tsx
```diff
@@ -0,0 +1,104 @@
+import { Button, Checkbox, Banner, Spacer } from "@artsy/palette"
+import React, { useRef, useState } from "react"
+import { data as sd } from "sharify"
+
+enum Mode {
```
jonallured, line 5: Cool!!

src/v2/Apps/Unsubscribe/Components/UnsubscribeLoggedIn.tsx
```diff
@@ -0,0 +1,96 @@
...
+  me: UnsubscribeLoggedIn_me
+}
+
+export const UnsubscribeLoggedIn: React.FC<UnsubscribeLoggedInProps> = ({
+  me,
+}) => {
+  const { relayEnvironment } = useSystemContext()
+
+  const [emailFrequency, setEmailFrequency] = useState(
+    me.emailFrequency ?? "daily"
```
jonallured, line 25: We could possibly come back to this page and use this same fallback for signed out users but not worth holding up this PR - just something I was chatting with Jenna about.

jonallured (approved):
One minor suggestion but this looks awesome!! I did a call with Jenna and she agreed that this is worth shipping so that we can use this page during IP warming. 👍 

/cc @isakelaris

PR #7734, opened 2021-06-09
feat(PX-4132): Collector sees updated design for shipping step

Serge0n:
Updated design for shipping step

JIRA: https://artsyproduct.atlassian.net/browse/PX-4132

https://user-images.githubusercontent.com/55637696/121357059-20c50100-c93a-11eb-9334-714e1f9f318e.mov

### DIFF
--- a/src/v2/Apps/Order/Components/SavedAddressItem.tsx
+++ b/src/v2/Apps/Order/Components/SavedAddressItem.tsx
@@ -1,4 +1,4 @@
-import { Flex, Text, RadioProps } from "@artsy/palette"
+import { Flex, Text, RadioProps, Banner } from "@artsy/palette"
 import React from "react"
 import styled from "styled-components"
 import { SavedAddresses_me } from "v2/__generated__/SavedAddresses_me.graphql"
@@ -39,13 +39,32 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
           nameAndAddressLine.map(
             (line: string, index: number) =>
               line && (
-                <Text
-                  style={{ textTransform: "capitalize" }}
-                  variant="text"
-                  key={index}
-                >
-                  {line}
-                </Text>
+                <Flex key={index} justifyContent="row" alignItems="center">
+                  <Text
+                    style={{ textTransform: "capitalize" }}
+                    textColor={index === 0 ? "black100" : "black60"}
+                    variant="text"
+                  >
+                    {line}
+                  </Text>
+                  {address?.isDefault && index === 0 && (
+                    <Banner
+                      ml={0.5}
+                      variant="defaultLight"
+                      px={0.5}
+                      py={0.2}
+                      borderRadius={2}
+                    >
+                      <Text
+                        textColor="black60"
+                        variant="caption"
+                        display="inline"
+                      >
+                        Default
+                      </Text>
+                    </Banner>
+                  )}
+                </Flex>
               )
           )}
         <Text textColor="black60" style={{ textTransform: "capitalize" }}>
--- a/src/v2/Apps/Order/Components/SavedAddressItem.tsx
+++ b/src/v2/Apps/Order/Components/SavedAddressItem.tsx
@@ -41,7 +41,7 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
               line && (
                 <Flex key={index} justifyContent="row" alignItems="center">
                   <Text
-                    style={{ textTransform: "capitalize" }}
+                    textTransform="capitalize"
                     textColor={index === 0 ? "black100" : "black60"}
                     variant="text"
                   >
@@ -67,7 +67,7 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
                 </Flex>
               )
           )}
-        <Text textColor="black60" style={{ textTransform: "capitalize" }}>
+        <Text textColor="black60" textTransform="capitalize">
           {formattedAddressLine}
         </Text>
         <Text textColor="black60">{phoneNumber}</Text>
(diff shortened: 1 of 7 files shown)


src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -171,20 +171,36 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
...
+              setModalDetails({
+                addressModalTitle: "Add address",
+                addressModalAction: "createUserAddress",
+              })
+          }}
+        >
+          Add a new address
+        </Button>
+      ) : (
+        <Button
```
sweir27, line 189: Can you describe this use-case more? I don't see it in the designs (but could be looking in the wrong place!).

Serge0n: As I understand, if there are no saved addresses, one primaryBlack button is shown. 
If there is at least one saved address then secondaryOutline button is shown. But primaryBlack can be shown only on Profile settings ...
I can't check this on localy Force... maybe I need to configure .env file

image

sweir27: Oh interesting, so this component is re-used in the profile view?

My first thought is it might be fine to keep it as a white button instead of updating to be black. If we do want to keep this component in-use in both places, then perhaps we should pass a prop from the parent view to determine button color (as it seems the logic is more reflective of that vs. how many addresses there are).

@sepans since I think you may have worked on saved addresses? -- how does this sound?

Serge0n: SavedAddresses component (which contains "Add new address" button) is re-used in the profile view.

src/v2/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -39,13 +39,32 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
...
-                <Text
-                  style={{ textTransform: "capitalize" }}
-                  variant="text"
-                  key={index}
-                >
-                  {line}
-                </Text>
+                <Flex key={index} justifyContent="row" alignItems="center">
+                  <Text
+                    style={{ textTransform: "capitalize" }}
```
damassi, line 44: We can use the `styled-system` prop `textTransform='uppercase'` here

src/v2/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -39,13 +39,32 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
...
+                  >
+                    {line}
+                  </Text>
+                  {address?.isDefault && index === 0 && (
+                    <Banner
+                      ml={0.5}
+                      variant="defaultLight"
+                      px={0.5}
+                      py={0.2}
+                      borderRadius={2}
```
damassi, line 56: @dzucconi - do we have patterns in place around border-radius? I haven't seen this around except for in the `BorderBox` component

dzucconi: Not really; might be worth surfacing to `#design-system`

src/v2/Apps/Order/Components/SavedAddresses.tsx
sepans: Yes this component is being reused in `UserSettingsAddresses`. great catch @Serge0n 💯 

@sweir27 if you mean having like (psydocoding...)
```tsx
const buttonVariant = whoIsMyParent === '....' && hasSavedAddress ? ... : ....
const addButtonProps = { variant: buttonVariant ...}
<Button {...addButtonProps}>...
```
It sounds good to me. Also wondering if we want to use this opportunity to make the address edit functionality under payment and addresses settings and checkout flow as identical as possible.

sweir27: @sepans Yes! And I agree with making the behavior the same eventually 😅.

Not for this PR, but we may want to move this shared component out of `Order` since it's a little confusing to have it there but also be shared!

src/v2/Apps/Order/Components/SavedAddressItem.tsx
sweir27: @Serge0n mind making this change? Thanks!

sweir27:
This looks good! @Serge0n -- one little comment that it would be good to address. I'll wait to merge this until I'm online tomorrow so we can keep an eye on failures in our integration testing job run (#integrity), in case it relies on specific form fields/elements that have changed.

src/v2/Apps/Order/Components/SavedAddressItem.tsx
Serge0n: @sweir27 Done)

src/v2/Apps/Order/Components/SavedAddressItem.tsx
dzucconi: BTW: apparently we've been trying to drop border radiuses everywhere

dzucconi: Also: I didn't even notice this until just now: you don't want to be giving custom borders/padding etc to these components.

dzucconi: Just looking at the comp now, this just isn't a `Banner` — maybe a `Pill`?

Serge0n: @dzucconi thanks, I'll use Pill

sweir27:
Merging this in to do a little testing on staging! @Serge0n will follow up with the `Pill` change. 🙏

PR #7739, opened 2021-06-09
Fixes image flashes on route change

dzucconi:
### Before:

https://user-images.githubusercontent.com/112297/121412681-fb80c480-c932-11eb-990d-cb864593720f.mov

### After:

https://user-images.githubusercontent.com/112297/121412785-12bfb200-c933-11eb-9601-a4fd1b6abcd7.mov

### DIFF
--- a/src/v2/Components/Artwork/SaveButton/index.tsx
+++ b/src/v2/Components/Artwork/SaveButton/index.tsx
@@ -1,234 +1,204 @@
 import { AuthContextModule, Intent } from "@artsy/cohesion"
 import { SaveButton_artwork } from "v2/__generated__/SaveButton_artwork.graphql"
-import * as Artsy from "v2/Artsy"
-import { extend, isNull } from "lodash"
-import React from "react"
+import React, { useState } from "react"
 import { RelayProp, createFragmentContainer, graphql } from "react-relay"
 import { TrackingProp } from "react-tracking"
 import * as RelayRuntimeTypes from "relay-runtime"
-import styled from "styled-components"
+import styled, { css } from "styled-components"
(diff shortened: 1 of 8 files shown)


src/v2/Components/Artwork/SaveButton/index.tsx
```diff
@@ -1,234 +1,204 @@
...
+  artwork: graphql`
+    fragment SaveButton_artwork on Artwork {
+      id
+      internalID
+      slug
+      is_saved: isSaved
+      title
+    }
+  `,
+})
```
dzucconi, line 204: This was just a mindless translation to functional component with the addition of using `useSystemContext` instead of accepting mediator/user as props. We could clean this up a lot but save that for another day.

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -295,7 +289,15 @@ export const BaseArtworkFilter: React.FC<
...
-            {children || <ArtworkGrid />}
+            {children || (
+              <ArtworkFilterArtworkGrid
+                // @ts-expect-error STRICT_NULL_CHECK
+                filtered_artworks={viewer.filtered_artworks}
+                isLoading={isFetching}
+                offset={offset}
+                columnCount={[2, 2, 2, 3]}
+              />
+            )}
```
damassi, line 300: jesus

dzucconi:
> Will this also fix the issue with lazy being added to image components?

What do you mean?

dzucconi:
Yeah this addresses it elsewhere; threading the user/mediator was what was causing it in the shelves. It's less clear to me *why* that was a problem exactly; but either way just pulling those props out contained in the save button fixes the issue and reduces prevents the image from re-rendering.

PR #7747, opened 2021-06-10
fix: Show fair booth when it's full featured only

anastasiapyzhik:
https://artsyproduct.atlassian.net/browse/FX-2535

**Before**
Screen Shot 2021-06-10 at 6 47 53 PM

**After**
Screen Shot 2021-06-10 at 6 48 36 PM

### DIFF
--- a/src/v2/Apps/Show/ShowApp.tsx
+++ b/src/v2/Apps/Show/ShowApp.tsx
@@ -104,8 +104,12 @@ export const ShowApp: React.FC<ShowAppProps> = ({ show }) => {
               <ShowArtworksEmptyState show={show} />
             </>
           )}
-          <Separator as="hr" my={6} />
-          <ShowContextCard show={show} />
+          {show.fair?.hasFullFeature !== false && (
+            <>
+              <Separator as="hr" my={6} />
+              <ShowContextCard show={show} />
+            </>
+          )}
         </AnalyticsContext.Provider>
       </>
     </>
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Show/ShowApp.tsx
```diff
@@ -104,8 +104,12 @@ export const ShowApp: React.FC<ShowAppProps> = ({ show }) => {
               <ShowArtworksEmptyState show={show} />
             </>
           )}
-          <Separator as="hr" my={6} />
-          <ShowContextCard show={show} />
+          {show.fair?.hasFullFeature !== false && (
```
dblandin, line 107: **minor / non-blocking**: I think a truthy conditional would work just as well here:

```suggestion
          {show.fair?.hasFullFeature && (
```

dblandin (approved):
Looks great! One minor suggestion which you can take or leave. Assigning back to you so you can merge when ready. 💯

src/v2/Apps/Show/ShowApp.tsx
anastasiapyzhik: That was intended. Fair may be either object with `hasFullFeature` prop or `null`.
For instance here https://www.artsy.net/show/loft-art-gallery-mous-lamrabat-digital-solo-show?sort=partner_show_position fair is `null` (hence `hasFullFeature` is `falsy`) but link is not broken, it just refers to _/loft-art-gallery_, not to _/fair/loft-art-gallery_

dblandin: Ahh, that makes sense. Thanks for clarifying!!

PR #7749, opened 2021-06-10
lib: Upgrade styled-components@5.x

damassi:
This upgrades styled-components to the latest: https://medium.com/styled-components/announcing-styled-components-v5-beast-mode-389747abd987.

Note this _does not_ update the typescript types package as that leads to a cascade of insane errors (even though styled-components 5 is backwards compatible with 4). Will take another look at this in a follow-up, as the priority in this PR is to increase performance:

- 26% smaller bundle size (16.2kB vs. 12.42kB min+gzip)
- 22% faster client-side mounting 
- 26% faster updating of dynamic styles
- 45% (!!!) faster server-side rendering

### DIFF
--- a/package.json
+++ b/package.json
@@ -305,7 +305,7 @@
     "babel-plugin-lodash": "3.3.4",
     "babel-plugin-module-resolver": "3.1.0",
     "babel-plugin-relay": "9.0.0",
-    "babel-plugin-styled-components": "1.11.1",
+    "babel-plugin-styled-components": "1.12.0",
     "benv": "3.3.0",
     "cache-loader": "1.2.2",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -164,6 +164,7 @@
     "knox": "0.9.2",
     "lodash": "4.17.21",
     "mailcheck": "1.1.1",
+    "memoize-one": "^5.2.1",
```
damassi, line 167: This weirdly keeps coming up as a missing package, not sure /shrug

damassi, line 167: looks like this was previously a transitive dep; its used in the artwork grid

dzucconi:
🙏 🙏

PR #7753, opened 2021-06-11
feat(PX-4133): Collector sees updated address form styling

Serge0n:
Updated design for address form

JIRA: https://artsyproduct.atlassian.net/browse/PX-4133

While I was working on this task, I had two questions:
1) How I can set modal width like in Figma ? ("618px").  Is it necessary? Because I can only choose from the available values:
`export declare enum ModalWidth {
    Narrow = "300px",
    Normal = "440px",
    Wide = "900px"
}`
or I can set `isWide` prop.
2) In Figma all input titles are bold. But I think I can't set bold prop into input title by using `Input` props... If needed bold font like in Figma, I can delete `title="Address Line 1"` and create `<Text variant="mediumText">Address Line 1</Text>`
But I think separating title from Input is bad practice. What should I do in the end?

image

### DIFF
--- a/src/v2/Components/Address/AddressModalFields.tsx
+++ b/src/v2/Components/Address/AddressModalFields.tsx
@@ -1,5 +1,14 @@
 import React from "react"
-import { Input, Spacer, Text } from "@artsy/palette"
+import {
+  Box,
+  Column,
+  Flex,
+  GridColumns,
+  Input,
+  Join,
+  Spacer,
+  Text,
+} from "@artsy/palette"
 import { useFormikContext } from "formik"
 import { SavedAddressType } from "v2/Apps/Order/Utils/shippingAddressUtils"
 import { CountrySelect } from "../CountrySelect"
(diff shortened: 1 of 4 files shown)


ansor4:
@ArvindLall how do you feel about the labels for each field in this form not being bold? Looks like the specs have each label bolded, but using palette's `Input1, we're not able to style the field labels since it's built into the component. Just wondering if we should be adding a separate `Text` over each field just for the boldness of the field labels.

ArvindLall:
@ansor4 - Yes can be unbold but we should be plugging in the new design system which I don't think is in your spec. There is an overlap on add address and edit address between our two teams. I am updating a similar design for saved addresses in my profile which should supersede this styling. 

Hopefully that all makes sense.

Serge0n:
can it be merged?

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -15,70 +24,88 @@ export const AddressModalFields: React.FC = props => {
...
+                title="Address Line 2 (optional)"
+                placeholder="Add address line 2"
+                name="addressLine2"
+                onChange={handleChange}
+                onBlur={handleBlur}
+                error={touched.addressLine2 && errors.addressLine2}
+                value={values?.addressLine2}
+              />
+              <Input
+                title="State, provence, or region"
```
sweir27, line 97: Should be `province` (here and below)

PR #7754, opened 2021-06-11
fix: Removed ignoring scroll behavior for route

anastasiapyzhik:
Jira: [FX-3005]

[FX-3005]: https://artsyproduct.atlassian.net/browse/FX-3005

### DIFF
--- a/src/v2/Apps/Fair/fairRoutes.tsx
+++ b/src/v2/Apps/Fair/fairRoutes.tsx
@@ -46,7 +46,6 @@ export const fairRoutes: AppRouteConfig[] = [
   {
     path: "/fair/:slug",
     theme: "v3",
-    ignoreScrollBehavior: true,
     getComponent: () => FairApp,
     prepare: () => {
       FairApp.preload()


dblandin:
@anastasiapyzhik For this PR, I think before/after screen recordings would be very useful!

damassi:
Copying a comment [from here](https://artsy.slack.com/archives/C2TQ4PT8R/p1623430209128100?thread_ts=1623402235.119600&cid=C2TQ4PT8R) before we merge: 

We have an `ignoreScrollBehavior` prop that we can add to the route config: https://github.com/artsy/force/blob/a8203e756fcc0562d2d7aa35a2a210d7692bab43/src/v2/Apps/Artist/artistRoutes.tsx#L160, but if this was set to true, then that means that if a user was halfway scrolled down the page when they clicked a link to a fair, it would keep them halfway scrolled down the page when they navigated to the fair for the first time (if the fair page is long enough, I think). If they navigated away and then back, it would preserve their scroll position though. So its logically complicated. We don’t have a good answer right now, but I think this issue is more architectural than a bug, and we need to find a better solution.

dblandin:
@damassi In this case, the undesirable behavior is when a user is navigating _back_ to a fair page.

**Expected Behavior**

When a scroll down on a fair page
and click on an artwork link
if I navigate my browser back
the previous scroll position on the fair page should be retained

**Actual Behavior**

When I scroll down on a fair page
and click on an artwork link
if I navigate my browser back
the fair page borrows the scroll position of the artwork page I just navigated to, losing my previous position

So it sounds like we either want to set this explicitly to `false` or remove the config line entirely. Unclear if we're going to be trading this odd behavior for other odd behavior though.

dblandin:
@damassi Do you have any other thoughts on this? This might be a good topic for web practice but the scheduled time is a bit late in the day for me.

dblandin:
Going to go ahead and close this PR for now until we can come to a better architectural solution.

PR #7756, opened 2021-06-11
fix: restores tracking to route tabs

dzucconi:
Closes [GRO-372](https://artsyproduct.atlassian.net/browse/GRO-372)


damassi (approved):
Thanks for the fix 👍

PR #7760, opened 2021-06-11
feat: add buyer guarantee message to inquiry checkout works

xtina-starr:
Addresses: [PURCHASE-2703](https://artsyproduct.atlassian.net/browse/PURCHASE-2703)

This PR adds the buyer guarantee messaging behind `isOfferableFromInquiry` and the Web Inquiry Checkout feature flag.

Screen Shot 2021-06-11 at 3 44 45 PM

It's worth noting I didn't include a test for this. Testing this component requires a bit of setup which we have a [ticket](https://artsyproduct.atlassian.net/browse/PURCHASE-2736) for.


src/v2/Apps/Conversation/Components/BuyerGuaranteeMessage.tsx
```diff
@@ -0,0 +1,26 @@
+import { Flex, Text } from "@artsy/palette"
+import React from "react"
+
+export const BuyerGuaranteeMessage: React.FC = () => {
+  return (
+    <>
```
damassi, line 6: Unless there's an adjacent element no need for fragments

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -25,12 +37,29 @@ export interface ConversationProps {
...
+
 const Conversation: React.FC<ConversationProps> = props => {
   const { conversation, relay, showDetails, setShowDetails } = props
+  const { user } = useSystemContext()
 
   const bottomOfPage = useRef(null)
   const initialMount = useRef(true)
 
+  const isMakeOfferArtwork =
+    // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 54: Was this code copied from elsewhere? If so, we shouldn't also be copying over these ts ignores

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -25,12 +37,29 @@ export interface ConversationProps {
 
 export const PAGE_SIZE: number = 15
 
+const StyledBox = styled(Flex)`
+  border-top: 1px solid ${color("black10")};
+  padding: 15px 20px;
```
damassi, line 42: We should always stay within our theme spacing scale, and use props on the component (`p={2}`, for example -- this component can likely go away and `<Flex>` styled directly)

src/v2/Apps/Conversation/Components/BuyerGuaranteeMessage.tsx
xtina-starr: Ah good catch. I previously had adjacent elements but removed it and forgot to remove the fragment.

src/v2/Apps/Conversation/Components/Conversation.tsx
xtina-starr: It wasn't copied over `conversation.items[0]` was throwing a type error: `Object is possibly 'null'` then when I added `conversation.items?[0]?.item?.__typename` I was getting `Property 'item' does not exist on type 'number[]'.` for item.

damassi: Ah 👍If having a type issue like this then will want to ping #dev-help or reach out! Seems like a quick fix.

xtina-starr: I was think that was an appropriate use of `// @ts-expect-error STRICT_NULL_CHECK` since most of the properties on the conversation object that can be nullable use it and it was added fairly recent as part of the strict null check work.

xtina-starr: If you have a second I'd be down to resolve it.

damassi: So the history there is -- The `// @ts-expect-error STRICT_NULL_CHECK` flags were inserted by an automatic codemod script so that we could add strict type checking to Force without having to fix every error in the codebase first. With the codemod run, its understood that all new code should be type safe (and handle strict type errors like the above). 

(Strangely, CI should have been failed the build since it keeps track of the number of current flags in the codebase, and if any new ones are added should throw an error 🤔) 

Whats missing is a proper check after items: `conversation.items?.[0]...` (note the `?.` before array access).

src/v2/Apps/Conversation/Components/Conversation.tsx
xtina-starr: This makes sense to me

xtina-starr: Is there a way to make sure this is consistent with the designs we get from designers. Like are they aware of this (i.e. 15px not being an option in our theme and to not use it in specs)?

damassi: Other than reporting back to the designer directly and helping to educate, not too sure. We, in engineering, are the last guard against design drift! So if some design is "15px 20px", and we know that there's a `1 = 10px, and 2 -20px` in our design system theme, then we can either round that `15px` up or down using our best judgement / eye. But we never step out of the design system spacing scale which only creates further problems at scale. Most likely, the designer just isn't being pixel perfect here (which is fine since engineers can stick within the theme and values "snap" to their closest targets.)

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -27,10 +39,19 @@ export const PAGE_SIZE: number = 15
...
   const bottomOfPage = useRef(null)
   const initialMount = useRef(true)
 
+  const isMakeOfferArtwork =
+    conversation.items?.[0]?.item?.__typename === "Artwork" &&
+    conversation.items?.[0]?.item?.isOfferableFromInquiry
+
+  const showBuyerGuaranteeMessage =
+    // @ts-expect-error STRICT_NULL_CHECK
+    userHasLabFeature(user, "Web Inquiry Checkout") && isMakeOfferArtwork
```
damassi, line 53: Ahh, not sure i was clear in my previous comment. We should really strive to make sure that type issues are never ignored and that these flags are never included in PRs with new code. If a type error is encountered, someone on the team can help solve. 

`// @ts-expect-error STRICT_NULL_CHECK` in particular should never be used. That was only added via an automatic codemod to help us migrate to strict type checking. On the rare occasion where something needs to be ignore we can use `@ts-ignore`. 

Whats the type error here? If it's something about the `user` being null, we can add `!` to the end of `user` to tell TS that one is certain the user is not null. But that makes code unsafe. Its better to properly guard: 

```ts
if (user && userHasLabFeature(user, "Web Inquiry Checkout") && isMakeOfferArtwork) {
  ... 
}
```

xtina-starr: Ah okay. I understand. Thanks for explaining.

damassi: np! I noticed another dev add a few `// @ts-expect-error STRICT_NULL_CHECK` comments as well so I should probably drop a new message in #dev, I bet a lot of folks didn't see Justins message.

sepans (approved):
Looks great 🚀

PR #7762, opened 2021-06-11
fix(GRO-371): Updates popovers; prevents artist suggestion popover from overflowing

dzucconi:
Closes: [GRO-371](https://artsyproduct.atlassian.net/browse/GRO-371)

I'm impressed with how much of a rabbit hole I managed to turn this into. 😅 

I left the unstated stuff alone for now though I'd like to follow up with it's removal.


src/v2/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActions.tsx
```diff
@@ -129,23 +128,17 @@ export class ArtworkActions extends React.Component<ArtworkActionsProps> {
 
   renderShareButton() {
     return (
-      <MediocrePopover
```
damassi, line 132: lol

src/v2/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -169,66 +157,69 @@ export class FollowArtistButton extends React.Component<Props, State> {
...
+              artist={artist!}
             />
-          </SuggestionsPopoverContainer>
-        )}
+          }
+        >
+          {({ anchorRef, onVisible }) => {
+            const openSuggestions = () => {
+              if (triggerSuggestions && !artist?.is_followed) {
+                onVisible()
```
damassi, line 186: what does `onVisible` do?

damassi (approved):
(Thanks for the additional `STRICT_NULL_CHECK` cleanup)

src/v2/Components/FollowButton/FollowArtistButton.tsx
dzucconi: Opens the popover

damassi: I see, so you can set `<Popover visible /> or use this form to display 👍

dzucconi: Exactly

PR #7764, opened 2021-06-11
refactor: Renames v2/Artsy folder to v2/System

damassi:
Having our framework code named "Artsy" has always been a bit confusing. This renames it to `System`. 

(Pre-pr to replacing our system context stuff with easy-peasy)


dzucconi:
Yes!

PR #7766, opened 2021-06-11
Adds cypress smoketests from integrity

sweir27:
## Problems
1. We have a smoke test suite in [integrity](https://github.com/artsy/integrity) that is good at identifying bugs! However, the integrity suite doesn't run on force builds, so it is not effective at blocking merges that may include errors. (Someone has to explicitly check the output).
2. Force's own smoke test suite is pretty anemic and doesn't have full coverage.
3. It's a little confusing to have the same tests in both integrity and in force, especially since both instances query the same environment (staging!).

## Solution
- This PR moves all of the smoke tests from integrity into force. An eventual follow up will remove the smoke tests from integrity entirely.
- See the notes below: A few are disabled.

## Some notes/outstanding issues
- A few of the tests are failing (disabled via `.skip` and with corresponding `//TODO:` messages) due to missing environment variables. Currently, the cypress suite runs a force server using the `.env.oss` file, but that's missing a few necessary, but secret, keys. We'll need to update our smoke test script to run force using a different set of environment variables (ideally pulling from a shared configuration).
- This incarnation of a cypress suite doesn't have any CYPRESS_ variables set up (as far as I can tell). We'll need to set up at least a `TEST_USER` and `TEST_PASSWORD` in order to pull a viewing room (to replicate integrity's behavior 1-1).
- Future tests, such as for `/user/edit` will also need to ensure that logging in works from within the cypress run (this will also require secrets available within the cypress test).

### DIFF
--- a/cypress/integration/about.spec.ts
+++ b/cypress/integration/about.spec.ts
@@ -1,0 +1,10 @@
+import { visitWithStatusRetries } from "../helpers/visitWithStatusRetries"
+
+// TODO: Currently throwing error. Likely needs additional ENV var.
+describe.skip("About", () => {
+  it("/about", () => {
+    visitWithStatusRetries("about")
+    cy.get("h1").should("contain.text", "The Art WorldOnline")
+    cy.title().should("eq", "About | Artsy")
+  })
+})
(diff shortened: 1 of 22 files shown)


damassi (approved):
Thanks for moving these over 👍 As a follow-up I'll add a danger rule that will check for additions to [our routes file](https://github.com/artsy/force/blob/master/src/v2/routes.tsx) and warn about adding a corresponding smoke test

PR #7771, opened 2021-06-14
chore: remove ENABLE_NEW_ARTWORK_FILTERS feature flag

anastasiapyzhik:
https://artsyproduct.atlassian.net/browse/FX-2793

Now that the new artwork filters have been launched and we no longer have to show filters conditionally, `ENABLE_NEW_ARTWORK_FILTERS` flag can be removed.


dblandin (commented):
This is looking good! I think there might be a lingering import referencing a deleted or updated file which is causing the CI failure. I'd recommend searching for the deleted paths to see if anything comes up.

anastasiapyzhik:
> This is looking good! I think there might be a lingering import referencing a deleted or updated file which is causing the CI failure. I'd recommend searching for the deleted paths to see if anything comes up.

That looks strange, this route is valid. Can't figure out what might be the problem
Screen Shot 2021-06-14 at 4 18 34 PM

src/v2/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
```diff
@@ -13,15 +12,18 @@ export function getWorksForSaleRouteVariables({ artistID }, props) {
...
+    "MAJOR_PERIOD",
+    "PARTNER",
+    "LOCATION_CITY",
+    "MATERIALS_TERMS",
+  ]
 
   return {
     input: allowedFilters(filterParams),
-    aggregations: allAggregations,
+    aggregations,
```
damassi, line 26: Much cleaner 👍

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -1,65 +1,212 @@
...
+    if (reset) {
+      setCustomSize({ height: ["*", "*"], width: ["*", "*"] })
+    }
+  }, [reset])
+
+  // @ts-expect-error STRICT_NULL_CHECK
+  const selection = currentlySelectedFilters().sizes
+  // @ts-expect-error STRICT_NULL_CHECK
+  const customHeight = currentlySelectedFilters().height
+  // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 181: We'll want to avoid adding these `@ts-expect-error STRICT_NULL_CHECK` comments in favor of making sure things are type-safe. Would something like 
```
currentlySelectedFilters()?.width
```
fix it? 

Another way to do it is:

```tsx
const selectedFilters = currentlySelectedFilters()! 
```

of which the `!` says: we're certain that `currentlySelectedFilters()` returns a non-null value. 

If it is possible that `currentlySelectedFilters()` can be null, we'll want to handle that properly however.

```tsx
const selectedFilters = currentlySelectedFilters()

if (selectedFilters) {
  ... do stuff 
}
```

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -68,6 +215,90 @@ export const SizeFilter: React.FC<SizeFilterProps> = ({ expanded }) => {
           })}
         </Flex>
       </Flex>
+
+      <Text></Text>
```
damassi, line 219: stray

damassi:
@anastasiapyzhik - you'll want to rebase from master! The path errors come from this refactor PR: https://github.com/artsy/force/pull/7764

damassi:
This PR looks 💯 from my end, but will assign back to @dblandin since there's some filter-specific logic stuff that he should be able to review better than I.

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
anastasiapyzhik: Thank you, that's very helpful!

dblandin (approved):
Nice work!! 💯

PR #7778, opened 2021-06-14
fix(GRO-367): AuthBanner shouldn't be fixed

dzucconi:
Closes: [GRO-367](https://artsyproduct.atlassian.net/browse/GRO-367)

https://user-images.githubusercontent.com/112297/121929196-ef1fb180-cd0e-11eb-8d0a-16441cd8b737.mov

Draft because I'm debating what to do with the sticky image attachment header in the partner app.

A minor side effect of this approach is that on desktop in Chrome/FF if you scroll down quickly and for some reason are staring at the top nav you might see it jump a pixel or two. Since `IntersectionObserver` is async, it can be a bit behind. Interestingly in Safari it is rock solid. I don't think this is a huge issue since you really have to be looking for it to notice, but I'm pointing it out as a potential drawback.

### DIFF
--- a/src/v2/Components/NavBar/NavBar.tsx
+++ b/src/v2/Components/NavBar/NavBar.tsx
@@ -18,13 +18,13 @@ import { track, useTracking } from "v2/System/Analytics"
 import Events from "v2/Utils/Events"
 import { useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
 import { NavBarPrimaryLogo } from "./NavBarPrimaryLogo"
-import { NavBarSkipLink } from "./NavBarSkipLink"
 import { LoggedInActionsQueryRenderer as LoggedInActions } from "./LoggedInActions"
 import {
   NAV_BAR_TOP_TIER_HEIGHT,
   NAV_BAR_BOTTOM_TIER_HEIGHT,
 } from "./constants"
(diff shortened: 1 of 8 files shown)


src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -85,9 +85,7 @@ export const NavBar: React.FC = track(
   }
 
   return (
-    <>
-      <NavBarSkipLink />
-
+    <Sticky zIndex={2}>
```
damassi, line 88: For some reason I feel like this one line of code is gonna fix a bunch of future things

dzucconi: Have we run into too many z-index wars in recent code? I suppose there may have been a reason this was `100`...

dzucconi: I generally tend to think that if a z-index is greater than `1` something is wrong. Here I set it to `2` so that things scrolling past it that are `1` don't overlap. Though I wonder if I should really just portal this to the end of the page as if it were a modal so that it naturally stacks using `1`.

damassi: This _was_ 100 because of old force app shell. For new pages 2 should work, but I'd check some old pages to see if things are ok.

PR #7780, opened 2021-06-14
feat: add Zendesk integration to auctions pages

ashleyjelks:
Adds Zendesk integration to the following Auction pages:
- registration
- bid confirmation
- landing page
- auction lots

cc @artsy/purchase-devs 

https://artsyproduct.atlassian.net/browse/PURCHASE-2477

Screen Shot 2021-06-14 at 12 26 48 PM

Screen Shot 2021-06-14 at 12 27 28 PM
Screen Shot 2021-06-14 at 12 28 01 PM

Screen Shot 2021-06-14 at 1 01 18 PM
Screen Shot 2021-06-14 at 12 23 02 PM


src/desktop/components/main_layout/templates/scripts.jade
```diff
@@ -10,6 +10,7 @@ script( src=asset("/assets/legacy-common-jquery.js") )
 script( src=asset("/assets/legacy-common-react.js") )
 script( src=asset("/assets/legacy-common-utility.js") )
 script( src=asset("/assets/legacy-artsy.js") )
+script( id='ze-snippet' src='https://static.zdassets.com/ekr/snippet.js?key=#{sd.ZENDESK_KEY}' )
```
ashleyjelks, line 13: mounting the Zendesk script tag here so needn't do it on every single page we want to render the chat button

src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
```diff
@@ -130,34 +130,6 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
   }
 
   renderZendeskScript() {
-    const {
```
ashleyjelks, line 133: I removed all this logic since 1) auctions don't have price listings so the button would never show up and 2) the other checks apply to ecommerce artworks, however, we are now showing the ZD icon on those pages as well so I think its safe to clean this up here.

src/desktop/components/main_layout/templates/scripts.jade
damassi: So does this mean that the zendesk code will be embedded on pages that don't use the button/

src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
xtina-starr: I think we would still want to have the thresholds on the artwork pages thought right? Unless MSO change the requirements?

src/desktop/components/main_layout/templates/scripts.jade
xtina-starr: Maybe we may want to conditionally render this based on whether it's an auction url? Or even at the script tag in the body of auctions page versus the main layout for all pages?

damassi: > Or even at the script tag in the body of auctions page

This is what I was thinking! Another way of going about it would be to add the script tag directly to the ZenDesk component so when a user adds the component it automatically comes with the embed script.

src/desktop/apps/auction/components/layout/auction_info/Registration.tsx
```diff
@@ -81,6 +82,12 @@ const Registration: React.FC<RegistrationProps> = props => {
     shouldPromptIdVerification,
   } = bidderQualifications(auction, user, userRegistration)
 
+  const renderZendeskScript = () => {
+    if (typeof window !== "undefined" && window.zEmbed) return
```
sepans, line 86: suggestion:

seems this check is done every where the component is being used (in [OrderApp](https://github.com/artsy/force/blob/5e102304113da2a26dadbb285526acb872fa070e/src/v2/Apps/Order/OrderApp.tsx#L112) and [ArtworkMeta](https://github.com/artsy/force/blob/5e102304113da2a26dadbb285526acb872fa070e/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx#L161)) so it is possible to move the check inside the component. (is that the case @xtina-starr?)

src/desktop/components/main_layout/templates/scripts.jade
sepans: I was looking to find where the zendesk script is loaded in the order/artwork pages and didn't find anything. Then I looked into the `react-zendesk` component (which is being used in `ZendeskWrapper` and it [includes](https://github.com/B3nnyL/react-zendesk/blob/master/src/index.js#L33) the script. So I am wondering why it needs to be added in the template?

src/desktop/apps/auction/components/layout/auction_info/Registration.tsx
xtina-starr: That's a good suggestion @sepans!!

ashleyjelks: I tried this and the `window` object is undefined when not called in the React component this check is being rendered in, I think it is bc it isn't in the func's scope when called this way. If there's a workaround, such as having the function accept `window` as a parm that could work, but it seems hacky.

src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
ashleyjelks: We removed the thresholds on BNMO works in the checkout flow,  it like we should be consistent and do the same thing on the artwork page? Also auction works don't have artwork prices so we'd do an early return here without this change. I can follow up with someone on MSO to see if this change is ok though.

src/desktop/components/main_layout/templates/scripts.jade
ashleyjelks: > I was looking to find where the zendesk script is loaded in the order/artwork pages and didn't find anything. Then I looked into the `react-zendesk` component (which is being used in `ZendeskWrapper` and it [includes](https://github.com/B3nnyL/react-zendesk/blob/master/src/index.js#L33) the script. So I am wondering why it needs to be added in the template?

Good question that I don't have an answer to other than to say that the library isn't mounting the script on our legacy pages so it needs to be injected in order to get the component to render.

xtina-starr (commented):
> The chat department to be used is “Auctions Chat”. 

Another thing worth noting is that the auction zendesk integration should launch a different channel and should use a different key from the one we are currently using (in `sd.ZENDESK`). I think we need to do the following:

1. Create a new environment variable for `AUCTION_ZENDESK`.
2. Update the ZendeskWrapper to accept the key for the correct channel.

I'm down to pair on this as well or add more context where needed.

PR #7769, opened 2021-06-12
feat: Confirm artwork modal for web inquiry checkout

jo-rs:
Resolves [PURCHASE-2706]

Includes modal and mutation, as well as edition mp query for [oncoming ticket](https://artsyproduct.atlassian.net/browse/PURCHASE-2708).

desk
mob

[PURCHASE-2706]: https://artsyproduct.atlassian.net/browse/PURCHASE-2706

PR #7781, opened 2021-06-14
fix: Updated ConfirmArtworkModal deps to system instead of artsy

jo-rs:
Updated deps for `ConfirmArtworkModal` to use `System` instead of `Artsy` as per [this](https://github.com/artsy/force/pull/7769#discussion_r651158340) comment.


dzucconi:
🙏

PR #7783, opened 2021-06-14
GRO-350: User sees broken link to gallery with partner slug instead of profile handle

The-Beez-Kneez:
This is to update the ArtistCareerHighlights to use the  `partner.profile.href` being passed in via metaphysics for slug generation on the Career Highlight.


jonallured (approved):
LGTM!! 🎉

PR #7784, opened 2021-06-14
feat: Suppress product artwork schema GRO-360

jonallured:
This PR updates our artwork structured data components to switch from `Product` to `CreativeWork` when we don't have price info. The ticket called for suppression of the structured data altogether, but I chatted with Jenna and we decided there was still some value to running the more generic type here.

I tried really hard to verify this fix prior to creating a PR, but for whatever reason the [Rich Results Test](https://search.google.com/test/rich-results) does not like ngrok URLs, haha. Furthermore, staging can't be tested either because we block robots:

```
jon@juggernaut:~/code/force(master)% http https://staging.artsy.net/robots.txt --body
User-agent: *
Disallow: /
```

So then I updated my local env file to allow robots but it still didn't work, probably because of whatever shenanigans ngrok does. 🤷 I don't think this is a very risky change to test in prod but open to other ideas. I did try to copy/paste the html but that doesn't work because so much happens client side.

https://artsyproduct.atlassian.net/browse/GRO-360

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Artwork/Components/Seo/SeoDataForArtwork.tsx
+++ b/src/v2/Apps/Artwork/Components/Seo/SeoDataForArtwork.tsx
@@ -44,11 +44,12 @@ export const SeoDataForArtwork: React.FC<SeoDataForArtworkProps> = ({
 
   // @ts-expect-error STRICT_NULL_CHECK
   const partnerType = get(artwork, a => a.partner.type)
-  if (partnerType === "Institution") {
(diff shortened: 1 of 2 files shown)


damassi:
👍 Lets see how it goes

jonallured:
Follow up - it worked!

Screen Shot 2021-06-15 at 10 28 15 AM

And by "it worked" I mean that we no longer have an error. 😝

PR #7785, opened 2021-06-15
[WIP] refactor: Add easy-peasy and refactor SystemContext to support it 

damassi:
Following Eigen's lead, this PR adds [easy peasy](https://easy-peasy.vercel.app) to the app, which is a modern, more performant, and more ergonomic way to manage state within a react app. Under the hood it uses [immer](https://github.com/immerjs/immer) and lives on top of redux for its root functionality, giving us access to a wide range of middleware and other tooling from that ecosystem. 

The first place I've updated is our SystemContext component. Next we'll be updating our Artwork filter.


damassi:
Gonna close this for now. Thinking it through a bit, until we need more sophisticated global functionality the knowledge cost of adding a new library here isn't worth it.

PR #7786, opened 2021-06-15
fix: Fix cropping artwork in narrow viewport (PURCHASE-2741)

copperkraft:
fixes https://artsyproduct.atlassian.net/browse/PURCHASE-2741

Change styles to limit the width of an artwork image in conversation to fit a narrow viewport
image


jo-rs (approved):
Nice and simple fix! 🏅

PR #7790, opened 2021-06-15
feature: Webpack 5

icirellik:
The long awaited Webpack 5 upgrade has finally arrived. This PR simply
upgrades all the nescesary plugins and Webpack to be running their
latest compatible versions.

There is a 50 KB (gzip) or 3.5% decrease in bundle size due to better
tree shaking.


src/v2/DevTools/renderUntil.tsx
```diff
@@ -16,15 +16,15 @@ function renderUntil<
       if (predicate(this)) {
         resolve(this)
       } else {
-        setImmediate(() => {
+        setTimeout(() => {
```
icirellik, line 19: These were causing browser errors as `setImmediate` is a node-only function and Webpack no longer auto polyfills.

package.json
```diff
@@ -175,6 +175,7 @@
     "nouislider": "9.2.0",
     "numeral": "2.0.6",
     "openseadragon": "2.4.2",
+    "os-browserify": "0.3.0",
```
damassi, line 178: What is this one?

src/v2/index.ejs
```diff
@@ -42,7 +42,7 @@
   <div id="react-root"><%%- content.body %></div>
 
   <!-- String concatenation is needed allow outputting a template tag -->
-  <%- htmlWebpackPlugin.tags.bodyTags.map((originalTag) => { const tag = { ...originalTag, attributes: { ...originalTag.attributes } }; tag.attributes['src'] = "<" + "%= cdnUrl %" + ">" + originalTag.attributes['src']; tag.attributes['async'] = true; return tag; }) %>
+  <%- htmlWebpackPlugin.tags.headTags.map((originalTag) => { const tag = { ...originalTag, attributes: { ...originalTag.attributes } }; tag.attributes['src'] = "<" + "%= cdnUrl %" + ">" + originalTag.attributes['src']; return tag; }) %>
```
damassi, line 45: Could this `headTags` switch potentially output head-only tags here (vs body-safe script tags). I guess I'm just wondering why the switch 😄)

webpack/envs/clientDevelopmentConfig.js
```diff
@@ -55,16 +54,18 @@ export const clientDevelopmentConfig = {
     ...standardPlugins,
     new webpack.HotModuleReplacementPlugin(),
     new ForkTsCheckerWebpackPlugin({
-      checkSyntacticErrors: true,
-      formatter: "codeframe",
-      formatterOptions: "highlightCode",
-      watch: ["./src/v2"],
```
damassi, line 61: Non-blocking - `watch` was pointed at v2 to speed things up a bit, since most product devs never leave the `v2` folder

damassi (approved):
A couple questions, but nothing blocking! Thanks for coming back to this and finishing it up now that the ecosystem supports it better 👍

joeyAghion (approved):

mc-jones (approved):

package.json
icirellik: Yeah, recommended in the Webpack 5 migration guide.

src/v2/index.ejs
icirellik: The v1.x and v2.x plugins are entirely different. 🤷 

They moved where the tags are inserted internally, I'm not entirely sure why however there are some newly added compiler hooks that could possibly simplify this logic. The way we inject the CDN being the only unresolved issue.

webpack/envs/clientDevelopmentConfig.js
icirellik: Literally doesn't exist anymore, from what I can gather in their changelog it never really worked, caused major performance issues, and duplicated watch functionality that was already in place.

src/v2/index.ejs
icirellik: They also added `defer` by default so I dropped `async`.

Dockerfile
```diff
@@ -82,7 +82,7 @@ COPY patches ./patches
 # ---------------------------------------------------------
 FROM yarn-base as yarn-deps
 
-RUN --mount=type=cache,target=/usr/local/share/.cache \
+RUN --mount=type=cache,target=/usr/local/share/.cache/yarn \
```
icirellik, line 85: This should not be necessary however docker would not run without it.

https://app.circleci.com/pipelines/github/artsy/force/15114/workflows/b7572059-a807-4da2-922b-c6cd72c5e4ef/jobs/111695

src/v2/DevTools/renderUntil.tsx
icirellik: After further testing, this one should not be required.

src/v2/Utils/Hooks/useRouteComplete.ts
```diff
@@ -19,9 +19,9 @@ export const useRouteComplete = ({
 
       // Wait till after the next tick to set to false, to give react tree
       // ability to execute related tracking handlers
-      setImmediate(() => {
+      setTimeout(() => {
```
icirellik, line 22: This one is required.

PR #7791, opened 2021-06-15
fix: TOS checkbox links (GRO-359)

ovasdi:
https://artsyproduct.atlassian.net/browse/GRO-359

~This change sets the Terms of Service checkbox link underline behavior to default which displays underlines under links.~
This change removes Terms of Service checkbox link `underlineBehavior` prop which defaults to links having andunderline.

:

| `underlineBehavior=default`  |  prop removed |
|---|---|
| GDPR sign up modal Screen Shot 2021-06-15 at 1 54 15 PM | GDPR sign up modal Screen Shot 2021-06-15 at 3 43 55 PM |
| US sign up modal Screen Shot 2021-06-15 at 1 54 53 PM | US sign up modal Screen Shot 2021-06-15 at 3 44 22 PM |

~I noticed that removing the prop entirely accomplishes the same and that we don't have any links with `underlineBehavior=default` across the code base. Would it be better to remove those props entirely from the TOS checkbox links?~

### DIFF
--- a/src/v2/Components/Authentication/TermsOfServiceCheckbox.tsx
+++ b/src/v2/Components/Authentication/TermsOfServiceCheckbox.tsx
@@ -21,7 +21,7 @@ const GdprLabel = color => {
         href="https://www.artsy.net/terms"
         target="_blank"
         color={color}
-        underlineBehavior="hover"
+        underlineBehavior="default"
       >
         Terms of Use
       </Link>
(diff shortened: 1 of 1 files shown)


jonallured:
> Would it be better to remove those props entirely

Yeah that seems slightly better to me. 👍

ovasdi:
> > Would it be better to remove those props entirely
> 
> Yeah that seems slightly better to me. 👍

Cool. I've added screen grabs for comparison just to be sure it looks the same.

jonallured:
TY! 🏆

PR #7792, opened 2021-06-15
feat: Rearrange sign up field order GRO-269

jonallured:
After some non-zero amount of panic that maybe we leaked user passwords, we discovered that a cohort of users had entered their password into the name field. 😝 

It's understandable because so many sign up forms ask you to enter and then re-enter your password so people are probably just really used to that flow. So then all this PR does is pull that name field up to first place. Our theory is that this change will prevent users from making this mistake in the future. It's also more natural to ask for name first.

More chatter here:

https://artsy.slack.com/archives/CTZA5NBB9/p1617135285003100

Looks like this:

Screen Shot 2021-06-15 at 2 54 02 PM

https://artsyproduct.atlassian.net/browse/GRO-269


dzucconi (approved):
Makes sense 👍

PR #7793, opened 2021-06-15
feat(GRO-214): sets up HomeApp

dzucconi:
Sets up the above the fold content. I figure the remainder will be all loaded client-side.

### DIFF
--- a/cypress/integration/home2.spec.ts
+++ b/cypress/integration/home2.spec.ts
@@ -1,0 +1,11 @@
+import { visitWithStatusRetries } from "../helpers/visitWithStatusRetries"
+
+describe("Home", () => {
+  it("/home", () => {
+    visitWithStatusRetries("home")
+    cy.get("h1").should(
+      "contain",
+      "Collect art from leading galleries, fairs, and auctions"
+    )
+  })
+})
--- a/src/v2/Apps/Home/Components/HomeHeroUnit.tsx
+++ b/src/v2/Apps/Home/Components/HomeHeroUnit.tsx
@@ -1,0 +1,111 @@
+import React from "react"
+import {
+  Image,
+  ResponsiveBox,
+  Spacer,
+  Text,
+  Button,
+  GridColumns,
+  Column,
+} from "@artsy/palette"
+import { createFragmentContainer, graphql } from "react-relay"
+import { HomeHeroUnit_heroUnit } from "v2/__generated__/HomeHeroUnit_heroUnit.graphql"
+import { cropped } from "v2/Utils/resized"
+import { RouterLink } from "v2/System/Router/RouterLink"
+
+interface HomeHeroUnitProps {
+  heroUnit: HomeHeroUnit_heroUnit
+}
+
+export const HomeHeroUnit: React.FC<HomeHeroUnitProps> = ({ heroUnit }) => {
+  const image = heroUnit.backgroundImageURL
+    ? cropped(heroUnit.backgroundImageURL, {
+        width: 910,
+        height: 683,
+        quality: 75,
+      })
+    : null
+
+  return (
+    <>
+      <RouterLink
+        to={heroUnit.href ?? ""}
+        style={{ display: "block" }}
+        tabIndex={-1}
+      >
+        <ResponsiveBox
+          aspectWidth={4}
+          aspectHeight={3}
+          maxWidth="100%"
+          bg="black10"
+        >
+          {image && (
+            <Image
+              src={image.src}
+              srcSet={image.srcSet}
+              width="100%"
+              height="100%"
+              lazyLoad
+            />
+          )}
+        </ResponsiveBox>
+      </RouterLink>
+
+      <Spacer mt={2} />
+
+      <Text variant="xs" textTransform="uppercase">
+        {heroUnit.heading}
+      </Text>
+
+      <Spacer mt={1} />
+
+      <RouterLink
+        to={heroUnit.href ?? ""}
+        tabIndex={-1}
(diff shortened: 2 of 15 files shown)


src/v2/Apps/Home/HomeApp.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { HomeApp_homePage } from "v2/__generated__/HomeApp_homepage.graphql"
+import { HomeHeroUnitFragmentContainer } from "./Components/HomeHeroUnit"
+import { HomeInfoBlurb } from "./Components/HomeInfoBlurb"
+
+interface HomeAppProps {
+  homePage: HomeApp_homePage
+}
+
+export const HomeApp: React.FC<HomeAppProps> = ({ homePage }) => {
+  const heroUnits = compact(homePage.heroUnits?.slice(0, 2)) ?? []
```
dzucconi, line 16: We definitely consistently have more than 2 of these — am curious what we want to do with them @deliciousnachos

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 113d74e7be365d10fc6d4e0d852f976437e82893
</p>

src/v2/Apps/Home/Components/HomeHeroUnit.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </GridColumns>
+    </>
+  )
+}
+
+export const HomeHeroUnitFragmentContainer = createFragmentContainer(
+  HomeHeroUnit,
+  {
+    heroUnit: graphql`
+      fragment HomeHeroUnit_heroUnit on HomePageHeroUnit {
```
damassi, line 100: (nonblocking) its always nicer long-term if the prop name being injected matches the type 1:1 -- less information in the system

src/v2/Apps/Home/__tests__/HomeApp.jest.tsx
```diff
@@ -0,0 +1,69 @@
...
+        homePage {
+          ...HomeApp_homePage
+        }
+      }
+    `,
+  })
+
+  describe("logged out", () => {
+    beforeAll(() => {
+      ;(useSystemContext as jest.Mock).mockImplementation(() => ({
```
damassi, line 30: Maybe 
```tsx 
const mockuseSystemContext = useSystemContext as jest.Mock
```

as we'll likely want to add future tests and 
```tsx
;(useSystemContext as jest.Mock)
``` 

could get a bit old

src/v2/System/useSystemContext.ts
```diff
@@ -6,5 +6,5 @@ import { SystemContext } from "./SystemContext"
  */
 export const useSystemContext = () => {
   const systemContext = useContext(SystemContext)
-  return systemContext
+  return { ...systemContext, isLoggedIn: !!systemContext.user }
```
damassi, line 9: We move this to the `SystemContextProvider` component where the rest of these initializers live

damassi (approved):
Couple minor comments, looks great 👌 

_Edit:_ just noticed the new danger warning: 

Screen Shot 2021-06-15 at 2 15 38 PM

Lets get this in from the get go before merge ;)

src/v2/Apps/Home/HomeApp.tsx
deliciousnachos: Ah yeah I need to give you a spec for that - will do

src/v2/Apps/Home/Components/HomeHeroUnit.tsx
dzucconi: Except that the field is called `heroUnits` — I can go either way; but I'm gonna lean towards the less verbose one.

damassi: Either way is fine with me 👍(This fresh on my mind after rebuilding artist page and noticing a lot of mismatch)

src/v2/System/useSystemContext.ts
dzucconi: This creates a bunch of annoying type errors. Doing a computed prop like this is fine IMO.

dzucconi: I can weaken the type of course; which is I guess why all of the other fields are optional.

dzucconi: What I mean is:



This kind of thing bothers me since there's no way this is actually undefined — in practice... I guess this is fine since you're typically only using it for a switch and the undefined would be falsey anway. Three-valued booleans just weird me out a bit.

damassi: If you feel the call to fix those types, they need it, but we should def be putting this stuff with the rest of the bunch rather than introducing drift

dzucconi:
Odd; type-check is passing locally. Unsure what's going on here.

damassi:
(Confirmed good over here too)

cypress/integration/home2.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe("Home", () => {
+  it("/home", () => {
+    visitWithStatusRetries("home")
+    cy.get("h1").should(
+      "contain",
+      "Collect art from leading galleries, fairs, and auctions"
+    )
+  })
+})
```
damassi, line 11: 🙏

PR #7794, opened 2021-06-15
docs: Add info about adding global script tags

damassi:
Per [this slack discussion](https://artsy.slack.com/archives/CP9P4KR35/p1623792192103600), wanted to add some docs to our FAQ about adding global script tags.


jonallured:
Perfect, thanks!! 🙌

PR #7797, opened 2021-06-16
feat(PX-4134): Collector can delete an address from the edit form

Serge0n:
First steps of add/delete logic in the shipping step

JIRA: https://artsyproduct.atlassian.net/browse/PX-4134

https://user-images.githubusercontent.com/55637696/122225492-cb4e9e00-cebd-11eb-820c-7abd300adfa8.mov


Serge0n:
can it be merged?

sweir27 (commented):
Couple of minor things!

In the video, it looks like there's a little bit of a delay from when you delete/add an address and when it actually disappears (or appears) on the page, which creates sort of a weird UX. One way around that could be to add an [`optimisticResponse`](https://relay.dev/docs/guided-tour/updating-data/graphql-mutations/#optimistic-response) to the mutations. Could you take a look to see if that helps?

Serge0n:
> Couple of minor things!
> 
> In the video, it looks like there's a little bit of a delay from when you delete/add an address and when it actually disappears (or appears) on the page, which creates sort of a weird UX. One way around that could be to add an [`optimisticResponse`](https://relay.dev/docs/guided-tour/updating-data/graphql-mutations/#optimistic-response) to the mutations. Could you take a look to see if that helps?

Yes, I started to implement this in the next PR.

src/v2/Apps/Order/Components/__tests__/SavedAddresses.jest.tsx
```diff
@@ -109,12 +113,11 @@ describe("SavedAddresses in collector profile", () => {
         addressConnection: mockAddressConnection,
       }),
     })
-    const editAddressComponent = wrapper.find("[data-test='editAddress']").at(0)
-    expect(editAddressComponent).toHaveLength(1)
+    const button = wrapper.find(Button)
```
pepopowitz, line 116: Blocking: 

Can we target this button more precisely than `Button`? This test could break pretty easily by adding new components to the page.

src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -36,6 +36,8 @@ interface SavedAddressesProps {
   commitMutation?: CommitMutation
   relay: RelayRefetchProp
   addressCount?: number
+  showModal?: boolean
```
pepopowitz, line 39: Is this needed? I don't see it referenced anywhere in this file.

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -29,6 +29,7 @@ const testAddressModalProps: Props = {
   address: savedAddress,
   onSuccess: jest.fn(),
   onError: jest.fn(),
+  onDeleteAddress: jest.fn(),
```
pepopowitz, line 32: I'd expect to see some tests in here to cover...

* the delete button appears at the right time
* clicking the delete button spawns a dialog
* when the dialog is confirmed, the "delete" action happens
* when the dialog is cancelled, the "delete" action doesn't happen

Are these intended to be added in a follow-up PR, or as part of this work?

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -129,19 +128,31 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+      saveAddress,
+    } = this.state
 
     if (shippingOption === "SHIP") {
       if (this.isCreateNewAddress()) {
         // validate when order is not pickup and the address is new
-        const { errors, hasErrors } = validateAddress(this.state.address)
-        const { error, hasError } = validatePhoneNumber(this.state.phoneNumber)
+        const { errors, hasErrors } = validateAddress(address)
+        const { error, hasError } = validatePhoneNumber(phoneNumber)
```
pepopowitz, line 155: 👍   Nice catches here, thank you for cleaning them up!

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -183,14 +194,12 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
         ? address
         : convertShippingAddressForExchange(
             // @ts-expect-error STRICT_NULL_CHECK
-            this.getAddressList()[parseInt(this.state.selectedSavedAddress)]
-              .node
+            this.getAddressList()[parseInt(selectedSavedAddress)].node
```
pepopowitz, line 197: Does anything bad happen here with `parseInt` when `selectedSavedAddress === NEW_ADDRESS`? L133 sets it to that value.

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -420,6 +409,8 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
                     onSelect={value => onSelectSavedAddress(value)}
                     handleClickEdit={this.handleClickEdit}
                     inCollectorProfile={false}
+                    showModal={showModal}
```
pepopowitz, line 412: Related to my previous comment, I don't think this is actually used anywhere within SavedAddresses, and management of it here could be removed?

pepopowitz: In fact, I think that cleanup could go further -- `handleClickEdit` and `this.state.editAddressIndex` I don't think are used in this component now that the editing happens within `SavedAddresses`.

src/v2/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -619,10 +622,13 @@ describe("Shipping", () => {
...
         expect(page.find(ShippingRoute).state().editAddressIndex).toBe(-1)
-        await page.find(`[data-test="edit-address"]`).first().simulate("click")
+        await page
+          .find(`[data-test="editAddressInShipping"]`)
+          .first()
+          .simulate("click")
         expect(page.find(ShippingRoute).state().editAddressIndex).toBe(0)
 
-        expect(page.find("AddressModal").length).toBe(2)
+        expect(page.find("AddressModal").length).toBe(1)
```
pepopowitz, line 631: 😅  I'm not sure why we were previously expecting 2 of these modals, but this assertion makes me feel much more comfortable now. 👍

pepopowitz (commented):
Thanks @Serge0n !!! I'm jumping in here to review based on @sweir27's request in slack, so I apologize if my comments are missing context.

I _think_ there's more cleanup we could do, and I have a couple questions that might prompt additions to this PR.

src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -171,23 +181,9 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
 
   const addAddressButton = (
     <>
-      {addressList.length > 0 ? (
+      {inCollectorProfile ? (
```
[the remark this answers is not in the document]
Serge0n: Yes, If Users not in their profile and no saved addresses - no need to show the button. In this situation, we show opened address form.

src/v2/Apps/Order/Components/__tests__/SavedAddresses.jest.tsx
Serge0n: Yes, you are right. Done)

src/v2/Apps/Order/Components/SavedAddresses.tsx
Serge0n: Sorry, my mistake. Done

src/v2/Apps/Order/Routes/Shipping/index.tsx
Serge0n: selectedSavedAddress === NEW_ADDRESS, means that need to show a clean Address form when you deleted all your saved addresses.

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
Serge0n: I think it's done now)

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -24,40 +25,112 @@ const savedAddress: SavedAddressType = {
...
+      ...testAddressModalProps,
+      modalDetails: {
+        addressModalTitle: "Add address",
+        addressModalAction: "createUserAddress",
+      },
+    })
+    expect(wrapper.text()).toContain("Add address")
     expect(wrapper.find("input").length).toBe(7)
     expect(wrapper.find("select").length).toBe(1)
+    expect(wrapper.find("Text[data-test='deleteButton']").length).toBe(0)
```
pepopowitz, line 77: non-blocking:

This particular line is the only unique assertion in this test; everything else is also covered by the previous test. I personally find it helpful to remove setup & assertions that aren't unique to a test, so that when someone is reading this test in the future, they can quickly identify the unique scenario it is testing. 

I'm not suggesting you change it for this PR to be merged; just pointing out that duplication in tests like this can mask the intention of the test.

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -24,40 +25,112 @@ const savedAddress: SavedAddressType = {
...
+    deleteButton.simulate("click")
+
+    setTimeout(() => {
+      expect(dialog).toHaveLength(1)
+      expect(dialog.props().title).toBe("Delete address?")
+      expect(dialog.props().detail).toBe(
+        "This will remove this address from your saved addresses."
+      )
+      expect(dialog.props().primaryCta.text).toContain("Delete")
+      expect(dialog.props().secondaryCta.text).toContain("Cancel")
```
pepopowitz, line 95: non-blocking

I like the pattern we used above for looking for text on a component -- `expect(wrapper.text()).toContain("Edit address")` -- it is a bit more robust in that there are less chances for the test to break due to an implementation change like renaming a prop. 

Again, I'm not suggesting this change must be made to merge the PR; just suggesting that avoiding checking of `props` directly in an assertion eliminates one more surface for tests breaking due to an unimportant change.

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -24,39 +25,112 @@ const savedAddress: SavedAddressType = {
...
+
+  it("when the dialog is cancelled, the delete action doesn't happen", () => {
+    const wrapper = getWrapper(testAddressModalProps)
+    const deleteButton = wrapper.find("Text[data-test='deleteButton']")
+    deleteButton.simulate("click")
+    const dialog = wrapper.find(Dialog)
+    const dialogCancel = dialog
+      .findWhere(node => node.text() === "Cancel")
+      .first()
+    console.log(dialogCancel.length)
```
pepopowitz, line 123:
```suggestion
```

😄

src/v2/Apps/Order/Routes/Shipping/index.tsx
pepopowitz: Ah -- now I see that these lines that do `parseInt(selectedSavedAddress)` are all guarded by the check to `this.isCreateNewAddress()`, so they'll never be hit when `selectedSavedAddress === NEW_ADDRESS`. Thanks for clarifying!

pepopowitz (changes requested):
Thanks for the changes @Serge0n! They look great.

I'm requesting one more change, regarding using the `Clickable` component for text that acts like a button. I'll reassign to you to merge once that's resolved, as I'm guessing you won't want to wait until I sign on late in your day 😄

PR #7799, opened 2021-06-16
feat: Preload default artwork image to speed up LCP GRO-382

jonallured:
This PR adds a preload tag for the default artwork image.

## Artwork Images Confuse Me

I figured what you see here in this diff would be all that's required for this task, but I'm confused because the image urls don't match:

```js
$('img[data-is-default=true]').src
"https://d32dm0rphc51dk.cloudfront.net/RtSWWtWGDI11QPzBuW4dsA/large.jpg"
$('link[as=image]').href
"https://d196wkiy8qx2u5.cloudfront.net/?resize_to=fit&width=640&height=394&quality=80&src=https%3A%2F%2Fd32dm0rphc51dk.cloudfront.net%2FRtSWWtWGDI11QPzBuW4dsA%2Flarge.jpg"
```

So there's something where the image url I have access to is a meta image? @dzucconi do you know what's going on here?

https://artsyproduct.atlassian.net/browse/GRO-382

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
@@ -37,6 +37,7 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
     if (is_shareable && imageURL) {
       return (
         <>
+          <Link rel="preload" as="image" href={imageURL} />
           <Meta property="twitter:card" content="summary_large_image" />
           <Meta property="og:image" content={imageURL} />
           {/* @ts-expect-error STRICT_NULL_CHECK */}
--- a/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
@@ -37,7 +37,6 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
     if (is_shareable && imageURL) {
       return (
         <>
-          <Link rel="preload" as="image" href={imageURL} />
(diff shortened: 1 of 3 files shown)


dzucconi:
Yeah: I'd include this tag directly in the lightbox component: https://github.com/artsy/force/blob/ef1c65306afd82fa67172a527522933cfb5bc384/src/v2/Components/Lightbox/Lightbox.tsx#L318

When this component gets rebuilt we're going to be dynamically resizing the image anyway so this will drift if it's separate.

jonallured:
Ok I moved it with a4e0d47 - is that what you were thinking @dzucconi?

dzucconi:
I think so: does the rendered HTML look OK to you? I also think what this is going to do is swap out the link tag when the image changes (when there are multiple images); which isn't what we want but also probably wouldn't harm anything...

jonallured:
Well this looks better to me!

```js
$('link[as=image]').href
"https://d32dm0rphc51dk.cloudfront.net/RtSWWtWGDI11QPzBuW4dsA/large.jpg"
$('img[data-is-default=true]').src
"https://d32dm0rphc51dk.cloudfront.net/RtSWWtWGDI11QPzBuW4dsA/large.jpg"
```

And yeah it's rendering all those images instead of just the default:

Screen Shot 2021-06-16 at 10 41 00 AM

Could avoid that by checking for `isDefault` - worth it?

dzucconi:
Yeah, we definitely don't want to preload ALL the images, just the first one; so the `isDefault` guard should be 👍

PR #7800, opened 2021-06-16
fix: Show spinning during redirect

anastasiapyzhik:
[FX-1846](https://artsyproduct.atlassian.net/browse/FX-1846)

Before
before

After
after


dblandin (approved):
Looks good! And awesome that we were able to clean up a few of those `@ts-expect-error` lines! 💯

damassi:
Why didn't we do this long ago, so much nicer 👍

dblandin:
> Why didn't we do this long ago, so much nicer 👍

@ashfurrow Thank you for the ticket! Glad we finally got to this :)

PR #7802, opened 2021-06-16
chore: Update Sentry libs and refactor implementation

damassi:
This updates our sentry lib to the latest, and adds a handful of new pre-send error filter entries to the list to help clean up our error logs.

Also renames `global_client_setup` to `deprecated_global_client_setup` just so its clear that folks shouldn't be adding stuff to there. (It was confusing when updating sentry stuff as there's two locations.)


src/middleware.ts
```diff
@@ -19,9 +19,13 @@ import sharify from "sharify"
 import siteAssociation from "artsy-eigen-web-association"
 import timeout from "connect-timeout"
 import bodyParser from "body-parser"
-import RavenServer from "raven"
 import config from "./config"
 
+// NOTE: Previoiusly, when deploying new Sentry SDK to prod we quickly start to
+// see errors like "`CURRENT_USER` is undefined". We need more investigation
+// because this only appears in prod, under load, and seems fine on staging.
+import * as Sentry from "@sentry/node"
```
damassi, line 27: Wanted to leave this comment in place because of the issues we saw last time we tried to upgrade (which i think was 2019). Not sure if we'll still encounter problems but just to be safe.

src/dev.js
```diff
@@ -60,6 +60,10 @@ const wdm = webpackDevMiddleware(compiler, {
   },
 })
 
+wdm.waitUntilValid(() => {
+  console.log(`\n[Force] Booted on ${process.env.APP_URL}.\n`)
+})
```
damassi, line 65: Noticed this was missing post webpack-5 upgrade due to no more friendly errors plugin

src/lib/analytics/sentryFilters.ts
```diff
@@ -0,0 +1,23 @@
+export const IGNORED_ERRORS = [
+  "AbortError: Fetch is aborted",
+  "AbortError: Request signal is aborted",
+  "AbortError: The operation was aborted",
+  'Blocked a frame with origin "https://www.artsy.net" from accessing a cross-origin frame',
+  "cancelled",
+  "Cannot set headers after they are sent to the client",
```
joeyAghion, line 7: This one seems like it could hide legitimate server-side bugs.

src/lib/analytics/sentryFilters.ts
```diff
@@ -0,0 +1,23 @@
...
+  "Failed to fetch",
+  "Origin is not allowed by Access-Control-Allow-Origin",
+  "pktAnnotationHighlighter", // Pocket ios app errors on opening articles
+  "Request has been terminated Possible causes: the network is offline, Origin is not allowed by Access-Control-Allow-Origin",
+  "Request has been terminated Possible causes: the network is offline",
+  "Response timeout",
+  "SecurityError: Blocked a frame with origin", // Facebook scripts error on opening articles
+  "TypeError Qg.m(gpt/pubads_impl_2021052601)",
+  "undefined is not an object (evaluating 'g.toString')",
```
joeyAghion, line 16: And this one. (Or is the `g` here some known-bad client-side object?)

joeyAghion (approved):

src/lib/analytics/sentryFilters.ts
damassi: Looks like this is coming from braze, added a filter and will clear it in sentry settings.

src/dev.js
damassi: This plugin not being here is def a big hit on the DX as we no longer get nice in-browser overlays pointing at syntax errors, and other clear console output. Will explore in another PR as I think react-fast-refresh comes with something packaged in; otherwise, there's likely a webpack 5 fork of friendly errors

icirellik: Could you elaborate on what you mean here `this was missing post webpack-5`. The Webpack 5 upgrade did not add nor remove any plugins.

PR #7814, opened 2021-06-17
fix: spacing between artist filter options

tr-ann:
Updated artists filter options style to v3
Jira: [FX-3028](https://artsyproduct.atlassian.net/browse/FX-3028)

**Before**
image

**After**
image

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/ArtistsFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/ArtistsFilter.tsx
@@ -16,18 +16,19 @@ export interface ArtistsFilterProps {
   user?: User
 }
 
-const ArtistItem: React.FC<{
-  slug: string
-  name: string
-  followedArtistSlugs: string[]
-  isFollowedArtistCheckboxSelected: boolean
-  CheckboxProps: CheckboxProps
-}> = ({
+const ArtistItem: React.FC<
+  {
+    slug: string
+    name: string
+    followedArtistSlugs: string[]
+    isFollowedArtistCheckboxSelected: boolean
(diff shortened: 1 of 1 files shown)


src/v2/Components/ArtworkFilter/ArtworkFilters/ArtistsFilter.tsx
```diff
@@ -21,11 +21,13 @@ const ArtistItem: React.FC<{
   name: string
   followedArtistSlugs: string[]
   isFollowedArtistCheckboxSelected: boolean
+  CheckboxProps: CheckboxProps
```
dzucconi, line 24: You'd want to extend or intersect with `CheckboxProps` and then just spread `...rest` — Rather than trying to encapsulate the props in another prop.

```tsx
{  slug: string  name: string  followedArtistSlugs: string[]  isFollowedArtistCheckboxSelected: boolean } & CheckboxProps
```

```tsx
({  slug,  name,  followedArtistSlugs,  isFollowedArtistCheckboxSelected,  ...rest }) => {
```

```tsx
    <Checkbox
      selected={
        currentlySelectedFilters().artistIDs.includes(slug) ||
        (isFollowedArtistCheckboxSelected && isFollowedArtist)
      }
      onSelect={selected => {
        return toggleArtistSelection(selected, slug)
      }}
      {...rest}
    >
```

-----

Alternatively — it'd be also fine to just recall the theme config within this component itself.

tr-ann: I like the idea of incapsulating props, thank you

dblandin (approved):
LGTM 👍🏼

PR #7815, opened 2021-06-17
Style fixes for inbox

lilyfromseattle:
[PURCHASE-2748]

There are broken styles on Inbox from the new site header and sitewide margins, as well as others

- scrollbar where there shouldn't be

- extra margins pushing sidebar inward/looking weird

- breakpoint issues - header disappears at one point

- The conversation list header and the conversation header are slightly misaligned with on another

_Before:_
Screen Shot 2021-06-16 at 3 24 32 PM

_After:_
Screen Shot 2021-06-16 at 3 24 46 PM

[PURCHASE-2748]: https://artsyproduct.atlassian.net/browse/PURCHASE-2748

### DIFF
--- a/src/v2/Apps/Conversation/Routes/Conversation/index.tsx
+++ b/src/v2/Apps/Conversation/Routes/Conversation/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Title } from "@artsy/palette"
+import { Box, Flex, FullBleed, Title } from "@artsy/palette"
 import { Conversation_me } from "v2/__generated__/Conversation_me.graphql"
 import { ConversationPaginationContainer as Conversation } from "v2/Apps/Conversation/Components/Conversation"
 import { ConversationListPaginationContainer as ConversationList } from "v2/Apps/Conversation/Components/ConversationList"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Conversation/Routes/Conversation/index.tsx
```diff
@@ -16,7 +16,7 @@ interface ConversationRouteProps {
 }
 
 const ConstrainedHeightContainer = styled(Box)`
-  height: calc(100vh - 60px);
+  height: calc(100vh - 103px);
```
xtina-starr, line 19: What does the `103px` in this calculation account for? I ask because 103 feels a bit more arbitrary than 60 and it makes wonder what it represents?

xtina-starr (approved):
I had one minor question which isn't a launch blocker. Feel free to merge this if you're ready.

src/v2/Apps/Conversation/Routes/Conversation/index.tsx
lilyfromseattle: Its to accommodate he new header. `100px` left a gap

PR #7816, opened 2021-06-17
fix: removes Zendesk script from auctions pages

ashleyjelks:
As a followup to #7780 the Zendesk scripts were being mounted twice on some auctions pages, this PR is a fix for that and should wrap up the work for https://artsyproduct.atlassian.net/browse/PURCHASE-2477.

### DIFF
--- a/src/v2/Apps/Auction/Routes/ConfirmBid/index.tsx
+++ b/src/v2/Apps/Auction/Routes/ConfirmBid/index.tsx
@@ -37,6 +37,8 @@ import {
   toStripeAddress,
 } from "v2/Apps/Auction/Components/Form"
 import { createStripeWrapper } from "v2/Utils/createStripeWrapper"
+import { ZendeskWrapper } from "v2/Components/ZendeskWrapper"
+import { data as sd } from "sharify"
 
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Auction/Routes/ConfirmBid/index.tsx
```diff
@@ -315,6 +317,12 @@ export const ConfirmBidRoute: React.FC<
     }
   }
 
+  function renderZendeskScript() {
+    if (typeof window !== "undefined" && window.zEmbed) return
+
+    return <ZendeskWrapper zdKey={sd.AUCTION_ZENDESK_KEY} />
```
damassi, line 323: It looks like the reason this issue happened is because `ZendeskWrapper` has no types in the file, so there was no error stating that `zdKey` was missing (and so it returned `null`. 

If the key is _actually_ null then we should def return null from the component so that things don't blow up, but we'll want to add `React.FC<{ zdKey: string }>` (a required prop) to ZendeskWrapper so that at the very least we'll get a type error if the key isn't being passed.

starsirius (approved):
👍

xtina-starr (approved):
👍

PR #7826, opened 2021-06-21
feat: Ensure ip address is never null GRO-383

jonallured:
We were seeing noise in our Sentry logs where sometimes Force would send an empty IP address to MP so this PR ensures we have a fallback. This fallback will result in the county coming back as empty:

Screen Shot 2021-06-21 at 9 11 55 AM

So that seems like it solves the problem but maybe there's a better way to not make this query at all? Open to other/better ideas here!

https://artsyproduct.atlassian.net/browse/GRO-383

/cc @artsy/grow-devs


The-Beez-Kneez:
We could do like we were doing before, and have the `country-code` default to a GDP designated country. We are only using the `country-code` to trigger the boolean on hiding or showing a certain sign up flow, so no functionality would be altered other than that.

Would be best to default to the state that would legally protect us (GDPR opt in).

PR #7811, opened 2021-06-17
chore: (GRO-381) remove sign in with apple feature flag

ovasdi:
https://artsyproduct.atlassian.net/browse/GRO-381

## Primary change
This removes the ENV var and supplementary logic (including tests) that put _sign in with apple_ behind a feature flag. Here is the [original PR](https://github.com/artsy/force/pull/5251) and the [specific commit](https://github.com/artsy/force/pull/5251/commits/6d11afa33db7535206323cef50091af34dcafd32).

## Secondary change
When committing my changes I experienced an error during the pre-commit hook that executes `yarn lint --fix`:

```sh
error  Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid
```

I reached out to [#dev-help](https://artsy.slack.com/archives/CP9P4KR35/p1623862247122500) and the recommendation was to replace the `<Link>` component with `<Clickable>`.  The scope of this change affects only `src/v2/Components/Authentication/Footer.tsx` elements. This change also affected a number of tests that were updated in this PR. I tried to keep the behavior and look exactly how it was before.

## Demonstration of changes

| staging login | local login |
|-|-|
| staging-login | local-login |

| staging signup | local signup |
|-|-|
| staging-signup | local-signup |

---

Lastly there was an extra space after the word Facebook which was removed in both Login and Signup auth footer.

Screen Shot 2021-06-17 at 1 37 38 PM

[Link to a review app](https://remove-appl-signup-ff.artsy.net/)

PR #7832, opened 2021-06-21
fix: (GRO-384) jsx-a11y/anchor-is-valid linter error in auth footer

ovasdi:
This PR contains the "_Secondary Change_" part of a reverted PR [#7821](https://github.com/artsy/force/pull/7821).

From the [description of reverted PR](https://github.com/artsy/force/pull/7811):

> When committing my changes I experienced an error during the pre-commit hook that executes yarn lint --fix:

```
error  Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid
```

> I reached out to [#dev-help](https://artsy.slack.com/archives/CP9P4KR35/p1623862247122500) and the recommendation was to replace the `<Link>` component with `<Clickable>`. The scope of this change affects only src/v2/Components/Authentication/Footer.tsx elements. This change also affected a number of tests that were updated in this PR. I tried to keep the behavior and look exactly how it was before.

Find the [review app here](https://jsx-a11y.artsy.net/).
[Integrity change that goes along with this](https://github.com/artsy/integrity/pull/233).

### DIFF
--- a/src/v2/Components/Authentication/Footer.tsx
+++ b/src/v2/Components/Authentication/Footer.tsx
@@ -1,8 +1,9 @@
-import { Box, Flex, Link } from "@artsy/palette"
+import { Box, Flex, Clickable, Sans } from "@artsy/palette"
 import React from "react"
-import { data as sd } from "sharify"
+import styled from "styled-components"
 import { CaptchaTerms, FooterText } from "./commonElements"
 import { ModalType } from "./Types"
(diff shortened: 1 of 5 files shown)


damassi:
LGTM 👍

PR #7835, opened 2021-06-21
GRO-376: Artist has displayed CV even without Career Highlights

The-Beez-Kneez:
This is an update to fix the bug noticed on the Artist page. The career highlights portion usually includes a link to a CV for artists to view "See all past shows and fair booths" link. But artists without career highlights were not able to have this CV link.

This adds in the "See all past shows and fair booths" link for artists regardless if they have highlights or not.

Past Behavior:
Screen Shot 2021-06-21 at 2 46 22 PM

New Behavior: 
Screen Shot 2021-06-21 at 2 33 41 PM

### DIFF
--- a/src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
+++ b/src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
@@ -104,6 +104,21 @@ describe("SelectedCareerAchievements", () => {
 
     expect(text.length).toEqual(0)
   })
+
+  it("renders the Artists CV RouterLink regardless of career achievements", async () => {
+    // @ts-expect-error STRICT_NULL_CHECK
+    wrapper = await getWrapper({
+      ...artistResponse,
+      auctionResultsConnection: null,
+      highlights: {
+        // @ts-expect-error STRICT_NULL_CHECK
+        ...artistResponse.highlights,
+        partnersConnection: null,
+      },
+      insights: null,
+    })
+    expect(wrapper.find("RouterLink").length).toBe(1)
+  })
 })
 
 const artistResponse: SelectedCareerAchievementsTestQueryRawResponse["artist"] = {
--- a/src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
+++ b/src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
@@ -69,7 +69,7 @@ describe("SelectedCareerAchievements", () => {
     expect(text).toContain("Included in a major biennial")
   })
 
-  // TODO
+  // TODO https://artsyproduct.atlassian.net/browse/GRO-393
   it("doesn't render selected career achievements if no auction results, partner highlights, or insights", async () => {
     // @ts-expect-error STRICT_NULL_CHECK
     wrapper = await getWrapper({
@@ -87,7 +87,7 @@ describe("SelectedCareerAchievements", () => {
     expect(text.length).toEqual(0)
   })
 
-  // TODO
+  // TODO https://artsyproduct.atlassian.net/browse/GRO-393
   it("doesn't render selected career achievements if no auction results or partner highlights and insights is null", async () => {
     // @ts-expect-error STRICT_NULL_CHECK
     wrapper = await getWrapper({
@@ -105,7 +105,8 @@ describe("SelectedCareerAchievements", () => {
     expect(text.length).toEqual(0)
   })
 
-  it("renders the Artists CV RouterLink regardless of career achievements", async () => {
+  // TODO https://artsyproduct.atlassian.net/browse/GRO-393
+  it("renders the Artists CV link regardless of career achievements", async () => {
     // @ts-expect-error STRICT_NULL_CHECK
     wrapper = await getWrapper({
       ...artistResponse,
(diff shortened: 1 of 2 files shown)


jonallured (commented):
Wow, this component is complex! I think the changes here make sense but I'd love to see a test - what do you think?

The-Beez-Kneez:
Added in a test for that @jonallured

jonallured (approved):

damassi:
Oh sorry for the drive by, caught a couple things! (commenting inline)

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
```diff
@@ -104,6 +104,21 @@ describe("SelectedCareerAchievements", () => {
 
     expect(text.length).toEqual(0)
   })
+
+  it("renders the Artists CV RouterLink regardless of career achievements", async () => {
+    // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 109: These comments were added automatically by a script when we were migrating to strict type checking and needed a way to get the type-checker to pass without fixing every error that came up, so we should never add more, but rather address (and fix) the type error.

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
```diff
@@ -104,6 +104,21 @@ describe("SelectedCareerAchievements", () => {
...
+      ...artistResponse,
+      auctionResultsConnection: null,
+      highlights: {
+        // @ts-expect-error STRICT_NULL_CHECK
+        ...artistResponse.highlights,
+        partnersConnection: null,
+      },
+      insights: null,
+    })
+    expect(wrapper.find("RouterLink").length).toBe(1)
```
damassi, line 120: As there could be additional `RouterLink`s added in the future (which would then fail this test), we can make the test more precise by checking that the router-link's URL and text matches what was added in the PR. 

Can get at `props` (referring to `to`) via enzyme's [props()](https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/props.html) method and can get at the router link's text via `text()`.

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
The-Beez-Kneez: @damassi  Do you have a suggestion on how to fix this typing? They are in two of the above unit tests as well, that you wrote from a year ago.

damassi: I'm not sure, these `STRICT_NULL_CHECK` were added a month or so ago, automatically via a script, irrespective of any other tests or todos or tickets

damassi: > Do you have a suggestion on how to fix this typing

What is the error?

The-Beez-Kneez: Argument of type '{ auctionResultsConnection: null; highlights: { partnersConnection: null; }; insights: null; slug?: string | undefined; id?: string | null | undefined; }' is not assignable to parameter of type '{ readonly highlights: { readonly partnersConnection: { readonly edges: readonly ({ readonly node: { readonly categories: readonly ({ readonly slug: string; readonly id: string | null; } | null)[] | null; readonly id: string | null; } | null; readonly id: string | null; } | null)[] | null; } | null; } | null; readon...'.
  Types of property 'slug' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.ts(2345)

The-Beez-Kneez: If this is something that is widespread, and is out of scope for this fix, I am going to add a JIRA ticket to this TODO so that we can track it and add in this change.

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
The-Beez-Kneez: Updated to look for the specific `<RouterLink>` url via text.

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
damassi: In that case we can use `// @ts-ignore` with a FIXME pointing at the ticket 👍 (not sure what the issue is with the test, but its no big deal)

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
damassi: Looks good 👍

src/v2/Components/__tests__/SelectedCareerAchievements.jest.tsx
damassi: The larger point is that anytime we see a _new_ 
```
```
in code in a PR we can be sure that something has been copied from somewhere so we'll want to fix, or replace with `// @ts-ignore - FIXME: do something` 

In fact, any new `ts-expect-error` additions should fail on CI, but not sure why that isn't happening. Need to look into it.

damassi (approved):

PR #7841, opened 2021-06-22
feat: artsy.net cookie consent.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3176

Add OneTrust javascripts and attempt to make them fire first on page load. This is to implement [Cookie Consent](https://www.notion.so/artsy/Cookie-Consent-01dace52a765476ea4e9353260ce9a7f). The tags are copied from [OneTrust's UI](https://app-eu.onetrust.com/cookies/websites).  Over there we configure consent banner, preference-center, cookie-blocking, and other things. The scripts implement those settings. Generally, changes on their UI take effect only when the scripts are re-published, here:

Screenshot from 2021-08-30 16-19-51

There  are 2 set of scripts, `Test` and `Production`. They differ only in an ID which is parameterized into the `ONETRUST_SCRIPT_ID` ENV var whose presence in config enables the scripts.

- `Test` tag ID: `12abb958-ed18-407e-8829-ea60cdd69a86-test` <--- we will use for staging
- `Productio` tag ID: `12abb958-ed18-407e-8829-ea60cdd69a86` <--- use for production

After this PR is merged, we will enable the tag in staging by:

```
hokusai staging env set ONETRUST_SCRIPT_ID=12abb958-ed18-407e-8829-ea60cdd69a86-test
```
Then we will announce to a broader (than previous announcements) audience and have them test on staging.

Once that's cleared, we will enable in production by:

```
hokusai staging env set ONETRUST_SCRIPT_ID=12abb958-ed18-407e-8829-ea60cdd69a86
```

### DIFF
--- a/src/v2/index.ejs
+++ b/src/v2/index.ejs
@@ -2,6 +2,18 @@
 <html>
 
 <head>
+
+  <!-- onetrust cookie consent, must be the first js loaded -->
+  <%% if ( !disable.onetrust && !sd.THIRD_PARTIES_DISABLED && sd.ONETRUST_SCRIPT_ID ) { %>
+    <script src="https://cdn.cookielaw.org/consent/<%%= sd.ONETRUST_SCRIPT_ID %>/OtAutoBlock.js" type="text/javascript" >
+    </script>
+    <script src="https://cdn.cookielaw.org/scripttemplates/otSDKStub.js"  type="text/javascript" charset="UTF-8" data-domain-script="<%%= sd.ONETRUST_SCRIPT_ID %>" >
+    </script>
+    <script type="text/javascript">
+      function OptanonWrapper() { }
+    </script>
+  <%% } %>
+
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
   <meta charset='utf-8' />
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -118,6 +118,7 @@ export const MOBILE_MEDIA_QUERY: any = "only screen and (max-width: 640px)"
 export const NETWORK_CACHE_SIZE: any = 2000
 export const NETWORK_CACHE_TTL: any = 3600000
 export const NODE_ENV: any = "development"
+export const ONETRUST_SCRIPT_ID: any = null
 export const OPENREDIS_URL: any = null
 export const PAGE_CACHE_ENABLED: any = false
 export const PAGE_CACHE_EXPIRY_SECONDS: any = 600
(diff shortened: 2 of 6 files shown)


src/desktop/components/main_layout/templates/scripts.jade
```diff
@@ -1,4 +1,4 @@
-- defaultOptions = {grecaptcha: true, stripev3: false, sailthru: true, marketo: true}
+- defaultOptions = {grecaptcha: true, marketo: true, onetrust: true, sailthru: true, stripev3: false}
```
artsyjian, line 1: how do we over-ride these default options? via ENV?

joeyAghion: Not sure. @icirellik ?

damassi: `options` are overwritten on a template-by-template basis, and merged in via [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) below. 

If you search force for `-options` things should pop up:

Screen Shot 2021-06-22 at 11 47 06 AM

src/v2/index.ejs
```diff
@@ -5,6 +5,17 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
   <meta charset='utf-8' />
 
+  <!-- onetrust cookie consent, must be the first js loaded -->
+  <%% if ( !sd.THIRD_PARTIES_DISABLED && sd.ONETRUST_SCRIPT_ID ) { %>
```
damassi, line 9: @icirellik can chime in on the necessity of this, but we've been adding all 3rd party scripts to this additional `disable` object so that we can toggle off and on specifically: https://github.com/artsy/force/blob/master/src/v2/server.ts#L97

artsyjian: oh. what's the use case? is it just for local dev?

damassi: I _think_ this is lighthouse related, for narrowing down on script that impact performance.

artsyjian: i see. thanks. @icirellik - i will add if helps.

icirellik: Yup, it was added as a way to disable locally and possibly in review apps.

icirellik: There are ways to disable scripts. `THIRD_PARTIES_DISABLED` is meant to disable all 3rd party scripts for testing locally and possibly in a review app.  What @damassi is pointing out is used to disable individual 3rd party scripts using a URL parameter that is Calibre uses.

artsyjian: thanks. added to `disable` object.

artsyjian:
Revamped PR. It's finally ready for final review!

mc-jones:
Looks great! 🚢

mc-jones (approved):

artsyjian:
Update: We switched staging.artsy.net to use OneTrust staging domain's `TEST` tag, due to [problems with cookie blocking](https://artsy.slack.com/archives/CA8SANW3W/p1630511029124100). That tag's ID is:
```
artsy:force jxu$ hokusai staging env get | grep -i onetrust
ONETRUST_SCRIPT_ID=e23bc15c-797a-4376-86af-5b3481ed6a1e-test
```

PR #7842, opened 2021-06-22
fix (CX-1618): change label name for auction house filter

kalshannikava:
Addresses [CX-1618]

**Description**
The label for the "Auction house" filter for "auction results" should be "Auction house" instead of "medium".
The bug was introduced in this PR: https://github.com/artsy/force/pull/7719/files

**The issue**
image

**The fix**
Screenshot 2021-06-22 at 17 48 05

[CX-1618]: https://artsyproduct.atlassian.net/browse/CX-1618


olerichter00:
I wonder if it should be "Auction House" instead of "Auction house" to be consistent with "Year Created" 🤔

damassi (approved):
> I wonder if it should be "Auction House" instead of "Auction house" to be consistent with "Year Created"

Seconding @olerichter00 on this! Once that's fixed this can be merged on 🍏

olerichter00 (approved):

PR #7851, opened 2021-06-23
feature: Stub out new /auctions v2 routes

damassi:
This stubs out a few new routes for the auctions page and adds a couple more tabs (temporarily, while we're building this out). As this currently in development, lets hide behind an `.env` flag for now:

```
ENABLE_NEW_AUCTIONS_FILTER=true
```

Screen Shot 2021-06-22 at 10 43 53 PM


src/v2/Apps/Auctions/auctionsRoutes.tsx
```diff
@@ -11,6 +11,24 @@ const AuctionsApp = loadable(
...
+      /* webpackChunkName: "auctionBundle" */ "./Routes/Auctions/AuctionsRoute"
+    ),
+  {
+    resolveComponent: component => component.AuctionsRouteFragmentContainer,
+  }
+)
+const ArtworksRoute = loadable(
+  () =>
+    import(
+      /* webpackChunkName: "auctionBundle" */ "./Routes/Auctions/AuctionsRoute"
```
ovasdi, line 26: Just wondering if this should be `./Routes/Artworks/ArtworksRoute`?

ovasdi (approved):
Left one question otherwise this looks good.

So is the "Current Auctions" tab being replaced by "Auctions" tab that you are laying ground for here?

src/v2/Apps/Auctions/auctionsRoutes.tsx
```diff
@@ -11,6 +11,24 @@ const AuctionsApp = loadable(
...
+    ),
+  {
+    resolveComponent: component => component.AuctionsRouteFragmentContainer,
```
ovasdi, line 29: Also wondering if this should be `component.ArtworksRouteFragmentContainer`?

src/v2/Apps/Auctions/auctionsRoutes.tsx
damassi: Ah, good eye, will update the variable name 👍

damassi:
> So is the "Current Auctions" tab being replaced by "Auctions" tab that you are laying ground for here?

Yup! eventually the current/upcoming/past tabs will go away and the two new stubbed ones will be the remainders.

PR #7855, opened 2021-06-23
fix: Added terms of showing fair booth

anastasiapyzhik:
[FX-2727](https://artsyproduct.atlassian.net/browse/FX-2727)

Unfortunately, there are no suitable examples for the demo. We need to wait until upcoming shows come up to test these changes


dblandin:
Closing as it seems like this was resolved with [FX-2535].

PR #7857, opened 2021-06-23
feat(GRO-214): adds featured articles rail

dzucconi:
Simple enough, however we don't yet expose the `featured` flag on the articles field in Metaphysics — we'll have to follow up with that.


src/v2/Apps/Home/HomeApp.tsx
```diff
@@ -44,6 +45,8 @@ export const HomeApp: React.FC<HomeAppProps> = ({ homePage }) => {
         <HomeArtworkModulesFragmentContainer homePage={homePage} />
 
         <HomeFeaturedShowsLazyQueryRenderer />
+
+        <HomeFeaturedArticlesLazyQueryRenderer />
```
damassi, line 49: I wonder if i could copy these shows / articles rails over to /artist/id 🤔

PR #7859, opened 2021-06-23
Retire <meta... fragment> tags and AJaX pre-rendering protocol (PLATFORM-3363)

joeyAghion:
First step of retiring AJaX pre-rendering support: remove `<meta name="fragment content="!">` tags from the few legacy pages that still render them.

As part of a later release, we can remove the `proxyReflection` middleware.

Co-authored by: @mc-jones

https://artsyproduct.atlassian.net/browse/PLATFORM-3363


ovasdi (approved):
:shipit:

PR #7860, opened 2021-06-23
feat(GRO-214): remaining rails

dzucconi:
Gets the rest of the rails in. I had some faulty assumptions about them, the only weird one is active bids, which we'll replace with the MyBids implementation next.

### DIFF
--- a/src/v2/Apps/Home/Components/HomeArtworkModuleContext.tsx
+++ b/src/v2/Apps/Home/Components/HomeArtworkModuleContext.tsx
@@ -51,6 +51,7 @@ const HomeArtworkModuleContext: React.FC<HomeArtworkModuleContextProps> = ({
         </Flex>
       )
     }
+
     case "Sale": {
       return (
         <Flex justifyContent="space-between">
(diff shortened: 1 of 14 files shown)


src/v2/Apps/Home/Components/HomeFeaturedEvents.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  if (events.length === 0) return null
+
+  return (
+    <>
+      <Text variant="xl">Featured events</Text>
+
+      <Spacer mt={4} />
+
+      <Shelf>
```
damassi, line 27: Any reason this component doesn't have `Rail` in the name?

damassi (approved):
One comment about the name, otherwise LGTM 👍

src/v2/Apps/Home/Components/HomeFeaturedEvents.tsx
dzucconi: Ah, I think when I made it I thought it wasn't a rail until I looked later. I'll rename and merge 👍

PR #7861, opened 2021-06-23
(HOLD) patch: React Spring - WeakSet

icirellik:
Update react-spring to use a WeakSet so that it doesn't indefinitely
hold references to each animation controller it creates. The controllers
are only cleaned up during `componentWillUnmount` which is why we
are seeing this leak during SSR.

### DIFF
--- a/HACKS.md
+++ b/HACKS.md
@@ -14,6 +14,18 @@ Tell us when we can remove this hack.
 
 > 👀 When adding a new item, see comment on top of file for template.
 
+## patches/react-spring+8.0.27.patch
+
+#### Explanation/Context:
+
+The `react-spring` library leaks locally due to a cleanup routine that only executes in `componentWillUnmount`, this patches the controller cache `Set` to be a `WeakSet` so that references are no longer held onto when the components have been garbage collected.
+
+Slack: https://artsy.slack.com/archives/CA8SANW3W/p1624453974296000
+
+#### When can we remove this:
+
+When we have removed the references to `react-spring` from `@artsy/palette`
+
 ## patches/relay-mock-network-layer+2.0.0.patch
 
 #### Explanation/Context:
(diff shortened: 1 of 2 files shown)


damassi (approved):
Looks good @icirellik - one request before merge, however: would you mind creating a HACKS.md file and dropping in a note about this new patch / link to slack thread? See https://github.com/artsy/README/issues/390#issuecomment-859962331

damassi:
Thanks for updating hacks doc. That actually made me realize that we're patching this in force (which already has one) vs Palette. Any reason we're applying it here to a transitive dep instead of where it's a direct dependency (over in palette)?

icirellik:
It's easier to test and won't break any other downstream projects. It's beginning to look like the solution is to remove `react-spring` node 12s implementation of weak references isn't complete enough to make this work properly with jest.

damassi:
> It's beginning to look like the solution is to remove react-spring node 12s implementation of weak references isn't complete enough to make this work properly with jest.

All good, so no patch necessary? 

> It's easier to test

It might be easier, but we still shouldn't patch transitive deps when we explicitly add it to `@artsy/palette`, and not to force. It's confusing and unnecessary. 

FWIW, steps to test this between palette and force:

1) `cd palette/packages/palette`
1) edit stuff 
1) `yarn patch-package`
1) `yarn link`
1) `yarn compile`
1) `cd force` 
1) `yarn link @artsy/palette` 
1) `yarn start`

icirellik:
The fix is necessary.

A patch cannot be the solution because node is lacking the nescesary support to make this change work with jest. You can also read this as Jestis the worst 😄

icirellik:
The patch won't work, need to remove `react-spring` properly. 😭

damassi:
> Unrelated, this DX is awful. Y'all do this regularly?

Thankfully not since most dev now takes place in storybooks 😄 Regarding monorepo, we could move all of this stuff together, but palette is more of its own thing, and we import it into any number of apps across artsy as our UI layer. Would feel a bit uncomfortable importing from force since it could just as well be Eigen's child.

icirellik:
The solution is simple! Move Eigen into Force as well 😄

PR #7862, opened 2021-06-23
sentry: Add dedupe lib and filter out issues

damassi:
Address voluminous issues around `Non-Error promise rejection captured with value: Object Not Found`, which are eating up much of our quota. See [example](https://sentry.io/organizations/artsynet/issues/2460521645/?project=28316&query=is%3Aunresolved&sort=freq&statsPeriod=90d).

Followed instructions from this [sentry thread](https://forum.sentry.io/t/unhandledrejection-non-error-promise-rejection-captured-with-value/14062/13), and added the [dedupe integration](https://docs.sentry.io/platforms/javascript/configuration/integrations/plugin/#dedupe) to help clean up logs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "@loadable/component": "5.14.1",
     "@loadable/server": "5.14.0",
     "@sentry/browser": "6.7.1",
+    "@sentry/integrations": "^6.7.2",
     "@sentry/node": "^6.7.1",
     "@sentry/tracing": "^6.7.1",
     "@stripe/react-stripe-js": "1.1.2",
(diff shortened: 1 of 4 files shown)


joeyAghion (approved):
Seems reasonable given the thread, and the wide variety of clients that might be generating these errors (including some without full JS support?!).

Note that dedupe might be a default again soon: https://github.com/getsentry/sentry-javascript/pull/3730

damassi:
> Note that dedupe might be a default again soon

Oh nice, i'll add a reminder to follow-up on that 👍

PR #7863, opened 2021-06-23
Retire proxying to Reflection for pre-rendered page HTML (PLATFORM-3364) #migration #wip

joeyAghion:
This follows up https://github.com/artsy/force/pull/7859 and removes the server-side proxying behavior that currently serves pre-rendered, JS-free HTML to specially crafted search crawler requests (`...?_escaped_fragment_=`). This shouldn't be merged until the earlier work has been released.

### Post-deploy migration:

```
hokusai [staging|production] env unset REFLECTION_URL
```

I'll leave a few questions inline...

https://artsyproduct.atlassian.net/browse/PLATFORM-3364


src/desktop/components/split_test/split_test.coffee
```diff
@@ -63,8 +60,6 @@ module.exports = class SplitTest
     outcome =
       if (@admin() and @edge?)
         @edge
-      else if @reflection()
-        if @control_group then @control_group else 'control'
```
joeyAghion, line 67: I actually like the idea of crawlers getting the "control" treatment of split tests, but it doesn't look like the code works that way. The `REFLECTION` value (below) is _only_ true for the specific `Artsy/Reflection` user-agent.

If there's a convenient way to identify crawlers here, we could keep this logic.

src/desktop/components/split_test/test/split_test.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-done-callback */
```
joeyAghion, line 1: Happy to convert the `beforeEach` below to a Promise if you can offer a pointer or example.

src/lib/middleware/proxyReflection.ts
```diff
@@ -1,50 +0,0 @@
-import type { NextFunction } from "express"
-import type { ArtsyRequest, ArtsyResponse } from "./artsyExpress"
-
-import httpProxy from "http-proxy"
```
joeyAghion, line 4: I hoped to drop some of these dependencies, but it looks like they're all used elsewhere.

damassi:
@joeyAghion - to be clear 
> This shouldn't be merged until the earlier work has been released.

Meaning, https://github.com/artsy/force/pull/7859 needs to be deployed, then this merged, and then another deploy goes out?

src/desktop/components/split_test/test/split_test.test.js
damassi: The `no done` rule is a bit extreme IMO /shrug. All good.

joeyAghion:
Yup I expect we'll continue to see the `?_escaped_fragment_=` requests for a short while after we remove the meta tags, so wanted to space these out. There's little actual risk though, since in the worst case we serve a JS-heavy page to a crawler who can't handle it. (Converted to draft for now.)

src/middleware.ts
```diff
@@ -93,8 +92,6 @@ export function initializeMiddleware(app) {
   // Sharify and async locals
   app.use(bootstrapSharifyAndContextLocalsMiddleware)
 
-  app.use(proxyReflectionMiddleware)
```
damassi, line 96: 🎉

damassi (approved):
LGTM 👍

joeyAghion:
Unset `REFLECTION_URL` in both environments.

PR #7869, opened 2021-06-24
feat(PX-4135): Collector can set an address as default

Serge0n:
The collector can set an address as default

Jira: https://artsyproduct.atlassian.net/browse/PX-4135

<details>
  <summary>Edit modal when isDefault true</summary>

https://user-images.githubusercontent.com/55637696/125309131-2b5c2700-e33a-11eb-93ea-8fb0e158f126.mov
</details>

<details>
  <summary>Edit modal when isDefault false</summary>

https://user-images.githubusercontent.com/55637696/125309297-50e93080-e33a-11eb-891e-920d9408d39d.mov
</details>

<details>
  <summary>Add modal(checkbox is always shown)</summary>

https://user-images.githubusercontent.com/55637696/125309822-ba693f00-e33a-11eb-98c4-0654d9a12cb7.mov
</details>

<details>
  <summary>Correct state when uncheck</summary>

https://user-images.githubusercontent.com/55637696/125309372-6199a680-e33a-11eb-86f5-16da6a34f9a4.mov
</details>

<details>
  <summary>How hover should work</summary>

https://user-images.githubusercontent.com/55637696/125445341-a91c8858-b8ea-45d7-8169-fb51ade8b8b4.mov
</details>

### DIFF
--- a/src/v2/Apps/Order/Components/AddressModal.tsx
+++ b/src/v2/Apps/Order/Components/AddressModal.tsx
@@ -105,7 +105,6 @@ export const AddressModal: React.FC<Props> = ({
             }
 
             const handleSuccess = savedAddress => {
-              console.log(savedAddress)
               if (isDefault) {
                 updateUserDefaultAddress(
                   // @ts-expect-error STRICT_NULL_CHECK
@@ -169,7 +168,7 @@ export const AddressModal: React.FC<Props> = ({
                     setIsDefault(selected)
                   }}
                   selected={formik.values?.isDefault}
-                  data-test="set-as-default-checkbox"
+                  data-test="setAsDefault"
                 >
                   Set as default
                 </Checkbox>
--- a/src/v2/Apps/Order/Components/AddressModal.tsx
+++ b/src/v2/Apps/Order/Components/AddressModal.tsx
@@ -23,6 +23,7 @@ import { createUserAddress } from "v2/Apps/Order/Mutations/CreateUserAddress"
 import { SavedAddresses_me } from "v2/__generated__/SavedAddresses_me.graphql"
 import { AddressModalFields } from "v2/Components/Address/AddressModalFields"
 import { useSystemContext } from "v2/System/SystemContext"
+import { updateUserDefaultAddress } from "../Mutations/UpdateUserDefaultAddress"
 export interface Props {
   show: boolean
   closeModal: () => void
@@ -75,7 +76,7 @@ export const AddressModal: React.FC<Props> = ({
   // @ts-expect-error STRICT_NULL_CHECK
   const [createUpdateError, setCreateUpdateError] = useState<string>(null)
   const [showDialog, setShowDialog] = useState<boolean>(false)
-
+  const [isDefault, setIsDefault] = useState<boolean>(false)
   return (
     <>
       <Modal
@@ -128,6 +129,14 @@ export const AddressModal: React.FC<Props> = ({
                   handleSuccess,
                   handleError
                 )
+            isDefault &&
+              updateUserDefaultAddress(
+                // @ts-expect-error STRICT_NULL_CHECK
+                relayEnvironment,
+                address.internalID,
+                onSuccess,
+                onError
+              )
           }}
         >
           {(formik: FormikProps<SavedAddressType>) => (
@@ -152,6 +161,7 @@ export const AddressModal: React.FC<Props> = ({
                 <Checkbox
                   onSelect={selected => {
                     formik.setFieldValue("isDefault", selected)
+                    setIsDefault(selected)
                   }}
                   selected={formik.values?.isDefault}
                   data-test="set-as-default-checkbox"
--- a/src/v2/Apps/Order/Components/AddressModal.tsx
+++ b/src/v2/Apps/Order/Components/AddressModal.tsx
@@ -1,4 +1,5 @@
 import React, { useState } from "react"
+import { Environment } from "react-relay"
 import {
   Button,
   Clickable,
@@ -27,15 +28,15 @@ import { updateUserDefaultAddress } from "../Mutations/UpdateUserDefaultAddress"
 export interface Props {
   show: boolean
   closeModal: () => void
-  address?: SavedAddressType
-  onSuccess: (address) => void
+  address: SavedAddressType
+  onSuccess: () => void
   onDeleteAddress: (addressID: string) => void
   onError: (message: string) => void
-  modalDetails?: {
+  modalDetails: {
     addressModalTitle: string
     addressModalAction: AddressModalAction
   }
-  me?: SavedAddresses_me
+  me: SavedAddresses_me
 }
 
 const SERVER_ERROR_MAP: Record<string, Record<string, string>> = {
@@ -61,7 +62,6 @@ export const AddressModal: React.FC<Props> = ({
   me,
 }) => {
   const title = modalDetails?.addressModalTitle
-  // @ts-expect-error STRICT_NULL_CHECK
   const createMutation = modalDetails.addressModalAction === "createUserAddress"
   const validator = (values: any) => {
     const validationResult = validateAddress(values)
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -102,10 +104,21 @@ export const AddressModal: React.FC<Props> = ({
               onError && onError(message)
             }
 
-            const handleSuccess = address => {
+            const handleSuccess = savedAddress => {
+              if (isDefault) {
+                updateUserDefaultAddress(
+                  // @ts-expect-error STRICT_NULL_CHECK
```
pepopowitz, line 110: blocking: this line was added in a bunch of places when we initially turned on strict null checking, but we should not be adding new ones. If this line is needed to prevent a linting error in this call, we should fix the arguments instead.

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -102,10 +104,21 @@ export const AddressModal: React.FC<Props> = ({
               onError && onError(message)
             }
 
-            const handleSuccess = address => {
+            const handleSuccess = savedAddress => {
+              if (isDefault) {
```
pepopowitz, line 108: Blocking if my understanding is correct: Do we not have access to the value through `values.isDefault`? It looks like we should ([here's the underlying type](https://github.com/artsy/force/blob/1db17a9b7ee6ff5c44401e0b46160d04cde68ab0/src/v2/__generated__/Shipping_me.graphql.ts#L20)). If we do, we should use that instead of creating another state variable to manage this value.

src/v2/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -48,15 +48,22 @@ export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
...
                   </Text>
                   {address?.isDefault && index === 0 && (
-                    <Pill ml={0.5} variant="textSquare" hover focus>
+                    <StyledPill
+                      ml={0.5}
+                      variant="textSquare"
+                      disabled
+                      hover
+                      width={53}
+                      maxHeight={24}
```
pepopowitz, line 57: Not blocking but pretty important visually: In the video named "Edit modal when isDefault false", the addresses jump around a bit when the pill moves from the 2nd to the 1st -- this makes me think there's a vertical margin or maxHeight or something else that's not quite right, or maybe even some missing margin on line 1 of the address text. Could we adjust this?

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -59,11 +59,28 @@ describe("AddressModal", () => {
     expect(wrapper.text()).toContain("Edit address")
     expect(wrapper.find("input").length).toBe(7)
     expect(wrapper.find("select").length).toBe(1)
+    expect(wrapper.find("Checkbox[data-test='setAsDefault']").length).toBe(1)
     expect(wrapper.find("Text[data-test='deleteButton']").length).toBe(1)
     expect(wrapper.find("Button[data-test='saveButton']").length).toBe(1)
   })
 
-  it("renders AddModal with the title, input fields and button", () => {
+  it("renders EditModal withoun checkbox when address is default", () => {
```
pepopowitz, line 67:
```suggestion
  it("renders EditModal without checkbox when address is default", () => {
```

(very minor)

src/v2/Apps/Order/Components/__tests__/AddressModal.jest.tsx
```diff
@@ -59,11 +59,28 @@ describe("AddressModal", () => {
...
+        ...savedAddress,
+        isDefault: true,
+      },
+    })
+    expect(wrapper.text()).toContain("Edit address")
+    expect(wrapper.find("input").length).toBe(7)
+    expect(wrapper.find("select").length).toBe(1)
+    expect(wrapper.find("Checkbox[data-test='setAsDefault']").length).toBe(0)
+    expect(wrapper.find("Text[data-test='deleteButton']").length).toBe(1)
+    expect(wrapper.find("Button[data-test='saveButton']").length).toBe(1)
```
pepopowitz, line 80:
```suggestion
    expect(wrapper.find("Checkbox[data-test='setAsDefault']").length).toBe(0)
```

Not blocking but fairly important: The majority of these assertions are already covered in other tests. The only one I see that is important here is to verify the checkbox isn't appearing in these conditions. 

Having more assertions here than necessary will lead to the test requiring changes for unrelated reasons.

pepopowitz (changes requested):
Looks good! I pointed out a couple small things we should change.

src/v2/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -88,3 +95,8 @@ const EditButton = styled(Text)`
     text-decoration: underline;
   }
 `
+
+const StyledPill = styled(Pill)`
+  background-color: ${color("black10")} !important;
+  pointer-events: none;
+`
```
damassi, line 102: Blocking: As this is a component in our design system the more constrained the better, meaning we should avoid styling palette components in this way. If we need to add the ability to control background colors we should update the palette `Pill` component directly. But more ideally, we push back at design on these one-off changes and instead go with current defaults.

src/v2/Apps/Order/Components/AddressModal.tsx
Serge0n: fixed

src/v2/Apps/Order/Components/SavedAddressItem.tsx
Serge0n: I took the height value from Figma. I think 21 will be correct.

src/v2/Apps/Order/Components/SavedAddressItem.tsx
Serge0n: Yes, totally agree with this.  What should I do at the moment? Leave it to work like this? As I understand I can't make corrections for the Pill component at Palette by myself.

https://user-images.githubusercontent.com/55637696/125597431-cb2abeb5-86e8-4d2d-aefd-956c5bf2583d.mov

src/v2/Apps/Order/Components/AddressModal.tsx
Serge0n: fixed

pepopowitz (approved):

src/v2/Apps/Order/Components/SavedAddressItem.tsx
damassi: I would suggest shipping this for now, and then we can iterate on adding a new property to Palette! Thoughts @pepopowitz?  

It would be worth raising in the #design-system channel in our slack so that designers (and other devs) can comment.

pepopowitz: A few thoughts: 

- +1 to shipping & iterating.
- Is there a bug here somewhere? In the video (1) the Pill starts with a medium gray background; (2) it gets a very light gray background when we hover over the second address box; (3) it stays light gray when you hover over the Pill directly; (4) it changes to a medium gray background when leaving the Pill but still hovering the address box. I'd expect 4 & 2 to match states, but they don't. If there's a bug within this PR that's causing it to turn light gray in step 2, I don't think the video would look as bad. (Is a hover state being propagated to the Pill somehow?) On the other hand, if the bug is within palette, we should track it there but I wouldn't expect to hold up this PR for it.
- It seems especially important to call out this exact demo to the design system group. The reason the default `Pill` styling doesn't look great is the combination of the Pill with an area that's turning gray due to hover. I'd expect the group to come back with either (1) we need an alternate style for the Pill because we'll be doing this more, or (2) this current UI could be re-worked to better utilize the components & design we already have, avoiding the need for an alternate `Pill` style. 
- It also would be worth raising to @tinn, to collaboratively work toward a short-term and long-term solution that utilizes palette completely instead of dropping down into styled-components to override the background. 

That's a lot, so to be clear, I wouldn't personally hold up a merge of either the styled-components solution or the solution in this video, as long as we track the work to follow up and make it better.

damassi: For reference, here's a link to [Palette's Pill storybook](https://palette-storybook.artsy.net/?path=/story/components-pill--states)

> as long as we track the work to follow up and make it better

I'd still encourage us to avoid merging the overrides with `!important` for the time being, unless there's a ticket and an immediate follow-up. The visual hover state is weird but it's noticeable, where stray overrides and `!important` could cause real confusion down the road, as they tend to do.

sweir27: I'm going to take a stab at writing up a ticket for this and pull it into the sprint as a follow up! Will ping you @pepopowitz to make sure it's accurate. 🙏

PR #7872, opened 2021-06-24
GRO-378: Copy Script changes suggested for Artist and associated pages

The-Beez-Kneez:
This covers the following copy script changes:
__________________________

Artwork page

CTAs should be title case (ie “Contact Gallery”)

Artist page

Cap “sale” in the part next to “Overview” → “Works for Sale”

Sentence case all “sub headers” — ie “Notable works,” “Career highlights,”  “Works for sale,” “Related artists” (“Shows/Articles featuring ____” is already sentence case)

“Followers” instead of "Following" for `ArtistHeader`

Hyphenate “blue-chip” before noun


damassi (approved):
Looks great 👍

PR #7876, opened 2021-06-24
fix: (GRO-374) contact for price sale message to title case

ovasdi:
https://artsyproduct.atlassian.net/browse/GRO-374

To match our new design system this updates "Contact For Price" and "Contact for price" to title case "Contact for Price" on the artwork page as well as the artwork grid/recommended rails.

| Before update | After update |
--- | --- |
|  |  |
|  |  |


The-Beez-Kneez:
> TODO

Whats is the `TODO` declaration referencing?

ovasdi:
> > TODO
> 
> Whats is the `TODO` declaration referencing?

Note to self.

ovasdi:
I've excluded the following file from making similar changes to as this seems to only affect v2 components: https://github.com/artsy/force/blob/master/src/desktop/models/artwork.coffee#L282

Please let me know if the legacy code needs these changes applied as well.

damassi:
Looks good!

PR #7880, opened 2021-06-25
feat:second button + deleted excess padding

alexj105:
This is a webview shared between web and app. We should display this second submit button only on the app, not on the web.

https://artsyproduct.atlassian.net/browse/PURCHASE-2688
ezgif com-gif-maker (2)


xtina-starr (approved):
This looks good to me! I'll go ahead and merge 🚀

PR #7881, opened 2021-06-25
fix[FX-2825]: reset size filter

tr-ann:
**Jira**: [FX-2825](https://artsyproduct.atlassian.net/browse/FX-2825)

**Description**
When user set custom size more then one time size filter doesn't reset

*Before*
reset-filter-before-2

*After*
reset-filter-after

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
@@ -146,6 +146,7 @@ export const SizeFilter: React.FC<SizeFilterProps> = ({ expanded }) => {
       sizes,
       height: "*-*",
       width: "*-*",
+      reset: false,
     }
 
     setFilters?.(newFilters, { force: false })
@@ -156,6 +157,7 @@ export const SizeFilter: React.FC<SizeFilterProps> = ({ expanded }) => {
     const newFilters = {
       ...currentlySelectedFilters?.(),
       sizes: [],
+      reset: false,
       ...mapSizeToRange(convertSizeToInches(customSize) as CustomSize),
     }
     setFilters?.(newFilters, { force: false })


MounirDhahri:
@tr-ann I am unfortunately not familiar with Fitlers on Force and can't offer good review here. check with your team who is the right person to assign here, I am sure they know better

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -146,6 +146,7 @@ export const SizeFilter: React.FC<SizeFilterProps> = ({ expanded }) => {
       sizes,
       height: "*-*",
       width: "*-*",
+      reset: false,
```
dblandin, line 149: Good find! Do you have a link to where `reset: false` is thread through and used? Wondering how this ends up fixing up the problem even though your screen recording clearly demonstrate that it does.

dblandin (approved):
This LGTM but I think it would be useful to go into more detail in the PR description on _how_ this fixes the bug.

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
tr-ann: The `reset` field used in useEffect hook used for reset custom size values in component.
When a user reset filters for the first time, `reset` field set true. Then, when the user click reset for the second time, `reset` field again set true, but in component it seems like `reset` don't updated, so effect doesn't work and custom size doesn't reset anymore.

The alternative way to fix this bug is update all filters and set `reset: false` when custom size reset. I mean 
```
useEffect(() => {
    // if filter state is being reset, then also clear local input state
    if (reset) {
      setCustomSize({ height: ["*", "*"], width: ["*", "*"] })
      setFilters?.({
          ...currentlySelectedFilters?.(),
          reset: false
      }, { force: false })
    }
  }, [reset])
```

tr-ann: But I think the current solution is cleaner 🙂

PR #7883, opened 2021-06-25
fix: Remove React Spring

icirellik:
This change replaces `react-spring` with styled component keyframe
animations and is necessary to remove a significant memory leak caused
by `react-spring` that is crashing our production pods.

Review App:
https://novo.artsy.net/

https://user-images.githubusercontent.com/403814/123460768-210a0100-d5b6-11eb-8912-5f56c771adbb.mov


damassi:
Simplified it a bit by just using `transition` and added some smoother easing, matching what we use in palette

PR #7886, opened 2021-06-28
fix: sanitizes inputs

dzucconi:
Re: https://artsy.slack.com/archives/C2TQ4PT8R/p1624812250180200

Anytime we don't sanitize inputs like this, Force will 500. But, these are at least easy to spot: anytime there's a variable input on a route query that isn't `String` — we need to prepare the input in some way. There's probably a few more of these lurking around so feel free to point any out.

### DIFF
--- a/src/v2/Apps/Artists/artistsRoutes.tsx
+++ b/src/v2/Apps/Artists/artistsRoutes.tsx
@@ -61,7 +61,7 @@ export const artistsRoutes: AppRouteConfig[] = [
           return {
             ...params,
             ...props,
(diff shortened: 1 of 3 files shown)


damassi:
I think @mzikherman is out, but this looks good to me!

dzucconi:
Aren't you supposed to be out too @damassi ❓

damassi:
I'm out, but these *#^&$ mortgage underwriters are causing me some hell right now and so i'm also at the computer 😢

PR #7891, opened 2021-06-29
fix: selecting size generate events for custom size

tr-ann:
[FX-2826](https://artsyproduct.atlassian.net/browse/FX-2826)

**Description**
When user select small/medium/large size, generates also events for custom size

**Before**
2826-before

**After**
2826-after


dzucconi:
I don't understand why we would do this? This behavior was intentional — you can't have both selected simultaneously. The ticket looks like it's just asking for analytics events?

tr-ann:
The wrong functionality. Close PR

PR #7893, opened 2021-06-29
feat: Sorting of exhibitors on the fair page

anastasiapyzhik:
[FX-2992]

[FX-2992]: https://artsyproduct.atlassian.net/browse/FX-2992

**_Desktop_**

**Before**
Screen Shot 2021-07-13 at 2 03 36 PM

**After**
after

**_Mobile_**

**Before**
Screen Shot 2021-07-13 at 2 11 29 PM

**After**
after_mob


src/v2/Apps/Fair/Utils/CheckSort.ts
```diff
@@ -0,0 +1,6 @@
+export const supportedSorts = ["FEATURED_DESC", "NAME_ASC"]
+export const defaultSort = supportedSorts[0]
+
+export const checkSort = (sort: string = "") => {
```
dimatretyak, line 4: Maybe it would be better to rename it to `checkIsAllowedSortValue`?

dimatretyak (approved):

src/v2/Apps/Fair/Utils/CheckSort.ts
anastasiapyzhik: @dimatretyak yeah, I agree it's better to give the function a clearer name but I'd rather name it `isSortValid`. What do you think?

dimatretyak: Yes, this name is even better

src/v2/Apps/Fair/Routes/__tests__/ExhibitorFilterContext.jest.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(context.filters).toEqual(initialExhibitorFilterState)
+  })
+
+  describe("behaviors", () => {
+    it("#onFilterClick", () => {
+      const spy = jest.fn()
+      getWrapper({ onFilterClick: spy })
+      context.onFilterClick!(
+        "sort",
+        "FAETURED_DESC",
```
dblandin, line 37: **typo**: Looks like this should be `FEATURED_DESC`?

src/v2/Apps/Fair/Routes/__tests__/ExhibitorFilterContext.jest.tsx
```diff
@@ -0,0 +1,55 @@
...
+        initialExhibitorFilterState
+      )
+      expect(spy).toHaveBeenCalledWith(
+        "sort",
+        "FAETURED_DESC",
```
dblandin, line 42: **typo**: Looks like this should be `FEATURED_DESC`?

src/v2/Apps/Fair/Components/FairExhibitorSortFilter.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react"
+import { SortFilter } from "v2/Components/SortFilter"
+import { useExhibitorsFilterContext } from "../Routes/ExhibitorFilterContext"
```
dblandin, line 3: **suggestion**: What do you think of moving this context component to `Fair/Components` instead of `Fair/Routes`? I think routes are usually pretty specific to the routing layer.

dblandin (commented):

src/v2/Apps/Fair/Components/FairExhibitorSortFilter.tsx
anastasiapyzhik: That's a good point, I'll move it. Thanks!

src/v2/Apps/Fair/Routes/FairExhibitors.tsx
```diff
@@ -55,61 +110,113 @@ const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair, relay }) => {
...
+}) => {
+  const {
+    match: { location },
+  } = useRouter()
+  return (
+    <ExhibitorFilterContextProvider
+      filters={location.query}
+      sortOptions={[
+        { text: "Relevance", value: "FEATURED_DESC" },
+        { text: "Alphabetical (A-Z)", value: "NAME_ASC" },
```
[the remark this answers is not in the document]
anastasiapyzhik: [this line](https://github.com/artsy/gravity/blob/f3748343a32d85ec2afc9817e287b3964fa7667d/app/api/v1/fairs_shows_endpoint.rb#L27) adds `"sortable"` to the string start in case the string contains `"name"`. So if we pass `SORTABLE_NAME_ASC` an error occurs (screenshot below). But if we pass `NAME_ASC` gravity turns it into `"sortable_name"`

Screenshot 2021-07-19 at 10 27 55

src/v2/Apps/Fair/Routes/__tests__/ExhibitorFilterContext.jest.tsx
anastasiapyzhik: Not sure how I missed that twice 😅  Thanks!

src/v2/Apps/Fair/Routes/FairExhibitors.tsx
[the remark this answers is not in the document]
dblandin: Ahh, strange! Thanks for the context.

dblandin (approved):
Looks good to me! Thanks for making those changes 👍🏼

PR #7895, opened 2021-06-29
chore: Remove unused code based on #7797 

pepopowitz:
In #7797 [I mentioned that I thought we could clean up a few more stateful properties](https://github.com/artsy/force/pull/7797#discussion_r660128541) in the `Shipping` route component, because they're being set but never read. This PR does that!

@Serge0n you're closer to this work, let me know if I'm missing something and these things can't be removed. 

I created the PR based on the branch for #7797, so this _could_ be merged into that branch before it's merged, if we want. Or we could merge this into master after #7797 is merged.

### DIFF
--- a/src/v2/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/v2/Apps/Order/Routes/Shipping/index.tsx
@@ -88,9 +88,7 @@ export interface ShippingState {
   addressErrors: AddressErrors
   addressTouched: AddressTouched
   selectedSavedAddress: string
-  editAddressIndex: number
   saveAddress: boolean
-  showModal: boolean
 }
 
 const logger = createLogger("Order/Routes/Shipping/index.tsx")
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -233,7 +229,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
             }
           }, // onError
           this.props.me, // me
-          () => this.setState({ showModal: false }) // closeModal
```
pepopowitz, line 236: I investigated if this argument could be completely removed from the function being called, but it is still needed for other code paths.

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -110,9 +108,7 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
     phoneNumberError: "",
     phoneNumberTouched: false,
     selectedSavedAddress: defaultShippingAddressIndex(this.props.me),
-    editAddressIndex: -1,
     saveAddress: true,
-    showModal: false,
```
pepopowitz, line 115: These are two properties that I don't think are needed anymore. Both are set, but neither is ever read as far as I can tell.

src/v2/Apps/Order/Routes/Shipping/index.tsx
Serge0n: Yes, you are right. Now modals state based in SavedAddresses.tsx

Serge0n (approved):

PR #7897, opened 2021-06-30
fix: size events fire only by changes explicitly

tr-ann:
Jira: [FX-2826](https://artsyproduct.atlassian.net/browse/FX-2826), [FX-2827](https://artsyproduct.atlassian.net/browse/FX-2827)

Solution for issues:
- user select/unselect small/medium/large size and filter fires events for custom width/height -- if custom width/height have null value (it may be undefined or *-*) and new value is also null, they don't update, in other way width/height should reset
- user select a custom width and filter fire event for custom height -- if custom height have null value and new value is also null, it don't update (the same for width)

**Before**
Selecting size fires additional analytics events
Ex. selecting medium size fire events for custom width/height
image
image

**After**
Size events fire only by changes explicitly
image

**Additional**
Fix the bug with reset flag (when user clear all flags the _Apply_ button can not be clicked, so the filters don't apply)

### DIFF
--- a/src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
+++ b/src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
@@ -60,7 +60,7 @@ describe("ArtworkFilterMobileActionSheet", () => {
 
     expect(context.stagedFilters).toEqual({
       ...initialArtworkFilterState,
-      reset: true,
+      reset: false,
     })
   })
 
--- a/src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
+++ b/src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
@@ -60,7 +60,7 @@ describe("ArtworkFilterMobileActionSheet", () => {
 
     expect(context.stagedFilters).toEqual({
       ...initialArtworkFilterState,
-      reset: false,
+      reset: true,
     })
   })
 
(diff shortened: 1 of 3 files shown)


dzucconi:
One thing to double check here, because I'm not sure if it's expressed (or easily expressible) in the specs: be sure the backspace works correctly in the custom inputs. I recall that being a bit tricky to get the interaction correct.

tr-ann:
> One thing to double check here, because I'm not sure if it's expressed (or easily expressible) in the specs: be sure the backspace works correctly in the custom inputs. I recall that being a bit tricky to get the interaction correct.

I checked all cases editing size. I think it works good

src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
```diff
@@ -60,7 +60,7 @@ describe("ArtworkFilterMobileActionSheet", () => {
 
     expect(context.stagedFilters).toEqual({
       ...initialArtworkFilterState,
-      reset: true,
+      reset: false,
```
dzucconi, line 63: This is my only ❓ — this spec seems like it _should_ be resetting this flag. It's been a while since I've looked at this code, so I don't even recall the "why" of this... but if the "Clear all" button doesn't flip this flag to `true`, then what does?

dimatretyak (approved):

src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
tr-ann: You are right, the "Clear button" set `reset: true`. But immediately effect reacted for `reset` changes and updates filters (set `false` value to `reset` flag).

Unfortunately I found the bug with this solution. When user clear all filters "apply" button are disable and it is impossible to apply reset filters. So I try to fix it

tr-ann: Fix the bug, so `reset` flat works correctly now

dzucconi (approved):

dblandin (approved):

PR #7899, opened 2021-06-30
fix: discernible name for artwork grid item links

tr-ann:
Jira: [FX-2383](https://artsyproduct.atlassian.net/browse/FX-2383)

Added aria-label attribute for artwork image links
Updated GraphQL schemas
Updated mock data for tests (artworks, fair, AKGroute)

**Result**
image


mdole (approved):
Nice! Seems like there might be a couple more mocks we need to update to get tests to pass, but this looks great otherwise. Assigning back to you @tr-ann so you can merge once the PR is green.

dimatretyak (commented):
LGTM 👍 I agree with Matt

PR #7903, opened 2021-06-30
feat(home): updates above the fold design/layout

dzucconi:
This is primarily a set of visual changes to bring the above the fold area on the new homepage [in line with Figma](https://www.figma.com/file/3zvqokEqQ3pvwt6iul0a9D/Homepage).

https://user-images.githubusercontent.com/112297/124023496-ff48b980-d9bb-11eb-853f-f609e5c05d35.mov

https://user-images.githubusercontent.com/112297/124023442-f0fa9d80-d9bb-11eb-9167-ea064912969f.mov


src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnitLayout.tsx
```diff
@@ -0,0 +1,113 @@
+import React from "react"
```
dzucconi, line 1: Woops this file is no longer needed.

PR #7904, opened 2021-06-30
fix: Remove Underscore from Asset Bundle

icirellik:
This change removes underscore from the asset bundle decreasing it's
size by 30KB gzip.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "test:ci:mocha": "NODE_ENV=test yarn mocha src/**/*.test.js src/**/*.test.ts",
     "type-check": "tsc",
     "unlink-all": "yalc remove --all && yarn --check-files",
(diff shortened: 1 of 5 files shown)


jonallured (approved):
🥇 🏆 😍

dzucconi:
omg

joeyAghion (approved):
Nice! I wonder what wonky bugs we'll encounter. 😬

dzucconi:
@joeyAghion Shouldn't be any bugs since this is just two minor instances where a dev mistakenly imported underscore instead of lodash in new code. (Evergreen reminder to callout usage of underscore if you encounter it in new code)

icirellik:
@joeyAghion Lodash and Underscore have very similar APIs, I did verify that the functionality and signature for all these methods were identical.

PR #7912, opened 2021-07-01
feature: Pod Stabilization (Staging)

icirellik:
This change alters the force pod autoscaler to wait 30 minutes before
removing nodes after a scaleUp event. Looking at the graphs in Datadog
there is a significant amount of pod trashing and this will remediate
that.

This change does not alter upscale timing which will stay at
"immediate".

### DIFF
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -83,7 +83,7 @@ spec:
                       - foreground
 
 ---
-apiVersion: autoscaling/v1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: force-web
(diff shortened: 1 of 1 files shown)


hokusai/staging.yml
```diff
@@ -83,19 +83,25 @@ spec:
...
   name: force-web
   namespace: default
 spec:
-  scaleTargetRef:
-    apiVersion: apps/v1
-    kind: Deployment
-    name: force-web
   minReplicas: 2
   maxReplicas: 3
   targetCPUUtilizationPercentage: 70
```
ovasdi, line 94: Wonder if we need this now that you added the `metrics` annotation.

hokusai/staging.yml
```diff
@@ -83,19 +83,25 @@ spec:
...
 
 ---
-apiVersion: autoscaling/v1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: force-web
   namespace: default
 spec:
-  scaleTargetRef:
```
ovasdi, line 92: This seems to be required according to the [spec](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v2beta2/#HorizontalPodAutoscalerSpec)
> scaleTargetRef (CrossVersionObjectReference), required

ovasdi (commented):
Left a couple comments.

artsyjian (approved):
down for trying it!

mc-jones (approved):

PR #7913, opened 2021-07-01
feat(GRO-403): fleshes out meta tags

dzucconi:
This implements a generic meta tag component and then uses it on the new homepage. I realize that these are kind of fiddly tedious to set up. Easy to forget things, image specifications, etc... so this takes the guess work out of it.

I've included usage in a couple other places. I'll have to go around and replace *all* meta tag components, which I'll do in a separate PR.

### DIFF
--- a/src/desktop/components/fair_week_marketing/__tests__/BannerPopUp.jest.tsx
+++ b/src/desktop/components/fair_week_marketing/__tests__/BannerPopUp.jest.tsx
@@ -9,7 +9,7 @@ jest.mock("sharify", () => ({
       {
         slug: "ca3",
         copy: "Discover and Buy Works from Art Fairs",
-        image: "http://files.artsy.net/images/art-fair.jpg",
+        image: "https://files.artsy.net/images/art-fair.jpg",
       },
     ],
   },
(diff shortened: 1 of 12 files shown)


src/desktop/components/fair_week_marketing/__tests__/BannerPopUp.jest.tsx
```diff
@@ -9,7 +9,7 @@ jest.mock("sharify", () => ({
       {
         slug: "ca3",
         copy: "Discover and Buy Works from Art Fairs",
-        image: "http://files.artsy.net/images/art-fair.jpg",
+        image: "https://files.artsy.net/images/art-fair.jpg",
```
dzucconi, line 12: Random thing I noticed when doing this; the copy-to-clipboard button in Vanity copies an http URL instead of https.

src/v2/Components/MetaTags.tsx
```diff
@@ -0,0 +1,76 @@
...
+    getENV("APP_URL"),
+    pathname.startsWith("/") ? pathname : `/${pathname}`,
+  ].join("")
+
+  const src = imageURL
+    ? cropped(imageURL, { width: 1200, height: 630 }).src
+    : null
+
+  const card =
+    !src || src === "DEFAULT_IMAGE_URL" ? "summary" : "summary_large_image"
```
dzucconi, line 46: Woops, this comparison is wrong. Will update...

PR #7916, opened 2021-07-01
feat(PX-4293): rename Shows tab to Events

kiryl-zubarau:
This PR renames **Shows** tab to **Events**

JIRA - https://artsyproduct.atlassian.net/browse/PX-4293

image

### DIFF
--- a/src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
+++ b/src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
@@ -39,7 +39,7 @@ export const NavigationTabs: React.FC<NavigationTabsProps> = ({ partner }) => {
         exact: true,
       },
       {
-        name: "Shows",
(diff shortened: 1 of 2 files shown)


sweir27:
#squashongreen

kiryl-zubarau:
I updated only the tab title. Let me know if you want to update routing too.

sweir27:
Oh good question @kiryl-zubarau -- would be worth asking Joh that. (And if we do update the route, do we need to change our redirect logic at all?)

kiryl-zubarau:
@sweir27 it will require to add section to the redirection logic like this one, but for the `/shows` route https://github.com/artsy/force/blob/dd646110bfd5098b297e46fa2270398be0d92681/src/desktop/apps/partner_redirect/partnerRedirectionMiddleware.ts#L23-L27

PR #7917, opened 2021-07-02
feature: Pod Scaling Stabilization (Production)

icirellik:
This change alters the force pod autoscaler to wait 30 minutes before
removing nodes after a scaleUp event. Looking at the graphs in Datadog
there is a significant amount of pod trashing and this will remediate
that.

This change does not alter upscale timing which will stay at
"immediate".

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -79,7 +79,7 @@ spec:
                     values:
                       - foreground
 ---
-apiVersion: autoscaling/v1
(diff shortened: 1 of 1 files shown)


artsyjian (approved):
lgtm

mc-jones (approved):

joeyAghion (approved):
I presume the behavior in staging was as expected?

icirellik:
Yes it looks good in staging.

PR #7918, opened 2021-07-02
chore: Remove @artsy/reaction from v2 Bundle

icirellik:
This change uses an existing local file over the same file in
`@artsy/reaction` thus eliminating it from the v2 bundle.


mc-jones (approved):

mc-jones:
LGTM

PR #7919, opened 2021-07-02
[PX-2975] Updates galleries_institutions to use metaphysics v2

sweir27:
Blocked by: https://github.com/artsy/metaphysics/pull/3259

Jira ticket: https://artsyproduct.atlassian.net/browse/PX-2975

This PR updates the galleries/institutions app to use metaphysics v2 instead of v1. It relies on the schema updates in the ☝️ metaphysics PR, hence the `[Don't Merge]` 😄.

The updates are because the schemas are slightly different, but I clicked around a bit and I think we're all good! 🤞

### DIFF
--- a/src/desktop/apps/galleries_institutions/queries/partner_categories_query.coffee
+++ b/src/desktop/apps/galleries_institutions/queries/partner_categories_query.coffee
@@ -1,14 +1,14 @@
 module.exports =
 
   """
-  query GalleriesInstitutionsPartnerCategoriesQuery($category_type: CategoryType, $type: [PartnerClassification]){
-    partner_categories(category_type: $category_type, size: 50, internal: false){
+  query GalleriesInstitutionsPartnerCategoriesQuery($categoryType: PartnerCategoryType, $type: [PartnerClassification]){
+    partner_categories: partnerCategories(categoryType: $categoryType, size: 50, internal: false){
       name
-      id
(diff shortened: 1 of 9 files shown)


src/desktop/apps/galleries_institutions/queries/partner_categories_query.coffee
```diff
@@ -1,14 +1,14 @@
...
 
   """
-  query GalleriesInstitutionsPartnerCategoriesQuery($category_type: CategoryType, $type: [PartnerClassification]){
-    partner_categories(category_type: $category_type, size: 50, internal: false){
+  query GalleriesInstitutionsPartnerCategoriesQuery($categoryType: PartnerCategoryType, $type: [PartnerClassification]){
+    partner_categories: partnerCategories(categoryType: $categoryType, size: 50, internal: false){
       name
-      id
-      primary: partners(eligible_for_listing: true, eligible_for_primary_bucket: true, type: $type, sort: RANDOM_SCORE_DESC, default_profile_public: true) {
+      id: slug
```
sweir27, line 7: This change is a little weird-- we use this to populate our list of partners/categories on the server, then the `partner_filter_query` below to populate on the client, and we need a way to connect the two.

With v2, we no longer allow you to query for an ID as part of aggregations (you just have `value`, which matches the `slug` here).

anandaroop (approved):
Nice to see more v1 ✂️ !

sweir27:
Alright! Metaphysics has been deployed and hopefully the tests pass now 🤞 . Will keep an eye on it!

PR #7920, opened 2021-07-03
feat: initial real time conversations implementation

jo-rs:
Resolves [PURCHASE-2716], [PURCHASE-2743], [PURCHASE-2811]

Mainly includes a naive implementation of "real-time" conversations. Currently polling every 10 seconds, and only while the tab is active so as to not spam the server. Eventually we'll move into better patterns.

Also includes a handful of fixes for scrolling bugs on conversations, and some refactoring.

ezgif com-gif-maker

[PURCHASE-2716]: https://artsyproduct.atlassian.net/browse/PURCHASE-2716
[PURCHASE-2743]: https://artsyproduct.atlassian.net/browse/PURCHASE-2743
[PURCHASE-2811]: https://artsyproduct.atlassian.net/browse/PURCHASE-2811


src/v2/Apps/Conversation/Components/ConfirmArtworkModal.tsx
```diff
@@ -33,7 +33,7 @@ export const ConfirmArtworkModal: React.FC<ConfirmArtworkModalProps> = ({
   show,
   closeModal,
 }) => {
-  const { editionSets, isEdition } = artwork
+  const { editionSets, isEdition } = { ...artwork }
```
jo-rs, line 36: This is just a workaround for an issue that kept popping up. Unrelated.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -49,27 +58,39 @@ const Conversation: React.FC<ConversationProps> = props => {
...
+    ) {
+      scrollToBottom()
+    }
+  }, [conversation?.lastMessageID])
 
   useEffect(() => {
-    UpdateConversation(relay.environment, conversation)
-  }, [conversation, relay.environment, conversation.lastMessageID])
+    markRead(!conversation.isLastMessageToUser)
+  }, [lastMessageID])
```
jo-rs, line 93: This can be probably condensed

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -134,6 +161,33 @@ const Conversation: React.FC<ConversationProps> = props => {
     orderID: orderID,
   })
 
+  // New Messages
+  const isBottomVisible = useOnScreen(bottomOfPage)
```
jo-rs, line 165: Checking if the bottom is visible - or the user has scrolled to know when to display the cta.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -134,6 +161,33 @@ const Conversation: React.FC<ConversationProps> = props => {
...
+  useEffect(() => {
+    if (isBottomVisible) refreshData()
+  }, [isBottomVisible])
```
jo-rs, line 168: For when the cta appears, but instead of clicking it, the user scrolls down.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -134,6 +161,33 @@ const Conversation: React.FC<ConversationProps> = props => {
...
+      () => {
+        setLastMessageID(conversation?.lastMessageID)
+        UpdateConversation(relay.environment, conversation)
+      },
+      delayed ? 3000 : 0
+    )
+  }
+  const viewUnreadMessages = () => {
+    refreshData()
+  }
```
jo-rs, line 189: This can defo be condensed

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -149,18 +203,21 @@ const Conversation: React.FC<ConversationProps> = props => {
               {isOfferable && <BuyerGuaranteeMessage />}
               {inquiryItemBox}
               <Waypoint onEnter={loadMore} />
-              {fetchingMore ? (
-                <SpinnerContainer>
-                  <Spinner />
-                </SpinnerContainer>
-              ) : null}
+              {fetchingMore ? <Loading /> : null}
```
jo-rs, line 206: I refactored this, initially, to use at the bottom as well but opted against it. Will probs revert this.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -149,18 +203,21 @@ const Conversation: React.FC<ConversationProps> = props => {
...
               <ConversationMessages
                 // @ts-expect-error STRICT_NULL_CHECK
                 messages={conversation.messagesConnection}
+                lastViewedMessageID={+(lastMessageID || -1)}
```
jo-rs, line 210: This ended up being weird. I need to clean that up.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -149,18 +203,21 @@ const Conversation: React.FC<ConversationProps> = props => {
...
               />
               <Box ref={bottomOfPage}></Box>
             </Flex>
           </Box>
+          <UnreadMessagesToastQueryRenderer
+            conversationID={conversation?.internalID!}
+            onOfferable={!!(artwork && isOfferable)}
+            hasScrolled={!isBottomVisible}
+            onClick={viewUnreadMessages}
+          />
```
jo-rs, line 220: This is the actual cta. `onOfferable` is used for positioning, but I'm sure there's a million smarter and less verbose ways of accomplishing the same.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -256,6 +315,7 @@ export const ConversationPaginationContainer = createPaginationContainer(
               id
             }
           }
+          totalCount
```
jo-rs, line 318: Ended up not using this. Delete.

src/v2/Apps/Conversation/Components/ConversationMessages.tsx
```diff
@@ -42,8 +47,22 @@ export const ConversationMessages = ({
...
+              if (
+                !newFlagShown &&
+                !!lastViewedMessageID &&
+                message.internalID > lastViewedMessageID &&
+                !message.isFromUser
+              ) {
+                // This marks visibility
+                showNewFlag = true
+                // This makes sure that it appears only on the first new message
+                newFlagShown = true
```
jo-rs, line 60: Self explanatory comment

src/v2/Apps/Conversation/Components/ConversationMessages.tsx
```diff
@@ -42,8 +47,22 @@ export const ConversationMessages = ({
...
+              }
+
               return (
                 <React.Fragment key={`message-${message.internalID}`}>
+                  {showNewFlag && <NewMessageMarker />}
```
jo-rs, line 65: This is the "new" marker that shows on top of new messages.

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
+import {
+  UnreadMessagesToastQuery,
+  UnreadMessagesToastQueryResponse,
+} from "v2/__generated__/UnreadMessagesToastQuery.graphql"
+import { UnreadMessagesToast_conversation } from "v2/__generated__/UnreadMessagesToast_conversation.graphql"
+
+// I assume this will have to be refactored into one of the newer components. Potentially one of the new buttons.
+const Container = styled(Flex)<{ bottomMargin?: boolean }>`
+  background-color: #1023D7; //${color("blue100")};
```
jo-rs, line 19: Had some issues getting the right blue/version-of-palette. Can defo refactored into one of the new blue buttons...I think

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+  border: none;
+  border-radius: 30px;
+  height: 40px;
+  width: 180px;
+  cursor: pointer;
+  position: fixed;
+  bottom: -40px;
+  transition: all 300ms;
```
jo-rs, line 27: Random transition speed choice. I'm sure this is standardised.

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+  &.visible {
+    bottom: ${props => (props.bottomMargin ? 180 : 80)}px;
```
jo-rs, line 29: Also, this is very CSS-y. Probably more React-y is preferable.

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+      relay.refetch(
+        {
+          conversationID: conversation?.internalID,
+        },
+        null
+      )
+    },
+    10,
+    match.params?.conversationID
+  )
```
jo-rs, line 73: This is where the checking every 10 seconds happens.

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  useEffect(() => {
+    setVisible(
+      hasScrolled &&
+        conversation?.lastMessageID !== conversation?.fromLastViewedMessageID
+    )
+  }, [
+    hasScrolled,
+    conversation?.lastMessageID,
+    conversation?.fromLastViewedMessageID,
```
jo-rs, line 59: Can probably be condensed, and made to avoid too many triggers.

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
jo-rs: This should probably have an idle time to not spam mp when the user leaves the tab open.

jo-rs: Probably on active tab

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -0,0 +1,142 @@
...
+        ({ me }) => (
+          <UnreadMessagesToastRefetchContainer
+            conversation={me?.conversation}
+            {...rest}
+          />
+        )
+      )}
+    />
+  )
+}
```
jo-rs, line 142: This is just a bit of a mess atm

src/v2/Apps/Conversation/Utils/useOnScreen.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, MutableRefObject, useState } from "react"
+
+export default function useOnScreen<T extends HTMLElement | null>(
```
jo-rs, line 3: Simple intersection observer hook, for visibility.

src/v2/Apps/Conversation/Utils/usePoll.ts
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from "react"
+
+type IntervalFunction = () => unknown | void
+
+export default function usePoll(
```
jo-rs, line 5: Less horrible `setInterval` hook. Ideally we can replace this with subscriptions in the future, or switch to using relay hooks as an intermediary step.

jo-rs:
And, of course tests are failing, because they're currently missing 🙃

src/v2/Apps/Conversation/Components/UnreadMessagesToast.tsx
damassi: Can we maybe add an `ease-out` ease so it can smooth in?

src/v2/Apps/Conversation/Utils/usePoll.ts
damassi: We should [avoid default exports](https://github.com/artsy/force/blob/master/docs/best_practices.md#naming-imports-and-exports) as they're not statically analyzable like named exports are (a named export can be one thing and one thing only when imported, unless deliberately aliased which takes extra effort). 

The reason is default exports often lead to drift because devs will import them and assign a variety of names to the variable.

src/v2/Apps/Conversation/Utils/usePoll.ts
```diff
@@ -0,0 +1,28 @@
...
+  callback: IntervalFunction,
+  seconds: number,
+  key: string | null,
+  stop?: boolean
+) {
```
damassi, line 10: Totally non-blocking form suggestion: We should use a single object as the argument (like how `props` works in a react component), versus a series of unnamed "bare" args. 

For example, a future dev at a glance will have a hard time understanding what each slot does ("what does '10' mean 🤔 ):  

```tsx
usePoll(
    () => {
      relay.refetch({ conversationID: conversation?.internalID }, null)
    },
    10,
    match.params?.conversationID,
    !tabActive
  )
```

Where something like the below, while more verbose, leaves no room for ambiguity:

```tsx
usePoll({
    callback: () => {
      relay.refetch({ conversationID: conversation?.internalID }, null)
    },
    intervalTime: 10,
    clearWhen: !tabActive 
    key: match.params?.conversationID,
  })
```

By using objects as args the code becomes self-documenting (since an object's keys describe its function), closer to React's idioms (which we spend the most time in) as well as less brittle (since you can move object keys all around, where naked args have fixed positions).

src/v2/Apps/Conversation/Utils/usePoll.ts
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useRef } from "react"
+
+type IntervalFunction = () => unknown | void
+
+export default function usePoll(
+  callback: IntervalFunction,
+  seconds: number,
```
damassi, line 7: this should be in milliseconds to match native `useInterval`

src/v2/Apps/Conversation/Utils/usePoll.ts
jo-rs: The avoiding default exports RFC was mine, so I should've known better 🙈

src/v2/Apps/Conversation/Utils/usePoll.ts
jo-rs: Yeah, makes sense 👍

damassi (approved):

damassi (approved):
Also, thanks for all of the inline self-comments describing implementation 💯  that's a really helpful practice for reviewers.

PR #7923, opened 2021-07-05
Fix:target _blank for The Artsy guarantee link

alexj105:
https://artsyproduct.atlassian.net/browse/PURCHASE-2791
The Artsy guarantee link opens in new window


xtina-starr (approved):
👍

PR #7926, opened 2021-07-06
feat(PX-4226): list of ARTA-generated shipping quotes

kiryl-zubarau:
This PR adds the list of ARTA-generated shipping quotes. 

JIRA - https://artsyproduct.atlassian.net/browse/PX-4226

<details>
  <summary>Default address in USA</summary>

  https://user-images.githubusercontent.com/79979820/124576790-51bf2580-de55-11eb-8702-9fec007e7302.mov
</details>

<details>
  <summary>Change address to the address in USA</summary>

  https://user-images.githubusercontent.com/79979820/124578361-d199bf80-de56-11eb-8384-4a100ad751da.mov
</details>

<details>
  <summary>Change address to the address not in USA</summary>

  https://user-images.githubusercontent.com/79979820/124578711-20475980-de57-11eb-8cbb-20ad812e6000.mov
</details>

<details>
  <summary>Default address not in USA</summary>

  https://user-images.githubusercontent.com/79979820/124579269-a06dbf00-de57-11eb-8f9c-4a1bbfff68a9.mov
</details>

<details>
  <summary>Change shipping option to Pickup and back</summary>

  https://user-images.githubusercontent.com/79979820/124580016-5cc78500-de58-11eb-864d-555f38e17968.mov
</details>

<details>
  <summary>Change shipping address after back from Payment step</summary>

  https://user-images.githubusercontent.com/79979820/124580700-11fa3d00-de59-11eb-846a-b7249a2c62db.mov
</details>

<details>
  <summary>Select shipping quote if collector has no saved addresses</summary>

  https://user-images.githubusercontent.com/79979820/124581936-3b679880-de5a-11eb-8081-bc524502434c.mov
</details>


src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -130,18 +142,53 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+  }
+
   // @ts-expect-error STRICT_NULL_CHECK
   getAddressList = () => this.props.me.addressConnection.edges
 
+  getOrderArtwork = () => this.props.order.lineItems?.edges?.[0]?.node?.artwork
+
   isCreateNewAddress = () => this.state.selectedSavedAddress === NEW_ADDRESS
 
+  isArtaShipping = () => {
```
jpotts244, line 162: There might be an opportunity to add a lab feature check here!


The lab feature will allow us to gate this logic from being shown to users and only show it to admins/individual internal users before we launch. (and allow us to set `artaShippingEnabled=true` on some artworks for testing without exposing to users) Some additional documentation [here](https://github.com/artsy/gravity/blob/50aeb677c8cddc401161b65bffa09dba57167f82/doc/LabFeatures.md)

In Force, the check for a user's eligibility could look similar to [this check](https://github.com/artsy/force/pull/5579/files).

Curious to hear your thoughts, since we are working async, I have generated a lab feature in gravity staging called `"Collector Arta Shipping"` for admin users and added it to your artsymail user account in case you would like to use it.

src/v2/Apps/Order/Components/ShippingQuotes.tsx
```diff
@@ -0,0 +1,83 @@
...
+  {
+    shippingQuotes: graphql`
+      fragment ShippingQuotes_shippingQuotes on CommerceShippingQuoteEdge
+        @relay(plural: true) {
+        node {
+          id
+          tier
+          name
+          isSelected
+          price(precision: 2)
```
jpotts244, line 78: Thank you for the price formatting updates!

src/v2/Apps/Order/Routes/Shipping/index.tsx
kiryl-zubarau: Yea, we discussed it with @ansor4, and if I remember correctly we decided to use **artaShippingEnabled** that comes from CMS as a lab feature key. I think if the ability to set **arta_enabled**(in CMS, `false` by default) will be hidden by the lab feature and collector won't have a chance to see the ARTA shipping quotes in Force. 

Also, I think it might break the page: for example, if we allow working with ARTA only for people with enabled lab feature and admin sets **arta_enabled** to **true** for an artwork what price should we show for the collectors without enabled lab feature? (according to the design there is no way to set shipping price if ARTA shipping is selected)

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -145,7 +145,9 @@ export const AddressModal: React.FC<Props> = ({
                 onChange={formik.handleChange}
                 onBlur={formik.handleBlur}
                 error={formik.touched.phoneNumber && formik.errors.phoneNumber}
-                value={formik.values?.phoneNumber}
+                value={
+                  formik.values?.phoneNumber ? formik.values?.phoneNumber : ""
```
ansor4, line 149: non-blocking, minor: Is the purpose of this to return an empty string instead of undefined? Maybe this could be easier to read?:
```
value={formik.values?.phoneNumber || ""}
```

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -183,7 +185,7 @@ export const AddressModal: React.FC<Props> = ({
           action: () => {
             setShowDialog(false)
             closeModal()
-            onDeleteAddress(address.internalID)
+            onDeleteAddress(address?.internalID)
```
ansor4, line 188: non-blocking, info: any reason to use safe accessor here vs failing loudly?

src/v2/Apps/Order/Components/ShippingQuotes.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <RadioGroup
+      {...rest}
+      onSelect={onSelect}
+      defaultValue={selectedShippingQuoteId}
+    >
+      {quotes.map(shippingQuote => {
+        const { id, name, price, tier } = shippingQuote
+        const shippingQuoteName = name || tier
```
ansor4, line 43: non-blocking, info: Hmm, this feels like it might cause mishaps in the future but I'm not close to the work. Is this always going to be the case where there's no name for `select` and `premium` tiers? What if Arta suddenly starts returning a name for quotes in those tiers?

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -34,7 +34,7 @@ export const AddressModalFields: React.FC = props => {
           onChange={handleChange}
           onBlur={handleBlur}
           error={touched.name && errors.name}
-          value={values?.name}
+          value={values?.name ? values?.name : undefined}
```
ansor4, line 37: Is this supposed to return `undefined` if `values?.name` is `null`?

I think `values?.name || undefined` should work here

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -82,7 +84,7 @@ export const AddressModalFields: React.FC = props => {
                 onChange={handleChange}
                 onBlur={handleBlur}
                 error={touched.postalCode && errors.postalCode}
-                value={values?.postalCode}
+                value={values?.postalCode ? values?.postalCode : ""}
```
ansor4, line 87: Same here: `values?.postalCode || ""`

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -91,7 +93,7 @@ export const AddressModalFields: React.FC = props => {
                 onChange={handleChange}
                 onBlur={handleBlur}
                 error={touched.addressLine2 && errors.addressLine2}
-                value={values?.addressLine2}
+                value={values?.addressLine2 ? values?.addressLine2 : ""}
```
ansor4, line 96: And here: `values?.addressLine2 || ""`

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -100,7 +102,7 @@ export const AddressModalFields: React.FC = props => {
                 onChange={handleChange}
                 onBlur={handleBlur}
                 error={touched.region && errors.region}
-                value={values?.region}
+                value={values?.region ? values?.region : ""}
```
ansor4, line 105: And last but not least: `values?.region || ""`

src/v2/Components/Payment/PaymentModal.tsx
```diff
@@ -100,12 +100,12 @@ export const PaymentModal: React.FC<PaymentModalProps> = props => {
...
-      name: values.name,
+      name: values.name ? values.name : undefined,
       address_line1: values.addressLine1,
-      address_line2: values.addressLine2,
+      address_line2: values.addressLine2 ? values.addressLine2 : undefined,
       address_city: values.city,
-      address_state: values.region,
-      address_zip: values.postalCode,
+      address_state: values.region ? values.region : undefined,
+      address_zip: values.postalCode ? values.postalCode : undefined,
```
ansor4, line 108: These too can be converted into `values.__field__ || undefined`

ansor4 (approved):
Some minor comments, mostly stylistic. One concern on displaying quotes via `name || tier `, any way we can future-proof that? I'm ok with a follow up PR for that.

src/v2/Apps/Order/Components/AddressModal.tsx
kiryl-zubarau: Since the delete button visible only for the address update, `address` and `internalID` can't be falsy. Updated code to verify `internalID` value before call `onDeleteAddress`

src/v2/Apps/Order/Components/ShippingQuotes.tsx
kiryl-zubarau: In a case when Arta starts returning a name for quotes according to current logic we will show only the name and price(without description). 

I think it's okay to use the name because if for example, we receive 2 quotes with a `premium` tier (quotes with `Premium 1` and `Premium 2` names) that will be useful to identify each quote(by name in this case). Also since we use our own quote descriptions it's not a good idea to my mind to use the current `premium` description for `Premium 1` and `Premium 2` quotes because they may differ a lot.

PR #7927, opened 2021-07-06
fix:right width for protectedBy block

alexj105:
This box is placed on full conversation width. Set maxWidth as Figma required 
https://artsyproduct.atlassian.net/browse/PURCHASE-2792
Screen Shot 2021-07-06 at 1 36 29 PM
Screen Shot 2021-06-26 at 00 47 35


xtina-starr (approved):
Looks good 👍

PR #7929, opened 2021-07-06
feature: App Bundling

icirellik:
Reduces the size of the shared `artsy-entry` bundle by encapsulating
more app logic into specific application bundles.

The entry bundle is reduced by 961 KB gzip to 794 KB gzip.

Bundles Before:
bundle-pre

Bundles After:
bundle-post


damassi (approved):

jonallured (approved):
I'm not sure how to give meaningful feedback here - seems legit I guess!! 😂

mc-jones (approved):

icirellik:
@jonallured  I will post screenshots of the bundle output...not the easiest to compare however it does show that the entry bundle, which is shared among all pages, is smaller. Which should lead to faster page loads overall.

PR #7932, opened 2021-07-06
fix: anchor to the last message in conversation

ivan-badyulya:
**JIRA** -> [Web - Anchoring to the latest message in a conversation](https://artsyproduct.atlassian.net/browse/PURCHASE-2805)

**Expected Outcome**
When a user opens a chat without new messages, the conversation should be scrolled down to the last seen message.

**Approach notes**
The problem of incorrect scrolling down is in children. The children of the Conversation components are rendered after the scrollTo method is invoked. The approach was based on creating a flags which would indicate that children are rendered and invoke scrolling after. Sometimes the scrolling isn't scrolling all the way down, thus looking forward of the best solution.

**GIF**:
ezgif com-gif-maker (4)

### DIFF
--- a/src/v2/Apps/Conversation/Components/Reply.tsx
+++ b/src/v2/Apps/Conversation/Components/Reply.tsx
@@ -61,7 +61,7 @@ const StyledTextArea = styled.textarea<{ height?: string }>`
 interface ReplyProps {
   conversation: Conversation_conversation
   environment: Environment
-  onScroll?: () => void
+  onScroll: () => void
   refetch: RelayRefetchProp["refetch"]
   openInquiryModal: () => void
   openOrderModal: () => void
@@ -154,6 +154,7 @@ export const Reply: React.FC<ReplyProps> = props => {
           conversation={conversation}
           openInquiryModal={() => openInquiryModal()}
           openOrderModal={() => openOrderModal()}
+          scrollMessageContainer={onScroll}
         />
         <StyledFlex p={1}>
           <FullWidthFlex width="100%">
--- a/src/v2/Apps/Conversation/Components/Reply.tsx
+++ b/src/v2/Apps/Conversation/Components/Reply.tsx
@@ -142,7 +142,7 @@ export const Reply: React.FC<ReplyProps> = props => {
       <Flex
         right={[0, null]}
         zIndex={[null, 2]}
-        position={["fixed", "fixed", "fixed", "static"]}
+        position={["fixed", "fixed", "static", "static"]}
         bottom={0}
         left={0}
         flexShrink={0}
--- a/src/v2/Apps/Conversation/Components/Reply.tsx
+++ b/src/v2/Apps/Conversation/Components/Reply.tsx
@@ -62,6 +62,7 @@ interface ReplyProps {
   conversation: Conversation_conversation
   environment: Environment
   onScroll: () => void
+  onMount: () => void
   refetch: RelayRefetchProp["refetch"]
   openInquiryModal: () => void
   openOrderModal: () => void
@@ -72,6 +73,7 @@ export const Reply: React.FC<ReplyProps> = props => {
     environment,
     conversation,
     onScroll,
+    onMount,
     openInquiryModal,
     openOrderModal,
   } = props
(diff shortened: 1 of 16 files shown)


dzucconi:
I'm maybe missing something obvious (forgive me I'm not too familiar with this part of the codebase) — but could the message components simply call [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) on their own refs when they mount?

dzucconi:
Or, rather, have a single empty div at the end that does that ([I've done something similar in the past for a chat interface](https://github.com/dzucconi/chat/blob/450c1b966a31c9d3d4f0da1bfe86381ebaaace47/src/components/Messages/Messages.tsx#L26))

ivan-badyulya:
> I'm maybe missing something obvious (forgive me I'm not too familiar with this part of the codebase) — but could the message components simply call [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) on their own refs when they mount?
@dzucconi 
Yes It's a good point also, I tried to do it, but I found out imperatively, that scrollTo works a bit smoother, thus I left it with scroll to as a result. I don't mind of both these approaches, so I can change it if it needed.

ivan-badyulya:
@dzucconi also it would be great of assistance if you see a more appropriate approach to make the message container scroll after all children render. Because sometimes, with this approach I still notice that the scrolling goes not all the way down, some component make a rerender after scrolling and it is impact the scrolling. Mb it is lazy loading of image?

dzucconi:
> it is lazy loading of image

If images don't have width/height attributes that'd do it sure (and would be a bad user experience that definitely needs to be fixed). Lazy loading itself wouldn't matter as long as the placeholders are sized.

dzucconi:
> works a bit smoother

As in the FPS of the scroll? I can't imagine it'd be _that_ different? IMO if it scrolls to the bottom simply, that'd be good enough for me — but again, I'm unfamiliar with this part of the product.

My problem with this approach is that it just seems very unintuitive — like looking at this code I'd be hard pressed to understand why any of this would have to happen.

ivan-badyulya:
hey @dzucconi!
I refactored a bit the functionality:
1) Replaced scrollTo with scrollIntoWview
2) Removed the flags due to their uselessness
3) Forwarded the scroll handler into the make offer button component and execute it on the component didMount because of only the button was impacting the layout of the message container, that's why the scrolling worked not all the way sometimes down, because it scrolled before the button is rendered

src/v2/Apps/Conversation/Components/Reply.tsx
```diff
@@ -154,6 +154,7 @@ export const Reply: React.FC<ReplyProps> = props => {
           conversation={conversation}
           openInquiryModal={() => openInquiryModal()}
           openOrderModal={() => openOrderModal()}
+          scrollMessageContainer={onScroll}
```
dzucconi, line 157: So this is the `onScroll` callback from some other component that gets called on mount in this component? The naming here is a little unclear to me — I would assume a container to be a component of some sort, not a callback. Can this be `onMount` or something similar?

dzucconi (approved):
This looks good. Minor comment about a naming thing. Feel free to merge when you're ready.

src/v2/Apps/Conversation/Components/Reply.tsx
ivan-badyulya: well, the thing is we have scrollToBottom handler in a parent component, and I have to throw it via several nodes. And the Reply component has the onScroll={scrollToBottom} prop, and use it for scrolling the message container on updating the chat after sending a message, not for execute it on mounting. So I decided not to create other prop with the same link to the same handler but use the existing onScroll prop.

ivan-badyulya: Conversation -> Reply (uses onScroll when message is sent and updated) -> ... -> MakeOfferButton (has to use onScroll for executing it on didMount)

ivan-badyulya: `
        <Reply
          onScroll={scrollToBottom}
          onMount={scrollToBottom}
        />
`
Do you want me to do like that?

dzucconi: Whatever makes sense to you — but in general I think passing any callbacks using the `onFoobar` naming convention is desirable.

ivan-badyulya: Refactored the naming issue

PR #7944, opened 2021-07-08
Bugfix: fix refetching on scroll in conversation list

lilyfromseattle:
[PURCHASE-2806]

Shoutout to @xtina-starr for finding this solution!

Conversations weren't refetching on scroll, and we realized that the `onScroll` method wasn't being hit. Moving the onScroll to a different container fixed this issue, and we were able to trace it back to this PR: https://github.com/artsy/force/pull/6135

[PURCHASE-2806]: https://artsyproduct.atlassian.net/browse/PURCHASE-2806

### DIFF
--- a/src/v2/Apps/Conversation/Components/ConversationList.tsx
+++ b/src/v2/Apps/Conversation/Components/ConversationList.tsx
@@ -13,7 +13,6 @@ import { ConversationListHeader } from "./ConversationListHeader"
 const Container = styled(Box)`
   height: 100%;
   overflow: hidden;
-  border-bottom: 30px solid ${color("white100")};
   border-right: 1px solid ${color("black10")};
   ${media.xs`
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Conversation/Components/ConversationList.tsx
```diff
@@ -13,7 +13,7 @@ import { ConversationListHeader } from "./ConversationListHeader"
 const Container = styled(Box)`
   height: 100%;
   overflow: hidden;
-  border-bottom: 30px solid ${color("white100")};
+  /* border-bottom: 30px solid ${color("white100")}; */
```
xtina-starr, line 16: I think you forgot to remove this.

xtina-starr (approved):
Other than the stray comment this looks good to me!

PR #7946, opened 2021-07-09
fix: the sending message container in a conversation overlaps the last message

ivan-badyulya:
**JIRA** -> [Inbox: Inquiry checkout: Mobile view: the latest message is overlapped by "The Artsy Guarantee" panel](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-2794&assignee=60469b1e43eac6006f636d07)

**Notes**:
The overlapping problem was in fixed position of sending message container with make offer button. Fixed position pulled the container out of the stream, so the chat container filled the space under it.
**Before**:
Screen Shot 2021-07-09 at 2 05 38 PM
**After**: 
Screen Shot 2021-07-09 at 2 11 05 PM

Other problem was in the empty after sending message container in a mobile media screen:
Screen Shot 2021-07-09 at 2 13 51 PM

was fixed by adding the media query:
```
const ConstrainedHeightContainer = styled(Box)`
  height: calc(100vh - 103px);
  @media ${themeGet("mediaQueries.xs")} {
    height: calc(100vh - 60px);
  }
`
```
GIF:
ezgif com-gif-maker (5)

### DIFF
--- a/src/v2/Apps/Conversation/Components/Conversation.tsx
+++ b/src/v2/Apps/Conversation/Components/Conversation.tsx
@@ -48,8 +48,6 @@ const Conversation: React.FC<ConversationProps> = props => {
     userHasLabFeature(user, "Web Inquiry Checkout") &&
     isArtworkOfferable
 
-  // Keeping track of this for scroll on send
-  const [lastMessageID, setLastMessageID] = useState<string | null>("")
   const [showConfirmArtworkModal, setShowConfirmArtworkModal] = useState<
     boolean
   >(false)
@@ -58,7 +56,6 @@ const Conversation: React.FC<ConversationProps> = props => {
 
   const scrollToBottom = () => {
     if (bottomOfMessageContainer.current) {
-      setLastMessageID(conversation?.lastMessageID)
       setTimeout(() => {
         bottomOfMessageContainer!.current!.scrollIntoView({
           behavior: "smooth",
@@ -67,7 +64,7 @@ const Conversation: React.FC<ConversationProps> = props => {
     }
   }
 
-  useEffect(scrollToBottom, [conversation, lastMessageID])
+  useEffect(scrollToBottom, [conversation])
 
   useEffect(() => {
     UpdateConversation(relay.environment, conversation)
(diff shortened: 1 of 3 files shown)


lilyfromseattle (approved):
Looks good! 🎉

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -67,7 +64,7 @@ const Conversation: React.FC<ConversationProps> = props => {
     }
   }
 
-  useEffect(scrollToBottom, [conversation, lastMessageID])
+  useEffect(scrollToBottom, [conversation])
```
jo-rs, line 67: @ivan-badyulya this actually triggers that on loading more messages on infinite scroll, every time, the user will be scrolled to the bottom instead of where they were. Please revert anything related to `lastMessageID`.

jo-rs (changes requested):
The issue seems to be fixed and looks good, but removing the `lastMessageID` logic causes other problems. So let's fix that before merging. Feel free to ping me if you have questions.

src/v2/Apps/Conversation/Components/Conversation.tsx
ivan-badyulya: @jo-rs sorry, I tried to understand why does it track the last ID but didn't think about infinite scroll. I wanted to ask you, but you were on vacation. Moved it back

ivan-badyulya: @jo-rs yes, I did a demo last week to discuss this stuff. The thing is the scrollIntoView works before some child component rerender and impact the scrolling. 1 stuff I noticed the wierd scrolling behavior is when we added the new make offer button. The button renders after scrollIntoView it impacts the message container height that's why scrolling wasn't working all the way down, that's why I execute the scrollToBottom on makeOfferButton did mount.

Bot now as I see the thing is not only in make offer button

ivan-badyulya: @jo-rs i tried to fix the stuff in the [pr ](https://github.com/artsy/force/pull/7932)

PR #7950, opened 2021-07-09
chore(shows2): roughs in current shows

dzucconi:
https://user-images.githubusercontent.com/112297/125113730-04da8980-e0b7-11eb-9d15-06b2cc8d3285.mov

-----

In general this isn't too notable yet. I need to add pagination but pagination doesn't work in Metaphysics for the top-level `showsConnection` field. I need to wrap our shows/feed endpoint for that to work, so...

But, in particular if you could take a look at aeed09cca3105c187ca1ab02830ca2fc439aad22 — I was a little frustrated with `RouterLink` being unintuitive, so now it supports all anchor and box props and no longer will complain about null hrefs.

### DIFF
--- a/src/v2/Apps/Partner/Components/PartnerShows/ShowCard.tsx
+++ b/src/v2/Apps/Partner/Components/PartnerShows/ShowCard.tsx
@@ -13,8 +13,7 @@ const ShowCard: React.FC<ShowCardProps> = ({ show }): JSX.Element => {
   const showType = isFairBooth ? "fair booth" : "show"
 
   return (
-    // @ts-expect-error STRICT_NULL_CHECK
-    <RouterLink to={href} noUnderline>
+    <RouterLink to={href} textDecoration="none">
       {coverImage && (
         <ResponsiveBox
           // @ts-expect-error STRICT_NULL_CHECK
--- a/src/v2/Apps/Shows/Components/ShowsCurrentShow.tsx
+++ b/src/v2/Apps/Shows/Components/ShowsCurrentShow.tsx
@@ -25,6 +25,8 @@ const ShowsCurrentShow: React.FC<ShowsCurrentShowProps> = ({ show }) => {
   const artworks = extractNodes(show.artworksConnection)
   const count = show.artworksConnection?.totalCount ?? 0
   const remaining = count - 15
+  const dateRange = [show.startAt, show.endAt].filter(Boolean).join(" – ")
+  const subtitle = [dateRange, show.location?.city].filter(Boolean).join(", ")
 
   return (
     <>
@@ -37,14 +39,7 @@ const ShowsCurrentShow: React.FC<ShowsCurrentShowProps> = ({ show }) => {
 
         <Spacer mt={1} />
 
-        <Text variant="md">
-          {[
-            [show.startAt, show.endAt].filter(Boolean).join(" – "),
-            show.location?.city,
-          ]
-            .filter(Boolean)
-            .join(", ")}
-        </Text>
+        <Text variant="md">{subtitle}</Text>
       </RouterLink>
 
       <Spacer mt={4} />
@@ -114,7 +109,7 @@ export const ShowsCurrentShowFragmentContainer = createFragmentContainer(
   }
 )
 
-export const SHOWS_CURRENT_SHOW_PLACEHOLDER = (
+export const ShowsCurrentShowPlaceholder: React.FC = () => (
   <Skeleton>
     <SkeletonText variant="lg">Partner Name</SkeletonText>
 
--- a/src/v2/Apps/Shows/Components/ShowsCurrentShow.tsx
+++ b/src/v2/Apps/Shows/Components/ShowsCurrentShow.tsx
@@ -1,0 +1,154 @@
+import {
+  Text,
+  Spacer,
+  Button,
+  Flex,
+  Skeleton,
+  SkeletonText,
+  SkeletonBox,
+  ResponsiveBox,
+  Box,
+} from "@artsy/palette"
+import React from "react"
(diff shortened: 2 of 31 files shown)


src/v2/System/Router/RouterLink.tsx
```diff
@@ -1,93 +1,73 @@
...
-        "aria-label",
-        "Component",
-        "activeClassName",
-        "className",
-        "data-test",
-        "exact",
-        "replace",
-        "role",
-        "style",
-        "tabIndex",
```
damassi, line 53: Are we now getting way more warnings in the dom? Other react-ish props that are passed down used to trigger a waterfall of warnings in the console

src/v2/System/Router/RouterLink.tsx
```diff
@@ -1,93 +1,73 @@
...
-          style={{ ...styleProps, ...(props as LinkPropsSimple).style }}
-          {...omit(props, [
-            "active",
-            "activeClassName",
-            "alignItems",
-            "hasLighterTextColor",
-            "justifyContent",
-            "noUnderline",
-            "textAlign",
-            "underlineBehavior",
```
damassi, line 83: ditto

src/v2/Apps/Partner/Components/PartnerShows/ShowCard.tsx
```diff
@@ -13,8 +13,7 @@ const ShowCard: React.FC<ShowCardProps> = ({ show }): JSX.Element => {
   const showType = isFairBooth ? "fair booth" : "show"
 
   return (
-    // @ts-expect-error STRICT_NULL_CHECK
-    <RouterLink to={href} noUnderline>
```
damassi, line 17: Thanks for the fix

src/v2/Apps/Shows/Components/ShowsCurrentShow.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+        <Spacer mt={1} />
+
+        <Text variant="md">
+          {[
+            [show.startAt, show.endAt].filter(Boolean).join(" – "),
+            show.location?.city,
+          ]
+            .filter(Boolean)
+            .join(", ")}
```
damassi, line 46: Can we move this to a variable?

src/v2/Apps/Shows/Components/ShowsCurrentShows.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }
+)
+
+const SHOWS_CURRENT_SHOWS_PLACEHOLDER = (
+  <Join separator={<Separator my={6} />}>
+    {SHOWS_CURRENT_SHOW_PLACEHOLDER}
+    {SHOWS_CURRENT_SHOW_PLACEHOLDER}
+    {SHOWS_CURRENT_SHOW_PLACEHOLDER}
+  </Join>
+)
```
damassi, line 100: This will feel better if its just `ReactComponentCase`, even if its just referring to already-invoked JSX -- future devs will see this CONSTANT and thing its like a string or something.

src/v2/System/Router/RouterLink.tsx
dzucconi: Nah, those stopped being a thing a while ago. (We can actually fix that correctly now in Styled Components 5, I believe)

src/v2/Apps/Shows/Components/ShowsCurrentShows.tsx
```diff
@@ -0,0 +1,128 @@
...
+      query={SHOWS_CURRENT_SHOWS_QUERY}
+      placeholder={SHOWS_CURRENT_SHOWS_PLACEHOLDER}
+      render={({ error, props }) => {
+        if (error) {
+          console.error(error)
+          return null
+        }
+
+        if (!props) {
+          return SHOWS_CURRENT_SHOWS_PLACEHOLDER
```
damassi, line 117: Or, re the above: `return <CurrentShowsPlaceholder />`. I don't think we need to worry to much about perf here, vs following existing convention. 

```
const CurrentShowsPlaceholder = () => ...
```

damassi (approved):
There's a couple questions in here about `RouterLink` now potentially emitting lots of new warnings (due to lack of props guards) and a suggestion about naming / form, but nothing blocking! Thanks for the cleanup work 👍

src/v2/System/Router/RouterLink.tsx
damassi: Ah, nice. What do we need to do in SC 5 (now that its already in?)

src/v2/Apps/Shows/Components/ShowsCurrentShows.tsx
dzucconi: Sort of disagree: PascalCase should definitely only be used for constructors/components. This is a very frustrating style inconsistency in our codebase where variables are sometimes in pascal case.

Though according to Airbnb style guide: https://github.com/airbnb/javascript — [uppercase consts that are exported

src/v2/Apps/Shows/Components/ShowsCurrentShows.tsx
dzucconi: Yeah, that I'm fine with.

src/v2/System/Router/RouterLink.tsx
dzucconi: lol I stand corrected, I was checking the browser console not the server console. Anyway we can fix this the right way sans safelist now: https://styled-components.com/docs/api#shouldforwardprop

> shouldForwardProp can take a second parameter that provides access to the default validator function. This function can be used as a fallback, and of course, it also works like a predicate, filtering based on known HTML attributes.

PR #7951, opened 2021-07-09
chore: bump Cohesion to 2.9.0

ashleyjelks:
Update's Cohesion version to 2.9.0.


xtina-starr:
It's weird that the PartnerProfile integration spec is failing. I'm going to try re-running the suit. It that doesn't work it might be worth trying to see if this spec is also failing locally @ashleyjelks.

PR #7957, opened 2021-07-09
chore: add inq checkout tracking analytics

ashleyjelks:
Adds tracking to Make Offer Inquiries when user taps Make Offer and when user confirms artwork before entering checkout flow.

https://artsyproduct.atlassian.net/browse/PURCHASE-2801


xtina-starr (approved):
🎉🎉

PR #7963, opened 2021-07-12
refactor(GRO-430): reduce dom node count; refactors class to functional components; defer save button rendering until hover

dzucconi:
These refactors reduce the number of DOM nodes by between a few hundred to a few thousand, depending on the page.

* A number of extraneous wrapper divs were removed from artwork metadata details
* Most significantly: we no longer render the save button until the parent is hovered. This necessitated the conversion of all it's consumers from class to functional components to use the hooks I wrote to assist in doing this. Previously we were rendering them but adjusting opacity to show and hide them.


dzucconi:
OK, it seems that we do want to fix the currently broken save button analytics — will do in the morning then this PR will be ready to go.

src/v2/Components/Artwork/Details.tsx
```diff
@@ -49,35 +40,33 @@ const ArtistLine: React.FC<DetailsProps> = ({
...
-                  <Text variant={tokens.variant}>{artist.name}</Text>
-                </ConditionalLink>,
-              ])
-            }, [])
-            .slice(1)}
-        </Text>
-      </TruncatedLine>
+      <Text variant={tokens.variant} overflowEllipsis>
+        {artists
+          .reduce((acc, artist, index) => {
```
jonallured, line 53: Off topic but I have a strong preference for extracting these types of values into a local constant. I pulled up this file and found it tricky to read this component but I now see there are three ways through:

* if there's a cultural maker, return that and bail on the rest of the function
* if there's no artists, bail
* if there are artists, join them with spaces and run them

I've got a local branch with this extracted and cleaned up a bit - you cool with me following up here after your works lands?

dzucconi: Yah that's fine. I also intend to go back over this component with the updated RouterLink as there's more nodes to be removed. Feel free to PR me once this is merged.

src/v2/Apps/Partner/Components/PartnerArtists/PartnerArtistDetails/PartnerArtistArtworksRail.tsx
```diff
@@ -61,6 +61,8 @@ export const PartnerArtistArtworksRail: React.FC<PartnerArtistArtworksRailProps>
               artwork={artwork}
               imageHeight={160}
               lazyLoad
+              // @ts-ignore TODO: Add relevant contextModule
+              contextModule={null}
```
damassi, line 65: This something that needs to be added to cohesion?

src/v2/Components/Artwork/FillwidthItem.tsx
```diff
@@ -43,129 +31,94 @@ export interface FillwidthItemContainerProps
...
-    // little safer and less confusing when that happens.
-    if (typeof aspectRatio !== "number") {
-      aspectRatio = 1
-    }
-
-    // NOTE: `aspectRatio` will default to `1` if image geometry is missing.
-    // Given that `imageHeight` is required, this means that any missing geometry
-    // will default to a square thumbnail.
-    return Math.floor(this.props.imageHeight * aspectRatio)
+export const FillwidthItem: React.FC<FillwidthItemProps> = ({
```
damassi, line 34: 🎉

damassi (approved):
Major props on converting those old components. Code wise things look good, but since this is a somewhat sensitive refactor PR can we: 
- Update PR title to include something like "+ refactors artwork class compnoents to functional components" 
- Update PR description with a _why_ about how this removes dom nodes. Its not immediately unclear from looking at code.

That way we'll be able to search / find stuff a little more easily should things need to addressed post-merge.

dzucconi:
🙏 — updated title/description

PR #7964, opened 2021-07-13
fix[PURCHASE-2793]: remove unwanted warning on redirect from the orders page

copperkraft:
solves [PURCHASE-2793](https://artsyproduct.atlassian.net/browse/PURCHASE-2793)

After the collector completes creating an order from the inquiry (using a button in conversation with the seller) user is redirected back to the conversation page. This redirect is interrupted by an "unsaved data" warning that makes no sense as the collector already completed the editing order at the moment. 

Solution: add check for "PUSH" action on the orders app to prevent "unsaved changes" warning on the application page.


jo-rs (approved):
It looks good, just curious if we should still have a default return. What do you think @copperkraft ?

PR #7967, opened 2021-07-13
feat(homepage): retires old home apps; promotes new home app

dzucconi:
Launches new homepage


src/v2/Apps/Home/homeRoutes.tsx
```diff
@@ -10,7 +10,7 @@ const HomeApp = loadable(
 export const homeRoutes: AppRouteConfig[] = [
   {
     theme: "v3",
-    path: "/home",
+    path: "//",
```
dzucconi, line 13: This is the only slightly weird thing. I guess maybe this warrants a comment. @damassi do you know why this has to happen? We're stripping the leading `/`?

dzucconi:
Ah, way to go smoke tests: we need to implement the flash message!

src/v2/Apps/Home/homeRoutes.tsx
damassi: Not sure if you saw the slack thread, but reposting here: https://github.com/artsy/force/blob/master/src/v2/System/Router/buildAppRoutes.tsx#L71-L75. 

Perhaps this has something to do with it?

damassi (approved):
This looks good to me, though before merge we should follow through with the convo around moving the Shows rail up closer to the top fold (above 19 gene rails) to make the home page more interesting. See thread [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1626196024491500).  

(Follow-up item will be A/B testing this placement.)

src/v2/Apps/Home/homeRoutes.tsx
dzucconi: > If the last character is a `/` trim the first character of the path.

What does this even mean? Do you happen to know the why of this?

dzucconi: These are the routes that get mounted:

```
/artist/:artistID
/artist-series/:slug
/artists
/artwork/:artworkID/:optional?
/auctions
/buyer-guarantee
/collect/:medium?
/collect/color/:color?
/collections
/collection/:slug
/consign
/consign/offer/:offerID
/user/conversations
/user/conversations/:conversationID
/example
/fair/:slug
/fair/:slug
/fairs
/art-fairs
/feature/:slug
/gene/:slug
/
/identity-verification/processing
/identity-verification/error
/identity-verification/:id
/order(2|s)/:orderID
/partner/:partnerId
/user/payments
/user/purchases
/search
/show/:slug
/show/:slug
/shows2
/tag/:slug
/unsubscribe
/viewing-rooms
/viewing-room/:slug
/debug
```

The only one that would actually match this is the new homepage. I think this code is safe to remove?

dzucconi: eea0223c9562c6a538219be6f4c7fd10e00300e1

damassi: I'd be triple sure its safe to remove, but 👍

dzucconi: Lemme try to hunt down the PR it was introduced in.

dzucconi:
🤔

src/v2/Apps/Home/homeRoutes.tsx
dzucconi: This was added in this commit: https://github.com/artsy/reaction/commit/5ee5f1fee212439b6ec8bd6e02a4fbdf93ddf044 / this PR: https://github.com/artsy/reaction/pull/3043

I don't know what its original intent was; but this code hasn't executed in a very long time.

dzucconi:
Thinking I'm just going to pend the flashMessage spec for now. They pass locally; and the flash messages work in staging, so I'm not sure what's up.

damassi:
Np, let's just add a todo or file a ticket and let's get this out! No more hard jumps from the home page 🥳 Also curious how this fixes our homepage deepcrawl meta tag errors.

PR #7969, opened 2021-07-13
fix: save button tweaks

dzucconi:
OK, this adds some `data-test` selectors to the remaining artwork items so that I can select them in Integrity.

Though this prompted me to find a comment I left for myself to disable the save buttons on touch (which is the current prod behavior). And removing another DOM node from the stack as well.


src/v2/Components/Artwork/SaveButton/useSaveButton.tsx
```diff
@@ -13,18 +14,18 @@ export const useSaveButton = ({
...
     },
     onMouseLeave: (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+      if (isTouch) return
       !isSaved && setIsHovered(false)
       onMouseLeave?.(event)
     },
   }
 
-  return { isSaveButtonVisible: isHovered, containerProps }
+  return { isSaveButtonVisible: isTouch ? false : isHovered, containerProps }
```
dzucconi, line 30: I guess this can be `isHovered && !isTouch`

dzucconi:
Deleting this src/v2/Components/__tests__/Artwork.jest.tsx snapshot spec...

PR #7974, opened 2021-07-14
fix(PURCHASE-2810): unwanted redirect to conversations from eigen

copperkraft:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-2810

Add check for the eigen to prevent unwanted redirect to the conversation. Instead, the user stays at the order status page.


jo-rs (approved):
Nice nice 💯

PR #7441, opened 2021-04-30
feature: Add artists filter to gene page

olerichter00:
Resolves [FX-2859]

[FX-2859]: https://artsyproduct.atlassian.net/browse/FX-2859

## Screenshots

image
image

image

PR #7976, opened 2021-07-14
fix: fetch follow artists

dimatretyak:
Type: **Fix**
Jira: [FX-2859]

### Description
The initial implementation was made in PR #7441, but later an error was discovered (relay sends `null` as a default value for optional arguments – on our case this is `fairId`), which this PR corrects

[FX-2859]: https://artsyproduct.atlassian.net/browse/FX-2859

### DIFF
--- a/src/v2/Components/ArtworkFilter/Utils/fetchFollowedArtists.ts
+++ b/src/v2/Components/ArtworkFilter/Utils/fetchFollowedArtists.ts
@@ -1,11 +1,50 @@
 import { Environment, fetchQuery } from "relay-runtime"
 import { graphql } from "react-relay"
-import { fetchFollowedArtistsQuery } from "v2/__generated__/fetchFollowedArtistsQuery.graphql"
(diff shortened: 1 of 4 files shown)


dzucconi:
I'll be honest: I somehow missed this `fetchFollowedArtists` — Can we do this a bit more conventionally: associate a fragment with a component, etc...

dimatretyak:
> I'll be honest: I somehow missed this `fetchFollowedArtists` — Can we do this a bit more conventionally: associate a fragment with a component, etc...

I think that it will be possible to do it in 2 ways:
1. Wrap the `ArtistsFilter` component in a `QueryRender` and get the artists that the user is subscribed to
2. Wrap the `ArtistsFilter` component in `createFragmentContainer,` pass `me` to the prop from which we will receive the artists that the user is subscribed to

dzucconi (approved):
If this fixes the immediate issue I'd say go ahead and merge. If you want to follow up with something more conventional — or at least leave a comment or something — I leave that up to you.

dimatretyak:
I discussed the current implementation with Matt Zikherman and decided to merge the current implementation. In the future, we can refactor the code, but test it very carefully

PR #7977, opened 2021-07-14
sentry: Add `Illegal Invocation` to list

damassi:
Noticing that this error keeps spamming sentry, and is coming from a 3rd party script.

### DIFF
--- a/src/lib/analytics/sentryFilters.ts
+++ b/src/lib/analytics/sentryFilters.ts
@@ -8,6 +8,7 @@ export const IGNORED_ERRORS = [
   "cancelled",
   "Failed to fetch",
   "gpt/pubads_impl",
+  "Illegal invocation", // From   "gpt/pubads_impl", script (3rd party)
   "Non-Error exception captured",
   "Non-Error promise rejection captured",
   "Origin is not allowed by Access-Control-Allow-Origin",


src/lib/analytics/sentryFilters.ts
```diff
@@ -8,6 +8,7 @@ export const IGNORED_ERRORS = [
   "cancelled",
   "Failed to fetch",
   "gpt/pubads_impl",
+  "Illegal invocation", // From   "gpt/pubads_impl", script (3rd party)
```
joeyAghion, line 11: This is a very low-level js error message (something about an improperly bound `this`), right? My worry is that this filter could suppress serious errors in our own code. Is there a way to scope it to the 3rd party scripts?

damassi: This error is pretty rare in my experience, at least in terms of stuff we need to watch. But I agree, it is low-level. I tried to filter by the `gpt/pubads` above; it's not catching it unfortunately.

joeyAghion: Unrelated to this PR, I continue to see some of the messages ignored above in our Sentry events, like `AbortError: The user aborted a request.`. Does this list not apply everywhere I think it does?

damassi: I wonder if there might be a slight difference in the message, but yeah, what happens is that this filter should match against the error and prevent it from dispatching to Sentry. (Will need to look at things more closely.)

joeyAghion: I'll merge but am a little worried this will be a future head-scratcher. Hopefully the decreased noise is worth the risk!

joeyAghion: (Or, rather, I'll label `squash on...`.)

PR #7980, opened 2021-07-14
chore(home): moves shows module up

dzucconi:

### DIFF
--- a/src/v2/Apps/Home/Components/HomeArtworkModules.tsx
+++ b/src/v2/Apps/Home/Components/HomeArtworkModules.tsx
@@ -1,5 +1,5 @@
 import React from "react"
-import { Join, Spacer } from "@artsy/palette"
+import { Join, Separator, Spacer } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
 import { HomeArtworkModules_homePage } from "v2/__generated__/HomeArtworkModules_homePage.graphql"
 import {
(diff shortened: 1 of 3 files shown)


damassi:
@dzucconi  - mind adding a screenshot of logged in view? Noticing over here that there's 
- Recently Viewed
- Similar to ...
- Recently Saved 
- Similar to ... 

And those user-centric items seem more important than shows being second.

dzucconi:
That is the logged in view

dzucconi:
OK so I'll have to get crafty about threading this in.

damassi:
Can we move it below the user-centric items? Or are you logged in from a fresh account and so those rails don't appear? 

The main idea is to pull this shows content up above the sea of gene rails, which are numerous and somewhat arbitrary.

damassi:
> OK so I'll have to get crafty about threading this in.

(Don't want to block things -- only if its straightforward)

dzucconi:
@damassi take a look at aa1cb6ca794f16751b5f22400719c86754dbbca8 — not pretty but it does the trick.

### Logged in:
https://user-images.githubusercontent.com/112297/125680766-4cc86829-1db1-4ff8-bec7-93675b6fc570.mov

### Logged out:
https://user-images.githubusercontent.com/112297/125683497-360644fd-b598-4f73-a531-a013ce2996cf.mov

src/v2/Apps/Home/Components/HomeArtworkModules.tsx
```diff
@@ -16,12 +18,50 @@ const HomeArtworkModules: React.FC<HomeArtworkModulesProps> = ({
...
+              relatedArtistID: "IGNORE",
+              followedArtistID: "IGNORE",
+            },
+          },
+          artworkModule,
+        ]
+      }
+
+      return artworkModule
+    })
```
damassi, line 47: 🚢 🚀

damassi (approved):
💯

PR #7981, opened 2021-07-14
chore(show): adds scrollbar; minor clean up

dzucconi:
Now that we have the scrollbars we can enable them for the install shots shelf. Fixes a few other minor visual inconsistencies.


damassi:
mergeOnGreen

PR #7982, opened 2021-07-15
FX-3066 feat: Auction House filter is not expanded by default

anastasiapyzhik:
[FX-3066]

Before
Screenshot 2021-07-15 at 14 24 41

After
Screenshot 2021-07-15 at 14 23 12

[FX-3066]: https://artsyproduct.atlassian.net/browse/FX-3066


mdole (approved):
Sweet! Let's ship it

PR #7984, opened 2021-07-15
FX-3067 feat: Rearranged existing filters

anastasiapyzhik:
[FX-3067]

**Before**
Screenshot 2021-07-15 at 17 53 30

**After**
Screenshot 2021-07-15 at 17 53 45

[FX-3067]: https://artsyproduct.atlassian.net/browse/FX-3067


dblandin:
I'm a bit hesitant to merge this without having a cap on the number of artists we show in the filter menu. As is, the artist filter is very likely to push the medium and price filters below the fold.

I'll sync up with Thomas and Jonathan on a recommendation.

dblandin:
Closing for now as we don't want to reorder until we're also able to update the filter controls here.

PR #7985, opened 2021-07-15
feat(PX-4234): review page

kiryl-zubarau:
This PR updates review page regarding Arta shipping. Also added a redirect to the shipping page from payment, review and status pages if shipping quote is not selected.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4234

image


ansor4 (approved):
Nice! this looks good to me! Only #minor thing I noticed is that the pattern of 
```
      const selectedShippingQuote = extractNodes(
        extractNodes(order.lineItems)?.[0].shippingQuoteOptions
      ).find(shippingQuote => shippingQuote.isSelected)
```
is repeated a lot, so if you see opportunities to pull that into a helper in the future, I would! Not blocking though, just a note!

PR #7987, opened 2021-07-15
fix(details): fixes artist name with multiple artists

dzucconi:
Was scrolling down the new homepage when I encountered this beauty: 

Happened to be the exact piece of code you called out earlier. Here's a much simpler way to do this and fix the problem in one go 😬


jonallured (commented):
Lol oops I looked closely at this and did not even notice that mistake - good catch and love the refactor. 🤗

PR #7996, opened 2021-07-16
feat(GRO-386): rebuilds artwork lightbox; resizes primary artwork image

dzucconi:
Closes [GRO-386](https://artsyproduct.atlassian.net/browse/GRO-386)

The main focus of this change is the new `ArtworkLightbox.tsx` — which replaces `Lightbox.tsx`. `Lightbox.tsx` becomes `DeepZoom.tsx` for now pending it's own rewrite. I'm passing its onClick handler via function children for now, but this will go away with the rewrite.

The only change that will be noticeable to the end user is that we now will have larger and higher quality images for the primary artwork image. Despite being larger and looking better, most of the time they are smaller. _Much_ smaller in the case of any 1x screens. (like if the `large.jpg` was ~200kb; the new 2x weighs around ~150kb and the new 1x weighs around ~60kb — though I didn't take a very exhaustive survey here.)

<del>Leaving this in draft over the weekend and we'll get it merged on Monday.</del>


src/v2/Utils/Hooks/useNextPrevious.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export interface UseNextPrevious {
+  onNext(): void
+  onPrevious(): void
+}
+
+/**
+ * Binds left and right keyboard arrows to next/previous functions if the element
+ * with `containerRef` if anything within it is in focus.
+ */
```
damassi, line 11: 👍

damassi (approved):
A+

PR #8001, opened 2021-07-19
chore: Cleanup Old Code

icirellik:
Removes the following two folders:

/src/desktop/apps/ssr-experiments
/src/desktop/apps/webpack

The following folder was cleaned up:

/src/desktop/apps/auctions/utils


damassi:
👍

PR #8005, opened 2021-07-19
fix: collection filter disappear

dimatretyak:
Type: **Fix**
Jira: [FX-3082]

### Description
When the user selects any filter value on collection page, all the filters whose values we get from aggregations disappear (only those filters with hardcoded values don't disappear - medium, rarity, size, ways to buy, time period, color)

#### Demo
https://user-images.githubusercontent.com/3513494/126481577-672abe25-0803-4231-a915-70ece9466c41.mp4

[FX-3082]: https://artsyproduct.atlassian.net/browse/FX-3082


dblandin (approved):

src/v2/Apps/Collect/Routes/Collection/Components/CollectionArtworksFilter.tsx
```diff
@@ -0,0 +1,91 @@
...
+    <ArtworkFilterContextProvider
+      // Reset state of filter context without calling reset; which would
+      // affect analytics.
+      key={pathname}
+      filters={match.location.query}
+      sortOptions={[
+        { text: "Default", value: "-decayed_merch" },
+        {
+          text: "Price (desc.)",
+          value: "sold,-has_price,-prices",
```
dblandin, line 38: I'm not familiar with these sort options. Why do we have three values here?

dimatretyak: This value [was specified](https://github.com/artsy/force/pull/8005/files#diff-0e882eaf2d4c702041d1a6a2ec52c47a931b5f44cbebd4ec1637284832aaea2fL137-L138) for the sort filter. I just moved the existing code to a separate component to use `createRefetchContainer`

PR #8006, opened 2021-07-19
fix: address line rendering

kiryl-zubarau:
This PR fixes partner profile page rendering if does not have a specified city.

image


sweir27:
Thank you!

PR #8010, opened 2021-07-20
fix(PURCHASE-2768): Anchoring of Nav items to the correct section

gkartalis:
# Description

Resolves [PURCHASE-2768](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-2768)

**Problem:** When a user selects Payments & Shipping from the burger menu they arrive at the ‘My Profile’ page with the wrong horizontal nav item in view. The right item is selected but not in view.

**Solution:** Adds scrollIntoView on active tab

### Video

https://user-images.githubusercontent.com/21178754/126458687-b9a1c798-020c-4359-95ae-4c8d036e15e6.mov

### DIFF
--- a/src/v2/Components/UserSettings/UserSettingsTabs.tsx
+++ b/src/v2/Components/UserSettings/UserSettingsTabs.tsx
@@ -58,7 +58,6 @@ export const UserSettingsTabs: React.FC<UserSettingsTabsProps> = track()(
             const isActive = tabClass(route.url, route) === "active"
             const { ref } = useScrollIntoView(isActive)
             return (
-              // @ts-ignore
               <Box ref={ref} key={index}>
                 <RouteTab className={tabClass(route.url, route)} to={route.url}>
                   {route.name}
--- a/src/v2/Components/UserSettings/UserSettingsTabs.tsx
+++ b/src/v2/Components/UserSettings/UserSettingsTabs.tsx
@@ -42,6 +42,22 @@ const tabClass = (tabUrl, route) => {
   return currentRoute === tabUrl ? "active" : undefined
 }
 
+const useScrollIntoView = (isRouteActive: boolean) => {
+  const ref = React.useRef() as React.MutableRefObject<HTMLDivElement>
+  React.useEffect(() => {
+    if (isRouteActive && ref?.current) {
+      ref.current?.scrollIntoView &&
+        ref.current.scrollIntoView({
+          behavior: "smooth",
+          block: "end",
+          inline: "nearest",
+        })
+    }
+  }, [isRouteActive])
+
+  return { ref }
+}
+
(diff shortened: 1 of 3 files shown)


src/v2/Components/UserSettings/UserSettingsTabs.tsx
```diff
@@ -42,6 +42,22 @@ const tabClass = (tabUrl, route) => {
...
+        ref.current.scrollIntoView({
+          behavior: "smooth",
+          block: "end",
+          inline: "nearest",
+        })
+    }
+  }, [isRouteActive])
+
+  return { ref }
+}
```
damassi, line 59: Note that we already have [this helper](https://github.com/artsy/force/blob/3afca276d9dcc6050e762264fc8497511e5ab8da/src/v2/Utils/scrollHelpers.ts#L15), and we use by attaching an `#scrollTo--whatever` id directly on the target dom node. Short and sweet. 

That said, if we need to use something more dynamic with reffs and so on, feel free to add a new function within the `scrollHelpers` file linked above.

gkartalis: I moved the hook into `scrollHelpers`. Tried to use [this helper](https://github.com/artsy/force/blob/3afca276d9dcc6050e762264fc8497511e5ab8da/src/v2/Utils/scrollHelpers.ts#L15) but in this case it wasn't working (I guess because the scroll area is nested in another one and it is horizontal).

But if I am missing something and it can be done with [this helper](https://github.com/artsy/force/blob/3afca276d9dcc6050e762264fc8497511e5ab8da/src/v2/Utils/scrollHelpers.ts#L15) I would love to reuse that.

damassi: The hook seems useful! all good 👍

src/v2/Utils/scrollHelpers.ts
```diff
@@ -24,3 +26,19 @@ export const scrollIntoView = (props: ScrollIntoViewProps) => {
     })
   }
 }
+
+export const useScrollIntoView = (isRouteActive: boolean) => {
+  const ref = React.useRef() as any
+  React.useEffect(() => {
+    if (isRouteActive && ref?.current) {
```
damassi, line 33: For something generic like `scroll thing into view` whats the purpose of `isRouteActive`?

damassi:
@gkartalis  - can this be closed?

gkartalis:
I think yes, maybe @jo-rs knows more about this since I did that during my rotation

damassi:
@gkartalis - can this be closed?

gkartalis:
yup, closing it!

PR #8027, opened 2021-07-20
feature: Server Timing Headers

icirellik:
Enables server timing headers to help identify performance issues. Then
headers when enabled follow the Server Timing spec:

https://w3c.github.io/server-timing/#the-server-timing-header-field

Additionally, there is a separate commit that correctly sets the types for
Response and Request options to fix type errors.

server-timing

### DIFF
--- a/src/lib/middleware/serverTimingHeaders.ts
+++ b/src/lib/middleware/serverTimingHeaders.ts
@@ -1,0 +1,52 @@
+import type { NextFunction } from "express"
+import type { ArtsyRequest, ArtsyResponse } from "./artsyExpress"
+
+export function setTimingHeader(
+  res: ArtsyResponse,
+  name: string,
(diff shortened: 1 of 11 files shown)


src/lib/middleware/serverTimingHeaders.ts
```diff
@@ -0,0 +1,52 @@
...
+        timingHeaders.push(key)
+      } else {
+        timingHeaders.push(`${key};dur=${duration}`)
+      }
+    }
+
+    res.serverTimingHeadersWritten = true
+    res.set("Server-Timing", timingHeaders.join(", "))
+
+    return defaultWritehead.apply(this, arguments)
```
mc-jones, line 49: Nice!

mc-jones (approved):

PR #8028, opened 2021-07-20
refactor(artwork): rebuilds deepzoom

dzucconi:
Closes: [GRO-387](https://artsyproduct.atlassian.net/browse/GRO-387)

No surprises here.


src/v2/Components/DeepZoom/DeepZoomSlider.tsx
```diff
@@ -1,165 +1,134 @@
...
 `
 
 const knobStyles = css`
   user-select: none;
   cursor: pointer;
-  width: ${space(2)}px;
-  height: ${space(2)}px;
-  border-radius: 20px;
-  background-color: ${color("white100")};
+  width: ${themeGet("space.2")};
```
damassi, line 59: cc @pvinis - what we were talking about

PR #8031, opened 2021-07-20
feat: (PLATFORM-3135) turn-key dev env

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-3135

This PR introduces "_turn key [development environments](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#development-environments)_" pattern into force.

The changes include:

1. Introduce a setup script that automates project setup.
2. Introduce a set of dot env configuration files:
    1. Adhering to [shared-configuration](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#shared-configuration) introduce `.shared.env` (for shared & sensitive values) and `.env.example` (for dev overrides)
3. Reduce `.env.oss` and use the `.env.oss` in place of `.env.shared` for OSS setup.
4. Change how `dotenv` is used by removing preloading from scripts and instead leveraging [loadenv.js](https://github.com/artsy/force/pull/8031/files#diff-7947041f69488c02bb7c290b35815b7136d488d6f44b8a68218658a365cff5ad) to detect/parse dot env files. The script loads file contents into `process.env` during development and when running in production mode locally:
    1. `loadenv.js` will only modify `process.env` if dot env files exist and are not empty
    2. Update/Remove `dotenv/config` imports
5. Update documentation to reflect the new setup process

---

Additional points to note:

* A number of empty environment variables have been removed from both, `.env.oss` and `.env.shared` (version in s3). If this is potentially problematic and I've overlooked something, please point that out.
* After discussing with @joeyAghion, we decided to keep the `.example.env` as slim as possible and future amendments to dot env files should likely go into the shared copy that lives in s3.
* As it stands the setup directions have been trimmed down to 3 commands:
    * `git clone git@github.com:artsy/force.git && cd force`
    * `./scripts/setup.sh && nvm use`
    * `yarn start`
 Please validate the ^ commands. 🙏 
---

More comments/questions below.


CONTRIBUTING.md
```diff
@@ -96,16 +67,13 @@ that behavior even for subdomains (such as local.artsy.net).
...
-3. Update your `.env` file's `APP_URL`, `RECAPTCHA_KEY`.
+2. Update your `.env` file's `APP_URL`:
 
 ```
 APP_URL=http://local.artsy.net:5000
-RECAPTCHA_KEY=REPLACE_ME
 ```
 
-4. Visit [`http://local.artsy.net:5000`](http://local.artsy.net:5000).
+3. Restart the server and navigate to [`http://local.artsy.net:5000`](http://local.artsy.net:5000).
```
damassi, line 76: These changes are all 🔥 🔥 👌

package.json
```diff
@@ -40,11 +40,11 @@
     "sentry": "scripts/sentry.sh",
     "start": "scripts/start.sh",
     "start:dev": "concurrently 'yarn start:dev:relay' 'yarn start:dev:server'",
-    "start:dev:server": "NODE_ENV=development node --max_old_space_size=3072 -r dotenv/config ./src/dev.js",
```
ovasdi, line 43: I've removed a number of `dotenv` preloads and instead import/require `loadenv` within scripts. I tested and they seemed to work. Please let me know if this may cause issues i've not accounted for.

src/dev.js
```diff
@@ -18,9 +18,6 @@ const { createReloadable } = require("@artsy/express-reloadable")
 const { createConfig } = require("../webpack")
 const { initializeMiddleware } = require("./middleware")
 
-// @ts-ignore
-const { APP_URL, PORT } = require("./config")
```
ovasdi, line 22: Removed these since the vars aren't used in here. Is this 🆗 ?

packages/publish/uploadToS3.js
```diff
@@ -1,4 +1,4 @@
-require("dotenv/config")
+require("../../src/loadenv")
```
ovasdi, line 1: I've updated this require to use `loadenv` but having a hard time figuring out how this gets called/triggered. Could someone provide context around this one?

.env.example
```diff
@@ -0,0 +1,39 @@
...
+# -- Full logs from Express
+# VERBOSE_LOGGING=true
+#
+# -- Display analytics calls in browser console
+# SHOW_ANALYTICS_CALLS=true
+
+# ------------------------
+# Localhost Environment
+# ------------------------
+# REQUIRED SETUP: See https://github.com/artsy/force/blob/master/CONTRIBUTING.md#authentication-in-your-local-force-app
```
damassi, line 36: Nice 👍

packages/publish/uploadToS3.js
damassi: Hokusai triggers the main `package.json` script [here](https://github.com/artsy/force/blob/b6c35ae463eed83ba5a0e99086a96f69c65eaed9/hokusai/config.yml#L2), which then calls out to this sub-package [here](https://github.com/artsy/force/blob/b6c35ae463eed83ba5a0e99086a96f69c65eaed9/packages/publish/package.json#L11).

I'm not sure exactly why this script was moved to this two-call form, but I would advice moving the second call directly into force and avoiding the one-off confusion around nesting mono-repo-esque packages in force (which isn't a mono-repo). 

We used to have this `uploadToS3` file in the root `scripts`, which was simply called from the root package.json. Now theres a few levels of unnecessary indirection.

scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+# This assumes you have general prerequisites installed as by:
+# https://github.com/artsy/potential/blob/master/scripts/setup
+
+# Exit if any subcommand fails
+set -e
+
+echo "Brew update and bundle install..."
+brew update
+brew bundle --file=- <<EOF
+brew 'yarn'
```
damassi, line 12: Maybe we can test if yarn is already installed before adding here?

damassi (approved):
Nice work on this @ovasdi 👍 One suggestion around removing confusion but otherwise looks great.

.env.oss
```diff
@@ -1,85 +1,28 @@
...
-ADMIN_URL=http://admin.artsy.net
-CMS_URL=http://cms.artsy.net
-CONSIGNMENTS_APP_URL=https://sell.artsy.net/
-CONVECTION_APP_URL=
-CONVECTION_APP_ID=
-CRITEO_ARTWORKS_ACCOUNT_NUMBER=
-CRITEO_AUCTIONS_ACCOUNT_NUMBER=
-GALAXY_PUBLISHABLE_TOKEN=
-GALAXY_URL=https://galaxy-staging.artsy.net/
+CONSIGNMENTS_APP_URL=https://sell.artsy.net
```
joeyAghion, line 22: I think this app no longer exists, and can be removed along with [its references](https://github.com/artsy/force/search?q=CONSIGNMENTS_APP_URL).

.env.oss
```diff
@@ -1,85 +1,28 @@
...
-
-# Needed to run Reaction via yarn link
-GRAPHQL_ENDPOINT=https://metaphysics-staging.artsy.net
-
-# Mobile
-CALENDAR_URL=
-EMBEDLY_KEY=
-GOOGLE_ANALYTICS_ID=
-MIXPANEL_ID=
+GALAXY_URL=https://galaxy-production.herokuapp.com
```
joeyAghion, line 28: This app hasn't been on Heroku for a while. I think the removed setting (pointing at galaxy-staging.artsy.net) is more accurate.

CONTRIBUTING.md
```diff
@@ -4,57 +4,21 @@ This project is work of [many developers](https://github.com/artsy/force/graphs/
...
 
-## Run Force
-
-Install [NVM](https://github.com/creationix/nvm) and Node 12:
+Execute the setup script:
 
 ```sh
-nvm install 12
-nvm alias default 12
+./scripts/setup.sh
```
joeyAghion, line 18: Based on our experimentation, I thought `source ./scripts/setup.sh` was the way to ensure the `nvm` choice persisted.

Dockerfile
```diff
@@ -109,6 +109,7 @@ COPY patches ./patches
 COPY src ./src
 COPY webpack ./webpack
 COPY .env.oss \
+  .env.example \
```
joeyAghion, line 112: ~Why does the container depend on this file now?~ _I see how the smoke tests uses it._

scripts/smoke_test.sh
```diff
@@ -4,8 +4,8 @@ set -ex
 
 # provide a default environment if none exists
 if [ ! -f ".env" ]; then
-    echo "Using .env.oss"
-    cp .env.oss .env
+    echo "Using .env.example and .env.oss as .env"
+    cat .env.example > .env && cat .env.oss >> .env
```
joeyAghion, line 8:
```suggestion
    cat .env.example .env.oss > .env
```

scripts/setup.sh
```diff
@@ -0,0 +1,42 @@
...
+# Test s3 access and download the shared force .env or use .env.oss as shared
+if [[ $(aws s3 ls s3://artsy-citadel/dev/ > /dev/null 2>&1 && echo $?) != 0 ]]; then
+  echo "Unable to download shared config from s3. Using .env.oss!"
+  echo "This is expected for open source contributors."
+  echo "If you work at Artsy, please check your s3 access."
+  cp .env.oss .env.shared
+else
+  echo "Downloading .env.shared (for common local dev config)..."
+  aws s3 cp s3://artsy-citadel/dev/.env.force .env.shared
+fi
```
joeyAghion, line 33: Might be simpler to just attempt the `aws s3 cp ...` and fall back if that fails (rather than 2 separate AWS calls).

packages/publish/uploadToS3.js
joeyAghion: I can understand the intent to _not_ require an entire Force environment (with all of its dependencies) just to upload to S3, or for Force dependencies to be bloated by AWS requirements. On the other hand, if this command is being run from an already-built Force image and it already depends on S3 in other ways, it doesn't make much of a difference.

Either way, assuming this script continues to work, that can be a separate PR.

joeyAghion (approved):
* Would you recommend other node projects borrow this `loadenv` script in order to respect multiple environment files? If so, we should update Metaphysics for consistency as well as the [doc's examples](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#examples).

I'm noticing that [defaults in the source code](https://github.com/artsy/force/blob/master/src/desktop/config.coffee) are inconsistent about referring to staging or production services. If we could agree to prefer staging values as defaults, it looks like we could clean up the environment further. (_Not_ for this PR.)

Now, to try the updated script out on a fresh environment...

packages/publish/uploadToS3.js
icirellik: This is to keep CI dependencies out of our production images, they use a significant amount of disk. The perfect solution here would be a separate image that volume mounts force to upload our assets if anyone wants to take that one!

scripts/smoke_test.sh
ovasdi: going to undo this change as `.env.example` has been reduced to app settings and a couple of local vars.

CONTRIBUTING.md
ovasdi: I chose not to add that and instead add a "_note to nvm users_" because I was experiencing terminal window exit when script threw an error. I'll test this further and experiment with `|| return` that you pointed out so as to add `source` prefix.

ovasdi:
> * Would you recommend other node projects borrow this `loadenv` script in order to respect multiple environment files? If so, we should update Metaphysics for consistency as well as the [doc's examples](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#examples).
> 
> I'm noticing that [defaults in the source code](https://github.com/artsy/force/blob/master/src/desktop/config.coffee) are inconsistent about referring to staging or production services. If we could agree to prefer staging values as defaults, it looks like we could clean up the environment further. (_Not_ for this PR.)
> 
> Now, to try the updated script out on a fresh environment...

Regarding `loadenv`, let's see how it works out with this change but yes a followup to update MP and/or docs would be a good next step.

Regarding "staging as defaults", yea the values are a bit scattered in the checked in configs. I updated dotenvs to default to staging in a few vars that didnt.

CONTRIBUTING.md
joeyAghion: This works for me.

joeyAghion (approved):
Feel free to merge when satisfied!

scripts/diagnose.js
```diff
@@ -0,0 +1,40 @@
...
+  if (task.status != 0) {
+    throw new Error(task.stderr.toString())
+  }
+  return task.stdout.toString()
+}
+
+const checkEnvVariables = () => {
+  // exec("touch .env.temp")
+  try {
+    exec("aws s3 cp s3://artsy-citadel/dev/.env.eigen .env.temp")
```
damassi, line 15: should this be `.env.force`?

src/index.js
```diff
@@ -3,20 +3,19 @@
...
 import express from "express"
-import config from "./config"
 import chalk from "chalk"
 import { startServer } from "./lib/startServer"
 import legacyDesktopApp from "./desktop"
 import legacyMobileApp from "./mobile"
 import forceV2 from "./v2/server"
 
-console.log(chalk.green(`\n[Force] NODE_ENV=${config.NODE_ENV}\n`))
+console.log(chalk.green(`\n[Force] NODE_ENV=${process.env.NODE_ENV}\n`))
```
icirellik, line 18: ✨

icirellik (approved):
Aside from the one item @damassi pointed out this looks awesome.

scripts/diagnose.js
ovasdi: I didn't meant to check that one in. That was a mistake.

ovasdi:
> Aside from the one item @damassi pointed out this looks awesome.

Thanks @icirellik, I mean't to leave `diagnose.js` to tackle on _Future Friday_ and checked it in mistakenly.

docs/env_configuration.md
```diff
@@ -1,14 +1,18 @@
+## Shared Configuration
```
ovasdi, line 1: @damassi I've added a section to explain and link to shared configuration process/docs.

damassi: 👀

docs/env_configuration.md
```diff
@@ -1,14 +1,18 @@
...
+
+Force uses [shared configuration](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#shared-configuration) to distribute common and sensitive configuration values. The [setup script](../scripts/setup.sh) will download `.shared.env` and also initialize a `.env` (from `.env.example`) for developer custom configuration and any overrides.
+
 ## Adding an ENV var
 
 ### Server side only
 
-1. Add the variable to your `.env` file.
+1. Add the variable to `.shared.env` file and update the S3 version.
```
damassi, line 9: (Looks like suggest edit button went away)

Even though the shared config link is just above, it'd still be helpful to link it here to, so

```
1. Add the variable to `.shared.env` file and [update the S3 version](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#shared-configuration).
```

(Side note -- very much share the same concern [as @jonallured here](https://artsy.slack.com/archives/CP9P4KR35/p1626960318203300?thread_ts=1626932540.200100&cid=CP9P4KR35) about these values not living in version control. At the very least it'd be interesting to see historical values as they change through time. ENVs very much tell a story.)

ovasdi: I added the links.

Will think on / start a discussion about how we should handle tracking of ENVs over time. Preferably the process should work for both, private and OSS projects that use this `.env.shared` configuration approach.

docs/env_configuration.md
```diff
@@ -1,14 +1,18 @@
...
 1. Edit `src/desktop/config.coffee` and `src/mobile/config.coffee` to include default for your env variable. You can safely use `null` to indicate an empty variable.
 1. Use it in the runtime via `/src/config.js`
 
 ### Server + Client
 
-1. Add the variable to your `.env` file.
+1. Add the variable to your `.shared.env` file and update the S3 version.
```
joeyAghion: (Late comment)

`.env.shared`

docs/env_configuration.md
```diff
@@ -1,14 +1,18 @@
+## Shared Configuration
+
+Force uses [shared configuration](https://github.com/artsy/README/blob/main/playbooks/development-environments.md#shared-configuration) to distribute common and sensitive configuration values. The [setup script](../scripts/setup.sh) will download `.shared.env` and also initialize a `.env` (from `.env.example`) for developer custom configuration and any overrides.
```
joeyAghion, line 3: `.env.shared`

docs/env_configuration.md
ovasdi: Thanks will get these tomorrow.

PR #8033, opened 2021-07-20
relay cache: Bump up cache expiry times

damassi:
#### History:  

When we launched force as an SPA last year we also launched the ability to tap into the relay network layer as a server-side response cache, stored in Redis. Since the launch, we've migrated dozens of pages, with each being cacheable for logged-out users during the SSR pass, and for logged-in users, on the client (unless we explicitly decide to not cache, like on artwork). 

#### The Change: 
This PR bumps up our initial launch values from 100 cache entries and a 15min expiry to 2000 entries and a 1hr expiry. We should continue playing with it, but this should account for a vastly expanded SPA page count a year in.


jonallured (approved):
❤️

PR #8035, opened 2021-07-20
chore: silences "react does not recognize..." warnings

dzucconi:
https://styled-components.com/docs/api#shouldforwardprop

We just don't have updated typings yet.


damassi (approved):
Thanks for the quick fix 👍

PR #8036, opened 2021-07-21
fix(PURCHASE-2821): Forward search string in order "respond" step redirects

copperkraft:
Solves https://artsyproduct.atlassian.net/browse/PURCHASE-2821
Collector sends an offer in inquiry, gallery sends a counteroffer, the collector opens counteroffer respond modal and sees there "respond" step (without a header and a footer, as intended)
image
After accepting, declining, or sending a counteroffer the modal gets a header and a footer as on a full page. 
image
That happens because the "isModal" state is passing in the search string and moving to the next step is handled by the router without preserving the search string. 

This PR resolves this issue by passing the current search string to a new route on redirect. 
As a result: 
image


lilyfromseattle (approved):
Looks good! 🎉

PR #8042, opened 2021-07-21
fix: Added optional parameter to updateUrl method

anastasiapyzhik:
Initial sort value for fair exhibitors previously was set as `-decayed_merch` in order to initially hide this param in url. That caused a bug where fetching data with invalid sort value when paginating to a new page would fail.

This PR:
- adds optional parameter `defaultValues` to `updateUrl` method so that we can hide in url filter values other than initial
- fixes initial data fetching on fair booths page by changing initial sort value that is hidden in url with the help of ^ update

### DIFF
--- a/src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
@@ -2,8 +2,9 @@ import { ArtworkFilters } from "../ArtworkFilterContext"
 
 export const isDefaultFilter: (
   name: keyof ArtworkFilters,
-  value: any
-) => boolean = (name, value) => {
+  value: any,
+  defaultValues?: Partial<ArtworkFilters>
+) => boolean = (name, value, defaultValues = {}) => {
   if (!value) {
     return false
   }
@@ -20,6 +21,8 @@ export const isDefaultFilter: (
       name === "artistNationalities" ||
       name === "materialsTerms":
       return value.length === 0
+    case !!defaultValues[name]:
+      return value === defaultValues[name]
     case name === "sort":
       return value === "-decayed_merch"
     case name === "medium":
--- a/src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
@@ -9,6 +9,8 @@ export const isDefaultFilter: (
     return false
   }
 
+  const isCustomDefaultValuePassed = !!defaultValues[name]
+
   switch (true) {
     case name === "sizes" ||
       name === "artistIDs" ||
@@ -21,7 +23,7 @@ export const isDefaultFilter: (
       name === "artistNationalities" ||
       name === "materialsTerms":
       return value.length === 0
-    case !!defaultValues[name]:
+    case isCustomDefaultValuePassed:
       return value === defaultValues[name]
     case name === "sort":
       return value === "-decayed_merch"
--- a/src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
+++ b/src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
@@ -34,8 +34,10 @@ export const paramsToCamelCase = params => {
 
 export const buildUrl = (
   state: ArtworkFilters,
-  pathname?: string,
-  defaultValues?: Partial<ArtworkFilters>
+  {
+    pathname,
+    defaultValues,
+  }: { pathname?: string; defaultValues?: Partial<ArtworkFilters> }
 ): string => {
   const params = removeDefaultValues(state, defaultValues)
   const queryString = qs.stringify(paramsToSnakeCase(params))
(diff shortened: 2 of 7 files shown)


src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
```diff
@@ -46,18 +50,24 @@ export const buildUrl = (state: ArtworkFilters, pathname?: string): string => {
   return url
 }
 
-export const updateUrl = (state: ArtworkFilters) => {
-  const url = buildUrl(state)
+export const updateUrl = (
+  state: ArtworkFilters,
+  defaultValues?: Partial<ArtworkFilters>
+) => {
+  const url = buildUrl(state, "", defaultValues)
```
mdole, line 57: **minor:** we should be able to write this as `buildUrl(state, defaultValues: defaultValues)` to avoid passing the `""`

mdole (changes requested):
Nice work @anastasiapyzhik ! Honestly it took me a while to understand how this was working 😅 but now I get our routing a bit better!

Marking this as "request changes" because I think adding a test for the new `isDefaultFilter` functionality would be very helpful.

src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
```diff
@@ -20,6 +21,8 @@ export const isDefaultFilter: (
       name === "artistNationalities" ||
       name === "materialsTerms":
       return value.length === 0
+    case !!defaultValues[name]:
+      return value === defaultValues[name]
```
mdole, line 25: Oop, thought I had included this comment in my PR review. Let's add a test that incorporates `defaultValues` to [our tests](https://github.com/artsy/force/blob/a340dd41856f0ece085920053144836f50ad4116/src/v2/Components/ArtworkFilter/Utils/__tests__/isDefaultFilter.jest.tsx#L1) for this utility!

src/v2/Components/ArtworkFilter/Utils/isDefaultFilter.tsx
```diff
@@ -20,6 +21,8 @@ export const isDefaultFilter: (
       name === "artistNationalities" ||
       name === "materialsTerms":
       return value.length === 0
+    case !!defaultValues[name]:
```
damassi, line 24: This line is kinda hard to understand at a glance; can we assign it to a self-describing variable name and then use that here instead?

src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
```diff
@@ -32,8 +32,14 @@ export const paramsToCamelCase = params => {
...
 }
 
-export const buildUrl = (state: ArtworkFilters, pathname?: string): string => {
-  const params = removeDefaultValues(state)
+export const buildUrl = (
+  state: ArtworkFilters,
+  {
+    pathname,
+    defaultValues,
+  }: { pathname?: string; defaultValues?: Partial<ArtworkFilters> }
```
damassi, line 40: (non-blocking) - since the second arg is expanding into something more options centric it might be helpful to describe it like so:

```jsx
const buildUrl = (state: ArtworkState, options: BuildUrlOptions) => { ... }
```

mdole (commented):
Agreed with @damassi 's comments! I think those naming changes will make it easier for future developers to understand. Let's make those adjustments and then merge it :)

src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
```diff
@@ -32,10 +32,20 @@ export const paramsToCamelCase = params => {
   }, {})
 }
 
-export const buildUrl = (state: ArtworkFilters, pathname?: string): string => {
-  const params = removeDefaultValues(state)
+interface BuildUrlOptions {
+  pathname?: string
+  defaultValues?: Partial<ArtworkFilters>
+}
```
mdole, line 38: nice, this feels really clear and readable! 👏

mdole (approved):
Sweet! Looking good to me 😎

src/v2/Components/ArtworkFilter/Utils/urlBuilder.tsx
```diff
@@ -46,18 +56,24 @@ export const buildUrl = (state: ArtworkFilters, pathname?: string): string => {
...
   if (typeof window !== "undefined") {
     window.history.replaceState({}, "", url)
   }
 }
 
-export const removeDefaultValues = (state: ArtworkFilters) => {
+export const removeDefaultValues = (
+  state: ArtworkFilters,
+  defaultValues?: Partial<ArtworkFilters>
+) => {
```
damassi, line 73: So that the function signatures are all the same across the board, you can share the newly created interface in both `updateUrl` and `removeDefaultValues`: 

```
state: ArtworkFilters,
options: BuildUrlOptions
```

and now all of the code (even how one accesses `defaultValues` is consistent). Since the second `options` param is an object, it's fine that the latter two functions don't actually use `pathname`.

dblandin (approved):
Looks good to me! 👍🏼

PR #8043, opened 2021-07-21
feat(FX-3078): create overview tab

tr-ann:
Jira: [FX-3078](https://artsyproduct.atlassian.net/browse/FX-3078)

**Description**
Create a tab on the fair page that: 
- Is titled "Overview"
- Is the default landing spot for users who land on the fair page
- Remove the "more info" functionality so content is presented in full
- Is visually laid out as follows: https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=2891%3A0

**Changes**
- Moved all fair info to overview tab
- Updated analytics events for tabs changes
- Updated tests for fair tabs

**Before**

https://user-images.githubusercontent.com/56556580/126508668-3d6869b4-824d-499b-af4a-83aec21bb7a1.mov

**After**

https://user-images.githubusercontent.com/56556580/126761652-8c8a3be7-6e8b-47b3-8542-ab5541b63759.mov


src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -10,18 +10,51 @@ import {
...
+  label,
+  info,
+  isHTML = false,
+}) => (
+  <>
+    {label && <Text variant="md">{label}</Text>}
+    {isHTML
+      ? info && <HTML variant="md" html={info} />
+      : info && <TextWithNewlines variant="md">{info}</TextWithNewlines>}
+
```
dimatretyak, line 46: You can use such code, because [you check](https://github.com/artsy/force/blob/e7a7b3ff11b03d8c484771dcc3240b3fd8cb577e/src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx#L90) the `info` prop before rendering the `FairInfoSection` component

```javascript
{isHTML ? <HTML variant="md" html={info} /> : <TextWithNewlines variant="md">{info}</TextWithNewlines>}
```

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -65,20 +87,56 @@ const FairHeader: React.FC<FairHeaderProps> = ({ fair, ...rest }) => {
...
+          {links && <FairInfoSection label="Links" isHTML info={links} />}
+          {contact && <FairInfoSection label="Contact" isHTML info={contact} />}
         </Column>
       </GridColumns>
+
+      {hasArticles && (
+        <Box my={4} pt={4} borderTop="1px solid" borderColor="black10">
+          <Box display="flex" justifyContent="space-between">
+            {/* @ts-expect-error STRICT_NULL_CHECK */}
+            {fair.articlesConnection.totalCount > FAIR_EDITORIAL_AMOUNT && (
```
dimatretyak, line 108: Can you please fix the `STRICT_NULL_CHECK` error?

src/v2/Apps/Fair/Components/FairHeader/FairHeaderImage.tsx
```diff
@@ -10,9 +10,8 @@ interface FairHeaderImageProps {
 export const FairHeaderImage: React.FC<FairHeaderImageProps> = ({
   fair: { image },
 }) => {
-  if (!image) return null
+  if (!image?.url) return null
 
-  // @ts-expect-error STRICT_NULL_CHECK
   return <FullBleedHeader src={image.url} />
```
dimatretyak, line 15: I think that this option will be more readable

```javascript
if (image?.url) {
 return <FullBleedHeader src={image.url} />
}
return null
```

src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -24,104 +17,86 @@ import {
...
-  const clickedArtworksTabTrackingData: ClickedNavigationTab = {
-    action: ActionType.clickedNavigationTab,
-    context_module: ContextModule.exhibitorsTab,
-    context_page_owner_id: contextPageOwnerId,
-    context_page_owner_slug: contextPageOwnerSlug,
-    // @ts-expect-error STRICT_NULL_CHECK
-    context_page_owner_type: contextPageOwnerType,
-    destination_path: `${fair.href}/artworks`,
-    subject: "Artworks",
+  const [lastClickedTab, setLastClickedTab] = useState(ContextModule.fairInfo)
```
dimatretyak, line 33: The current code works and everything is ok. But it seems to me it would be preferable to keep the last taboo using `useRef`

```javascript
const lastClickedTab = useRef(ContextModule.fairInfo)
...
lastClickedTab.current = newValue
```

src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -24,104 +17,86 @@ import {
...
+      <FairHeaderImageFragmentContainer fair={fair} />
+
+      <Spacer my={[4, 30]} />
+
+      <RouteTabs my={[0, 2]} fill>
+        <RouteTab
+          to={fairHref}
+          exact
+          onClick={getTabTrackingData(
+            fair.href ?? "",
```
dimatretyak, line 73: I think `fair.href` should always be here

src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -24,104 +17,86 @@ import {
...
-    context_page_owner_slug: contextPageOwnerSlug,
-    // @ts-expect-error STRICT_NULL_CHECK
-    context_page_owner_type: contextPageOwnerType,
-    destination_path: `${fair.href}/artworks`,
-    subject: "Artworks",
+  const [lastClickedTab, setLastClickedTab] = useState(ContextModule.fairInfo)
+
+  const fairHref = fair.href ?? ""
+
+  const getTabTrackingData = (
```
dimatretyak, line 37: I think it's better to rename the variable (for example, to `trackTabData`)

dimatretyak (changes requested):

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -10,18 +10,53 @@ import {
...
+  label,
+  info,
+  isHTML = false,
+}) => (
+  <>
+    {label && <Text variant="md">{label}</Text>}
+    {isHTML ? (
+      <HTML variant="md" html={info} />
+    ) : (
+      <TextWithNewlines variant="md">{info}</TextWithNewlines>
```
dzucconi, line 46: What's an example of when this is desired?

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -10,18 +10,53 @@ import {
...
+    )}
+
+    <Spacer my={2} />
```
dzucconi, line 49: Would make more sense to use `<Join separator={<Spacer mt={4} />}>` in the outer context.

Also this is a bit of a foot gun with the `Spacer` component but using `my` allows for values that exist outside of the design system spacing scale. In this case it's fine since `4` exists inside of the v2 theme. It doesn't exist there in v3 though.

I'm going to make a ticket to fix this component by having a single space value prop and a single direction, rather than supporting all of `Box`

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -65,20 +89,63 @@ const FairHeader: React.FC<FairHeaderProps> = ({ fair, ...rest }) => {
...
+          {tickets && <FairInfoSection label="Tickets" isHTML info={tickets} />}
+          {ticketsLink && (
             <>
-              <Spacer mt={2} />
-
-              <RouterLink to={`/fair/${fair.slug}/info`}>
-                <Text variant="sm">More info</Text>
-              </RouterLink>
+              <a href={ticketsLink}>
+                <Text>Buy Tickets</Text>
```
dzucconi, line 103: `variant` is a required prop in v3 settings.

src/v2/Apps/Fair/Components/FairHeader/FairHeaderImage.tsx
dzucconi: Funnily enough I think the opposite is the case — makes more sense to early return nulls and the default case is the existy thing. (It doesn't really matter to me either way though)

src/v2/Apps/Fair/FairApp.tsx
dzucconi: Yeah — in general use state for things that require re-renders. Use refs for all other values.

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
tr-ann: @dzucconi  For example, `<TextWithNewLines />` used for display location which is a string with `\r\n` symbols. So if we use `<TextWithNewLines />` it will be rendered the string below:
image

In other way we receive one line string:

image

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
tr-ann: @dzucconi Looks good, thanks. I'll use `<Join />` in this case, but `mt={4}` is too large in this case. I use `mt={2}`

mdole (approved):
rockin, nice work @tr-ann! merging!

PR #8048, opened 2021-07-21
fix: Legacy Asset Loading

icirellik:
Creates a reusable component that loads both the legacy JS and CSS
assets for modern pages.


damassi (approved):

lilyfromseattle (approved):
🎉

PR #8050, opened 2021-07-21
GRO-465: Empty WorksByArtistYouFollow Rail in Curatorial Rails when following no artists

The-Beez-Kneez:
This is to add in a zero state for the curatorial rails.

Current Behavior:
Screen Shot 2021-07-21 at 1 48 18 PM

New Behavior:
Screen Shot 2021-07-21 at 1 34 05 PM

Mobile View of New Behavior: 
Screen Shot 2021-07-22 at 7 43 05 AM


src/v2/Apps/Auctions/Components/StandoutLots/StandoutLots.tsx
```diff
@@ -22,7 +22,20 @@ const StandoutLots: React.FC<StandoutLotsProps> = ({ viewer }) => {
...
 
   if (nodes.length === 0) {
-    return null
+    return (
+      <Box>
+        <Text
+          as="h3"
+          color="black60"
+          mb={12}
+          mt={6}
```
damassi, line 31: Do these margins work in the mobile view?

src/v2/Apps/Auctions/Components/WorksByArtistsYouFollowRail/WorksByArtistsYouFollowRail.tsx
```diff
@@ -23,7 +23,20 @@ const WorksByArtistsYouFollowRail: React.FC<WorksByArtistsYouFollowRailProps> =
...
+          as="h3"
+          color="black60"
+          mb={12}
+          mt={6}
+          textAlign="center"
+          variant="mediumText"
+        >
+          No Works To Show
+        </Text>
+      </Box>
```
damassi, line 38: Since this bit is duplicated a few times we could probably extract it into a `CuritorialRailZeroState` component and share it across all three spots

damassi (commented):
Looking good! Two small requests:
- We should add a zero state component to clean up some duplication 
- Can we add tests for the zero state in those three spots?

src/v2/Apps/Auctions/Components/WorksByArtistsYouFollowRail/WorksByArtistsYouFollowRail.tsx
The-Beez-Kneez: That seems to me like it would be a pretty small component, especially as the only text in it would be a 5 lines worth of code....

Is that worth it?

src/v2/Apps/Auctions/Components/StandoutLots/StandoutLots.tsx
The-Beez-Kneez: They do, updated with Mobile screen shot above.

The-Beez-Kneez:
@damassi  I updated the unit tests to account for Zero State. But I honestly do not see the worth in creating a zero state component for these (as they contain only 5 lines of code), other than to cut down on a light duplication. But I am leaning towards duplication as opposed to a component bloat.

If we were adding in complex logic or designs I might feel differently 🤷‍♀️

src/v2/Apps/Auctions/Components/WorksByArtistsYouFollowRail/WorksByArtistsYouFollowRail.tsx
damassi: Small components are fine; thinking: why make three changes to three places when you can make one change to one place and all three locations update as if by magic? 

More generally tho, I stick pretty religiously to the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) axiom. If there were just two spots i'd probably say meh, but since there are three it pings this rule and led me to this comment.

damassi:
> But I do not see the worth in creating a zero state component for these 

See [comment here](https://github.com/artsy/force/pull/8050#discussion_r674980193).

src/v2/Apps/Auctions/Components/WorksByArtistsYouFollowRail/WorksByArtistsYouFollowRail.tsx
The-Beez-Kneez: Added `CuratorialRailsZeroState`

damassi (approved):
Once things are 🍏 LGTM :shipit:

PR #8053, opened 2021-07-22
refactor(GRO-388): rebuilds view in room

dzucconi:
<del>Drafting this while I write some specs.</del>

The architecture here is: There's a component called `ViewInRoomRoom` which displays the floor/bench and will scale itself and everything in it to fit in the browser, the same way object-fit: contain would. Nested in that is both the actual artwork image and the 8ft scale. We know that the scale should be the width of the bench and comes out to eight feet, so we use that to convert the artworks from centimeters into pixels. With the hanging level, we hardcode the distance from the ceiling to where the floor starts and then use that same scale conversion to measure out the hanging heights from the floor.

-----

<del>I basically can't get opacity animations to run on this container in browsers other than Chrome, so I've temporarily disabled the transition and I can tackle that as it's own ticket. I think the browsers are just dropping all of the transition frames. Interestingly even just doing a simple fade in with react-transition-group doesn't animate in Safari/Firefox/Mobile Safari — I have a few theories but I'd rather not get too bogged down when I think this is shippable sans-transition.</del> It already fixes a number of bugs with some artworks being misshapen and corrects both the scaling and the hanging at eye level distance.

I also moved the close button to the right to match it up with the zoom view's close button.

Here's a comparison (new on left, existing on right):

nQ0CeMDD


dzucconi:
Update: a quick theory I had was that since the room actually extends extremely far outside the browser window that this was the cause of the poor performance. Wrapping it in an overflow: hidden div and we've got the opacity transition back. 🧠

src/v2/Components/ViewInRoom/ViewInRoomArtwork.tsx
```diff
@@ -0,0 +1,83 @@
...
+        srcSet={image.srcSet}
+        alt=""
+        mb={
+          // If too big to hang at eye-level
+          artwork.heightCm > eyeLevelSizeLimit
+            ? // Bottom edge rests at ground level
+              groundLevelPx
+            : // Centered on eye-level
+              eyeLevelPx - height / 2
+        }
```
damassi, line 43: I like the top-down readability form of this, but at some future point a dev would have to cut this out and paste it into a variable to `console.log` it during debug -- might as well just set that up now.

src/v2/Components/ViewInRoom/ViewInRoomCloseButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+)
+
+const Svg = styled.svg`
+  display: block;
+  width: 30px;
+  height: 30px;
+  stroke: ${themeGet("colors.black60")};
+  stroke-width: 3px;
+  overflow: visible;
+`
```
damassi, line 24: Do we not have anything in palette that would suffice? (I'm thinking the mobile nav X button)

src/v2/Components/ViewInRoom/util.ts
```diff
@@ -0,0 +1,13 @@
+import { EIGHT_FEET_CM, EIGHT_FEET_PX } from "./ViewInRoomScale"
+
+export type Range = { min: number; max: number }
+
+export const remap = (n: number, from: Range, to: Range) => {
+  return ((n - from.min) * (to.max - to.min)) / (from.max - from.min) + to.min
+}
```
damassi, line 7: Based on the shorthand name alone, not too sure what this does. `remapWhat`?

damassi (approved):
A couple #superminor non-blocking comments. 

This component is so easy to follow and read! Very nicely done 💯 

Feel free to merge once there are some specs covering the logic.

src/v2/Components/ViewInRoom/ViewInRoomTransition.tsx
```diff
@@ -0,0 +1,103 @@
...
+        src: imgFrom.src,
+        srcSet: imgFrom.srcset,
+        from: imgFrom.getBoundingClientRect(),
+        to: imgTo.getBoundingClientRect(),
+      })
+
+      await wait(100)
+      setTransitionStage("TRANSITION")
+
+      await wait(500)
```
damassi, line 47: Since the 500 and 100 are used in a few places it might be worth moving them to a const at the top of the file

dzucconi: I'm not sure how that'd improve the readability? It's pretty clear: wait for 100ms, wait for 500ms.

dzucconi: Unless you're talking about interpolating them into the transition durations!

dzucconi: I immediately wanted to re-write this with the [Animation API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API) but got bogged down in just making the transitions work... 🤔

src/v2/Components/ViewInRoom/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./ViewInRoom"
```
damassi, line 1: @dzucconi - As you're the only one that uses them across Force / Eigen, can you elaborate on your fondness for index files a bit? I know we've discussed this before, but we've been trying to factor this pattern out and were almost successful.

src/v2/Utils/remapValue.ts
```diff
@@ -0,0 +1,12 @@
+type Range = { min: number; max: number }
+
+/**
+ *
```
damassi, line 4: Can we add a description here for what this does / is intended to do?

src/v2/Components/ViewInRoom/index.ts
dzucconi: Makes it clear what can/should be imported from a subfolder, lets you avoid doing the goofy `import Foo from '../Foo/Foo'`

damassi: > import Foo from '../Foo/Foo'

This is very direct though, and how we've been doing it. If you're insistent on this pattern all good, but we had a whole discussion before you rejoined about working away from this bad habit (according to Node.js creator). It's just more files in the system.

dzucconi: Nah, it's fine with me.

PR #8057, opened 2021-07-22
chore: Add NETWORK_CACHE env variables

damassi:
Follow-up to https://github.com/artsy/force/pull/8033 

This makes it so that our network cache is configurable via ENV vars. 

#### Todo:
- [x] Update `NETWORK_CACHE_SIZE` on staging 
- [x] Update `NETWORK_CACHE_SIZE` on prod
- [x] Update `NETWORK_CACHE_TTL` on staging
- [x] Update `NETWORK_CACHE_TTL` on prod


src/v2/Apps/Artist/Routes/Consign/Components/__tests__/ArtistConsignMeta.jest.tsx
```diff
@@ -3,10 +3,6 @@ import { mount } from "enzyme"
 import React from "react"
 import { ArtistConsignMeta } from "../ArtistConsignMeta"
 
-jest.mock("v2/Utils/getENV", () => ({
-  getENV: () => "https://artsy.net",
-}))
```
damassi, line 8: this was blowing up our tests, since we're now grabbing isomorphic values from the relay network middleware

src/typings/sharify.d.ts
```diff
@@ -39,6 +39,8 @@ declare module "sharify" {
       IS_MOBILE: boolean
       readonly IP_ADDRESS: string
       readonly METAPHYSICS_ENDPOINT: string
+      readonly NETWORK_CACHE_SIZE: string
+      readonly NETWORK_CACHE_TTL: string
```
joeyAghion, line 43: Aren't these integers?

src/v2/System/Relay/createRelaySSREnvironment.ts
```diff
@@ -104,8 +98,8 @@ export function createRelaySSREnvironment(config: Config = {}) {
     }),
     relaySSRMiddleware.getMiddleware(),
     cacheMiddleware({
-      size: 2000, // max 2000 requests
-      ttl: 3600000, // 1 hour
+      size: Number(getENV("NETWORK_CACHE_SIZE")) ?? 2000, // max 2000 requests
+      ttl: Number(getENV("NETWORK_CACHE_TTL")) ?? 3600000, // 1 hour
```
joeyAghion, line 102: Can this use the defaults already implemented above rather than repeat them?

joeyAghion (commented):
Do these need to be repeated in `mobile/config.coffee`? I see that many similar settings are repeated there, but am never sure.

src/typings/sharify.d.ts
damassi: the issue is that all items that come through `process.env` are strings. JS is often smart enough to cast, but its probably best to be honest and to manually cast when needed.

src/v2/System/Relay/createRelaySSREnvironment.ts
damassi: Good q. Unfortunately no, not in this case. The defaults that are set through sharify apply to a running app, but the network layer is used extensively through our tests, which is independent of of sharify's bootstrapping. Hence the need to duplicate.

starsirius: Perhaps we can import the defaults directly from src/desktop/config.coffee? (Sorry if I missed something obvious.)

damassi: Good suggestion @starsirius, that should work 👍

damassi:
> Do these need to be repeated in mobile/config.coffee - nope, no relay in our mobile code

PR #8059, opened 2021-07-23
fix: [PURCHASE-2817] AdressForm inputs caret

ivan-badyulya:
JIRA -> [Unexpected behavior when entering text in the shipping form](https://artsyproduct.atlassian.net/browse/PURCHASE-2817)

The bug was about weird input caret behavior of AddressForm component. It jumps to the end every re-render, so when a user wants to add a text somewhere in a middle of text, then the caret jumps to the end after first element input.

The problem was in 2 sources of truth. We have AdrdessForm state, where we store the values of inputs, and also we store the same data higher in the parent. And in the value attributes of each input in AddressForm `props.data` was passed, not `state.data`, that’s why it cause such a strange behavior. So, the fix is easy, just to replace `this.props.value` with `this.state.value`.

GIF:
ezgif com-gif-maker (1)

### DIFF
--- a/src/v2/Apps/Order/Components/PaymentPicker.tsx
+++ b/src/v2/Apps/Order/Components/PaymentPicker.tsx
@@ -252,6 +252,7 @@ export class PaymentPicker extends React.Component<
       addressErrors,
       addressTouched,
       creditCardSelection,
+      hideBillingAddress,
     } = this.state
     const {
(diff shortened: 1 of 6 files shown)


jo-rs (approved):
That's great! Nice job for figuring out a simple solution to the issue 🙌

PR #8061, opened 2021-07-23
feature: In Memory Cache

icirellik:
A super basic in-memory full page cache for modern pages with static content.

(Dependent on final solution):
Currently caches html, utf-8 encoded, as a buffer. Long term there are more
optimized ways to cache this data. Most obviously caching the already compressed
response buffer would save CPU cycles and reduce memory load.

The goal is to leverage this in production to test the impact of full-page caching
and use the collected metrics to drive future page caching techniques. In review
apps, this eliminates up to 95%+ of the TTFB for cached pages depending on
network connection overhead: latency, bandwidth, etc.

Note: Disabled by default.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -32,6 +32,7 @@ import {
   FACEBOOK_ID,
   FACEBOOK_SECRET,
   IP_DENYLIST,
+  MEMORY_PAGE_URL_FILTER,
   NODE_ENV,
   SEGMENT_WRITE_KEY_SERVER,
   SENTRY_PRIVATE_DSN,
@@ -57,10 +58,11 @@ import { downcaseMiddleware } from "./lib/middleware/downcase"
 import { hardcodedRedirectsMiddleware } from "./lib/middleware/hardcodedRedirects"
 import { localsMiddleware } from "./lib/middleware/locals"
 import { marketingModalsMiddleware } from "./lib/middleware/marketingModals"
-import { pageCacheMiddleware } from "./lib/middleware/pageCache"
+import { pageCacheMiddleware } from "./lib/middleware/redisPageCache"
 import { sameOriginMiddleware } from "./lib/middleware/sameOrigin"
 import { unsupportedBrowserMiddleware } from "./lib/middleware/unsupportedBrowser"
 import { backboneSync } from "lib/backboneSync"
+import { memoryPageCacheMiddleware } from "lib/middleware/memoryPageCache"
 import { serverTimingHeaders } from "lib/middleware/serverTimingHeaders"
 
 // App-specific V2 server-side functionality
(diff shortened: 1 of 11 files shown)


src/lib/nodeCache.ts
```diff
@@ -0,0 +1,58 @@
...
+const debugLog = util.debuglog("cache")
+
+const DEFAULT_OPTIONS = {
+  checkperiod: 120,
+  maxKeys: 10000,
+  stdTTL: 10 * 60,
+  useClones: false,
+}
+
+export class MemoryCache {
```
damassi, line 14: If this class is called `MemoryCache` can we rename the file to `MemoryCache`? Or vice-versa.

damassi (commented):
@icirellik - Can you update the PR description with 
- exactly what this PR does? There's a lot going on.
- link to review app so that we can see this in action and get some numbers before merging

src/middleware.ts
```diff
@@ -66,8 +70,41 @@ const {
...
+          return true
+        }
+      }
+      return false
+    })
+  }
+} else {
+  cachableModernRoutes = []
+}
+
```
damassi, line 107: Can we move this chunk of code into its own file and import it. (No need to have include logic where the rest of the file is a series of middleware declarations.)

icirellik:
It doesn't supercede any of the previous work, however, it may be worth removing the previous cache as it's been disabled for a reasonable amount of time.

> link to review app so that we can click around and see this in action and get some numbers before merging

Since the review app doesn't remotely reflect real-world usage these numbers would be far too open to personal interpretation. In several tightly controlled experiments, the results indicate, as one would expect, that full-page caching works amazingly well for static content. 

Because there have been so many attempts to "do caching" in the past, we are planning to do some blue-green testing so that we can compare side-by-side system metrics before anyone jumps the gun and start making personal assumptions about how well it "may" perform.

damassi:
Oh I was just referring to the PR description comment

> In review apps, this eliminates 95%+ of the TTFB for cached pages.

And was wondering where this # came from and was hoping to test it out. Not a blocking comment or anything. But if we can update the main PR description with what this PR is doing exactly (is it caching html? is it caching the `res` object? is it some other cache? etc) that would be great.

icirellik:
> And was wondering where this # came

I'm in the process of generating images, I knew you'd ask this! :smile:

> There's a lot of subtle stuff

Node streams definitely aren't the most obvious things unless you've been deep in the weeds for sure!

mc-jones (approved):

mc-jones:
LGTM, pending the merge conflicts and agree with @damassi it would be good to see some additional description for quick reference on the implementation!

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 5ed22d92270c51f67acb388a762747fc0a92d179
</p>

PR #8062, opened 2021-07-23
chore: Decaffeinate Config Files

icirellik:
There has been a lot of confusion recently about the multiple config
files in force. This change decaffeinates the old configuration files
and merges them together into a single unified `config.ts` file.


src/middleware.ts
```diff
@@ -19,7 +19,17 @@ import sharify from "sharify"
...
+  APP_TIMEOUT,
+  IP_DENYLIST,
+  NODE_ENV,
+  CLIENT_ID,
+  CLIENT_SECRET,
+  API_URL,
+  SEGMENT_WRITE_KEY_SERVER,
+  SENTRY_PRIVATE_DSN,
+} from "./config"
+import * as config from "./config"
```
ovasdi, line 32: This is likely a micro optimization but perhaps move the second import up and then initialize vars from that single import?

ovasdi (approved):
This is awesome to see having worked through config files recently. Thank you for taking this on. 💯

artsyjian (approved):
lgtm

src/middleware.ts
icirellik: This is required so that we can pass the entire config to `@artsy/passport` as we were previously. In the future I would love to purposefully pass the values we need to `@artsy/passport`.

icirellik: https://github.com/artsy/force/pull/8062/files#diff-0626369278c89781505d2c04865ecce1302e979271bcfb471e029c19addc13a0R205

PR #8064, opened 2021-07-23
WIP - feature: Webpack Build Stats

icirellik:
Enables detailed Webpack build stats and sends them along to Datadog.

From time to time it comes up in conversation that the force build is slow
on X's computer, or that a change to the build configuration has impacted
developer experience. 

If we collect the build stats and send them to Datadog we will be able to
detect anomalies and find ways to remediate them.

An incomplete Dashboard: [link](https://app.datadoghq.com/dashboard/bt2-c8n-vpp?from_ts=1626993129102&to_ts=1627079529102&live=true)

TODO:

[ ] Add shared datadog secret
[ ] Track CI builds separately

### DIFF
--- a/package.json
+++ b/package.json
@@ -256,6 +256,7 @@
     "@babel/preset-react": "7.12.13",
     "@babel/preset-typescript": "7.13.0",
     "@babel/register": "7.13.8",
+    "@datadog/build-plugin": "^0.4.3",
     "@graphql-inspector/core": "1.14.0",
     "@loadable/babel-plugin": "5.13.2",
     "@loadable/webpack-plugin": "5.14.0",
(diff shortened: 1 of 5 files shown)


mc-jones (approved):

mc-jones:
@icirellik - This looks good and interested to see the outcome. Few questions on the use of the dashboard. Is this just something that we are going to actively monitor or just reference when there seem to be issues? Do we care about any of the other metrics: https://github.com/DataDog/build-plugin/tree/master/src/hooks/datadog#metrics.

icirellik:
@mc-jones I do think adding monitors to chunk size and overall build time is a must, they are good indicators that something unexpected was committed. Overall, I feel like the best use of this dashboard would be to track the developer experience for force and optimize areas that are causing issues.

I hope to add all the metrics to the dashboard! I also hope they release the dashboard config: https://github.com/DataDog/build-plugin/issues/26

PR #8066, opened 2021-07-26
fix:handle on setShipping error [PURCHASE-2819]

alexj105:
Just moved up handleError function to avoid updateAddress call, if setShipping mutation returns an error 
https://artsyproduct.atlassian.net/browse/PURCHASE-2819

https://user-images.githubusercontent.com/37010996/126953497-9772f235-2e44-40ee-8200-2bad5ffb4639.mov


jo-rs (approved):
Nice nice 💯

PR #8075, opened 2021-07-27
FX-2849 feat: Artwork grids page change event

anastasiapyzhik:
[FX-2849]

### Description
Currently when user changes page on an artwork grid on web, we fire a `Commercial_filter_params_changed` event. We should fire a different event because a page change is not a filter change.

#### This PR separates page change event

Screenshot 2021-07-27 at 14 39 00

[FX-2849]: https://artsyproduct.atlassian.net/browse/FX-2849

### DIFF
--- a/src/v2/Components/ArtworkFilter/index.tsx
+++ b/src/v2/Components/ArtworkFilter/index.tsx
@@ -173,8 +173,7 @@ export const BaseArtworkFilter: React.FC<
   useDeepCompareEffect(() => {
     Object.entries(filterContext.filters ?? {}).forEach(
       ([filterKey, currentFilter]) => {
-        // @ts-expect-error STRICT_NULL_CHECK
-        const previousFilter = previousFilters[filterKey]
+        const previousFilter = previousFilters?.[filterKey]
         const filtersHaveUpdated = !isEqual(currentFilter, previousFilter)
 
         if (filtersHaveUpdated) {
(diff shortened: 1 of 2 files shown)


src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -167,29 +171,37 @@ export const BaseArtworkFilter: React.FC<
...
-              current: JSON.stringify(filterContext.filters),
+          if (filterKey === "page") {
+            tracking.trackEvent({
+              action: ActionType.clickedChangePage,
+              context_module: ContextModule.artworkGrid,
+              context_page_owner_type: contextPageOwnerType,
+              context_page_owner_id: contextPageOwnerId,
+              context_page_owner_slug: contextPageOwnerSlug,
+              page_current: previousFilter,
+              page_changed: currentFilter,
```
mdole, line 190: **change requested:** let's import `ClickedChangePage` from Cohesion (the version should be updated automatically, but if you don't see the event, try updating it with `yarn`) and explicitly type this event. It could look like:
```ts
const PageTrackingParams: ClickedChangePage = {
action: ActionType.clickedChangePage,
context_module: ContextModule.artworkGrid,
context_page_owner_type: contextPageOwnerType,
context_page_owner_id: contextPageOwnerId,
context_page_owner_slug: contextPageOwnerSlug,
page_current: previousFilter,
page_changed: currentFilter
}
```

That way we're 100% sure we have all of the desired attributes for this event and they're typed correctly :)

mdole (changes requested):
Good job! One small adjustment and then we're good to go

src/v2/Components/ArtworkFilter/index.tsx
anastasiapyzhik: Sounds good!

mdole (approved):
sweet thanks for making that change!

PR #8076, opened 2021-07-27
chore: Simplify Yarn Install

icirellik:
Do not use brew to install yarn. There is no need as the preferred
method is to install it with `npm`, this also has the benefit of being
cross platform.


ovasdi (approved):
:shipit: 

Also made same change in MP: https://github.com/artsy/metaphysics/pull/3306

PR #8078, opened 2021-07-27
fix[FX-1758]: adds og meta tags to collection pages

mdole:
So turns out that in order to display a card for a tweet, Twitter demands that you add a `twitter:card` meta tag. We didn't add this tag on collection pages, so they looked terrible in tweets.

Tested using https://cards-dev.twitter.com/validator and [ngrok](https://ngrok.com/docs) for port forwarding.

One note: when I tested [this collection](https://www.artsy.net/collection/bridget-riley-black-and-white), I received the following warning:

```
WARN:  The image URL http://files.artsy.net/images/bridget-riley-black-white.png specified by the 'og:image' metatag may be restricted by the site's robots.txt file, which will prevent Twitter from fetching it.
```

So we _might_ still have problems if an image was uploaded using [Vanity](https://vanity.artsy.net/).

Before:
Screen Shot 2021-07-27 at 7 28 14 PM

After:
Screen Shot 2021-07-27 at 7 27 43 PM

Jira ticket: [FX-1758](https://artsyproduct.atlassian.net/browse/FX-1758)

### DIFF
--- a/src/v2/Apps/Collect/Routes/Collection/index.tsx
+++ b/src/v2/Apps/Collect/Routes/Collection/index.tsx
@@ -69,6 +69,10 @@ export const CollectionApp: React.FC<CollectionAppProps> = props => {
       <Meta property="og:url" content={collectionHref} />
       <Meta property="og:image" content={socialImage} />
       <Meta property="og:description" content={metadataDescription} />
+      <Meta property="og:title" content={title} />
+      <Meta property="og:type" content="website" />
+      <Meta property="og:site_name" content="Artsy" />
+      <Meta property="twitter:card" content="summary_large_image" />
       <Meta property="twitter:description" content={metadataDescription} />
       <Link rel="canonical" href={collectionHref} />
       <BreadCrumbList


src/v2/Apps/Collect/Routes/Collection/index.tsx
```diff
@@ -69,6 +69,10 @@ export const CollectionApp: React.FC<CollectionAppProps> = props => {
       <Meta property="og:url" content={collectionHref} />
       <Meta property="og:image" content={socialImage} />
       <Meta property="og:description" content={metadataDescription} />
+      <Meta property="og:title" content={title} />
+      <Meta property="og:type" content="website" />
+      <Meta property="og:site_name" content="Artsy" />
+      <Meta property="twitter:card" content="summary_large_image" />
```
damassi, line 75: @mdole - note that we now have a `MetaTags` component. That might be helpful here.

damassi: https://github.com/artsy/force/blob/7a216215d6c9d0fe3fee63dcc8f6a1fc376dcd19/src/v2/Apps/ArtistSeries/Components/ArtistSeriesMeta.tsx#L24-L28

mdole: oooo that's awesome, thanks for the pointer! Updated in https://github.com/artsy/force/pull/8084

PR #8079, opened 2021-07-27
fix(artwork): fixes preload srcset

dzucconi:
Embarrassed I didn't catch this sooner: When I switched this component to use a `srcSet` also required was updating the preload tag to support it. So on 2x screens we've been preloading the 1x then not using it.


jonallured (approved):
LGTM!! 🎉

PR #8090, opened 2021-07-28
chore(GRO-452): preloads unica italic

dzucconi:
Closes: [GRO-452](https://artsyproduct.atlassian.net/browse/GRO-452)

There is an ongoing design issue with the goal of removing italics completely, but for the time being this font is used prominently on every page, so it makes sense to preload it in the meantime.


icirellik:
Thanks for doing this! I've been meaning to test this out myself. One day we'll implement progressive font loading and be happy.

icirellik (approved):

PR #8092, opened 2021-07-28
feature: Webpack Stats from Environment

icirellik:
Add support for loading the webpack datadog configuration from the
shared environment file and log some basic machine stats.

### DIFF
--- a/webpack/utils/env.js
+++ b/webpack/utils/env.js
@@ -1,9 +1,43 @@
 // @ts-check
 
+import { createHash } from "crypto"
+import { readFileSync, accessSync, constants } from "fs"
+import { parse } from "dotenv"
 import chalk from "chalk"
 import path from "path"
 import yn from "yn"
 import os from "os"
+import v8 from "v8"
+
+function loadEnv(env) {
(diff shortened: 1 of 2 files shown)


webpack/utils/env.js
```diff
@@ -1,17 +1,51 @@
...
+  try {
+    accessSync(filepath, constants.F_OK)
+  } catch (e) {
+    return false
+  }
+  return true
+}
+
+// Load the default envs and apply to process.env
+applyToEnv({
```
ovasdi, line 34: This looks like an improved `src/loadenv.js` implementation! Wondering why you didn't refactor that and `import src/loadenv.js` here instead?

ovasdi (approved):
LGTM ❗

webpack/utils/env.js
icirellik: I debated heavily, this actually started as a refactor of `src/loadenv.js`. My preference is to keep our build source completely separate from our application logic lest it accidentally be included in a production build by webpack due to a rogue import.

It made me think twice about making a separate npm module for it 😄

ovasdi: Thanks for clarifying. Let the telemetry roll in 😃

PR #8093, opened 2021-07-28
feature: Preload Fonts ASAP

icirellik:
Being preloading the fonts before the JS assets. With the additional 
chunks created by the new chunking algorithm, fonts are being delayed
for longer than we'd like. Overall, this should help with our initial render.


dzucconi (approved):
Nice

PR #8094, opened 2021-07-28
fix(artistrelatedartists): uses cropped image versions; cleans up

dzucconi:
Just a minor thing I noticed when I was auditing the images. Also un-nests the nested anchor tags.

### DIFF
--- a/src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedArtistsRail.tsx
+++ b/src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedArtistsRail.tsx
@@ -1,5 +1,5 @@
 import { ContextModule, OwnerType } from "@artsy/cohesion"
-import { Box, EntityHeader, Image, Shelf, Spacer, Text } from "@artsy/palette"
+import { EntityHeader, Image, Shelf, Text } from "@artsy/palette"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "react-tracking"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedArtistsRail.tsx
```diff
@@ -45,67 +45,66 @@ const ArtistRelatedArtistsRail: React.FC<ArtistRelatedArtistsRailProps> = ({
...
-                  destinationPageOwnerId: node.internalID,
-                  destinationPageOwnerSlug: node.slug,
-                  destinationPageOwnerType: OwnerType.artwork,
-                  horizontalSlidePosition: index + 1,
-                  type: "thumbnail",
-                })
-              }}
-            >
-              <Box>
+            <React.Fragment key={index}>
```
damassi, line 48: why move key to a fragment instead of keeping it on RouterLink?

src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedArtistsRail.tsx
```diff
@@ -140,9 +139,8 @@ export const ArtistRelatedArtistsRailFragmentContainer = createFragmentContainer
                     node {
                       internalID
                       slug
-
                       image {
-                        resized(width: 325, height: 230) {
+                        cropped(width: 325, height: 230) {
```
damassi, line 143: 👍  -- this actually fixes something I missed in the banner; was using `resized` instead of `cropped`.

src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedArtistsRail.tsx
dzucconi: its the outer container now. with the entity header this had invalid dom nesting

PR #8097, opened 2021-07-29
feat(FX-3131): Support following the fair organizer

dimatretyak:
This PR resolves [FX-3131]

### Description
* Should reflect the follow state of the authenticated user
* If **unauthenticated**, follow button should prompt to sign up / login
* Should be able to use the existing `followProfile` GraphQL mutation in Metaphysics

### Resources
* [Figma](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3139%3A5588)

#### Unauthenticated user
https://user-images.githubusercontent.com/3513494/127527209-9e685218-78fa-483c-8a02-dad8cef22e3f.mp4

#### Authenticated user
https://user-images.githubusercontent.com/3513494/127527239-df44a6be-752a-4104-b56d-84456722d2de.mp4

#### Desktop
image

#### Mobile
image

[FX-3131]: https://artsyproduct.atlassian.net/browse/FX-3131


src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -17,5 +18,17 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
...
+        id: params.slug,
+      }
+    },
+    query: graphql`
+      query fairOrganizerRoutes_FairQuery($id: String!) {
+        fair(id: $id) {
+          ...FairOrganizerApp_fair
+        }
+      }
+    `,
```
damassi, line 32: Note that if these `$id`s were `$slug` then we wouldn't need to remap the variable in `prepareVariables` up above; it would naturally flow in from the `path: "/fair-organizer/:slug"` up above.

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -17,5 +18,17 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
...
       FairOrganizerApp.preload()
     },
+    prepareVariables: params => {
+      return {
+        id: params.slug,
+      }
+    },
+    query: graphql`
+      query fairOrganizerRoutes_FairQuery($id: String!) {
+        fair(id: $id) {
```
damassi, line 28: Note that if these `$id`s were `$slug` then we wouldn't need to remap the variable in `prepareVariables` up above; it would naturally flow in from the `path: "/fair-organizer/:slug"` up above.

damassi (approved):
Code looks great 👍 One tiny request about renaming `$id` to `$slug` and deleting a no-longer-needed `prepareVariables` function, but after that's done 🚢

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
dimatretyak: Oh, right. That's a very good point.

mdole (approved):
looks good! the tracking schema updates have now been release by Cohesion - a PR will automatically be created and merged in Force, or if you're feeling impatient @dimatretyak, you could manually update it with `yarn`. once we update the tracking, I think this PR is ready to go!

PR #8098, opened 2021-07-29
Revert "Remove unused: desktop/components/error_handler"

icirellik:
It turns out the tooling does not detect usage when `path.resolve` is used to generate a static file path. 😬 

https://github.com/artsy/force/blob/master/src/lib/middleware/errorHandler.ts#L18

This reverts commit 3a73ff9acce8639c8384ecf154ccf210da04cd08.


dzucconi:
🙏

PR #8099, opened 2021-07-29
fix(types): passes through system query renderer types

dzucconi:
I cherry-picked these off my working branch because it turned into a _thing_. Noticing props being cast as `any` led me to fix the typing in `SystemQueryRenderer` which pulled on a bit of a thread.

For the most part I'm just asserting the existence of a bunch of props, which _should_ be more or less OK, the nullable fields from Metaphysics tend to be incorrect for the most part. I had to use a `ts-ignore` or two in Onboarding where the types are pretty rough from a bunch of other ignores in the file.

### DIFF
--- a/src/v2/Apps/Partner/Components/Overview/ShowBannersRail/ShowBannersRail.tsx
+++ b/src/v2/Apps/Partner/Components/Overview/ShowBannersRail/ShowBannersRail.tsx
@@ -170,7 +170,10 @@ export const ShowBannersRailRenderer: React.FC<
 
         return (
           <PartnerShowBannersContextProvider>
-            <ShowBannersRailFragmentContainer {...rest} {...props} />
+            <ShowBannersRailFragmentContainer
+              {...rest}
+              partner={props.partner!}
+            />
           </PartnerShowBannersContextProvider>
         )
       }}
--- a/src/v2/System/Relay/SystemQueryRenderer.tsx
+++ b/src/v2/System/Relay/SystemQueryRenderer.tsx
@@ -1,34 +1,52 @@
 import React from "react"
-// FIXME: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37950
-// @ts-ignore
-import { QueryRenderer, QueryRendererProps } from "react-relay"
+import { useEffect } from "react"
+import { useState } from "react"
+import { QueryRenderer } from "react-relay"
 import { OperationType } from "relay-runtime"
 
-interface SystemQueryRendererState {
-  isMounted: boolean
+type QueryRendererProps = React.ComponentProps<typeof QueryRenderer>
+
+export type SystemQueryRendererProps<T extends OperationType> = Omit<
+  QueryRendererProps,
+  "render" | "environment" | "variables"
+> & {
+  placeholder?: React.ReactNode
+  environment?: QueryRendererProps["environment"]
+  variables?: QueryRendererProps["variables"]
+  render(renderProps: {
+    error: Error | null
+    props: T["response"] | null
+    retry: (() => void) | null
+  }): React.ReactNode
 }
 
-/** A QueryRenderer that runs only on the client */
-export class SystemQueryRenderer<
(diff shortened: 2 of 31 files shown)


src/v2/System/Relay/SystemQueryRenderer.tsx
```diff
@@ -1,34 +1,52 @@
...
+  "render" | "environment" | "variables"
+> & {
+  placeholder?: React.ReactNode
+  environment?: QueryRendererProps["environment"]
+  variables?: QueryRendererProps["variables"]
+  render(renderProps: {
+    error: Error | null
+    props: T["response"] | null
+    retry: (() => void) | null
+  }): React.ReactNode
```
dzucconi, line 20: This is a bit annoying — the only way I was able to figure out how to thread the response props from generated types through was by manually re-typing `render` — if anyone has suggestions on how to pass this into `QueryRendererProps` above, I'm all ears. I'm probably missing something obvious.

src/v2/Apps/Partner/Components/Overview/ShowBannersRail/ShowBannersRail.tsx
```diff
@@ -170,7 +170,10 @@ export const ShowBannersRailRenderer: React.FC<
 
         return (
           <PartnerShowBannersContextProvider>
-            <ShowBannersRailFragmentContainer {...rest} {...props} />
+            <ShowBannersRailFragmentContainer
+              {...rest}
+              partner={props.partner!}
```
damassi, line 175: Seeing you use `!` all over makes me feel a bit better about doing this with Relay components, but I still always wonder whyyy

src/v2/System/Relay/SystemQueryRenderer.tsx
```diff
@@ -1,34 +1,52 @@
...
-// @ts-ignore
-import { QueryRenderer, QueryRendererProps } from "react-relay"
+import { useEffect } from "react"
+import { useState } from "react"
+import { QueryRenderer } from "react-relay"
 import { OperationType } from "relay-runtime"
 
-interface SystemQueryRendererState {
-  isMounted: boolean
+type QueryRendererProps = React.ComponentProps<typeof QueryRenderer>
```
dzucconi, line 7: Like... this extracted type should be generic but I'm not sure how you'd be able to preserve it when extracting with `React.ComponentProps`

src/v2/System/Relay/SystemQueryRenderer.tsx
```diff
@@ -1,34 +1,52 @@
...
-    })
-  }
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
 
-  render() {
-    const { placeholder, ...rest } = this.props
+  if (!environment) {
+    return <></>
```
damassi, line 37: Been noticing this used quite a bit over `null` - why one over the other?

damassi (approved):
👍

src/v2/Apps/Partner/Components/Overview/ShowBannersRail/ShowBannersRail.tsx
dzucconi: I don't necessarily want to here... path of least resistance. Technically this _could_ be `null` if there's an error... and you'd want to probably guard on those props specifically. If you think putting in some `// @ts-expect-error STRICT_NULL_CHECK` is clearer to future maintainers, then that's an option too.

dzucconi: The "why" is mostly because of how infrequently we use non-nullable fields when defining schema in Metaphysics.

src/v2/System/Relay/SystemQueryRenderer.tsx
dzucconi: Ah, in this case it's expecting a return type of `JSX.Element`, which `null` is not. Same difference really. I could loosen up the return type as well.

PR #8102, opened 2021-07-30
feat:toast component for save address notifications [PURCHASE-2771]

alexj105:
[PURCHASE-2771]
Create a new component for saving addresses notifications.
When a user add, edit or delete address we see notifications below nav panel
swipe right to hide this in mob view
https://artsyproduct.atlassian.net/browse/PURCHASE-2771

https://user-images.githubusercontent.com/37010996/127626073-be8dbe3a-2378-4f2a-8e56-8abe108d55fa.mov
https://user-images.githubusercontent.com/37010996/127626104-e9d6adc4-2336-4353-bc70-3ad0d26c7394.mov

[PURCHASE-2771]: https://artsyproduct.atlassian.net/browse/PURCHASE-2771


damassi (commented):
@dzucconi - do you know if we already have something like this in Force?

src/v2/Components/Toast/ToastComponent.tsx
```diff
@@ -0,0 +1,88 @@
+import styled from "styled-components"
+import colors from "../../Assets/Colors"
+import { color, Flex, Text, themeProps } from "@artsy/palette"
+import React, { useEffect, useState } from "react"
+import { useNavBarHeight } from "../NavBar/useNavBarHeight"
+import { useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
+
+const ToastContainer = styled.div<{
```
damassi, line 8: We should avoid CSS and `divs` (etc) where possible and instead leverage Palette, which has a lot of these properties:
```
const ToastContainer = styled(Box)`
  ...
```

src/v2/Components/Toast/ToastComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+  box-shadow: 0px 0px 1px 2rgba (0, 0, 0, 0.08), 0px 2px 8px rgba(0, 0, 0, 0.12);
+  border-radius: 2px;
+  border-bottom: 1px solid ${color("black10")};
+  transition: all 0.6s;
+  opacity: 0;
+  position: fixed;
+  left: 0;
+  ${({ isOpen }) => isOpen && "opacity:1; transform: translate3d(0,0,0); "};
+  ${({ isMoveToRight }) => isMoveToRight && "transform: translate3d(100%,0,0);"}
+`
```
damassi, line 25: #superminor - line breaks between discreet units of code helps with readability

src/v2/Components/Toast/ToastComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+interface ToastComponentProps {
+  showNotification: boolean
+  notificationAction: string
+  duration: number
+  title: string
+  onCloseToast: () => void
+}
```
damassi, line 32: Same here (and elsewhere)

dzucconi:
@damassi sort of. There's https://palette-storybook.artsy.net/?path=/story/components-banner--default which this looks like this is trying to be. 

We need a proper toasts system in Palette — though this isn't really a toast and the system would need to be much more robust than this.

dzucconi:
For now, this should probably just use `Banner`. We should get a legitimate toast ticket on the DS board though.

src/v2/Assets/Colors.tsx
```diff
@@ -18,4 +18,5 @@ export default {
   yellowBold: "#f1af1b",
   greenRegular: "#0eda83",
   greenBold: "#0aa462",
+  greenToast: "#127438",
```
damassi, line 21: This file has been deprecated in favor of Palette, and we can pull values out via our `useTheme` hook:
```tsx
import { useTheme } from '@artsy/palette'

... 

const { theme } = useTheme()
console.log(theme.colors)
```

All available design-system colors can be seen [here](https://palette.artsy.net/tokens/color)

src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -151,60 +172,70 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
...
+            index={index}
+            address={address}
+            handleClickEdit={() => handleEditAddress(address, index)}
+          />
+          <Separator my={1} />
+          <ModifyAddressWrapper>
+            {!isDefaultAddress && (
+              <Box mr={[3, 1]}>
+                <Text
+                  onClick={() => handleSetDefaultAddress(address.internalID)}
```
damassi, line 200: All non button / input elements that have mouse events like this should be wrapped in [Clickable](https://palette.artsy.net/components/clickable).

dzucconi: Disappointed the a11y linter isn't picking this up

damassi: I wonder if its because the [`styled-components`](https://github.com/artsy/force/blob/master/.eslintrc.js#L16) scope is too narrow?

dzucconi: We are using this though https://github.com/brendanmorrell/eslint-plugin-styled-components-a11y

src/v2/Components/Toast/ToastComponent.tsx
alexj105: I can't use `styled(Box)` because, if I'm not mistaken `styled` receives only native html elements. anyway I got rid of `styled-component` by using Box with props

damassi: Cool 👍 

`styled(SomeComponent)` works with any styled component. Its like `class Foo extends Bar`

dzucconi: (You can use `styled` with anything that passes the `className` into the component)

src/v2/Components/Toast/ToastComponent.tsx
```diff
@@ -0,0 +1,74 @@
...
+    setTouchStart(e.targetTouches[0].clientX)
+  }
+
+  const handleTouchMove = e => {
+    setTouchEnd(e.targetTouches[0].clientX)
+  }
+
+  const handleTouchEnd = () => {
+    if (touchStart - touchEnd < -30) {
+      // do your stuff here for right swipe
```
xtina-starr, line 44: I think this is a stray comment that can be deleted.

src/v2/Components/UserSettings/__tests__/UserSettingsAddresses.jest.tsx
```diff
@@ -0,0 +1,92 @@
...
+      .find(`[data-test="deleteAddressInProfile"]`)
+      .first()
+    mockDeleteUserAddress.mockResolvedValue(mockDeleteAddressSuccessResponse)
+    deleteButton.simulate("click")
+    await wrapper.update()
+    const toast2 = wrapper.find(ToastComponent)
+    // I run here onShowToast func, but I can't get updating tree
+    //I tried to use setTimeout, but it doesn't work
+    setTimeout(() => {
+      console.log(toast2.text())
```
xtina-starr, line 49: Did you mean to leave the `console.log` & comments?

src/v2/Components/UserSettings/__tests__/UserSettingsAddresses.jest.tsx
```diff
@@ -0,0 +1,92 @@
...
+      expect(toast2.props().showNotification).toBe(true)
```
xtina-starr, line 50: I know we talked a bit about tests for these and I think seeing as the `showNotification` prop is what is used to determine whether toast is visible this covers the functionality of the toast notif.

It might be worth adding tests for edit and save.

xtina-starr (approved):
Left some comments regarding cleaning up a left over `console.log` and comments.

I also suggest adding toast tests for tests for the save and edit, but will let you use your discretion.

alexj105:
> I



> Left some comments regarding cleaning up a left over `console.log` and comments.
> 
> I also suggest adding toast tests for tests for the save and edit, but will let you use your discretion.

src/v2/Components/UserSettings/__tests__/UserSettingsAddresses.jest.tsx
alexj105: I would be glad to create tests for save, but we call createAdressMutation inside AddressModal by clicking save(onSubmit for formik). I open this modal by simulating  click on edit but I can't get any components inside AddressModal after that to trigger createAdressMutation. Of course, I can achieve it by trigger inside AddressModal wrapper, but by ToastComponent is outside this wrapper, namely 2 level above

PR #8103, opened 2021-07-30
fix: Hide current password input in settings form conditionally

olerichter00:
Addresses [CX-294]

## Description

Hides current password input in the settings page if the user account doesn't have a password. This case can happen if the user created the account with a 3rd party provider which doesn't require a password.

Currently, the user has to enter an arbitrary password in the "current password" field to pass the form validation when creating a password for an account without a password.

Screenshot 2021-08-09 at 10 27 36

[CX-294]: https://artsyproduct.atlassian.net/browse/CX-294


damassi:
Thanks for the fix 👍 (This form is so so old, need to eventually replace..)

PR #8105, opened 2021-07-30
Remove special instructions for local auth now that localhost should work

joeyAghion:
As of https://github.com/artsy/gravity/pull/14315, redirects from Gravity's staging environment to Force on `localhost` are permitted, so these special instructions should no longer be necessary.

We left [this special webpack path logic](https://github.com/artsy/force/blob/78f3308351952c9d5f8ea188ee5a9cbad9472913/src/lib/webpackPublicPath.ts#L17) in place, in case developers continue to use `local.artsy.net`.

Co-authored by: @icirellik

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,8 +33,6 @@
 # ------------------------
 # Localhost Environment
 # ------------------------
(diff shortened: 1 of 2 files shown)


icirellik (approved):
✨

ovasdi (approved):
great! confirmed that redirect is working for auth flow on `http://localhost:5000`. thanks!

PR #8106, opened 2021-07-30
chore: Clearer Reaction Usage

icirellik:
I was investigating leaking graphql query in reaction and started by
cleaning up our type imports.


mc-jones (approved):

mc-jones:
Dug into the [docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html) on this a bit, interesting stuff. LGTM.

icirellik:
It's not often needed, however, the `@artsy/reaction` library is close to being deprecated and I find that using `import type` is a helpful indicator to simplify the amount of effort required to remove an import.

PR #8115, opened 2021-08-02
feat(FX-3133): fair organizer past events

dimatretyak:
This PR resolves [FX-3133]

### Description
* Use the palette V3 shelf
* Cap number of fetched events to first 20 (should cover all fair organizers)

#### Web
Fair Organizer | Artsy 2021-08-03 12-24-57

#### Mobile
Fair Organizer | Artsy 2021-08-03 12-25-28

### Resources
* [Figma](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3139%3A5588)

[FX-3133]: https://artsyproduct.atlassian.net/browse/FX-3133


src/v2/Apps/FairOrginizer/Components/PastEventsRail.tsx
```diff
@@ -0,0 +1,49 @@
...
+  if (pastFairs.length === 0) {
+    return null
+  }
+
+  return (
+    <Box mt={2}>
+      <Text variant="lg" as="h3" mb={4}>
+        Past Events
+      </Text>
+      <Carousel>
```
damassi, line 25: @dzucconi - would you say that `Carousel` is deprecated and that we should be using `Shelf` everywhere now?

dzucconi: Yeah, the `Shelf` would be used here. `Carousel` is still useful but has special/particular use-cases now. Though in this case I wonder if this is really just a 4 column grid? There's no specification for the overflow at all and it looks like it adheres to the grid.

dzucconi: Looking at the mobile view leads me to believe this is a `Shelf` but there's no scrollbar and we lack the the ability to use grid within a `Shelf`. I left a comment on the Figma. Just using a `Shelf` with a 4:3 aspect ratio image is probably the correct course of action here.

dimatretyak: Got it. I will use `Shell` and the ratio is 325x244

dzucconi (approved):

PR #8126, opened 2021-08-03
fix(FX-2825): analytics event for clearing custom size

tr-ann:
### Description
- User selects a custom size filter (custom width/height)
- User then clicks “clear all” filters
- No analytics event is fired to clear width/height

### Changes

- Set initial values for custom width/height to artworks filter state

**Note**: this bug can be solved by 2 ways ([described here](https://artsy.slack.com/archives/C9SATFLUU/p1627657974044000))

### Demo
When user set custom size and then clear all filters the analytics event is fired

image


src/v2/Components/ArtworkFilter/__tests__/ArtworkFilter.jest.tsx
```diff
@@ -76,19 +77,8 @@ describe("ArtworkFilter", () => {
       const { current, changed } = trackEvent.mock.calls[0][0]
 
       expect(JSON.parse(current)).toMatchObject({
+        ...initialArtworkFilterState,
```
dblandin, line 80: Nice that we can reuse the initial artwork filter state here! 👍🏼

dblandin (approved):
Looks great!

PR #8127, opened 2021-08-03
fix(PX-4326): Don't manually override background in Pill component in SavedAddresses

Serge0n:
Jira Ticket: https://artsyproduct.atlassian.net/browse/PX-4326
Discussion: https://github.com/artsy/force/pull/7869#discussion_r669827365

This PR deletes `Pill` component usage with  `!important`  in styles and add a custom "Default" label instead of `Pill` in BNMO Checkout flow. Also, I deleted deprecated `color` and put `themeGet("colors.black10")` instead.

This PR resolves this [Jira Ticket](https://artsyproduct.atlassian.net/browse/PX-4367) too, so I closed it.

https://user-images.githubusercontent.com/55637696/129696004-8590e1b1-5a09-4900-9ded-008db9c92c84.mov

### DIFF
--- a/src/v2/Apps/Order/Components/SavedAddressItem.tsx
+++ b/src/v2/Apps/Order/Components/SavedAddressItem.tsx
@@ -1,4 +1,5 @@
-import { Flex, Text, RadioProps, Pill, color } from "@artsy/palette"
+import { Flex, Text, RadioProps, Pill } from "@artsy/palette"
+import { themeGet } from "@styled-system/theme-get"
 import React from "react"
 import styled from "styled-components"
 import { SavedAddresses_me } from "v2/__generated__/SavedAddresses_me.graphql"
@@ -105,6 +106,8 @@ const EditButton = styled(Text)`
 `
 
 const StyledPill = styled(Pill)`
-  background-color: ${color("black10")} !important;
-  pointer-events: none;
+  background-color: ${themeGet("colors.black10")};
+  &:hover {
+    background-color: ${themeGet("colors.black10")};
+  }
 `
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -105,6 +106,8 @@ const EditButton = styled(Text)`
 `
 
 const StyledPill = styled(Pill)`
-  background-color: ${color("black10")} !important;
-  pointer-events: none;
+  background-color: ${themeGet("colors.black10")};
+  &:hover {
+    background-color: ${themeGet("colors.black10")};
+  }
```
dzucconi, line 112: I'm just wondering why there's this one-off Pill variation in the first place? Is this something that should be elevated to a proper variant in Pill in the design system?

Serge0n: I think the better solution is to update the `Pill` component in `Palette`. In the near future, I will try to figure out how to do it.

Serge0n: I know for sure that something is wrong with the definition of the background of the component. This is complicated by the fact that the background changes in the parent component (`RadioGroup`) on the `hover`

dzucconi: Honestly: Design System atoms like this aren't meant to have custom styling, that's really it. There's nothing broken or wrong, exactly. It may just be the case that design has asked for a snowflake here and you don't actually want a `Pill`. The valid variants can be seen here: https://palette-storybook.artsy.net/?path=/story/components-pill--variants

damassi: I agree with @dzucconi - this is technically a fine solution, but ideologically we're constrained by what the design system supports. If a designer is trying to pass off a design with a color (etc) that isn't in Palette then the design system wins over the designer, fortunately. 

I would just use the regular pill and not worry about this, even if it feels a little weird. If it's a big deal we can have a formal discussion. In the meantime tho, we shouldn't be doing 1-off things.

PR #8133, opened 2021-08-04
FX-3134 Added latest articles block on fair organizer page

anastasiapyzhik:
[FX-3134]

<details>
<summary>desktop Figma spec</summary>

Screenshot 2021-08-04 at 15 16 35
</details>

<details>
<summary>mobile Figma spec</summary>

Screenshot 2021-08-04 at 15 16 51
</details>

[FX-3134]: https://artsyproduct.atlassian.net/browse/FX-3134


src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
```diff
@@ -39,27 +55,54 @@ export const FairEditorialItem: React.FC<FairEditorialItemProps> = ({
...
-          srcSet={article.thumbnailImage?.cropped?.srcSet}
-          width="100%"
-          height={240}
-          lazyLoad={true}
-          alt={article.thumbnailTitle!}
-        />
+    <RouterLink
+      to={article.href!}
+      aria-label={`${article.title} (${article.publishedAt})`}
+      style={{ textDecoration: "none", display: "block" }}
```
dimatretyak, line 86: Replace `style={{ textDecoration: "none", display: "block" }}` with `textDecoration="none" style={{ display: "block" }}`

src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
```diff
@@ -39,27 +55,54 @@ export const FairEditorialItem: React.FC<FairEditorialItemProps> = ({
...
+  }) => (
+    <Image
+      width={width}
+      height={height}
+      src={image.src}
+      srcSet={image.srcSet}
+      lazyLoad={true}
+      alt={article.thumbnailTitle!}
+    />
+  )
```
dimatretyak, line 80: Maybe it would be better to put this in a separate component and extract data using `createFragmentContainer`?

dimatretyak (changes requested):

dimatretyak (approved):

src/v2/Apps/FairOrginizer/helpers/getArticlesColumns.ts
```diff
@@ -0,0 +1,10 @@
+export function getArticlesColumns<T>(articles: T[]) {
+  const leftColumn: T[] = []
+  const rightColumn: T[] = []
+
+  articles.forEach((article: T, i) => {
+    ;(i % 2 === 0 ? leftColumn : rightColumn).push(article)
```
iskounen, line 6: **question** (non-blocking): Is the semicolon at the front of this line necessary? If so, I would like to learn more about it!

src/v2/Apps/FairOrginizer/Components/FairOrganizerLatestArticles.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export const FairOrganizerLatestArticles: React.FC<FairOrganizerLatestArticlesProps> = ({
+  fairOrganizer,
+}) => {
+  const { articles, name } = fairOrganizer
+
+  const nodes = extractNodes(articles)
+  const [latestArticle, ...otherArticles] = nodes
+  const { leftColumn, rightColumn } = getArticlesColumns(otherArticles)
+
```
iskounen, line 23: **suggestion** (non-blocking): As an alternative to creating the articles columns helper, would something like this work as well?

```
const leftColumn = otherArticles.filter((v,i) => i % 2 === 0)
const reftColumn = otherArticles.filter((v,i) => i % 2 !== 0)
```

src/v2/Apps/FairOrginizer/Components/__tests__/FairOrganizerLatestArticles.jest.tsx
```diff
@@ -0,0 +1,43 @@
...
+    query FairOrganizerLatestArticles_Test_Query {
+      fairOrganizer(id: "example") {
+        ...FairOrganizerLatestArticles_fairOrganizer
+      }
+    }
+  `,
+})
+
+describe("FairOrganizerLatestArticles", () => {
+  it("renders correctly", () => {
```
iskounen, line 23: **suggestion** (non-blocking): I think that breaking this test into smaller tests will help clarify what "correctly" means for rendering this component. For example:

- it renders a section title with the fair name
- it renders a read all button
- it renders the most recent article in a large card
- it renders the next 6 articles in small cards

src/v2/Apps/FairOrginizer/Components/FairOrganizerLatestArticles.tsx
```diff
@@ -0,0 +1,106 @@
...
+                  <Article article={article} size={"small"} />
+                </Box>
+              ))}
+            </Column>
+
+            {/* Read All Articles button */}
+            <Column span={[12, 8, 6]} start={[1, 5, 7]}>
+              <Spacer mt={30} />
+              <RouterLink
+                to={"/"}
```
iskounen, line 64: **question**: Is this URL meant to be `/articles`?

src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
```diff
@@ -39,27 +55,55 @@ export const FairEditorialItem: React.FC<FairEditorialItemProps> = ({
     action: ActionType.clickedArticleGroup,
   }
 
+  const image = article.thumbnailImage?.[size]!
+  const variants = {
+    large: {
+      primary: ["xl", "xxl"] as TextVariant[],
+    },
+    small: {
+      primary: ["lg", "lg", "xl"] as TextVariant[],
```
iskounen, line 64: question (non-blocking): I haven't used the `variant` property of a Text component before, so I am curious what this array means.

iskounen (approved):
LGTM. I just have some non-blocking questions.

src/v2/Apps/FairOrginizer/helpers/getArticlesColumns.ts
anastasiapyzhik: That was auto formatted but let's see what I've found: _It is acceptable to begin the statement with a `;` to make sure that the statement is interpreted as a new one and not a continuation of the previous statement._
So as I understand it was put here because the array for pushing into is dinamically defined and that's not a regular syntax

src/v2/Apps/FairOrginizer/Components/FairOrganizerLatestArticles.tsx
anastasiapyzhik: That would cause double array iteration but as long as division into columns is not used elsewhere I think we can do without this helper and your variant fits as well 🙂

src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
anastasiapyzhik: That might be used if you want to show different text sizes depending on screen width

src/v2/Apps/FairOrginizer/Components/__tests__/FairOrganizerLatestArticles.jest.tsx
anastasiapyzhik: Good point 👍  Thanks!

src/v2/Apps/FairOrginizer/helpers/getArticlesColumns.ts
iskounen: OK. It makes sense that it was auto-formatted. I still don't understand why the syntax to dynamically pick an array to push to needs this, but it seems harmless 🤷‍♂️ Thanks for the explanation!

src/v2/Apps/FairOrginizer/Components/FairOrganizerLatestArticles.tsx
iskounen: Understood. I wasn't thinking the performance of my approach 😅 There's no need to make any changes on this comment.

src/v2/Apps/FairOrginizer/helpers/getArticlesColumns.ts
anastasiapyzhik: I think it's helpful in case developer forgot to close the previous line with `;` so that the new line might be interpreted as a continuation of it. If the code looked something like 
```
const value = someFn
(i % 2 === 0 ? leftColumn : rightColumn).push(article)
```
JS would probably consider it as 
`const value = someFn(i % 2 === 0 ? leftColumn : rightColumn).push(article)`

dblandin (approved):
LGTM! Will assign back to you to merge as soon as the merge conflicts are resolved 👍🏼

src/v2/Apps/FairOrginizer/helpers/getArticlesColumns.ts
iskounen: That example is very clear 😅 Thank you!

PR #8136, opened 2021-08-04
fix: put back the 'log in to request condition report' cta

pepopowitz:
Integrity tests have been failing for several days, due to a change in #8099 that prevented render of the "Log in to request condition report" CTA when users weren't logged in. 

This PR walks back part of that change so that the `RequestConditionReport` will be rendered even when the `me` sub-query from relay is null, which is the case when a user is unauthenticated.

[Slack thread for context](https://artsy.slack.com/archives/CNRF89P7Y/p1628008959005500)

## Proof that the button is back

image


dzucconi (approved):
👍

damassi:
Thanks for the sleuthing on this @pepopowitz

PR #8138, opened 2021-08-04
fix(gro-454): reduce dom node count in global nav; converts to v3

dzucconi:
I've got a bunch of specs to fix before merging but here's the desktop nav. I'm going to follow this up with some clean up and refactoring of the mobile nav as a separate PR.

This gets the number of DOM nodes down from `477` to `188`. Our hands are a bit tied with keeping those large sub-menus in the DOM unfortunately.

This will depend on https://github.com/artsy/palette/pull/986 — which should be just about ready to go now.

Will convert off draft once I have specs in.

Updated on left, existing on right:

ycDHqLKs


src/v2/Components/NavBar/NavBarItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+    css`
+      color: ${themeGet("colors.blue100")};
+    `}
+
+  &:hover {
+    color: ${themeGet("colors.blue100")};
+  }
+`
+
+// NOTE: Found's `Link` doesn't forward its ref, which is necessary for the UI.
```
dzucconi, line 25: This is an annoying discovery. I'd PR a fix but we're already a few versions behind.

src/v2/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -0,0 +1,169 @@
...
+})`
+  border-radius: 50%;
+  width: 6px;
+  height: 6px;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  top: ${themeGet("space.2")};
+  right: 5px;
+`
```
dzucconi, line 169: This is also removed from immediately executing so logged in users should see one less graphql request and a few less image requests on page loads

src/v2/System/Router/RouterLink.tsx
```diff
@@ -24,40 +24,40 @@ export type RouterLinkProps = Omit<
...
-    const matcher = context?.router?.matcher
-    const isSupportedInRouter = useMemo(
-      () => !!matcher?.matchRoutes(routes, to),
-      [matcher, routes, to]
-    )
-
-    // TODO: Bulk replace
-    const deprecated = noUnderline ? { textDecoration: "none" } : {}
+export const RouterLink: React.ForwardRefExoticComponent<
+  RouterLinkProps & { ref?: React.Ref<HTMLAnchorElement> }
```
dzucconi, line 28: I guess we should actually remove this `forwardRef` — since it's unsupported in router aware links.

src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -143,172 +173,155 @@ export const NavBar: React.FC = track(
...
-                        }
-                        onClick={() => {
-                          setIsVisible(false)
-                        }}
-                      />
-                    )
-                  }}
+              <Flex alignItems="stretch">
+                <Dropdown
+                  keepInDOM
```
dzucconi, line 221: If we ever feel like removing ~150 DOM nodes from every page. 🤷

src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -143,172 +173,155 @@ export const NavBar: React.FC = track(
...
-                onClose={() => toggleMobileNav(false)}
-                isOpen={showMobileMenu}
-                onNavButtonClick={handleMobileNavClick}
-              />
-            </>
-          )}
-        </NavBarContainer>
-      </header>
-    </>
+      {/* TODO: V3 */}
```
dzucconi, line 315: This is actually fine/looks identical since v2 text variants still work in v3. But there is a bunch of updates that I'll make here in a separate PR.

damassi:
> Our hands are a bit tied with keeping those large sub-menus in the DOM unfortunately.

Does this mean that those parts of the nav can't take advantage of the new Palette work?

dzucconi:
Nah, I added a boolean flag to keep things in the DOM — so we use the palette dropdown here for everything but flip that flag for the two sub-menus

src/v2/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -0,0 +1,169 @@
...
+              })
+            }}
+            {...anchorProps}
+          >
+            <NavBarItemUnfocusableAnchor href="/works-for-you" />
+
+            <BellIcon
+              title="Notifications"
+              // @ts-ignore
+              fill="currentColor"
```
damassi, line 72: What is this `currentcolor`/

dzucconi: A [widely supported CSS value](https://caniuse.com/currentcolor) — so it'll inherit a `:hover` color from the parent, which means you don't have to do swap out the fill. IMO this should be the default fill value for all icons.

dzucconi:
OK, this should be in good shape now if you feel like giving it a once over @damassi — I mostly just fixed the specs and some types.

src/v2/Components/NavBar/NavBarLoggedInActions.tsx
damassi: Very cool and hadn't heard of this before

src/v2/Components/NavBar/Menus/NavBarNotifications.tsx
```diff
@@ -0,0 +1,213 @@
...
+        environment={relayEnvironment}
+        query={graphql`
+          query NavBarNotificationsQuery {
+            me {
+              unreadNotificationsCount
+              followsAndSaves {
+                notifications: bundledArtworksByArtistConnection(
+                  sort: PUBLISHED_AT_DESC
+                  first: 10
+                ) @connection(key: "WorksForYou_notifications") {
```
damassi, line 170: We might want to swap out this key

damassi (approved):
One comment about a seemingly-harmless `connection` key but otherwise 👍

PR #8148, opened 2021-08-05
fix(GRO-454): mobile global nav cleanup; convert to v3; reduce dom nodes; etc.

dzucconi:
This cleans up the nav and improves the touch feedback, improves the tappable areas, lines things up correctly, etc.

I think the menu DOM nodes go from like ~600 to ~250 (the whole tree is always present in the DOM once it's rendered). Though that shouldn't matter too much since it's modal. The way we're rendering them has some accessibility issues, but that's a whole separate thing.

Realizing I can clean up a few more things before merging so I'll hit these in the morning.

- [x] Replace responsive `display` in `NavBar` with Fresnel
- [x] Pull `AuthBanner` into `NavBar`
- [x] Change the loading bar color from purple to blue


src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
```diff
@@ -3,7 +3,7 @@ import styled from "styled-components"
 import { Flex, HTML, Join, Spacer, Text, color } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FeatureHeaderDefault_feature } from "v2/__generated__/FeatureHeaderDefault_feature.graphql"
-import { NAV_BAR_HEIGHT } from "v2/Components/NavBar"
+import { DESKTOP_NAV_BAR_HEIGHT } from "v2/Components/NavBar"
```
dzucconi, line 6: Is good to rename this since I think the `NAV_BAR_HEIGHT` was sometimes imported thinking it's only one height ever. I added a note in the constants file to just use the hook. Though we can eventually get rid of this entirely.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

> `src/v2/Components/NavBar/NavBarMobileMenu/NavBarMobileMenuAuthentication.tsx`
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 81f38bb0e3aa4abee7782c8108195205c5c17835
</p>

src/v2/Apps/ViewingRoom/Components/ViewingRoomHeader.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react"
 import { Box, Flex, Link, ResponsiveImage, Sans, color } from "@artsy/palette"
-import { NAV_BAR_HEIGHT, MOBILE_NAV_HEIGHT } from "v2/Components/NavBar"
+import { DESKTOP_NAV_BAR_HEIGHT, MOBILE_NAV_HEIGHT } from "v2/Components/NavBar"
```
damassi, line 3: Seriously 😓 that we need this all over. There surely has to be a better way

src/v2/System/Router/PageLoader.tsx
```diff
@@ -73,7 +73,7 @@ export class PageLoader extends React.Component<
       <Box width="100%" style={style} className={className}>
         <ProgressBar
           percentComplete={progress}
-          highlight="purple100"
+          highlight="blue100"
```
damassi, line 76: Should this be `brand`?

dzucconi: I think this is outside the theme provider — brand wasn't working. Just being blue is fine in this case.

dzucconi:
@damassi This should be ready to go unless you've got any other feedback.

PR #8150, opened 2021-08-06
feat: Add structured data for partners GRO-379

jonallured:
This PR adds structured data to the partner pages per the `LocalBusiness` schema.org spec:

https://schema.org/LocalBusiness

I poked around and discovered some things about the `address` and `location` attributes plus learned more about how we model partner locations so ended up pitching what you see here - we run both the `legalName` and `name` always and then conditionally run the `address` as a `PostalAddress` including phone number and conditionally run the `location` as a `Place` with lat/long in case Google will pick that up and do a map. More docs for the curious:

https://schema.org/Place
https://schema.org/PostalAddress

Note that I discovered there was no need to use the scary `dangerouslySetInnerHTML` api and instead I just ran the content right inside the `script` tag.

Looks like this:

Screen Shot 2021-08-06 at 10 55 31 AM

And here's the validator output:

Screen Shot 2021-08-06 at 11 01 30 AM

## Extract Stand-alone Structured Data Component

My first step here was to read the existing code for running structured data. I felt there was room for improvement there so I started this line of work by extracting a component called `StructuredData` that does the work of converting a JS object to stringified data and sets the correct context. It also encapsulates the funky `application/ld+json` content type and knows to run it as a `script` tag. I got some tests around this too.

## Migrate to Extracted Component

Once I had my extracted component I could then look at the places were I wanted to migrate away from the repetition. I started with the artist one and then did the artwork ones. Added tests here too. Easy peasy.

That left the breadcrumb one and that took a bit more doing because there was actual (untested?) behavior. So I took a few commits to extract and wrap with tests to ensure I could migrate without any regressions.

## Partner Page

At this point I was ready to add my new `LocalBusiness` component following these new patterns and then update the partner app to use it. One thing I struggled with a bit was making it so that `undefined` properties didn't come through and end up rendering to the page. I found that [Underscore `pick`](https://underscorejs.org/#pick) worked so I pulled that in - open to other ideas here!

## Testing This Change

All we can really do here is copy/paste into the validator until this lands in production. From there we can wait for Google to crawl the site and report any errors in Google Search Console which I plan to do in a week or so.

https://artsyproduct.atlassian.net/browse/GRO-379

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
+++ b/src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
@@ -4,6 +4,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { PartnerMeta_partner } from "v2/__generated__/PartnerMeta_partner.graphql"
 import { getENV } from "v2/Utils/getENV"
 import { useRouter } from "v2/System/Router/useRouter"
+import { LocalBusiness } from "v2/Components/Seo/LocalBusiness"
 
 interface PartnerMetaProps {
   partner: PartnerMeta_partner
@@ -11,9 +12,11 @@ interface PartnerMetaProps {
 
 const PartnerMeta: React.FC<PartnerMetaProps> = ({
   partner: {
-    slug,
+    locationsConnection,
     // @ts-expect-error STRICT_NULL_CHECK
     meta: { description, image, title },
+    name,
+    slug,
   },
 }) => {
   const {
@@ -27,6 +30,9 @@ const PartnerMeta: React.FC<PartnerMetaProps> = ({
   const canonicalHref = artistId
     ? `${getENV("APP_URL")}/partner/${slug}/artists/${artistId}`
     : `${getENV("APP_URL")}/partner/${slug}`
+
+  const locationEdges = locationsConnection?.edges || []
+  const partnerLocation = locationEdges[0]?.node
 
   return (
     <>
@@ -45,6 +51,7 @@ const PartnerMeta: React.FC<PartnerMetaProps> = ({
 
       {image && <Meta property="og:image" content={image} />}
       {image && <Meta name="thumbnail" content={image} />}
+      <LocalBusiness partnerLocation={partnerLocation} partnerName={name} />
     </>
   )
 }
@@ -54,12 +61,30 @@ export const PartnerMetaFragmentContainer = createFragmentContainer(
   {
     partner: graphql`
       fragment PartnerMeta_partner on Partner {
-        slug
+        locationsConnection(first: 1) {
+          edges {
+            node {
+              address
+              address2
+              city
+              coordinates {
+                lat
+                lng
+              }
+              country
+              phone
+              postalCode
+              state
+            }
+          }
+        }
         meta {
           image
           title
           description
         }
+        name
+        slug
       }
     `,
   }
--- a/src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
+++ b/src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
@@ -12,13 +12,7 @@ interface PartnerMetaProps {
 }
 
 const PartnerMeta: React.FC<PartnerMetaProps> = ({
-  partner: {
-    locationsConnection,
-    // @ts-expect-error STRICT_NULL_CHECK
-    meta: { description, image, title },
-    name,
-    slug,
-  },
+  partner: { locationsConnection, meta, name, slug },
 }) => {
   const {
     match: {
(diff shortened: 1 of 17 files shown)


src/v2/Components/Seo/LocalBusiness.tsx
```diff
@@ -0,0 +1,96 @@
+import React from "react"
+import { StructuredData } from "./StructuredData"
+import _ from "underscore"
```
dzucconi, line 3: use lodash instead (we need a linter for this!)

dzucconi: also I'd just pull out the relevant functions instead of importing _

src/v2/Components/Seo/LocalBusiness.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+interface PostalAddressSchemaData {
+  "@type": string
+  addressCountry: string
+  addressLocality: string
+  addressRegion: string
+  postalCode: string
+  streetAddress: string
+  telephone: string
+}
```
dzucconi, line 13: I know we can import a package of these types instead of manually doing them. :looking:

dzucconi: https://www.npmjs.com/package/schema-dts

src/v2/Components/Seo/LocalBusiness.tsx
jonallured: Damn I def searched the package.json file for `lowdash` and not finding it searched for `underscore` found it and picked that instead, haha. Will update!

jonallured: Also confused by this feedback @dzucconi 🤔 

Screen Shot 2021-08-06 at 11 26 18 AM

dzucconi: Size concerns aside: we don't use underscore at all in the new app so importing it a single time will add the entire library unnecessarily. Theoretically lodash can be tree-shaken (hence advice importing just the functions you need) though I'm unclear as to whether or not we're reaping the benefit of that.

dzucconi: `lowdash` !== `lodash`

src/v2/Components/Seo/LocalBusiness.tsx
dzucconi: I've used this before and it really takes the guess work out of it, and types don't add any weight to the bundle since they are discarded

jonallured: Nice one, will add! 🙌

src/v2/Components/Seo/LocalBusiness.tsx
jonallured: Fixed with d7b13a1 - is that what you were thinking?

dzucconi: Exactly 👍

src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
```diff
@@ -4,16 +4,19 @@ import { createFragmentContainer, graphql } from "react-relay"
...
 
 interface PartnerMetaProps {
   partner: PartnerMeta_partner
 }
 
 const PartnerMeta: React.FC<PartnerMetaProps> = ({
   partner: {
-    slug,
+    locationsConnection,
     // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 16: as a helpful nudge, can we remove these as we come across them? Mostly just adding safe access operators (or guards)

src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
```diff
@@ -28,6 +31,9 @@ const PartnerMeta: React.FC<PartnerMetaProps> = ({
     ? `${getENV("APP_URL")}/partner/${slug}/artists/${artistId}`
     : `${getENV("APP_URL")}/partner/${slug}`
 
+  const locationEdges = locationsConnection?.edges || []
```
damassi, line 34: can use the `extractNodes` helper

src/v2/Components/Seo/StructuredData.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const StructuredData = props => {
+  const schemaData = {
+    "@context": "http://schema.org",
+    ...props.schemaData,
+  }
+
+  const schemaContent = JSON.stringify(schemaData, null, 2)
+
+  return (
+    <Meta tag="script" type="application/ld+json">
```
damassi, line 13: nice

damassi (approved):
Code and tests look great and I'll leave it to you to merge and follow-up with @dzucconi's lib suggestion, or hold this until that's in 👍

src/v2/Components/Seo/LocalBusiness.tsx
jonallured: More chatter here but I had trouble adding this package and then using it for the `PostalAddress` data:

https://artsy.slack.com/archives/CP9P4KR35/p1628277103099800

Going to bail on this and can return to this topic later if it's warranted - hope that's ok!

src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
jonallured: Took a shot at this with 620e78d - is that what you had in mind?

src/v2/Apps/Partner/Components/PartnerMeta/index.tsx
jonallured: Did this with fbc5cc4.

jonallured:
Going to merge this, thanks for the feedback all!! 🤗

src/v2/Components/Seo/LocalBusiness.tsx
dzucconi: Yah is fine. FWIW looking at this now — you wouldn't use `PostalAddress` directly but rather the entire `LocalBusiness` type. If we ever want to do JSON+LD in the future we should definitely just use the provided schema. I suspect that these will be invalid when there's some nulls around, etc.

PR #8152, opened 2021-08-06
Migrate payments app to palette v3

lilyfromseattle:
[PURCHASE-2878]

__Before:__

Screen Shot 2021-08-06 at 2 44 18 PM

Screen Shot 2021-08-05 at 4 50 40 PM

__After:__

Screen Shot 2021-08-06 at 2 43 51 PM

Screen Shot 2021-08-05 at 5 47 40 PM

__V2 styles preserved for other uses of the shared credit card field:__

Screen Shot 2021-08-05 at 6 14 22 PM

[PURCHASE-2878]: https://artsyproduct.atlassian.net/browse/PURCHASE-2878

### DIFF
--- a/src/v2/Apps/Order/Components/CreditCardInput.tsx
+++ b/src/v2/Apps/Order/Components/CreditCardInput.tsx
@@ -1,25 +1,52 @@
-import { BorderBox, Sans, color, themeProps } from "@artsy/palette"
-import { fontFamily } from "@artsy/palette/dist/platform/fonts"
 import {
-  BorderProps as InputBorderProps,
-  borderMixin as inputBorder,
-} from "v2/Components/Mixins"
+  BorderBox,
+  themeProps,
+  useThemeConfig,
+  Text,
+  getThemeConfig,
+  TextVariant,
+} from "@artsy/palette"
+import { borderMixin, v3BorderMixin } from "v2/Components/Mixins"
 import React from "react"
 import type {
   StripeCardNumberElementChangeEvent,
   StripeError,
 } from "@stripe/stripe-js"
 import { CardElement } from "@stripe/react-stripe-js"
-import styled from "styled-components"
+import styled, { css } from "styled-components"
+import { useState } from "react"
+import { fontFamily } from "@artsy/palette/dist/platform/fonts"
 
-export const StyledCardElement = styled(CardElement)`
+export const StyledCardElement = styled(CardElement)<{ isV3?: boolean }>`
+  ${props => {
+    const states = getThemeConfig(props, {
+      v2: { padding: "9px 10px" },
+      v3: { padding: "12px 10px" },
+    })
+
+    return css`
+      padding: ${states.padding};
+    `
+  }}
   width: 100%;
-  padding: 9px 10px;
 `
 
 // Re-uses old input border behavior
-const StyledBorderBox = styled(BorderBox)<InputBorderProps>`
-  ${inputBorder};
+export interface BorderProps {
+  hasError?: boolean
+}
+
+const StyledBorderBox = styled(BorderBox)<BorderProps>`
+  ${props => {
+    const states = getThemeConfig(props, {
+      v2: { marginBottom: "0px", mixin: borderMixin },
+      v3: { marginBottom: "20px", mixin: v3BorderMixin },
+    })
+
+    return css`
+      ${states.mixin}
+    `
+  }}
   padding: 0;
   height: 40px;
 `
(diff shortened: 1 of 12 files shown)


src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -1,10 +1,36 @@
...
+`
 
 export const AddressModalFields: React.FC = () => {
+  const elements = useThemeConfig({
+    v2: {
+      fieldLabel: <Text mb={0.5}>Country</Text>,
+    },
+    v3: {
+      fieldLabel: (
+        <AllCapsText fontSize="13px" mb={0.5}>
```
damassi, line 27: Why not use `<Text>`?

src/v2/Components/Address/AddressModalFields.tsx
```diff
@@ -42,7 +68,7 @@ export const AddressModalFields: React.FC = () => {
             />
           </Box>
         </Column>
-        <Column span={6}>
+        <Column span={12}>
```
damassi, line 71: I _think_ by default `<Column>` spans 12, so no need for prop

src/v2/Components/Mixins.tsx
```diff
@@ -24,6 +25,24 @@ export const borderMixin = css`
     border: 2px dotted ${colors.grayRegular};
   }
 `
+export const v3BorderMixin = css`
+  border: 1px solid
+    ${({ hasError }: BorderProps) =>
+      hasError ? color("red100") : color("black30")};
```
damassi, line 31: we'll want to use `themeGet("colors.someColor")` as the `color` helper has been deprecated (it can't infer the correct theme, where `themeGet` can)

src/v2/Components/Payment/PaymentModal.tsx
```diff
@@ -176,17 +176,14 @@ export const PaymentModal: React.FC<PaymentModalProps> = props => {
             <Text data-test="credit-card-error" color="red" my={2}>
               {createError}
             </Text>
-            <Text color="black60" mb={1}>
-              All fields marked * are mandatory
-            </Text>
-            <Text>Credit Card *</Text>
+            <Text fontSize="13px">CREDIT CARD</Text>
```
damassi, line 179: We should never step outside of our design system variants

src/v2/Components/Payment/SavedCreditCards.tsx
```diff
@@ -37,11 +37,11 @@ export class CreditCard extends React.Component<
   render() {
     return (
       <>
-        <BorderBox flexDirection="column" p={2} mb={2}>
+        <BorderBox flexDirection="column" p={2} mb={4}>
           <Flex justifyContent="space-between" alignItems="center">
             {/* @ts-expect-error STRICT_NULL_CHECK */}
             <CreditCardDetails {...this.props.creditCard} />
-            <Sans size="2" color="purple100">
+            <Text size="2" color="red100">
```
damassi, line 44: We should be using `variants` (no more `size`)

damassi:
(Added @dzucconi as a reviewer as there's some subtle stuff that he might have an opinion on)

src/v2/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -1,90 +1,103 @@
...
+export interface BorderProps {
+  hasError?: boolean
+  isV3?: boolean
+}
+
+const StyledBorderBox = styled(BorderBox)<BorderProps>`
   padding: 0;
   height: 40px;
+  ${props => (props.isV3 ? v3BorderMixin : borderMixin)}
+  ${props => props.isV3 && "margin-bottom: 20px"}
```
dzucconi, line 28: You should use [`getThemeConfig`](https://github.com/artsy/palette/blob/5a263d623b5eb27f6427489db05101af2285cadd/packages/palette/src/Theme.tsx#L72) instead here.

src/v2/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -1,90 +1,103 @@
...
+          isV3
+          options={{
+            hidePostalCode: true,
+            style: {
+              base: {
+                "::placeholder": { color: "black60" },
+                fontSize: styles.fontSize,
+                fontFamily: styles.fontFamily,
+                fontSmoothing: "antialiased",
+                lineHeight: styles.lineHeight,
```
dzucconi, line 87: Ugh, I forgot how annoying it was to style these.

src/v2/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -1,90 +1,103 @@
...
+      lineHeight: "20px",
+    },
+    v3: {
+      fontSize: "16px",
+      borderBox: "title",
+      fieldHeight: "50px !important",
+      fontColor: "black60",
+      isV3: true,
+      fontFamily: "inherit",
+      lineHeight: "24px",
```
dzucconi, line 56: Rather than hardcode these you could import the static themes for each version: https://github.com/artsy/palette/blob/master/packages/palette/src/Theme.tsx#L9 then pull out the tokens.

src/v2/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -1,90 +1,103 @@
...
-  }
+            },
+          }}
+          onChange={onChange.bind(this)}
+          onFocus={() => setFocused(true)}
+          onBlur={() => setFocused(false)}
+        />
+      </StyledBorderBox>
+      {message && (
+        <Text pt={1} size="2" color="red100">
```
dzucconi, line 97: Remove size and add a variant.

src/v2/Components/Address/AddressModalFields.tsx
dzucconi: That & use a variant. Never set a fontSize directly.

src/v2/Components/Address/AddressModalFields.tsx
dzucconi: Column spans 1 by default. I'm not sure grid columns are even needed here anymore.

src/v2/Components/Mixins.tsx
dzucconi: I guess this is to fake the input styling for the credit card input? These will diverge from the actual input: https://github.com/artsy/palette/blob/master/packages/palette/src/elements/Input/tokens/v3.ts — makes me wonder if we should expose a mixin?

src/v2/Components/Payment/SavedCreditCards.tsx
dzucconi: I'm gonna make an issue to just unset this prop since it probably causes more confusion than it's worth (it's a shorthand for setting width/height so it's technically valid)

src/v2/Components/Payment/SavedCreditCards.tsx
```diff
@@ -166,7 +166,7 @@ export class CreditCard extends React.Component<
 
 export const RemoveLink = styled.div`
   text-align: right;
-
+  text-decoration: underline;
```
dzucconi, line 169: This whole thing should be a `Clickable` instead. We shouldn't put onClick handlers on divs

src/v2/Components/Address/AddressModalFields.tsx
lilyfromseattle: 👍 Didn't realize we had a variant for 13px

src/v2/Apps/Order/Components/CreditCardInput.tsx
lilyfromseattle: Is there an example of how to use `getThemeConfig`? I don't see it being used anywhere in force.

dzucconi: It's used [frequently in Palette](https://github.com/artsy/palette/search?q=getthemeconfig) — I guess we haven't had much of a need for it in Force. But it's basically the exact same thing as `useThemeConfig` but with a slightly different call signature: the first argument is the props, second argument is the v2/v3 configuration.

damassi (approved):
Once [this comment](https://github.com/artsy/force/pull/8152/files#r684442943) is addressed this looks good to me 👍 (Though others might have more feedback.)

damassi:
@lilyfromseattle - if you rebase from master that acceptance test failure should be fixed.

Also, since @xtina-starr is leaving, we should reassign this PR so that we can get it merged and out.

xtina-starr (approved):

PR #8155, opened 2021-08-09
feat(FX-3149): move fair name and logo to page header

tr-ann:
Jira: [FX-3149](https://artsyproduct.atlassian.net/browse/FX-3149)

### Description

Users often find themselves on the Artworks or Booths tab when looking at a fair (either via organic browsing or email redirect). When on those tabs, they do not see the fair's name and so lose out on the context of what they are looking at. We would like to always have the user clearly oriented so they know which fair the works or exhibitors they are looking at pertain to.
So **purpose** of the task is move fair logo and name to page header.

### Changes
- Move fair name/logo to header
- Refactor fair components (fair overview and fair header)
- Add some tests for fair timer

### Demo

_Desktop_

https://user-images.githubusercontent.com/56556580/128702335-55c58952-7438-4b10-beed-192b0bdde8f6.mov

_Mobile_

https://user-images.githubusercontent.com/56556580/128702123-b2da1eb8-20bd-4af4-b50e-731fee63f07a.mov


iskounen (approved):
LGTM 👍

src/v2/Apps/Fair/Components/FairHeader/FairHeader.tsx
```diff
@@ -1,170 +1,55 @@
...
-          </Flex>
-        </Column>
-
-        <Column span={6}>
-          <Join separator={<Spacer mt={2} />}>
-            <Text variant="md" textTransform="uppercase">
-              About
+    <>
+      {/* Desktop Fair Header */}
+      <Media greaterThan="xs">
```
dblandin, line 19: **suggestion**: We usually use this `Media` component when the layout between sizes differ greatly, but in this case, I think that we should be able to get by using [responsive props](https://palette.artsy.net/guides/responsive/) so that we're not rendering the same content twice:

Would you mind taking a moment to see if we can consolidate this JSX and move from `Media` to responsive props where possible?

```jsx
<Flex my={[2]}>
  <FairHeaderIconFragmentContainer fair={fair} mr={2} />
  <Box display={["flex", "block"] flexDirection="column" justifyContent="center" >
    <Text as="h1" variant={["lg", "xl"]}>
      {name}
    </Text>
    <FairTiming fair={fair} />
  </Box>
</Flex>
```

src/v2/Apps/Fair/Components/FairHeader/FairTiming.tsx
```diff
@@ -12,25 +13,31 @@ interface Props {
...
+      <EventTiming currentTime={currentTime} startAt={startAt} endAt={endAt} />
+    )
+
   return (
     <>
-      <Text variant="xl" color="black60" mb={1}>
-        {exhibitionPeriod}
-      </Text>
+      {/* Desktop Fair Timing */}
+      <Media greaterThan="xs">
```
dblandin, line 27: **suggestion**: Similar comment to the one above to explore using [responsive props](https://palette.artsy.net/guides/responsive/) instead of the `Media` component.

src/v2/Apps/Fair/Components/FairOverview/FairTimer.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export const FairTimer: React.FC<FairTimerProps> = ({ fair: { endAt } }) => {
+  if (!endAt) {
+    return null
+  }
+
+  return (
+    <Column span={6}>
+      {/* Desktop Fair Timer */}
+      <Media greaterThan="xs">
```
dblandin, line 20: **suggestion**: Same here re: [responsive props](https://palette.artsy.net/guides/responsive/)

dblandin (commented):
This looks great!

I left a comment suggesting that we try to use responsive props instead of the `Media` component where possible.

tr-ann:
@dblandin thanks for the comments, it's very useful! I replaced `<Media />` with responsive props, so can you review it one more time? 🙂

dblandin (approved):
Nice! 💯  Looks great 👍🏼

PR #8156, opened 2021-08-09
feature: Authentication Local - Production Mode

icirellik:
This change enables authentication when running locally in production
mode.


src/lib/middleware/session.ts
```diff
@@ -15,7 +16,9 @@ export function sessionMiddleware(): RequestHandler {
     domain: process.env.NODE_ENV === "development" ? "" : COOKIE_DOMAIN,
     name: SESSION_COOKIE_KEY,
     maxAge: SESSION_COOKIE_MAX_AGE,
-    secure: process.env.NODE_ENV === "production" || NODE_ENV === "staging",
+    secure:
+      (process.env.NODE_ENV === "production" || NODE_ENV === "staging") &&
```
joeyAghion, line 20: It's weird that we're using both `process.env.NODE_ENV` and `NODE_ENV` here. Should they both be `NODE_ENV` which supports a default?

icirellik: Yes...I'm not even sure we want the option for "staging" anymore.

damassi: I recall us getting rid of `staging` related stuff a few years back; is it still in use somewhere?

damassi: Ah, I remember now -- it was removed, then re-added when we wanted to add a banner indicating that we were _on_ staging, so as to not confuse (weird req from random users). That banner is only on old pages and can be killed (we didn't bother migrating to new). So yeah, AFAIK we can remove this env.

ovasdi (approved):
LGTM ❗
@icirellik did you want to adjust anything else based on @damassi [last comment](https://github.com/artsy/force/pull/8156#discussion_r685395353) or can we merge this?

PR #8160, opened 2021-08-09
chore: Remove feature flag checkers for web inquiry checkout

jo-rs:
Removes feature flag checkers for Web Inquiry Checkout functionality.
cc: @artsy/purchase-devs

### DIFF
--- a/src/v2/Apps/Conversation/Components/Conversation.tsx
+++ b/src/v2/Apps/Conversation/Components/Conversation.tsx
@@ -15,7 +15,6 @@ import { UpdateConversation } from "../Mutation/UpdateConversationMutation"
 import styled from "styled-components"
 import { ConversationHeader } from "./ConversationHeader"
 import { ConfirmArtworkModalQueryRenderer } from "./ConfirmArtworkModal"
-import { useSystemContext } from "v2/System/SystemContext"
 import { BuyerGuaranteeMessage } from "./BuyerGuaranteeMessage"
 import { extractNodes } from "v2/Utils/extractNodes"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -51,10 +50,7 @@ const Conversation: React.FC<ConversationProps> = props => {
 
   const isArtworkOfferable = !!artwork && !!artwork?.isOfferableFromInquiry
 
-  const isOfferable =
-    user &&
-    userHasLabFeature(user, "Web Inquiry Checkout") &&
-    isArtworkOfferable
+  const isOfferable = user && isArtworkOfferable
```
xtina-starr, line 53: We only needed the user here to satisfy a type error that was being thrown we used `userHasLabFeature`. Since we are removing that I think we can refactor this to the following:

```ts
const isOfferable = !!artwork && !!artwork?.isOfferableFromInquiry
```

xtina-starr: It also might be a good idea (and fun) to clean up the lab feature in gravity. Let me know if you want to pair on that.

xtina-starr (changes requested):
I had a small refactor suggestion.

src/v2/Apps/Conversation/Components/Conversation.tsx
jo-rs: Gotcha. Comment addressed.
Yeah, just to make sure we get the feature out and that this gets deployed first I wanted to push this through, and then deal with the gravity cleanup 😊

PR #8007, opened 2021-07-19
moved timestamp to above order update and fixed mo button styling

lilyfromseattle:
# [PURCHASE-2816] 
Incorrect timestamp location when submitting an offer in inquiry

Expected behavior:
The inline offer status is preceded by the time stamp

Actual behavior:
The inline offer status is followed by the time stamp

__Before:__
Screen Shot 2021-07-19 at 2 32 43 PM

__After:__
Screen Shot 2021-07-19 at 2 32 35 PM

# [PURCHASE-2825]
Extra border shadow is still there even when MO button isn't visible

__Before:__
Screen Shot 2021-07-19 at 2 32 52 PM

__After:__
Screen Shot 2021-07-19 at 2 33 01 PM

[PURCHASE-2816]: https://artsyproduct.atlassian.net/browse/PURCHASE-2816
[PURCHASE-2825]: https://artsyproduct.atlassian.net/browse/PURCHASE-2825

PR #8165, opened 2021-08-10
revert: Adjust MO button style and move order update timestamp

lilyfromseattle:
Re-implementing changes from [this PR](https://github.com/artsy/force/pull/8007)


jo-rs (approved):
Thank you for taking care of this 💜 
Changing to semantic PR title and squashing on green.

PR #8166, opened 2021-08-10
Remove the Help chat bubble from Inquiry checkout modal

lilyfromseattle:
[PURCHASE-2879]

*No screenshots because help chat doesn't render locally*

Current behavior:

Given a buyer has submitted an offer through an inquiry,
And they have received a counteroffer from the seller,
When they open the counteroffer modal,
Then the Help chat bubble renders at the bottom right of the modal.

Improved behavior:

Given a buyer has submitted an offer through an inquiry,
And they have received a counteroffer from the seller,
When they open the counteroffer modal,
Then the Help chat bubble does not render.

[PURCHASE-2879]: https://artsyproduct.atlassian.net/browse/PURCHASE-2879

### DIFF
--- a/src/v2/Apps/Order/__tests__/OrderApp.jest.tsx
+++ b/src/v2/Apps/Order/__tests__/OrderApp.jest.tsx
@@ -571,6 +571,15 @@ describe("OrderApp", () => {
     expect(subject.find("ZendeskWrapper")).toHaveLength(1)
   })
 
+  it("does not show the Zendesk chat integration button if in a modal", () => {
+    const props = getProps()
+    const subject = getWrapper({
+      props: { ...props, location: { query: { isModal: true } } },
+    })
+
+    expect(subject.find("ZendeskWrapper")).toHaveLength(1)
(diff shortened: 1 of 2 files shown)


ashleyjelks (approved):

src/v2/Apps/Order/__tests__/OrderApp.jest.tsx
```diff
@@ -571,6 +571,15 @@ describe("OrderApp", () => {
     expect(subject.find("ZendeskWrapper")).toHaveLength(1)
   })
 
+  it("does not show the Zendesk chat integration button if in a modal", () => {
+    const props = getProps()
+    const subject = getWrapper({
+      props: { ...props, location: { query: { isModal: true } } },
+    })
+
+    expect(subject.find("ZendeskWrapper")).toHaveLength(1)
```
damassi, line 580: @lilyfromseattle - should this be `toHaveLength(0)` since we don't want it to be in the modal?

PR #8167, opened 2021-08-10
feat: Improve unsubscribe fallback GRO-475

jonallured:
We discovered some traffic was making it to our unsubscribe page without a token so this PR aims to improve that UX by nudging them to log in. Once they are logged in, then the page will behave as expected and they can change their email frequency. Here's how the fallback page looks:

Screen Shot 2021-08-10 at 3 19 15 PM

In terms of the actual code here my approach was:

* add some specs for the three paths through this component
* extract the existing fallback component
* update that fallback component

And that's it!

https://artsyproduct.atlassian.net/browse/GRO-475

### DIFF
--- a/src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx
+++ b/src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx
@@ -7,6 +7,14 @@ import { UnsubscribeApp_me } from "v2/__generated__/UnsubscribeApp_me.graphql"
 import { UnsubscribeLoggedInFragmentContainer } from "./Components/UnsubscribeLoggedIn"
 import { UnsubscribeLoggedOut } from "./Components/UnsubscribeLoggedOut"
 
+const UnsubscribeFallback = () => {
+  return (
+    <Message variant="error" my={4}>
+      Unable to update your email preferences
+    </Message>
+  )
+}
+
 interface UnsubscribeAppProps {
   me: UnsubscribeApp_me | null
 }
@@ -16,11 +24,7 @@ export const UnsubscribeApp: React.FC<UnsubscribeAppProps> = ({ me }) => {
   const { authentication_token: authenticationToken } = match.location.query
 
   if (!me && !authenticationToken) {
-    return (
-      <Message variant="error" my={4}>
-        Unable to update your email preferences
-      </Message>
-    )
+    return <UnsubscribeFallback />
   }
 
   return (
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx
```diff
@@ -7,37 +7,37 @@ import { UnsubscribeApp_me } from "v2/__generated__/UnsubscribeApp_me.graphql"
...
-      </Message>
-    )
-  }
+  const showFallback = !me && !authenticationToken
+  const showLoggedIn = me && !authenticationToken
+  const showLoggedOut = !me && authenticationToken
 
   return (
     <>
+      {showFallback && <UnsubscribeFallback />}
```
dzucconi, line 31: In this case it doesn't matter much, but in general, rather than redeclaring a component in a variable in the function body, just inline it (or extract it, in this case it's simple enough though). (This _can_ create issues when images are used)

dzucconi (approved):

src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx
jonallured: Hrm, not sure I follow this feedback - maybe some examples would help?

dzucconi:
```suggestion
      {showFallback && (
        <Message variant="error" my={4}>
          Please sign in to update your email preferences
        </Message>
      )}
```

dzucconi: But you'd have to update the specs, blah blah. Kind of just a general comment: I see this a lot and we should avoid doing it because it can cause very mysterious issues with re-renders. In this case; it's not really a big deal.

jonallured: > rather than redeclaring a component in a variable in the function body

I don't think I did this - I think I extracted it to a variable in the file, no? Or are you saying I should have extracted it all the way into a separate file?

dzucconi: Oh, lol you're absolutely right. __This is fine ignore me__

damassi: As a related aside, I was debugging some serious sidebar weirdness in the Palette docs and lo and behold, declared the component in the function, versus outside of it.

PR #8175, opened 2021-08-11
chore: removes icon fonts preload

dzucconi:
So the places the icon font is used on the artwork page are modal and once they appear on page the font will load, so there shouldn't be a need to preload this.


icirellik (approved):
Sounds like a great idea to me!

PR #8178, opened 2021-08-11
refactor: removes remaining instances of deprecated responsive image

dzucconi:
In an effort to fix an issue with Image width & height I'm removing this old `ResponsiveImage` component that shouldn't ever be used. Replaced with one-off inline styles.


src/v2/Apps/Artist/Routes/Consign/Components/ArtistConsignHeader/ArtistConsignHeaderImages.tsx
```diff
@@ -83,7 +86,16 @@ const LeftImagePhoto: React.FC<ImageProps> = ({ image }) => {
...
+  return (
+    <div
+      style={{
+        width: "100%",
+        paddingBottom: "100%",
+        backgroundImage: `url(${image.resized.url})`,
+        backgroundSize: "contain",
+      }}
+    />
+  )
```
damassi, line 98: @dzucconi - this would really benefit from using `ResponsiveBox` or at least creating a component that wraps this

dzucconi: I mean sure; we shouldn't do this _at all_. I'm simply replacing it with what was there previously. This is just exactly what `ResponsiveImage` was doing, not what it should be.

PR #8180, opened 2021-08-11
feature: Palette Charts

icirellik:
Use `@artsy/palette-charts` to limit the pages d3 is loaded on.

Depends on:

https://github.com/artsy/palette/pull/1016

Before
d3-before

After
d3-after


dzucconi:
🙏 🙏 🙏

PR #8186, opened 2021-08-11
fix: uses a consistent height for artwork context bar

dzucconi:
[Figma spec](https://www.figma.com/file/m6gDpKHEWDbYJyrwsVZDBr/Artsy-3.0-Design-System?node-id=1007%3A1392)


src/v2/Apps/ArtistSeries/Components/ArtistSeriesHeader.tsx
```diff
@@ -94,7 +94,7 @@ const ArtistSeriesHeader: React.FC<ArtistSeriesHeaderProps> = ({
           </Text>
 
           {descriptionFormatted && (
-            <HTML variant="text">
+            <HTML variant="sm">
```
dzucconi, line 97: Random thing I noticed. (We'll have to circle back here to refine the typography in this page's context bar as well.)

PR #8188, opened 2021-08-12
feat(FX-3132): Added link to the fair page on fair organizer page, show 404 when there is no profile

anastasiapyzhik:
[FX-3132]

**Demo**

link

_moving-image-istanbul_ organizer hasn't profile hence its page is not available
Screenshot 2021-08-16 at 10 23 18
Screenshot 2021-08-16 at 10 23 40

[FX-3132]: https://artsyproduct.atlassian.net/browse/FX-3132

### DIFF
--- a/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
+++ b/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
@@ -1,5 +1,7 @@
+import React from "react"
 import loadable from "@loadable/component"
 import { graphql } from "react-relay"
+import { ErrorPage } from "v2/Components/ErrorPage"
 import { AppRouteConfig } from "v2/System/Router/Route"
 
 const FairOrganizerApp = loadable(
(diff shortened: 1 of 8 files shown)


dimatretyak (approved):

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -29,9 +31,22 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
     prepare: () => {
       FairOrganizerApp.preload()
     },
+    render: ({ Component, props }) => {
+      if (Component && props) {
```
mdole, line 35: **Question:** is there a case when we wouldn't have component or props, so this would fail? If so, what happens?

dblandin (approved):
Looks great!

mdole (commented):
Oop, I see Devon beat me to it 😁 I didn't have any blocking feedback, just some thoughts!

I left one question below, and one other general question: should we address the case where a fair organizer exists but has no associated fairs? This is definitely an unlikely case, but I think right now, it would just break if we try to fetch `fairsConnection(first: 1)` and get back `null`

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
anastasiapyzhik: @mdole those data just come from the server side and `render` method may be executed without `props` for the first time. And in this case nothing is rendered 🤷‍♀️

anastasiapyzhik:
> Oop, I see Devon beat me to it 😁 I didn't have any blocking feedback, just some thoughts!
> 
> I left one question below, and one other general question: should we address the case where a fair organizer exists but has no associated fairs? This is definitely an unlikely case, but I think right now, it would just break if we try to fetch `fairsConnection(first: 1)` and get back `null`

Might be mistaken but I think that if organizer doesn't have any fairs it doesn't have a profile too hence it's not rendered

PR #8191, opened 2021-08-12
chore: adds restricted imports eslint config

dzucconi:
Any others? Sad we didn't think of doing this sooner!


dzucconi:
odd...

damassi:
That acceptance test failure generally means something has changed in CMS and the test needs to be updated.

PR #8192, opened 2021-08-12
feat(FX-3077): exhibitors tab

tr-ann:
Jira: [FX-3077](https://artsyproduct.atlassian.net/browse/FX-3077)

### Description
The purpose of the ticket is create Exhibitors tab that is visually laid out as follows: [Figma](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3474%3A15570)

### Changes
- Added exhibitors tab with lazy-loaded components
- Added tests for it
- Added new env variable (default `false`) to render new Exhibitors tab (render tab only if it is true)

- Partners info fetched from high level query (early info fetched separately for every exhibitors group)

Note: tabs are **not** sticky now (it is the other ticket)

### Demo

_Desktop_ 

https://user-images.githubusercontent.com/56556580/129546788-bd448dea-e3e0-4be8-9375-ae5001bc9ebe.mov

_Mobile_

https://user-images.githubusercontent.com/56556580/129546912-458364f8-eb25-4c69-817a-a2fc7c1edb1d.mov


src/v2/Apps/Fair/Routes/FairExhibitors.tsx
```diff
@@ -1,219 +1,60 @@
...
+          exhibitor => exhibitor?.partnerID || ""
+        )
+
+        return (
+          <React.Fragment key={exhibitorsGroup.letter}>
+            <Text variant="lg" my={4}>
+              {exhibitorsGroup.letter}
+            </Text>
+            {isEnteredView ? (
+              <FairExhibitorsGroup ids={partnerIds} />
```
dblandin, line 35: **question**: If Metaphysics supported a `partner` field under the `FairExhibitor` field, would we be able to fetch all of the data in the top-level query instead of fanning out to multiple QueryRenders? That might be worth adding in so we can simplify the data fetching here.

src/v2/Apps/Fair/Routes/FairBooths.tsx
```diff
@@ -0,0 +1,217 @@
...
+      </Media>
+
+      <Spacer mt={4} />
+
+      <LoadingArea isLoading={isLoading}>
+        {fair.exhibitors?.edges!.map((edge, index) => {
+          const show = edge?.node!
+          if (show?.counts?.artworks === 0 || !show?.partner) {
+            // Skip rendering of booths without artworks
+            return null
```
dblandin, line 128: **question**: I realize this isn't new code but as a follow-up I wonder if we can use the `isDisplayable` field to determine whether we should render or not.

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorCard.tsx
```diff
@@ -0,0 +1,174 @@
...
+            cropped(width: 50, height: 50) {
+              src
+              srcSet
+            }
+          }
+          image {
+            url(version: "medium")
+          }
+        }
+        locations: locationsConnection(first: 20) {
```
dblandin, line 163: **suggestion**: I believe the partner type already has a `cities` field which is a unique list of cities the partner has locations within.

dblandin (commented):
Looking good! I think we might have some opportunities to simplify the implementation, such as supporting `partner` under the `FairExhibitor` type in Metaphysics and using `Partner#cities` vs `Partner#locationsConnection`.

src/v2/Apps/Fair/Routes/FairExhibitors.tsx
tr-ann: `FairExhibitor` type includes only partner `id` and `name`. I'm not sure if we can fetch all partners info in the top-level query.

dblandin: It doesn't now - but I believe Metaphysics could be updated to expose a `partner` field under `FairOganizer` that would allow you to query for everything in one QueryRenderer.

Something like:

```graphql
{
  fair(id: "the-armory-show-2020") {
    name
    exhibitorsGroupedByName {
      letter
      exhibitors {
        partner {
          name
          href
          profile {
            icon {
              cropped(width: 50, height: 50) {
                src
                srcSet
              }
            }
          }
        }
      }
    }
  }
}
```

What that be preferable to breaking out the groups and querying separately?

tr-ann: Yes, I agree, maybe it will be better to update metaphysics and add `Partner`  type to `FairExhibitors`.
But for now it implemented like this 🙂

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+      variables={{ ids }}
+      render={({ error, props }) => {
+        if (error || !props) {
+          return <FairExhibitorsGroupPlaceholder />
+        }
+
+        return (
+          <FairExhibitorsGroupFragmentContainer {...rest} {...(props as any)} />
+        )
+      }}
```
dimatretyak, line 81: chore: maybe it would be better to render placeholder only when there are no errors and data?

```javascript
render={({ error, props }) => {
  if (error) {
    return null
  }

  if (!props) {
    return <FairExhibitorsGroupPlaceholder />
  }

  if (props.partnersConnection) {
    <FairExhibitorsGroupFragmentContainer {...rest} partnersConnection={props.partnersConnection} />
  }

  return null
}}
```

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
```diff
@@ -0,0 +1,84 @@
...
+  partnersConnection: FairExhibitorsGroup_partnersConnection
+}
+
+export const FairExhibitorsGroup: React.FC<FairExhibitorsGroupProps> = ({
+  partnersConnection,
+}) => {
+  return (
+    <Box>
+      <GridColumns position="relative" gridRowGap={[4, 4]}>
+        {partnersConnection?.edges?.map(exhibitor => {
```
dimatretyak, line 21: advice: you can use `src/v2/Utils/extractNodes.ts` helper method

dimatretyak (approved):

dblandin:
I feel like some organization might help with the review, QA, and rollout of these changes. Here's what I would propose:

- Open a separate PR that _just_ renames the existing `FairExhibitors` route to `FairBooths`
- Update this PR to include the new `FairExhibitors` route (now without a naming collision)
- Update this PR to conditionally render the new tab via an environment variable (`/exhibitors route accessible regardless)

@tr-ann What do you think? Does that make sense?

tr-ann:
@dblandin Sounds good. So I need just to create a new PR and rename `FairExhibitors` to `FairBooth`, right?

dblandin:
> @dblandin Sounds good. So I need just to create a new PR and rename FairExhibitors to FairBooth, right?

Yes, this shouldn't have any user-facing changes. Just renaming the current `FairExhibitors` route (labeled as `Booths` in the UI) to `FairBooths` in the code.

And I'm imagining this PR, once updated, could also include a feature flag check backed by an environment variable. Something like `ENABLE_FAIR_PAGE_EXHIBITORS_TAB`. And if true, it'll render the new "Exhibitors" tab alongside Overview and Booths.

And if the .env variable is true, the we can render booths (tab with exhibitors artworks). Is it right?

src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -74,13 +74,26 @@ const FairApp: React.FC<FairAppProps> = ({ children, fair }) => {
...
+        </RouteTab>
+
         <RouteTab
           to={`${fairHref}/booths`}
           exact
           onClick={trackTabData(
             `${fairHref}/booths`,
             "Booths",
-            ContextModule.exhibitorsTab
+            ContextModule.artistsTab
```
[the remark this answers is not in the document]
louislecluse: Added [here](https://github.com/artsy/cohesion/pull/222) @dblandin

tr-ann:
@dblandin I added env variable (hope I understand you correctly 🙂 ), so new Exhibitors tab renders only if `ENABLE_FAIR_PAGE_EXHIBITORS_TAB` is true. Can you check the code please?

dblandin (approved):
Looks great!

dblandin:
Going to go ahead and merge this in.

PR #8200, opened 2021-08-13
feature: Minimize Moment Timezone

icirellik:
Moment Timezone provides a webpack plugin that minimizes the timezone
data we need to include client-side, this adds the plugin and limits the
timezones to the only one that is used in v2 apps `America/New_York`

Timezone data from 32 KiB gzip -> 300 bytes gzip

Before
moment-timezone-old

After
moment-timezone-new

### DIFF
--- a/webpack/envs/clientPlugins.js
+++ b/webpack/envs/clientPlugins.js
@@ -1,0 +1,17 @@
+// @ts-check
+
+import MomentTimezoneDataPlugin from "moment-timezone-data-webpack-plugin"
+const currentYear = new Date().getFullYear()
+
+export const clientPlugins = [
+  // To include only specific zones, use the matchZones option
+  new MomentTimezoneDataPlugin({
+    matchZones: /^America\/New_York/,
+  }),
+
+  // To keep all zones but limit data to specific years, use the year range options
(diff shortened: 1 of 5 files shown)


webpack/envs/clientPlugins.js
```diff
@@ -0,0 +1,17 @@
...
+  // To include only specific zones, use the matchZones option
+  new MomentTimezoneDataPlugin({
+    matchZones: /^America\/New_York/,
+  }),
+
+  // To keep all zones but limit data to specific years, use the year range options
+  new MomentTimezoneDataPlugin({
+    startYear: currentYear - 5,
+    endYear: currentYear + 5,
+  }),
```
damassi, line 16: Should [these two](https://github.com/artsy/force/blob/13194796c96d16dda38be0ca05e7d6ceec02c84d/webpack/envs/commonPlugins.js#L13-L20) also be added to this list?

icirellik: They are already in the common plugins.

damassi: The thinking was: both of those have to do with client, and you explicitly created a new `clientPlugins` file; now there are client plugins in two locations, and an additional new file.

PR #8210, opened 2021-08-14
feature: Slim Lodash

icirellik:
Round 2 of slim lodash, it turns out that the lodash-webpack-plugin
is aggressive by default. An extra configuration option was
required to enable all the functionality we use.

The new change:

https://github.com/artsy/force/compare/feature-loadash-slim?expand=1#diff-fbd53dc365f51bf49b96ebd837db791e14e93411af4f3b66980a61ec563efddcR10

The previous PR:

https://github.com/artsy/force/pull/8206/files

### DIFF
--- a/webpack/envs/commonLoaders.js
+++ b/webpack/envs/commonLoaders.js
@@ -21,6 +21,30 @@ export const babelLoader = {
   ],
 }
 
+/**
+ * This loader ensures that all requisite babel configurations are in place to
+ * properly optimize lodash.
+ */
+export const babelLoaderWithLodashOptimization = {
+  exclude: /(node_modules)/,
+  include: path.resolve(basePath, "src"),
+  test: /(\.(js|ts)x?$)/,
+  use: [
+    {
+      loader: "babel-loader",
+      options: {
+        cacheDirectory:
+          !env.onCi && path.join(basePath, ".cache", "babel/force"),
+        plugins: [
+          "lodash",
+          env.isDevelopment && require.resolve("react-refresh/babel"),
+        ].filter(Boolean),
+        presets: [["@babel/preset-env", { modules: false }]],
+      },
+    },
+  ],
+}
+
 export const coffeeLoader = {
   exclude: /(node_modules)/,
   include: path.resolve(basePath, "src"),
--- a/package.json
+++ b/package.json
@@ -352,6 +352,7 @@
     "konami-js": "0.0.2",
     "levenary": "1.1.1",
     "lint-staged": "10.2.6",
+    "lodash-webpack-plugin": "0.11.6",
     "mocha": "3.4.1",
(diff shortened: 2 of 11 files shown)


webpack/envs/commonLoaders.js
```diff
@@ -21,6 +21,26 @@ export const babelLoader = {
...
+  include: path.resolve(basePath, "src"),
+  test: /(\.(js|ts)x?$)/,
+  use: [
+    {
+      loader: "babel-loader",
+      options: {
+        cacheDirectory:
+          !env.onCi && path.join(basePath, ".cache", "babel/force"),
+        plugins: [
+          "lodash",
```
damassi, line 35: Why are we now [duplicating config](https://github.com/artsy/force/blob/master/babel.config.js#L43)?

icirellik: In response to the original, edit.

"Why are we now duplicating config?"

It's not duplicated when it's different.

In response to the future edits, naming is hard. It's tailored for lodash, so the name reflects that. Helpful suggestions welcomed.

"It seems a bit specific when folks looking at config know the usual suspects."

I'm not following this? Pretty sure only us two touch these files. Mostly myself.

damassi: > In response to the original, edit.

Everything does not need to be a defensive fight. The edit button is there for a reason and you don't need to pedantically comment on every minor revision that I make on a PR comment that I'm using to craft a thought. I asked a simple question about there seemingly being two lodash babel plugins used in two separate spots, and then was wondering about naming. Chill.

webpack/envs/commonLoaders.js
```diff
@@ -21,6 +21,30 @@ export const babelLoader = {
...
+          env.isDevelopment && require.resolve("react-refresh/babel"),
+        ].filter(Boolean),
+        presets: [["@babel/preset-env", { modules: false }]],
```
icirellik, line 42: Required for proper `LodashModuleReplacementPlugin` optimization.

webpack/envs/commonLoaders.js
```diff
@@ -21,6 +21,30 @@ export const babelLoader = {
...
+  exclude: /(node_modules)/,
+  include: path.resolve(basePath, "src"),
+  test: /(\.(js|ts)x?$)/,
+  use: [
+    {
+      loader: "babel-loader",
+      options: {
+        cacheDirectory:
+          !env.onCi && path.join(basePath, ".cache", "babel/force"),
+        plugins: [
```
icirellik, line 38: Ensure the `lodash` plugin is enabled for for `LodashModuleReplacementPlugin` optimization.

ovasdi (approved):
LGTM❗

mc-jones:
👍

mc-jones (approved):

PR #8212, opened 2021-08-16
[GRO-460] Update image/copy in homepage banner for logged out users

TMcMeans:
Marketing banners are updated in the admin tool but the sign up banner for logged out users is hardcoded in the Home app. Looking in the code, it seems to me the reason why this is hardcoded is because it renders conditionally based on the isLoggedOut user status... but I don't see why this can't also be moved into the admin app? The copy lives in something called a HeroUnit, which is just a statically-typed object. I'm not sure how admin app stuff gets into force, but I could probably figure it out.  

Waiting on new sub-header copy from the brand team and then this pr is ready to go. 

### Screenshots
Screen Shot 2021-08-16 at 10 54 06 AM

### DIFF
--- a/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx
+++ b/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx
@@ -276,10 +276,9 @@ export const HomeHeroUnitFragmentContainer = createFragmentContainer(
 )
 
 export const LOGGED_OUT_HERO_UNIT: StaticHeroUnit = {
-  title: "Collect art from leading galleries, fairs, and auctions",
+  title: "Collect art by the world's leading artists",
   subtitle: "Sign up to get updates about your favorite artists",
   href: "/signup",
   linkText: "Sign up",
-  backgroundImageURL:
-    "https://files.artsy.net/images/alexander-calder-rouge-triomphant-triumphant-red-1959-1965.jpg",
+  backgroundImageURL: "http://files.artsy.net/homepage/signup-banner.png",
 }
--- a/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx
+++ b/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx
@@ -277,7 +277,7 @@ export const HomeHeroUnitFragmentContainer = createFragmentContainer(
 
 export const LOGGED_OUT_HERO_UNIT: StaticHeroUnit = {
   title: "Collect art by the world's leading artists",
-  subtitle: "Sign up to get updates about your favorite artists",
+  subtitle: "Register for updates on available works, market news, and more.",
   href: "/signup",
   linkText: "Sign up",
   backgroundImageURL: "http://files.artsy.net/homepage/signup-banner.png",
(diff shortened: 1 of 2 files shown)


TMcMeans:
Looks like its easier to keep this copy hardcoded since its rarely touched and moving it to the admin app would be a pain.

damassi:
> Looks like its easier to keep this copy hardcoded since it's rarely touched and moving it to the admin app would be a pain.

This has come up a lot in the past, meaning: we've created CMSs, weird admin features and other convoluted tools to address copy changes that we assumed would be changed regularly, but in practice rarely. So 👍 on keeping things simple.

src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx
```diff
@@ -276,10 +276,9 @@ export const HomeHeroUnitFragmentContainer = createFragmentContainer(
...
 
 export const LOGGED_OUT_HERO_UNIT: StaticHeroUnit = {
-  title: "Collect art from leading galleries, fairs, and auctions",
+  title: "Collect art by the world's leading artists",
   subtitle: "Sign up to get updates about your favorite artists",
   href: "/signup",
   linkText: "Sign up",
-  backgroundImageURL:
-    "https://files.artsy.net/images/alexander-calder-rouge-triomphant-triumphant-red-1959-1965.jpg",
+  backgroundImageURL: "http://files.artsy.net/homepage/signup-banner.png",
```
damassi, line 283: Should this be a `.png`? We should be able to get a much lower file-size if it were a jpg.

dzucconi: Gemini will convert it here: https://github.com/artsy/force/blob/d6f7d907c5db1b1f0ae2c9ea09225331513db800/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnit.tsx#L61-L68

TMcMeans: Gotcha, I'll change the file extension in vanity

dzucconi: You'd have to re-encode it (in Photoshop for instance) rather than change it's file extension. But this isn't necessary, Gemini (our image processing service), will re-encode this to JPG (or WebP) automatically.

TMcMeans: Wait, does just changing the file extension change the file type?

TMcMeans: If not, I might have to petition to design to get a .jpg which isn't a big deal.

damassi: I believe what @dzucconi is saying is that all is ok -- that `cropped` function up above will take care of everything automatically behind the scenes (read png -> user gets served a jpg).

TMcMeans: Gotcha, I was going to mention that there are other imgs in vanity that are .pngs but it looks like Gemini does the file conversion behind the scenes so folks don't have to worry about that.

The-Beez-Kneez (approved):

The-Beez-Kneez:
LGTM, mergin!

PR #8214, opened 2021-08-16
feat(FX-3158): Fair exhibitors A-Z navigation

anastasiapyzhik:
[FX-3158]

[FX-3158]: https://artsyproduct.atlassian.net/browse/FX-3158

**The purpose of this PR** is to add the ”A B C … X Y Z 0-9” navigator to the top of the new Exhibitors A-Z tab.
**Acceptance criteria:**  clicking on a single “letter” (A, B, C…) in the filter tab scrolls down the page

scroll

scroll-mob

### DIFF
--- a/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
+++ b/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
@@ -3,6 +3,7 @@ import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
 import { Text, Flex, Swiper } from "@artsy/palette"
 import { Media } from "v2/Utils/Responsive"
+import { scrollIntoView } from "v2/Utils/scrollHelpers"
 import { ExhibitorsLetterNav_fair } from "v2/__generated__/ExhibitorsLetterNav_fair.graphql"
 
 const LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ".split("").concat(["0-9"])
@@ -28,6 +29,15 @@ export const ExhibitorsLetterNav: React.FC<ExhibitorsLetterNavProps> = ({
               title={
                 isEnabled ? `View exhibitors starting with “${letter}”` : ""
               }
+              onClick={() => {
+                if (isEnabled) {
+                  scrollIntoView({
+                    selector: `#jump--letter${letter}`,
+                    offset: 120,
+                    behavior: "smooth",
+                  })
+                }
+              }}
             >
               {letter}
             </Letter>
--- a/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
+++ b/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
@@ -1,10 +1,12 @@
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
-import { Text, Flex, Swiper } from "@artsy/palette"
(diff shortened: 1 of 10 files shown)


src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
```diff
@@ -0,0 +1,83 @@
...
+              key={letter}
+              color={isEnabled ? "black100" : "black10"}
+              title={
+                isEnabled ? `View exhibitors starting with “${letter}”` : ""
+              }
+              onClick={() => {
+                if (isEnabled) {
+                  scrollIntoView({
+                    selector: `#jump--letter${letter}`,
+                    offset: 120,
```
dblandin, line 36: **question**: Where does this offset come from?

I wonder if we can reuse the offset calculation found here, although I'm not sure what 30 refers to.

https://github.com/artsy/force/blob/f23ebc54255c2de0df547f6443390d00472a833b/src/v2/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx#L69

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
```diff
@@ -0,0 +1,83 @@
...
+        })}
+      </Flex>
+    )
+  }
+
+  return (
+    <>
+      <Media lessThan="md">
+        <Swiper snap="start">
+          <Letters width={1300} />
```
dblandin, line 54: **question**: Is there a way to support the scrolling rail without the fixed width here?

dblandin (approved):
This looks great! Added a couple non-blocking questions in the review.

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
anastasiapyzhik: This offset comes from my mind, I just chose the fitting size 😄  Your suggestion looks much better!

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
anastasiapyzhik: Yes, we can use margin for each letter to make space between them

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
```diff
@@ -0,0 +1,83 @@
...
+      </Media>
+
+      <Media greaterThanOrEqual="md">
+        <Letters />
+      </Media>
+    </>
+  )
+}
+
+const Letter = styled(Text).attrs({ p: 1, variant: "md" })`
```
mdole, line 65: @anastasiapyzhik I think we should remove this padding around each letter - in the specs, the letters are all the way to the left side of the screen and they line up perfectly with the exhibitors. Because they have padding now, they are slightly toward the middle of the screen and look slightly strange.

mdole (commented):
Nice work! Left one small comment about an alignment issue, otherwise looking great to me

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
anastasiapyzhik: 👍  That makes sense for desktop but it seems that we should keep the padding on mobile

Screenshot 2021-08-23 at 15 55 45

dblandin (approved):
💯

PR #8215, opened 2021-08-16
feat: [PURCHASE-2872] - migrate Conversation from v2 to v3

ivan-badyulya:
JIRA -> [Web - Conversation](https://artsyproduct.atlassian.net/secure/RapidBoard.jspa?rapidView=41&projectKey=PURCHASE&modal=detail&selectedIssue=PURCHASE-2872&assignee=60469b1e43eac6006f636d07)

What I should do:
Migrate [Conversation](https://github.com/artsy/force/tree/master/src/v2/Apps/Conversation) app on force to use Palette V3.

What I did:
- migrated Conversation module to v3
- refactored code
- got rid of ts strict null check disabled comment (most of them)

gif BEFORE:
ezgif com-gif-maker

gif AFTER:
ezgif com-gif-maker (1)

### DIFF
--- a/src/v2/Apps/Conversation/Components/ConversationList.tsx
+++ b/src/v2/Apps/Conversation/Components/ConversationList.tsx
@@ -1,19 +1,23 @@
-import { Box, Spinner, color, media } from "@artsy/palette"
-import { ConversationList_me } from "v2/__generated__/ConversationList_me.graphql"
 import React, { useState } from "react"
 import {
   RelayPaginationProp,
   createPaginationContainer,
   graphql,
 } from "react-relay"
-import { ConversationSnippetFragmentContainer as ConversationSnippet } from "./ConversationSnippet"
 import styled from "styled-components"
+import compact from "lodash/compact"
+import { themeGet } from "@styled-system/theme-get"
+import { Box, Spinner, media } from "@artsy/palette"
+
+import { ConversationSnippetFragmentContainer as ConversationSnippet } from "./ConversationSnippet"
 import { ConversationListHeader } from "./ConversationListHeader"
+
+import { ConversationList_me } from "v2/__generated__/ConversationList_me.graphql"
 
 const Container = styled(Box)`
   height: 100%;
   overflow: hidden;
-  border-right: 1px solid ${color("black10")};
+  border-right: 1px solid ${themeGet("colors.black10")};
   ${media.xs`
     border-right: none;
     border-bottom: none;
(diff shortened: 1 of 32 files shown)


src/v2/Apps/Conversation/Components/ConversationSnippet.tsx
```diff
@@ -105,10 +95,10 @@ const ConversationSnippet: React.FC<ConversationSnippetProps> = props => {
         href={`/user/conversations/${conversation.internalID}`}
         underlineBehavior="none"
       >
-        <Flex alignItems="center" px="5px" width="100%" height="120px">
+        <Flex alignItems="center" px={0.5} width="100%" height="120px">
           <Flex
             alignItems="center"
-            width={space(2)}
+            width={`${themeGet("space.2")}`}
```
dzucconi, line 101: Does this work? (In cases like this it's fine to just use `width={20}`)

src/v2/Apps/Conversation/Components/ConversationList.tsx
```diff
@@ -74,21 +75,14 @@ const ConversationList: React.FC<ConversationsProps> = props => {
...
-              // @ts-expect-error STRICT_NULL_CHECK
+              isSelected={edge?.node?.internalID === selectedConversationID}
+              conversation={edge.node!}
               key={edge.cursor}
               hasDivider={
-                // @ts-expect-error STRICT_NULL_CHECK
                 conversations.indexOf(edge) !== selectedConversationIndex &&
-                // @ts-expect-error STRICT_NULL_CHECK
                 conversations.indexOf(edge) !== selectedConversationIndex - 1 &&
-                // @ts-expect-error STRICT_NULL_CHECK
```
damassi, line 91: thanks for fixing these 👍

src/v2/Apps/Conversation/Components/Details.tsx
```diff
@@ -89,25 +91,23 @@ export const Details: FC<DetailsProps> = ({
...
+    ?.reduce((previous, current) => previous?.concat(current!), [])
+    ?.filter(attachment => !attachment?.contentType.includes("image"))
 
   const attachmentItems = attachments?.map(attachment => {
     return (
       <Link
-        key={attachment.id}
-        href={attachment.downloadURL}
+        key={attachment?.id ?? undefined}
+        href={attachment?.downloadURL ?? undefined}
```
damassi, line 104: Rather than creating undefined links, might we guard against the existence of `downloadUrl` and then conditionally render? That or filter `attachments` so only truthy values appear.

damassi (approved):
Nice work 👍 

One minor design comment is: the right hand conversation section feels much too wide on larger screens as chat windows are generally compact inside of UI. Perhaps that can be capped as something less than the max breakpoint size or maybe capped at a max width and then _centered_ within a wide right hand column. 

(This is obviously a #design question, so non-blocking, but wanted to call it out)

dzucconi (approved):

ivan-badyulya:
> Nice work 👍
> 
> One minor design comment is: the right hand conversation section feels much too wide on larger screens as chat windows are generally compact inside of UI. Perhaps that can be capped as something less than the max breakpoint size or maybe capped at a max width and then _centered_ within a wide right hand column.
> 
> (This is obviously a #design question, so non-blocking, but wanted to call it out)

Yeah, there is something like that. But it was the same thing with v2 :)

PR #8217, opened 2021-08-16
feat: Temporarily load recaptcha in form

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-443 

This is a temporary fix for loading recaptcha only when the (old) Inquiry form loads. The new one loads things dynamically so this will quickly be phased out.


src/desktop/components/inquiry_questionnaire/views/inquiry.coffee
```diff
@@ -8,11 +8,17 @@ sd = require('sharify').data
...
 
 module.exports = class Inquiry extends StepView
 
   initialize: ({ @user, @inquiry, @artwork, @state, @trail, @modal }) ->
-    recaptcha('inquiry_impression')
+    loadRecaptcha()
+
+    setTimeout(() =>
+      recaptcha('inquiry_impression')
+    , 1000)
```
damassi, line 20: this is a game of wack-a-mole, even though `async` is set to false below. Seems fine for the time being.

dzucconi (approved):

PR #8219, opened 2021-08-16
chore: Move loadenv to /lib

damassi:
Something I missed during code-review! Since this file is more of an implementation detail wanted to move it out of the main project root and into `lib` where we have other similar files. Now the project root remains scoped to the primary motifs --boot, dev, middleware, config.


ovasdi (approved):
LGTM❗

PR #8220, opened 2021-08-16
feat: Disable search box for marketing page GRO-485

jonallured:
This is a quick PR to just add a particular meta tag to a marketing page. The goal here is to signal to Google that we don't want the Buying with Artsy marketing page to show up as a site link.

Looks like this:

Screen Shot 2021-08-16 at 4 55 58 PM

https://artsyproduct.atlassian.net/browse/GRO-485

/cc @artsy/grow-devs


damassi (approved):

damassi:
#mergeOnGreen

PR #8223, opened 2021-08-17
refactor: exhibitors tab

tr-ann:
The purpose of the PR is **rename** existing "Exhibitors" tab and route to "Booths" to avoid naming collision when add new "Exhibitors A-Z" tab.

Made some changes to improve the code


src/v2/Apps/Fair/FairApp.tsx
```diff
@@ -80,7 +80,7 @@ const FairApp: React.FC<FairAppProps> = ({ children, fair }) => {
           onClick={trackTabData(
             `${fairHref}/booths`,
             "Booths",
-            ContextModule.exhibitorsTab
+            "boothsTab" as ContextModule
```
dblandin, line 83: **suggestion / non-blocking**: Let's make sure to ticket this new context so we can swap `ContextModule.boothsTab` in here.

cc/ @louislecluse FYI ☝🏼

dblandin (approved):
LGTM! A couple minor comments. Feel free to merge when ready!

src/v2/Apps/Fair/FairApp.tsx
tr-ann: @dblandin should I do something with it? Or leave it as it is?

dblandin: You can leave it as is for this PR! I'll follow up with another Jira ticket.

PR #8224, opened 2021-08-17
feat(PX-4286): Users can see viewing rooms on the gallery profile page [WIP]

Serge0n:
Jira Ticket: https://artsyproduct.atlassian.net/browse/PX-4286
Discussion: https://artsy.slack.com/archives/C9ZJYFDNF/p1629130378499300

This PR creates a new `Viewing Rooms` tab in PPP.


src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  const {
+    viewingRoomsList: {
+      edges: viewingRooms,
+      pageInfo: { hasNextPage, endCursor },
+      pageCursors,
+    },
+    slug,
+  } = partner
```
ansor4, line 48: non-blocking: Small preference for splitting this into 3 lines:
```js
const { viewingRoomsList, slug } = partner
const { edges, pageInfo, pageCursors } = viewingRoomsList
const { hasNextPage, endCursor } = pageInfo
```
I find it easier when all available variables are on the left-most side of the line, but I understand the succinctness of how you did it!

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRooms.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import { ViewingRoomCardFragmentContainer } from "./ViewingRoomCard"
+import { ViewingRooms_edges } from "v2/__generated__/ViewingRooms_edges.graphql"
+
+interface ViewingRoomsProps {
+  edges: ViewingRooms_edges
+  eventTitle: string
+}
+
+const ViewingRooms: React.FC<ViewingRoomsProps> = ({ edges, eventTitle }) => (
```
damassi, line 12: We should avoid implicitly returned components and always explicitly use a `return`

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRooms.tsx
```diff
@@ -0,0 +1,43 @@
...
+  <>
+    <Text color="black" variant="title" mb={6}>
+      {eventTitle}
+    </Text>
+
+    <GridColumns mb={6} gridRowGap={[2, 4]}>
+      {edges.map(({ node: viewingRoom }) => {
+        if (!viewingRoom) return
```
ansor4, line 20: discussion: Something about this is off to me, I'm not sure if it's the two return statements back-to-back or that we're returning `undefined` in the early exit and hoping that `<GridColumns>` will handle that correctly.

I would feel better if we filtered out the falsey `node`s from `edges` earlier, maybe before returning this component?

```js
const ViewingRooms: React.FC<ViewingRoomsProps> = ({ edges, eventTitle }) => {
  const viewingRooms = edges.filter(node => node.viewingRoom)
  return (
    <>
      <Text color="black" variant="title" mb={6}>
        {eventTitle}
      </Text>

      <GridColumns mb={6} gridRowGap={[2, 4]}>
        {viewingRooms.map((viewingRoom) => {
  		  return (
            <Column key={viewingRoom.internalID} span={[6, 6, 3, 3]}>
              <ViewingRoomCardFragmentContainer viewingRoom={viewingRoom} />
            </Column>
          )
        })}
  ...
}
```
What do you think? It iterates through the `edges` array twice instead of once, but we don't have to worry about falsey `viewingRoom` values affecting `<GridColumns>`

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx
```diff
@@ -0,0 +1,195 @@
...
+  offset,
+  paramsPage,
+}) => {
+  const {
+    match: { location },
+    router,
+  } = useRouter()
+  const [isLoading, setIsLoading] = useState(false)
+
+  if (!partner?.viewingRoomsList) {
```
ansor4, line 37: non-blocking: I would put early exits as early as possible, so in this case, at the top of the function!

ansor4 (commented):
This is coming along nicely!

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRooms.tsx
Serge0n: Okay, thanks! Fixed

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx
```diff
@@ -0,0 +1,195 @@
...
+    },
+    slug,
+  } = partner
+
+  const handleClick = (cursor: string, page: number) => {
+    const canRefetch = paramsPage !== page
+
+    canRefetch && setIsLoading(true)
+
+    canRefetch &&
```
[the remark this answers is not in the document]
Serge0n: yeah) Fixed

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx
Serge0n: fixed!

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRooms.tsx
Serge0n: I've done this before) in: https://github.com/artsy/force/blob/ef6ee3e70f572cdb24d67992b8770078b21f87f6/src/v2/Apps/Partner/Routes/ViewingRooms/index.tsx#L24

This is fine?

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx
Serge0n: Totally agree. Fixed

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

> `src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomsPaginated.tsx`
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 34db8d7a93bfe6cd6f43fadd3c4016de076501f3
</p>

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRooms.tsx
ansor4: Yeah, that works for me!

kiryl-zubarau (changes requested):
LGTM! Could you also add a placeholder for the page? Not very good when the user sees a blank page when loading data.

src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
```diff
@@ -46,6 +47,12 @@ export const NavigationTabs: React.FC<NavigationTabsProps> = ({ partner }) => {
         exact: true,
         hidden: !counts?.displayableShows,
       },
+      {
+        name: "Viewing Rooms",
+        href: route("/viewing-rooms"),
+        exact: true,
+        hidden: !viewingRooms || !viewingRooms.totalCount,
```
oxaudo, line 54: #superminor - I think this line can be `!viewingRooms?.totalCount`

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <Image
+            {...coverImageURL}
+            alt={title}
+            width="100%"
+            height="100%"
+            lazyLoad
+          />
+        </ResponsiveBox>
+      )}
+      {coverImage && title && href && exhibitionPeriod && (
```
oxaudo, line 33: This condition reads a bit funny to me - why we only display `Viewing Room` only if all of those are present? What will happen if let's say we are missing only one?

Should we either display it always or not display it ever?

src/v2/Apps/Partner/Routes/ViewingRooms/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+// import { useRouter } from "v2/System/Router/useRouter"
+
+interface PartnerShowsProps {
+  partner: ViewingRooms_partner
+}
+
+export const ViewingRooms: React.FC<PartnerShowsProps> = ({ partner }) => {
+  const { currentEvents, upcomingEvents, slug } = partner
+
+  // const {
```
oxaudo, line 18: seems like some forgotten comments here?

oxaudo (approved):

src/v2/Apps/Partner/Components/PartnerViewingRooms/ViewingRoomCard.tsx
Serge0n: I added this condition for VR label because of this test:

https://github.com/artsy/force/blob/bae52d423f07ea21f7dbe11155cdd5f0671b683f/src/v2/Apps/Partner/Components/__tests__/PartnerViewingRooms/ViewingRoomCard.jest.tsx#L65

PR #8226, opened 2021-08-18
fix(artist): corrects v3 styling

dzucconi:
This PR fixes a few pieces of the artist page that weren't converted to V3. In particular the gallery submitted bio lacked correct line-spacing which is what prompted this discovery.


damassi:
Thank you!

PR #8228, opened 2021-08-18
feature: Enable Lodash Currying

icirellik:
**Description**

This PR enables Lodash currying in the `lodash-webpack-plugin`. This functionality 
is required for `.reverse()` to work.

There may be a couple of extra feature sets that will need enabling to support all of
the features we used from Ldoash. These will be enabled as needed to limit how
much scope is brought into our client bundles.


dzucconi (approved):
🙏

PR #8242, opened 2021-08-20
GRO-469: Updated Trending Lots to remove closed lots

The-Beez-Kneez:
This is for [GRO-469].

We wanted to update Trending Lots to remove any closed lots from being displayed.

Note: Currently we do not have any "Trending Lots" in staging that are open. So we do not have local screen shots to show other then the empty state curatorial rail.

[GRO-469]: https://artsyproduct.atlassian.net/browse/GRO-469


dzucconi (approved):
👍 

I'm sort of surprised we can't filter this out with an arg to the `saleArtworksConnection`

The-Beez-Kneez:
@dzucconi that was the first approach, but I do not see a filter that allows that and could not find one after messing around with GraphiQL. 

If you got one readily known, open to it. Otherwise this does handle it.

dzucconi:
Yeah, I don't see one. The inconsistency in what various fields accept is consistently frustrating.

PR #8243, opened 2021-08-20
feat(inquiry): roughs in full inquiry structure

dzucconi:
This PR converts over the inquiry questionnaire workflow into React. The actual views, along with the conditions that drive them, are just stubs currently. So I'll merge this and then take them one view at a time with a PR dedicated to each one. I suspect there might be a few Metaphysics PRs when it comes to updating the collector profile, etc, but we'll cross that bridge when we come to it.

The "engine" code is a pretty straight port, along with the test suite. I can think of much better ways to handle that configuration now, but it's not _that_ bad for the time being. Particularly since we can prune the tree now that pre-qualification is no longer relevant.


src/v2/Apps/Debug/DebugInquiryApp.tsx
```diff
@@ -0,0 +1,18 @@
...
+  })
+
+  return (
+    <>
+      <Button onClick={showInquiry}>Open inquiry questionnaire</Button>
+
+      {inquiryQuestionnaire}
+    </>
+  )
+}
```
dzucconi, line 18: Rather than trying to deal with Relay etc in Storybook; just adding a simple route here that I'll teardown when it's done.

src/v2/Components/Inquiry/config.ts
```diff
@@ -0,0 +1,179 @@
...
+                    },
+                  ],
+                },
+              },
+            ],
+          },
+        },
+        "Done",
+      ],
+      conditions: {
```
dzucconi, line 111: These should really just be inlined with the actual steps for clarity; but I'm just leaving it alone.

src/v2/Components/Inquiry/config.ts
```diff
@@ -0,0 +1,179 @@
...
+  GalleriesYouWorkWith: InquiryGalleriesYouWorkWith,
+  HowCanWeHelp: InquiryHowCanWeHelp,
+  Inquiry: InquiryInquiryQueryRenderer,
+  InstitutionalAffiliations: InquiryInstitutionalAffiliations,
+  Specialist: InquirySpecialist,
+}
+
+type View = keyof typeof VIEWS
+
+export const useEngine = () => {
```
damassi, line 36: Would it make sense to co-locate this hook within `Engine`?

damassi (approved):
Nice job setting this work up to be incremental

laurabeth (approved):

PR #8245, opened 2021-08-23
feat(FX-3208): hide empty about section

tr-ann:
Jira: [FX-3208](https://artsyproduct.atlassian.net/browse/FX-3208)

### Description
The purpose of the ticket is hide About header if no about content present (example below)
image

### Changes
- Added condition to render About info
- Move About content to separate component
- Refactor spaces between Overview tab sections (replace borders with `<Separator />`)

### Demo
_Desktop_
image

_Mobile_
image

### DIFF
--- a/src/v2/Apps/Fair/Components/FairOverview/FairAbout.tsx
+++ b/src/v2/Apps/Fair/Components/FairOverview/FairAbout.tsx
@@ -1,11 +1,12 @@
-import React, { useMemo } from "react"
-import { Column, GridColumns, Separator, Spacer, Text } from "@artsy/palette"
-import { moment } from "desktop/lib/template_modules"
+import React from "react"
+import { Column, GridColumns, Spacer, Text } from "@artsy/palette"
 import { some } from "lodash"
+import { DateTime } from "luxon"
 import { InfoSection } from "v2/Components/InfoSection"
 import { FairTimerFragmentContainer as FairTimer } from "./FairTimer"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FairAbout_fair } from "v2/__generated__/FairAbout_fair.graphql"
+import { useCurrentTime } from "v2/Utils/Hooks/useCurrentTime"
 
 const aboutInfoTypes = [
   "about",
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Fair/Components/FairOverview/FairAbout.tsx
```diff
@@ -0,0 +1,116 @@
...
+    ticketsLink,
+    hours,
+    links,
+    contact,
+    summary,
+    tickets,
+    endAt,
+  } = fair
+
+  const hasEnded = endAt && moment().isAfter(new Date(endAt))
```
dblandin, line 40: **suggestion**: Can we take this opportunity to drop the moment import here?

As mentioned [here](https://github.com/artsy/force/pull/8169#discussion_r687913313), we're trying to move away from moment entirely. If we still need this `hasEnded` conditional, there should be a similar solution available using luxon.

But I'm wondering, even if the fair hasn't ended, should we still hide the about section if there's no content to display? I forget why we have this `hasEnded` conditional to begin with.

mdole (approved):
Nice! Looks great to me

src/v2/Apps/Fair/Components/FairOverview/FairOverview.tsx
```diff
@@ -17,72 +10,23 @@ import { FairFollowedArtistsFragmentContainer } from "./FairFollowedArtists"
...
-              <Text variant="md">Buy Tickets</Text>
-            </a>
-          )}
-          {links && <InfoSection label="Links" type="html" info={links} />}
-          {contact && (
-            <InfoSection label="Contact" type="html" info={contact} />
-          )}
-        </Column>
-      </GridColumns>
+      <FairAbout fair={fair} />
```
dblandin, line 26: **praise**: Really nice to have this extracted to a new component! 💯

src/v2/Apps/Fair/Components/FairOverview/FairAbout.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const hasAboutContent = useMemo(
+    () => some(aboutInfoTypes, field => !!fair[field]),
+    [fair]
+  )
```
dimatretyak, line 44: **suggestion**: don't use useMemo

```javascript
const hasAboutContent = some(aboutInfoTypes, field => !!fair[field])
```

dimatretyak, line 44: **suggestion**: don't use `useMemo`

```javascript
const hasAboutContent = some(aboutInfoTypes, field => !!fair[field])
```

src/v2/Apps/Fair/Components/FairOverview/FairAbout.tsx
tr-ann: If a fair has ended we shouldn't display timer. I'm not sure if we should hide empty About section in this case

dblandin: Right, hiding the timer makes sense although I think we can continue to show the about content if it's available.

tr-ann: You are right. And we dislpay about info if it`s available. In the component we return null only if fair has ended and there are not anything info about it
It checked in 47-49 lines in FairAbout component

dimatretyak (approved):
LGTM 💯

dblandin (approved):
LGTM 👍🏼

PR #8246, opened 2021-08-23
feature: Add quick navigation to search items

olerichter00:
Addresses [CX 1761]

## Description

Adds quick navigation buttons (artworks and auction results) to artist auto-suggest search items.


src/v2/Components/Search/SearchBar.tsx
```diff
@@ -266,11 +266,14 @@ export class SearchBar extends Component<Props, State> {
   )
   onSuggestionSelected({
     suggestion: {
-      node: { href, displayType },
+      node: { href, __typename },
     },
+    method,
   }) {
+    if (method === "click") return
```
olerichter00, line 273: Navigating on click will prevent the quick navigation items from working properly.

src/v2/Components/Search/SearchBar.tsx
```diff
@@ -320,6 +323,9 @@ export class SearchBar extends Component<Props, State> {
     return displayLabel
   }
 
+  getLabel = ({ displayType, __typename }) =>
+    displayType || (__typename === "Artist" ? "Artist" : null)
```
olerichter00, line 327: `displayType` doesn't exist on `Artist` results.

src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +98,45 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+  const { router } = useRouter()
+
+  const onClick = event => {
+    event.preventDefault()
+
+    router ? router.push(to) : window.location.assign(to)
+  }
+  return (
+    <Clickable display="flex" onClick={onClick} as="div">
+      <Pill variant="textSquare" mr="1">
```
olerichter00, line 137: Not sure if pill is the correct component here to use...

damassi:
Looks great 👌

src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +98,45 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+const QuickNavigationItem: React.FC<{ to: string }> = ({ children, to }) => {
+  const { router } = useRouter()
+
+  const onClick = event => {
+    event.preventDefault()
+
+    router ? router.push(to) : window.location.assign(to)
+  }
+  return (
+    <Clickable display="flex" onClick={onClick} as="div">
```
dzucconi, line 136: This sort of negates the benefit of using a `Clickable` — I realize it's invalid to nest buttons like this. Really we'd want to either adjust the hit areas so as to be distinct or handle this using `role="button" tabIndex={0} onKeyPress={...}` etc. I'd leave it alone for now since I'm re-working a bunch of this, but more of a heads up.

olerichter00: Thanks!  You're right, adjusting the hit area sounds cleaner. I used `onClick` as a workaround here to not nest buttons...

PR #8249, opened 2021-08-23
fix: login spinner handles repeat error responses

mc-jones:
This PR addresses a bug introduced in the [PR](https://github.com/artsy/force/pull/7800) that displayed the submit button spinner while the page redirect was being completed. 

Issue
The submit spinner would spin indefinitely if the same error response was received a second time (e.g., the user input the wrong username, password, or otp twice in a row). 

This was the result of the `isLoading` flag not being reset to `false` upon the second submission. This has been resolved by adding the formik `status` prop to the `useEffect` subscriptions. 

Before:
before

After:
after


anastasiapyzhik (approved):
Looks great!

PR #8254, opened 2021-08-23
feat: Add price in USD to auction results list item

olerichter00:
Addresses [CX-1707]

## Description

Adds the price in US$ to auction results if the price is available and shown in another currency.

[CX-1707]: https://artsyproduct.atlassian.net/browse/CX-1707
Screenshot 2021-08-26 at 15 08 28


src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -371,13 +384,28 @@ const renderPricing = (
...
+            <Flex alignItems="center" justifyContent="flex-end">
+              <Text variant="md" fontWeight="bold">
+                {salePrice}
+              </Text>
+              {!!showPriceUSD && (
+                <Text
+                  variant="xs"
+                  lineHeight={1}
+                  color="black60"
+                  style={{ paddingTop: "2px" }}
```
olerichter00, line 402: Couldn't find a way a better way to have `salePrice` and `salePriceUSD` be aligned correctly without applying top padding. Any ideas?

damassi: Only suggestion would be to remove the `lineHeight` prop above, as that might be throwing off the `alignItems=center` on the flex container; those values are baked into the variants and shouldn't be adjusted if possible. 

@dzucconi might have some suggestions!

damassi (approved):

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
dzucconi: Setting `lineHeight={1}` on both should work fine.

olerichter00: Thanks for the help! Adding `alignItems="flex-end"` to the parent container did the trick 🎉

olerichter00:
#squashongreen

PR #8265, opened 2021-08-24
feat(inquiry): implements basic info step

dzucconi:
Implements the "basic info" step. I'm going to follow this up with an update to Metaphysics to expand the mutation there to support these other fields (location, phone number).

https://user-images.githubusercontent.com/112297/130673517-5cfd74d9-0ae8-4549-9f98-fa946ba0f7fc.mov


src/v2/Components/LocationAutocompleteInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  })
+
+  const inputRef = useRef<HTMLInputElement | null>(null)
+  const autocompleteServiceRef = useRef<any | null>(null)
+
+  useEffect(() => {
+    if (!ready || !inputRef.current) return
+
+    // @ts-ignore
+    autocompleteServiceRef.current = new google.maps.places.Autocomplete(
```
dzucconi, line 42: I didn't hunt down the typings for this since it's loaded into the global async, blah blah. Probably a better way to handle this; but the component is simple enough.

src/v2/Components/LocationAutocompleteInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  return {
+    city: place.name,
+    state: components.state?.long_name,
+    stateCode: components.state?.short_name,
+    postalCode: components.postalCode?.long_name,
+    country: components.country?.long_name,
+    coordinates: [place.geometry.location.lat(), place.geometry.location.lng()],
+  }
+}
```
dzucconi, line 161: What we're doing in the current (old) version is taking the lat/long coordinates then passing them through an old/archived library called Geoformatter which is a port of some Ruby code. Unnecessary of course :)

dzucconi:
Woops, forgot to add the spec for this one. BRB

src/v2/Components/Inquiry/views/InquiryBasicInfo.tsx
```diff
@@ -1,19 +1,215 @@
...
+    event.preventDefault()
+
+    setMode(Mode.Loading)
+
+    try {
+      await submitUpdateMyUserProfile({ phone: state.phoneNumber })
+      setMode(Mode.Success)
+      next()
+    } catch (err) {
+      console.error(err)
```
damassi, line 80: If possible can we use `logger`? Then, in the absence of perfect sentry sourcemaps we get the file name where the error is originating: https://github.com/artsy/force/blob/master/src/v2/System/Router/buildClientApp.tsx#L36

src/v2/Components/LocationAutocompleteInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  onChange(place: Place): void
+}
+
+export const LocationAutocompleteInput: React.FC<LocationAutocompleteInputProps> = ({
+  onChange,
+  ...rest
+}) => {
+  const [ready, setReady] = useState(false)
+
+  useLoadScript({
```
damassi, line 22: Can we do a follow-up and audit force for other locations where similar functionality is being used and update? I'm thinking the recaptcha component.

dzucconi: :waits for @damassi to scroll down: 😀

src/v2/Components/Inquiry/views/InquiryBasicInfo.tsx
dzucconi: Ah yeah, I didn't even know this was a thing. Nice.

src/v2/Components/LocationAutocompleteInput.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  const components = place.address_components.reduce(
+    (
+      acc: {
+        state?: AddressComponent
+        postalCode?: AddressComponent
+        country?: AddressComponent
+      },
+      component
```
damassi, line 134: totally non-blocking - i know there's the `address_components` array from google, but maybe we can use a less overloaded term here. Devs might see `component.foo` below and think we're doing some funny business with react

src/v2/Utils/Hooks/useAppendStylesheet.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    const css = document.createElement("style")
+
+    css.id = id
+    css.setAttribute("type", "text/css")
+    css.textContent = body
+
+    document.body.appendChild(css)
+  }, [body, id])
+}
```
damassi, line 22: I think recaptcha might also be appending a stylesheet, so that can be updated as well

src/v2/Utils/ReCaptchaContainer.tsx
```diff
@@ -1,29 +1,18 @@
...
-      return
-    }
-    const css = document.createElement("style")
-    css.setAttribute("type", "text/css")
-    css.textContent = ".grecaptcha-badge { visibility: hidden; }"
-    document.body.appendChild(css)
+  useLoadScript({
+    id: "google-recaptcha-js",
+    src: `https://www.google.com/recaptcha/api.js?render=${sd.RECAPTCHA_KEY}`,
+  })
```
damassi, line 15: Oh lol @dzucconi - you beat me to it 😄

damassi: Bonus points 💯

src/v2/Components/LocationAutocompleteInput.tsx
damassi: hah

src/v2/Components/LocationAutocompleteInput.tsx
dzucconi: Yeah, I had the same though. 👍

damassi (approved):
Lookin' good 👍 Merge once the spec is in.

PR #8267, opened 2021-08-24
chore(stories): adds story for handling sticky within app container

dzucconi:
Re: https://artsyproduct.atlassian.net/browse/FX-3118

So just adding an example here since it's a bit confusing. We should revisit this, but for now here's how you'd do it.

I wrote this in the story below but copied here for visibility:

This is a bit verbose but we currently don't match <code>position: sticky;</code> exactly. One of the side effects of this is that content when stuck will break out of the container completely. At the moment this is desirable though we may revisit in the future. In the meantime simply break out of the container with the <code>FullBleed</code> component, then wrap it in our <code>AppContainer</code>, which controls max-width; and <code>HorizontalPadding</code> which controls the page's padding.


package.json
```diff
@@ -264,10 +264,10 @@
...
     "@pmmmwh/react-refresh-webpack-plugin": "0.4.3",
     "@sentry/cli": "1.41.2",
-    "@storybook/addon-actions": "6.2.0-beta.11",
-    "@storybook/addon-essentials": "6.2.0-beta.11",
-    "@storybook/addon-links": "6.2.0-beta.11",
-    "@storybook/react": "6.2.0-beta.11",
+    "@storybook/addon-actions": "6.3.7",
+    "@storybook/addon-essentials": "6.3.7",
+    "@storybook/addon-links": "6.3.7",
+    "@storybook/react": "6.3.7",
```
dzucconi, line 270: Unrelated but story reloading was being weird; just ran the upgrade and seems to have fixed.

dzucconi:
Long story but we don't use `position: sticky` for the time being because of problems with how it interacts with `overflow: hidden` containers. As a result we have a custom sticky implementation. We'll probably move that implementation closer to match the spec over time though. I've got some updates to this queued up soon-ish. 🤷

PR #8290, opened 2021-08-27
feat(viewing room): redirect viewing room route /works to /artworks

anastasiapyzhik:
[FX-3200]

The URL structures of viewing room works tab were not consistent on Eigen and Force:

This worked in Force but not Eigen: https://www.artsy.net/viewing-room/pearl-lam-galleries-mr-doodle-the-pop-heart-collection/works

And this worked in Eigen but not Force: https://www.artsy.net/viewing-room/pearl-lam-galleries-mr-doodle-the-pop-heart-collection/artworks

This PR changes url structure for viewing room works tab in Force(`:id/works` → `:id/artworks`)  so that this page will be accessible everywhere via the same url

Demo

artworks

Redirect

redirect

### DIFF
--- a/src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
+++ b/src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
@@ -1,4 +1,5 @@
 import loadable from "@loadable/component"
+import { RedirectException } from "found"
 import { graphql } from "react-relay"
 import { AppRouteConfig } from "v2/System/Router/Route"
 
@@ -92,6 +93,12 @@ export const viewingRoomRoutes: AppRouteConfig[] = [
       },
       {
         path: "works",
+        render: () => {
+          throw new RedirectException("artworks", 301)
+        },
+      },
+      {
+        path: "artworks",
         Component: WorksRoute,
         ignoreScrollBehavior: true,
         query: graphql`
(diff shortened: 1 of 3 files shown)


src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
```diff
@@ -92,6 +93,12 @@ export const viewingRoomRoutes: AppRouteConfig[] = [
       },
       {
         path: "works",
+        render: () => {
+          throw new RedirectException("artworks", 301)
+        },
+      },
+      {
+        path: "artworks",
         Component: WorksRoute,
```
dblandin, line 102: **question**: Are we able to remove some of the other properties/queries on this route if we're always responding with a 301 redirect?

dblandin (commented):
Looks good! Added a question about the /works route redirection.

src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
anastasiapyzhik: @dblandin I’m not sure I follow this question, there is no other properties for `/works` route
https://github.com/artsy/force/blob/9db09af6e8c4d4676b28f1afb589cbb178a8d6af/src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx#L94-L99

dblandin: Oh sorry, you're totally right. I was looking down at the next block and thinking it was part of `/works`. Never mind!

dblandin (approved):

PR #8293, opened 2021-08-27
chore: add admin flag to show VR in PPP

ansor4:
Let's put the viewing room tab behind an admin flag so we can properly QA and such!

cc @Serge0n


src/v2/Apps/Partner/Components/NavigationTabs/index.tsx
```diff
@@ -8,6 +8,8 @@ import { Sticky } from "v2/Components/Sticky"
 import { ScrollIntoView } from "v2/Utils"
 import { Media } from "v2/Utils/Responsive"
 import { NavigationTabs_partner } from "v2/__generated__/NavigationTabs_partner.graphql"
+import { useSystemContext } from "v2/System"
+import { userIsAdmin } from "v2/Utils/user"
```
sweir27, line 12: In general, for internal testing, it's better to use [userIsTeam](https://github.com/artsy/force/blob/master/src/v2/Utils/user.ts#L42) as we are stricter now about who has the `admin` role!

PR #8305, opened 2021-08-31
feat(FX-3237): Redirect to active fair page

anastasiapyzhik:
[FX-3237]

When a user lands on a fair organizer route and there’s an active fair for that organizer then the user should be automatically directed to the fair page

**Demo**

With active fair:

act-fair

Without active fair:

no-a-fair

[FX-3237]: https://artsyproduct.atlassian.net/browse/FX-3237

### DIFF
--- a/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
+++ b/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
@@ -61,7 +61,7 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
           profile {
             __typename
           }
-          runningFairs: fairsConnection(first: 1, status: RUNNING) {
+          runningFairs: fairsConnection(first: 1, status: RUNNING, sort: START_AT_DESC) {
             edges {
               node {
                 href
--- a/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
+++ b/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
@@ -1,5 +1,4 @@
 import React from "react"
-import { DateTime } from "luxon"
 import loadable from "@loadable/component"
 import { graphql } from "react-relay"
 import { ErrorPage } from "v2/Components/ErrorPage"
@@ -37,24 +36,16 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
     render: ({ Component, props }) => {
       if (Component && props) {
         const { fairOrganizer } = props as any
-        const { profile, fairsConnection } = fairOrganizer
+        const { profile, runningFairs } = fairOrganizer
 
         if (!profile) {
           return <ErrorPage code={404} />
         }
 
-        const fair = extractNodes(fairsConnection)[0]
-        if (fair) {
-          const { startAt, endAt, href } = fair as any
-          const currentTime = DateTime.fromJSDate(new Date())
-          const startTime = DateTime.fromISO(startAt!)
-          const endTime = DateTime.fromISO(endAt!)
-
-          const isFairActive = currentTime >= startTime && currentTime < endTime
-
-          if (isFairActive) {
-            throw new RedirectException(href, 302)
-          }
+        const activeFair = extractNodes(runningFairs)[0]
+        if (activeFair) {
+          const { href } = activeFair as any
+          throw new RedirectException(href, 302)
         }
 
         return <Component {...props} />
@@ -66,11 +57,9 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
           profile {
             __typename
           }
-          fairsConnection(first: 1, sort: START_AT_DESC) {
+          runningFairs: fairsConnection(first: 1, status: RUNNING) {
             edges {
               node {
-                startAt
-                endAt
                 href
               }
             }
--- a/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
+++ b/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
@@ -61,7 +61,11 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
           profile {
             __typename
           }
-          runningFairs: fairsConnection(first: 1, status: RUNNING, sort: START_AT_DESC) {
+          runningFairs: fairsConnection(
+            first: 1
+            status: RUNNING
+            sort: START_AT_DESC
+          ) {
             edges {
               node {
                 href
--- a/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
+++ b/src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
@@ -3,8 +3,13 @@ import loadable from "@loadable/component"
 import { graphql } from "react-relay"
 import { ErrorPage } from "v2/Components/ErrorPage"
 import { AppRouteConfig } from "v2/System/Router/Route"
-import { RedirectException } from "found"
+import { RedirectException, RenderProps } from "found"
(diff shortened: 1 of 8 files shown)


src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -47,6 +57,13 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
           profile {
             __typename
           }
+          runningFairs: fairsConnection(first: 1, status: RUNNING) {
```
dblandin, line 60: **suggestion**: This might not ever happen, but, if an organizer happens to have two concurrently running fairs, it might be nice to have an explicit sort defined here.

```suggestion
          runningFairs: fairsConnection(first: 1, status: RUNNING, sort: START_AT_DESC) {
```

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -34,10 +36,18 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
...
-        const { profile } = fairOrganizer
+        const { profile, runningFairs } = fairOrganizer
+
         if (!profile) {
           return <ErrorPage code={404} />
         }
+
+        const activeFair = extractNodes(runningFairs)[0]
+        if (activeFair) {
+          const { href } = activeFair as any
```
dblandin, line 47: **question**: I'm surprised to see `as any` here. Is `activeFair` not already typed?

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
anastasiapyzhik: @dblandin that's a good point
But maybe in case of several concurrently running fairs it'd be better to display all of them? Otherwise they won't be displayed anywhere on the page

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
anastasiapyzhik: I haven't found any examples of typing props in routes so followed existing practice `props as any`. So `extractNodes(runningFairs)` _(which is of type `any` too)_ was returning `object[]`
But if we want to get props values typed we can extend props type with query response type

src/v2/Apps/FairOrginizer/fairOrganizerRoutes.tsx
```diff
@@ -33,11 +40,18 @@ export const fairOrganizerRoutes: AppRouteConfig[] = [
...
+        const { fairOrganizer } = props as Props
+        const { profile, runningFairs } = fairOrganizer!
         if (!profile) {
           return <ErrorPage code={404} />
         }
+
+        const activeFair = extractNodes(runningFairs)[0]
+        if (activeFair) {
+          const { href } = activeFair
+          throw new RedirectException(href!, 302)
```
dblandin, line 52: Looks like the Circle CI smoke test got caught in a redirect loop. If `href` is ever null / empty, we might want to guard for that here. Wondering if there's something else going on... 🤔

anastasiapyzhik: We are [testing art-paris organizer](https://github.com/artsy/force/blob/cafd8984c52563bc7f82411b458323fa8e5ac544/cypress/integration/fairOrganizer.spec.ts#L5) which has running fair currently and this fair is only accessable to admins so we get `404 Not Found` error

dblandin: Ahh, I see. I believe this would introduce some bad behavior if not addressed. We don't want a user to be redirected to a fair page they can't see.

I see [in the GraphQL schema](https://github.com/artsy/metaphysics/blob/a7d0d1df9beb965574a53f229758fb19237d9084/src/schema/v2/fair.ts#L197-L200) that we _could_ query for `isPublished` and only redirect if `isPublished=true` but then I wondered why we are returning fairs that a user (in this case the Cypress user) cannot navigate to and it looks like we could also add some additional visibility controls to the backend API endpoint [here](https://github.com/artsy/gravity/blob/master/app/api/v1/fairs_endpoint.rb#L170-L174) 🔒 .

Seems like that admin visibility consideration is usually in place for fairs but this `fair_organizer_id` param branch gets around it.

dblandin: So I think the next steps here are to:

- Make that change in artsy/gravity, adding the same checks to further scope the visibility of `fairs`
- Use the same art-paris example here and re-run CI to make sure it passes

anastasiapyzhik: @dblandin so gravity shouldn't return fairs user doesn't have access to and they simply shouldn't be visible on fair-organizer page?

dblandin: @dblandin I think I might have misread this issue a bit last week. It looks like the API endpoint in gravity actually _does_ chain the additional `organizer` clause onto the previous query which includes the visibility scoping.

I now think the difference is in how we're currently not considering the profile's visibility. You can see the previous handling of the profile visibility in the old fair organizer app [here](https://github.com/artsy/force/blob/master/src/desktop/apps/fair_organizer/routes.coffee#L52-L54).

So I think we might need additional check to see if `activeFair.profile` is present. If so, we can redirect to the fair, otherwise skip the redirection.

anastasiapyzhik: @dblandin should we also add this check when rendering fairs or when redirecting only?

dblandin: It looks like we have some version of that check already in the new fair app [here](https://github.com/artsy/force/blob/master/src/v2/Apps/Fair/FairApp.tsx#L164-L169).

anastasiapyzhik: I mean, maybe it'd be better not to render such fairs on fair organizer page so that users don't get the ability to get to a broken page?

anastasiapyzhik: Or at least not to provide such fairs with links?

dblandin: Oh, I see what you mean - when rendering fairs on the fair organizer page. That might be rare at the moment for past fairs. Let's leave the rendering as is and I can follow up with the fairs team on this.

anastasiapyzhik: @dblandin even though we prevented redirect to unaccessible fairs, redirect to accessible ones has been remained and currently CI fails because the page user is redirected to displays different titles. As redirect is temporary, tests for `/fair-organizer/:slug` are unstable.
I would either replace the `art-paris` slug to some old organizer's one or temporarily remove current test (and possibly add redirect tests) and document the follow-up work on updating it once the fair is not active
What do you think? Is there a better solution?

dblandin: I think a redirect leading to a 200 is still considered successful. Can we loosen the text assertions to match against either the fair organizer page or the fair page?

```ts
visitWithStatusRetries("fair-organizer/art-paris")
    cy.contains("Art Paris").should("exist")
    cy.title().should("include", "Art Paris")
```

anastasiapyzhik: Sounds good to me! Thank you!

dblandin (approved):
LGTM!

PR #8307, opened 2021-08-31
fix(PX-4489): Viewing rooms: View works: 404 error page is opened

Serge0n:
Jira Ticket:ticket: : [PX-4489](https://artsyproduct.atlassian.net/browse/PX-4489)
Discussion:scroll: : https://artsy.slack.com/archives/C9ZJYFDNF/p1630397677128800

This PR fixing the 404 error when clicking on the `View Works` button.

**VIDEO:clapper: :**

https://user-images.githubusercontent.com/55637696/131490371-c8528fa0-0346-4b37-a8d9-44aa960f716a.mov

### DIFF
--- a/src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
+++ b/src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
@@ -94,8 +94,12 @@ export const viewingRoomRoutes: AppRouteConfig[] = [
       },
       {
         path: "works",
-        render: () => {
-          throw new RedirectException("artworks", 301)
+        render: ({
+          match: {
+            params: { slug },
+          },
+        }) => {
+          throw new RedirectException(`/viewing-room/${slug}/artworks`, 301)
         },
       },
       {


src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
```diff
@@ -94,8 +94,12 @@ export const viewingRoomRoutes: AppRouteConfig[] = [
...
       {
         path: "works",
-        render: () => {
-          throw new RedirectException("artworks", 301)
+        render: ({
+          match: {
+            params: { slug },
+          },
+        }) => {
+          throw new RedirectException(`/viewing-room/${slug}/artworks`, 301)
```
mdole, line 102: thanks for looking into this @Serge0n! I'm a bit confused here: won't this always throw an exception? I think it will render OK since it returns `null` which I believe is expected when there are no children ([here](https://github.com/4Catalyzer/found/blob/master/README.md#render)), but it would still 301 every time this link is clicked.

301 also seems like a surprising code to throw, as we are not redirecting to a page whose route has changed. Can you help me understand what's happening here?

Sorry if I'm missing something! I'm new to routing with `found`, so please correct me if I'm wrong.

mdole: OH I was just pointed to [this PR](https://github.com/artsy/force/pull/8290/files) that actually did change the route, from /works to /artworks. So that makes sense! I think I understand now

mdole (approved):

src/v2/Apps/ViewingRoom/viewingRoomRoutes.tsx
dblandin: Throwing an error should halt the execution of the render function right away. I believe the found routing layer will rescue the `RedirectException` error and instruct the browser to redirect to the given path. We're also specifying that this should be a permanent redirect with the `301` status code.

Seems like the error we introduced was using a partial relative path instead of the full `/viewing-room/:slug/artworks` path.

PR #8308, opened 2021-08-31
fix(fair organizer): sort fairs by -start_at to determine upcoming show

dblandin:
The default fair sort in Gravity is -created_at. When looking for an upcoming fair for an organizer, we currently fetch the first fair and check if the start time is in the future. If a past fair is created after a future fair, we might be checking the wrong entity in the system.

This updates the GraphQL query to sort fair results by `-start_at` to mimic the behavior in the previous implementation [here][commit].

Related to https://github.com/artsy/metaphysics/pull/3377

Jira 🔒 : [FX-3243]

[commit]: https://github.com/artsy/force/blob/4de69dfc0d2d26cbe43b7e5ffefc0a88e44333ac/src/desktop/apps/fair_organizer/routes.coffee#L36-L41

[FX-3243]: https://artsyproduct.atlassian.net/browse/FX-3243

### Before

Screen Shot 2021-08-31 at 12 57 47 PM

### After

Screen Shot 2021-08-31 at 1 09 15 PM

### DIFF
--- a/src/v2/Apps/FairOrginizer/Components/DedicatedArticlesBreadcrumbs.tsx
+++ b/src/v2/Apps/FairOrginizer/Components/DedicatedArticlesBreadcrumbs.tsx
@@ -47,7 +47,7 @@ export const DedicatedArticlesBreadcrumbsFragmentContainer = createFragmentConta
       fragment DedicatedArticlesBreadcrumbs_fairOrganizer on FairOrganizer {
         slug
         name
-        fairsConnection(first: 1) {
+        fairsConnection(first: 1, sort: START_AT_DESC) {
           edges {
             node {
               image {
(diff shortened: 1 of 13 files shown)


dimatretyak (approved):
Nice 🚀

dimatretyak:
Perhaps it is also necessary to pass the `sort` param when [getting the header background image](https://github.com/artsy/force/blob/dblandin/update-organizer-running-show/src/v2/Apps/FairOrginizer/FairOrganizerApp.tsx#L52)?

dblandin:
> Perhaps it is also necessary to pass the sort param when getting the header background image?

Good call. But I wonder why aren't using the fair organizer's image for the header and articles breadcrumb link...

PR #8309, opened 2021-08-31
fix: Always clear search term after navigating away from search results

olerichter00:
Follows up on [CX-1761]

https://github.com/artsy/force/pull/8246 introduced a bug that prevented the `SearchBar` from clearing the search term after clicking on a suggestion item.

This fix ensures the search term gets cleared every time the user clicks on a suggestion item.

**Before:**

https://user-images.githubusercontent.com/4691889/131504182-5eec1cc6-6c6c-402d-93d6-a3413e1d4bf9.mov

**After:**

https://user-images.githubusercontent.com/4691889/131504515-13faa211-d019-4418-a7b1-e0329dad1911.mov

[CX-1761]: https://artsyproduct.atlassian.net/browse/CX-1761

### DIFF
--- a/src/v2/Components/Search/SearchBar.tsx
+++ b/src/v2/Components/Search/SearchBar.tsx
@@ -166,15 +166,19 @@ export class SearchBar extends Component<Props, State> {
     // Clear the search term once you navigate away from search results
     this.removeNavigationListener = this.props.router
       ? this.props.router.addNavigationListener(location => {
-          if (!location.pathname.startsWith("/search")) {
-            this.setState({ term: "" })
-          }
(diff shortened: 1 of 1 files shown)


src/v2/Components/Search/SearchBar.tsx
```diff
@@ -270,10 +274,11 @@ export class SearchBar extends Component<Props, State> {
     },
     method,
   }) {
-    if (method === "click") return
-
+    this.clearSearchTerm()
```
olerichter00, line 277: Clearing the search term in `onSuggestionSelected`.

PR #8311, opened 2021-08-31
fix(FX-3236): update timers on the fair page (remove duplicate, show "closed" state)

dimatretyak:

This PR resolves [FX-3236]

### Description
* As a user
* When I navigate to the fair page
* I should only see the timer in the overview tab

### Acceptance criteria
* The timer from title bar should be removed
* When the fair has passed, the timer should display the "closed" label in the overview tab

[FX-3236]: https://artsyproduct.atlassian.net/browse/FX-3236


src/v2/Apps/Fair/Components/FairOverview/FairTimer.tsx
```diff
@@ -3,20 +3,25 @@ import { Box, Text } from "@artsy/palette"
 import { Timer } from "v2/Components/Timer"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FairTimer_fair } from "v2/__generated__/FairTimer_fair.graphql"
+import { useTimer } from "v2/Utils/Hooks/useTimer"
 
 interface FairTimerProps {
   fair: FairTimer_fair
 }
 
 export const FairTimer: React.FC<FairTimerProps> = ({ fair: { endAt } }) => {
```
dimatretyak, line 12: is there any cases when `endAt` is `null`?

dblandin: I don't believe so. We have some presence validations in Gravity [here](https://github.com/artsy/gravity/blob/master/app/models/domain/fair.rb#L135).

dblandin (approved):
Looks good to me! Might want to ping @thomasmeerschwam or Jonathan when this up for review in staging. Since we're not guarding this behind a feature flag, it'll go out to production with the next deploy.

dimatretyak:
> Looks good to me! Might want to ping @thomasmeerschwam or Jonathan when this up for review in staging. Since we're not guarding this behind a feature flag, it'll go out to production with the next deploy.

I agreed with Jonathan on the font size of the" closed " label and sent a screenshot. But I will ask him to check again when the changes will be available on staging

thomasmeerschwam:
@dblandin @dimatretyak in future, for any user-facing changes on web/mweb, would it be possible for either Jonathan or myself to review _before_ we merge the PR to staging? As you know, this is because once it’s in staging and not behind a flag, it will get deployed to production in a cadence that is beyond our control.

dblandin:
> In future, for any user-facing changes on web/mweb, would it be possible for either Jonathan or myself to review before we merge the PR to staging? As you know, this is because once it’s in staging and not behind a flag, it will get deployed to production in a cadence that is beyond our control.

@thomasmeerschwam We should chat more about how strict we want to be here. I know you and @georgeclarke23 are working on a QA guidelines document that should bring some more clarity around what type of QA process we need for various user-facing changes.

For this example, the PR was merged at 4:20pm GMT+2 and checked by Jonathan just a few minutes later in staging. That seems totally reasonable to me.

We have the option to QA against a local dev instance over Zoom or spin up a review app, but those approaches take more effort+time and I don't think we need that level of scrutiny for every user-facing change.

thomasmeerschwam:
@dblandin +1 to your points. I agree that having one of us simply checking staging soon after after a merge would work efficiently, we'd just need to make sure that it happens well before prod deploy so we can catch any potential errors.

PR #8316, opened 2021-08-31
feat: get past events from fair organizer

dimatretyak:

### Description
After the changes to metaphysics ([PR #3377](https://github.com/artsy/metaphysics/pull/3377)), we can now get a list of past fairs from `fairOrganizer`


dblandin (approved):
LGTM!

PR #8321, opened 2021-08-31
fix(home): fixes relay type error; refactors to be simpler

dzucconi:
This PR silences the warning regarding masking and overall refactors it into a slightly simpler component structure.


src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnits.tsx
```diff
@@ -1,23 +1,64 @@
...
+  const { isLoggedIn } = useSystemContext()
+
+  const heroUnits = [
+    ...(isLoggedIn ? [] : [LOGGED_OUT_HERO_UNIT]),
+    ...compact(homePage.heroUnits),
+  ]
+
+  const rendered = heroUnits.map((heroUnit, i) => {
+    // If logged out, the first hero unit is static and
+    // needs to be rendered without the fragment container
```
dzucconi, line 30: Crux of the issue, this

damassi (approved):
Thanks for the quick fix 👍

PR #8325, opened 2021-09-01
feat(PURCHASE-2895): Remove back to artsy button

ivan-badyulya:
**JIRA** -> [Remove Back to Artsy link from inquiry checkout counteroffer modal](https://artsyproduct.atlassian.net/jira/software/c/projects/PURCHASE/boards/41?modal=detail&selectedIssue=PURCHASE-2895&assignee=60469b1e43eac6006f636d07)

**What I've done**:
- removed the `go back to artsy` button from the modal offer status, when buyer/seller declines a counteroffer
- refactored code
- added isModal checker due this behavior won't just be for declined statuses, the CTA opens the page modally whenever the partner takes any action

**GIF**
ezgif com-gif-maker

p.s. sorry for the fast gif... just wanted to show that we don't have a button for declined offers from now :)

### DIFF
--- a/src/v2/Apps/Order/Routes/Status/index.tsx
+++ b/src/v2/Apps/Order/Routes/Status/index.tsx
@@ -12,6 +12,7 @@ import {
   media,
 } from "@artsy/palette"
 import styled from "styled-components"
+import { RouterLink } from "v2/System/Router/RouterLink"
 import { TransactionDetailsSummaryItemFragmentContainer as TransactionDetailsSummaryItem } from "v2/Apps/Order/Components/TransactionDetailsSummaryItem"
 import { TwoColumnLayout } from "v2/Apps/Order/Components/TwoColumnLayout"
 import { get } from "v2/Utils/get"
@@ -174,7 +175,7 @@ export class StatusRoute extends Component<StatusProps> {
         return {
           title: "Offer expired",
           description: (
-            <>The seller’s offer expired because you didn’t respond iFn time.</>
+            <>The seller’s offer expired because you didn’t respond in time.</>
           ),
           showTransactionSummary: false,
         }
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Order/Routes/Status/index.tsx
```diff
@@ -233,9 +234,41 @@ export class StatusRoute extends Component<StatusProps> {
...
+
+    if (isModal || isDeclined) {
+      return null
+    }
+
+    return (
+      <Button
+        onClick={() => {
+          window.location.href = "/"
+        }}
```
dzucconi, line 261:
```tsx
// @ts-ignore
as={RouterLink}
to="/"
```

The types for polymorphic `as` are weird so that's what's necessitating the `@ts-ignore`.

In general you never want to do a link like this since it breaks shift+cmd click, etc. (You can of course handle that if necessary but there's no need here.)

src/v2/Apps/Order/Routes/Status/index.tsx
```diff
@@ -173,7 +174,7 @@ export class StatusRoute extends Component<StatusProps> {
         return {
           title: "Offer expired",
           description: (
-            <>The seller’s offer expired because you didn’t respond in time.</>
+            <>The seller’s offer expired because you didn’t respond iFn time.</>
```
dzucconi, line 177: Typo

src/v2/Apps/Order/Routes/Status/index.tsx
dzucconi: (I see this is just moved from below but might as well fix it anyway.)

ivan-badyulya:
@dzucconi hey, I've fixed the things you commented above, could you check it again?

dzucconi (approved):

PR #8330, opened 2021-09-01
fix(loadenv): prevent modifying already set env vars

ovasdi:
This CL adds a condition to `lib/loadenv` and `utils/env` that prevents overwriting existing env vars that are already set. This matches the same rule that `dotenv` package adheres to: [never modify any environment variables that have already been set](https://github.com/motdotla/dotenv#what-happens-to-environment-variables-that-were-already-set). The `package.json` scripts that export `NODE_ENV=production` will now behave as expected even if a contributor has `NODE_ENV=development` in `.env`.

Additionally:
- updated `loadenv` to match the implementation in `utils/env` as I find it better
- moved the `loadenv` imports to the top of the files

### DIFF
--- a/src/dev.js
+++ b/src/dev.js
@@ -1,11 +1,11 @@
 // @ts-check
 
+require("./lib/loadenv")
 require("coffeescript/register")
 require("@babel/register")({
   extensions: [".ts", ".js", ".tsx", ".jsx"],
   plugins: ["babel-plugin-dynamic-import-node"],
 })
-require("./lib/loadenv")
 
 const express = require("express")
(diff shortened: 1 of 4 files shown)


src/dev.js
```diff
@@ -1,11 +1,11 @@
 // @ts-check
 
+require("./lib/loadenv")
```
icirellik, line 3: Preferably this should be after we register the transpilers. However, since `loadenv` is vanilla JS there shouldn't be any issues as a result of putting it here.

icirellik (approved):

PR #8337, opened 2021-09-01
lib: Upgrade relay to 10

damassi:
This was a fun one 😓 

Learnings: 
- Relay test auto fixtures no longer auto-fixture booleans. Had to update a lot of tests
- Sometime one needs to add the @relay_test_operation directive to the query in the test. I'm not sure exactly when. 

---- 

When attempting to migrate the `AboveTheFoldQueryRenderer` over from Eigen I noticed that we were behind a version in relay, blocking the implementation. This upgrades things so that we can proceed. 

Note: This upgrade reminded me that when performing relay upgrades there's often changes to how `relay-typescript-compiler` outputs types, requiring us to manually update many of our old-school fixtures in order to get type-checking to pass. This wouldn't be a problem if we were using modern testing practices that dynamically generate fixture data. The most critical of these is `Apps/Order`, which we should make an effort to update.


damassi:
Tons of tests are failing and will need to return to this when I get back from OOO.

pepopowitz:
Nice -- I'm going to try isolating a component that has nothing but that isClosed aliased field and see if the payload generated for that contains the dupes.

damassi:
This madness is 🍏 , gonna merge

pepopowitz:
Epic work @damassi 💪 🎉 🕺

PR #8338, opened 2021-09-02
fix: stop fair organizer page from crashing with no articles

mdole:
A newly-created Fair Organizer was failing to load on production with 500 errors. After investigating, @dblandin and I figured out that the problem was that we assumed that all fair organizers had associated articles. 

Slack thread: https://artsy.slack.com/archives/C9SATFLUU/p1630573114272300


dblandin (approved):
LGTM! 👍🏼

anastasiapyzhik:
Very nice change! Thank you 🙏🏼

PR #8339, opened 2021-09-02
feat(FX-3249): updated exhibitors tab style

tr-ann:
Jira: [FX-3249](https://artsyproduct.atlassian.net/browse/FX-3249)

### Description
The exhibitors listed in the A-Z tab should be formatted as per the latest 8/23 “Work in Progress” figma file. In particular we shouldn't display exhibitors image.

[Figma spec](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3904%3A8029)

### Changes
- Removed image from exhibitor card
- Wrap text when overflow
- Updated exhibitors placeholder

### Demo
image

### DIFF
--- a/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorCard.tsx
+++ b/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorCard.tsx
@@ -1,12 +1,5 @@
 import React from "react"
-import {
-  Box,
-  Flex,
-  ResponsiveBox,
-  Image,
-  Text,
-  BorderBox,
-} from "@artsy/palette"
+import { Box, Flex, Image, Text, BorderBox } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useAnalyticsContext, useSystemContext, useTracking } from "v2/System"
 import {
@@ -17,7 +10,6 @@ import {
   PageOwnerType,
 } from "@artsy/cohesion"
 import { FairExhibitorCard_partner } from "v2/__generated__/FairExhibitorCard_partner.graphql"
-import { Media } from "v2/Utils/Responsive"
 import { FollowProfileButtonFragmentContainer as FollowProfileButton } from "v2/Components/FollowButton/FollowProfileButton"
 import { RouterLink } from "v2/System/Router/RouterLink"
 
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorCard.tsx
```diff
@@ -62,75 +54,51 @@ export const FairExhibitorCard: React.FC<FairExhibitorCardProps> = ({
...
+            />
+          )}
+        </BorderBox>
+        <Box>
+          <Text variant="md" overflow="clip">
+            {name}
+          </Text>
+          {cities?.length ? (
+            <Text variant="md" color="black60" overflow="clip">
+              {partnerAddress([...cities])}
```
dblandin, line 81: **question**: Does `partnerAddress([...cities])` have a different effect than `partnerAddress(cities)`?

dblandin (approved):
Looks good! Left one minor style question.

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorCard.tsx
tr-ann: Yes, because `cities` are read only, so to avoid linter errors we need to create new array

dblandin: Ahh, that makes sense. I see that linter error now:

Screen Shot 2021-09-02 at 1 45 54 PM

In that case, would it be better to update the argument types to also expect a readonly array?

```typescript
const partnerAddress = (cities: readonly (string | null)[]) => {
  const visibleCities = cities?.slice(0, VISIBLE_CITIES_NUM).join(", ")

  if (cities.length > VISIBLE_CITIES_NUM) {
    return `${visibleCities}, +${cities.length - VISIBLE_CITIES_NUM} more`
  }

  return visibleCities
}
```

It looks like the `partnerAddress` already creates a copy right away with `#slice` and doesn't mutate the array afterwards.

dimatretyak (approved):
LGTM 👍

tr-ann:
@dblandin I fixed tests and all checks are 🍏 . Can we merge it?

PR #8340, opened 2021-09-02
fix(artist): fixes consign button link

dzucconi:
Looks like this got un-nested in the v3 conversion.

### DIFF
--- a/src/v2/Apps/Artist/Routes/Overview/Components/ArtistConsignButton.tsx
+++ b/src/v2/Apps/Artist/Routes/Overview/Components/ArtistConsignButton.tsx
@@ -80,11 +80,24 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = ({
       </RouterLink>
 
       <Media greaterThanOrEqual="md">
-        <Button variant="secondaryGray">Get started</Button>
+        <Button
+          variant="secondaryGray"
+          // @ts-ignore
+          as={RouterLink}
+          to={consignURL}
+        >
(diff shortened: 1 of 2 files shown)


mdole (approved):
👍 #mergeongreen

src/v2/Apps/Artist/Routes/Overview/Components/ArtistConsignButton.tsx
```diff
@@ -80,11 +80,24 @@ export const ArtistConsignButton: React.FC<ArtistConsignButtonProps> = ({
       </RouterLink>
 
       <Media greaterThanOrEqual="md">
-        <Button variant="secondaryGray">Get started</Button>
+        <Button
+          variant="secondaryGray"
+          // @ts-ignore
+          as={RouterLink}
```
damassi, line 86: @dzucconi - can we type `as` in Palette as Any and then just forward that down with a ts-ignore?

dzucconi: Yeah; `Button` is an easy target; you can't exactly just type `as` since it wouldn't ignore subsequent props like `to`

I'll do something like this: https://github.com/artsy/palette/blob/56f21896941be8c28e63a5c17566fbd610cf0e4d/packages/palette/src/elements/BaseTabs/BaseTab.tsx#L16-L23

MounirDhahri:
Thank you for fixing this Damon

PR #8341, opened 2021-09-02
fix: Increase artist auction results default end date

olerichter00:
Addresses [CX-1874]

Increases artist auction result's default createdBeforeYear value to fix a bug where recent auction results were filtered out on initial load.

[CX-1874]: https://artsyproduct.atlassian.net/browse/CX-1874


MounirDhahri (approved):

MounirDhahri:
🛳🛳🛳

PR #8355, opened 2021-09-03
(PLATFORM-3537) feat: login form supports on-demand otp

mc-jones:
This PR enables Force to handle on-demand otp related errors, which were enabled in Gravity in this [PR](https://github.com/artsy/gravity/pull/14395/files).

When we are ready to enable this functionality we will set `enforces_on_demand_otp` to `true` on the `ClientApplicaiton`. This will be tested out in staging before being set on production.

When an on-demand otp is required from the User, they will see the following prompt:
Screen Shot 2021-09-03 at 3 10 34 PM

### DIFF
--- a/src/v2/Components/Authentication/Desktop/LoginForm.tsx
+++ b/src/v2/Components/Authentication/Desktop/LoginForm.tsx
@@ -24,8 +24,8 @@ interface ConditionalOtpInputProps {
 }
 
 const ConditionalOtpInput: React.FC<ConditionalOtpInputProps> = props => {
-  const [showOtpInput, setShowOtpInput] = useState(false)
-  const [showOtpOnDemandPrompt, setShowOtpOnDemandPrompt] = useState(false)
+  const [showOtpInput, setShowOtpInput] = useState(true)
+  const [showOtpOnDemandPrompt, setShowOtpOnDemandPrompt] = useState(true)
   const {
     errors,
     values,
@@ -48,7 +48,7 @@ const ConditionalOtpInput: React.FC<ConditionalOtpInputProps> = props => {
     <>
       {showOtpOnDemandPrompt && (
         <Message>
-          This login requires additional authroization. Please check your email
+          This login requires additional authorization. Please check your email
           for a one-time authentication code.
         </Message>
       )}
--- a/src/v2/Components/Authentication/Desktop/LoginForm.tsx
+++ b/src/v2/Components/Authentication/Desktop/LoginForm.tsx
@@ -24,8 +24,8 @@ interface ConditionalOtpInputProps {
 }
 
 const ConditionalOtpInput: React.FC<ConditionalOtpInputProps> = props => {
-  const [showOtpInput, setShowOtpInput] = useState(true)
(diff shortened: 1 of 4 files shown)


src/v2/Components/Authentication/Desktop/LoginForm.tsx
```diff
@@ -80,8 +98,11 @@ export class LoginForm extends Component<FormProps, LoginFormState> {
         isLoading: true,
       },
       () => {
-        values.otpRequired = this.state.otpRequired
-        this.props.handleSubmit?.(values, formikBag)
+        const submitValues: SubmitValues = {
```
mc-jones, line 101: @damassi - Minor, but wanted to revet a change I had made to our previous work on this. Felt it safer to not mutate the values object directly without a better understanding of how this is working under the hood in formik. This now matches our original approach.

damassi (approved):
Sorry for not getting to this sooner as I'm OOO! If there are tests in the authentication app we might want to add some coverage for this new behavior. Otherwise LGTM 👍

src/v2/Components/Authentication/Desktop/LoginForm.tsx
```diff
@@ -17,28 +17,42 @@ import QuickInput from "v2/Components/QuickInput"
...
 
-  if (!show && props.error === "missing two-factor authentication code") {
-    setShow(true)
+  if (!showOtpInput) {
+    if (error === "missing two-factor authentication code") {
+      setShowOtpInput(true)
+    } else if (error === "missing on-demand authentication code") {
+      setShowOtpInput(true)
+      setShowOtpOnDemandPrompt(true)
+    }
```
dblandin, line 44: **question**: Could this be simplified a bit by using the new helper function defined below?

```tsx
if (!showOtpInput) {
  if (isOtpError(error)) {
    setShowOtpInput(true)
  }

  // Maybe we should add an `isMissingOnDemandOtpError` helper function to co-locate the magic strings?
  if (error === "missing on-demand authentication code") {
    setShowOtpOnDemandPrompt(true)
  }
}
```

dblandin (approved):
Left a comment but nothing blocking. Looks good!

src/v2/Components/Authentication/Desktop/LoginForm.tsx
mc-jones: This makes sense, to me! Was erring on the side of being more explicit for those unfamiliar with the otp flow, but give we had already have utilized a helper, which would likely need to be reference anyways, agree it is likely a better trade off to co-locate the messages.

src/v2/Components/Authentication/__tests__/Desktop/LoginForm.jest.tsx
```diff
@@ -239,6 +239,90 @@ describe("LoginForm", () => {
...
+
+      expect(wrapper.html()).not.toMatch("Error")
+      expect(props.handleSubmit.mock.calls[2][0]).toEqual({
+        email: "email@email.com",
+        password: "password",
+        otp_attempt: "123456",
+        otpRequired: true,
+      })
+    })
+
```
damassi, line 325: Nice tests 👍 ty

PR #8357, opened 2021-09-06
fix(FX-3252): Featured fairs on homepage link to fair 'Overview' tab by default

anastasiapyzhik:
[FX-3252]

**Before**

before

**After**

after

[FX-3252]: https://artsyproduct.atlassian.net/browse/FX-3252

### DIFF
--- a/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
@@ -27,6 +27,19 @@ const HomeFeaturedEventsRail: React.FC<HomeFeaturedEventsRailProps> = ({
     4
   )
 
+  const getHref = (initHref?: string | null): string => {
+    let href = ""
+
+    if (initHref?.includes("/fair/")) {
+      href = initHref
+        .split("/")
+        .filter(part => part !== "booths")
+        .join("/")
+    }
+
+    return href
+  }
+
   if (events.length === 0) return null
 
   return (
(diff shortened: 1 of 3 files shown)


mdole:
Thanks for working on this @anastasiapyzhik! Before we commit to just clobbering the URLs, I wanted to understand where they came from. Here's what I found out (**TL;DR the URLs for featured fairs are set by hand at admin.artsy.net/featured_links)**:

---

- Ok, so these are fetched via a query to OrderedSets, and we actually just have a hardcoded ID that corresponds to the homepage rail:

```graphql
{
  orderedSet(id: "529939e2275b245e290004a0") {
    items {
      __typename
      ... on FeaturedLink {
        id
        internalID
        title
        subtitle
        href
        image {
          cropped(width: 95, height: 63) {
            src
            srcSet
          }
        }
      }
    }
  }
}
```

- And here's the API call that returns those items: https://api.artsy.net/api/v1/set/529939e2275b245e290004a0/items
- An orderedSet can be a bunch of different things. In the case of the homepage, we only care about FeaturedLinks
- Those are set in the admin panel: https://admin.artsy.net/featured_links
- Here's an example of a specific featured link for one of the current fairs: https://admin.artsy.net/featured_link/6135d728fbee4c000c9bb02c
- **So currently those links are simply set in a way that includes /booths or /artworks**
- Also, we do something really goofy with these links: our MP resolver sets `internalID` to be whatever the last thing in the href is. So if the href is `https://www.artsy.net/fair/the-armory-show-2020/booths`, `internalID` will be `booths`. That doesn't seem like ideal behavior! https://github.com/artsy/metaphysics/blob/e5b974c2047d36f707ea2a8a23cc276410f13144/src/schema/v2/featured_link.ts#L15

---

I'm going to start a thread in #fairs-and-friends with members of the fairs team to understand why those links are being set with /booths or /artworks. I think probably that's either intentional and we shouldn't change it, or it's a mistake and they can simply set them differently in the future!

mdole:
Closing this PR - confirmed with Fairs team that those links are intentional for now and that they will link to the overview page once there are artworks there https://artsy.slack.com/archives/C01129C3MTJ/p1631015214030600, https://artsy.slack.com/archives/C01129C3MTJ/p1631028504048400

PR #8365, opened 2021-09-07
fix: use sharify instance from res.locals (PURCHASE-2924)

starsirius:
Fixes https://artsyproduct.atlassian.net/browse/PURCHASE-2924

The [previous attempt](https://github.com/artsy/force/pull/8345) to exclude cookie consents JS for Eigen didn't work. The reason was that we [used a new instance of sharify](https://github.com/artsy/force/blob/07ddb110a241bebb9e948493321d81f5fef0e1a7/src/v2/server.ts#L96), which didn't go through the [bootstrap process](https://github.com/artsy/force/blob/6045498ca44242ca0d556b67f94ac24dc00d8074/src/middleware.ts#L106) that `res.locals.sharify` did. Therefore, `sd.Eigen` wasn't present.

This updates the use of sharify in v2 server to always use the instance from `res.locals.sharify`, instead of importing a new instance again. Looks like this also fixes `CURRENT_PATH` that `apple-itunes-app` meta tag relies on.

cookie-consents-app

### DIFF
--- a/src/v2/server.ts
+++ b/src/v2/server.ts
@@ -1,4 +1,3 @@
-import sharify from "sharify"
 import type { NextFunction } from "express"
 import type { ArtsyRequest, ArtsyResponse } from "lib/middleware/artsyExpress"
 import { buildServerApp } from "v2/System/Router/server"
@@ -56,7 +55,7 @@ app.get(
       }
 
       const headTagsString = ReactDOM.renderToString(headTags as any)
-      const sharifyData = res.locals.sharify.script()
+      const sharify = res.locals.sharify
 
       const { APP_URL, CURRENT_PATH, WEBFONT_URL } = sharify.data
 
@@ -64,7 +63,7 @@ app.get(
         cdnUrl: NODE_ENV === "production" ? CDN_URL : "",
         content: {
           body: bodyHTML,
-          data: sharifyData,
+          data: sharify.script(),
           head: headTagsString,
           scripts,
           style: styleTags,


src/v2/server.ts
```diff
@@ -56,15 +55,15 @@ app.get(
       }
 
       const headTagsString = ReactDOM.renderToString(headTags as any)
-      const sharifyData = res.locals.sharify.script()
+      const sharify = res.locals.sharify
```
starsirius, line 58: Use the sharify instance from `res.locals`, instead of importing it again.

src/v2/server.ts
```diff
@@ -56,15 +55,15 @@ app.get(
...
 
       const { APP_URL, CURRENT_PATH, WEBFONT_URL } = sharify.data
```
starsirius, line 60: The update turns out to fix the `CURRENT_PATH` that `apple-itunes-app` meta tag relies on. See below for before and after:

Screen Shot 2021-09-07 at 1 47 10 PM

src/v2/server.ts
starsirius: I couldn't find tests for v2 server. Is there a good place to test this?

damassi:
@starsirius - Forgive me if i'm missing something. So before it was
```ts
const sharifyData = res.locals.sharify.script()
…
data: sharifyData,
```

and after

```ts
const sharify = res.locals.sharify
…
data: sharify.script()
```

It seems like the only change is that you’re moving the script() invocation down to the bottom? Not sure why that would make any difference though since in both before / after instances its being called withing the req/res route cycle. Is this a race condition issue? Where when it was invoked at the top there was some kind of timing issue by the time the variable assignment happened at data ? (Sorry if i’m missing something obvious). 

Also, not sure [if this line matters](https://github.com/artsy/force/blob/9f53e0e4707166545ec6f81ed2c5dc45ea52d284/src/v2/server.ts#L94-L95) at all.

starsirius:
@damassi Sorry I realized the changes in this PR is not very clear.

The real impact of this change is the `sharify` instance on this line
https://github.com/artsy/force/blob/aca3f492c1b4ce6a8a75a0425d57cf3d65dae394/src/v2/server.ts#L96

Previously it was imported through `import sharify from "sharify"` at the top of the file, and it's missing some data that got later bootstrapped. With the change, now it's grabbed from `res.locals` which should have everything.

damassi:
Oh, I see! This all makes sense now 👍

src/v2/server.ts
damassi: Tests are typically in the closest `__tests__` folder; if there's not a test there then it likely doesn't exist.

damassi (approved):

artsyjian:
Looks like all questions are resolved. Merging!

icirellik:
This is such a neat find! We should also be cautious as this issue has the potential to appear in all other locations where we import `sharify` directly for any SSR code. At a quick glance is over 100 other locations, oh my 😱 

Otherwise, this change looks great! ✨

mzikherman:
This seemed confusing to me, but a the end it makes sense- just need to look at the whole file. The moving of `.script()` from one line to another (when referring to `res.locals.sharify`) shouldn't matter at all (cc @damassi ). Truly the _only_ change is removing the `import sharify..` line. There are other places in the code where `sharify.data` was accessed and are unchanged (so seen in this exact diff), but now refer to  `sharify = res.locals.sharify; sharify.data`.

Amazing catch!

What's the 'rule of thumb' for sharify and server-side use, and can we codify that in a Danger or custom lint rule? Within our React code I think we _generally_ prefer to read it from the context, and any other server-side code, it _must_ be from locals. Other client-side code is fine to import it. (is that correct? That does seem pretty tricky to 'automate', and definitely seems error-prone).

damassi:
To echo @icirellik - we generally just `import { data as sharify} from "sharify"` in our isomorphic components and don't think twice about it, and a lot of devs aren't that aware of the client/server divide.  

If there's some divergence in behavior between SSR and client, we'll need to make sure we start pulling things from the `SystemContext` by a) passing it into context in `buildServerApp`; and b) passing corresponding client sharify instance into context in `buildClientApp`. This is a significantly different flow, however, and we'd need to likely create a danger rule that checks for direct imports from sharify.

PR #8370, opened 2021-09-08
fix(PURCHASE-2909): Show zendesk only for BNMO artworks exceeding price threshold

copperkraft:
Solves [PURCHASE-2909]

Makes the behaviour of the Zendesk "help" chat bubble match the acceptance criteria detailed in the ticket [https://artsyproduct.atlassian.net/browse/PURCHASE-2442] that implemented it.

* On the artwork page, the widget needs to appear if the artwork price is over $10,000 or equivalent (see below)
* On the checkout / make offer flow, the widget needs to appear if the artwork price is over $10,000 or equivalent (see below)

Threshold:
* USD: US$10000
* AUD: A$13000
* EUR: €8000
* HKD: HK$77000
* GBP: £7000

[PURCHASE-2909]: https://artsyproduct.atlassian.net/browse/PURCHASE-2909

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkMeta.tsx
@@ -31,21 +31,18 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
 
   renderImageMetaTags() {
     const { artwork } = this.props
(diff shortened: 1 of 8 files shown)


erikdstock:
Opening the identical pr #8431 above seems to have triggered a new peril run that avoided whatever strange issue happened above.

PR #8377, opened 2021-09-09
[WIP]feat(CX-1837): price database search scrolls to auction results

rquartararo:
**Description** 

Ticket: [CX-1837](https://artsyproduct.atlassian.net/browse/CX-1837)

When a user searches for auction results on the Price Database landing page, navigating to the artist's page triggers the page to scroll to the auction results.

**Screen recording** 

https://user-images.githubusercontent.com/50849237/132647265-e1da8530-3082-4728-9186-ad62fd3b0364.mov


src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -66,6 +66,19 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
   const { pageInfo } = artist.auctionResultsConnection ?? {}
   const { hasNextPage, endCursor } = pageInfo ?? {}
   const artistName = artist.name
+  const queryString = require("query-string")
+
+  // scroll to auction results if param flag is present
+  useEffect(() => {
+    const queryParams = queryString.parse(location.search)
```
olerichter00, line 73: We can use `useRouter` here to get the query string (here we get the filters the same way https://github.com/artsy/force/blob/6700454cc5fd4ad090a2e92f6523293f2b15e21c/src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx#L291).

```
  const { match } = useRouter()
  const { scrollToAuctionResults } = paramsToCamelCase(match?.location.query)
```

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -66,6 +66,19 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
...
+    if (queryParams.scroll_to_auction_results) {
+      scrollIntoView({
+        selector: "#scrollTo--artistAuctionResultsTop",
+        behavior: "smooth",
+        offset: 150,
+      })
+    }
+  }, [queryString])
```
olerichter00, line 81: Maybe we should run this effect only if `scroll_to_auction_results` changes in the URL. Right now we would scroll to the "Auction Results" section any time `scroll_to_auction_results` and any query param changes (this happens every time we change a filter on the page).

olerichter00 (commented):
That looks pretty good! 

I checked the scroll behaviour for a few artists and it seems that [some artists](http://localhost:5000/artist/ger-lataster/auction-results) don't have a "Market Signals" section. For these artists the scrolling to `scrollTo--artistAuctionResultsTop` doesn't work as expected.

rquartararo:
@olerichter00 I also noticed that the scroll after pagination is scrolling to market signals, but I feel it should scroll to the top of the artwork list instead. Thoughts?

src/v2/Apps/PriceDatabase/Components/PriceDatabaseSearch.tsx
```diff
@@ -36,13 +36,15 @@ export const PriceDatabaseSearch: React.FC = () => {
 
   const handleSearch = () => {
     const pathName = `/artist/${artistSlug}/auction-results`
-
     const searchFilters = filterSearchFilters(filters, ALLOWED_FILTERS)
     const queryString = qs.stringify(paramsToSnakeCase(searchFilters))
+    const paramFlag = "scroll_to_auction_results=true"
```
damassi, line 41: We should avoid `snake_case` if possible

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -67,6 +67,20 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
   const { hasNextPage, endCursor } = pageInfo ?? {}
   const artistName = artist.name
 
+  const { match } = useRouter()
+  const queryParams: any = paramsToCamelCase(match?.location.query)
```
olerichter00, line 71: We should try to avoid using `any` wherever we can. Something like this could work:

```
  const { scrollToAuctionResults } = paramsToCamelCase(
    match?.location.query
  ) as { scrollToAuctionResults?: boolean }
  ```

File: src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx [hunk 5]
```diff
@@ -179,13 +193,18 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
...
 
       <MarketStatsQueryRenderer
         artistInternalID={artist.internalID}
         environment={relay.environment}
       />
 
+      <Text variant={["md", "lg"]}>Auction Results</Text>
+      <Spacer my={2} />
+
+      <Box id="scrollTo--artistAuctionResultsTop" />
```
olerichter00, line 206: 🙌

olerichter00 (approved):
Looks good to me! Just added a comment about typing.

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
rquartararo: Fixed!

damassi (approved):

PR #8378, opened 2021-09-09
fix(PX-4495): make continue button disabled if shipping quote not selected

kiryl-zubarau:
JIRA - https://artsyproduct.atlassian.net/browse/PX-4495

Before
image

Now
image

### DIFF
--- a/src/v2/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/v2/Apps/Order/Routes/Shipping/index.tsx
@@ -581,7 +581,10 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
     const isArtaShipping = this.isArtaShipping()
     const isContinueButtonDisabled = isCommittingMutation
       ? false
-      : isArtaShipping && !shippingQuoteId && !this.isCreateNewAddress()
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -581,7 +581,10 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
     const isArtaShipping = this.isArtaShipping()
     const isContinueButtonDisabled = isCommittingMutation
       ? false
-      : isArtaShipping && !shippingQuoteId && !this.isCreateNewAddress()
+      : isArtaShipping &&
+        !!shippingQuotes &&
+        shippingQuotes.length > 0 &&
+        !shippingQuoteId
```
ansor4, line 587: minor, non-blocking: I think we could probably get rid of this ternary operator:
```js
const isContinueButtonDisabled = !isCommittingMutation &&
							  	 isArtaShipping &&
						         !!shippingQuotes &&
						         shippingQuotes.length > 0 &&
						         !shippingQuoteId
```

ansor4 (approved):

PR #8380, opened 2021-09-09
feat(inquiry): decision making

dzucconi:
Ultimately this winds up being sort of simple but was pretty unintuitive to arrive at.

Since a lot of this is a pretty straight port from the Backbone code; the `Engine` is expecting to accept some data which it uses to reason about the steps to move through. Originally this data was stored as models; which mutate their internal state in place. In React/Relay this becomes tricky because we'd either have to update the state functionally by re-instantiating new `Engine`s or pull things directly out of the Relay store. Simplest solution here is just to store the data in a ref; then provide a function which mutates the data directly. This works since we don't need to re-render on any of these changes; just choose which view to transition to.


src/v2/Components/Inquiry/Components/InquiryBackdrop.tsx
```diff
@@ -0,0 +1,19 @@
+import { ModalBase, ModalBaseProps } from "@artsy/palette"
+import React, { useEffect, useState } from "react"
+import styled from "styled-components"
+
+export const InquiryBackdrop: React.FC<ModalBaseProps> = props => {
+  const [isMounted, setIsMounted] = useState(false)
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
```
damassi, line 10: Have noticed this snippet a few times lately -- maybe we can move it to Utils/Hooks

src/v2/Components/Inquiry/Logger.ts
```diff
@@ -0,0 +1,44 @@
+import Cookies from "cookies-js"
+import { uniq } from "lodash"
+
+export class Logger {
```
damassi, line 4: So this isn't confused with our normal logger, can this be renamed to `InquiryLogger`?

src/v2/Components/Inquiry/config.ts
```diff
@@ -109,71 +112,95 @@ export const useEngine = () => {
...
+  )
 
-  const [current, setCurrent] = useState(engine.current())
+  const [current, setCurrent] = useState(engine.current.current())
 
   const View = VIEWS[current as View]
 
+  // Log each step as it updates
+  useEffect(() => {
+    logger.current.log(current)
```
damassi, line 186: Should we log outside of `NODE_ENV=development`?

dzucconi: This is a very different kind of logger; I'll just rename it. Basically it's just a thing that keeps track of what views you've seen and persists between sessions.

src/v2/Components/Inquiry/Components/InquiryBackdrop.tsx
dzucconi: Apparently there's already a hook for this in there. Swapping

PR #8382, opened 2021-09-10
feat(FX-3170): refresh search results page

tr-ann:
Jira: [FX-3170](https://artsyproduct.atlassian.net/browse/FX-3170)

### Description
The purpose of this ticket is to refresh the styles and layout of the search results page in Force to match the latest design system.
Figma spec is [here](https://www.figma.com/file/0svJJjlcrCFBkVYzagKP3Z/FX-v3-Surfaces?node-id=4021%3A2880) 

### Changes
- Updated components style
- Updated tests for search app
- Removed unused files
- Small refactor of components

### Demo

[review app](https://pr-8882.artsy.net/search?term=andy)


src/v2/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -17,17 +20,55 @@ export interface Props extends SystemContextProps {
...
+
+const RoundedRouteTab: React.FC<RouteTabProps> = ({ text, count, ...rest }) => {
+  const isActive = useIsRouteActive(rest.to, { exact: rest.exact ?? true })
+
+  return (
+    <RouteTab
+      height="auto"
+      borderWidth="1px"
+      borderStyle="solid"
+      borderRadius="20px / 50%"
```
dimatretyak, line 49: **nitpick**: can we specify `10px` here right away?

src/v2/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -17,17 +20,55 @@ export interface Props extends SystemContextProps {
...
+          <Text
+            variant="xs"
+            color={isActive ? "blue100" : "black60"}
+            display="inline"
+          >
+            {count}
+          </Text>
+        )}
+      </Media>
+    </RouteTab>
```
dimatretyak, line 70: **suggestion**: please clarify with Jonathan whether there should be such a style with the hover effect?

https://user-images.githubusercontent.com/3513494/132865254-71e83241-e774-4b25-8f90-4fbd9aa7e174.mp4

dimatretyak:
**nitpick**:  pills is different from the design in Figma on small devices

Search Results for 'aa' | Artsy 2021-09-10 17-06-17

src/v2/Apps/Search/Components/NavigationTabs.tsx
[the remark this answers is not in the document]
dzucconi: These should just be `Pill` components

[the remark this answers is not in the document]
dimatretyak: Oh, I didn't know that palette has [such a component](https://palette.artsy.net/elements/pill/#gatsby-focus-wrapper). Thanks!

src/v2/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -17,17 +20,55 @@ export interface Props extends SystemContextProps {
...
+        {count && (
+          <Text
+            variant="xs"
+            color={isActive ? "blue100" : "black60"}
+            display="inline"
+          >
+            {count}
+          </Text>
+        )}
+      </Media>
```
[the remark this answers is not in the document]
dzucconi: We have a component for doing the sizing correctly called `Sup`: https://palette-storybook.artsy.net/?path=/story/components-sup--default

src/v2/Apps/Search/Components/SearchMeta.tsx
```diff
@@ -1,24 +1,20 @@
...
+  const title = `Search Results for '${term}' | Artsy`
+  const href = `/search?term=${term}`
+  return (
+    <>
+      <Title>{title}</Title>
+      <Link rel="canonical" href={`${sd.APP_URL}${href}`} />
+      <Meta property="og:title" content={title} />
+      <Meta property="og:url" content={`${sd.APP_URL}${href}`} />
+    </>
+  )
```
dzucconi, line 19: You can use https://github.com/artsy/force/blob/master/src/v2/Components/MetaTags.tsx here

src/v2/Apps/Search/Components/GenericSearchResultItem.tsx
```diff
@@ -54,52 +45,31 @@ export class GenericSearchResultItem extends React.Component<
...
+              <Text size="md">{name}</Text>
 
               {description && (
-                <>
-                  <Spacer mb={0.5} />
-                  <Serif color="black60" size="3" maxWidth={536}>
-                    <Truncator maxLineCount={3}>{description}</Truncator>
-                  </Serif>
-                </>
+                <Text mt={0.5} variant="sm" color="black60">
```
dzucconi, line 67: This component has a `lineClamp` prop — so you can get rid of that truncator below

src/v2/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -59,88 +106,67 @@ export class NavigationTabs extends React.Component<Props> {
...
+  restAggregationCount > 0 &&
+    tabs.push(
+      renderTab("More", route("/more"), {
+        count: restAggregationCount,
+      })
     )
-  }
+
+  return (
+    <Swiper my={2} snap="start">
```
dzucconi, line 162: I'm sure this is fine; though generally we use this for things of a consistent width. It might be more appropriate to just use `BaseTabs` here — which is a free scroll with no snapping but has the gradient fade overlay as an affordance when items go off-screen. I _think_ it should work fine. I may be mistaken.

src/v2/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -169,10 +195,11 @@ export interface TabCounts {
...
 }
 
-export const tabCountMap: (props: Props) => TabCounts = props => {
+export const tabCountMap: (
+  searchableConnection: NavigationTabs_searchableConnection
+) => TabCounts = props => {
   return Object.entries(TAB_NAME_MAP).reduce((acc, [key, val]) => {
-    // @ts-expect-error STRICT_NULL_CHECK
-    let count = get(aggregationFor(props, key), agg => agg.count, 0)
+    let count = get(aggregationFor(props, key), agg => agg?.count, 0)
```
dzucconi, line 202: `non-blocking:` if you see these `get` calls you can just replace them with the safe operator

```suggestion
    let count = aggregationFor(props, key)?.count ?? 0
```

dzucconi (approved):
Looks good. I think if you rebase with master you should get that acceptance test passing. I'd consider swapping in `BaseTabs` (we should maybe rename that) for `Swiper` to get the gradient affordance; but non-blocking.

src/v2/Apps/Search/Components/NavigationTabs.tsx
tr-ann: Thanks Damon, but when I use BaseTabs I get a separator at the bottom of tabs
image

And I don't understand how I can remove it. Can you help me with it?

dzucconi: Ah, OK maybe punt on this for now and I'll extract the underlying primitive so that we can follow up. Thanks for looking!

mdole (commented):
Really nice work @tr-ann!! Just checked this PR out locally and there are a few small visual tweaks that we should make before merging. Also two questions I'll ask Jonathan in Slack: should these pills be sticky? And should a scroll rail be visible on mobile/small window width?

### Pills missing margin between: 
Screen Shot 2021-09-15 at 11 08 12 AM

---

### Filter button doesn't match design specs
It's a black/blue pill here instead of the minimal text + icon

Screen Shot 2021-09-15 at 11 09 06 AM

---

### Missing 1 space between text + numbers in pill

Design:
Screen Shot 2021-09-15 at 11 15 17 AM

Current:
Screen Shot 2021-09-15 at 11 13 49 AM

---

### Pill "selected" state differs from design:

Frankly, this might be fine or it might be something we should change everywhere at the same time. I'm not 100% sure if this is already being discussed with the design team.

Design: 
Screen Shot 2021-09-15 at 11 16 46 AM

Current:
Screen Shot 2021-09-15 at 11 12 07 AM

mdole:
Jonathan got back to us on Slack - we don't want a scroll rail, we do want sticky.

tr-ann:
@mdole thanks for the clarifying. I updated the pills style and made them sticky.
What about visible scroll bar, I think we need to investigate it separately, because the scroll disappear when scrolling at least one and (example in attached video)

https://user-images.githubusercontent.com/56556580/133429808-cb2ff9c0-6e51-42f4-861d-e1370349e20b.mov

I think a new style of filter button is common (not only for search results page), so it will be better to refresh it separately (maybe we need to create ticket for that?). What do you think?

And the last, pill selected state. We need to create another one variant for search tabs (now we have pill styles for filters, artist and text). There is discussion about it on [slack](https://artsy.slack.com/archives/C9XJKPY9W/p1631283326018400)

dzucconi:
@mdole Regarding the Pill design discrepancies: this was already discussed: there's a Pill filter variant. That's what this should use. The design _is_ updated but it's not reflected in Palette yet. Leave these alone and DSWG will take care of the Pill updates down the line.

dzucconi:
But the various spacing issues you point out are valid

dzucconi:
I'll also pull out a new component for doing that scroll thing. I'll see if I can get to that today.

dimatretyak (approved):
LGTM 👍

mdole:
Thank you both for clarifying! @tr-ann I think we can use the HorizontalOverflow component Damon added to Palette in [this PR](https://github.com/artsy/palette/pull/1034) to address the scrollbar behavior, and then I think we're ready to merge!

Because this is a large visual change, I would also suggest that we notify Jonathan and Thomas and make sure they have time to QA shortly after it is deployed to staging. Or, as an alternative, we could spin up a review app (see instructions [here](https://github.com/artsy/force/blob/bd325f12d3f8296cabdfab6901f9edbdad9ab9d9/docs/creating_review_app.md)) to give them time to QA without blocking deployments to production.

@tr-ann want to try creating a review app first, and if it's not too much work, we can use that for QA before we merge this PR? Please let me know if I can help!

tr-ann:
Sure, thanks @mdole. I'll use HorisontalComponent and create review app after that

tr-ann:
@mdole I notice interesting thing. BaseTabs component implementation already include HorizontalOverflow component ([here](https://github.com/artsy/palette/blob/694063c1a8ec3e42d36014e81541b08879cd6217/packages/palette/src/elements/BaseTabs/BaseTabs.tsx#L34)), but I don't understand why scroll bar displayed anyway.
@dzucconi can you clarify it?

dzucconi:
@tr-ann I extracted that `HorizontalOverflow` so that you didn't have to use the `BaseTabs` (which has a border and in-built spacing). Though I guess this is fine.

WRT to the scrollbar. This is what it's using: https://github.com/artsy/palette/blob/ea901c51f6fbbed8a89c6aa87e973dd10c306325/packages/palette/src/helpers/visuallyDisableScrollbar.ts#L6-L15 — it's possible that there's just some browser setting there that's overruling it. Feel free to PR this mixin in Palette since you have a repro; but I'd punt as I suspect it's not affecting a major percentage. Also it's not like it's a bad thing that it's more usable IMO 😬

mdole:
While looking at the review app, I noticed something that is also an issue with the current search results page (it is NOT new with this PR): when you click an artwork, your search term is replaced by “Undefined”. See the ticket below for gifs.

I’ve ticketed that [here](https://artsyproduct.atlassian.net/browse/FX-3328), just mentioning in this ticket in case anyone else notices it. Considering this is not new behavior, I would suggest we don’t block this PR and instead follow up!

dzucconi:
Yeah that seach bug is very long standing and I believe ticketed within FX?

PR #8384, opened 2021-09-10
refactor(PURCHASE-2774): Update Saved Address and Payment to the new design system

alexj105:
Saved Address and Payment were ubdated!
Handle on empty cards and empty addresses lists
**Desktop:**
Screen Shot 2021-09-10 at 5 23 45 PM

Screen Shot 2021-09-10 at 5 22 36 PM
Screen Shot 2021-09-10 at 4 37 10 PM
Screen Shot 2021-09-10 at 4 37 23 PM

**Mobile:**
Screen Shot 2021-09-10 at 4 42 22 PM
Screen Shot 2021-09-10 at 4 45 33 PM
Screen Shot 2021-09-10 at 4 56 23 PM
Screen Shot 2021-09-10 at 5 14 13 PM


src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -36,6 +35,7 @@ export const NEW_ADDRESS = "NEW_ADDRESS"
 const PAGE_SIZE = 30
 
 interface SavedAddressesProps {
+  handeChangesAddressCount?: (active?: number) => void
```
damassi, line 38: The general idiomatic rule of thumb with callbacks in React is that callbacks passed down as props use the "onSomething" naming form, and callbacks that are defined internal to the component use the `handleSomething` form.

src/v2/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -177,11 +173,11 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
     }
 
     const isDefaultAddress = address.isDefault
-
     return (
       <BorderBox
         p={2}
-        width="100%"
+        mb={2}
+        width={340}
```
damassi, line 180: (non-blocking) Fixed widths like this don't scale so well when the the browser is at various sizes. Can we use a percent width instead of something fixed? If not, no prob; we just want to be sure that it works in different browsers and on mobile web.

src/v2/Apps/Payment/Components/PaymentSection.tsx
```diff
@@ -24,22 +24,27 @@ export const PaymentSection: React.FC<PaymentSectionProps> = props => {
...
     <>
-      <Text variant="lg" my={4}>
-        Saved Cards
+      <Text variant={["sm", "lg"]} mt={4} mb={[2, 4]}>
+        {creditCards.length ? "Saved Cards" : "No Saved Cards"}
       </Text>
+
       {creditCards?.length ? (
-        <Box maxWidth={542}>
+        <Box maxWidth={940}>
```
damassi, line 34: Can we make the width a %?

src/v2/Components/UserSettings/UserSettingsAddresses.tsx
```diff
@@ -55,6 +59,7 @@ export const UserSettingsAddresses: React.FC<UserSettingsAddressesProps> = props
       <SavedAddresses
         // @ts-expect-error STRICT_NULL_CHECK
         me={me}
+        handeChangesAddressCount={handeChangesAddressCount}
```
damassi, line 62: `onChangeAddressCount={handleChangeAddressCount}`

damassi (approved):
There's a couple minor comments that we should address, but otherwise LGTM!

PR #8386, opened 2021-09-10
fix(gro-514): homepage featured rail typography/layout updates

dzucconi:
Closes: [GRO-514](https://artsyproduct.atlassian.net/browse/GRO-514)

### DIFF
--- a/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
@@ -1,16 +1,17 @@
 import {
-  Flex,
   Box,
   Image,
   Text,
-  Spacer,
   GridColumns,
   Column,
+  ResponsiveBox,
+  Flex,
 } from "@artsy/palette"
 import { compact, take } from "lodash"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
```diff
@@ -30,47 +31,78 @@ const HomeFeaturedEventsRail: React.FC<HomeFeaturedEventsRailProps> = ({
...
+                    </Box>
+                  </Flex>
+                </Media>
 
-                <Flex flexDirection="column" justifyContent="center">
-                  <Text variant="xs" color="black60">
+                <Media greaterThanOrEqual="sm">
+                  <ResponsiveBox
+                    aspectWidth={3}
+                    aspectHeight={2}
```
damassi, line 77: What do these numbers do?

dzucconi: They set the aspect ratio. So this becomes a responsive 3:2 box. In this case I could also use the width/height coming off the crop

PR #8387, opened 2021-09-10
Refactor: convert AddressForm to functional component to and add v3 styles

lilyfromseattle:
# PURCHASE-2929

paired with @copperkraft and @ivan-badyulya 

Styles on checkout shipping form country selector are inconsistent with the rest of the form for v3.

AddressForm lives in src/v2/components so was not converted to v3 along with the order app, should be styled to work with both v2 and v3 components. In order to do this, it needed to be converted to a functional component so that we could use `useThemeConfig` to style conditionally so as not to affect v2 components also using it

__After:__
_V3:_
Screen Shot 2021-09-09 at 11 34 02 AM
.
_V2_
Screen Shot 2021-09-09 at 2 58 17 PM

__Before:__
_V3:_
Screen Shot 2021-09-10 at 2 08 22 PM


src/v2/Components/AddressForm.tsx
```diff
@@ -44,190 +53,195 @@ export interface AddressFormProps {
   touched: AddressTouched
 }
 
-interface AddressFormState {
-  address: Address
-}
-
-export class AddressForm extends React.Component<
```
damassi, line 51: Thanks for converting this to a FC

damassi (approved):

copperkraft:
Locally tests pass, but on ci one fails.

dzucconi:
So this got merged without being green and the test is currently failing in master

src/v2/Components/__tests__/Utils/addressForm.ts
```diff
@@ -37,6 +37,9 @@ export const fillInPhoneNumber = (
 
 export const fillCountrySelect = (component, value) => {
   const input = component.find(CountrySelect)
+  // input.simulate("select", {
+  //   target: { value: "GB" },
+  // })
```
dzucconi, line 42: ☝️

PR #8390, opened 2021-09-10
feat: Remove homepage modules

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-515 

Related: 
- https://github.com/artsy/force/pull/8390

Removes the following home page modules: 
- Featured categories
- Works by popular artists
- Category/Gene carousels


src/v2/Apps/Home/Components/HomeArtworkModules.tsx
```diff
@@ -10,6 +10,8 @@ import { MyBidsQueryRenderer } from "v2/Apps/Auctions/Components/MyBids/MyBids"
 import { HomeFeaturedShowsLazyQueryRenderer } from "./HomeFeaturedShows"
 import { compact } from "lodash"
 
+const EXCLUDED_MODULES = ["popular_artists", "generic_gene"]
```
damassi, line 13: @dzucconi - It seems like this should be elevated to MP. Thoughts?

dzucconi, line 13: The query for this supports an `exclude` argument. So we can just adjust the query:

```graphql
artworkModules(exclude: [POPULAR_ARTISTS, GENERIC_GENES]) {
  ...
}
```

dzucconi: [Already is!](https://github.com/artsy/force/pull/8390#discussion_r706443835)

damassi: Ah, looks like I was on the wrong field: 

Screen Shot 2021-09-10 at 1 13 06 PM

Will fix 👍

damassi: (Ah -- was looking at v1.)

dzucconi: Ugh, this happens to me frequently

PR #8393, opened 2021-09-13
feat: add flag country code dropdown to profile form

ashleyjelks:
https://artsyproduct.atlassian.net/browse/PURCHASE-2913

cc @artsy/purchase-devs 

This PR introduces a country/flag dropdown list to enable users to select their country code separately from their phone number. For now, we are going to add the country code and phone number to the mutation, so every number is store as  `+country code-phone number`. 

We should see this both on the Saved Addresses form and in the Payment checkout flow.

Kapture 2021-09-13 at 11 43 59


src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -72,6 +75,14 @@ export const AddressModal: React.FC<Props> = ({
   modalDetails,
   me,
 }) => {
+  const [countryCode, setCountryCode] = useState<string>("us")
```
erikdstock, line 78: non-blocking - can we use a [user's location](https://github.com/artsy/metaphysics/blob/master/_schema.graphql#L15028-L15040) as the default country?

src/v2/Apps/Order/Utils/shippingUtils.ts
```diff
@@ -120,12 +120,15 @@ export const convertShippingAddressForExchange = (
 }
 
 export const convertShippingAddressToMutationInput = (
+  phoneNumber: SavedAddressType["phoneNumber"],
   address: SavedAddressType
 ): UserAddressAttributes => {
+  const modifiedAddress = assign(address, { phoneNumber })
```
erikdstock, line 126: Can this be done with something like `const modifiedAddress = { ...address, phoneNumber }`?

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -180,17 +194,49 @@ export const AddressModal: React.FC<Props> = ({
...
-                value={formik.values?.phoneNumber || ""}
-              />
+              <Flex>
+                <Box style={{ maxWidth: "35%" }}>
+                  <Select
+                    title="Phone number"
+                    description="Only used for shipping purposes"
+                    options={countries}
+                    onSelect={cc => {
+                      setCountryCode(cc)
```
erikdstock, line 204: Should we be calling `mergeCountryCodeToPhoneNumber` here in case someone changes this value without triggering the `onBlur` in the phone number input?

erikdstock (commented):
Cool! Left a few comments, mostly minor.

src/v2/Apps/Order/Components/AddressModal.tsx
ashleyjelks: Yes, that's been discussed [here](https://artsy.slack.com/archives/C2TQ4PT8R/p1629920350084200) and also [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1628605963138300), but I asked Charles if this geolocating should be done on our first pass and the answer was no (for now).

src/v2/Apps/Order/Utils/shippingUtils.ts
ashleyjelks: Good call!

src/v2/Apps/Order/Components/AddressModal.tsx
ashleyjelks: I updated this to account for that edge case...we only call it from here now if there's a value in the phone input field.

PR #8394, opened 2021-09-13
feat: Update homepage article layout

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-516 

This updates the homepage articles section to the new design.

articles

### DIFF
--- a/src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
+++ b/src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
@@ -7,17 +7,23 @@ import {
   Skeleton,
   SkeletonBox,
   SkeletonText,
-  Shelf,
+  GridColumns,
+  Column,
+  ResponsiveBox,
 } from "@artsy/palette"
-import { compact } from "lodash"
+import { compact, take } from "lodash"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { Masonry } from "v2/Components/Masonry"
 import { useSystemContext } from "v2/System"
 import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
 import { RouterLink } from "v2/System/Router/RouterLink"
 import { useLazyLoadComponent } from "v2/Utils/Hooks/useLazyLoadComponent"
+import { Media } from "v2/Utils/Responsive"
 import { HomeFeaturedArticlesQuery } from "v2/__generated__/HomeFeaturedArticlesQuery.graphql"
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
```diff
@@ -26,53 +30,92 @@ interface HomeFeaturedArticlesProps {
...
+                          aspectHeight={image.height}
+                          maxWidth={image.width}
+                        >
+                          <Image
+                            width="100%"
+                            src={image.src}
+                            srcSet={image.srcSet}
+                            // FIXME: Not sure why second column of masonry grid
+                            // doesn't lazy load. Disabling for now.
+                            // lazyLoad={true}
```
damassi, line 97: Not sure whats up here -- masonry doesn't like it /shrug. It's not really noticeable and only 6 images, but heads up.

dzucconi: This should have a `height="100%"` as well; it could be that it's just collapsing to 0 so the intersection observer isn't being triggered.

src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
```diff
@@ -7,72 +7,117 @@ import {
...
-              </Text>
-            </RouterLink>
-          )
-        })}
-      </Shelf>
+              return (
+                <RouterLink
+                  key={article.internalID}
+                  to={article.href ?? ""}
+                  style={{ display: "block", textDecoration: "none" }}
```
dzucconi, line 83: `RouterLink` accepts `BoxProps` now

src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
```diff
@@ -124,29 +180,78 @@ export const HomeFeaturedArticlesFragmentContainer = createFragmentContainer(
...
+        </Column>
 
-              <Spacer mt={0.5} />
+        <Column span={[12, 6]}>
+          <Masonry columnCount={[1, 2]}>
+            {[...new Array(8)].map((_, i) => {
+              return (
+                <React.Fragment key={i}>
+                  <Box mb={4}>
+                    <>
```
dzucconi, line 209: Could get rid of both of these fragments

src/v2/Apps/Home/Components/HomeFeaturedArticles.tsx
damassi: ahh 🍝

PR #8395, opened 2021-09-14
fix cypress test

alexj105:
The reason for this error was "Presented by" string that is a bit specific for different show items(_show/daniel-cooney-fine-art-daniel-cooney-fine-art-at-art-on-paper-2021?sort=partner_show_position_). It was replaced by **Sow** because this string is common for all items!
Thanks @copperkraft for paring!
Screen Shot 2021-09-14 at 3 02 15 PM


joeyAghion (approved):
Thanks! I didn't realize that some show pages didn't have that text!

PR #8396, opened 2021-09-14
feat(inquiry): integrates new inquiry behind env variable

dzucconi:
This PR adds an additional branch to every call site for the existing inquiry where we check the `ENABLE_V3_INQUIRY` and potentially use that instead of the old flow.

Currently there's no analytics and there hasn't been an Q/A so we don't want to enable this yet.

I'd recommend looking at this with disabled whitespace diffing: https://github.com/artsy/force/pull/8396/files?diff=unified&w=1


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarExtraLinks.tsx
```diff
@@ -85,6 +93,11 @@ class ArtworkSidebarExtraLinksContainer extends React.Component<
...
   }))
   onClickBuyNowAskSpecialist() {
+    if (sd.ENABLE_V3_INQUIRY) {
+      this.props.showInquiry({ askSpecialist: true })
+      return
+    }
+
     this.props.mediator &&
       this.props.mediator.trigger &&
       this.props.mediator.trigger("openBuyNowAskSpecialistModal", {
```
dzucconi, line 103: There's no actual difference between `onClickBuyNowAskSpecialist` and `onClickAuctionAskSpecialist`

PR #8397, opened 2021-09-14
Revert "feat: add flag country code dropdown to profile form"

ashleyjelks:
Reverts artsy/force#8393


damassi:
@ashleyjelks - mind adding a reason for the revert in the PR description? 🙏

ashleyjelks:
This PR was reverted because when it was QA'd on staging we found a bug that shouldn't be deployed. Will send a fix out today.

damassi:
Sorry @ashleyjelks, meaning: what bug did you see? If something is reverted its usually from a `bug`, but that doesn't provide much info to other devs. Since reverts can be ambiguous and red-herring-ish its best to treat them like regular PRs with a properly fleshed out description.

PR #8399, opened 2021-09-15
fix(PURCHASE-2931):Editing saved address in form cause page to crash

alexj105:
I found a small bug, I don't know the reason for it. Initial values for Formik is gotten from relay and marked like readonly. So we get an error when we try to change this field.An Spread operator solved this issue. (https://artsyproduct.atlassian.net/browse/PURCHASE-2931)

### DIFF
--- a/cypress/integration/shows.spec.ts
+++ b/cypress/integration/shows.spec.ts
@@ -15,6 +15,6 @@ describe("Shows", () => {
     const showLink = cy.get('a[href*="/show/"]:first')
     showLink.click()
     cy.url().should("contain", "/show/")
-    cy.contains("Presented by")
+    cy.contains("Show")
   })
 })
--- a/src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
+++ b/src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
@@ -8,7 +8,12 @@ import {
 type UseArtworkInquiryRequestInput = Omit<
   SubmitInquiryRequestMutationInput,
   "inquireableID" | "inquireableType" | "message"
-> & { relayEnvironment?: Environment; artworkID: string; message: string }
+> & {
+  relayEnvironment?: Environment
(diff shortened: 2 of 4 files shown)


src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
```diff
@@ -8,7 +8,12 @@ import {
 type UseArtworkInquiryRequestInput = Omit<
   SubmitInquiryRequestMutationInput,
   "inquireableID" | "inquireableType" | "message"
-> & { relayEnvironment?: Environment; artworkID: string; message: string }
+> & {
+  relayEnvironment?: Environment
+  artworkID: string
+  message: string
+  contactGallery?: boolean | null
```
ashleyjelks, line 15: Why was the contactGallery prop needed?

ashleyjelks (approved):

cypress/integration/shows.spec.ts
```diff
@@ -15,6 +15,6 @@ describe("Shows", () => {
     const showLink = cy.get('a[href*="/show/"]:first')
     showLink.click()
     cy.url().should("contain", "/show/")
-    cy.contains("Presented by")
+    cy.contains("Show")
```
starsirius, line 18: Is this relevant?

alexj105: absolutely!

alexj105: I dont know why git set it like new changes, because we have this string [here](https://github.com/artsy/force/blob/master/cypress/integration/shows.spec.ts) so it is absolutely safe

src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
alexj105: > Why was the contactGallery prop needed?

I got pre push type error from prev changes

PR #8400, opened 2021-09-15
feat: Add autosuggest quick navigation (second try)

olerichter00:
Addresses [CX-1761]

## Description

Adds quick navigation items ("Artworks" & "Auction Results") to the search bar's autosuggest items for artists.

Screenshot 2021-09-17 at 10 19 53

[CX-1761]: https://artsyproduct.atlassian.net/browse/CX-1761

### DIFF
--- a/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
+++ b/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
@@ -1,11 +1,4 @@
-import {
-  ArtworkIcon,
-  AuctionIcon,
-  Clickable,
-  Flex,
-  Pill,
-  Text,
-} from "@artsy/palette"
+import { ArtworkIcon, AuctionIcon, Flex, Pill, Text } from "@artsy/palette"
 import match from "autosuggest-highlight/match"
 import parse from "autosuggest-highlight/parse"
 import React from "react"
--- a/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
+++ b/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
@@ -126,7 +126,8 @@ const QuickNavigationItem: React.FC<{ to: string }> = ({ children, to }) => {
     event.stopPropagation()
     event.preventDefault()
 
-    router ? router.push(to) : window.location.assign(to)
+    // Fixme: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
+    window.location.assign(to)
   }
   return (
     <ClickableFlex onClick={onClick} mt={1} mr={1}>
--- a/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
+++ b/src/v2/Components/Search/Suggestions/SuggestionItem.tsx
@@ -1,9 +1,18 @@
-import { Text } from "@artsy/palette"
+import {
+  ArtworkIcon,
+  AuctionIcon,
+  Clickable,
+  Flex,
+  Pill,
+  Text,
+} from "@artsy/palette"
 import match from "autosuggest-highlight/match"
 import parse from "autosuggest-highlight/parse"
 import React from "react"
 import styled from "styled-components"
+import { Container } from "v2/Components/Sticky"
 import { RouterLink } from "v2/System/Router/RouterLink"
+import { useRouter } from "v2/System/Router/useRouter"
 
 interface SuggestionItemProps {
   display: string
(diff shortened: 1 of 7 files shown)


src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +92,47 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+  )
+}
+
+const QuickNavigationItem: React.FC<{ to: string }> = ({ children, to }) => {
+  const { router } = useRouter()
+
+  const onClick = event => {
+    // Stopping the event from propagating to prevent SearchBar from navigation to the main suggestion item url.
+    event.stopPropagation()
+    event.preventDefault()
```
olerichter00, line 127: Stopping the event from propagating to prevent `SearchBar` from navigation to the main suggestion item URL.

src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +92,47 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+          Auction Results
+        </QuickNavigationItem>
+      )}
+    </Flex>
+  )
+}
+
+const QuickNavigationItem: React.FC<{ to: string }> = ({ children, to }) => {
+  const { router } = useRouter()
+
```
olerichter00, line 123:
```suggestion
```

src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +92,46 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+  )
+}
+
+const QuickNavigationItem: React.FC<{ to: string }> = ({ children, to }) => {
+  const onClick = event => {
+    // Stopping the event from propagating to prevent SearchBar from navigation to the main suggestion item url.
+    event.stopPropagation()
+    event.preventDefault()
+
+    // Fixme: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
```
olerichter00, line 127:
```suggestion
    // FIXME: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
```

src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -78,3 +92,46 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
...
+  const onClick = event => {
+    // Stopping the event from propagating to prevent SearchBar from navigation to the main suggestion item url.
+    event.stopPropagation()
+    event.preventDefault()
+
+    // FIXME: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
+    window.location.assign(to)
+  }
+  return (
+    <ClickableFlex onClick={onClick} mt={1} mr={1}>
```
dzucconi, line 131: If you want to use `Clickable` get it from palette. This is apparently just a flex with a pointer for a cursor: which should never be used (I'd consider this a bug in the conversation app). Clickable is an actual `<button>` and supports all `BoxProps`

dzucconi: To be clear: the reason why we'd use an actual button is [so that it is accessible](https://tech.boxed.com/accessible-ui-stop-making-clickable-divs-264115938716)

dzucconi (commented):
Overall I think the more appropriate approach would be to absolutely position these in a parent above the actual suggestion and make them actual links because this will break cmd/ctrl+clicks the way it's written. Obviously this autosuggest library makes that difficult. You could also check keydown in that click handler as an alternative. However I wouldn't sink too much time into this as it can now be replaced with the Autocomplete in Palette. I've got a PR I'm sitting on for that here so I think the best thing is to get this working/merged and we'll follow up.

damassi:
Note that this search bar code is a great candidate for a complete rewrite

dzucconi:
@damassi I've got a branch where i rewrote it already :p

damassi:
oh what! Damn 👌 👌

dzucconi:
https://github.com/artsy/force/compare/dzucconi/refactor/autocomplete-integration?expand=1#diff-e310f1cc3bb59bb47b8f02ad047e93ffbe464faf299a09afa405fa5c3d1a0d8d — but I need to port this work over

damassi:
Needless to say, your version simpler 😄

PR #8401, opened 2021-09-15
feat(FX-3280): added back link for fair shows

tr-ann:
Jira: [FX-3280](https://artsyproduct.atlassian.net/browse/FX-3280)

### Description
As a collector, I should be able to go back to the Fairs 'Exhibitors A-Z' tab after clicking through to view a Fairs 'Show'. Currently, there is no navigational banner making it explicit and easy for the collector to go back to the Fair page to continue browsing exhibitors within a Fair.

[Figma](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3474%3A17827)

### Changes
- Added navigation bar with link to `/fair/:slug/exhibitors`

### Demo
When redirect from fair

https://user-images.githubusercontent.com/56556580/133590538-91e5a1fd-d89c-4899-8f10-ed320385be22.mov

When redirect from shows page

https://user-images.githubusercontent.com/56556580/133590568-f2e7b15a-df3d-450d-9f38-eb0bb7e86902.mov

### DIFF
--- a/src/v2/Apps/Show/Components/ShowHeader.tsx
+++ b/src/v2/Apps/Show/Components/ShowHeader.tsx
@@ -13,8 +13,11 @@ export const ShowHeader: React.FC<ShowHeaderProps> = ({ show, ...rest }) => {
   const { name, startAt, endAt, formattedStartAt, formattedEndAt } = show
 
   const currentTime = useCurrentTime({ syncWithServer: true })
-  // @ts-expect-error STRICT_NULL_CHECK
-  const { formattedTime } = useEventTiming({ currentTime, startAt, endAt })
+  const { formattedTime } = useEventTiming({
+    currentTime,
+    startAt: startAt!,
+    endAt: endAt!,
+  })
 
   return (
     <Box {...rest}>
@@ -51,15 +54,6 @@ export const ShowHeaderFragmentContainer = createFragmentContainer(ShowHeader, {
       status
       formattedStartAt: startAt(format: "MMMM D")
       formattedEndAt: endAt(format: "MMMM D, YYYY")
-      partner {
-        ... on Partner {
-          name
-        }
-        ... on ExternalPartner {
-          name
-        }
-      }
-
       ...ShowContextualLink_show
     }
   `,
--- a/src/v2/Apps/Show/Components/ShowNavigationBanner.tsx
+++ b/src/v2/Apps/Show/Components/ShowNavigationBanner.tsx
@@ -1,5 +1,5 @@
-import { BoxProps } from "@artsy/palette"
 import React from "react"
+import { BoxProps } from "@artsy/palette"
 import { BackLink } from "v2/Components/Links/BackLink"
 import { ShowNavigationBanner_show } from "v2/__generated__/ShowNavigationBanner_show.graphql"
 import { createFragmentContainer, graphql } from "react-relay"
@@ -12,16 +12,11 @@ const ShowNavigationBanner: React.FC<ShowNavigationBannerProps & BoxProps> = ({
   show,
   ...rest
 }) => {
-  const { name, fair, isFairBooth, partner } = show
-
-  if (!fair?.href) {
-    return null
-  }
+  const { fair } = show
 
   return (
-    <BackLink {...rest} to={`${fair.href}/exhibitors`}>
-      Back to {name}
-      {!isFairBooth && partner?.name && `at ${partner.name}`}
+    <BackLink {...rest} to={`${fair?.href}/exhibitors`}>
+      Back to {fair?.name}
     </BackLink>
   )
 }
@@ -31,17 +26,8 @@ export const ShowNavigationBannerFragmentContainer = createFragmentContainer(
   {
     show: graphql`
       fragment ShowNavigationBanner_show on Show {
-        name
-        isFairBooth
-        partner {
-          ... on Partner {
-            name
-          }
-          ... on ExternalPartner {
-            name
-          }
-        }
         fair {
+          name
           href
         }
       }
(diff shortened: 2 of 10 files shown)


src/v2/Apps/Show/Components/ShowHeader.tsx
```diff
@@ -13,8 +13,11 @@ export const ShowHeader: React.FC<ShowHeaderProps> = ({ show, ...rest }) => {
   const { name, startAt, endAt, formattedStartAt, formattedEndAt } = show
 
   const currentTime = useCurrentTime({ syncWithServer: true })
-  // @ts-expect-error STRICT_NULL_CHECK
-  const { formattedTime } = useEventTiming({ currentTime, startAt, endAt })
+  const { formattedTime } = useEventTiming({
+    currentTime,
+    startAt: startAt!,
+    endAt: endAt!,
```
iskounen, line 19: **question:** I assume that we were suppressing `STRICT_NULL_CHECK` errors because we couldn't guarantee that `startAt` and `endAt` would be non-null and non-undefined. Has something changed with these props that allows us to ensure that they are non-null and non-undefined?

src/v2/Apps/Show/ShowApp.tsx
```diff
@@ -52,9 +53,13 @@ export const ShowApp: React.FC<ShowAppProps> = ({ show }) => {
             contextPageOwnerType,
           }}
         >
-          <Box mt={4} mb={[4, 6]}>
-            <ShowInstallShots show={show} mt={4} mb={6} />
-          </Box>
+          <ShowNavigationBanner show={show} mt={2} mb={4} />
```
iskounen, line 56: **question:** It looks like we will be showing this navigation banner regardless of how a user arrives at a show page (ie. from `/shows` or the search bar). Is that the intended behavior, or should we only show it if a user arrives at a show page from a fair?

src/v2/Apps/Show/Components/ShowNavigationBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { name, fair, isFairBooth, partner } = show
+
+  if (!fair?.href) {
+    return null
+  }
+
+  return (
+    <BackLink {...rest} to={`${fair.href}/exhibitors`}>
+      Back to {name}
+      {!isFairBooth && partner?.name && `at ${partner.name}`}
```
iskounen, line 24: **suggestion:** This implementation matches the [design spec](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3474%3A17827). However it seems odd that a user who is on the "MOVART at African Galleries Now 2021" show page is being asked to go "Back to MOVART at African Galleries Now 2021."

I feel like the text should be "Back to African Galleries Now 2021" instead, and the show name should be omitted from this back link.

iskounen (commented):
I have a few questions about when this back-link should be displayed and whether the copy of the back-link is accurate.

src/v2/Apps/Show/Components/ShowHeader.tsx
tr-ann: When we use comments like `@ts-expect-error STRICT_NULL_CHECK` we just bypass `STRICT_NULL_CHECK` errors, we don't guarantee that variables are not null/undefined. In this case the changes also used to bypass the error. Just another way to avoid linter errors

dzucconi: The errors are still valid in this case though. This hook is expecting a start/end time and I'm sure there are shows in the database that lack one or both. Somewhat problematic here in that we can't conditionally call hooks. Your best bet would be to refactor the hook to deal with the cases when one or both of these are missing (possibly just returning a null for that field, etc)

src/v2/Apps/Show/ShowApp.tsx
dzucconi: I see that in the component it's checking for the fair + href and returning null but I concur that's weird; it feels like the check should be in here rather than the other way around.

iskounen (commented):
These changes look good to me. I think it would also be helpful to write a test for this behavior.

src/v2/Apps/Show/ShowApp.tsx
```diff
@@ -52,11 +56,17 @@ export const ShowApp: React.FC<ShowAppProps> = ({ show }) => {
...
           }}
         >
-          <Box mt={4} mb={[4, 6]}>
-            <ShowInstallShots show={show} mt={4} mb={6} />
-          </Box>
+          {arrivedFromFair && (
+            <ShowNavigationBanner show={show} mt={2} mb={4} />
+          )}
+
+          {show?.images!.length > 0 && (
```
mdole, line 63: #minor is there a reason we're using `images!` instead of `images?`? I'm not sure if it's possible for a show to have no associated images, but I think this would crash if so

mdole (approved):
nice! agreed that a test would be good to have, otherwise I think this is ready to go

src/v2/Apps/Show/ShowApp.tsx
tr-ann: "!" using to avoid null_check error. We always get array (empty or with items). If we use "?" we should cast null/unndefined types to number to bypass linter error

src/v2/Apps/Show/Components/ShowNavigationBanner.tsx
```diff
@@ -0,0 +1,36 @@
...
+  show,
+  ...rest
+}) => {
+  const { fair } = show
+
+  return (
+    <BackLink {...rest} to={`${fair?.href}/exhibitors`}>
+      Back to {fair?.name}
+    </BackLink>
+  )
```
dimatretyak, line 21: ** thought**: maybe it's better not to render the BackLink at all, if not fair.name?

dimatretyak (approved):

mdole (approved):
sweet! let's merge!

PR #8402, opened 2021-09-15
feat(FX-3303): Added Read More functionality

anastasiapyzhik:
[FX-3303]

Before

Screenshot 2021-09-15 at 16 05 29

After

Screenshot 2021-09-15 at 16 07 04

[FX-3303]: https://artsyproduct.atlassian.net/browse/FX-3303


dimatretyak (approved):
LGTM 👍

mdole (approved):
great! good catch about the `about` data too, I didn't realize we used that field to collect all of the fair data together.

PR #8409, opened 2021-09-16
fix(FX-3231): Removed max width of fair article

anastasiapyzhik:
[FX-3231]

This fixes broken layout of articles on fair organizer page at large screen width 

**Before**

Screenshot 2021-09-16 at 14 37 51

Screenshot 2021-09-16 at 16 17 28

**After**

Screenshot 2021-09-16 at 16 17 53

Screenshot 2021-09-16 at 16 18 12

[FX-3231]: https://artsyproduct.atlassian.net/browse/FX-3231

### DIFF
--- a/src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
+++ b/src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
@@ -94,7 +94,7 @@ export const FairEditorialItem: React.FC<FairEditorialItemProps> = ({
   }
 
   return (
-    <Box maxWidth={image.width}>
+    <Box>
       {/* Devided link into separate parts in order to avoid linking via empty
         space when responsive box is applied */}
       <ItemLink>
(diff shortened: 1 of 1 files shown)


dzucconi (approved):

anastasiapyzhik:
Reverting this because it breaks articles layout on another page

Screenshot 2021-09-17 at 13 55 36

dzucconi:
IMO those should be two different components.

anastasiapyzhik:
> IMO those should be two different components.

Those are both fair articles looking similarly with slight differences. Am I missing some reasons to separate them?

dzucconi:
Different image sizes and text behavior.

anastasiapyzhik:
@dzucconi different image sizes might be handled by passing prop `size` and text just takes all available space so we can adjust the size of container by passing `Box` props to the component. This is [related PR](https://github.com/artsy/force/pull/8420).

I think these differences are too small to separate the component, we would have to duplicate a lot of code

PR #8414, opened 2021-09-16
refactor: replaces old buyers premium dialog

dzucconi:
So there are three modal dialogs referenced in the `legacyAssetDll`

1. `openCollectorFAQModal` — this one isn't actually called anywhere anymore and can be deleted.
2. `openAuctionBuyerPremium` — that's what this PR concerns. It fetches the sale and then renders the schedule for the buyer's premium at different hammer prices.
3. `openAuctionFAQModal` — this one is currently disabled. It was disabled when it we realized the entire 'multiPageModal' thing doesn't work anymore. I'm going to see if I can just get the data and re-build it statically since it seems like this is actually something we want.


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCurrentBidInfo.tsx
```diff
@@ -14,171 +11,162 @@ import {
   Text,
   WinningBidIcon,
 } from "@artsy/palette"
+import { useDialog } from "v2/Utils/Hooks/useDialog"
+import { AuctionBuyersPremiumDialogQueryRenderer } from "v2/Components/AuctionBuyersPremiumDialog"
 
 export interface ArtworkSidebarCurrentBidInfoProps {
   artwork: ArtworkSidebarCurrentBidInfo_artwork
 }
```
dzucconi, line 20: Most of the diff here is just converting this component to an FC.

src/v2/Utils/Hooks/useDialog.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+import { useState } from "react"
+
+export interface UseDialogProps {
+  Dialog: React.ComponentType<any>
+  onShow?(): void
+  onHide?(): void
+}
+
+export const useDialog = ({ Dialog, onShow, onHide }: UseDialogProps) => {
```
damassi, line 10: Since we have a few modals and dialogs in Palette, should we move this hook over there?

damassi (approved):
One question, but otherwise looks great 👍

src/v2/Utils/Hooks/useDialog.tsx
dzucconi: The legacy modal might do this but it's weird and has been really hard to kill. At some point I need to implement a dialog based on `ModalBase` to replace that one in which case this could move over, yeah.

PR #8410, opened 2021-09-16
feat[FX-3291]: add MetaTags to the FairOrganizerApp

araujobarret:
Description
Add the missing MetaTags of the Fair Organizer app.

[Ticket](https://artsyproduct.atlassian.net/browse/FX-3291)

Screenshot
image

Validated with the Twitter URL checked and used [ngrok](https://ngrok.com/) as the helper tool.

PR #8419, opened 2021-09-17
fix(FX-3291): adjust meta description on FairOrganizerApp

araujobarret:
Related to #8410 

- Adjust the description to use truncated text from `about` in the `MetaTags`.
- Clean up test
- Remove Title not necessary anymore

Screenshot 2021-09-17 at 10 17 25

Now the description makes more sense to me, but would like to hear about it 👍 

PS: followed the Twitter recommendation to not go over 200 characters on the description.

### DIFF
--- a/src/v2/Apps/FairOrginizer/FairOrganizerApp.tsx
+++ b/src/v2/Apps/FairOrginizer/FairOrganizerApp.tsx
@@ -1,6 +1,7 @@
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { Box, Spacer, Title } from "@artsy/palette"
+import truncate from "trunc-html"
+import { Box, Spacer } from "@artsy/palette"
 import { FairOrganizerApp_fairOrganizer } from "v2/__generated__/FairOrganizerApp_fairOrganizer.graphql"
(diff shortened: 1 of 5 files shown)


src/v2/Apps/FairOrginizer/FairOrganizerApp.tsx
```diff
@@ -56,6 +55,7 @@ export const FairOrganizerAppFragmentContainer = createFragmentContainer(
       fragment FairOrganizerApp_fairOrganizer on FairOrganizer {
         name
         slug
+        about(format: HTML)
```
mdole, line 58: Are other formats available? I worry a little that me might end up having tags or escaped characters in our descriptions

mdole (approved):
Nice! One question on the format of the description text. Assigning back to you to merge once you feel confident about the descriptions :)

src/v2/Apps/FairOrginizer/FairOrganizerApp.tsx
araujobarret: They are available but we use [truncate](https://www.npmjs.com/package/trunc-html) to not escape the tags 😬 
I avoided using plain because one of the fragments uses `about` with HTML, so would cause a conflict. Anyways `PLAIN` also has special characters to be removed with truncate, I can explore changing to it if you want, np

PR #8420, opened 2021-09-17
fix(FX-3231): Aligned articles layout

anastasiapyzhik:
[FX-3231]

**Before**

staging artsy net_fair-organizer_the-armory-show_articles_page=8 (4)

**After**

localhost_5000_fair-organizer_the-armory-show_articles_page=8 (2)

[FX-3231]: https://artsyproduct.atlassian.net/browse/FX-3231

### DIFF
--- a/src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
+++ b/src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx
@@ -1,6 +1,5 @@
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { useTracking } from "react-tracking"
 import { FairEditorialItem_article } from "v2/__generated__/FairEditorialItem_article.graphql"
 import {
   Box,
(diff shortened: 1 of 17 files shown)


dimatretyak (approved):
LGTM 👍

dimatretyak (approved):
LGTM 👍

dzucconi (changes requested):
A few things here:

* Avoid declaring components within the function body. _Particularly_ with images. Ever re-render will cause a flicker because of how images are loaded.
* The image should be resized according to the max dimensions it can be displayed at. This will be different for each size.

dimatretyak:
> The image should be resized according to the max dimensions it can be displayed at. This will be different for each size.
If it's not difficult, can you explain it in more detail?

dzucconi:
Ah, sorry I missed this: https://github.com/artsy/force/blob/09cd0cb3e686e1053113b3edb5e32729e75f48bc/src/v2/Apps/Fair/Components/FairEditorial/FairEditorialItem.tsx#L148-L159

So just inlining the component declarations should be fine.

dzucconi (approved):

PR #8426, opened 2021-09-17
fix(GRO-509): Update homepage banner copy

jacobherrington:


dzucconi (approved):
😗👌

dzucconi:
[https://danger.systems/js/](Danger) is complaining about not having a description. For something as simple as this maybe just a link to the JIRA ticket would be appropriate. Not sure what the other errors there are about. Merging! Thanks

damassi:
> For something as simple as this maybe just a link to the JIRA ticket would be appropriate.

Even though this is simple, as a best practice we should avoid bare links to GH tickets as that makes our GH papertrail dependent on a 3rd party (which we could switch away from in the future), while additionally limiting search functionality.

jacobherrington:
@damassi I agree. In this case, I think the commit message (PR title) conveys any information someone might need.

I tend to bias toward keeping as much information in the commits vs the PR for the same reason. I'll definitely keep this in mind as I do more feature development and work in Force.

PR #8433, opened 2021-09-20
feat: [Home] Add trending artists & auction lots rail

damassi:
Addresses 
- https://artsyproduct.atlassian.net/browse/GRO-524
- https://artsyproduct.atlassian.net/browse/GRO-520

This adds a new Trending Artists rail to the home page: 

Screen Shot 2021-09-20 at 4 19 44 PM

As well as the Auction Lots rail: 

Screen Shot 2021-09-20 at 5 28 15 PM

### DIFF
--- a/src/v2/Apps/Home/Components/HomeAuctionLotsRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeAuctionLotsRail.tsx
@@ -1,0 +1,160 @@
+import {
+  Box,
+  Text,
+  Flex,
+  Spacer,
+  Shelf,
+  Skeleton,
+  SkeletonText,
+  SkeletonBox,
+  Sup,
+} from "@artsy/palette"
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { useSystemContext } from "v2/System"
+import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
+import { RouterLink } from "v2/System/Router/RouterLink"
+import { HomeAuctionLotsRail_viewer } from "v2/__generated__/HomeAuctionLotsRail_viewer.graphql"
+import { HomeAuctionLotsRailQuery } from "v2/__generated__/HomeAuctionLotsRailQuery.graphql"
+import { extractNodes } from "v2/Utils/extractNodes"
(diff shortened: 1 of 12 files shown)


src/v2/Apps/Home/Components/HomeAuctionLotsRail.tsx
```diff
@@ -0,0 +1,162 @@
...
+  if (nodes.length === 0) {
+    return null
+  }
+
+  return (
+    <HomeAuctionLotsRailContainer lotCount={nodes.length}>
+      <Shelf>
+        {nodes.map((node, index) => {
+          return (
+            <Box>
```
dzucconi, line 42: Why the extra `Box`? Would have to move the `key` up if there's a reason

src/v2/Apps/Home/Components/HomeAuctionLotsRail.tsx
```diff
@@ -0,0 +1,162 @@
...
+              <ShelfArtworkFragmentContainer
+                artwork={node}
+                key={node.slug}
+                // TODO: Add home type to cohesion once we have tracking
+                contextModule={null as any}
+                hidePartnerName
+                // FIXME: For some reason this is leaves the gray loading
+                // background post load
```
dzucconi, line 50: That's always there (it's on my grievances list). But why is it a problem in this case? It should only matter for transparent images.

src/v2/Apps/Home/Components/HomeTrendingArtistsRail.tsx
```diff
@@ -0,0 +1,196 @@
...
+                <Spacer mt={1} />
+
+                <EntityHeader
+                  name={node.name!}
+                  meta={node.formattedNationalityAndBirthday!}
+                  smallVariant
+                  FollowButton={
+                    <FollowArtistButtonFragmentContainer
+                      user={user}
+                      artist={node as any}
```
dzucconi, line 73: Why's this get cast as `any`?

src/v2/Apps/Home/Components/HomeAuctionLotsRail.tsx
damassi: Wrong width set for image, fixed

PR #8444, opened 2021-09-21
fix(inquiry): inquiry q/a

dzucconi:
Along with the recent Palette updates, this PR handles the majority of [the Q/A items](https://www.notion.so/artsy/Artwork-Inquiry-Form-QA-7102f7c454a54003b5263ed13fa8a54e). I'm going to take a look at GDPR separately (which AFAIK we don't address in the current implementation either).


src/desktop/components/location_search/index.styl
```diff
@@ -16,41 +16,3 @@
...
-  &.pac-item-selected
-    border-color transparent
-    background-color #000
-    > .pac-item-query
-      color #fff
-
-.pac-item,
-.pac-item-query
-  garamond()
-  font-size default-garamond-font-size
```
dzucconi, line 56: Old instances of this will just use the default Google styling, which is fine. The new component has styling that matches the rest of our autocompletes.

src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
```diff
@@ -1,55 +1,56 @@
...
 > & {
-  relayEnvironment?: Environment
   artworkID: string
   message: string
   contactGallery?: boolean | null
 }
 
 export const useArtworkInquiryRequest = () => {
-  const { relayEnvironment: defaultRelayEnvironment } = useSystemContext()
+  const { relayEnvironment } = useInquiryContext()
```
damassi, line 18: Why do we pull the `relayEnvironment` from inquiry context rather than system context?

dzucconi: This gets set in login/sign up. We authenticate the environment and the one in system context is still logged out.

src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -116,6 +122,15 @@ const InquiryProvider: React.FC<InquiryProviderProps> = ({
     onDone: onClose,
   })
 
+  const { relayEnvironment: defaultRelayEnvironment } = useSystemContext()
+
+  const relayEnvironment = useRef(defaultRelayEnvironment!)
+
+  const setRelayEnvironment = (updatedEnvironment: Environment) => {
```
damassi, line 129: A comment here describing why this is necessary might be helpful for later

src/v2/Components/Inquiry/Views/InquiryLogin.tsx
```diff
@@ -51,10 +57,15 @@ export const InquiryLogin: React.FC = () => {
     try {
       const { user } = await login({ email: inquiry.email!, ...state })
 
+      const relayEnvironment = createRelaySSREnvironment({ user })
```
damassi, line 60: Why is this necessary? (Code comment here would be good since we don't have to do this anywhere else in the app)

damassi (approved):
Just a couple non-blocking questions / suggestions. Looks good 🍏

src/v2/Components/Inquiry/Hooks/useArtworkInquiryRequest.ts
damassi: Another thought is: can we move this behavior up to the system context itself and add the function there? (This would resolve other questions in PR, as code comments there would be unnecessary.) Thinking through it, currently when we log out we do a hard reload, but in the future that doesn't have to be so, and in the meantime we could at least provide a mechanism to refresh relay's environment from where it's typically accessed.

dzucconi: Explored that a bit; it won't work because we'd need to return the environment ref rather than the environment itself. Meaning it'd lead to a much larger refactor. It's a good idea to figure this out but I'm just going to add some comments here for now.

damassi: Np 👍

src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -126,6 +127,11 @@ const InquiryProvider: React.FC<InquiryProviderProps> = ({
 
   const relayEnvironment = useRef(defaultRelayEnvironment!)
 
+  /**
+   * In this flow we go from a logged out state to a logged in/signed up state
+   * where we have to execute mutations like sending the inquiry, saving your
+   * information, etc. We store the Relay environment in a ref then update it here.
+   */
```
damassi, line 134: 👍

PR #8446, opened 2021-09-22
fix: Expand auction house filter by default

olerichter00:
## Description

Expands the `auction house` filter by default to show the filter to improve the UX when the user searches for auction results on the new price database landing page.

Screenshot 2021-09-22 at 11 08 43


brainbicycle (approved):
🪗 !

PR #8447, opened 2021-09-22
fix(PURCHASE-2923): add zIndex to toast component

tam-kis:
[PURCHASE-2923]
Added zIndex to prevent other elements from overlapping the toast component. If overlapping happens in the future, should be enough to increase the zIndex number.

Before:
Screenshot 2021-09-22 at 12 33 52

After:
Screenshot 2021-09-22 at 12 35 25

[PURCHASE-2923]: https://artsyproduct.atlassian.net/browse/PURCHASE-2923

### DIFF
--- a/src/v2/Components/Toast/ToastComponent.tsx
+++ b/src/v2/Components/Toast/ToastComponent.tsx
@@ -53,12 +53,12 @@ const ToastComponent: React.FC<ToastComponentProps> = ({
         boxShadow:
           "0px 0px 1px 2rgba (0, 0, 0, 0.08), 0px 2px 8px rgba(0, 0, 0, 0.12)",
         left: isMoveToRight ? "100%" : 0,
-        zIndex: 1,
       }}
       borderRadius={2}
       position="fixed"
       left={0}
       width="100%"
+      zIndex={1}
       opacity={showNotification ? 1 : 0}
       onTouchStart={handleTouchStart}
       onTouchMove={handleTouchMove}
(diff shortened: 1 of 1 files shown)


src/v2/Components/Toast/ToastComponent.tsx
```diff
@@ -53,6 +53,7 @@ const ToastComponent: React.FC<ToastComponentProps> = ({
         boxShadow:
           "0px 0px 1px 2rgba (0, 0, 0, 0.08), 0px 2px 8px rgba(0, 0, 0, 0.12)",
         left: isMoveToRight ? "100%" : 0,
+        zIndex: 1,
```
jo-rs, line 56: I think that the `Box` component takes a `zIndex` prop directly, which we prefer doing over adding to the style prop. So, adding a `zIndex={1}` directly should do the trick.

jo-rs (changes requested):
Congrats on your first Purchase PR; nice and easy solution! 💯 
Just a small comment regarding our preferred approach. Give it a shot and see if it works as well 😄

jo-rs:
Also, we have a few hooks set up for Jira and PRs (that sometimes don't work terribly well 😅 ) so we add the ticket number to the PR body for that and future reference. Just including "[PURCHASE-2923]" on the body, should automatically link to it.

src/v2/Components/Toast/ToastComponent.tsx
tam-kis: Updated to zIndex as a prop.

jo-rs (approved):
Woohoo, nice! 🚀 
Checking what's happening with the build.

PR #8448, opened 2021-09-22
remove clickedArtworkGroup helpers

pvinis:
Removing the use of clickedArtworkGroup helpers, after the cohesion RFC to remove helpers.

Sibling PR https://github.com/artsy/cohesion/pull/236

### DIFF
--- a/src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
+++ b/src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
@@ -9,9 +9,11 @@ import { AuctionArtworksRailPlaceholder } from "../AuctionArtworksRailPlaceholde
 import { tabTypeToContextModuleMap } from "../../Utils/tabTypeToContextModuleMap"
 import { useTracking } from "react-tracking"
 import {
-  clickedArtworkGroupHeader,
-  ClickedArtworkGroupHeaderArgs,
+  ActionType,
+  ClickedArtworkGroup,
+  ContextModule,
   OwnerType,
+  PageOwnerType,
 } from "@artsy/cohesion"
 import { useAnalyticsContext } from "v2/System"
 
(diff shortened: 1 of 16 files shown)


src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
```diff
@@ -9,9 +9,11 @@ import { AuctionArtworksRailPlaceholder } from "../AuctionArtworksRailPlaceholde
 import { tabTypeToContextModuleMap } from "../../Utils/tabTypeToContextModuleMap"
 import { useTracking } from "react-tracking"
 import {
-  clickedArtworkGroupHeader,
-  ClickedArtworkGroupHeaderArgs,
+  ActionType,
+  ClickedArtworkGroup,
```
damassi, line 13: Is it ok that the helper types are still here? This is pretty confusing

src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
```diff
@@ -113,3 +112,20 @@ export const AuctionArtworksRailFragmentContainer = createFragmentContainer(
...
   }
 )
+
+const tracks = {
+  clickedArtworkGroupHeader: (
+    contextModule: ContextModule,
+    contextPageOwnerType: PageOwnerType,
+    saleID: string,
+    saleSlug: string
+  ): ClickedArtworkGroup => ({
```
damassi, line 122: this should either be removed or defined in force... or?

src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
pvinis: That's the schema type. small `c` was the helper. The schema types make sense to be there, to verify we send the right things.

src/v2/Apps/Auctions/Components/AuctionArtworksRail/AuctionArtworksRail.tsx
pvinis: You mean the schema type? That's here https://github.com/artsy/cohesion/blob/d9f09f800282a372e2e182a407b42ee153181491/src/Schema/Events/Click.ts#L164.

mdole (approved):
LGTM!

mdole:
Oh, although...we should probably introduce regression tests in the same pattern as with this Eigen PR, right? https://github.com/artsy/eigen/pull/5492

PR #8449, opened 2021-09-22
feat(FX-3280): navigation banner at fair show page

tr-ann:
Jira: [FX-3280](https://artsyproduct.atlassian.net/browse/FX-3280)

### Description
As a collector, I should be able to go back to the Fairs 'Exhibitors A-Z' tab after clicking through to view a Fairs 'Show'. Currently, there is no navigational banner making it explicit and easy for the collector to go back to the Fair page to continue browsing exhibitors within a Fair.

[Figma](https://www.figma.com/file/aM3b5sodn04vMTigbQXACi/Fairs-2021?node-id=3474%3A17827)

### Changes
- Added navigation banner
- Small refactor of fair components

### Demo

https://user-images.githubusercontent.com/56556580/134356999-ce1b8cd1-9361-4b25-94d1-428fba855752.mov

https://user-images.githubusercontent.com/56556580/134358180-52d99e9a-77dd-4da2-888e-221a7d3611f1.mov

### DIFF
--- a/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
+++ b/src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
@@ -1,13 +1,12 @@
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import styled from "styled-components"
-import { Text, Flex, Swiper, themeProps } from "@artsy/palette"
+import { Text, Flex, Swiper } from "@artsy/palette"
 import { Media } from "v2/Utils/Responsive"
 import { scrollIntoView } from "v2/Utils/scrollHelpers"
 import { ExhibitorsLetterNav_fair } from "v2/__generated__/ExhibitorsLetterNav_fair.graphql"
-import { __internal__useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
-import { useNavBarHeight } from "v2/Components/NavBar/useNavBarHeight"
 import { getExhibitorSectionId } from "../Utils/getExhibitorSectionId"
+import { useExhibitorsTabOffset } from "../Utils/useExhibitorsTabOffset"
 
 const LETTERS = "ABCDEFGHIJKLMNOPQRSTUVWXYZ#".split("")
 
@@ -19,15 +18,7 @@ export const ExhibitorsLetterNav: React.FC<ExhibitorsLetterNavProps> = ({
   fair,
 }) => {
   const letters = fair?.exhibitorsGroupedByName?.map(group => group?.letter)
-  const {
-    height: [mobileNavBarHeight, desktopNavBarHeight],
-  } = useNavBarHeight()
-
-  const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
-  const stickyTabsHeight = 150
-
-  const offset =
-    (isMobile ? mobileNavBarHeight : desktopNavBarHeight) + stickyTabsHeight
+  const offset = useExhibitorsTabOffset()
 
   const Letters = ({ withSwiper = false }) => {
     return (
--- a/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
+++ b/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
@@ -1,9 +1,10 @@
 import React from "react"
-import { Column, GridColumns } from "@artsy/palette"
+import { Column, GridColumns, themeProps } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FairExhibitorsGroup_exhibitorsGroup } from "v2/__generated__/FairExhibitorsGroup_exhibitorsGroup.graphql"
 import { FairExhibitorCardFragmentContainer as FairExhibitorCard } from "./FairExhibitorCard"
 import { useRouter } from "v2/System/Router/useRouter"
+import { useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
 
 interface FairExhibitorsGroupProps {
   exhibitorsGroup: FairExhibitorsGroup_exhibitorsGroup
@@ -14,6 +15,7 @@ export const FairExhibitorsGroup: React.FC<FairExhibitorsGroupProps> = ({
 }) => {
   const { match } = useRouter()
   const { focused_exhibitor: focusedExhibitorID } = match.location.query
+  const isMobile = useMatchMedia(themeProps.mediaQueries.xs)
 
   const focusedExhibitorStyles = {
     borderColor: "brand",
@@ -30,7 +32,8 @@ export const FairExhibitorsGroup: React.FC<FairExhibitorsGroupProps> = ({
         }
 
         const focused = focusedExhibitorID === exhibitor.partner.internalID
-        const exhibitorCardStyle = focused ? focusedExhibitorStyles : {}
+        const exhibitorCardStyle =
+          focused && !isMobile ? focusedExhibitorStyles : {}
 
         return (
           <Column
--- a/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
+++ b/src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
@@ -3,6 +3,7 @@ import { Column, GridColumns } from "@artsy/palette"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FairExhibitorsGroup_exhibitorsGroup } from "v2/__generated__/FairExhibitorsGroup_exhibitorsGroup.graphql"
 import { FairExhibitorCardFragmentContainer as FairExhibitorCard } from "./FairExhibitorCard"
+import { useRouter } from "v2/System/Router/useRouter"
 
 interface FairExhibitorsGroupProps {
   exhibitorsGroup: FairExhibitorsGroup_exhibitorsGroup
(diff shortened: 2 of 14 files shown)


dzucconi:
Design related question: we have this: https://github.com/artsy/force/blob/master/src/v2/Components/TopContextBar.tsx

Which is used in a few places and looks like:



Should we be using that?

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
```diff
@@ -11,15 +13,34 @@ interface FairExhibitorsGroupProps {
 export const FairExhibitorsGroup: React.FC<FairExhibitorsGroupProps> = ({
   exhibitorsGroup: { exhibitors },
 }) => {
+  const { match } = useRouter()
+  const { focused_exhibitor: focusedExhibitorID } = match.location.query
+  const isMobile = useMatchMedia(themeProps.mediaQueries.xs)
```
dzucconi, line 18: This won't be safe for SSR

@damassi wondering how to discourage this usage?

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
```diff
@@ -11,15 +13,34 @@ interface FairExhibitorsGroupProps {
...
+
+  const focusedExhibitorStyles = {
+    borderColor: "brand",
+    borderStyle: "solid",
+    borderWidth: "1px",
+    padding: "2px",
+  }
```
dzucconi, line 25: What exactly is this? I don't see it in the Figma

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
```diff
@@ -11,15 +13,34 @@ interface FairExhibitorsGroupProps {
...
   return (
     <GridColumns position="relative" gridRowGap={4}>
       {exhibitors?.map(exhibitor => {
         if (!exhibitor?.partner) {
           return null
         }
 
+        const focused = focusedExhibitorID === exhibitor.partner.internalID
+        const exhibitorCardStyle =
+          focused && !isMobile ? focusedExhibitorStyles : {}
```
dzucconi, line 36: For mobile styles you'd just use responsive props instead of the hook

dzucconi: `borderColor: ["brand", undefined]` etc...

src/v2/Apps/Fair/Utils/useExhibitorsTabOffset.ts
```diff
@@ -0,0 +1,17 @@
...
+    height: [mobileNavBarHeight, desktopNavBarHeight],
+  } = useNavBarHeight()
+
+  const isMobile = useMatchMedia(themeProps.mediaQueries.xs)
+  const stickyTabsHeight = 150
+
+  const offset =
+    (isMobile ? mobileNavBarHeight : desktopNavBarHeight) + stickyTabsHeight
+
+  return offset
```
dzucconi, line 16: I guess we should just write a utility function to detect the sticky height and handle all this automatically since this has come up a few times.

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
tr-ann: All this changes deleted. We should not use the border.

src/v2/Apps/Fair/Utils/useExhibitorsTabOffset.ts
tr-ann: I think we should create a new ticket for this, do you?

dzucconi: 👍

tr-ann:
@dzucconi what about `TopContextBar`, we have not a separator below a link, so I think we should not use it
image

dzucconi:
@tr-ann yeah you're right. I guess I'm hinting at a larger discussion about whether or not we have a consistent pattern for this. Like there's a difference between an entity being 'part of' something larger and calling that out with `TopContextBar` vs a sub-page which has a back button.

tr-ann:
@dzucconi yes, you are right. Should we discuss it with Jonathan?

dzucconi:
I think it's sufficient to go with this PR and maybe just leave a comment in the Figma. Can always circle back. Show pages are another one which have this same back button pattern: https://www.artsy.net/show/moskowitz-gallery-transitive-horizon/info so it's consistent with that.

src/v2/Apps/Fair/Components/FairExhibitors/FairExhibitorsGroup.tsx
damassi: Yeah this is the like the 4th or 5th time i've seen it used within a week, having never really seen it be used before. Lemme see where else its being used in the codebase real quick and report back here.

damassi: ~6 spots: 
Screen Shot 2021-09-22 at 11 28 03 AM

We could refactor it out and delete, or mark it as `@deprecated` with a note about our other strategies. I'm fine with the latter as that big strike through will discourage use, but some people don't use VSCode. Thoughts?

damassi: Right right, its just that I can't think of a way to scare away folks other than that strike-through and / or renaming the function to something like `__internal__useMatchMedia` (or something)

dzucconi: Yeah: works for me; maybe just adding a note about that it's not _really_ deprecated... 🤔

tr-ann: `useMatchMedia` is no loger used in my chages, so it don't block the pr
сс/ @dzucconi @damassi

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
```diff
@@ -19,15 +18,7 @@ export const ExhibitorsLetterNav: React.FC<ExhibitorsLetterNavProps> = ({
...
-  const {
-    height: [mobileNavBarHeight, desktopNavBarHeight],
-  } = useNavBarHeight()
-
-  const isMobile = useMatchMedia(themeProps.mediaQueries.xs)
-  const stickyTabsHeight = 150
-
-  const offset =
-    (isMobile ? mobileNavBarHeight : desktopNavBarHeight) + stickyTabsHeight
+  const offset = useExhibitorsTabOffset()
```
mdole, line 21: Sorry if this is a dumb question, but I'm still working on understanding how SSR works in Force: is this component being rendered server-side? If so, I think we would want to avoid using this hook, right?

mdole (approved):
One question about hook usage, otherwise this is looking great!

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
tr-ann: I'm not sure, but I think you are right and we want to create a utility function instead of hook.
@dzucconi does it's right?

src/v2/Apps/Show/ShowApp.tsx
```diff
@@ -26,13 +26,20 @@ import {
...
   show: ShowApp_show
 }
 
 export const ShowApp: React.FC<ShowAppProps> = ({ show }) => {
   const { contextPageOwnerSlug, contextPageOwnerType } = useAnalyticsContext()
+  const { match } = useRouter()
+  const { from_fair } = match.location.query
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const arrivedFromFair = useMemo(() => from_fair ?? false, [])
```
dimatretyak, line 42: **suggestion**: I think it is not necessary to use `useMemo` here

```javascript
const arrivedFromFair = from_fair ?? false
```

dimatretyak (approved):

src/v2/Apps/Show/ShowApp.tsx
mdole: I wondered about that too - I think the idea is that we don't need to recompute `arrivedFromFair` on every re-render because `from_fair` will not change the whole time we're on the page. So seems like a slight performance optimization?

dimatretyak: There is no point in optimization here, since either `true` or `false` is expected all the time

tr-ann: When we use `useMemo` we compute `arrivedFromFair` only one time and then even if user click "Back to _Fair_" the banner are not disappeared

dimatretyak: Aah, I totally forgot about it. Thank you for the reminder!

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
dzucconi: Here it's not being used to _render_ anything so it's safe. That's sort of the distinction here: it's not SSR-safe but that's with regard to the actual rendering. Here it's being used to scroll to some position on screen after-mount; so all of that happens on the client-side.

dzucconi: But yeah, I had mentioned we should just do this kind of thing _once_ everywhere since we're never not going to want to account for the sticky-stack when scrolling to a position. (Which reminds me to ticket it)

dzucconi (approved):

src/v2/Apps/Fair/Components/ExhibitorsLetterNav.tsx
mdole: thanks, that helps my understanding a lot!

PR #8466, opened 2021-09-23
bugfix: [Home] Fix homepage module display

damassi:
A bit of fallout from https://github.com/artsy/force/pull/8455. 

Fixes issue where we updated module string type to a union, but didn't `toUpperCase` on the module key when making a request to MP, leading to a type error. 

Example of auth'd rails now coming through: 

Screen Shot 2021-09-23 at 1 15 09 PM


dzucconi (approved):
🙏

PR #8471, opened 2021-09-23
fix: [Router] Fix cannot update state after unmount error + cleanup

damassi:
Noticed that after our React 17 upgrade we started getting "cannot set state after unmount" errors when transitioning between routes. Wrapping our state calls with `setImmediate` fixes it. 

Also cleans up all of the `// @ts-expect-error` flags in `v2/System/Router`.

### DIFF
--- a/src/v2/System/Router/RenderStatus.tsx
+++ b/src/v2/System/Router/RenderStatus.tsx
@@ -22,8 +22,7 @@ export const RenderPending = () => {
    * duration of the fetch.
    */
   if (!isFetching) {
-    // @ts-expect-error STRICT_NULL_CHECK
-    setFetching(true)
+    setImmediate(() => setFetching?.(true))
   }
 
   if (isFetching) {
@@ -50,15 +49,11 @@ export const RenderPending = () => {
   }
 }
 
-// @ts-expect-error STRICT_NULL_CHECK
-export const RenderReady: React.FC<{
-  elements: React.ReactNode
-}> = props => {
+export const RenderReady = (props: { elements: React.ReactNode }) => {
   const { isFetching, setFetching } = useSystemContext()
 
   if (isFetching) {
-    // @ts-expect-error STRICT_NULL_CHECK
-    setFetching(false)
+    setImmediate(() => setFetching?.(false))
   }
 
   if (!isFetching) {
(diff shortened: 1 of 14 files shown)


src/v2/System/Router/RenderStatus.tsx
```diff
@@ -50,15 +49,11 @@ export const RenderPending = () => {
   }
 }
 
-// @ts-expect-error STRICT_NULL_CHECK
-export const RenderReady: React.FC<{
-  elements: React.ReactNode
-}> = props => {
+export const RenderReady = (props: { elements: React.ReactNode }) => {
```
damassi, line 52: There's an unfortunate property about `found` where in order to trigger certain router behaviors one has to return `undefined` (vs `null`). If we're currently fetching we return `undefined` here, which keeps the render pending component running smoothly and displaying the top loading indicator bar.

src/v2/System/Router/RenderStatus.tsx
```diff
@@ -22,8 +22,7 @@ export const RenderPending = () => {
    * duration of the fetch.
    */
   if (!isFetching) {
-    // @ts-expect-error STRICT_NULL_CHECK
-    setFetching(true)
+    setImmediate(() => setFetching?.(true))
```
damassi, line 25: 🙏 @olerichter00 for the idea from your other PR

src/v2/System/Router/Utils/findCurrentRoute.tsx
```diff
@@ -1,25 +1,17 @@
-import { Match } from "found"
-import { AppRouteConfig } from "../Route"
-
-export const findCurrentRoute = ({
-  route: baseRoute,
-  routes,
-  routeIndices,
-}: Match & { route?: AppRouteConfig }) => {
+export const findCurrentRoute = (
+  { route: baseRoute, routes, routeIndices }: any /* FIXME: Add type */
```
damassi, line 2: I took the type off of here because sometimes all the finagling with `!` and `?` and so on just makes things even more unsafe. This file was a particularly bad offender. (I welcome another set of eyes.)

dzucconi: What type does this wind up returning?

damassi: It returns a [Route](https://github.com/4Catalyzer/found/blob/ac5b727d094588992743fbe9b3ef1343483f870f/types/index.d.ts#L213)

dzucconi (approved):

PR #8461, opened 2021-09-23
Revert "Refactor: convert AddressForm to functional component to and add v3 styles"

starsirius:
Reverts artsy/force#8387

https://artsy.slack.com/archives/C9YNS4X32/p1632408694257600

PR #8472, opened 2021-09-24
Refactor: convert AddressForm to functional component to and add v3 styles

olerichter00:
Reverts artsy/force#8461 to restore artsy/force#8387.

https://artsy.slack.com/archives/C9YNS4X32/p1632408694257600


src/v2/Apps/Order/Components/__tests__/PaymentPicker.jest.tsx
```diff
@@ -230,32 +230,6 @@ describe("PaymentPickerFragmentContainer", () => {
     expect(page.addressFormIsVisible).toBe(true)
   })
 
-  it("removes all data when the billing address form is hidden", async () => {
```
olerichter00, line 233: @lilyfromseattle Removing the whole test because we don't remove the data anymore when the billing address form is hidden.

PR #8475, opened 2021-09-24
fix(PX-4574): fix Arta error message

kiryl-zubarau:
This PR fixes an error message that we show if something went wrong on the Arta side.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4574

image


ansor4 (approved):

jpotts244:
Very nice, thank you for this @kiryl-zubarau !

PR #8476, opened 2021-09-24
fix: Adjust price database tracking

olerichter00:
## Description

Moves price database screen tracking into `useEffect` (which should fix the flaky tests as well) and adds keys to `PopularArtistsList`.

### DIFF
--- a/src/v2/Apps/PriceDatabase/PriceDatabase.tsx
+++ b/src/v2/Apps/PriceDatabase/PriceDatabase.tsx
@@ -1,5 +1,5 @@
 import { Flex, Separator, Spacer } from "@artsy/palette"
-import React from "react"
+import React, { useEffect } from "react"
 import { useTracking } from "react-tracking"
 import { PriceDatabaseBenefits } from "./Components/PriceDatabaseBenefits"
 import { PriceDatabaseSearch } from "./Components/PriceDatabaseSearch"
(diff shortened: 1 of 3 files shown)


src/v2/Apps/PriceDatabase/PriceDatabase.tsx
```diff
@@ -11,11 +11,13 @@ import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
...
-    context_module: ContextModule.priceDatabaseLanding,
-    context_screen_owner_type: OwnerType.priceDatabase,
-  })
+  useEffect(() => {
+    trackEvent({
+      action: ActionType.screen,
+      context_module: ContextModule.priceDatabaseLanding,
+      context_screen_owner_type: OwnerType.priceDatabase,
+    })
+  }, [])
```
olerichter00, line 20: @MounirDhahri I'm not quite sure if this is the correct way to do screen tracking.

MounirDhahri: this is a good catch
I am also not sure if that's the way to do screen tracking here but I think's fine for now

olerichter00: @damassi is there a better way to do screen tracking in Force?

PR #8478, opened 2021-09-24
fix: add some space between shipping options and button

pepopowitz:
Adds a spacer after the shipping options. The size aligns with the other spaces on this form.

Addresses [PX-4594]

Before:

image

After:

image

[PX-4594]: https://artsyproduct.atlassian.net/browse/PX-4594


annacarey:
Yay so much better! Generally, I don't worry about mobile so much in CMS so I think we can merge. And I'm sure it's fine anyways. Merging!

PR #8479, opened 2021-09-24
feat(gro-517): improved sticky

dzucconi:
I'm going to let this bake in draft over the weekend and take a closer look on Monday; but here's the general idea:

You can see the two sentinels as red lines. The top one corresponds with when the content hits the bottom edge of the nav, the bottom sentinel corresponds with when the bottom edge of the content hits the bottom edge of its container.  The gap between them is the distance from the bottom edge of the content to the bottom edge of its container. You can see that when resized wide enough such that the content is as tall or taller than the right-hand side, nothing happens: ie: when both sentinels are in the screen at the same time, we don't need to stick anything.

https://user-images.githubusercontent.com/112297/134717009-9ad6621f-a600-438f-bc18-9f0e0021d604.mov

Now the problem I just ran into with integration is that we don't actually want to start sticking immediately in some instances, we want to wait to get to the bottom of the content then stick with that offset. This only manifests itself as a possibility when the content you want to stick is taller than the entire screen. 😅

### DIFF
--- a/src/v2/Components/Sticky/Sticky.story.tsx
+++ b/src/v2/Components/Sticky/Sticky.story.tsx
@@ -1,4 +1,12 @@
-import { Box, FullBleed, Separator, Spacer, Text } from "@artsy/palette"
+import {
+  Box,
+  Column,
+  FullBleed,
+  GridColumns,
+  Separator,
+  Spacer,
+  Text,
+} from "@artsy/palette"
 import React from "react"
 import { AppContainer } from "v2/Apps/Components/AppContainer"
 import { HorizontalPadding } from "v2/Apps/Components/HorizontalPadding"
@@ -10,9 +18,9 @@ export default {
   title: "Components/Sticky",
 }
 
-export const Example = () => {
+const HeaderPlaceholder: React.FC = () => {
   return (
-    <StickyProvider>
+    <>
       <Box
         position="fixed"
         top={0}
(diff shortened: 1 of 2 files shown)


src/v2/Components/Sticky/Sticky.story.tsx
```diff
@@ -61,67 +79,72 @@ export const Example = () => {
...
+  return (
+    <AppContainer>
+      <HorizontalPadding>
+        <HeaderPlaceholder />
+
+        <Filler />
+
+        <GridColumns my={2}>
+          <Column span={6}>
+            <Sticky proportional contained>
```
damassi, line 105: 👍

src/v2/Components/Sticky/Sticky.tsx
```diff
@@ -1,99 +1,252 @@
...
- *      Everything that is stuck is the full width of the screen. See the
- *      `GridExample` story for an example that works around this by re-wrapping
- *      everything in an AppContainer + HorizontalPadding. (This is non-ideal)
- *
- * - Q: How do I unstick an element once it reaches the end of its parent container?
- * - A: This is also not currently supported.
  */
-export const Sticky: React.FC = ({ children }) => {
-  const sentinelRef = useRef<HTMLDivElement | null>(null)
+export const Sticky: React.FC<StickyProps> = ({
```
damassi, line 50: One Q is: should this go in Palette? Seems appropriately lib-ish?

dzucconi: Well right now it's got hardcoded assumptions about our nav in Force. Those [can/will go away](https://github.com/artsy/force/pull/7778) but for now we'll have to keep it here.

PR #8480, opened 2021-09-24
chore: More // @ts-expect-error cleanup

damassi:
Following up on work in https://github.com/artsy/force/pull/8471, this cleans up more `// @ts-expect-error` strict null-check migration flags (this time from the entire `System` folder) and cleans up some long-standing type issues around our `User` type.

### DIFF
--- a/src/typings/gravity.d.ts
+++ b/src/typings/gravity.d.ts
@@ -1,14 +1,17 @@
-interface User {
-  roles?: UserRole[]
-  accessToken?: string
-  appToken?: string
-  email?: string
-  has_partner_access?: string
-  id?: string
-  lab_features?: string[]
-  type?: string
-  name?: string
-}
+type User =
+  | {
+      roles?: UserRole[]
+      accessToken?: string
+      appToken?: string
+      email?: string
+      has_partner_access?: string
+      id?: string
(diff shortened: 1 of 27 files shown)


src/typings/gravity.d.ts
```diff
@@ -1,14 +1,17 @@
...
-  name?: string
-}
+type User =
+  | {
+      roles?: UserRole[]
+      accessToken?: string
+      appToken?: string
+      email?: string
+      has_partner_access?: string
+      id?: string
```
dzucconi, line 8: If the User is exist-y; then I think they are _always_ going to have an `id`

src/typings/gravity.d.ts
```diff
@@ -1,14 +1,17 @@
...
+      lab_features?: string[]
+      type?: string
+      name?: string
+    }
+  | null
+  | undefined
```
dzucconi, line 14: 🙏

dzucconi (approved):

src/typings/gravity.d.ts
damassi: Ah shoot, that blew up all over 😬 gonna just keep it how it was for now

PR #8481, opened 2021-09-24
refactor: Rename useMatchMedia to __internal__useMatchMedia

damassi:
Addresses https://artsyproduct.atlassian.net/browse/WP-71 

This has come up quite a bit lately, where folks have been relying on this client-side-only, SSR-incompatable hook for writing responsive code. Since Palette's responsive props and fresnel's `<Media>` approaches will cover the vast majority of responsive use-cases, we're going to rename this hook in order to discourage use as well as add some documentation around where devs should go if they encounter it.

### DIFF
--- a/src/v2/Utils/Hooks/useMatchMedia.ts
+++ b/src/v2/Utils/Hooks/useMatchMedia.ts
@@ -1,6 +1,14 @@
 import { useEffect, useState } from "react"
 
 /**
+ * @note This function ***IS NOT*** server-side rendering friendly, and should be
+ * avoided in the vast majority of  use-cases. For writing responsive UI, please
+ * see our docs on:
+ *
+ * - Responsive Props: https://palette.artsy.net/guides/responsive/ and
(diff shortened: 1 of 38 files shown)


dzucconi (approved):
👍 👍

src/v2/Utils/Hooks/useMatchMedia.ts
```diff
@@ -1,24 +1,32 @@
...
+ * see our docs on:
+ *
+ * - Responsive Props: https://palette.artsy.net/guides/responsive/ and
+ * - Server-side Rendering Responsively: https://artsy.github.io/blog/2019/05/24/server-rendering-responsively/
+ *
+ *
  * Checks to see if the browser matches a particular media query
  *
- * Thanks! https://github.com/olistic/react-use-media/
+ * Thanks! https://github.com/olistic/reat-use-media/
```
dzucconi, line 14: typo

PR #8482, opened 2021-09-24
refactor: Rename Jest .v1 config to .legacy

damassi:
This renames our old jest config file to .legacy, to indicate its deprecated status.


icirellik (approved):
This is awesome! 💖

PR #8483, opened 2021-09-24
fix(inquiry): adds preliminary inquiry analytics tracking

dzucconi:
I'll add specs on Monday. Waiting on https://github.com/artsy/cohesion/pull/238 as well as some updates to the auth event types.


dzucconi:
This is ready to go and the only thing holding back launching the new inquiry flow is https://github.com/artsy/cohesion/pull/238 — @anipetrov @abhitip if there's anything I can do to answer https://github.com/artsy/cohesion/pull/238#discussion_r717046674 let me know

damassi (approved):

PR #8484, opened 2021-09-24
test(categories): convert spec to RTL

anandaroop:
Lil Friday exercise to work those React Testing Library muscles, now that it's been [added to Force](https://artsy.slack.com/archives/C2TQ4PT8R/p1632434644096600).

- converts the basic rendering spec for the `CategoriesApp` to use RTL instead of Enzyme:
- demonstrates the basic boilerplate required by our current pattern
- demonstrates how following the grain of RTL leads to more a11y-friendly markup

### DIFF
--- a/src/v2/Apps/Categories/__tests__/CategoriesApp.jest.tsx
+++ b/src/v2/Apps/Categories/__tests__/CategoriesApp.jest.tsx
@@ -30,4 +30,31 @@ describe("CategoriesApp", () => {
     renderWithRelay()
     expect(screen.getByText("The Art Genome Project")).toBeInTheDocument()
   })
+
+  it("displays families and genes", () => {
+    renderWithRelay({
+      GeneFamilyConnection: () => ({
+        edges: [
+          {
+            node: {
+              name: "Styles and Movements",
+              genes: [
+                {
+                  displayName: "Early Randomcore",
+                },
+              ],
+            },
+          },
+        ],
+      }),
+    })
+
+    expect(
+      screen.getByRole("heading", { name: "Styles and Movements" })
+    ).toBeInTheDocument()
+
+    expect(
+      screen.getByRole("link", { name: "Early Randomcore" })
+    ).toBeInTheDocument()
+  })
 })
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Categories/__tests__/CategoriesApp.jest.tsx
```diff
@@ -30,7 +27,7 @@ const { getWrapper } = setupTestWrapper<CategoriesApp_Test_Query>({
 
 describe("CategoriesApp", () => {
   it("renders", () => {
-    const wrapper = getWrapper()
-    expect(wrapper.find("CategoriesApp").exists()).toBe(true)
+    renderWithRelay()
+    expect(screen.getByText("The Art Genome Project")).toBeInTheDocument()
```
anandaroop, line 31: The RTL way is to eschew internal implementation details like the name of the React component in favor of user-visible text.

src/v2/Apps/Categories/__tests__/CategoriesApp.jest.tsx
```diff
@@ -30,4 +30,31 @@ describe("CategoriesApp", () => {
...
+              ],
+            },
+          },
+        ],
+      }),
+    })
+
+    expect(
+      screen.getByRole("heading", { name: "Styles and Movements" })
+    ).toBeInTheDocument()
```
anandaroop, line 54: My first move was to try the familiar query

```js
screen.getByText("Styles and Movements")
```

…but when I did that I got this helpful error message:

multi


…Thus realizing that I was matching both the horizontal pill nav as well as the subheading for the "Styles and Movements" family.

I was only concerned with the subheading, which led me to write the query this way…

```js
screen.getByRole("heading", { name: "Styles and Movements" })
```

…thus leaning into the [ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#roles) that are kind of the [bread and butter](https://testing-library.com/docs/queries/about/#priority) of working with RTL. 




**_That_** assertion in turn led me to change the markup in the component from a vanilla `div` to a semantic `h2` [up there](https://github.com/artsy/force/pull/8484/commits/23e4ec89d164fd2425f8e5f1f19b8750f68c6388#diff-cf581cd1b36174c1baf0e997d9501fe5fe6feb104500a86cc5a15252e0af9c72R30) 

A nice little illustration of how RTL thinking can actually lead us to write more a11y-friendly components.

Thank you for coming to my ted talk

mdole (approved):
nice! thanks for doing this @anandaroop; it's a good reminder to try using RTL next time I'm writing tests in Force :)

PR #8491, opened 2021-09-27
Revert "chore(deps): update dep typescript from 4.3.5 to v4.4.3"

dzucconi:
Reverts artsy/force#8281

### DIFF
--- a/package.json
+++ b/package.json
@@ -390,7 +390,7 @@
     "stylus-loader": "3.0.2",
     "supertest": "2.0.1",
     "terser-webpack-plugin": "4.1.0",
-    "typescript": "4.4.3",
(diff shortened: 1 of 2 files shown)


damassi:
@dzucconi - can you update the PR description with details about why we had to revert? Then we'll know later when it inevitably fails 🔮

dzucconi:
https://github.com/artsy/force/pull/8281 was automatically merged despite not passing type-check

damassi:
oh no, so this is still happening?

dzucconi:
in this case it appears it was Renovate that merged not Peril. 🤖 uprising?

damassi:
😓

PR #8495, opened 2021-09-28
fix: Adjust artist app render or redirect checks

olerichter00:
Addresses [CX-1644]

## Description

This PR fixes the redirect logic in the artist app to fix a bug related to the "Auction Results" tab. 

With the old logic the "Auction Results" tab was visible on some artist pages, but clicking on the tab forced a redirect to `/artist/:slug/works-for-sale`.

Some of the artists where the bug appears have auction results and some of them don't (in this case the "Auction Results" [renders](https://github.com/artsy/force/blob/master/src/v2/Apps/Artist/ArtistApp.tsx#L50) but shouldn't). To hide the "Auction Results" tab in the second case this PR changes the check to use `artist?.counts?.auctionResults` instead of `artist?.statuses?.auctionLots`.

## Test Cases

**Artists with auction results and without overview**
- http://artsy.net/artist/kaws-x-pushead/auction-results

**Artists with auction results and overview**

- http://artsy.net/artist/gerhard-richter/auction-results

**Artists without auction results and overview**

- http://artsy.net/artist/guillaume-bonn/auction-results
- https://www.artsy.net/artist/ioye-iorgos-pavlopoulos/works-for-sale
- http://localhost:5000/artist/banksy-x-clown-skateboards/works-for-sale

[CX-1644]: https://artsyproduct.atlassian.net/browse/CX-1644


src/v2/Apps/Artist/Routes/Overview/Utils/renderOrRedirect.tsx
```diff
@@ -20,39 +19,10 @@ export function renderOrRedirect({
     return undefined
   }
 
-  const showMarketInsights = () => {
-    // Is there a gallery representation section? Is there an auction highlights
-    // section? Is there a permanent collections section?
-    if (
-      artist?.artistPartnersConnection?.edges?.length ||
-      artist?.auctionResults?.edges?.length ||
```
olerichter00, line 28: This seems to be always undefined because `auctionResults` doesn't exist on `artist` at this point.

src/v2/Apps/Artist/Routes/Overview/Utils/renderOrRedirect.tsx
```diff
@@ -39,17 +39,22 @@ export function renderOrRedirect({
...
+  const onAuctionResultsPath = pathname.includes(
+    `${artist.slug}/auction-results`
+  )
+  const canShowAuctionResults = artist?.counts?.auctionResults
+
   if (pathname === `/artist/${artist.slug}/`) {
     throw new RedirectException(`/artist/${artist.slug}`, 301)
   }
 
-  if (!canShowOverview && !alreadyAtWorksForSalePath) {
```
olerichter00, line 52: This condition can be true even if the user is at the "Auction Results" path.

src/v2/Apps/Artist/Routes/Overview/Utils/renderOrRedirect.tsx
```diff
@@ -39,17 +39,22 @@ export function renderOrRedirect({
...
+  if (
+    (onOverviewPath && !canShowOverview) ||
+    (onAuctionResultsPath && !canShowAuctionResults)
```
olerichter00, line 56: I'm not 100% sure if this covers all cases in which we want to redirect to `/works-for-sale` or if we have to add more cases.

src/v2/Apps/Artist/ArtistApp.tsx
```diff
@@ -47,7 +47,7 @@ const ArtistApp: React.FC<ArtistAppProps> = ({ artist, children, match }) => {
 
   const showOverviewTab = hasOverviewContent(artist)
   const showArtworksTab = artist?.statuses?.artworks
-  const showAuctionLotsTab = artist?.statuses?.auctionLots
+  const showAuctionResultsTab = artist?.counts?.auctionResults
```
olerichter00, line 50: `artist?.statuses?.auctionResults` doesn't seem to be reliable. We use `artist?.counts?.auctionResults` already for the [search auto-suggest quick navigation links](https://github.com/artsy/force/blob/master/src/v2/Components/Search/SearchBar.tsx#L356) in Force and Eigen.

damassi (approved):
@olerichter00 - if this is ready to be merged, be sure to use the assignee field to `assign` to someone! Otherwise, LGTM 👍

PR #8498, opened 2021-09-28
fix(inquiry): adds support for on-demand auth

dzucconi:
Implements the same as in https://github.com/artsy/force/pull/8355

### DIFF
--- a/src/v2/Components/Inquiry/Views/InquiryLogin.tsx
+++ b/src/v2/Components/Inquiry/Views/InquiryLogin.tsx
@@ -4,6 +4,7 @@ import {
   Clickable,
   Flex,
   Input,
+  Message,
   Separator,
   Spacer,
   Text,
@@ -29,6 +30,7 @@ enum Mode {
   Pending,
   Loading,
   TwoFactor,
+  OnDemand,
   Error,
   Success,
 }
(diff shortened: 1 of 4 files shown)


src/v2/Components/Inquiry/Views/InquiryLogin.tsx
```diff
@@ -146,7 +153,14 @@ export const InquiryLogin: React.FC = () => {
           my={1}
         />
 
-        {mode === Mode.TwoFactor && (
+        {mode === Mode.OnDemand && (
+          <Message mt={2} mb={1}>
+            This login requires additional authorization. Please check your
+            email for a one-time authentication code.
+          </Message>
```
jacobherrington, line 160: **question (non-blocking):** The message component renders a `div`, should we be passing a paragraph tag?

```suggestion
          <Message mt={2} mb={1}>
            <p>This login requires additional authorization. Please check your
            email for a one-time authentication code.</p>
          </Message>
```

jacobherrington (commented):
LGTM! I just had a question. 👍

src/v2/Components/Inquiry/Views/InquiryLogin.tsx
dzucconi: If we wanted to do this we'd handle it in the Palette level rather than rely on consumers of it to use semantic markup. (and emit an `<aside role="alert"><p>...` etc...). And differentiate between text children and markup children to support multiple paragraphs.

(Which is a good idea, and one I'll ticket in DSWG)

PR #8501, opened 2021-09-29
[WIP]: Update feature page design

araujobarret:
[Ticket](https://artsyproduct.atlassian.net/browse/FX-3169) 

This PR updates the feature page with the latest style from [Figma](https://www.figma.com/file/0svJJjlcrCFBkVYzagKP3Z/FX-v3-Surfaces?node-id=4128%3A3944)

**DESKTOP**

Before:

https://user-images.githubusercontent.com/15792853/135229739-f98738b2-b37f-40ec-91de-a8700814ae00.mov

After:

https://user-images.githubusercontent.com/15792853/135231104-d51ccf2d-d7b4-4453-b000-615d81911722.mov

**MOBILE**

Before:

https://user-images.githubusercontent.com/15792853/135231215-3e820b04-9612-4e76-9627-48736eaad6c5.mov

After:

https://user-images.githubusercontent.com/15792853/135231234-d4c97d98-970c-4510-bf3e-37e7768e740d.mov


src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
```diff
@@ -45,46 +42,51 @@ export interface FeatureHeaderDefaultProps {
...
+        <Container
+          display={["block", "flex"]}
+          height={[
+            "auto",
+            !!image ? `calc(95vh - ${DESKTOP_NAV_BAR_HEIGHT}px)` : "50vh",
+          ]}
+        >
+          <Figure height={["50vh", "auto"]} backgroundColor="black10">
+            <picture>
+              {/* @ts-expect-error STRICT_NULL_CHECK */}
```
damassi, line 57: 👋 not sure if you've come across these `ts-expect-error`s before, but here's a [little history](https://github.com/artsy/force/blob/master/docs/best_practices.md#avoid-copying-and-try-to-fix--ts-expect-error-strict_null_check-flags). 

If you ever see them when working on something its super helpful to try to cleanup -- usually its as easy as adding a guard or a `?` between props chaining 🙏

araujobarret: thanks for the heads up, I was wondering about them and had no clue about the history.
Gonna fix clean them 🙏

araujobarret: removed a few ones in this [commit](https://github.com/artsy/force/pull/8501/commits/ab72d5a946ce1d92266a284c76ef4aa9fcbf610d), the other that I didn't touch looked a bit more tricky, probably with some nested NonNullable types, please let me know if I should fix them here in this PR 😄

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
```diff
@@ -18,6 +17,12 @@ import { RouterLink } from "v2/System/Router/RouterLink"
 import { FeatureFeaturedLink_featuredLink } from "v2/__generated__/FeatureFeaturedLink_featuredLink.graphql"
 import { themeGet } from "@styled-system/theme-get"
 
+const Container = styled(Flex)<FlexProps>`
+  &:hover > div {
```
dzucconi, line 21: What's this about? It doesn't look like there's links in here?

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
```diff
@@ -18,6 +17,12 @@ import { RouterLink } from "v2/System/Router/RouterLink"
...
+    color: ${color("blue100")};
```
dzucconi, line 22:
```suggestion
    color: ${themeGet("colors.blue100")};
```

dzucconi: Though, I don't like this, exactly. Shoudn't this be on the `RouterLink`?

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
```diff
@@ -115,9 +78,13 @@ export const FeatureFeaturedLink: React.FC<FeatureFeaturedLinkProps> = ({
 
       <Flex flexDirection={size === "large" ? ["column", "row"] : "column"}>
         {subtitle && (
-          <HTML variant="mediumText" html={subtitle} mt={2} flexBasis="50%" />
+          <Text my={1} variant={["xs", "xs"]} textTransform="uppercase">
```
dzucconi, line 81: No need for the array

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
```diff
@@ -142,32 +109,32 @@ export const FeatureFeaturedLinkFragmentContainer = createFragmentContainer(
       fragment FeatureFeaturedLink_featuredLink on FeaturedLink {
         href
         title
-        subtitle(format: HTML)
+        subtitle(format: PLAIN)
```
dzucconi, line 112: So we're stripping formatting?

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
```diff
@@ -45,46 +42,47 @@ export interface FeatureHeaderDefaultProps {
...
-            {/* @ts-expect-error STRICT_NULL_CHECK */}
-            <source srcSet={image.lg.srcSet} media="(min-width: 1200px)" />
-            {/* @ts-expect-error STRICT_NULL_CHECK */}
-            <Image src={image.sm.src} alt={name} loading="lazy" />
-          </picture>
-        </Figure>
-      )}
+  if (image) {
+    return (
+      <FullBleed>
```
dzucconi, line 47: Wondering if we could use https://github.com/artsy/force/blob/master/src/v2/Components/FullBleedHeader.tsx rather than this be a snowflake. The design is slightly different; just wondering if we want to consolidate for consistency. Would run it by design

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
```diff
@@ -45,46 +42,47 @@ export interface FeatureHeaderDefaultProps {
...
-            variant="largeTitle"
-            as="h1"
-            fontSize="size10"
-            textAlign="center"
-          >
-            {name}
-          </Text>
+          <Meta p={4} flexBasis={image ? "50%" : "100%"}>
+            <Text
+              variant="largeTitle"
```
dzucconi, line 66: This is no longer a valid variant under v3. Should be `xl` or `xxl`

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
```diff
@@ -45,46 +42,47 @@ export interface FeatureHeaderDefaultProps {
...
+              as="h1"
+              fontSize="size10"
+              textAlign="center"
+            >
+              {name}
+            </Text>
+            {subheadline && (
+              <HTML variant="subtitle" html={subheadline} textAlign="center" />
```
dzucconi, line 74: No longer a valid variant

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
```diff
@@ -29,45 +37,43 @@ export const FeatureHeaderFull: React.FC<FeatureHeaderFullProps> = ({
...
-      )}
-
-      <Box p={4}>
-        <Join separator={<Spacer my={1} />}>
-          <Text
-            variant="largeTitle"
-            as="h1"
-            fontSize="size10"
-            textAlign="center"
+    <FullBleed>
```
dzucconi, line 40: Ditto re: https://github.com/artsy/force/blob/master/src/v2/Components/FullBleedHeader.tsx

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
```diff
@@ -29,45 +37,43 @@ export const FeatureHeaderFull: React.FC<FeatureHeaderFullProps> = ({
...
+              <source srcSet={image?.lg?.srcSet} media="(min-width: 1200px)" />
+              <Image src={image?.sm?.src} alt={name} loading="lazy" />
+            </picture>
+          </Figure>
+        )}
+
+        <Box p={4}>
+          <Join separator={<Spacer my={1} />}>
+            <Text
+              variant="largeTitle"
```
dzucconi, line 62: Invalid variant

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
```diff
@@ -29,45 +37,43 @@ export const FeatureHeaderFull: React.FC<FeatureHeaderFullProps> = ({
...
+            >
+              {name}
+            </Text>
 
-          {subheadline && (
-            <HTML variant="subtitle" textAlign="center" html={subheadline} />
-          )}
-        </Join>
+            {subheadline && (
+              <HTML variant="subtitle" textAlign="center" html={subheadline} />
```
dzucconi, line 71: Invalid variant

dzucconi:
Would also Q/A with a few other features. They sometimes vary _dramatically_.

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
araujobarret: The subtitle should have no formatting AFAIK from the new design perspective, should be that small text before the title.

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
araujobarret: The design is indeed different, would love to reuse it but to make it I would require to change the `FullBleedHeader` implementation and the idea of it is quite different imo. If you think it's worthy abstracting more the `FullBleedHeader` to reuse here, I can try, but I guess the ideas of both situations are similar but different, we would be creating much more complexity on `FullBleedHeader` just to use here.

Example of feature page with FullBleed layout
image

src/v2/Apps/Feature/Components/FeatureFeaturedLink.tsx
araujobarret: Yeah makes sense there, I wanted to avoid bc of the use the sibling selector(personal preference), I've changed in this [commit](https://github.com/artsy/force/pull/8501/commits/559e998d999e873f6b55855d42dd608df12b6a04)

araujobarret:
> Would also Q/A with a few other features. They sometimes vary _dramatically_.

Thanks a lot for the review @dzucconi , sorry about missing v3 variants and using the v2, I'm aware now 🙂 

Regarding the QA, Jonathan did one big yesterday with other features before I opened it to QA, this blue text on hover for example he asked to implement it. In any case after the CR we'll pass again through QA because we don't have designs for all pages, and one specific `FeatureSet` `type` we are not using, `Artist`, so we can't implement one of the Design's idea 😞

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
dzucconi: Rather than change `FullBleedHeader` my suggestion was that we should change the design to be more consistent with the patterns throughout the site. It'd be a question for design. Some of the header images would have to be updated, sure.

dzucconi:
> this blue text on hover for example he asked to implement it

We're just going to do that globally for the `Text` component, unless there winds up being any weird side-effects to such a sweeping change.

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
araujobarret: Ah okay, now I got it. I will raise this flag to Jonathan, this image is one example of something that we would need to change, this feature sets freedom is a bit funky from what I see

araujobarret:
> We're just going to do that globally for the `Text` component, unless there winds up being any weird side-effects to such a sweeping change.

Humm so will default on Palette, how de we do in this case when a sibling has hovered and the Text color should change?

dzucconi:
> Humm so will default on Palette, how de we do in this case when a sibling has hovered and the Text color should change?

Ah it's probably the case that what I'm saying isn't really relevant for your use case. It's hard to tell from the code/screenshots. So just do whatever you think is best. So long as the only thing that's triggering a hover is a valid anchor tag hit area, you're good.

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
damassi: Thanks for taking care of those 👍 

As a general rule, if they're easy to fix, then we should fix (think, adding a `?` to make object access safe) but if the error is not clear and you end up wading around in the weeds then def don't worry about it.

src/v2/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx
damassi: Here's a list of all of the Typography options and how they map between v2 and v3: https://palette.artsy.net/atoms/typography/

damassi: Here's the mapping between, with a migration guide: https://palette.artsy.net/guides/migration/

damassi:
@araujobarret - when this is ready for final review, be sure to use the `assignee` field to assign to someone; that person has ultimate authority over approving / merging the PR. Check out [these docs](https://github.com/artsy/README/blob/main/playbooks/engineer-workflow.md#assignees-and-reviewers) for a bit more info.

PR #8506, opened 2021-09-29
Revert "Refactor: convert AddressForm to functional component to and add v3 styles"

dzucconi:
I'm not 100% this fixes things. Let me open this and check it out locally.

Auction registration is totally broken because of some circular/infinite update when you go to enter any values.


dzucconi:
OK, yeah; this fixes auction registration. When we revert this for the 5th time let's make sure we Q/A all instances of the form and get Integrity fully green before merging.

PR #8515, opened 2021-09-29
fix(inquiry): reload on close after authentication

dzucconi:
Re: https://github.com/artsy/integrity/pull/251

@anipetrov noticed while Q/A-ing the inquiry analytics that if she drops off the flow before reaching the end then we don't actually trigger the reload; which has caused some problems in the past.

This was a little tricky to do because the initial user fetch has to happen _inside_ of the modal; but the modal also needs to be inside of the context provider. Previously the modal was on the outside. Relay hooks would have made this _much_ easier but here we go. I had to separate out the initial context fetch into its own component where all it does is fetch, then set the context in the hook. This lets us pull it out so the provider can be moved up, and as a result so can the `onClose` function.

I _also_ happened to notice a small bug where we weren't properly populating basic info after login — so this fixes that as well.

Will leave some comments throughout.

### DIFF
--- a/src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
+++ b/src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
@@ -1,4 +1,11 @@
-import React, { createContext, useContext, useRef, useState } from "react"
+import React, {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from "react"
 import { Engine } from "../Engine"
 import { useEngine } from "../config"
 import { createFragmentContainer, graphql, Environment } from "react-relay"
@@ -14,21 +21,23 @@ import { Spinner } from "@artsy/palette"
 export type Context = {
   askSpecialist: boolean
   collectorLevel?: number | null
+  isLoggedIn: boolean
   location?: Location | null
   phone?: string | null
   profession?: string | null
+  requiresReload: boolean
   shareFollows: boolean
-  isLoggedIn: boolean
 }
 
 export const DEFAULT_CONTEXT: Context = {
   askSpecialist: false,
   collectorLevel: null,
+  isLoggedIn: false,
   location: null,
   phone: null,
   profession: null,
+  requiresReload: false,
   shareFollows: false,
-  isLoggedIn: false,
 }
 
 export const DEFAULT_MESSAGE =
@@ -82,15 +91,13 @@ const InquiryContext = createContext<{
 interface InquiryProviderProps {
   artworkID: string
   askSpecialist?: boolean
-  me: useInquiryContext_me | null
   onClose(): void
 }
 
-const InquiryProvider: React.FC<InquiryProviderProps> = ({
+export const InquiryProvider: React.FC<InquiryProviderProps> = ({
   artworkID,
   askSpecialist,
   children,
-  me,
   onClose,
 }) => {
   /**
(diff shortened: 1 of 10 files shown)


src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -14,21 +21,23 @@ import { Spinner } from "@artsy/palette"
 export type Context = {
   askSpecialist: boolean
   collectorLevel?: number | null
+  isLoggedIn: boolean
   location?: Location | null
   phone?: string | null
   profession?: string | null
+  requiresReload: boolean
```
dzucconi, line 28: This is the new addition here; we flip this flag after both login and sign up. And then the close function looks at it and decides to reload the page.

src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -99,28 +106,32 @@ const InquiryProvider: React.FC<InquiryProviderProps> = ({
...
+  }, [])
 
   const [inquiry, setInquiry] = useState<InquiryState>({
     message: DEFAULT_MESSAGE,
   })
 
+  const handleClose = () => {
+    if (context.current.requiresReload) {
+      window.location.reload()
+      return
```
dzucconi, line 126: Said reload

src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -160,8 +171,32 @@ const InquiryProvider: React.FC<InquiryProviderProps> = ({
...
+      isLoggedIn: !!me,
+      location: !!me?.location?.city ? { city: me.location.city } : null,
+      phone: me?.phone,
+      profession: me?.profession,
+      shareFollows: !!me?.shareFollows,
+    })
+  }, [me, setContext])
+
+  return <>{children}</>
+}
```
dzucconi, line 196: Relay hooks would make this much simpler 😓

src/v2/Components/Inquiry/config.ts
```diff
@@ -192,13 +192,6 @@ export const useEngine = ({ context, onDone }: UseEngine) => {
   }, [current, visited])
 
   const next = () => {
-    // At the end, and previously logged out; refresh the page
-    // to log in the entire stack
-    if (engine.current.isEnd() && !context.current?.isLoggedIn) {
-      window.location.reload()
-      return
-    }
```
dzucconi, line 200: Previously this would only reload at the end.

src/v2/Components/Inquiry/Hooks/useInquiryContext.tsx
```diff
@@ -99,28 +106,32 @@ const InquiryProvider: React.FC<InquiryProviderProps> = ({
...
-  }
+  }, [])
 
   const [inquiry, setInquiry] = useState<InquiryState>({
     message: DEFAULT_MESSAGE,
   })
 
+  const handleClose = () => {
+    if (context.current.requiresReload) {
+      window.location.reload()
```
damassi, line 125: Def bringing back code ptsd from when we launched new app shell - can we add a comment here about why this is needed?

dzucconi: Yeah, good idea.

damassi (approved):
😅

PR #8516, opened 2021-09-29
feat: Track user interactions in new checkout flow - order review page [PX-4617]

pepopowitz:
Partially addresses [PX-4617]

Collaboration with @annacarey, @ansor4, and @jpotts244 

@brendamanganelli the only event **from the order review page events** not included in this PR is `clickedOnSubmitOrder` -- we skipped this one because [we're already tracking something for that event](https://github.com/artsy/force/blob/5afd5f6f6e3de5d87dfa70d34058e23f3a41dd14/src/v2/Apps/Order/Routes/Review/index.tsx#L54-L66). Let us know if we need to make changes to the existing tracked event, or if it's something you can use on your end without changes.

## follow-up work
- There is a TODO to move an OwnerType constant to cohesion, but we're hoping to launch shipping tomorrow and we plan to do that after launch.
- Tests!

[PX-4617]: https://artsyproduct.atlassian.net/browse/PX-4617

### DIFF
--- a/src/v2/Apps/Order/Routes/Review/index.tsx
+++ b/src/v2/Apps/Order/Routes/Review/index.tsx
@@ -55,6 +55,9 @@ export interface ReviewProps {
 
 const logger = createLogger("Order/Routes/Review/index.tsx")
 
+// TODO: move this to cohesion
+const OrdersReviewOwnerType = "orders-review"
+
 @track()
 export class ReviewRoute extends Component<ReviewProps> {
   @track<ReviewProps>(props => ({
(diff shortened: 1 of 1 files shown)


jpotts244 (approved):
Not super knowledgeable in this area but here if you nee help shipping/testing

annacarey (approved):
Nice work on this!

brendamanganelli:
Great job - and thank you! We would need to move all existing events to Cohesion at some point, but it's not a launch blocker - so let's skip it for now.

chr-tatu:
> Great job - and thank you! We would need to move all existing events to Cohesion at some point, but it's not a launch blocker - so let's skip it for now.

@brendamanganelli : If there's no ticket for moving them to Cohesion, could you create it, so we don't forget about it?

ansor4 (approved):

PR #8520, opened 2021-09-30
feat: implement tracking in force for shipping page [PX-4619]

pepopowitz:
Addresses [PX-4619]

Collab with @ansor4 and @annacarey 

## Progress

- [ ] ClickedDeliveryMethod -- skipped because we're [already tracking that as an event](https://github.com/artsy/force/blob/5ed9a05a1b2d293df7b2c0a428336eb8c8fc4680/src/v2/Apps/Order/Routes/Shipping/index.tsx#L462-L468)
- [x] ClickedShippingAddress
- [x] ClickedAddNewShippingAddress
- [x] ClickedSelectShippingOption -- we are passing the ShippingQuote ID instead of the text for the subject -- @brendamanganelli this means more work for you when looking things up, but it would have been a risky refactor the night before a launch to send the text instead. 
- [x] ClickedBuyerProtection -- we're sending this event only on the shipping and review pages, because those were the only ones with a context module defined in cohesion.

[PX-4619]: https://artsyproduct.atlassian.net/browse/PX-4619


src/v2/Apps/Order/Components/__tests__/SavedAddresses.jest.tsx
```diff
@@ -21,209 +23,219 @@ class SavedAddressesTestPage extends RootTestPage {
...
-        me {
-          ...SavedAddresses_me
-        }
-      }
-    `,
+describe("Saved Addresses", () => {
+  beforeAll(() => {
+    ;(useTracking as jest.Mock).mockImplementation(() => ({
+      trackEvent: jest.fn(),
+    }))
```
pepopowitz, line 30: most of this file diff is noise from adding a top-level describe, so that all tests can take advantage of this `beforeEach`.

PR #8524, opened 2021-09-30
fix: [Home] Fix columns on CurrentFairs

damassi:
Fixes an issue where the columns weren't flowing correctly in Home -> CurrentFairs. 

Also adjusts some spacing per pairing Jonathan. 

Screen Shot 2021-09-30 at 1 19 25 PM

### DIFF
--- a/src/v2/Apps/Home/Components/HomeCurrentFairsRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeCurrentFairsRail.tsx
@@ -4,7 +4,6 @@ import {
   Text,
   Flex,
   Spacer,
-  Shelf,
   Skeleton,
   SkeletonText,
   SkeletonBox,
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Home/Components/HomeCurrentFairsRail.tsx
```diff
@@ -49,8 +48,9 @@ const HomeCurrentFairsRail: React.FC<HomeCurrentFairsRailProps> = ({
           return (
             <RouterLink
               to={fair.href}
-              textDecoration="none"
               key={index}
+              textDecoration="none"
+              style={{ display: "block " }}
```
damassi, line 53: (no display mixin)

dzucconi: Weird it should be there? https://github.com/artsy/force/blob/53843fa49965a1fc7570ad9bb4806153f1740800/src/v2/System/Router/RouterLink.tsx#L3

dzucconi (approved):

src/v2/Apps/Home/Components/HomeCurrentFairsRail.tsx
damassi: Looks like VSCode was buggin, will fix 👍

PR #8528, opened 2021-10-01
fix(PX-4608): Checkout: Artsy Shipping prices should be right-justified

Serge0n:
Jira Ticket:ticket: : [PX-4608](https://artsyproduct.atlassian.net/browse/PX-4608)

This PR changes the shipping prices alignment.

Now the prices of shipping options in checkout are `right-justified`, so they're `easier to scan by buyers`.

### DIFF
--- a/src/v2/Apps/Order/Components/__tests__/ShippingQuotes.jest.tsx
+++ b/src/v2/Apps/Order/Components/__tests__/ShippingQuotes.jest.tsx
@@ -112,7 +112,7 @@ describe("ShippingQuotes", () => {
     const ascendingPrices = ["$1.00", "$2.00", "$3.00", "$4.00", "$5.00"]
 
     shippingQuotes.forEach((node: ReactWrapper, index: number) => {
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Components/__tests__/ShippingQuotes.jest.tsx
```diff
@@ -112,7 +112,7 @@ describe("ShippingQuotes", () => {
     const ascendingPrices = ["$1.00", "$2.00", "$3.00", "$4.00", "$5.00"]
 
     shippingQuotes.forEach((node: ReactWrapper, index: number) => {
-      expect(node.find("Text").first().text()).toContain(ascendingPrices[index])
```
pepopowitz, line 115: non-blocking because this is a test, but can we update this to target the element in a more predictable way? `.at(1)` is prone to breaking on UI changes; using a `data-test-id` or searching for the text regardless of position would be more robust.

pepopowitz (approved):
This looks beautiful!!!

PR #8531, opened 2021-10-01
test(hooks): Add RTL hooks

araujobarret:
- Adds React Testing Library Hooks dev dependency
- Provides 2 test examples using the library

Related to #8484

### DIFF
--- a/src/v2/Utils/Hooks/__tests__/useDialog.jest.tsx
+++ b/src/v2/Utils/Hooks/__tests__/useDialog.jest.tsx
@@ -1,0 +1,50 @@
+import React from "react"
+import { renderHook, act } from "@testing-library/react-hooks"
+import { render, screen } from "@testing-library/react"
+import { useDialog } from "../useDialog"
+
+const onShow = jest.fn()
+const onHide = jest.fn()
+const hookProps = { Dialog: () => <div>Dialog</div>, onShow, onHide }
+
+describe("useDialog", () => {
+  afterEach(() => jest.clearAllMocks())
+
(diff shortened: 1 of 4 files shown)


src/v2/Utils/Hooks/__tests__/useDialog.jest.tsx
```diff
@@ -0,0 +1,50 @@
...
+    expect(screen.queryByText("Dialog")).not.toBeInTheDocument()
+  })
+
+  it("visibility is set to true", () => {
+    const { result, waitForNextUpdate } = renderHook(() => useDialog(hookProps))
+
+    act(() => {
+      result.current.showDialog()
+    })
+    waitForNextUpdate()
```
anandaroop, line 27: This function [returns a promise](https://react-hooks-testing-library.com/reference/api#waitfornextupdate), right? 

Do we not need to `await waitForNextUpdate()`? (Noticed this in the [official example](https://react-hooks-testing-library.com/usage/advanced-hooks#async))

anandaroop (commented):
NIce, this looks like a very useful addition 🙏🏽

src/v2/Utils/Hooks/__tests__/useDialog.jest.tsx
araujobarret: Good point, I needed because ran into some flakiness, but from my previous experience with the module never needed it, very rare occasions. Removed 👍

damassi (approved):
This is awesome 👍 thanks for adding the additional examples. (Will note this PR for later when we get to some docs.)

PR #8532, opened 2021-10-01
feat(shows2): adds pagination

dzucconi:
Also takes care of [DSWGW-92](https://artsyproduct.atlassian.net/browse/DSWGW-92) cc @tr-ann

I need to follow this up with another PR where we support `?page=n` URL params, but need to make another change in Metaphysics to support.

### DIFF
--- a/src/v2/Apps/Shows/Routes/ShowsCity.tsx
+++ b/src/v2/Apps/Shows/Routes/ShowsCity.tsx
@@ -7,8 +7,8 @@ import {
   Sup,
   Message,
 } from "@artsy/palette"
-import React, { useMemo } from "react"
-import { createFragmentContainer, graphql } from "react-relay"
+import React, { useMemo, useRef, useState } from "react"
+import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
 import { ShowsHeaderFragmentContainer } from "../Components/ShowsHeader"
 import { ShowsCity_viewer } from "v2/__generated__/ShowsCity_viewer.graphql"
 import { ShowsCity_city } from "v2/__generated__/ShowsCity_city.graphql"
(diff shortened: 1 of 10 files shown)


src/v2/Apps/Shows/Routes/ShowsCity.tsx
```diff
@@ -48,6 +57,44 @@ export const ShowsCity: React.FC<ShowsCityProps> = ({ viewer, city }) => {
     )
   }, [city.upcomingShows])
 
+  const [loading, setLoading] = useState(false)
+
+  const currentShowsRef = useRef<HTMLDivElement | null>(null)
+  const { scrollTo } = useScrollTo({
+    selectorOrRef: currentShowsRef,
```
dzucconi, line 64: Should maybe rename to `element`? 🤷

damassi: I like selector or ref 👍 

Thoughts on moving this to Palette?

src/v2/Utils/Hooks/useScrollTo.ts
```diff
@@ -0,0 +1,45 @@
...
+  behavior,
+  offset = 0,
+  selectorOrRef,
+}: UseScrollTo) => {
+  const el =
+    typeof selectorOrRef === "string"
+      ? document.querySelector(selectorOrRef)
+      : selectorOrRef.current
+
+  const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
```
damassi, line 27: Ah, this might prevent things from being moved. All good 👍

damassi (approved):

src/v2/Apps/Shows/Routes/ShowsCity.tsx
dzucconi: It's specific to Force, since it's specific to our header and our sticky implementation. Maybe everything can go over there once we remove hard-coding, but I don't necessarily think it'd be that useful as part of a UI package. Maybe a separate `palette-hooks` package or something down the line.

PR #8535, opened 2021-10-01
feat(shows2): supports page param

dzucconi:
This PR adds support to current shows pagination such that the page param updates on click and full page reloads point to the correct page.

### DIFF
--- a/src/v2/Apps/Shows/Routes/ShowsCity.tsx
+++ b/src/v2/Apps/Shows/Routes/ShowsCity.tsx
@@ -65,13 +65,13 @@ export const ShowsCity: React.FC<ShowsCityProps> = ({
     offset: 20,
   })
 
-  const handleClick = (cursor: string) => {
+  const handleClick = (cursor: string, page: number) => {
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Shows/Routes/ShowsCity.tsx
```diff
@@ -65,25 +65,27 @@ export const ShowsCity: React.FC<ShowsCityProps> = ({
...
+      { slug: city.slug, first: 18, after: cursor, page: null },
       null,
       error => {
         if (error) {
           console.error(error)
         }
 
         setLoading(false)
+
+        window.history.replaceState({}, "", `?page=${page}`)
```
damassi, line 83: Should we be using `router.replace`?

dzucconi: That triggers a full page render (and disrupts the scroll). This is apparently how we handle things in the artwork filter, for instance.

damassi (approved):

PR #8540, opened 2021-10-04
feat(shows2): supports online shows

dzucconi:
This PR adds support for the "Online Exclusive" shows

### DIFF
--- a/src/v2/System/Router/RouterLink.tsx
+++ b/src/v2/System/Router/RouterLink.tsx
@@ -56,11 +56,28 @@ type RouterLinkMixinProps = BoxProps & {
   textDecoration?: ResponsiveValue<string>
 }
 
+const VALID_ROUTER_LINK_PROPS = [
+  "activeClassName",
+  "activeStyle",
+  "exact",
+  "target",
+  "to",
+]
+
+const routerLinkValidator = (prop: string) => {
+  return VALID_ROUTER_LINK_PROPS.includes(prop)
+}
+
 export const RouterAwareLink: React.FC<LinkPropsSimple & RouterLinkMixinProps> =
   // TODO: Update styled-components types
   // @ts-ignore
   styled(Link).withConfig({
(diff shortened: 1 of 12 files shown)


src/v2/System/Router/RouterLink.tsx
```diff
@@ -56,11 +56,28 @@ type RouterLinkMixinProps = BoxProps & {
   textDecoration?: ResponsiveValue<string>
 }
 
+const VALID_FOUND_LINK_PROPS = [
+  "activeClassName",
```
dzucconi, line 60: I guess I previously broke active highlighting when I enabled the prop filtering here.

src/v2/System/Router/RouterLink.tsx
```diff
@@ -56,11 +56,28 @@ type RouterLinkMixinProps = BoxProps & {
...
 export const RouterAwareLink: React.FC<LinkPropsSimple & RouterLinkMixinProps> =
   // TODO: Update styled-components types
   // @ts-ignore
   styled(Link).withConfig({
-    shouldForwardProp: (prop, defaultValidatorFn) => defaultValidatorFn(prop),
+    shouldForwardProp: (
+      prop: string,
+      defaultValidatorFn: (prop: string) => boolean
+    ) => {
+      return defaultValidatorFn(prop) || foundValidatorFn(prop)
```
damassi, line 79: `FOUND` and `found` feel too close to encoding implementation details. can we just say `routerLinkProps`, et?

damassi (approved):

src/v2/System/Router/RouterLink.tsx
dzucconi: Ah, yeah sure

damassi (approved):
Lets get this out the door but at some future point be sure to add a few tests (if they're not already there).

dzucconi:
Yeah, there's already specs for City/Index routes. This is just adding a new city, so

PR #8543, opened 2021-10-04
feat(shows): supports remaining routes/redirects; retires old shows app

dzucconi:
This PR:

* Puts a few redirects in place (`/city/:slug` => `/shows/:slug`), as well as adding a few hardcoded redirects which we were previously supporting
* Adds a /shows/all-cities route which is _very_ simple: just lists out all the cities with links. We previously only supported this in the `/mobile/` app. I figured I'd include it so that all the cities are crawlable, since I don't think the dropdown would be.
* Retires the old desktop and mobile shows apps.


src/v2/Apps/Shows/redirects.ts
```diff
@@ -0,0 +1,89 @@
...
+  tbilisi: "tbilisi-georgia",
+  "tel-aviv": "tel-aviv-yafo-israel",
+  tokyo: "tokyo-japan",
+  toronto: "toronto-canada",
+  vancouver: "vancouver-canada",
+  venice: "venice-italy",
+  vienna: "vienna-austria",
+  washington: "washington-dc-usa",
+  zurich: "zurich-switzerland",
+}
```
damassi, line 89: Out of curiosity, where did this list come from?

damassi (approved):
🚢

PR #8551, opened 2021-10-05
chore: removes legacy asset dll

dzucconi:
This PR migrates the last bit of code `recordArtworkView` and removes the legacy asset DLL from both the Artwork and Order apps.

I believe there's a large network of CoffeeScript I can now delete; but I'm just beginning to unravel it. Will follow up with a separate PR for that once this is merged.


src/v2/Apps/Artwork/useRecordArtworkView.ts
```diff
@@ -0,0 +1,50 @@
...
+  }, [isLoggedIn, artworkID, recordArtworkView])
+}
+
+/**
+ * Component wrapper to enable hook usage in a class component
+ */
+export const UseRecordArtworkView: React.FC = () => {
+  useRecordArtworkView()
+  return null
+}
```
damassi, line 50: Hey, haven't seen this before 👍 👍

dzucconi:
This should be safe to deploy now. Merging.

PR #8552, opened 2021-10-05
feat(Home): Move all rails to <Rail>

damassi:
This moves the remainder of our homepage rails to `<Rail>`. 

Next up, going to look around force and see where our `<Shelf>` usage matches up (title, view all button, possible count, rail items0 and then migrate those as well.


src/v2/Apps/Home/__tests__/HomeArtworkModuleContext.jest.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 import { graphql } from "relay-runtime"
 import { setupTestWrapper } from "v2/DevTools/setupTestWrapper"
-import { HomeArtworkModuleContextFragmentContainer } from "../Components/HomeArtworkModuleContext"
+import { HomeArtworkModuleContextFragmentContainer } from "../Components/HomeArtworkModules/HomeArtworkModuleContext"
```
damassi, line 4: Still feel a bit weird about deleting these files. Lots of possible future stuff we could pull from in that homepage graphql type.

PR #8554, opened 2021-10-05
fix: update inquiry on-demand otp user prompt

mc-jones:
Updates the user prompt in #8498 to match the updated copy from the design team (see: [slack thread](https://artsy.slack.com/archives/C02BGC5PW/p1630601789004800) and implementation #8430).


dzucconi (approved):
Thanks!

PR #8555, opened 2021-10-06
WIP: Migrate all rails to <Rail>

damassi:
WIP


src/v2/Apps/Auctions/Components/AuctionArtworksRail.tsx
```diff
@@ -0,0 +1,208 @@
...
+        })
+      }}
+    />
+  )
+}
+
+const PLACEHOLDER = (
+  <Skeleton>
+    <Rail
+      isLoading
```
damassi, line 97: Swaps out `<Text>` with `<SkeletonText>`

damassi:
Ah, looks like I missed a couple rails due to bad fuzzy matching. Will follow up in a second PR as this is already getting kinda big.

PR #8567, opened 2021-10-06
feat: Handle malformed auth tokens for unsubscribe

jonallured:
We have some number of emails in circulation that have a malformed unsubscribe URL:

```
expected:
https://www.artsy.net/unsubscribe?authentication_token=shhhh

malformed:
https://www.artsy.net/unsubscribe?authentication_token=shhhh?utm=braze
```

So all this PR does is attempt to detect and fix a malformed token.

https://artsy.slack.com/archives/C9RK0BLEP/p1633557767012900


dzucconi:
Ugh; what's minting URLs like that?

PR #8588, opened 2021-10-08
refactor(SystemQueryRenderer): Remove aliasing on imports and more lazyLoading 

damassi:
⚠️ Branched from https://github.com/artsy/force/pull/8587 so that needs to be merged first 

This goes around and removes all of our `{ SystemQueryRender as QueryRenderer }` aliasing we had in the app and switches on `lazyLoading` here and there where appropriate.


dzucconi:
Ah, OK: looks like you're only turning it on when there's actually a placeholder. 👍

damassi:
Yup, sticking to mainly recent / modern code

PR #8608, opened 2021-10-11
chore: track analytics for all buyer guarantee clicks (px-4631)

pepopowitz:
Addresses [PX-4631].

This is more follow-up work to the tracking changes we made right before launching ARTA. We were tracking clicks on the BuyerGuarantee component when it was rendered from the Order app's Shipping route, but _only_ that route. This PR updates the Order app to track those clicks from every route.

[PX-4631]: https://artsyproduct.atlassian.net/browse/PX-4631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brendamanganelli:
👍

PR #8610, opened 2021-10-12
feat(FX-3442): Create alert button for Saved Search

anastasiapyzhik:
[FX-3442]

This adds a _Create Alert_ button under feature flag `ENABLE_SAVED_SEARCH`. The button should be visible on artist page _Works for Sale_ tab when at least 1 filter is active, in future the click on it will open a modal for creating email alert

**Demo**

Screenshot 2021-10-25 at 11 56 57

https://user-images.githubusercontent.com/44819355/138666295-c304e8b2-6f59-438e-b765-17572cbe920f.mp4

[FX-3442]: https://artsyproduct.atlassian.net/browse/FX-3442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -153,6 +157,9 @@ export const BaseArtworkFilter: React.FC<
   const { filtered_artworks } = viewer
   const hasFilter = filtered_artworks && filtered_artworks.id
 
+  const isCreateAlertEnabled = sd.ENABLE_SAVED_SEARCH && enableCreateAlert
+  const showCreateAlert = isCreateAlertEnabled && filterContext.hasFilters
```
dblandin, line 161: **question**: Should we consolidate this logic into the `ArtistArtworkFilter.tsx` component instead of updating this component to be aware of the feature flag?

The addition of the component prop here makes sense but I think deciding whether we enable/disable that option should be a bit further up the stack, closer to the artist page's artwork grid, especially since that will be our initial focus as we prep to expand to other surfaces.

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -317,7 +324,7 @@ export const BaseArtworkFilter: React.FC<
       {/* Desktop Artwork Filter */}
       <Media greaterThan="xs">
         {tokens.version === "v3" && (
-          <Flex justifyContent="space-between" alignItems="flex-start" pb={4}>
+          <Flex justifyContent="space-between" alignItems="flex-start" pb={40}>
```
dblandin, line 327: **question**: What prompted the change from 4 to 40 here? Should we try to keep to our defined design system spacing options?

dblandin (commented):
Looking good!!

src/v2/Components/ArtworkFilter/index.tsx
anastasiapyzhik: That makes sense 👍

src/v2/Components/ArtworkFilter/index.tsx
anastasiapyzhik: That wasn't intentional, I was just playing around with layout styles and accidentally changed it😅

src/v2/Components/ArtworkFilter/CreateAlertButton.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react"
+import styled from "styled-components"
+import { themeGet } from "@styled-system/theme-get"
+import { BellIcon, Button, Flex, Spacer, Text } from "@artsy/palette"
+
+const StyledCreateAlertButton = styled(Button).attrs({
+  variant: "secondaryOutline",
+  size: "small",
+  width: 130,
```
dimatretyak, line 9: **nitpick**: do we have to specify a fixed width here?

dimatretyak, line 9: do we necessarily need to specify a fixed width here?

anastasiapyzhik: I believe the answer is yes since we want it to be of the same width as in figma spec

Screenshot 2021-11-03 at 13 30 55

dimatretyak: I am afraid that when using a fixed width, there may be problems with displaying the button label (for example, the user may have the page zoomed in the browser and the fixed width will not be enough for label). For example, as in the screenshot below, the button label goes beyond the offsets

Image 2021-11-03 14-01-08

dimatretyak (approved):

dblandin (approved):
LGTM! 👍🏼

PR #8621, opened 2021-10-13
chore: updates to relay bump PR

pepopowitz:
I noticed the target versions of relay dependencies varied in #8337, with some bumping to v10.0.1 and some bumping to v10.1.3. 

This PR is based on that branch and bumps the ones that were at 10.0.1 to 10.1.3. It **does not** fix the failing tests in that branch; this was just a thread I pulled at to see if it would fix them. 

It includes removing a patch!!! I felt comfortable removing it because it contains a comment that reads `// This patch is just to make an error message more helpful. It can be deleted safely.`. In addition, it looks like that area of code has changed between v9 and v10, and the patch would need to be rewritten to keep it.

I'm happy to take time to rewrite the patch if we think it's important; my preference is to remove it and see if the error messages are frustrating and need to be patched again.


damassi:
Totally agree on removing the patch. Lets trust in the library authors (who work on incredible scale) knowing what's best.

PR #8634, opened 2021-10-14
chore: Move detect-secrets call to formal lit-staged list

damassi:
This formalizes our use of `detect-secrets` + git hooks by placing it within the `lint-staged` chain of command:
 
Screen Shot 2021-10-14 at 11 57 24 AM

This also fixes the cannot use `--no-verify` issue.


ovasdi:
> @ovasdi, not sure why CI is failing as the detect-secrets command ran successfully and nothing was changed in that file?

👋, I replied in slack with some details as to why this is happening.
Also re-ran the workflow manually from failed. Passing now.

PR #8646, opened 2021-10-15
feat: integrates toasts

dzucconi:
Re: https://github.com/artsy/palette/pull/1060

I've got some things to confirm on the design end, and some clean up to do in the PR, but this is what integration and usage should look like.

This will replace the `[ToastComponent](https://github.com/artsy/force/blob/master/src/v2/Components/Toast/ToastComponent.tsx)` cc @tam-kis @alexj105


dzucconi:
After consulting with Jonathan: these move to the top center.

dzucconi:
This should prove quite useful for the settings rebuild

PR #8649, opened 2021-10-16
feat(PX-4502): adjust the display for order details page

oxaudo:
Adjusting this display based in displayState.


oxaudo:
Ok. Resolved some TODOs here in logic.

Still need:
- fix specs
- add new specs
- link tracking id if tracking URL is present
- account for the case where no tracking URL or tracking Id is present. 

cc @annacarey @jpotts244 @pepopowitz

oxaudo:
Ok friends! @annacarey @jpotts244 @pepopowitz - removing WIP here as I added minimal spec and the state is currently much better than what we had before!

There is still more things to do on this specific one:
- link tracking id if tracking URL is present
- account for the case where no tracking URL or tracking Id is present.
- write even more specs!
But I think all those can come as a follow up PRs. The main logic here is ok and I think it would be helpful to have it on staging to look at things.

PR #8682, opened 2021-10-20
feat(React): Auto import React in our components 

damassi:
Addresses https://artsyproduct.atlassian.net/browse/WP-72 

Been wanting to get to [this for a while](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). 

Ended up being a bit tricky due to how we alias things in our webpack config, but figured it out. Issue is we were aliasing `react`, which then prevented us from properly aliasing `react/jsx-runtime` (aliases aren't recursive). And so i just removed the `react` alias since we no longer need to commonly `yarn link` with reaction, etc. 

I then updated all of our files via a codemod: `npx react-codemod update-react-imports`.


damassi:
> Some conflicts here; I also worry about conflicts in open and about to be opened PRs

The conflicts aren't extreme. Remove react import here, destructure it there. I'll put a notice out to #dev for folks who need help.

PR #8684, opened 2021-10-20
docs(detect-secrets): Update CI output to indicate how to fix detect-secrets failures

damassi:
This updates the CI job with some instructions on how to fix if the job fails:

```
Changes to baseline results need to be committed.
If seeing this message in CI you may have ran your commit using --no-verify, thereby skipping the detect-secrets git-hook.

To fix, run:
  git reset --soft HEAD^
  yarn detect-secrets
```

Note that this approach will only reset the latest commit so we might want to make this more robust in the future. I ran into issues executing [the approach suggested here](https://artsy.slack.com/archives/CA8SANW3W/p1634768262207200?thread_ts=1634764642.195600&cid=CA8SANW3W):

```sh
detect-secrets-hook --baseline .secrets.baseline $(git ls-files)
zsh: argument list too long: detect-secrets-hook
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -414,7 +414,7 @@
   "husky": {
     "hooks": {
       "post-merge": "sh scripts/pull-lock.sh",
-      "pre-commit": "lint-staged",
+      "pre-commit": "lint-staged --no-stash",
       "pre-push": "yarn type-check"
     }
   },


damassi:
@ovasdi - feel free to push to this branch if you think of a better way! Not sure whats up with my env in terms of the suggested command, but that concerns me a bit.

PR #8777, opened 2021-10-30
refactor(Auth2): Move files to new Authentication app 

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-576 .

This PR takes our static (static = non-modal) auth-related components and moves them into a new `/v2/Apps/Authentication` app, and then sets up some routes so that all of our pages are in the new framework, piggy backing on the new `onServerSideRender` router hook. 

There's still further testing we need to do before this flips live but I think we're in a good incremental spot to merge. 

cc @artsy/grow-devs


src/v2/Components/Authentication/Types.ts
```diff
@@ -48,31 +48,14 @@ export interface AfterSignUpAction {
...
-   */
-  destination?: string
-  /**
-   * the page path the user is redirected to after successfully
-   * login or account creation (skips onboarding).
-   */
-  redirectTo?: string
-  /**
-   * the action taken that prompted user to signup or login.
+   * Hook to be called after the modal has closed
```
damassi, line 51: just alphabetization

src/v2/Apps/Authentication/authenticationRoutes.tsx
```diff
@@ -9,14 +21,80 @@ const ResetPasswordRoute = loadable(
...
+
+const SignupRoute = loadable(
+  () =>
+    import(
+      /* webpackChunkName: "authenticationBundle" */ "./Routes/SignupRoute"
+    ),
+  { resolveComponent: component => component.SignupRoute }
+)
+
+const runAuthMiddleware = compose(checkForRedirect, setReferer)
```
damassi, line 40: is there a lodash function like `compose` that instead of threading args through functions that return an output, simply invokes said functions with initial args?

src/desktop/apps/authentication/__tests__/loginRoute.jest.ts
```diff
@@ -1,4 +1,6 @@
-import { FullPageAuthStatic } from "../components/FullPageAuthStatic"
+/* eslint-disable jest/no-done-callback */
+
+import { FullPageAuthStatic } from "v2/Apps/Authentication/Components/FullPageAuthStatic"
```
dzucconi, line 3: I'd probably rename this to something more sensible

src/v2/Apps/Authentication/Utils/useAuthFormData.ts
```diff
@@ -0,0 +1,77 @@
...
+  } = match.location.query as ModalOptions
+
+  const title = computeTitle({
+    copy: copyQueryParam,
+    intent,
+    pageTitle,
+  })
+
+  const redirectTo = getENV("AUTHENTICATION_REDIRECT_TO")
+  const signupReferer = getENV("AUTHENTICATION_REFERER")
```
dzucconi, line 39: Wouldn't we use sharify for this?

src/v2/Apps/Authentication/authenticationRoutes.tsx
dzucconi: You're using styled-system's compose here which... I don't think is going to work how you expect it to.

Are you looking for `curry`?

src/v2/Apps/Authentication/authenticationRoutes.tsx
```diff
@@ -9,14 +21,80 @@ const ResetPasswordRoute = loadable(
...
+      return LoginRoute.preload()
+    },
+  },
+  {
+    path: "/sign_up2",
+    render: ({ match }) => {
+      throw new RedirectException(`/signup2${match.location.search}`, 301)
+    },
+  },
 ]
```
dzucconi, line 100: Ah, cool I guess we can turn these all on at once

src/v2/Apps/Authentication/authenticationRoutes.tsx
damassi: Nah, compose is just an exposed helper function that composes two functions via curry. It works as one would expect. I don't know why lodash doesn't have it; it has flow, which is the same, only the other direction. I'm just looking for a way to say `invoke([fnA, fnB], withArgs)` -- something like that.

src/desktop/apps/authentication/__tests__/loginRoute.jest.ts
damassi: I didn't want to mess with any naming because this stuff is used elsewhere. I was gonna do an auth form cleanup after this PR, which basically just moves stuff and rearranges a bit.

src/v2/Apps/Authentication/Utils/useAuthFormData.ts
damassi: `getENV` uses sharify under the hood, and is the SSR friendly way to access.

src/v2/Apps/Authentication/authenticationRoutes.tsx
damassi: Can you elaborate?

src/v2/Apps/Authentication/authenticationRoutes.tsx
dzucconi: It's not though — https://github.com/styled-system/styled-system/blob/6d822f27a486e5e2bc1cae1f44dd151ce874a863/packages/core/src/index.js#L186-L190 - it news up some parsers etc.

But yeah as you said there's flow and also flowRight

src/v2/Apps/Authentication/Routes/__tests__/ForgotPasswordRoute.jest.tsx
```diff
@@ -0,0 +1,73 @@
...
+        title: "Reset your password",
+        type: ModalType.forgot,
+      },
+      options: queryParams,
+      type: ModalType.forgot,
+    }))
+
+    render(<ForgotPasswordRoute />)
+    expect(screen.getAllByText("Reset your password").length).toBe(2)
+    expect(screen.queryByText("Send me reset instructions")).toBeInTheDocument()
```
damassi, line 45: Note that all of the actual form component behavior is tested elsewhere. These route tests just test whats unique to these static routes.

damassi: That being said, more tests can be added elsewhere in a follow-up

src/v2/Apps/Authentication/authenticationRoutes.tsx
dzucconi: Oh — I mean there's a ticket to promote the reset_password2 — but we can just promote all the routes at once.

src/v2/Apps/Authentication/Utils/useAuthFormData.ts
dzucconi: Wait — Sharify isn't SSR friendly? I thought that was the whole point of it?

dzucconi (approved):

src/v2/Apps/Authentication/Utils/useAuthFormData.ts
damassi: As far as I understand, since our JS is now fully isomorphic when you try to import `sd` it will look for the window and fail. So getENV will first check environment -- either localStorage or process.env on server, and then sharify on client. Old force wasn't executing JS isomorphically like we are now, it was mostly just rendering jade templates on the server with data passed in (from express). Now that we're using sharify much more widely (and without consideration of where things are executing) we needed a helper.

PR #8785, opened 2021-11-02
feat(router): Add onServerSideRender and onClientSideRender callbacks to route configuration

damassi:
This addresses some tech debt by adding a new pre-render hook behavior to our framework router. 

The gist of things is: 
- Before, if a `v2/App/<AppName>` server-side Express functionality, we'd have to mount said code within [`middleware.ts`](https://github.com/artsy/force/blob/master/src/middleware.ts) and break co-location
- To fix, this adds a new `onServerSideRender` prop to our route, and if found, will execute when a route matches
- This passes Express's `req`, `res`, and `next` functions, as well as the `route` being executed.

Example:

```ts
const routes = [
  {
    path: '/foo',
    Component: () => <div>hello how are you?</div>
    onServerSideRender: ({ req, res, next }) => {
      if (req.path === '/foo/bar') {
        res.redirect(301, '/bar')
      }
    }
  }
]
```

Now, everything can live comfortably right beside everything else. 

cc @artsy/grow-devs


src/v2/System/Router/Route.tsx
```diff
@@ -74,34 +72,7 @@ function createRender({
...
-          See: https://graphql.slack.com/archives/C0BEXJLKG/p1561741782163900
-
-          export const setLocal = (query: GraphQLTaggedNode, localData: object) => {
-            const request = getRequest(query);
-            const operation = createOperationDescriptor(request, {});
-
-            env.commitPayload(operation, localData);
-            env.retain(operation.root);  // <== here @en_js magic :wink:
-          };
-        */
```
damassi, line 103: No longer relevant

src/middleware.ts
```diff
@@ -149,20 +144,6 @@ export function initializeMiddleware(app) {
...
-  /**
-   * FIXME: Move this into function
-   * Mount middleware for handling server-side portions of apps mounted into
-   * global router.
-   */
-  app.get("/artist/*", artistMiddleware)
-  app.get("/search*", searchMiddleware)
-
-  // TODO: Artwork download, does this belong here.
-  app.get("/artwork/:artworkID/download/:filename", handleArtworkImageDownload)
```
damassi, line 165: Moved all of this functionality into the new pattern. Code didn't really change at all, just relocated. (Thankful we didn't have much more server-side middleware tho 😅)

src/v2/Apps/Artist/artistRoutes.tsx
```diff
@@ -86,7 +87,8 @@ export const artistRoutes: AppRouteConfig[] = [
     path: "/artist/:artistID",
     ignoreScrollBehaviorBetweenChildren: true,
     getComponent: () => ArtistApp,
-    prepare: () => {
+    onServerSideRender: enableArtistPageCTA,
```
damassi, line 90: Add the callback to the route

src/v2/Apps/Artwork/Server/handleArtworkImageDownload.ts
```diff
@@ -14,9 +14,5 @@ export const handleArtworkImageDownload = async (req, res, next) => {
       imageRequest.set("X-ACCESS-TOKEN", req.user.get("accessToken"))
     }
     req.pipe(imageRequest, { end: false }).pipe(res)
-  } else {
```
araujobarret, line 17: which error are we sending back to the user if he's not authorized to download an artwork image?

src/v2/Apps/Search/Server/redirectQueryToTerm.ts
```diff
@@ -1,19 +1,15 @@
 import { stringify } from "qs"
 
-export const searchMiddleware = async (req, res, next) => {
-  if (!req.query.term) {
+export function redirectQueryToTerm({ req, res, next }) {
```
araujobarret, line 3: maybe better to specify the argument type and destructure only the variables that the fn is going to use, we can remove next from here in this case

araujobarret (approved):
Thanks again for the great PR 💯

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -35,24 +35,20 @@ export const AppShell: React.FC<AppShellProps> = props => {
   const showFooter = !isEigen && !routeConfig.hideFooter
   const appContainerMaxWidth = routeConfig.displayFullPage ? "100%" : null
 
-  /**
-   * Check to see if a route has a prepare key; if so call it. Used typically to
-   * preload bundle-split components (import()) while the route is fetching data
-   * in the background.
-   */
+  // Check to see if a route has a prepare key; if so call it. Used typically
```
dzucconi, line 38:
```suggestion
  // Check to see if a route has a `onClientSideRender` key; if so call it. Used typically
```

dzucconi (approved):
Very nice

AlicanAkyuz (approved):
🚀 🚀

src/v2/Apps/Artwork/Server/handleArtworkImageDownload.ts
damassi: This is no longer relevant because we only show the download image button to authorized (admin) users. It's hidden for everyone else. (This code is superrrr old.)

damassi:
Addressed feedback! will #squashOnGreen

PR #8788, opened 2021-11-02
[DO NOT MERGE] Feat: [Swa 49] - Handle errs of image upload in consignment flow

AlicanAkyuz:
### Description 

Ticket: [SWA-49](https://artsyproduct.atlassian.net/browse/SWA-49)

This ticket involves displaying an error modal to the user when uploading an image fails. Note that this ticket deals with image upload errors on a granular basis; the network errors on a more higher level will be handled within the scope of [another ticket](https://artsyproduct.atlassian.net/browse/SWA-79). 

### Video 

https://user-images.githubusercontent.com/42584148/139831889-c478523b-92b9-4b90-89d6-9fd371af07a6.mov


src/v2/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
```diff
@@ -97,6 +112,13 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
 
   return (
     <Box {...rest}>
+      <ErrorModal
+        show={showErrorModal}
+        headerText="An error occurred"
+        contactEmail="orders@artsy.net"
```
AlicanAkyuz, line 118: the email address might change as per product requirements.

iskounen (commented):
These changes look good to me.

**question:** I am surprised by the amount of setup that is needed to test the error modal! Are there any opportunities to simplify that setup code (or share it with other tests related to the photo upload step)?

src/v2/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
iskounen: Thanks for pointing this out. I see that this is being discussed in the [Jira ticket](https://artsyproduct.atlassian.net/browse/SWA-49?focusedCommentId=43152), so let's keep discussing in that thread before merging this ticket.

Would you mind making this a WIP PR until we resolve this?

AlicanAkyuz:
I agree that the setup for tests can be better. 

This test could be written in UploadPhotos.jest as well, but _uploadPhoto_ needs to reject for us to imitate a failure. I couldn't find a way to use the same mock promise with different resolutions. I'll look into this more within the scope of SWA-79 unless someone already knows how to resolve ```jest.mock("../../Utils/fileUtils", () => ({...})``` differently in a single test file.

kiryl-zubarau:
@AlicanAkyuz You can mock the function and import it. Then cast the function to jest.Mock and mock the result value: 

https://github.com/artsy/force/blob/cbfd67714f850dc0924a4f0a89cd35e7f45f70be/src/v2/Apps/Consign/Routes/SubmissionFlow/Utils/__tests__/fileUtils.jest.ts#L58-L60

PR #8792, opened 2021-11-02
feat(SWA-84): AutocompleteInput component from Palette instead Autosuggest

Serge0n:
Jira Ticket:ticket: : [SWA-84](https://artsyproduct.atlassian.net/browse/SWA-84)
Discussion:scroll: : https://github.com/artsy/force/pull/8590#discussion_r726368466

This PR replaces `Autosuggest` usage with `AutocompleteInput`  in the new `Submission flow`.
**VIDEO:clapper: :**

https://user-images.githubusercontent.com/55637696/143873929-4e488901-9ea5-4701-8e36-7ea857f6f04b.mov

https://user-images.githubusercontent.com/55637696/143874357-21dd00d2-55fe-4b05-93f9-7ef170398237.mov

https://user-images.githubusercontent.com/55637696/143885687-4ca23488-385e-4d25-96d1-c44d2e09c642.mov


damassi:
@Serge0n / @kiryl-zubarau - if y'all ran into any API roughness with our implementation do report back! Thanks for taking the time to swap this new component in 👍

Serge0n:
> @Serge0n / @kiryl-zubarau - if y'all ran into any API roughness with our implementation do report back! Thanks for taking the time to swap this new component in 👍

Yeah) I prepared a couple of questions:

1.  About `Tab` navigation
- Is it correct that we can navigate to options by pushing `Tab`? (because we are able to do that by arrows)
- if I will continue pushing `Tab` and move to the next input it didn't close options. And I can navigate to the fields under them

https://user-images.githubusercontent.com/55637696/140022391-b3e54564-3865-460d-92ae-7567f94822c2.mov

2. About `onBlur`
- Is it correct that `onBlur` on `AutocompleteInput` triggered equally `when I'm hovering to any option `and `when I'm clicking outside the component`? I asked because I want to catch the event when a user moved out the focus from the field to show him an error if he didn't choose one of the suggested artists. 

https://user-images.githubusercontent.com/55637696/140021991-0dd67cfc-7779-4a67-9632-499a8e82cb1d.mov

3. About `onSelect`
- If I understood it correct, `onSelect` should be called if I'll hover (or navigate by arrows) to option, and as result input will fill by option text. But it don't work for me now, I didn't figure why yet.

https://user-images.githubusercontent.com/55637696/140024871-5afce0bf-586e-43f5-984c-8843f3dcb2f8.mov

dzucconi:
> Is it correct that we can navigate to options by pushing Tab? (because we are able to do that by arrows)
You're right, I think this should be disabled. We should move focus to the next tab stop outside of the input.

> if I will continue pushing Tab and move to the next input it didn't close options. And I can navigate to the fields under them
Yep, this is a bug as well. We'll fix and close it once you leave the input.

> Is it correct that onBlur on AutocompleteInput triggered equally when I'm hovering to any option (or pushing Tab or arrows) and when I'm clicking outside the component? I asked because I want to catch the event when a user moved out the focus from the field to show him an error if he didn't choose one of the suggested artists.
Hm, this sounds tricky. I think that we should blur the input on keyboard up/down because then you're explicitly moving the focus to/from each option. And right now it's implemented such that mouse enter/leave also moves the focus around.

We could have a synthetic `onBlur` which we trigger when focus leaves the _entire_ component, rather than just the input

> maybe it will be better if onMouseEnter will trigger onSelect like onMouseDown? Because if I'll hover on option and after that, I'll move focus out the input value will be filled from option value, but onSelect not called (and in the case of Submission flow form it didn't set artistId in formik and as a result form will invalid)
I think this is just a bug. If you don't select anything explicitly with a click or an <enter> we shouldn't leave the staged option present in the input. Will fix.

I'll make a ticket that encompasses the above points. Thank you for the great detailed feedback!

dzucconi:
https://artsyproduct.atlassian.net/browse/DSWGW-99

AlicanAkyuz (approved):
lgtm

iskounen (approved):

dzucconi (approved):

PR #8795, opened 2021-11-02
chore(PX-4677): small refactor to account for displayState being capitalized on Exchange side now

oxaudo:
Super minor PR using the refactor that we did in Exchange.

### DIFF
--- a/src/v2/Apps/Order/Routes/Status/index.tsx
+++ b/src/v2/Apps/Order/Routes/Status/index.tsx
@@ -60,7 +60,7 @@ export class StatusRoute extends Component<StatusProps> {
     const isArtaShipped: boolean =
       requestedFulfillment?.__typename === "CommerceShipArta"
 
-    switch (displayState?.toUpperCase()) {
+    switch (displayState) {
       case "SUBMITTED":
         return isOfferFlow
           ? {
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Order/Routes/Status/index.tsx
```diff
@@ -148,10 +148,10 @@ export class StatusRoute extends Component<StatusProps> {
         // accepted
         return this.canceledOfferOrderCopy()
       default:
-        // This should not happen. Check the order states are all accounted for:
+        // This should not happen. Check the order displayState here:
         // https://github.com/artsy/exchange/blob/master/app/models/order.rb
         // (Aside from PENDING and ABANDONED)
-        logger.error(`Unhandled order state: ${state}`)
+        logger.error(`Unhandled order state: ${displayState} in ${state} state`)
```
annacarey, line 154: Should this be:
```suggestion
        logger.error(`Unhandled order state: ${displayState} in ${displayState} state`)
```

Generally a little confused by this comment. Why are we ignoring `Pending` and `Abandoned`?

annacarey (commented):
Looks good other than a quick q on the error state

src/v2/Apps/Order/Routes/Status/index.tsx
oxaudo: @annacarey - no, I intentionally left the state in there as well as this should never be shown to users and if for some reason that is - it would be helpful for us to know both State and DisplayState for easier debugging.

Pending and Abandoned should be filtered way before and never shown to users in this view.

annacarey: Got it thank you! This is good to go for me!

PR #8796, opened 2021-11-02
feature: Dockerfile Tuning

icirellik:
Tuns the way that caches are used in the Dockerfile.

The first change separates the dev and prod caches this fixes a rare
issues caused when the cache is empty and both instances of yarn attempt
to write the same files at the same time. This will increase total disk
space in CI by about 2 GB.

The second change limits a single writer to the cypress directory at a
time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:experimental
+# syntax=docker/dockerfile:1.3
 #              +------------------+
 #              |                  |
 #              |  builder-base    |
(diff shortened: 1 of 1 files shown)


joeyAghion (approved):
> limits a single writer to the cypress directory

What sort of problem motivated this^?

icirellik:
@joeyAghion mostly it's preventative, I'm not sure we will ever run into this issue as cypress updates less regularly than the yarn cache. It should have zero impact.

PR #8797, opened 2021-11-02
feat: Update favicon to black GRO-579

jonallured:
All I did for this was search the project for filename that contained `favicon` like so:

```
jon@juggernaut:~/code/force(jonallured/feat-update-favicon-to-black)% fd favicon
src/desktop/public/images/favicon.ico
src/mobile/public/images/favicon.ico
```

And then I updated those images with the PNG provided by Jonathan. Here's some pretty pictures:

Screen Shot 2021-11-02 at 1 23 19 PM
Screen Shot 2021-11-02 at 1 23 31 PM

https://artsyproduct.atlassian.net/browse/GRO-579

/cc @artsy/grow-devs


dzucconi:
We can just use an SVG for a favicon now instead of this rasterized file.

That'd also help avoid the blurry rasterization and non-square file

dzucconi:
And we could do dark mode very easily with it.

jonallured:
More chatter here: https://artsy.slack.com/archives/C01ADJNCS5D/p1635887184301100 there might be a way to use a fallback for Safari but diminishing returns... 🤷‍♂️

PR #8807, opened 2021-11-03
[Feat]: SWA-51/Trim email address

AlicanAkyuz:
### Description 

Ticket: [SWA-51](https://artsyproduct.atlassian.net/browse/SWA-51)

This PR handles the case where user types spaces in contact information step's email input. The spaces are simply trimmed before pushed to formik's event handler. 

### Videos (updated after trim moved to validation)
https://user-images.githubusercontent.com/42584148/140296986-81d23465-7c45-4b91-b0dd-2c5cabe9c1af.mov

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
@@ -34,7 +34,10 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
         title="email"
         placeholder="Your Email Address"
         value={values.email}
-        onChange={handleChange}
+        onChange={e => {
+          e.target.value = e.target.value.trim()
+          handleChange(e)
+        }}
         onBlur={handleBlur}
       />
       <Input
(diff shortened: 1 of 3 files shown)


AlicanAkyuz:
@Serge0n I left out testing this trimming, since there's a ticket for unit testing assigned to you.

Serge0n:
I thought that this ticket is to change `validation` and use `trim()` before saving data to session storage. But I'm not sure... Let's ask @kiryl-zubarau and @iskounen

kiryl-zubarau:
I would say yes, let's add `trim` to validation, and before saving the form. I suppose here we should not restrict users to enter spaces. 
Let's also add such behavior for the name and phone number fields.

Serge0n:
Also, you forgot to trim values before saving form data to session storage. Like here:
https://github.com/artsy/force/blob/9b22244da128622d760ad28a270d3edc47e81798/src/v2/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/ArtworkDetails.tsx#L40

and to put trimmed values into saveSubmission() like here:
https://github.com/artsy/force/blob/9b22244da128622d760ad28a270d3edc47e81798/src/v2/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/ArtworkDetails.tsx#L48

AlicanAkyuz:
oh thanks for catching this @Serge0n, having a look 👀

Serge0n:
> oh thanks for catching this @Serge0n, having a look 👀

No problem) And we should trim all values in the form (phone and name too)

kiryl-zubarau (approved):
LGTM!

PR #8808, opened 2021-11-03
fix: User Page Import

icirellik:
An error occurred on the user page where the default export was changed
to a named export during decaffeination and the name as imported was
different.


jonallured (approved):
LGTM!!

PR #8814, opened 2021-11-03
feat(FX-3527): List of saved search alerts

dimatretyak:
This PR resolves [FX-3527]


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f299e86bac1fcaa8cabc6d5cae0d28073350615b
</p>

src/v2/Apps/Settings/SettingsApp.tsx
```diff
@@ -39,6 +44,14 @@ const tabs = [
   },
 ]
 
+const tabs = sourceTabs.filter(tab => {
```
anastasiapyzhik, line 47: Can we render tabs something like this?
```
{tabs.map(tab => {
  if (tab.name === "Alerts" && !getENV("ENABLE_SAVED_SEARCH")) {
    return null
  }
  return <RouteTab to={tab.url}>{tab.name}</RouteTab>
})}
```

So that we do not need to make extra iteration

src/v2/Components/UserSettings/UserSettingsTabs.tsx
```diff
@@ -56,6 +61,13 @@ export const UserSettingsTabs: React.FC<UserSettingsTabsProps> = track()(
         variant: "xl" as TextVariant,
       },
     })
+    const routes = sourceRoutes.filter(route => {
```
anastasiapyzhik, line 64: The same here :)

anastasiapyzhik (dismissed):
LGTM!

src/v2/Apps/Settings/SettingsApp.tsx
dimatretyak: I don't think if using extra iteration will affect the performance in any way. And also in the future, when we get rid of feature flag, we will have to make fewer changes in the code (we can just remove `filter`)

anastasiapyzhik: Just a suggestion, non blocking, feel free to keep it as it is

src/typings/sharify.d.ts
```diff
@@ -25,7 +25,7 @@ declare module "sharify" {
       CURRENT_USER: User
       EIGEN: boolean
       readonly ENABLE_FAIR_PAGE_EXHIBITORS_TAB: boolean
-      readonly ENABLE_SAVED_SEARCH: boolean
+      ENABLE_SAVED_SEARCH: boolean
```
dblandin, line 28: **question**: Are we able to keep this variable readonly? Maybe by mocking sharify in the specs?

src/v2/Components/NavBar/Menus/__tests__/NavBarUserMenu.jest.tsx
```diff
@@ -37,6 +43,22 @@ describe("NavBarUserMenu", () => {
...
+    sd.ENABLE_SAVED_SEARCH = true
+
+    const wrapper = getWrapper()
+    const links = wrapper.find("a")
+
+    expect(links.map(a => [a.prop("href"), a.text()])).toEqual([
+      // Label also includes SVG image title
+      ["/user/purchases", "Pending Order History"],
+      ["/user/saves", "Save Saves & Follows"],
+      ["/user/alerts", "Watch lot Alerts"],
```
dblandin, line 56: **question**: is the title of "Watch lot Alerts" correct here?

src/v2/Components/NavBar/Menus/NavBarUserMenu.tsx
```diff
@@ -73,6 +76,16 @@ export const NavBarUserMenu: React.FC = () => {
...
       </NavBarMenuItemLink>
 
+      {isSavedSearchEnabled && (
+        <NavBarMenuItemLink
+          aria-label="View your saved alerts"
+          to="/user/alerts"
+          onClick={trackClick}
+        >
+          <BellIcon mr={1} aria-hidden="true" /> Alerts
+        </NavBarMenuItemLink>
```
dblandin, line 86: **question**: I noticed some flashing between server/client renders in the tab bar. If it's tricky to resolve in this PR another option would be to remove the conditional tab/menu renders and have this PR only expose the route, accessible manually via `/user/alerts`.

dblandin (commented):

src/v2/Components/NavBar/Menus/__tests__/NavBarUserMenu.jest.tsx
dimatretyak: Yes, because it also includes SVG image title (see the comment above)

src/v2/Components/NavBar/Menus/NavBarUserMenu.tsx
dimatretyak: > it's tricky to resolve in this PR

You are absolutely right. I tried to solve this problem, but in the end nothing helped. If the option to leave only the path `/user/alerts` is suitable, then I will do it 👍

src/typings/sharify.d.ts
dimatretyak: If the alert pages will only be accessible via a direct link `users/alerts`, then we don't need to delete `readonly` for tests.

src/v2/Components/NavBar/Menus/NavBarUserMenu.tsx
dblandin: @dimatretyak Yes, I think that would be fine for this iteration.

damassi:
As a heads up, not that grow team is rebuilding the /settings app under `v2/Apps/Settings` ATM and these files will eventually be moved over there (but not immediately).

dimatretyak:
> As a heads up, not that grow team is rebuilding the /settings app under `v2/Apps/Settings` ATM and these files will eventually be moved over there (but not immediately).

`/settings2` will not be used for [this reason](https://github.com/artsy/force/pull/8814#discussion_r743890884)

damassi:
@dimatretyak - mind elaborating on whats going on? Its not clear from that thread.

dimatretyak:
> @dimatretyak - mind elaborating on whats going on? Its not clear from that thread.

Page with the list of alerts will be available only by direct link `users/alerts` (tabs will not be used)

damassi:
That makes sense. We'll still eventually move the files over since technically this is still settings-related code, it just wont appear in tab bar. (No changes needed for this PR! More of a future heads up.)

dimatretyak:
> No changes needed for this PR! More of a future heads up

Got it. Thanks for the clarifications

dimatretyak:
Based on [the comment](https://artsyproduct.atlassian.net/browse/FX-3527?focusedCommentId=43433), it was decided to close PR

dimatretyak:
Based on [the comment](https://artsyproduct.atlassian.net/browse/FX-3527?focusedCommentId=43433), it was decided to close PR

PR #8818, opened 2021-11-03
fix(router): Make onServerSideRender calls async

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-650 

We noticed that for logged in artsymail users sometimes the image would appear, rather than the page. This is due to the route being improperly mounted. 

Looking deeper, noticed that we weren't properly `await`ing our `onServerSideRender` calls, so this updates that.

### DIFF
--- a/src/v2/Apps/Artwork/Server/handleArtworkImageDownload.ts
+++ b/src/v2/Apps/Artwork/Server/handleArtworkImageDownload.ts
@@ -13,6 +13,6 @@ export const handleArtworkImageDownload = async ({ req, res }) => {
     if (req.user) {
       imageRequest.set("X-ACCESS-TOKEN", req.user.get("accessToken"))
     }
-    req.pipe(imageRequest, { end: false }).pipe(res)
+    res.redirect(302, imageRequest.url.replace("original", "normalized"))
   }
 }
--- a/src/v2/Apps/Artwork/Server/__tests__/handleArtworkImageDownload.jest.ts
+++ b/src/v2/Apps/Artwork/Server/__tests__/handleArtworkImageDownload.jest.ts
@@ -27,12 +27,11 @@ describe("artworkMiddleware", () => {
         canDownload: true,
         get: jest.fn(),
       },
-      pipe: () => ({
-        pipe: spy,
-      }),
     }
 
-    const res = {}
+    const res = {
+      redirect: spy,
+    }
     await handleArtworkImageDownload({ req, res })
     expect(spy).toHaveBeenCalled()
   })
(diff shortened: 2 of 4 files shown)


dzucconi (approved):

src/v2/Apps/Artwork/Server/handleArtworkImageDownload.ts
```diff
@@ -13,6 +13,6 @@ export const handleArtworkImageDownload = async ({ req, res }) => {
     if (req.user) {
       imageRequest.set("X-ACCESS-TOKEN", req.user.get("accessToken"))
     }
-    req.pipe(imageRequest, { end: false }).pipe(res)
+    res.redirect(302, imageRequest.url.replace("original", "normalized"))
```
damassi, line 16: Rather than piping and streaming, we can just open the image in a new tab. (Not sure whats up with `pipe` right here, but this achieves the same end goal. Also worth noting that if signed in as an admin, one can right click and hit save as right on the image.)

dzucconi: wait this works? doesn't require auth?

dzucconi: we dont want to serve the normalized tho

dzucconi: id leave this alone. normalized is a totally differnet and lossy file

dzucconi: its streamed because we dont want to buffer the whole thing into memory

damassi: > doesn't require auth?

Can't get to this spot till you've auth'd.

damassi: or actually, that's the spot above. I don't know... we can ship this and update so folks aren't freaked out by not being able to see artwork pages occasionall

dzucconi: That's fine but we definitely need to serve the original here. normalized is not the same file. Why change this?

damassi: As mentioned in the thread `original` would only return a weird small square

dzucconi: `original` definitely works — it was working fine in replication Isabelle showed me.

PR #8823, opened 2021-11-04
fix: updates icons

dzucconi:
Closes: [GRO-579](https://artsyproduct.atlassian.net/browse/GRO-579)

This PR:
* Adds an SVG favicon that adapts to dark-mode and prefers it
* Gets rid of the many-sized apple-touch-icons in favor of the current size: 180x180, and updates it getting rid of the purple
* Adds a manifest.webmanifest file for Androids icons... We could flesh this out with some other data.

Those last two aren't really critically important — I doubt anyone is adding our PWA to their home screen but so be it.

@icirellik — would it be preferable to use the `NOVO_MANIFEST.lookup` for the image files? Also it seems like the images are uploaded, un-fingerprinted so that manifest shouldn't be a problem but obviously the paths won't be pointed at the Cloudfront distro. I don't think this is going to be a problem. I should try to track down where we're serving static files in the app.


damassi:
Re static files, see [here](https://github.com/artsy/force/blob/master/src/middleware.ts#L146)

PR #8830, opened 2021-11-05
feat(SWA-102): Prepopulate contact information step with information from the session storage

Serge0n:
Jira Ticket:ticket: : [SWA-102](https://artsyproduct.atlassian.net/browse/SWA-102)

This PR adds the ability to pre-populate `ContactInformationForm` fields from `session storage`. Fields from session storage override values that had been taken from the user profile. It is really helpful for example when the user  lost connection after he clicked `Submit Artwork`

**VIDEO:clapper: :**

https://user-images.githubusercontent.com/55637696/140531705-4b00799f-c7b8-437e-9e33-afb37f3afec4.mov

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
@@ -5,13 +5,6 @@ import { useRouter } from "v2/System/Router/useRouter"
 import { ContactInformation_me } from "v2/__generated__/ContactInformation_me.graphql"
 import { useSubmission } from "../../Utils/useSubmission"
 
-export const getContactInformationFormInitialValues = (
-  me: ContactInformation_me
-) => ({
-  name: me?.name || "",
-  email: me?.email || "",
-  phone: me?.phone || "",
-})
 export interface ContactInformationFormModel {
   name: string
   email: string
@@ -32,7 +25,6 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
     handleBlur,
     resetForm,
     validateForm,
-    setErrors,
   } = useFormikContext<ContactInformationFormModel>()
 
   const {
@@ -45,13 +37,7 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
   useEffect(() => {
     if (submission) {
       resetForm({ values: submission.contactInformationForm })
-      validateForm(submission.contactInformationForm).then(e => {
-        setErrors(e)
-      })
-    } else {
-      resetForm({
-        values: getContactInformationFormInitialValues(me),
-      })
+      validateForm(submission.contactInformationForm)
     }
   }, [submission])
 
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
@@ -1,20 +1,59 @@
 import { Box, BoxProps, Input } from "@artsy/palette"
 import { useFormikContext } from "formik"
+import { useEffect } from "react"
+import { useRouter } from "v2/System/Router/useRouter"
+import { ContactInformation_me } from "v2/__generated__/ContactInformation_me.graphql"
+import { useSubmission } from "../../Utils/useSubmission"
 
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
```diff
@@ -1,20 +1,59 @@
...
+  phone: me?.phone || "",
+})
 export interface ContactInformationFormModel {
   name: string
   email: string
   phone: string
 }
 
-export interface ContactInformationFormProps extends BoxProps {}
+export interface ContactInformationFormProps extends BoxProps {
```
AlicanAkyuz, line 21: is there any reason why form props extend BoxProps?

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
```diff
@@ -1,20 +1,59 @@
...
+      params: { id },
+    },
+  } = useRouter()
+  const { submission } = useSubmission(id)
+
+  useEffect(() => {
+    if (submission) {
+      resetForm({ values: submission.contactInformationForm })
+      validateForm(submission.contactInformationForm).then(e => {
+        setErrors(e)
```
AlicanAkyuz, line 49: it looks like these errors are effective only for enabling/disabling the submit button, but we don't seem to display them in the ui. I'm wondering if there's any reason why we don't display them to the user?

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
Serge0n: To be able manage `ContactInformationForm` like this (by any box props): https://github.com/artsy/force/blob/8259853334cff10c71753923d20333f8e9cea084/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx#L107

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
Serge0n: requirements)

Serge0n: You can check this ticket for example: [SWA-66](https://artsyproduct.atlassian.net/browse/SWA-66)

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
AlicanAkyuz: oh I see! 

**non-blocking**: in this case, my suggestion would be to control the spacing in the parent, something like the below. it'll increase the readability, and we can get rid of passing/spreading the props while we can. 

```
<Box my={6}>
     <ContactInformationForm me={me} />
</Box>
```

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
AlicanAkyuz: then we can get rid of the chaining `.then(e => { setErrors(e) })` -> because it's creating the impression that errors are handled, although they are not. 

what do you think?

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
Serge0n: This is an existing pattern in Force to avoid unnecessary wrappers.

AlicanAkyuz (approved):

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
Serge0n: I think you were right, fixed

kiryl-zubarau (approved):
🚀

PR #8833, opened 2021-11-05
chore(buttons): use "Show More" / primary variant for pagination buttons

dzucconi:
No ticket here; just noticed we didn't have a consistent label/variant for these and after consulting with Jonathan we are settling on `"Show More"` with the `primaryBlack` (default) button variant.


src/v2/Apps/FeatureAKG/__tests__/FeatureAKGRoute.jest.tsx
```diff
@@ -272,7 +272,7 @@ describe("FeatureAKG", () => {
           action_type: "Click",
           context_module: "Editorial",
           destination_path: "/articles",
-          subject: "View more",
+          subject: "Show More",
```
dzucconi, line 275: Oh this one is probably erroneous.

PR #8834, opened 2021-11-05
chore(home): use higher resolution image versions

dzucconi:
Minor thing — other day noticed a very poor quality image on this rail. We weren't using the largest possible source versions so this changes that to prefer them.

### DIFF
--- a/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
@@ -14,7 +14,7 @@ import {
   Flex,
 } from "@artsy/palette"
 import { compact, take } from "lodash"
-import * as React from "react";
+import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useTracking } from "v2/System"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Home/Components/HomeFeaturedEventsRail.tsx
```diff
@@ -138,13 +138,21 @@ export const HomeFeaturedEventsRailFragmentContainer = createFragmentContainer(
...
                 src
                 srcSet
                 width
                 height
               }
-              large: cropped(width: 445, height: 297) {
+              large: cropped(
+                width: 445
+                height: 297
+                version: ["main", "wide", "large_rectangle"]
```
damassi, line 154: Interesting - i've seen this around but unsure what it does?

dzucconi: Chooses which version to process — images aren't guaranteed to have all the versions. In this case it was defaulting to `large_rectangle` which is actually pretty small. So we're saying here to prefer main, if that's not available, then go to wide, then to large_rectangle.

PR #8836, opened 2021-11-05
feat(GRO-635): Add unique title on for sale page

jacobherrington:
Solves: https://artsyproduct.atlassian.net/browse/GRO-635

I wasn't sure if it was overkill to add a Cypress test here, so I just left things at a unit test.

### DIFF
--- a/src/v2/Apps/Artist/Utils/computeTitle.ts
+++ b/src/v2/Apps/Artist/Utils/computeTitle.ts
@@ -1,13 +1,16 @@
 /**
- * Computes title meta tage value for an artist
+ * Computes title meta tag value for an artist
  */
-export function computeTitle(name: string, count: number) {
-  const parts = [
-    name,
-    "-",
-    count ? count + " Artworks," : null,
-    "Bio & Shows on Artsy",
-  ]
-  const titleString = parts.filter(Boolean).join(" ")
-  return titleString
+export function computeTitle(
+  name: string,
+  count: number,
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Artist/Utils/computeTitle.ts
```diff
@@ -1,13 +1,16 @@
...
-  ]
-  const titleString = parts.filter(Boolean).join(" ")
-  return titleString
+export function computeTitle(
+  name: string,
+  count: number,
+  forSaleOnly: boolean = false
+) {
+  if (count > 0) {
+    return `${name} - ${count} Artworks${
```
dzucconi, line 10: I realize this logic was already here as is but:

* The count might be 1 so we'd want to conditionally add the s to artworks
* We might not have a count at all

dzucconi: (Oh I see you accommodate the count not existing)

dzucconi (approved):

src/v2/Apps/Artist/Utils/computeTitle.ts
jacobherrington: In the case that count is undefined, it should evaluate as false, right?

dzucconi:
```js
undefined > 0
// false
```

👍

PR #8846, opened 2021-11-08
feat(GRO-596): Edit profile / About You

dzucconi:
This PR implements the top half of the form. You can see in the TODOs I need to expose some Metaphysics fields for price range to support this. Will follow up.


src/v2/Apps/Settings/Routes/EditProfile/Components/__tests__/SettingsEditProfileAboutYou.jest.tsx
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  it("submits the form", async () => {
+    renderWithRelay()
+
+    fireEvent.change(screen.getByPlaceholderText("Profession"), {
+      target: { name: "profession", value: "Artist" },
+    })
+
+    await waitFor(() => {
```
dzucconi, line 50: In the future you'll probably run into this again: with RTL, Formik's `onSubmit` needs to be waited on with `waitFor` otherwise you get an unmounted component update error.

src/v2/Apps/Settings/settingsRoutes.tsx
```diff
@@ -152,20 +159,41 @@ export const settingsRoutes: AppRouteConfig[] = [
         onClientSideRender: () => {
           PurchasesRoute.preload()
         },
+        query: graphql`
+          query settingsRoutes_PurchasesRouteQuery {
+            me {
+              ...SettingsPurchasesRoute_me
+            }
+          }
+        `,
```
dzucconi, line 168: I added these remaining queries to all the routes missing them

damassi (approved):
We should keep a list of the patterns being established here (toast, etc) so that we can apply them uniformly across the other tabs 👍

PR #8847, opened 2021-11-09
refactor(NX-2867): refactor inquiryModal and upgrades tests

araujobarret:
[NX-2867]

This PR was supposed to refactor a bunch of things in order to make everything testable, checking the code and discussing with the team I haven't found any major anti-pattern that blocks testing in the Conversation App Components, but I see some attempts to test bigger components instead of testing the smaller ones first.

- Upgrade all conversation tests to use RTL
- Refactor ConversationCTA to improve readability and import the right fragment
- OpenInquiryModalButton is now OpenInquiryCTA and was refactored to not use QueryRenderer, we only need a Fragment there

[NX-2867]: https://artsyproduct.atlassian.net/browse/NX-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Conversation/Components/ConversationCTA.tsx
```diff
@@ -19,30 +19,31 @@ export const ConversationCTA: React.FC<ConversationCTAProps> = ({
   openOrderModal,
 }) => {
   // Determine whether we have a conversation about an artwork
-
   const firstItem = conversation?.items?.[0]?.item
   const artwork = firstItem?.__typename === "Artwork" ? firstItem : null
-  const artworkID = artwork?.internalID
 
-  let CTA: JSX.Element | null = null
```
araujobarret, line 27: improved removing this CTA assignment for readability

src/v2/Apps/Conversation/Components/OpenInquiryModalButton.tsx
```diff
@@ -1,108 +0,0 @@
...
-          onClick={() => handleOpenModal()}
-        >
-          Make Offer
-        </Button>
-      </Flex>
-    </>
-  )
-}
-
-export const OpenInquiryModalButtonQueryRenderer: React.FC<{
```
araujobarret, line 70: This wasn't necessary, this acts as a very dummy component, just a Fragment does the job in this case.
please @damassi correct me if I'm wrong

src/v2/Apps/Conversation/Components/OpenInquiryModalCTA.tsx
```diff
@@ -0,0 +1,75 @@
...
+          onClick={() => handleOpenModal()}
+        >
+          Make Offer
+        </Button>
+      </Flex>
+    </>
+  )
+}
+
+export const OpenInquiryModalCTAFragmentContainer = createFragmentContainer(
```
araujobarret, line 66: the only change from OpenInquiryModalButton is we have only a Fragment and we control if we're going to render it or not from the caller, we don't need to check `isOfferableFromInquiry` here.

src/v2/Apps/Conversation/Components/__tests__/Conversation.jest.tsx
```diff
@@ -55,28 +64,14 @@ describe("Conversation", () => {
       }
     })
 
-    it("shows the buyer guarantee message", () => {
```
araujobarret, line 58: some pointless test separated in different tests, grouped them in the test below

src/v2/Apps/Conversation/Components/__tests__/Item.jest.tsx
```diff
@@ -1,10 +1,10 @@
-import { mount } from "enzyme"
+import { screen, render } from "@testing-library/react"
 import { Item } from "../Item"
 import { Conversation_conversation } from "v2/__generated__/Conversation_conversation.graphql"
 
 describe("Item", () => {
   describe("when inquiry item is an artwork", () => {
-    // @ts-expect-error STRICT_NULL_CHECK
+    // @ts-ignore
```
araujobarret, line 7: I'm still not 100% which to use, I saw other reviews people asking to use @ts-ignore instead of the first.

src/v2/Apps/Conversation/Components/ConversationCTA.tsx
```diff
@@ -52,11 +53,8 @@ export const ConversationCTA: React.FC<ConversationCTAProps> = ({
       )
     }
   }
-  if (!CTA) {
-    return null
-  } else {
-    return <>{CTA}</>
-  }
+
+  return null
```
damassi, line 57: Nice cleanup here ➕

src/v2/Apps/Conversation/Components/OpenInquiryModalCTA.tsx
```diff
@@ -0,0 +1,75 @@
...
+  Separator,
+} from "@artsy/palette"
+import { createFragmentContainer, graphql } from "react-relay"
+import { useTracking } from "react-tracking"
+import { tappedMakeOffer } from "@artsy/cohesion"
+import styled from "styled-components"
+import { OpenInquiryModalCTA_conversation } from "v2/__generated__/OpenInquiryModalCTA_conversation.graphql"
+
+export const ShadowSeparator = styled(Separator)`
+  box-shadow: 0 -1px 1px rgba(50, 50, 50, 0.1);
```
damassi, line 17: Can we use the shadows [from palette](https://palette.artsy.net/utils/shadow/)?

src/v2/Apps/Conversation/Components/__tests__/ConversationMessages.jest.tsx
```diff
@@ -1,90 +1,89 @@
...
-    )
-    expect(wrapper.find("Message").length).toBe(2)
-    expect(wrapper.find("OrderUpdate").length).toBe(2)
+            },
+          },
+        ],
+      },
+    },
+  ],
+} as any
```
damassi, line 76: (nonblocking) - with our relay test tools we can lean on auto fixtures and only mock out those properties that we need to test behavior of the component. Aware that this stuff was here before, but the less fixture data the better.

src/v2/Apps/Conversation/Components/__tests__/Item.jest.tsx
damassi: `// @ts-ignore` is correct! Here's the backstory: https://github.com/artsy/force/blob/master/docs/best_practices.md#avoid-copying-and-try-to-fix--ts-expect-error-strict_null_check-flags

damassi: One thing however, is rather than switch to `// @ts-ignore` we'll want to try to fix the error. With the `STRICT_NULL_CHECK` flag we know that this was added by the codemod, and is waiting for a user to address.

src/v2/Apps/Conversation/Components/__tests__/Item.jest.tsx
```diff
@@ -22,25 +22,24 @@ describe("Item", () => {
...
+      )
+      expect(screen.getByText("Banksy")).toBeInTheDocument()
+      expect(screen.getByText("Untitled / June 22, 2020")).toBeInTheDocument()
+      expect(screen.getByText("$2000")).toBeInTheDocument()
     })
   })
 
   describe("when inquiry item is a show", () => {
-    // @ts-expect-error STRICT_NULL_CHECK
+    // @ts-ignore
```
damassi, line 42: ditto here (and elsewhere where STRICT_NULL_CHECK was swapped)

damassi (approved):
PR looks great @araujobarret. Lots of good cleanup / simplification all about. Only a couple of minor comments, but after that 🚢

src/v2/Apps/Conversation/Components/OpenInquiryModalButton.tsx
damassi: Yup, no need for unnecessary stuff! thanks for cleaning up.

src/v2/Apps/Conversation/Components/__tests__/Conversation.jest.tsx
```diff
@@ -2,13 +2,22 @@ import { ConversationPaginationContainer } from "../Conversation"
...
 jest.unmock("react-relay")
 jest.mock("v2/System/useSystemContext")
-
-const { getWrapper } = setupTestWrapper<ConversationPagination_Test_Query>({
+jest.mock("v2/Apps/Conversation/Components/ConfirmArtworkModal", () => ({
+  ConfirmArtworkModalQueryRenderer: () => null,
+}))
+jest.mock("v2/Apps/Conversation/Components/UnreadMessagesToast", () => ({
+  UnreadMessagesToastQueryRenderer: () => null,
+}))
```
damassi, line 16: One thing we've been doing for testing existence-y components like this in RTL is to do 
```tsx
jest.mock("v2/Apps/Conversation/Components/UnreadMessagesToast", () => ({
  UnreadMessagesToastQueryRenderer: () => "UnreadMessagesToastQueryRenderer",
}))
```
which then mirrors our old enzyme tests in that we can do 
```tsx
expect(screen.queryByText("UnreadMessagesToastQueryRenderer")).toBeInDocument()
```
This may not be relevant here, but just a heads up for testing overall container-component composition.

damassi (approved):
Killer PR @araujobarret. Thanks for moving this stuff over 👍

PR #8849, opened 2021-11-09
feat(FX-3443): create saved search alert modal

tr-ann:
Jira: [FX-3443](https://artsyproduct.atlassian.net/browse/FX-3443)

### Blocked by
- spec for `CreateAlertButton` behavior (should add new button variant to palette)
- how to pass saved search info (for creating alert) to `ArtworkFilters`

### Description
- add Formik form setup
- Includes inputs for name and notification preferences
- Triggers GraphQL mutation on save
- Has cancel button

### Changes
- Added modal with simple Formik form
- Added mutation to create saved search alert
- Added base filters parser to create pill (used for generate default saved search alert name)

### Demo
**Desktop**

https://user-images.githubusercontent.com/56556580/140916324-7528ffe2-5ad2-4fd2-940b-d3ea637a4eb1.mov

**Mobile**

https://user-images.githubusercontent.com/56556580/140926174-ac37e426-4d77-4b1d-9ceb-80aa2107d492.mov

_check if alert have created_
image

### Notes
- When user create alert on web and check it on Eigen there are also width and height pills in alert page. It happens because on Force custom size always defined (\*-\* by default) unlike the Eigen
image
- Pills extractor has a lot of bugs and need improvement
  - A: we have separate ticket for that
- We also should add alert when we create saved search alert like "success/fail creation"
- Should we refetch search criteria and disable "create alert" button (if we have criteria with the same filters) after creating alert?
  - A: also have to do this in separate ticket


dimatretyak:
> When user create alert on web and check it on Eigen there are also width and height pills in alert page. It happens because on Force custom size always defined (*-* by default) unlike the Eigen

I think it would be better in this case not to send the `width`, `height` values to the mutation at all

dimatretyak:
> Should we refetch search criteria and disable "create alert" button (if we have criteria with the same filters) after creating alert?

We definitely have to do this (I think it will be necessary to make this a separate task). @mdole what do you think?

dimatretyak:
> Pills extractor has a lot of bugs and need improvement

I also think it would be better to do this in a separate task

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -88,7 +91,12 @@ export const ArtworkFilter: React.FC<
       onChange={onChange}
       ZeroState={ZeroState}
     >
-      <ArtworkFilterRefetchContainer viewer={viewer} {...rest} />
+      <ArtworkFilterRefetchContainer
+        viewer={viewer}
+        artistId={artistId}
+        artistName={artistName}
```
dimatretyak, line 97: **suggestion**: instead of passing `artistName`, `artistId`, we can pass `viewer`. In the future, we will be able to extract all the values we need from the `viewer` directly and this will make it easier for us to implement the "saved search" button on other artwork grids (artist series, collection, etc).

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -309,6 +321,18 @@ export const BaseArtworkFilter: React.FC<
 
           <Spacer mb={2} />
 
+          {enableCreateAlert && artistId && artistName && (
+            <FiltersPills
+              artistId={artistId}
+              artistName={artistName}
+              aggregations={filterContext.aggregations!}
+              filters={filterContext.currentlySelectedFilters?.() || {}}
```
dimatretyak, line 329: **suggestion**: we don't have to pass `filters` as a separate prop. We can just use the `useArtworkFilterContext` hook in the component (in our case, `SavedSearchAlertForm`)

dimatretyak (approved):
LGTM. If it's not too much trouble, can you add tests for your code?

src/v2/Components/SavedSearchAlert/SavedSearchAlertModel.tsx
```diff
@@ -0,0 +1,25 @@
...
+  artistName: string
+}
+
+export interface SavedSearchAlertMutationResult {
+  id: string
+}
+
+export interface SavedSearchAlertUserAlertSettings {
+  name: string
+  email?: boolean
```
dimatretyak, line 24: you can make `email` **not as optional**. In Eigen it is now optional for the reason that notification toggles under the feature flag

src/v2/Components/SavedSearchAlert/CreateSavedSearchAlert.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react"
```
damassi, line 1: If `SavedSearchAlert` isn't shared across other apps / components we should avoid putting it in the shared `v2/Components` library and instead co-locate it with the ArtworkFilter (or wherever most appropriate)

src/v2/Components/ArtworkFilter/SavedSearch/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+  "acquireable",
+  "atAuction",
+  "inquireableOnly",
+  "offerable",
+  "materialsTerms",
+  "priceRange",
+  "dimensionRange",
+  "height",
+  "width",
+]
```
damassi, line 34: into -> SavedSearch/Utils

src/v2/Components/SavedSearchAlert/SavedSearchAlertForm.tsx
```diff
@@ -0,0 +1,97 @@
...
+          criteria
+        )
+
+        const result = {
+          id: response.createSavedSearch?.savedSearchOrErrors.internalID!,
+        }
+
+        onComplete?.(result)
+      } catch (error) {
+        console.error(error)
```
damassi, line 63: we should use our logger (v2/Utils/logger)

src/v2/Components/SavedSearchAlert/CreateSavedSearchAlert.tsx
tr-ann: As I know `SavedSearchAlert ` will be used by [collection](https://artsyproduct.atlassian.net/browse/FX-3551) page, [categories](https://artsyproduct.atlassian.net/browse/FX-3552) etc.
So I think we should leave it in Components

damassi: 👍

src/v2/Components/ArtworkFilter/SavedSearch/CreateAlertButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+  &:hover {
+    div {
+      color: ${themeGet("colors.white100")};
+    }
+    svg {
+      fill: ${themeGet("colors.white100")};
+    }
+  }
+`
```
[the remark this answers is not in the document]
dzucconi: Yeah I'm not sure what's going on here but definitely don't override anything. If you want the SVG to match the button hover colors just pass it fill="currentColor"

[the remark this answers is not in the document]
tr-ann: Maybe we should add this component to palette? Or just creating separate component which is don't related to palette's button?
@damassi @dzucconi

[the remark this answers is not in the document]
dzucconi: What is this supposed to look like? Can you show a screen shot? It doesn't appear to be particularly novel?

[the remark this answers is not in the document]
tr-ann: https://user-images.githubusercontent.com/56556580/141318664-42b31da5-7073-4e98-9f18-ebbeccc3f0af.mov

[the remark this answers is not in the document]
dzucconi: I'll release this in a moment: https://github.com/artsy/palette/pull/1071

Just easier support for this sort of thing. Basically just drop in an icon with `fill="currentColor"`. We don't want to neutralize the padding either.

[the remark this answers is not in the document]
dzucconi: Also this should not have blue text. I'm not sure where that's coming from but there's nothing in the design system like that — that goes against our hover states and is confusing.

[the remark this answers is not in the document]
tr-ann: Oh, thank you, it is very helpful. I'll fix it asap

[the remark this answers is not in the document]
dzucconi: This should get released in a bit; so you'll have to bump Palette. But once that's done:
https://github.com/artsy/palette/blob/0765be7978e5c4fcf6d389c0e22150050f5c4492/packages/palette/src/elements/Button/Button.story.tsx#L172-L175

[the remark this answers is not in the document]
tr-ann: But [here](https://www.figma.com/file/zbEFZqjKgKquZwtU8U44j9/Alerts?node-id=3475%3A6831) is design of the button and it is blue.
Should I clarify it with designers?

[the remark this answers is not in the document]
dzucconi: Yeah: we'd need a spec for every single state and we'd have to add a new variant to Palette if this is really something that they want.

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -130,6 +138,8 @@ export const BaseArtworkFilter: React.FC<
     Filters?: JSX.Element
     offset?: number
     enableCreateAlert?: boolean
+    artistId?: string
+    artistName?: string
```
damassi, line 142: (blocking) Similar to @dimatretyak's comment above, we shouldn't be passing feature-specific props to our generic artwork filter

src/v2/Components/SavedSearchAlert/SavedSearchAlertForm.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const SavedSearchAlertForm: React.FC<SavedSearchAlertFormProps> = ({
+  artistId,
+  artistName,
+  initialValues,
+  onComplete,
+}) => {
+  const logger = createLogger(
+    "v2/Components/SavedSearchAlert/SavedSearchAlertForm"
+  )
```
damassi, line 30: this can be moved outside of the react component

src/v2/Components/ArtworkFilter/index.tsx
tr-ann: Can we pass an object with needed data?
Something like this:
```
{
  type: "artist",
  name: "artist name",
  id: "artist id",
}
```
Or we should pass all info with viewer? It may cause some problems with viewer types, because we have viewer type for artist, collection, artist series etc. So we'll should check type of viewer and if it is `ArtistArtworkFilter_artist` render `CreateAlertButton`

tr-ann: @damassi What do you think? ☝️

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -29,10 +31,16 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
     return null
   }
 
+  const savedSearchAttributes: SavedSearchAttributes = {
+    type: "artist",
+    id: artist.internalID,
+    name: artist.name || "",
```
dimatretyak, line 37: **suggestion**: `artist.name ?? ""`

src/v2/Components/SavedSearchAlert/SavedSearchAlertModel.tsx
```diff
@@ -0,0 +1,13 @@
+export interface SavedSearchAleftFormValues {
+  name: string
+  email: boolean
+}
+
+export interface SavedSearchAlertFormPropsBase {
+  artistId: string
+  artistName: string
+}
```
dimatretyak, line 9: this interface is not used anywhere

dimatretyak (approved):

src/v2/Components/SavedSearchAlert/SavedSearchAlertForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+          name="name"
+          placeholder={namePlaceholder}
+          value={formik.values.name}
+          onChange={formik.handleChange("name")}
+          onBlur={formik.handleBlur("name")}
+          error={formik.errors.name}
+          maxLength={75}
+        />
+        <Box display="flex" justifyContent="space-between" my={4}>
+          <Text>Email Alerts</Text>
```
dzucconi, line 93: Anytime you use `Text` in v3 it should have a `variant`

tr-ann:
@dzucconi Can we merge this changes without right _create alert_ button behavior because they block other work? This changes are behind feature flag and we can fix button behavior at any time.
What do you think?

dzucconi:
@tr-ann Yeah the button shouldn't be blocking anything: feel free

PR #8853, opened 2021-11-09
feat: selectively load Segment destinations based on OneTrust consent

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3593

To integrate OneTrust with Segment.

So far we have been treating Segment as `Performance` category such that consent for that category is required for Segment to fire, and if there's no such consent, Segment does not fire. This does not work for us because Segment manages a bunch of 3rd parties each of which can belong to a different category. Also, we have since [decided](https://artsyproduct.atlassian.net/browse/PLATFORM-3397?focusedCommentId=43093) that Redshift and Braze should be categorized as `Strictly Necessary` and should always fire.

Therefore, we have to treat each Segment destination separately. We have to map each destination to a OneTrust category (OneTrust categories are listed in [this notion doc](https://www.notion.so/artsy/Cookie-Consent-01dace52a765476ea4e9353260ce9a7f)), and load that destination only if there's consent.

The solution here is borrowed from:
-  a Github [gist](https://gist.github.com/sperand-io/4725e248a35d5005d68d810d8a8f7b29) provided by Segment as an example for integrating with TrustArc (similar to OneTrust).
- [code](https://github.com/littledata/shopify-tracker/blob/ec4fac54cd56344980aa481800ee8dba550ea97d/src/customerPrivacyAPI/oneTrust.ts#L31) from LittleData integrating Shopify with Onetrust.

How it works:

- If OneTrust is disabled, we load Segment normally, all Segment destinations will be loaded.

- If OneTrust is enabled:
  - Fetch list of destinations from Segment integrations API. In the fetch URL, supply Force staging's or prod's `SEGMENT_WRITE_KEY`. The key maps to Segment `force-staging` and `force-production` Sources. There are other Segment write keys in Force's config but they don't seem applicable. Each Segment Source has Enabled and Disabled Destinations. This API returns the Enabled ones only, and it omits Redshift (but Redshift will get events as I've [verified](https://artsyproduct.atlassian.net/browse/PLATFORM-3728).) If the fetch fails (e.g. API down, bad data returned, ...), we load Segment without 3rd party destinations, that is, only Segment itself and Redshift will get events.
  - Obtain OneTrust consent which is stored in OneTrust's `OnetrustActiveGroups` object. Wait for OneTrust to initialize, if required. If OneTrust never becomes ready, we set consent to only `Strictly Necessary`.
  - For each destination returned by Segment API, consult the mapping (currently stored in the code) that maps destination names to OneTrust consent categories. See whether OneTrust has consent for the destination. If it has, set the destination's preference to `true` (will be loaded), otherwise `false` (will not be loaded). If mapping is not found for the destination, which would be the case if we add a new destination on Segment's side and have not yet updated the mapping in the code, set the destination's preference to `false`.
  - When all the destinations are examined, call Segment's `window.analytics.load`, pass it the preferences (via `integrations` param).
  - Register with OneTrust's hook (`OptanonWrapper` function) for triggering Segment reload (via page reload) when there is a consent change.


src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+    const interval = setInterval(() => {
+      if (oneTrustReady() || attempts > maxAttempts) clearInterval(interval);
+      attempts++;
+    }, 10);
+  }
+}
+
+function oneTrustReady() {
+    // OneTrust is ready if OnetrustActiveGroups contains at least Strictly Necessary group.
+    if ((typeof window.OnetrustActiveGroups === 'string') && (window.OnetrustActiveGroups.split(',').includes('C0001'))) {
```
artsyjian, line 44: Observation: On page load, `OnetrustActiveGroups` is `,,` (just commas). It looks like after OneTrust banner script loaded, the string becomes `,C0001,`.

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+
+  if (!res.ok) {
+    throw new Error(
+      `Failed to fetch Segment destinations: HTTP ${
+        res.status
+      } ${res.statusText}`
+    );
+  }
+
+  const destinations = await res.json();
```
artsyjian, line 145: returns:
```
[
  {
    "name": "Indicative",
    "creationName": "Indicative",
    "description": "Indicative is a behavioral analytics platform designed to help Marketing and Product teams optimize user engagement, conversion, and retention. It's easy to setup and easy to use—no SQL or coding knowledge required—and offers analysts a single view of customer behavior across all digital touchpoints. Get started by creating an account.",
    "website": "https://app.indicative.com/?utm_source=segment&utm_medium=partners&utm_campaign=integrations-page#/login/register",
    "category": "Analytics"
  },
  {
    "name": "Braze",
    "creationName": "Appboy",
    "description": "Braze lets you understand, engage, monetize and maximize the lifetime value of your app users.",
    "website": "https://www.braze.com/",
    "category": "SMS & Push Notifications"
  },
...
```

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+    writeKey: SEGMENT_WRITE_KEY,
+    destinations: SEGMENT_DESTINATIONS,
+    destinationPreferences
+  });
+}
+
+function setSegmentDestinationPref(oneTrustConsent, destinations) {
+  // maps Segment destination's category to OneTrust cookie category.
+  const segmentToOneTrust= {
+    'SMS & Push Notifications': 'C0001', // OneTrust Strictly Necessary
```
artsyjian, line 84: because Segment's category for Braze is `SMS & Push Notifications`

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+    'Analytics': 'C0002', // OneTrust Performance
```
artsyjian, line 85: includes Segment `Amazon S3` destination.

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+      }
+    })
+    .reduce(
+      (acc, val) => {
+        return {
+          ...val,
+          ...acc
+        };
+      },
+      { "Segment.io": oneTrustConsent.some(d => d === SEGMENT_MAPPING) }
```
artsyjian, line 105: Not exactly sure what `Segment.io` maps to.

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+
+  for (const destination in destinationPreferences) {
+    const isEnabled = destinationPreferences[destination];
+    if (isEnabled) {
+      isAnythingEnabled = true;
+    }
+  }
+
+  if (isAnythingEnabled) {
+    window.analytics.load(writeKey, { integrations: destinationPreferences });
```
artsyjian, line 126: On page load this call works. But when I accept all cookies via OneTrust banner, in GDPR view, it errors:
Screen Shot 2021-11-09 at 9 36 20 AM

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+    waitForOneTrust();
+
+    if (oneTrustReady()) {
+      getConsentAndLoadSegment();
+    }
+  }
+)
+
+// OneTrust calls OptanonWrapper when there's a consent change.
+window.OptanonWrapper = getConsentAndLoadSegment;
```
artsyjian, line 27: Observation: This is always triggered once even when there's no consent change.

artsyjian:
TOODO: update script so that if OneTrust is disabled, load Segment normally.

src/v2/index.ejs
```diff
@@ -57,13 +57,7 @@
...
-    !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on","addSourceMiddleware","addIntegrationMiddleware","setAnonymousId","addDestinationMiddleware"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t,e){var n=document.createElement("script");n.type="text/javascript";n.async=!0;n.src="https://cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var a=document.getElementsByTagName("script")[0];a.parentNode.insertBefore(n,a);analytics._loadOptions=e};analytics.SNIPPET_VERSION="4.1.0";
-    analytics.load("<%%= sd.SEGMENT_WRITE_KEY %>");
-    }}();
-  </script>
-  <%% } %>
+  <% include ./segment.v2.ejs %>
```
artsyjian, line 60: Not sure about this. Perhaps the script should be somewhere else in `src`? Or be a NPM lib?

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+      (acc, val) => {
+        return {
+          ...val,
+          ...acc
+        };
+      },
+      { "Segment.io": oneTrustConsent.some(d => d === SEGMENT_MAPPING) }
+    );
+
+  return destinationPreferences;
```
artsyjian, line 108: Example of its content:
Screen Shot 2021-11-03 at 4 54 57 PM

src/v2/index.ejs
damassi: `src/lib` is a good place

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+    delete destination.creationName;
+  }
+
+  return destinations;
+}
+
+}}();
+</script>
+
+<%% } %>
```
damassi, line 159: We can likely move all of this script code into a proper JS file and initialize it on load (within the app)

src/v2/index.ejs
artsyjian: Got it. And thanks for the walk through during frontend office hours. I shall revise.

src/v2/segment.v2.ejs
joeyAghion: Why map the destinations by category (rather than, say, name)? I could imagine some destinations with the same category having different degrees of necessity.

artsyjian: Wanted to avoid coding Destination names in the script, but yeah, by category does seem coarse. Best if we can straight up tag each destination on Segment's UI with appropriate OneTrust category, and have Segment API return that. Time to involve @abhitip ! @abhitip - do you know if that's possible?

src/v2/index.ejs
artsyjian: revised.

src/v2/client.tsx
```diff
@@ -34,6 +35,8 @@ async function setupClient() {
       })
     }
 
+    fetchSegmentDestinationsAndLoad()
```
damassi, line 38: 👍

damassi (approved):
🚀

src/v2/segment.v2.ejs
artsyjian: I think this is b/c after Segment loads, it cleans up after itself - it deletes `window.analytics.load` function. That might be why the [Gist](https://gist.github.com/sperand-io/4725e248a35d5005d68d810d8a8f7b29#file-granulardecision-js-L132) does a page reload. I took that out thinking it's not required. Now I have put it back. So now, whenever the visitor changes consent (via banner, preference-center), the page reloads. @joeyAghion , @damassi - is that okay for user experience? 
onetrust segment test consent change reload

src/v2/segment.v2.ejs
artsyjian: After testing it a bunch of times, it looks like OneTrust triggers `OptanonWrapper` once every page load, even when there's no consent change. No idea why the trigger and when. I thought that it happens when OneTrust has successfully set `window.OnetrustActiveGroups`, but that's likely not the case, because the trigger happens on a page reload (`window.OnetrustActiveGroups` survives page reloads). I wonder if there's potential for a reload loop. Scenario: Page reloads, script fires, loads Segment, registers `OptanonWrapper`, a little later OneTrust triggers `OptanonWrapper`, script sees that Segment has already been initialized, script reloads page, repeat. To guard against that, I am having `getConsentAndLoadSegment` abort if there's not really a consent change, which requires saving OneTrust consent into a var.

src/lib/analytics/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,179 @@
...
+
+export function fetchSegmentDestinationsAndLoad() {
+  console.log("fetch segment destinations")
+  fetchDestinations(SEGMENT_WRITE_KEY).then(destinations => {
+    SEGMENT_DESTINATIONS = destinations
+
+    waitForOneTrust()
+
+    if (oneTrustReady()) {
+      getConsentAndLoadSegment()
```
artsyjian, line 17: In the event that OneTrust never becomes ready, Segment is not loaded. That seems bad. I will revise this so that in that case we load Segment for `Strictly Necessary` (`C0001`) destinations.

src/v2/segment.v2.ejs
damassi: I'll leave that to Joey -- its not optimal UX and reloads are generally last resort kinda stuff (like how we do it when we log out / log in) but if the user generally only has to do it once then nbd.

artsyjian: Ah, good point. Indeed, we do not yet provide user a way to recall the OneTrust banner or preference center - there's no way for user to re-consent. OneTrust saves consent in a cookie that expires in 1 year, so user will be re-prompted after a year or after clearing browser cache.

src/lib/analytics/segmentOneTrustIntegration.ts
artsyjian: revised.

src/v2/segment.v2.ejs
artsyjian: Meanwhile, I will update the notion [doc](https://www.notion.so/artsy/Cookie-Consent-01dace52a765476ea4e9353260ce9a7f) with Segment destinations - a table that maps each destination to a OneTrust category.

src/v2/segment.v2.ejs
```diff
@@ -0,0 +1,159 @@
...
+)
+
+// OneTrust calls OptanonWrapper when there's a consent change.
+window.OptanonWrapper = getConsentAndLoadSegment;
+
+function waitForOneTrust() {
+  let attempts = 0;
+  const maxAttempts = 100;
+
+  // wait 1sec at most.
```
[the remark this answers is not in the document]
artsyjian: seems fine for a maximum wait. on my local, the wait is never more than ~100ms.

artsyjian:
> TODO: update script so that if OneTrust is disabled, load Segment normally.

done

src/v2/segment.v2.ejs
joeyAghion: I agree that it's not ideal, but I think it's acceptable given how we know it affects initial script execution (and how rarely these preferences should change).

src/lib/analytics/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,209 @@
+import { data as sd } from "sharify"
+
+const SEGMENT_MAPPING = "C0001" // Segment itself assigned OneTrust Strictly Necessary category.
+const SEGMENT_WRITE_KEY = sd.SEGMENT_WRITE_KEY
+
+let PREVIOUS_CONSENT = ""
+let SEGMENT_DESTINATIONS = []
+let ONETRUST_ENABLED = false
+
+if (sd.ONETRUST_SCRIPT_ID) {
```
artsyjian, line 10: @damassi - Wonder if this is sufficient, as over there in [src/v2/index.ejs](https://github.com/artsy/force/blob/0fdf31ad41419ac08562e2a979fb82b25d496531/src/v2/index.ejs#L7), the condition is more rigorous. I can copy the condition from there, but then it seems too much. wdyt?

artsyjian:
adding more reviewers now that this looks shippable.

src/v2/segment.v2.ejs
abhitip: Finally getting to this - the table in your cookie consent doc looks good to me!

artsyjian: Thanks @abhitip. I will update the code here to reflect the [table in the doc](https://www.notion.so/artsy/Cookie-Consent-01dace52a765476ea4e9353260ce9a7f#3ec509bfb40b4d08985f6d81a306f7bf). Let me open a ticket with Segment to see if we can add a OneTrust tag to the Destinations and have their API return that tag. That would be ideal.

ovasdi (approved):
LGTM

src/v2/segment.v2.ejs
artsyjian: @abhitip -  Here's the reply from Segment. Apparently there's no way to tag Destinations. We will have to hardcode Destination-to-OneTrust-ID mapping in this script. This means we will have to remember to update this script for Destination changes.

```
Priscila (Segment)

Nov 16, 2021, 7:09 AM MST

Hi Jian,

Thanks for reaching out - happy to help here!

1) Is it possible to assign a custom attribute to each destination (via Segment UI), and have the API return that attribute along with the current ones (name/CreationName/category/...)? We want an attribute to store the OneTrust consent group (Necessary, Functional, Performance, etc.), so that the script can simply parse it from the API's response. The existing 'category' attribute doesn't work because there are multiple destinations of the same category that should be mapped differently. For example, we want the 'Amazon S3' destination (Analyticss) to map to 'Necessary', while other Analytics destinations to map to 'Performance'.


I'm afraid it's not currently possible to add custom attributes to destinations. I'm happy to pass on your feedback formally by submitting a feature request to our Engineering team. Let me know if you would like me to open a feature request? If you are able to share any specifics on the business need for this feature and the impact without this functionality, that would be helpful context for me to include in the request, too.

In the spirit of transparency, our teams are currently working through a backlog of feature requests right now. Our Product team does review every request, but I wouldn't be able to offer you a timeline as to if and when it would be looked at quite yet. Our Product team does review each and every feature and takes many things into consideration for planning which feature will be prioritized for the roadmap.


2) The API returns all our destinations except 'Redshift'. What is the reason?


By default, data warehouses doesn't populate in the integrations endpoint. Sorry about that. 


Priscila
Product Technical Lead @ Segment
```

src/v2/segment.v2.ejs
artsyjian: we will map based on destination name. `Amazon S3` will map to `Necessary`.

artsyjian:
Hey @damassi - About this [jest typescript failure](https://app.circleci.com/pipelines/github/artsy/force/21624/workflows/d3435333-2987-4735-b529-e7929418de0b/jobs/166234), do you think it affects the test only? or is it a problem at runtime as well? I don't get this error locally (`git push`, `yarn start`). It looks like that test is not picking up the [type](https://github.com/artsy/force/pull/8853/files#diff-fef83147e282c873b1f6692c6d85c0c39ba66cf601281e108e67814914d0c4c0R5) that was added? Is it advisable to add a `@ts-ignore` right above the call to [bootstrap()](https://github.com/artsy/force/blob/ac2456166259966dcf09ad09067b1ab256910e23/src/mobile/components/layout/test/layout.jest.ts#L17)?

joeyAghion (dismissed):

src/lib/analytics/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,224 @@
...
+    Indicative: "C0002",
+    "Segment.io": "C0001",
+  }
+
+  const consentArray = consent.split(",")
+
+  const basePref = {
+    // default to false for all Segment destinations.
+    // so if we don't set preference for a destination, it will not be loaded.
+    All: false,
```
joeyAghion, line 139: Does this block the unrecognized destination for all visitors or those who'v opted out of some categories?

It seems like a safer default would be `true`, requiring destinations to be treated as "necessary" unless configured otherwise in the mapping above.

artsyjian: Correct, with `All: false`, any unrecognized destination (one returned by Segment API but not listed in `segmentToOneTrust` ) will not be loaded.

I wonder how much risk there is with that. This should only be of concern when we come to add a destination on Segment's side. I imagine we will discover early (in staging perhaps) that the destination does not work, and leaern to update `segmentToOneTrust`.

With `All: true` (the default), the destination will work from the get-go. We will likely never update `segmentToOneTrust` (even if there's a playbook). The destination is likely not `Necessary`. We might violate visitor's consent, for a long time.

src/v2/segment.v2.ejs
artsyjian: Found out that if `Segment.io` is false, no events are sent from browser to Segment.

src/mobile/components/layout/bootstrap.js
```diff
@@ -11,105 +11,118 @@
...
-const CurrentUser = require('../../models/current_user');
-const { init } = require("@sentry/browser");
-const globalReactModules = require('../../../desktop/lib/global_react_modules.tsx');
-const hydrateStitch = require('@artsy/stitch/dist/internal/hydrate').hydrate;
-const { mediator } = require('../../../lib/mediator');
-const syncAuth = require('../../../lib/syncAuth.ts').default;
-const { logoutEventHandler } = require('../../../desktop/lib/deprecated_global_client_setup.tsx');
-
-export const bootstrap = function() {
+require("jquery")
```
artsyjian, line 14: so much changes to this file. all i did was to add loadSegment call.

src/lib/analytics/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+  delay,
+  fetchDestinationForWriteKey,
+  fetchDestinations,
+  getConsent,
+  getConsentAndLoad,
+  getOneTrustConsent,
+  oneTrustReady,
+  setSegmentDestinationPref,
+  validateSegmentResponse,
+}
```
damassi, line 258: just want to 👍 this pattern! I use it a lot and haven't seen others do it before now. Generally people make everything public through exports even though said exports aren't actually used. By scoping things in this way it keeps things nice and focused.

src/lib/analytics/segmentOneTrustIntegration.ts
joeyAghion: Whatever we choose, we'll have the opposite problem (either integrations not operating as expected or consent not being respected).

This is acceptable as long as we're clear about the burden on developers/analysts.

src/lib/analytics/segmentOneTrustIntegration.ts
damassi: I'd consider whatever v2 is doing as canonical.

src/lib/analytics/__tests__/segmentOneTrustIntegration.jest.ts
```diff
@@ -0,0 +1,315 @@
...
+  })
+  it("returns false if a destination has no creationName key", () => {
+    const destinations = [{ creationName: "foo" }, { blah: "blah" }]
+    expect(SOI.validateSegmentResponse(destinations)).toBe(false)
+  })
+  it("returns false if a destination has creationName key but its value is not a string", () => {
+    const destinations = [{ creationName: "foo" }, { blah: {} }]
+    expect(SOI.validateSegmentResponse(destinations)).toBe(false)
+  })
+})
```
damassi, line 315: @artsyjian - i think the testing approach you've laid out here is totally fine 👍

src/lib/analytics/__tests__/segmentOneTrustIntegration.jest.ts
```diff
@@ -0,0 +1,315 @@
...
+describe("fetchDestinationForWriteKey", () => {
+  afterAll(() => {
+    SOI.validateSegmentResponse = originalValidateSegmentResponse
+  })
+  it("returns destinations if segment api returns them", async () => {
+    const res = {
+      ok: true,
+      json: () => Promise.resolve([{ creationName: "foo" }]),
+    }
+    // @ts-ignore
```
damassi, line 142: to fix all of these `ts-ignores` can just drop 

```
declare const window: any
``` 
before the tests start

src/lib/analytics/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,258 @@
...
+  if (window.analytics.initialized) {
+    // Segment has been initialized, a page reload is required to reload Segment.
+    if (shouldReload) {
+      window.location.reload()
+    }
+    return
+  }
+
+  if (isAnythingEnabled) {
+    // @ts-ignore
```
damassi, line 193: putting 
```
declare const window: any
```
at top of file will fix all of these

src/mobile/components/layout/test/layout.jest.ts
```diff
@@ -14,6 +14,7 @@ describe("bootstrap", () => {
   })
 
   it("sets up mediator to call #logoutEventHandler", () => {
+    // @ts-ignore
```
damassi, line 17: why is this suddenly needed?

damassi:
So stunned about what is involved to get this GDPR stuff to conform! If we're paying one-trust to handle this, one would expect support for this kind of thing for a popular analytics lib would be included.

src/mobile/components/layout/bootstrap.js
artsyjian: the conversion was done by linter. have cut a [separate pr](https://github.com/artsy/force/pull/8983) for the conversion. once that pr is merged i'll rebase, then the change here will only be adding `loadSegment`.

artsyjian: done.

src/lib/analytics/__tests__/segmentOneTrustIntegration.jest.ts
artsyjian: done. thanks!

src/lib/analytics/segmentOneTrustIntegration.ts
artsyjian: done. thanks!

src/mobile/components/layout/test/layout.jest.ts
artsyjian: i was trying to resolve this error (only happens in [CI](https://app.circleci.com/pipelines/github/artsy/force/22348/workflows/ac01593a-b5ea-428e-b91f-22ecdfbcaf40/jobs/170014)):
```
FAIL   legacy  src/mobile/components/layout/test/layout.jest.ts (568 MB heap size)
  ● Console

    console.log src/lib/analytics/segmentOneTrustIntegration/segmentOneTrustIntegration.ts:17
      loadSegment called

  ● bootstrap › sets up mediator to call #logoutEventHandler

    TypeError: Cannot read property 'load' of undefined

      34 |     // OneTrust is disabled. Load Segment normally - load all Segment destinations.
      35 |     // @ts-ignore
    > 36 |     window.analytics.load(SEGMENT_WRITE_KEY)
         |                      ^
```
but it didn't help.

src/lib/analytics/segmentOneTrustIntegration.ts
artsyjian: got it. changed conditional to match v2 (except test for [!disable.onetrust flag](https://github.com/artsy/force/blob/eb1997b04759458aedcbe2cfc0369b9c5ae6fb8c/src/v2/index.ejs#L7) which doesn't work in the integration script.)

src/lib/analytics/segmentOneTrustIntegration/setSegmentDestinationPref.ts
```diff
@@ -0,0 +1,53 @@
+export function setSegmentDestinationPref(
+  consent,
+  destinations,
+  segmentMapping
+) {
+  // map Segment destination id to OneTrust cookie category.
+  const segmentToOneTrust = {
```
artsyjian, line 7: would it be better to store this mapping in env config?

src/lib/analytics/segmentOneTrustIntegration/fetchDestinationForWriteKey.ts
```diff
@@ -0,0 +1,30 @@
...
+    // Rename creationName to id to abstract the weird data model
+    for (const destination of destinations) {
+      destination.id = destination.creationName
+      delete destination.creationName
+    }
+
+    return destinations
+  } catch (error) {
+    console.error("Error: fetching Segment destinations resulted in: ", error)
+    return []
```
artsyjian, line 28: might this be too strict? any error will cause 3rd party destinations to not be loaded, only Segment itself and Redshift will get events.

i guess we can avoid fetching api, by storing list of destinations in config, per environment.

src/lib/analytics/segmentOneTrustIntegration/segmentOneTrustIntegration.ts
```diff
@@ -0,0 +1,46 @@
...
+let ONETRUST_ENABLED = false
+
+if (!sd.THIRD_PARTIES_DISABLED && !sd.EIGEN && sd.ONETRUST_SCRIPT_ID) {
+  ONETRUST_ENABLED = true
+}
+
+export function loadSegment() {
+  if (ONETRUST_ENABLED) {
+    // OneTrust is enabled, load Segment based on OneTrust consent.
+    fetchDestinations(SEGMENT_WRITE_KEY).then(async destinations => {
```
artsyjian, line 20: the fetch happens on every page load/reload, so call volume to segment's api will increase.

src/lib/analytics/segmentOneTrustIntegration/fetchDestinationForWriteKey.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    // Rename creationName to id to abstract the weird data model
+    for (const destination of destinations) {
+      destination.id = destination.creationName
+      delete destination.creationName
+    }
+
+    return destinations
+  } catch (error) {
+    console.error("Error: fetching Segment destinations resulted in: ", error)
```
artsyjian, line 27: is it good practice to error to browser console? should this go to sentry instead?

joeyAghion (approved):
To paraphrase, the outstanding question is: since we already need to store a mapping of destinations to consent categories, maybe we should just store enough destination data that we don't need to fetch from Segment's API at all?

Does the Segment fetch affect the perceived performance or experience of the page? If so, I'd be tempted to just provide the necessary Segment data directly. It can be done via config if it's different across environments.

src/mobile/components/layout/test/layout.jest.ts
artsyjian: this was resolved by https://github.com/artsy/force/pull/8853/commits/fdc74ca09fefb92840e63b93625023b542ed022a

artsyjian:
> To paraphrase, the outstanding question is: since we already need to store a mapping of destinations to consent categories, maybe we should just store enough destination data that we don't need to fetch from Segment's API at all?

I feel the same. Our hard-coding segment-to-onetrust mapping has greatly reduced the benefit of the API fetch, as we have to update the mapping anyway when we enable a destination. I guess we can preempt the update by mapping all Segment destinations under the sun, including the ones we don't use, but that would be a large (?) list that we have to retrieve from somewhere (S3?), so still a fetch.

> Does the Segment fetch affect the perceived performance or experience of the page? If so, I'd be tempted to just provide the necessary Segment data directly. It can be done via config if it's different across environments.

I will try to test this. The fetch is done async so I imagine it does not affect page load.

artsyjian:
> > Does the Segment fetch affect the perceived performance or experience of the page? If so, I'd be tempted to just provide the necessary Segment data directly. It can be done via config if it's different across environments.
> 
> I will try to test this. The fetch is done async so I imagine it does not affect page load.

tested by adding a `delay` call:

```
export async function fetchDestinationForWriteKey(writeKey) {
  try {

    await delay(15000)

    const res = await window.fetch(
      `https://cdn.segment.com/v1/projects/${writeKey}/integrations`
    )
```

Looks like there's no impact to page load:
segment_fetch_delay

joeyAghion:
If the Segment call doesn't affect page experience I'm OK with this as-is. I _don't_ think we should try to map all possible destinations.

I was thinking that all we'd need would be an additional field (such as `creationName`) within our mapping's list of destinations. If that's all the API call is providing, and it's easy to embed the necessary data instead, it seemed like a worthwhile improvement, especially if the API call affected the page experience.

artsyjian:
> I was thinking that all we'd need would be an additional field (such as `creationName`) within our mapping's list of destinations. If that's all the API call is providing, and it's easy to embed the necessary data instead, it seemed like a worthwhile improvement, especially if the API call affected the page experience.

The API-call returns specifically those destinations that are enabled for a given write-key. The code reads `SEGMENT_WRITE_KEY` which contains Force' client-side key respective to the environment (staging, production). (There are server side keys too). The [mapping](https://github.com/artsy/force/pull/8853/files#diff-7addcf833d0c6e2f80f62bf5f670e0b230355d92f89833ad40b76935ba41b4c2R8) should ideally list all possible Segment destinations, but currently contains just the ones enabled (across Force staging and production Sources). The destination names in the list are taken from the `creationName` attribute as returned by the API (I curl'd it manually), not from the names as seen on Segment UI.

All that the API-call provides us is that we can send Segment an appropriate [destinationPref](https://github.com/artsy/force/pull/8853/files#diff-13645c895374b4c065976a0bae9cdced32e02fa14a2ad9fb55c6f9810c61c2daR28), that it contains only those destinations relevant to the write-key. I have tested adding an irrelevant destination (foobar) to `destinationPref`, and Segment seems to take it. So we can perhaps also just throw all the destinations from the mapping into `destinationPref` and not bother with the API-call, but that gives me the feeling that sooner or later we will get bitten by it.

damassi:
🎉 @artsyjian - your persistence on this def blew my mind a bit

PR #8859, opened 2021-11-10
[FX-3343] Remove clickedAppDownload helper

lidimayra:
[FX-3343]

Remove deprecated `clickedAppDownload` helper from the two components that were using it (`DownloadAppBadge` and `SellWithArtsy`), replacing it with `ClickedAppDownload` type that is [provided by cohesion](https://github.com/artsy/eigen/pull/5492/files#diff-f9f35322e1193efeb953aaf472a8d911d497795c25d0dd3ceaf6313e4ced9a8eR113).

[FX-3343]: https://artsyproduct.atlassian.net/browse/FX-3343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Consign/Routes/MarketingLanding/Components/SellWithArtsy.tsx
+++ b/src/v2/Apps/Consign/Routes/MarketingLanding/Components/SellWithArtsy.tsx
@@ -1,9 +1,15 @@
-import * as React from "react";
+import * as React from "react"
 import { Box, Button, Flex, Image, Link, Text, color } from "@artsy/palette"
 import { SectionContainer } from "./SectionContainer"
+import { useAnalyticsContext } from "v2/System"
 import { Media } from "v2/Utils/Responsive"
 import { useTracking } from "react-tracking"
-import { ContextModule, OwnerType, clickedAppDownload } from "@artsy/cohesion"
+import {
+  ActionType,
+  ContextModule,
+  OwnerType,
+  ClickedAppDownload,
+} from "@artsy/cohesion"
 
 const DOWNLOAD_URL =
   "https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080"
@@ -11,15 +17,20 @@ const DOWNLOAD_URL =
 export const SellWithArtsy: React.FC = () => {
   const tracking = useTracking()
 
+  const { contextPageOwnerId, contextPageOwnerSlug } = useAnalyticsContext()
+
+  const clickedAppDownload: ClickedAppDownload = {
+    action: ActionType.clickedAppDownload,
+    context_module: ContextModule.sellFooter,
+    context_page_owner_type: OwnerType.consign,
+    context_page_owner_slug: contextPageOwnerSlug,
+    context_page_owner_id: contextPageOwnerId,
+    destination_path: DOWNLOAD_URL,
+    subject: "Download the app",
+  }
+
   const trackDownloadAppClick = () => {
-    tracking.trackEvent(
-      clickedAppDownload({
-        context_module: ContextModule.sellFooter,
-        context_page_owner_type: OwnerType.consign,
-        destination_path: DOWNLOAD_URL,
-        subject: "Download the app",
-      })
-    )
+    tracking.trackEvent(clickedAppDownload)
   }
 
   return (
--- a/src/v2/Components/DownloadAppBadge.tsx
+++ b/src/v2/Components/DownloadAppBadge.tsx
@@ -1,6 +1,6 @@
-import * as React from "react";
+import * as React from "react"
 import track, { useTracking } from "react-tracking"
-import { clickedAppDownload, ContextModule } from "@artsy/cohesion"
+import { ActionType, ClickedAppDownload, ContextModule } from "@artsy/cohesion"
 import { useAnalyticsContext } from "v2/System"
 import Events from "v2/Utils/Events"
 import { Link, LinkProps } from "@artsy/palette"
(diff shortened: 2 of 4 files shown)


src/v2/Apps/Consign/Routes/MarketingLanding/Components/SellWithArtsy.tsx
```diff
@@ -1,25 +1,43 @@
...
   const tracking = useTracking()
 
+  const { contextPageOwnerId, contextPageOwnerSlug } = useAnalyticsContext()
+
+  const options: AnalyticsOptions = {
+    context_module: ContextModule.sellFooter,
+    destination_path: DOWNLOAD_URL,
+    context_page_owner_type: OwnerType.consign,
+    context_page_owner_slug: contextPageOwnerSlug,
+    context_page_owner_id: contextPageOwnerId,
```
lidimayra, line 36: Is this correct?
The only reason why I'm setting `context_page_owner_slug` and `context_page_owner_id` is that I saw these were already being generated when first writing the specs for this component, with the previous implementation.

Therefore, to have this available I'm setting these variables as a result of `useAnalyticsContext()` on L29
I did this because it's what we also do on [DownloadAppBadge](https://github.com/artsy/force/blob/c4e5f1cbdb805d8157ea9bdd5f8880f9841f9481/src/v2/Components/DownloadAppBadge.tsx#L46-L50) component. But I thought it would be good to confirm if this is correct (as we don't need to do this here with the current implementation)

src/v2/Components/DownloadAppBadge.tsx
```diff
@@ -18,6 +23,14 @@ const DOWNLOAD_IOS_APP_BADGE =
...
   "https://files.artsy.net/images/download-android-app.svg"
 
+interface AnalyticsOptions {
+  context_module: ContextModule
+  context_page_owner_id?: string
+  context_page_owner_slug?: string
+  destination_path: string
+  context_page_owner_type: PageOwnerType
+}
+
```
mdole, line 33: Let's avoid defining this type here and in theDownloadAppBadge file - we should be able to get what we need using the ClickedAppDownload interface exported by Cohesion.

Instead, we could use one of the following types:
`const options: Partial<ClickedAppDownload> = { ... }` 
This basically means "the type of `options` must be a subset of `ClickedAppDownload`." So `options` could have some of, or all of, the fields specified in `ClickedAppDownload`, but we couldn't add _new_ fields to it.

`const options: Omit<ClickedAppDownload, "action" | "subject"> = { ... }`
This means "we expect `options` to have all of the properties of `ClickedAppDownload` except for `action` and `submit`

src/v2/Components/DownloadAppBadge.tsx
```diff
@@ -83,3 +95,11 @@ export const DownloadAppBadge: React.FC<DownloadAppBadgeProps> = track(null, {
...
   )
 })
+
+const tracks = {
+  clickedAppDownload: (options: AnalyticsOptions): ClickedAppDownload => ({
+    action: ActionType.clickedAppDownload,
+    subject: "Download on the App Store",
+    ...options,
+  }),
+}
```
mdole, line 105: I'm tempted to say we actually don't need this `tracks` object in this case. It's a pattern we have in other places, but here, it feels like it just leads to misdirection. We're assembling an object out of two separate objects that don't really need to be separate, and that means we have to do some extra typing and type declaring.

Instead of what we have here, we could just do:

```ts
 const clickedAppDownload: ClickedAppDownload = {
      action: ActionType.clickedAppDownload,
      context_module: contextModule,
      context_page_owner_type: contextPageOwnerType!,
      context_page_owner_slug: contextPageOwnerSlug,
      context_page_owner_id: contextPageOwnerId,
      destination_path: downloadAppUrl,
      subject: "Download on the App Store",
    }

    tracking.trackEvent(clickedAppDownload)
```

That said, standards are useful to have, so I'm not sure what the right call is. What do you think @lidimayra?

src/v2/Components/DownloadAppBadge.tsx
mdole: Or see my other comment for another suggestion on typing 🤷‍♂️

mdole (changes requested):
Thanks for taking this on @lidimayra !! Left a few comments, and I also noticed a bug while reviewing this PR that I flagged to the Grow team ([Slack](https://artsy.slack.com/archives/C01ADJNCS5D/p1637148866132900)). I think we can resolve the comments that are here and merge this, and let the bug I flagged be a separate followup task.

src/v2/Components/DownloadAppBadge.tsx
lidimayra: I think this is much better!!!
I just followed the standard because I'm not that comfortable when working in the front-end, so I just copied what was done in other places... but I do think this is much simpler. I just applied the changes here and now it looks much easier to understand what's going on.

So I squashed this into 2ad327d95cc1387abb395b50200d204663b4177a (and applied similar changes to `SellWithArtsy` component and squashed it into 725bc1cda4a59e2b643dc0787bae079bac634d65)

src/v2/Components/DownloadAppBadge.tsx
lidimayra: Applied the changes suggested on https://github.com/artsy/force/pull/8859/files?diff=unified&w=0#r751185851

src/v2/Apps/Consign/Routes/MarketingLanding/Components/SellWithArtsy.tsx
mdole: I'd say either way is fine; separate branch is perhaps a little "cleaner" from a code history perspective, but it's a minor enough change that I don't feel strongly about it :)

mdole (approved):
this looks ready to me! let's go ahead and merge this one, and we can either PR or ticket a followup to deprecate the `SellWithArtsy` component.

PR #8864, opened 2021-11-11
[FX-3344] Remove clickedArticleGroup helper

lidimayra:
[FX-3344]

Replace clickedArticleGroup helper usage on ReadMore component with [ClickedArticleGroup type](https://github.com/artsy/cohesion/blob/01b821a0f3ea5e67eaaca9876533998b86ac8eee/src/Schema/Events/Click.ts#L88) that is provided by cohesion.

[FX-3344]: https://artsyproduct.atlassian.net/browse/FX-3344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Consign/Routes/MarketingLanding/Components/ReadMore.tsx
+++ b/src/v2/Apps/Consign/Routes/MarketingLanding/Components/ReadMore.tsx
@@ -1,23 +1,30 @@
-import * as React from "react";
+import * as React from "react"
 import { Box, Column, GridColumns, Image, Text } from "@artsy/palette"
 import { RouterLink } from "v2/System/Router/RouterLink"
 import { useTracking } from "react-tracking"
-import { ContextModule, OwnerType, clickedArticleGroup } from "@artsy/cohesion"
+import {
+  ActionType,
+  ClickedArticleGroup,
+  ContextModule,
+  OwnerType,
+} from "@artsy/cohesion"
 import { cropped } from "v2/Utils/resized"
 
 export const ReadMore: React.FC = () => {
   const tracking = useTracking()
 
   const trackClick = ({ slug, internalID }) => {
-    tracking.trackEvent(
-      clickedArticleGroup({
-        context_module: ContextModule.relatedArticles,
-        context_page_owner_type: OwnerType.consign,
-        destination_page_owner_id: internalID,
-        destination_page_owner_slug: slug,
-        destination_page_owner_type: OwnerType.article,
-      })
-    )
+    const clickedArticleGroup: ClickedArticleGroup = {
+      action: ActionType.clickedArticleGroup,
+      context_module: ContextModule.relatedArticles,
+      context_page_owner_type: OwnerType.consign,
+      destination_page_owner_id: internalID,
+      destination_page_owner_slug: slug,
+      destination_page_owner_type: OwnerType.article,
+      type: "thumbnail",
+    }
+
+    tracking.trackEvent(clickedArticleGroup)
   }
 
   return (


src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/ReadMore.jest.tsx
```diff
@@ -33,6 +33,25 @@ describe("ReadMore", () => {
...
+          "action": "clickedArticleGroup",
+          "context_module": "relatedArticles",
+          "context_page_owner_type": "consign",
+          "destination_page_owner_id": "5e6fc79cb77f450020738595",
+          "destination_page_owner_slug": "artsy-editorial-decide-consign-art",
+          "destination_page_owner_type": "article",
+          "type": "thumbnail",
+        },
+      ]
+    `)
```
lidimayra, line 53: I'm wondering what would be the best approach here...

The previous existent tests are still passing with the new implementation:
https://github.com/artsy/force/blob/a34d0f92af576f0e062e7db6b5cfc5f6352dedcb/src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/ReadMore.jest.tsx#L55-L69

I wasn't able to replace this with the `toMatchInlineSnapShot()` version because I can't use this matcher inside a loop.

I added this one here, but it's only checking for a single element instead of all of the 6 that are present.

Which solution would be better in this scenario?
1) Checking a single element with the `.toMatchInlineSnapshot()` matcher
2) Checking the 6 elements with the old implementation
3) Some other possibility that I'm not thinking of

src/v2/Apps/Consign/Routes/MarketingLanding/Components/ReadMore.tsx
```diff
@@ -1,23 +1,37 @@
...
+import {
+  ActionType,
+  ClickedArticleGroup,
+  ContextModule,
+  OwnerType,
+  PageOwnerType,
+} from "@artsy/cohesion"
 import { cropped } from "v2/Utils/resized"
 
+interface AnalyticsOptions {
```
mdole, line 14: My suggestions/questions from https://github.com/artsy/force/pull/8859/files also apply here (using `Partial/Omit` or removing `const tracks` - let's apply the same solution in both places once we've decided how to proceed :)

src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/ReadMore.jest.tsx
mdole: good question - I think either would be OK! Regardless of approach, we're confirming that tracking is being called and we're passing parameters with the right shape, which is what's important. So feel free to go with either one.

mdole (changes requested):
Nice work! Marking as "request changes" just so that we remember to update our types once we discuss on https://github.com/artsy/force/pull/8859/files.

src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/ReadMore.jest.tsx
lidimayra: Cool!! I'm keeping the previously existent test then!!

src/v2/Apps/Consign/Routes/MarketingLanding/Components/ReadMore.tsx
lidimayra: Thanks, Matt!! Just applied this (3724941)

mdole (approved):
nice! thanks for making those tweaks. ship it!

PR #8875, opened 2021-11-11
refactor(type-checking): Update strict null check migration flag to be less ambiguous 

damassi:
I'm not sure if `PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION` is too far over the top, but since the questions around `STRICT_NULL_CHECK` and copy-pasta nature of these flags keep propagating we should just... update em.


MounirDhahri (approved):
👏
I love this! I am going to do the same in Eigen

dzucconi:
Yes please.

PR #8883, opened 2021-11-12
chore(GRO-660): Change metadata for Price Database

jacobherrington:
Updating some metadata values for this page!

https://artsyproduct.atlassian.net/browse/GRO-660


damassi (approved):
🚀

PR #8884, opened 2021-11-12
chore: adds US$ to USD currency in order app

ashleyjelks:
On BNMO works priced in US dollars, we are not showing anything to indicate if the currency is US dollars, Hong Kong dollars, Australian dollars, etc. This PR adds a `US` prefix to our US priced artwork purchases and in the BNMO checkout flow.

https://artsyproduct.atlassian.net/browse/NX-2405

cc @artsy/negotiate-devs 

Screen Shot 2021-11-12 at 2 11 28 PM

Screen Shot 2021-11-12 at 2 12 30 PM

Screen Shot 2021-11-12 at 2 13 05 PM

Screen Shot 2021-11-12 at 2 13 14 PM

Screen Shot 2021-11-12 at 2 13 26 PM

Screen Shot 2021-11-12 at 2 13 45 PM

Screen Shot 2021-11-12 at 2 14 01 PM

Screen Shot 2021-11-12 at 2 14 51 PM

Screen Shot 2021-11-12 at 2 23 19 PM

Screen Shot 2021-11-12 at 2 25 22 PM

Screen Shot 2021-11-12 at 2 25 32 PM

Screen Shot 2021-11-12 at 2 25 40 PM

Screen Shot 2021-11-12 at 2 25 51 PM

Screen Shot 2021-11-12 at 2 26 19 PM


src/v2/Apps/Order/Utils/currencyUtils.ts
```diff
@@ -0,0 +1,11 @@
+export const appendCurrencySymbol = (
+  price: string | number | null | undefined,
+  currency: string
+): string | number | null | undefined => {
+  const pricePresent: boolean =
+    typeof price === "string" || typeof price == "number"
+  if (currency === "USD" && pricePresent) {
```
araujobarret, line 7: I'm curious about the other currencies, shouldn't we support them as well?

araujobarret (dismissed):
One more question about the inputs, did we check with Design about having masks there to inform the currency too?

I know you had to touch in many test files and I assume would cost some time to upgrade those to rtl, right?

I don't see any problems here, great job 💯

src/v2/Apps/Order/Utils/currencyUtils.ts
araujobarret: nvm this comment, I missed `we are not showing anything to indicate if the currency is US dollars, Hong Kong dollars, Australian dollars, etc.`

ashleyjelks: Yes, also since this only touches BNMO works and we only support USD, EUR, and GBP this works for now as well. Although I agree we should have a better way of dealing with our currency types!

src/v2/Apps/Order/Components/OfferHistoryItem.tsx
```diff
@@ -29,14 +31,14 @@ const OfferHistoryItem: React.FC<
             : "Your offer"}
         </Text>
         <Text variant={["xs", "sm"]} color="black100">
-          {lastOffer?.amount}
+          {appendCurrencySymbol(lastOffer?.amount, currency)}
```
starsirius, line 34: To avoid duplicating logic in many places and help consolidate the display across all surface areas, it will be great the implementation is in Metaphysics and clients can just query for it. For example, we can expand the [`amount`](https://github.com/artsy/metaphysics/blob/5e9eb1795fd261ecb8f94ad76a56e94436f56870/src/schema/v1/fields/money.ts#L28-L71) helper to consider displaying the currency code.

In parallel, I think design is working on [consolidating money displays](https://artsy.slack.com/archives/C9XJKPY9W/p1635442601002500) so we have some standard to follow.

PR #8887, opened 2021-11-15
feat(FX-3543): Add filters pills

anastasiapyzhik:
[FX-3543]

This PR:

- **adds default pill with artist name above their artworks**
- **adds CloseIcon to non-default pills**
- **removes animation of pills block** as it was implemented for the block only containing create alert button and it's height was predictable. Now we are going to be able to add and remove pills hence the block's height can't be hardcoded any more and animation implementation would be more complex

#### Demo

Screenshot 2021-11-25 at 13 27 10

[FX-3543]: https://artsyproduct.atlassian.net/browse/FX-3543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -61,6 +61,7 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
         }}
         savedSearchProps={savedSearchAttributes}
         enableCreateAlert={sd.ENABLE_SAVED_SEARCH}
+        defaultValues={[artist.name]}
```
dimatretyak, line 64: **suggestion**: pass it via `savedSearchProps`

src/v2/Components/ArtworkFilter/__tests__/FiltersPills.jest.tsx
```diff
@@ -0,0 +1,46 @@
+import { mount } from "enzyme"
```
dimatretyak, line 1: **suggestion**: use `@testing-library/react` instead of `enzyme`

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -137,6 +137,7 @@ export const BaseArtworkFilter: React.FC<
     offset?: number
     savedSearchProps?: SavedSearchAttributes
     enableCreateAlert?: boolean
+    defaultValues?: (string | null)[]
```
dimatretyak, line 140: **thoughts**: what do you think about the fact that only `string[]` is always expected here?

dimatretyak (approved):

src/v2/Components/ArtworkFilter/__tests__/FiltersPills.jest.tsx
damassi: Recently updated docs: https://github.com/artsy/force/blob/master/docs/best_practices.md#write-unit-tests-for-new-components

dimatretyak (approved):

dblandin (approved):
Looks great! 💯

PR #8889, opened 2021-11-15
feat(partners): lazily load individual rails

dzucconi:
This PR further optimizes the lazily loaded rails by lazily loading each individual rail instead of in just a big block.

### DIFF
--- a/src/v2/System/Relay/SystemQueryRenderer.tsx
+++ b/src/v2/System/Relay/SystemQueryRenderer.tsx
@@ -1,8 +1,10 @@
-import * as React from "react";
+import * as React from "react"
 import { QueryRenderer } from "react-relay"
 import { OperationType } from "relay-runtime"
 import { useDidMount } from "v2/Utils/Hooks/useDidMount"
 import { useLazyLoadComponent } from "v2/Utils/Hooks/useLazyLoadComponent"
+import { useSystemContext } from "v2/System/useSystemContext"
+import createLogger from "v2/Utils/logger"
 
 type QueryRendererProps = React.ComponentProps<typeof QueryRenderer>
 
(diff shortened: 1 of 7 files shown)


src/v2/System/Relay/SystemQueryRenderer.tsx
```diff
@@ -28,19 +29,26 @@ export function SystemQueryRenderer<T extends OperationType>({
...
+
   const showPlaceholder =
     debugPlaceholder || !isMounted || (lazyLoad && !isEnteredView)
 
+  const { relayEnvironment } = useSystemContext()
+
+  const environment = _environment ?? relayEnvironment
+
   if (!environment) {
+    console.warn("Requires an environment")
```
dzucconi, line 51: This took me way too long to track down 😅

damassi: For system-level stuff can we use `v2/Utils/logger`

src/v2/System/Relay/SystemQueryRenderer.tsx
```diff
@@ -28,19 +29,26 @@ export function SystemQueryRenderer<T extends OperationType>({
...
   const { isEnteredView, Waypoint } = useLazyLoadComponent({
     threshold: lazyLoadThreshold,
   })
+
   const showPlaceholder =
     debugPlaceholder || !isMounted || (lazyLoad && !isEnteredView)
 
+  const { relayEnvironment } = useSystemContext()
+
+  const environment = _environment ?? relayEnvironment
```
damassi, line 48: Why is this now needed?

damassi (approved):
One minor request and a q, otherwise LGTM.

src/v2/System/Relay/SystemQueryRenderer.tsx
dzucconi: It's passed in two places below — could inline the `??` in both otherwise

damassi: Oh i just mean, it seems strange that `relayEnvironment` should ever be passed into `SystemQueryRenderer`, unless that code is very old. It should always be available from `<Boot>`.

dzucconi: It has always needed it (every instance of `SystemQueryRenderer` explicitly passes it in, it was just never given a default before. What seems strange about this though — it's needed in `QueryRenderer`.

PR #8890, opened 2021-11-15
feat: update mobile/legacy pages segment snippet to 4.1.0.

artsyjian:
https://artsyproduct.atlassian.net/browse/PLATFORM-3730
https://artsy.slack.com/archives/CP9P4KR35/p1637009420458000

Different versions of segment snippet are being used. `4.1.0` is used for [v2](https://github.com/artsy/force/blob/2ae273eae1708e5b635ae1a9ffc2e950b666e51b/src/v2/index.ejs#L61). `3.0.1` for mobile/legacy.

I am working on [OneTrust/Segment integration](https://artsyproduct.atlassian.net/browse/PLATFORM-3593) which [requires](https://gist.github.com/sperand-io/4725e248a35d5005d68d810d8a8f7b29#file-granulardecision-js-L1) `4.1.0` at minimum.

This PR upgrades mobile/legacy to `4.1.0`. 

I _think_ `analytics.SNIPPET_VERSION` identifies the version of the code inline in the JS tag, so really just an annotation for humans. So I just copied the code from v2 and pasted it into mobile/legacy.


damassi:
👍

PR #8896, opened 2021-11-16
feat(home): updates set id for homepage featured galleries rail

dzucconi:
Closes: [GRO-670](https://artsyproduct.atlassian.net/browse/GRO-670)


dzucconi:
Note that this set isn't synced with staging yet so it won't show up on staging until next week.

PR #8898, opened 2021-11-17
feat: Add message to the order summary after purchase

rajsam003:

This PR resolves[ CX-2139](https://artsyproduct.atlassian.net/browse/CX-2139)

### Description

This PR adds the below message with Appstore and GooglePlay badge after the order gets placed. 

image
|Android|IPhone|
|---|---|
|image|image|


lordkiz (dismissed):
LGTM.
What's with all the graphql related changes? What triggered them?

rajsam003:
@lordkiz  It was because I added `state` prop in `TransactionDetailsSummaryItem_order`

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -52,6 +54,33 @@ export class TransactionDetailsSummaryItem extends React.Component<
         <Spacer mb={2} />
         <Entry label="Total" value={this.buyerTotalDisplayAmount()} final />
         {showOfferNote && order.mode === "OFFER" && this.renderNoteEntry()}
+        {order.state === "SUBMITTED" && (
```
oxaudo, line 57: Late to this party. But wondering if that is the correct state to be looking at. The Submitted state does not necessarily guarantee you the purchase and we def have a lot of orders and even more offers that will get rejected by galleries. Logically this collection item creation should happen on order Approval I think rather than on order submission.

Please ignore if that point was discussed and decided upon already :)

rajsam003: @oxaudo Very good catch. I'm not sure how quickly the order status can change from `Submitted` to `Approved` as the gallery could also approve the order very late. But the requirement is to display this message when the user clicks `Submit` which is on the BNMO page. If we are showing it based on the approval status, then there is no point in viewing it on this page.

I will post about this in the CX channel and ask PM and designer. Thanks for noticing it 🙌

brainbicycle (approved):

brainbicycle:
Looks good to me! Thanks for pushing through!

PR #8902, opened 2021-11-17
Issue with Buyer’s Premium modal display

lilyfromseattle:
# [NX-3004]
cc @negotiate-devs

Context:
https://artsy.slack.com/archives/CCS8TMQ0K/p1636044522073500

Example: 
https://www.artsy.net/artwork/sam-francis-blue-green-lembark-l-dot-56-sf-318

Before:
Screen Shot 2021-11-17 at 1 31 17 PM

After:
Screen Shot 2021-11-17 at 1 31 24 PM

[NX-3004]: https://artsyproduct.atlassian.net/browse/NX-3004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/AuctionBuyersPremiumDialog.tsx
+++ b/src/v2/Components/AuctionBuyersPremiumDialog.tsx
@@ -9,7 +9,7 @@ import {
   Text,
 } from "@artsy/palette"
 import { compact } from "lodash"
-import * as React from "react";
+import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useSystemContext } from "v2/System"
 import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
@@ -79,12 +79,16 @@ const AuctionBuyersPremiumDialog: React.FC<AuctionBuyersPremiumDialogProps> = ({
             </Text>
           ) : (
             schedule.map((premium, i) => {
+              const percent = (premium.percent ?? 0) * 100
+              const hasDecimal = percent - Math.floor(percent) !== 0
+              const fixed = percent.toFixed(1)
+              const displayPercentage = hasDecimal ? fixed : percent
               // First
               if (i === 0) {
                 return (
                   <Text key={i} variant="sm" mb={0.5}>
                     On the hammer price up to and including{" "}
-                    {schedule[i + 1]?.amount}: {(premium.percent ?? 0) * 100}%
+                    {schedule[i + 1]?.amount}: {displayPercentage}%
                   </Text>
                 )
               }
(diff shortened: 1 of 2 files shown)


src/v2/Components/AuctionBuyersPremiumDialog.tsx
```diff
@@ -79,12 +79,16 @@ const AuctionBuyersPremiumDialog: React.FC<AuctionBuyersPremiumDialogProps> = ({
             </Text>
           ) : (
             schedule.map((premium, i) => {
+              const percent = (premium.percent ?? 0) * 100
```
araujobarret, line 82: non-blocking: if it's something we use shared across other components maybe putting in some utils folder

src/v2/Components/__tests__/AuctionBuyersPremiumDialog.jest.tsx
```diff
@@ -76,5 +76,19 @@ describe("AuctionBuyersPremiumDialog", () => {
         "On the portion of the hammer price in excess of $2,500,000: 12%"
       )
     })
+
+    describe("with a percentage that isn't a whole number", () => {
```
araujobarret, line 80: I would say to avoid adding more enzyme tests ([more info](https://github.com/artsy/force/blob/bb6cc363dfd11c71967e7b8767fe39d168025ffb/docs/best_practices.md#write-unit-tests-for-new-components)), we recommend using react testing library instead of adding more tests with the old approach.
[example](https://github.com/artsy/force/blob/bb6cc363dfd11c71967e7b8767fe39d168025ffb/src/v2/Apps/Settings/Routes/Saves/Components/__tests__/SettingsSavesArtists.jest.tsx)

damassi: Here's the latest best practices docs that cover the change: https://github.com/artsy/force/blob/master/docs/best_practices.md#write-unit-tests-for-new-components

lilyfromseattle: Going forward, should I be converting any test files I come across?

src/v2/Components/AuctionBuyersPremiumDialog.tsx
lilyfromseattle: So far its only used here

src/v2/Components/__tests__/AuctionBuyersPremiumDialog.jest.tsx
araujobarret: if it's not too complicated and the time is ok I would say yes, otherwise just to not increase the number of tests with enzyme use RTL

ashleyjelks (approved):

PR #8904, opened 2021-11-18
fix: NX-3020 remove text

patrinoua:
Resolves NX-3020 

Removes "Please note that all offers are binding."


tam-kis (approved):
Looks good, you can merge on green (when all tests have passed)

PR #8906, opened 2021-11-18
feat(GRO- 595): /settings2/saves

dzucconi:
This implements the top half of the page.

Remaining are followed categories and partners. Will switch to the Metaphysics work for both before continuing.

https://user-images.githubusercontent.com/112297/142468998-99117981-5ced-4b5e-b283-1616156ffed5.mov

### DIFF
--- a/src/v2/Apps/Settings/Routes/Saves/Components/SettingsSavesArtists.tsx
+++ b/src/v2/Apps/Settings/Routes/Saves/Components/SettingsSavesArtists.tsx
@@ -1,0 +1,165 @@
+import { FC, Fragment, useState } from "react"
+import {
+  createPaginationContainer,
+  graphql,
+  RelayPaginationProp,
+} from "react-relay"
+import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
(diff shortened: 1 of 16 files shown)


src/v2/Apps/Settings/Routes/Saves/Components/SettingsSavesArtists.tsx
```diff
@@ -0,0 +1,165 @@
...
+      return props.me.followsAndSaves?.artistsConnection
+    },
+    getFragmentVariables(prevVars, totalCount) {
+      return { ...prevVars, totalCount }
+    },
+    getVariables(_, { cursor: after }, fragmentVariables) {
+      return { ...fragmentVariables, after }
+    },
+    query: SETTINGS_SAVES_ARTISTS_QUERY,
+  }
```
damassi, line 127: This can be simplified. (There's a TLDR at the end of this post: https://artsy.github.io/blog/2020/08/31/relay-pagination-containers-demystified)

dzucconi: Good catch, I can remove the `getConnectionFromProps`. However, Relay is unable to synthesize a `getFragmentVariables` though in this case. Wondering why

PR #8908, opened 2021-11-18
fix(settings): restores old form; reorganization

dzucconi:
So: I've put back the old form (which I'm assuming is broken; but at least is broken in whatever the way it's always been).

For the new form I reorganized everything into the correct location; renamed things and started to refactor it to make more sense. I need to dig into some _fundamental_ issues with it however.

But you should be able to slot in the new work. I've included TODOs with the correct naming. Will point them out below.


src/desktop/components/react/stitch_components/index.tsx
```diff
@@ -19,7 +19,7 @@ export {
   TwoFactorAuthenticationQueryRenderer as TwoFactorAuthentication,
 } from "v2/Components/UserSettings/TwoFactorAuthentication"
 export { UserEmailPreferencesQueryRenderer as UserEmailPreferences, } from "v2/Components/UserSettings/UserEmailPreferences"
-export { UserInformationQueryRenderer as UserInformation } from "v2/Apps/Settings/Routes/EditProfile/UserInformation"
+export { UserInformationQueryRenderer as UserInformation } from "v2/Components/UserSettings/UserInformation"
```
dzucconi, line 22: This goes back to what it was before. For any future work let's just leave these code alone and consider it dead. Copy things rather than move them.

src/v2/Apps/Settings/Routes/EditProfile/SettingsEditProfileRoute.tsx
```diff
@@ -1,7 +1,6 @@
 import React from "react"
 import { SettingsEditProfileRoute_me } from "v2/__generated__/SettingsEditProfileRoute_me.graphql"
 import { createFragmentContainer, graphql } from "react-relay"
-import { UserInformationRefetchContainer } from "./UserInformation"
```
dzucconi, line 4: This had no business being a refetch container. I think someone just misunderstood that you can just fetch attrs in the mutation? I'm not sure.

src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
```diff
@@ -170,70 +164,40 @@ export const UserInformation: React.FC<UserInformationProps> = ({
                   />
                 </>
               )}
-              {status && !status.success && (
+
+              {/* TODO: Remove with toast usage */}
+              {status && status.error && (
```
dzucconi, line 169: The reason the red box was showing is that it was asserting on an attr that was never there to begin with. This is also a weird way to handle it in the first place. Will remove all of this in a refactor.

src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsRoute.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+          {/* TODO: SettingsEditSettingsPassword */}
+
+          {/* TODO: SettingsEditSettingsTwoFactor */}
+
+          {/* TODO: SettingsEditSettingsLinkedAccounts */}
+
+          {/* TODO: SettingsEditSettingsEmailPreferences */}
+
+          {/* TODO: SettingsEditSettingsDeleteAccount */}
```
dzucconi, line 26: So here you can see the components left on this page. We should follow this naming convention which prevents collisions. Relay related code needs to have globally unique names.

src/v2/Apps/Settings/settingsRoutes.tsx
```diff
@@ -68,10 +68,10 @@ const SavesRoute = loadable(
       component.SettingsSavesRouteFragmentContainer,
   }
 )
-const SettingsRoute = loadable(
+const EditSettingsRoute = loadable(
   () =>
     import(
-      /* webpackChunkName: "settingsBundle" */ "./Routes/Settings/SettingsEditRoute"
+      /* webpackChunkName: "settingsBundle" */ "./Routes/EditSettings/SettingsEditSettingsRoute"
```
dzucconi, line 74: This was infact unique (but poorly named) for whatever reason this route was literally importing and rendering another route inside of it.

src/v2/Components/UserSettings/UserInformation/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+// @ts-expect-error STRICT_NULL_CHECK
+type GravityFieldErrors = UpdateUserInformationMutationResponse["updateMyUserProfile"]["userOrError"]["mutationError"]
+
+const formatGravityErrors = ({ fieldErrors }: GravityFieldErrors) => {
+  const formatted = {}
+  fieldErrors.map(err => {
+    formatted[err["name"]] = err.message.split(", ")
+  })
+  return formatted
+}
```
dzucconi, line 232: Old code.

The-Beez-Kneez:
So the understanding being we are duplicating the code instead of moving it?

src/v2/Components/UserSettings/UserInformation/index.tsx
dzucconi: This will also fix Integrity for the time being. We'll have to update all the specs once the new page is launched, of course.

dzucconi:
@The-Beez-Kneez 

Yeah, just duplicate. If we were to move and keep them in two places at once it limits how we can refactor. (Toasts won't work, for instance)

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

> `src/v2/Components/UserSettings/UserInformation/index.tsx`
  </td>
    </tr>
  </tbody>
</table>




<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 305310cf6fe095f34cc3c4c71855e4e91294ace9
</p>

dzucconi:
Going to ignore the `check-pr` step which isn't relevant here. Will merge on the rest passing.

PR #8912, opened 2021-11-19
fix(settings): submits the password with mutation; improves validation and display

dzucconi:
So this PR primarily fixes the issue where we weren't sending the password through with the mutation and corrects the logic surrounding how it's displayed. 

So here you can see: as soon as you change the email the password input displays itself. If you type the correct password, it updates the email. If you type the wrong password, it prevents you from updating the email and errors.

https://user-images.githubusercontent.com/112297/142661528-4007db15-c7b6-4282-bc9f-42ec1e4ae922.mov

I'll leave a few comments below explaining what was going on and why.

### DIFF
--- a/src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
+++ b/src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
@@ -1,18 +1,18 @@
 import * as React from "react"
+import * as Yup from "yup"
 import { createFragmentContainer, graphql } from "react-relay"
-import { Form, Formik, FormikProps } from "formik"
+import { Form, Formik } from "formik"
 import { SettingsEditSettingsInformation_me } from "v2/__generated__/SettingsEditSettingsInformation_me.graphql"
 import {
   Button,
   Text,
-  Banner,
   Input,
   Join,
   Spacer,
   useToasts,
   PasswordInput,
 } from "@artsy/palette"
-import { ChangeUserInformationValidator } from "v2/Components/Authentication/Validators"
+import { email, password, name } from "v2/Components/Authentication/Validators"
 import { useUpdateSettingsInformation } from "./useUpdateSettingsInformation"
 
 interface SettingsEditSettingsInformationProps {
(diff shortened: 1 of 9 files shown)


src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
```diff
@@ -25,149 +25,115 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
...
+          password: "",
+        }}
+        validationSchema={Yup.object().shape({
+          email,
+          name,
+          // Requires password when email is changed
+          password: password.when("email", {
+            is: email => email !== me.email,
+            otherwise: field => field.notRequired(),
+          }),
```
dzucconi, line 50: I added this conditional password validation — if the email isn't the email initially returned from the server, then the password is required. If _somehow_ a user is able to submit the form before the password field shows itself, this will prevent submission.

src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
```diff
@@ -25,149 +25,115 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
...
+                <Input
+                  name="paddleNumber"
+                  title="Bidder Number"
+                  value={"me.paddleNumber"}
+                  readOnly
+                  disabled
+                />
               )}
 
-              {touched.email && values.email !== me.email && (
```
dzucconi, line 150: So you can see this was asking for `touched.email`... initially this was puzzling: apparently `touched` won't update until a blur event. So despite validation and values updating on change, you wouldn't see see this form until the blur. Simple solution: we don't need to check `touched` of course. If the value is different it's touched!

But below you can see what was happening that was letting submissions go through. If you just hit "enter" on the keyboard, it'd immediately trigger submission which would run before the input got onto the page. There was no actual validation aside from the required prop.

https://user-images.githubusercontent.com/112297/142662310-e58d9bdb-c8d0-498b-b8f2-f2a3f9bc928b.mov

src/v2/Apps/Settings/Routes/EditSettings/useUpdateSettingsInformation.ts
```diff
@@ -28,6 +28,9 @@ export const useUpdateSettingsInformation = () => {
             $input: UpdateMyProfileInput!
           ) {
             updateMyUserProfile(input: $input) {
+              me {
+                ...SettingsEditSettingsInformation_me
+              }
```
dzucconi, line 33: This updates the `me` object after mutation and essentially resets the field, re-hiding the password input. Previously this would also break repeated form submissions because if you try to submit the _correct_ password _without_ an email change, Gravity will error 😅

src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
```diff
@@ -25,149 +25,115 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
...
-                    name="paddleNumber"
-                    title="Bidder number"
-                    value={me.paddleNumber}
-                    readOnly
-                  />
-                </>
+                <Input
+                  name="paddleNumber"
+                  title="Bidder Number"
+                  value={"me.paddleNumber"}
```
The-Beez-Kneez, line 119: Actually wanted to ask about this section. In what instances is the `paddleNumber` present? No objections, just curious as to the purpose.

src/v2/Components/Authentication/Validators.tsx
```diff
@@ -4,9 +4,9 @@ export const email = Yup.string()
   .email("Please enter a valid email.")
   .required("Please enter a valid email.")
 
-const name = Yup.string().required("Name is required.")
+export const name = Yup.string().required("Name is required.")
 
-const password = Yup.string()
+export const password = Yup.string()
```
The-Beez-Kneez, line 9: What is `Yup`?

src/v2/Apps/Settings/Routes/EditSettings/SettingsEditSettingsInformation.tsx
dzucconi: Oh, woops your comment realized I also left this in a string (was just testing how it displayed).

You get a paddle number when you register for an auction.

The-Beez-Kneez (approved):
Jus some questions for clarification, but otherwise lgtm.

src/v2/Components/Authentication/Validators.tsx
dzucconi: It's a little library that let's you express a validation schema: https://github.com/jquense/yup

It plugs into Formik so that when inputs don't match the schema they get passed down as errors associated with the fields.

pepopowitz:
Nice cleanup 💪 

Thanks for doing this.

PR #8916, opened 2021-11-22
feat(SWA-132): Add recaptcha for the submission creation

Serge0n:
Jira Ticket:ticket: : [SWA-132](https://artsyproduct.atlassian.net/browse/SWA-132)

This PR adds `Recaptcha` usage to the `Contact Information` form submitting.

❓:
- Is it enough? Or we should use backend validation for our flow? @damassi, @mzikherman could you help, please? I noticed that in Force only `SignUpForm` has backend validation.
- Are those Recaptcha call's doing something? 
https://github.com/artsy/force/blob/0d76e5e579925c2027711ff725450001a6fc0882/src/v2/Components/Authentication/Views/ForgotPasswordForm.tsx#L23
https://github.com/artsy/force/blob/0d76e5e579925c2027711ff725450001a6fc0882/src/v2/Components/Authentication/Views/LoginForm.tsx#L46

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
@@ -15,6 +15,7 @@ import { contactInformationValidationSchema } from "../Utils/validation"
 import { BackLink } from "v2/Components/Links/BackLink"
 import { useErrorModal } from "../Utils/useErrorModal"
 import { data as sd } from "sharify"
+import { recaptcha, RecaptchaAction } from "v2/Utils/recaptcha"
 
 const getContactInformationFormInitialValues = (me: ContactInformation_me) => ({
   name: me?.name || "",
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
```diff
@@ -45,11 +46,16 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
     removeSubmission,
   } = useSubmission(id)
 
+  const handleRecaptcha = (action: RecaptchaAction) =>
+    new Promise(resolve => recaptcha(action, resolve))
```
iskounen, line 50: **question:** How do we know whether a user passed or failed the reCAPTCHA test?

kiryl-zubarau: recaptcha(action, **resolve**))

`resolve` function will receive the recaptcha token that as far as I understood indicates that the user passed.

kiryl-zubarau (approved):
🚀

joeyAghion:
From what I recall, the recaptcha version in use by Force is the "invisible recaptcha," which only generates a token representing the interaction. If no back-end validation of the token is done, aggregate data about the suspicion level (`0`-`1`) of these actions will be available in our dashboard, but no suspicious actions will actually be blocked.

iskounen:
> aggregate data about the suspicion level (`0`-`1`) of these actions will be available in our dashboard

@joeyAghion thanks for the context. Do you know how these dashboards are used?

joeyAghion:
The dashboard is for informational purposes only. E.g., if we experience abuse of inquiries, we can review that data and consider whether to enforce a check and at what threshold. We can also consider changes of the enforced thresholds for actions that are already enforced.

(This investigation made me realize that we recently _lost_ a bunch of our previous tracking around inquiries. I'll follow up w/the team working on that.)

PR #8918, opened 2021-11-22
[WIP] Convert shipping form to functional component

lean8086:
**[DO NOT MERGE YET]: Some use cases are being not covered after the latest commit**

As part of the [consolidation of the address forms into one component](https://artsyproduct.atlassian.net/browse/NX-2993), we need to convert the parent components (only Shipping and PaymentPicker) to functional components.

Ticket: [nx 2993/](https://artsyproduct.atlassian.net/browse/NX-3025)

Co-authored-by: @rquartararo @laurabeth 

Things to keep in mind when reviewing this PR:
- `state()` is not available anymore on a test level, so we are now testing the UI behaves as expected.
- when tracking user's selection of delivery method (ship, pickup) we were sending `flow: "buy now"` and `type: "button"` which is not part of our schema. Now we are using `flow: Schema.Flow.BuyNow, // Equals to "Buy now"` and `type: Schema.Type.Button, // Equals to "Button"` instead.
- `@track` class decorator was replaced for `useTracking` hook

### DIFF
--- a/src/v2/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/v2/Apps/Order/Routes/Shipping/index.tsx
@@ -36,8 +36,8 @@ import {
   validateAddress,
   validatePhoneNumber,
 } from "v2/Apps/Order/Utils/formValidators"
-import { track } from "v2/System/Analytics"
-import * as Schema from "v2/System/Analytics/Schema"
+import { useTracking } from "react-tracking"
+import { AnalyticsSchema } from "v2/System"
 import {
   Address,
   AddressChangeHandler,
@@ -94,25 +94,8 @@ export interface ShippingProps extends SystemContextProps {
   isCommittingMutation: boolean
 }
 
-export interface ShippingState {
-  shippingOption: CommerceOrderFulfillmentTypeEnum
-  address: Address
-  phoneNumber: PhoneNumber
-  phoneNumberError: PhoneNumberError
-  phoneNumberTouched: PhoneNumberTouched
-  addressErrors: AddressErrors | {}
-  addressTouched: AddressTouched
-  selectedAddressID: string
-  saveAddress: boolean
-  shippingQuotes: ShippingQuotesType
-  shippingQuoteId?: string
-  savedAddressID?: string
-}
-
 const logger = createLogger("Order/Routes/Shipping/index.tsx")
 
-track()
-// export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 export const ShippingRoute: React.FC<ShippingProps> = props => {
   const [shippingOption, setShippingOption] = useState<
     CommerceOrderFulfillmentTypeEnum
@@ -143,6 +126,9 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
   )
   const [saveAddressToggle, setSaveAddressToggle] = useState<boolean>(true)
 
+  const { trackEvent } = useTracking()
+
+  // TODO: use Formik instead?
   const touchedAddress = () => {
     return {
       name: true,
@@ -282,15 +268,12 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
   }
 
   const selectShippingQuote = async () => {
-    // const { shippingQuoteId } = state
-    const { order } = props
-
-    if (shippingQuoteId && order.internalID) {
+    if (shippingQuoteId && props.order.internalID) {
       try {
         const orderOrError = (
           await selectShippingOption(props.commitMutation, {
             input: {
-              id: order.internalID,
+              id: props.order.internalID,
               selectedShippingQuoteId: shippingQuoteId,
             },
           })
@@ -314,25 +297,15 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
   }
 
   const saveAddress = async () => {
-    // const {
-    //   address,
-    //   phoneNumber,
-    //   shippingOption,
-    //   saveAddress,
-    //   savedAddressID,
-    // } = state
-    const { relayEnvironment } = props
-    // const isCreateNewAddress = isCreateNewAddress()
-
     if (saveAddressToggle) {
       if (
         shippingOption === "SHIP" &&
         isCreateNewAddress() &&
-        relayEnvironment
+        props.relayEnvironment
       ) {
         if (savedAddressID) {
           updateUserAddress(
-            relayEnvironment,
+            props.relayEnvironment,
             savedAddressID,
             {
               ...address,
@@ -348,7 +321,7 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
           )
         } else {
           await createUserAddress(
-            relayEnvironment,
+            props.relayEnvironment,
             {
               ...address,
               phoneNumber: phoneNumber,
@@ -370,7 +343,7 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
       }
     } else if (savedAddressID) {
       deleteUserAddress(
-        relayEnvironment!,
+        props.relayEnvironment!,
         savedAddressID,
         () => {},
         message => {
@@ -442,16 +415,6 @@ export const ShippingRoute: React.FC<ShippingProps> = props => {
     setShippingQuoteId(undefined)
   }
 
-  track((props, state, args) => ({
-    action_type: Schema.ActionType.Click,
-    subject:
-      args[0] === "SHIP"
-        ? Schema.Subject.BNMOProvideShipping
-        : Schema.Subject.BNMOArrangePickup,
-    flow: "buy now",
-    type: "button",
-  }))
-
   useEffect(() => {
     const addressList = getAddressList()
     if (addressList && addressList.length > 0 && isArtaShipping()) {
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -146,211 +141,180 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+  useEffect(() => {
+    if (
+      isArtaShipping() &&
+      selectedAddressID &&
+      !shippingQuotes &&
+      !shippingQuoteId
+    ) {
+      selectShipping()
+    }
+  }, [shippingOption, shippingQuotes, shippingQuoteId])
```
pepopowitz, line 191:
```suggestion
  }, [])
```

I'm not totally sure about when this effect should run, but in the old version it ran on `componentDidMount`. To replicate that behavior, we should pass an empty array to this dependency array, which would prompt the effect to only run once, when the component mounts.

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -146,211 +141,180 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+          setPhoneNumberError(error!)
+          setPhoneNumberTouched(true)
           return
         } else if (hasErrors) {
-          this.setState({
-            addressErrors: errors!,
-            addressTouched: this.touchedAddress,
-          })
+          setAddressErrors(errors!)
+          setAddressTouched(touchedAddress())
```
pepopowitz, line 215: This seems like another good clue that there are some properties that are related & could live within the same state object.

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -146,211 +141,180 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
-      saveAddress,
-      savedAddressID,
-    } = this.state
-    const { relayEnvironment } = this.props
-    const isCreateNewAddress = this.isCreateNewAddress()
-
-    if (saveAddress) {
-      if (shippingOption === "SHIP" && isCreateNewAddress && relayEnvironment) {
+  const saveAddress = async () => {
+    if (saveAddressToggle) {
```
pepopowitz, line 309: Good rename here, I was confused reading the old code 👍

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -146,211 +141,180 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+      if (
+        shippingOption === "SHIP" &&
+        isCreateNewAddress() &&
+        props.relayEnvironment
```
pepopowitz, line 313: Is the removal of the destructuring of `relayEnvironment` from `props` intentional? I don't personally care either way but just noting that we do end up with `props.` sprinkled into this function multiple times to access `relayEnvironment`.

src/v2/Apps/Order/Routes/Shipping/index.tsx
pepopowitz: Actually I think I was wrong -- I just found https://github.com/artsy/force/pull/8918/files#diff-3450e2d03077be9b8efde9cbfc7e452910188d0d80ad0cd4848b242d63840b60L484 and now I see why we were firing the effect when any of those change.

Side note....this is a perfect example of how `useEffect` can help keep all these related things in one place instead of two

pepopowitz (commented):
This is awesome, thank you for doing this work!!!

The bit of feedback I'm most interested in is about grouping logical state properties. There were several times through reading (noted in multiple comments) that it looked like related properties were changing at the same time but with different calls to `useState`. This makes me think there is some grouping we can and should do. 

props to @annacarey for the co-review

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -94,46 +94,41 @@ export interface ShippingProps extends SystemContextProps {
...
+  >(false)
+  const [selectedAddressID, setSelectedAddressID] = useState<string>(
+    defaultShippingAddressIndex(props.me, props.order)
+  )
+  const [shippingQuotes, setShippingQuotes] = useState<ShippingQuotesType>(
+    getShippingQuotes(props.order)
+  )
+  const [shippingQuoteId, setShippingQuoteId] = useState<string | undefined>(
+    getSelectedShippingQuoteId(props.order)
+  )
```
damassi, line 126: If we're spending time moving this to `useState`, why not just use formik? It does all of this for us for free.

src/v2/Apps/Order/Routes/Shipping/index.tsx
araujobarret, line 191: the hook here is missing one dependency, `isArtaShipping`.

if the function signature changes very often, you can wrap this function in a useCallback

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -94,46 +94,41 @@ export interface ShippingProps extends SystemContextProps {
...
-      this.props.me,
-      this.props.order
-    ),
-    shippingQuotes: getShippingQuotes(this.props.order),
-    shippingQuoteId: getSelectedShippingQuoteId(this.props.order),
-    saveAddress: true,
-  }
 
-  get touchedAddress() {
+export const ShippingRoute: React.FC<ShippingProps> = props => {
```
araujobarret, line 99: nit: you can destructure the props here and this would save some verbosity along the component of calling `props.something`.

src/v2/Apps/Order/Routes/Shipping/index.tsx
lean8086: Good call. Looking into this I realized there was also a [remaining destructuring from the older render() function](https://github.com/artsy/force/pull/8918/files#diff-3450e2d03077be9b8efde9cbfc7e452910188d0d80ad0cd4848b242d63840b60R511).
I'm destructuring all props together now 👌

src/v2/Apps/Order/Routes/Shipping/index.tsx
lean8086: props to @laurabeth

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -146,211 +141,180 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
       // save address when user is entering new address AND save checkbox is selected
-      await this.saveAddress()
+      await saveAddress()
 
-      if (isArtaShipping) {
-        this.setState({
-          shippingQuotes: getShippingQuotes(orderOrError?.order),
-        })
+      if (isArtaShipping()) {
+        console.log({ shippingQuotes: getShippingQuotes(orderOrError?.order) })
```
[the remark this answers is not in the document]
lean8086: Oops

src/v2/Apps/Order/Routes/Shipping/index.tsx
lean8086: Glad you mentioned this! The purpose of moving this to a FC is so we can consolidate all our address forms using formik. There is a [work-in-progress ticket](https://artsyproduct.atlassian.net/browse/NX-2993) that is already using formik on all of the others FC address forms, except for Shipping and PaymentPicker which need to be converted.
So, indeed, we have an immediate follow-up task to use formik and remove most of the code that is validating address "manually".

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -94,46 +94,41 @@ export interface ShippingProps extends SystemContextProps {
...
+  const [selectedAddressID, setSelectedAddressID] = useState<string>(
+    defaultShippingAddressIndex(props.me, props.order)
+  )
+  const [shippingQuotes, setShippingQuotes] = useState<ShippingQuotesType>(
+    getShippingQuotes(props.order)
+  )
+  const [shippingQuoteId, setShippingQuoteId] = useState<string | undefined>(
+    getSelectedShippingQuoteId(props.order)
+  )
+  const [saveAddressToggle, setSaveAddressToggle] = useState<boolean>(true)
```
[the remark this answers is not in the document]
lean8086: Agreed. There is a [follow-up task](https://artsyproduct.atlassian.net/browse/NX-2993) to replace the address form and its validations with formik so we can have only one centralized address form. I'll check if grouping some `useState` can interfere with that.

src/v2/Apps/Order/Routes/Shipping/index.tsx
lean8086: Exactly, this is grouping not only `componentDidMount` but all the callbacks defined on `this.setState()` of the class component.

src/v2/Apps/Order/Routes/Shipping/index.tsx
damassi: Sweet 🙏

damassi:
@lean8086 - can this PR be closed, or is it still being worked on?

lean8086:
> @lean8086 - can this PR be closed, or is it still being worked on?

It's technically in progress but it's currently down in priority. Would it make sense to convert it into draft @damassi?

damassi:
Yup, draft would be better, assuming it preserves all of the comment history and so on! It's nbd keeping it open as is, just doing a little janitorial work.

PR #8927, opened 2021-11-22
feature: Upload Progress

icirellik:
Add the progress status and time taken to the upload to S3 task to
enabled a better understanding of which file take the longest.


ovasdi (approved):
LGTM

PR #8930, opened 2021-11-23
[PLATFORM-3280] Copy/Download recovery codes

lidimayra:
[PLATFORM-3280]

- [x] Add a `copy` button to the modal that is displayed by the time a user enables 2FA
- [x] Add a `copy` button to the modal that is displayed by the time a user clicks on `show` or `regenerate` backup codes
- [x] Add a `download` button to the modal that is displayed by the time a user enables 2FA
- [x] Add a `download` button to the modal that is displayed by the time a user clicks on `show` or `regenerate` backup codes
- [x] Refactor

Screenshot 2021-11-23 at 18 23 07

[PLATFORM-3280]: https://artsyproduct.atlassian.net/browse/PLATFORM-3280?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorActions.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorActions.tsx
@@ -1,0 +1,58 @@
+import { Button, BorderBoxProps, Flex } from "@artsy/palette"
+import React, { useEffect, useState } from "react"
+
+interface BackupSecondFactorActionsProps extends BorderBoxProps {
+  backupSecondFactors: string[]
+}
+
+export const BackupSecondFactorActions: React.FC<BackupSecondFactorActionsProps> = props => {
+  const { backupSecondFactors } = props
+
+  const [supportsClipboard, setSupportsClipboard] = useState(false)
+
+  useEffect(() => {
+    // Only render the copy button if browser supports the Clipboard API
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API
+    if ("clipboard" in navigator) setSupportsClipboard(true)
+  }, [])
+
+  function copyCodesToClipboard() {
+    navigator.clipboard.writeText(backupSecondFactors.join("\n"))
+  }
+
+  function downloadCodes() {
+    const codes = backupSecondFactors.join("\n")
+    const element = document.createElement("a");
+    const file = new Blob([codes], {type: 'text/plain'});
+    element.href = URL.createObjectURL(file);
+    element.download = "recovery_codes.txt";
+    document.body.appendChild(element); // Required for this to work in FireFox
+    element.click();
+  }
+
+  return (
+    <Flex justifyContent="center">
+      {supportsClipboard && (
+        <Button
+          onClick={copyCodesToClipboard}
+          variant="secondaryOutline"
+          size="small"
+          m={1}
+          data-test="copyButton"
+        >
+          Copy
+        </Button>
+      )}
+
+      <Button
+        onClick={downloadCodes}
+        variant="secondaryOutline"
+        size="small"
+        m={1}
+        data-test="downloadButton"
+      >
+        Download
+      </Button>
+    </Flex>
+  )
+}
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorActions.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorActions.tsx
@@ -22,12 +22,13 @@ export const BackupSecondFactorActions: React.FC<BackupSecondFactorActionsProps>
 
   function downloadCodes() {
     const codes = backupSecondFactors.join("\n")
-    const element = document.createElement("a");
-    const file = new Blob([codes], {type: 'text/plain'});
-    element.href = URL.createObjectURL(file);
-    element.download = "recovery_codes.txt";
-    document.body.appendChild(element); // Required for this to work in FireFox
-    element.click();
+    const element = document.createElement("a")
+    const file = new Blob([codes], { type: "text/plain" })
+    element.href = URL.createObjectURL(file)
+    element.download = "recovery_codes.txt"
+    document.body.appendChild(element) // Required for this to work in FireFox
+    element.click()
+    URL.revokeObjectURL(element.href)
   }
 
   return (
--- a/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorModalContent.tsx
+++ b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorModalContent.tsx
@@ -29,6 +29,16 @@ export const BackupSecondFactorModalContent: React.FC<BackupSecondFactorModalCon
     navigator.clipboard.writeText(codes)
   }
 
+  function downloadCodes() {
+    const codes = me.backupSecondFactors!.map(item => item!.code).join("\n")
+    const element = document.createElement("a")
+    const file = new Blob([codes], { type: "text/plain" })
+    element.href = URL.createObjectURL(file)
+    element.download = "recovery_codes.txt"
+    document.body.appendChild(element) // Required for this to work in FireFox
+    element.click()
+  }
+
   return (
     <Box minHeight="280px">
       <Sans size="3" color="black60">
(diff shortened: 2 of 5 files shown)


src/v2/Components/UserSettings/TwoFactorAuthentication/__tests__/TwoFactorAuthentication.jest.tsx
```diff
@@ -197,3 +199,47 @@ describe("TwoFactorAuthentication", () => {
...
+
+    it("enables user to copy the recovery codes", () => {
+      const copyButtonProps = getCopyButton().props()
+
+      if (copyButtonProps.onClick) copyButtonProps.onClick({} as any);
+
+      expect(navigator.clipboard.writeText).toBeCalledTimes(1)
+    })
+  })
+})
```
lidimayra, line 245: @mc-jones I added these tests to cover the newly added behavior and I have it on `TwoFactorAuthentication` tests at this point as we discussed yesterday. However, these are not being benefited from the setup that is required for the other tests that are present in this same file.

At this point, they feel much more like unit tests for me, and I would be inclined to actually have an exclusive file for these on `src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactorReminder/__tests__/index.jest.tsx` (following similar pattern that exists today for `BackupSecondFactor` component [here](https://github.com/artsy/force/blob/master/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/__tests__/index.jest.tsx))

What do you think?

mc-jones: @lidimayra - Yeah, I definitely think that is correct. 👍

lidimayra: Thanks for the response, @mc-jones 🙌 

I ended up extracting this whole behavior (and consequently the tests as well) to a new component (1b1034a) so we can reuse it on both modals.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/__tests__/BackupSecondFactorActions.jest.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    it("displays a download button", () => {
+      expect(downloadButton).toHaveLength(1)
+      expect(downloadButton.text()).toBe("Download")
+    })
+
+    it("enables user to download the recovery codes", () => {
+      global.URL.createObjectURL = jest.fn()
+
+      // @ts-ignore: Type
```
lidimayra, line 59: I wonder if there would be a better approach to test this respecting typing...

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorModalContent.tsx
```diff
@@ -33,6 +34,9 @@ export const BackupSecondFactorModalContent: React.FC<BackupSecondFactorModalCon
           </Box>
         ))}
       </Flex>
+
+      <BackupSecondFactorActions
+        backupSecondFactors={me.backupSecondFactors!.map(item => item!.code!.toString())} />
```
lidimayra, line 39: I guess it's safe to assume the presence of these elements, right?

mc-jones:
Still need to take a look at this, but adding @dzucconi to reviewers, as he may have some views from the web side of things!

dzucconi:
We'll have to migrate it to v3 — @The-Beez-Kneez is working on this component right now, actually, so there may be some conflicts.

lidimayra:
Thanks for the update!!

I rebased this branch with the latest master. I'll keep an eye open for conflicts from now on and will reach out to @The-Beez-Kneez in case of questions.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorModalContent.tsx
mc-jones: Upon initial mutation, we would handle with the APIErrorModal [here](https://github.com/artsy/force/blob/master/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/SmsSecondFactor/Modal.tsx#L77) and so not reach the Recovery Code modal without the backup codes having been created. 

Took a quick look but it is less clear when the user is just checking their already existing backup codes. Would be good to verify and handle gracefully if we aren't already.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorActions.tsx
```diff
@@ -0,0 +1,58 @@
...
+    // Only render the copy button if browser supports the Clipboard API
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API
+    if ("clipboard" in navigator) setSupportsClipboard(true)
+  }, [])
+
+  function copyCodesToClipboard() {
+    navigator.clipboard.writeText(backupSecondFactors.join("\n"))
+  }
+
+  function downloadCodes() {
```
mc-jones, line 23: Is there a security risk we need to consider in using a data blog for sensitive values? One thing is I believe you need to explicitly revoke the URL object to prevent memory leaks or unintended persistence. Any other security issues?

Good reference here: https://github.com/artsy/force/blob/master/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/SmsSecondFactor/Modal.tsx#L77

lidimayra: Hey @mc-jones , I guess the wrong link was pasted here by accident. This is the one contained in the other comment, related to the mutation handling.

I pushed this, though: 047ff31379846a01a446c899ea3473230e8f127d
Is this what you meant?

mc-jones: Hey @lidimayra. Sorry about the wrong link! Here is the correct one for reference: https://www.w3.org/TR/FileAPI/#creating-revoking. 

But as discussed, I'll take finish out from here once the migration from @The-Beez-Kneez  is done.

mc-jones:
@dzucconi, @The-Beez-Kneez - Just want to check if there is a ticket(s) I can follow to track the migration of the Backup code, so can make sure to finish out this work, when ready?

dzucconi:
@mc-jones thanks for reminding me — as I forgot about this PR.

I took over migration of these components and did this module on Friday.

Should be easy to drop in a copy/download button

https://github.com/artsy/force/blob/master/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/SettingsEditSettingsTwoFactorBackupCodesDialog.tsx

dzucconi:
(Note that this is a totally new component now — so probably warrants a new PR)

mc-jones:
@dzucconi - I think it makes sense to wait until App and SMS are also migrated, as this will also touch those components via the [Reminder component](https://github.com/artsy/force/blob/359cd6c4b65297d77f59edda29e5727ff414286b/src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactorReminder.tsx) (which is admittedly repetitive of the backup code modal and gets displayed after each enrollment as a way to encourage users to save their backup codes and limit lockout).

dzucconi:
@mc-jones cool — I'll keep you in the loop. Should have this done in the next few days.

src/v2/Components/UserSettings/TwoFactorAuthentication/Components/BackupSecondFactor/__tests__/BackupSecondFactorActions.jest.tsx
```diff
@@ -0,0 +1,73 @@
...
+      global.Blob = function (content, options) {
```
damassi, line 60: note its pretty rare to use `function` keyword like this with arrow functions now being the standard: 

```tsx
global.Blob = (content, options) => { ... } 
```

mc-jones:
@damassi - This good to finish out in a new PR, now that you have moved these components to v3 with #9048? Or is there more work to be done to migrate?

dzucconi:
@mc-jones yeah now this should be safe to finish up. Thanks for bearing with.

mc-jones:
No worries. Thanks, @dzucconi. Also, sorry @damassi, for the mistag just now.

lidimayra:
Closing this in favor of #9067

PR #8939, opened 2021-11-24
Remove SellWithArtsy component

lidimayra:
This component is not used anymore. For ref, see: https://github.com/artsy/force/pull/8859#discussion_r751168029

According to our history, this component was used inside `MarketingLandingApp` component in the past, however, usage was removed at this point: 26becb003c5acd1e2d9ba7c0a71d8739b0949bee


mdole (approved):
nice! thanks @lidimayra!

PR #8940, opened 2021-11-24
fix: Attachment display on conversation sidebar

jo-rs:
[NX-3027]

Fixed an issue causing the length of the attachments to display on the conversation sidebar if there was none, and no attachments if any were present.
Also included is an ellipsis for long file names while at it, but this area's design is set to be revisited soon.

Before (displaying the actual `attachments.length`):
image

After (with and without attachments):
Screenshot 2021-11-24 at 15 09 44
Screenshot 2021-11-24 at 12 18 58

cc: @artsy/negotiate-devs 

[NX-3027]: https://artsyproduct.atlassian.net/browse/NX-3027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Conversation/Components/Details.tsx
+++ b/src/v2/Apps/Conversation/Components/Details.tsx
@@ -101,11 +101,6 @@ export const Details: FC<DetailsProps> = ({
     ?.map(edge => edge?.node?.attachments)
     ?.filter(attachments => attachments?.length)
     ?.flat()
-  // ?.filter(attachment => !attachment?.contentType.includes("image"))
-
-  useEffect(() => {
-    console.log("Hey HEy", attachments)
-  }, [])
 
   const attachmentItems = attachments
     ?.filter(attachment => attachment?.id && attachment?.downloadURL)
--- a/src/v2/Apps/Conversation/Components/Details.tsx
+++ b/src/v2/Apps/Conversation/Components/Details.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect } from "react";
+import { FC, useEffect } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import {
   Box,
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Conversation/Components/Details.tsx
```diff
@@ -195,7 +200,7 @@ export const Details: FC<DetailsProps> = ({
           </Flex>
         </>
       )}
-      {attachments?.length && (
+      {!!attachments?.length && (
```
jo-rs, line 203: Fixes the length being displayed if there was none.

src/v2/Apps/Conversation/Components/Details.tsx
```diff
@@ -94,8 +100,7 @@ export const Details: FC<DetailsProps> = ({
   const attachments = conversation?.messagesConnection?.edges
     ?.map(edge => edge?.node?.attachments)
     ?.filter(attachments => attachments?.length)
-    ?.reduce((previous, current) => previous?.concat(current!), [])
-    ?.filter(attachment => !attachment?.contentType.includes("image"))
+    ?.flat()
```
jo-rs, line 103: Simplifies this logic, and removes the unnecessary image filter.

src/v2/Apps/Conversation/Components/Details.tsx
```diff
@@ -55,6 +55,12 @@ const DetailsContainer = styled(Flex)<{ transform?: string }>`
   `}
 `
 
+const TruncatedLine = styled(Text)`
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  overflow: hidden;
+`
```
jo-rs, line 62: Quick and simple ellipsis for long file names.

araujobarret (approved):

PR #8944, opened 2021-11-24
feat(FX-3607): Upon email alert click scroll down to artwork grid immediately

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3607]

### Description
* Clicking the CTA in the email alert should lead the user directly to the artworks they want to see on the Artist’s Works for Sale tab
* The effect should be that the screen “jumps” right to the artwork grid on the artist artwork tab, rather than land at the top of the mWeb page

### Demo
[Demo link](https://artist-artworks-scroll.artsy.net/artist/kaws/works-for-sale?attribution_class%5B0%5D=unique&additional_gene_ids%5B0%5D=sculpture&search_criteria_id=f5c5e83b-a53c-4461-a496-bba7ddc109ca)
#### Web
https://user-images.githubusercontent.com/3513494/143268199-14950fca-c7da-48b3-af4c-c797f658f7e4.mp4

#### mWeb
https://user-images.githubusercontent.com/3513494/143268226-118f2caf-e340-46c6-aa16-e42dce1c434e.mp4

[FX-3607]: https://artsyproduct.atlassian.net/browse/FX-3607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
@@ -31,7 +31,9 @@ const ArtistWorksForSaleRoute: React.FC<ArtistWorksForSaleRouteProps> = ({
 
   useEffect(() => {
     if (isMatchMediaParsed && match?.location?.query?.search_criteria_id) {
-      scrollTo()
+      requestAnimationFrame(() => {
+        scrollTo()
+      })
     }
   }, [isMatchMediaParsed])
 
--- a/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
@@ -29,6 +29,13 @@ const ArtistWorksForSaleRoute: React.FC<ArtistWorksForSaleRouteProps> = ({
     offset: 10,
   })
 
+  /**
+   * We should perform scroll only when isReadyForUse is true
+   * Otherwise the wrong offset will be used for mWeb
+   *
+   * scrollTo without requestAnimationFrame doesn't work in Safari
+   * when it is used in useEffect hook
+   */
   useEffect(() => {
     if (isReadyForUse && match?.location?.query?.search_criteria_id) {
       requestAnimationFrame(() => {
--- a/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
@@ -1,4 +1,4 @@
-import * as React from "react"
+import React, { useEffect } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { ArtistArtworkFilterRefetchContainer } from "./Components/ArtistArtworkFilter"
 import { ArtistWorksForSaleRoute_artist } from "v2/__generated__/ArtistWorksForSaleRoute_artist.graphql"
@@ -7,6 +7,8 @@ import { ArtistSeriesRailFragmentContainer } from "v2/Components/ArtistSeriesRai
 import { ContextModule } from "@artsy/cohesion"
 import { computeTitle } from "v2/Apps/Artist/Utils/computeTitle"
 import { Title } from "react-head"
+import { useScrollTo } from "v2/Utils/Hooks/useScrollTo"
+import { useRouter } from "v2/System/Router/useRouter"
 
 interface ArtistWorksForSaleRouteProps {
   artist: ArtistWorksForSaleRoute_artist
@@ -20,6 +22,18 @@ const ArtistWorksForSaleRoute: React.FC<ArtistWorksForSaleRouteProps> = ({
     artist?.counts?.forSaleArtworks!,
     true
   )
+  const { match } = useRouter()
+  const { scrollTo, isMatchMediaParsed } = useScrollTo({
+    selectorOrRef: "#jump--artworkFilter",
+    behavior: "smooth",
+    offset: 10,
+  })
+
+  useEffect(() => {
+    if (isMatchMediaParsed && match?.location?.query?.search_criteria_id) {
+      scrollTo()
+    }
+  }, [isMatchMediaParsed])
 
   return (
     <>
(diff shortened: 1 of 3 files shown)


src/v2/Utils/Hooks/useScrollTo.ts
```diff
@@ -47,5 +47,5 @@ export const useScrollTo = ({
     window.scrollTo({ top: position, behavior })
   }
 
-  return { scrollTo }
+  return { scrollTo, isMatchMediaParsed: isMobile !== null }
```
dzucconi, line 50: What's this mean?

dzucconi: / why would we include it in this hook? Doesn't seem relevant to scrolling.

dimatretyak: If we try to use `scrollTo` in `useEffect` hooks, we get that `isMobile` will be `null` and for this reason there will be an [incorrect offset](https://github.com/artsy/force/blob/dfc405fdc09032d01f881e1b21565451cb5cdb3d/src/v2/Utils/Hooks/useScrollTo.ts#L45) for mWeb

src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
```diff
@@ -20,6 +22,20 @@ const ArtistWorksForSaleRoute: React.FC<ArtistWorksForSaleRouteProps> = ({
...
+  const { match } = useRouter()
+  const { scrollTo, isMatchMediaParsed } = useScrollTo({
+    selectorOrRef: "#jump--artworkFilter",
+    behavior: "smooth",
+    offset: 10,
+  })
+
+  useEffect(() => {
+    if (isMatchMediaParsed && match?.location?.query?.search_criteria_id) {
+      requestAnimationFrame(() => {
```
dzucconi, line 34: Why wrap this in a RAF?

dimatretyak: To make `scrollTo` work for Safari

src/v2/Utils/Hooks/useScrollTo.ts
dzucconi: Could you add a comment explaining that? Is there a way to rename this variable such that it's not concerned with the internal implementation details?

src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
dzucconi: Why doesn't it work in Safari, and what about RAF makes it work? Does `setTimeout(() => {}, 0)` accomplish the same thing?

dimatretyak: > Does setTimeout(() => {}, 0) accomplish the same thing?

Yep

dimatretyak: > Why doesn't it work in Safari, and what about RAF makes it work?

Unfortunately, I could not figure out the reason why `scrollTo` doesn't work in Safari if used in `useEffect`

For example, if we use this code, it will work fine in Google Chrome, but it will not work in Safari
```javascript
React.useEffect(() => {
  window.scrollTo({
    top: 1000,
    behavior: "smooth",
  })
})
```

### Google Chrome
https://user-images.githubusercontent.com/3513494/143281470-c104c4f7-b090-4d1f-806f-1c63bec0a05b.mp4

### Safari
https://user-images.githubusercontent.com/3513494/143281514-e70690a5-46b1-43c4-ab88-4fe7d795d6e8.mp4

dzucconi: Got it — I'd just add a comment explaining that. Anything that's not immediately obvious, ya kno?

Also, I'm curious if `useLayoutEffect` makes any difference?

src/v2/Utils/Hooks/useScrollTo.ts
dimatretyak: > Is there a way to rename this variable such that it's not concerned with the internal implementation details

will the name `isReady` be suitable?

src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
dimatretyak: Unfortunately, `useLayoutEffect` doesn't work either

src/v2/Utils/Hooks/useScrollTo.ts
dzucconi: Why don't we just set the default value in useMatchMedia to `false` rather than having what is essentially a 3-state boolean? It's not immediately clear to me why this is happening? The matchMedia event runs after you want it to?

src/v2/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
dzucconi: 👍

src/v2/Utils/Hooks/useScrollTo.ts
dimatretyak: I think it is possible to refactor `__internal__use of matchMedia` hook so that the `matches` value is specified immediately and [not in `useEffect`](https://github.com/artsy/force/blob/a1d30f7aa4430b09496844f8e07183dd013f2b84/src/v2/Utils/Hooks/useMatchMedia.ts#L36).

dzucconi: If you see an angle on that, by all means fix the root cause. Though it does need to be in useEffect presumably because this can only run on client-side.

dimatretyak: > this can only run on client-side

Good point. I didn't think about it

dimatretyak: > Why don't we just set the default value in useMatchMedia to false

In this case we will get the wrong behavior. For example, a user will open Force on a mobile device -> we will pass `themeProps.mediaQueries.xs` in `__internal__useMatchMedia`. In this case, we will immediately get `isMobile = false`, which is _not entirely true_, because the user uses a mobile device

dimatretyak: In the current implementation:
- If `__internal__useMatchMedia` returned `null`, then the `mediaQueryString` check has not yet been performed
- If `__internal__useMatchMedia` returned `true`, then the device falls under the conditions of `mediaQueryString`
- If `__internal__useMatchMedia` returned `false`, then the device is **not** under the conditions of `mediaQueryString`

dzucconi (approved):
Thanks for the explanations. Feel free to merge when 🍏

PR #8957, opened 2021-11-25
feat: Upon email alert click scroll down to artwork grid immediately

dimatretyak:
Type: Feature
Jira ticket: [FX-3607]

Description
- Clicking the CTA in the email alert should lead the user directly to the artworks they want to see on the Artist’s Works for Sale tab
- The effect should be that the screen “jumps” right to the artwork grid on the artist artwork tab, rather than land at the top of the mWeb page

[FX-3607]: https://artsyproduct.atlassian.net/browse/FX-3607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilterAnchor.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilterAnchor.tsx
@@ -1,0 +1,29 @@
+/* tslint:disable */
+/* eslint-disable */
+
+import React from "react"
+import { themeProps } from "@artsy/palette-tokens"
+import { useNavBarHeight } from "../NavBar/useNavBarHeight"
+import { __internal__useMatchMedia } from "v2/Utils/Hooks/useMatchMedia"
+import { useSticky } from "../Sticky"
+
+// Based on this answer https://stackoverflow.com/a/13184714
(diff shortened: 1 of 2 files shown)


src/v2/Components/ArtworkFilter/ArtworkFilterAnchor.tsx
```diff
@@ -0,0 +1,29 @@
+/* tslint:disable */
+/* eslint-disable */
```
damassi, line 2: Out of curiosity, why add these?

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -268,8 +269,7 @@ export const BaseArtworkFilter: React.FC<
 
   return (
     <Box mt={tokens.mt} {...rest}>
-      <Box id="jump--artworkFilter" />
-
+      <ArtworkFilterAnchor />
```
damassi, line 272: if this going to be added in a follow-up PR, can we find other instances of this kind of use elsewhere in app and replace with `ArtworkFilterAnchor`?

PR #8963, opened 2021-11-25
bugfix: cleans price currency before rendering

ashleyjelks:
Currently, `price` can be any of the following formats:
`Under $1000`
`1000`
`$1000`
`"1000`
`$1000 - 2000`
`0`
`"0"`
`£150`
`€150`
`""`
`"-"`

This PR fixes a bug caused by not properly handling all the various formats that `price` is in when we pass that value to components that render an artwork's price in the BNMO flow. 

https://artsyproduct.atlassian.net/browse/NX-3029

This is a follow-up to #8884 

cc @artsy/negotiate-devs 

**The Bug:**
Screen Shot 2021-11-25 at 12 02 11 PM

**The Fix:**
Screen Shot 2021-11-25 at 12 02 23 PM

### DIFF
--- a/src/v2/Apps/Order/Utils/__tests__/currencyUtils.jest.ts
+++ b/src/v2/Apps/Order/Utils/__tests__/currencyUtils.jest.ts
@@ -1,0 +1,24 @@
+import { appendCurrencySymbol } from "v2/Apps/Order/Utils/currencyUtils"
+
+describe("#appendCurrencySymbol", () => {
+  it("correctly appends US currency symbol", () => {
+    expect(appendCurrencySymbol("11111", "USD")).toBe("US$11111")
+    expect(appendCurrencySymbol("-", "USD")).toBe("-")
+    expect(appendCurrencySymbol("gobledygook", "USD")).toBe("gobledygook")
+    expect(appendCurrencySymbol(22222, "USD")).toBe("US$22222")
+    expect(appendCurrencySymbol("Under $1000", "USD")).toBe("Under US$1000")
+    expect(appendCurrencySymbol("", "USD")).toBe("")
+    expect(appendCurrencySymbol(null, "USD")).toBe(null)
+    expect(appendCurrencySymbol(undefined, "USD")).toBe(undefined)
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Order/Utils/__tests__/currencyUtils.jest.ts
```diff
@@ -0,0 +1,24 @@
+import { appendCurrencySymbol } from "v2/Apps/Order/Utils/currencyUtils"
+
+describe("#appendCurrencySymbol", () => {
+  it("correctly appends US currency symbol", () => {
```
araujobarret, line 4: All this control in the frontend side worries me a bit, my gut feeling tells me that we should be controlling all cases from the backend, the frontend should just display the field 

I know that was already implemented, but reading it a second time I got this feeling

sweir27: I agree with @araujobarret here! Especially given recent talks around consolidating how we display currencies across the platform, I think we should avoid having logic like this in our front-ends (it makes it harder to consolidate & maintain later).

I know that @starsirius has also been thinking about this a bit! Might be a good opportunity to pair?

ashleyjelks: I agree with everyone as well. However, I discussed different implementation options [here](https://artsy.slack.com/archives/C9YNS4X32/p1636401545272200), and [another dev also looked into the work to see what's required](https://artsyproduct.atlassian.net/browse/NX-2405?focusedCommentId=40503) and the same conclusion was reached that handling this at the middleware or backend layer is ideal but since work on this page is coming down the Up Next pipeline, that the quickest approach to getting this done is ok for now. (Impact over perfection?) Anyways, I'm happy to revert all of this work and scope out tickets to get it done using one of the other approaches, so lemme circle back to the team and see what makes sense for next steps.

ashleyjelks: Merged this and created a followup ticket per https://artsy.slack.com/archives/C9YNS4X32/p1638211535024400!

PR #8966, opened 2021-11-26
fix(partners): adds links to hero units

dzucconi:
Embarrassingly, this shipped without being clickable.


src/v2/Components/ArtworkFilter/__tests__/FiltersPills.jest.tsx
```diff
@@ -6,6 +6,7 @@ const savedSearchAttributes: SavedSearchAttributes = {
   type: "artist",
   id: "test-artist-id",
   name: "test-artist-name",
+  slug: "example-slug",
```
dzucconi, line 9: Not sure how this passed on the PR but this was failing type check. cc @anastasiapyzhik

PR #8968, opened 2021-11-29
[Feat]: AlicanAkyuz/swa-135/session id in consignment flow

AlicanAkyuz:
### Description 

Ticket: [SWA-135](https://artsyproduct.atlassian.net/browse/SWA-135) 

This PR adds sessionId to the addAsset mutation input. As a part of securing that a user cannot update another user's submission, we already send sessionId to our backend to associate a submission with a session if user is not logged in. The next task is to send sessionId in AddAssetsToConsignmentSubmission as well, so that we can associate an addAsset request with the owner of the submission.


kiryl-zubarau (approved):
🚀

Serge0n (approved):

PR #8969, opened 2021-11-29
feat(FX-3595): Filter pills interaction

anastasiapyzhik:
[FX-3595]

Acceptance Criteria:
-When selecting a filter, user sees pill with selected option name on top of the artworks grid
-When tapping on the 'X' of a pill it disappears and the filter related to this pill gets unselected

Changes are available with `ENABLE_SAVED_SEARCH` feature flag enabled

### Demo

https://user-images.githubusercontent.com/44819355/144020160-671ea960-7770-4e67-816b-59a4a8d72b33.mp4

[FX-3595]: https://artsyproduct.atlassian.net/browse/FX-3595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/WaysToBuyFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/WaysToBuyFilter.tsx
@@ -1,5 +1,5 @@
 import { Checkbox, Flex, useThemeConfig } from "@artsy/palette"
-import { isEmpty } from "lodash"
+import { entries, isEmpty } from "lodash"
 import { FC } from "react"
 import {
   ArtworkFilters,
@@ -17,6 +17,25 @@ export interface WaysToBuyFilterProps {
   expanded?: boolean
 }
 
+export const WAYS_TO_BUY_OPTIONS = {
+  acquireable: {
+    name: "Buy Now",
+    countName: "ecommerce_artworks",
+  },
+  offerable: {
+    name: "Make Offer",
+    countName: "has_make_offer_artworks",
+  },
+  atAuction: {
+    name: "Bid",
+    countName: "auction_artworks",
+  },
+  inquireableOnly: {
+    name: "Inquire",
+    countName: "for_sale_artworks",
+  },
+}
+
 export const WaysToBuyFilter: FC<WaysToBuyFilterProps> = ({ expanded }) => {
   const filterContext = useArtworkFilterContext()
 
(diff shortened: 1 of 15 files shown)


src/v2/Components/ArtworkFilter/SavedSearch/Components/FiltersPills.tsx
```diff
@@ -20,17 +36,36 @@ export const FiltersPills: FC<FiltersPillsProps> = ({
...
       <Flex flexWrap="wrap" mx={-PILL_HORIZONTAL_MARGIN_SIZE}>
-        {pills.map((pill, index) => (
+        {pills.map(pill => (
           <Pill
-            key={`filter-label-${index}`}
+            key={`filter-label-${pill.name}`}
             variant="textSquare"
             mx={PILL_HORIZONTAL_MARGIN_SIZE}
             mb={1}
+            onClick={() => removePill(pill)}
```
dblandin, line 66: **question**: Might be worth clarifying with design whether the whole pill should be clickable - and trigger the removal of that filter criteria - or if only the close icon X should be clickable for the removal. It's a bit unclear from the Figma specs what's preferred.

src/v2/Components/ArtworkFilter/ArtworkFilters/WaysToBuyFilter.tsx
```diff
@@ -31,32 +50,17 @@ export const WaysToBuyFilter: FC<WaysToBuyFilterProps> = ({ expanded }) => {
...
-      disabled: isDisabled(filterContext.counts.ecommerce_artworks),
-      name: "Buy Now",
-      state: "acquireable",
-    },
-    {
-      // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-      disabled: isDisabled(filterContext.counts.has_make_offer_artworks),
-      name: "Make Offer",
-      state: "offerable",
+  const checkboxes: WayToBuy[] = entries(WAYS_TO_BUY_OPTIONS).reduce(
```
dblandin, line 53: **praise**: Nice refactoring here 👍🏼

dblandin (commented):
LGTM! Left a couple comments. I think we should double-check the desired hover/tap behavior with design before merging.

Also, this work is still behind a feature flag, right? If so, I think it would be useful to point that out in the PR description.

src/v2/Components/ArtworkFilter/SavedSearch/Components/FiltersPills.tsx
anastasiapyzhik: @dblandin it's been answered [here](https://artsyproduct.atlassian.net/browse/FX-3543#:~:text=Should%20tapping%20anywhere,and%20filter%20value), so a criteria should be removed on pill tap, not icon

dblandin: Ahh, I see that Q&A now:

> Should tapping anywhere on the pill dismiss, or only on the X?
> - Tapping a pill with a 'X' should remove the pill and filter value
> - Artist pill should not have an X

Thanks!

dblandin (approved):

PR #8973, opened 2021-11-30
fix: Hide app badges on Eigen

rajsam003:
This PR fixes the bug mentioned on this ticket [CX-2125](https://artsyproduct.atlassian.net/browse/CX-2125)

* Congratulation message will only be shown on the Order submitted page.
* App Badges will not be displayed in Eigen.


src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -318,7 +325,6 @@ export const TransactionDetailsSummaryItemFragmentContainer = createFragmentCont
             }
           }
         }
-        state
```
olerichter00, line 321: 👀

olerichter00 (approved):
LGTM 🚀

damassi (approved):
We should update corresponding tests before merge, but code looks good! 👍

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
rajsam003: We don't need this here.

olerichter00 (approved):

PR #8983, opened 2021-12-01
feat: linter convert mobile bootstrap

artsyjian:
https://artsy.slack.com/archives/CA8SANW3W/p1638375400188200

In an un-related [PR](https://github.com/artsy/force/pull/8853), I have an update to this file and notice that `git commit` converts the file to typescript (?). The conversion has no relationship to the work in that PR, so I thought it best to PR the conversion here without any actual code changes.

### DIFF
--- a/src/mobile/components/layout/bootstrap.js
+++ b/src/mobile/components/layout/bootstrap.js
@@ -10,106 +10,115 @@
 // Don't go too wild here, we want to keep this minimal and light-weight because it could be
 // included across most apps and any uncessary bloat should be avoided.
 //
+//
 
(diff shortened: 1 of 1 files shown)


damassi:
@artsyjian - looks like this is just [`prettier`](https://prettier.io/) changes which [run on commit](https://github.com/artsy/force/blob/master/package.json#L429).

artsyjian:
> @artsyjian - looks like this is just [`prettier`](https://prettier.io/) changes which [run on commit](https://github.com/artsy/force/blob/master/package.json#L429).

i see. thanks!

PR #8984, opened 2021-12-01
Revert "bugfix: cleans price currency before rendering"

ashleyjelks:
Reverts artsy/force#8963

Going to move this work to Metaphysics so reverting changes now.


damassi:
👍 on MP

PR #8985, opened 2021-12-01
refactor(PX-4712): shuffle around the shipping options display to match the design

oxaudo:
Ended up not adding space between the title and the description as this spacing seems fine now when things are in a table format.

Looks like:
Screen Shot 2021-12-01 at 12 44 36 PM


oxaudo:
Oh and for the record that was a reference figma: https://www.figma.com/file/J5bIQj4a0l7wYAHnhbVGTp/%5BCollector%5D-Checkout-Shipping?node-id=1761%3A0 but the spacing there is not consistent at all so I just did not do anything spacing wise.

pepopowitz (approved):
Looks great! 

Sorry for the slow review, I lost track of it.

PR #8988, opened 2021-12-01
feat(GRO-597): Order history

dzucconi:
Closes [GRO-597](https://artsyproduct.atlassian.net/browse/GRO-597)

### DIFF
--- a/src/v2/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
+++ b/src/v2/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
@@ -1,0 +1,351 @@
+import {
+  Box,
+  CheckCircleFillIcon,
+  Column,
+  EntityHeader,
+  Flex,
+  GridColumns,
+  HelpIcon,
+  Image,
+  PendingCircleIcon,
+  Separator,
+  Skeleton,
+  SkeletonBox,
+  SkeletonText,
+  Text,
+  XCircleIcon,
+} from "@artsy/palette"
+import { DateTime } from "luxon"
+import { FC } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { RouterLink } from "v2/System/Router/RouterLink"
+import { SettingsPurchasesRow_order } from "v2/__generated__/SettingsPurchasesRow_order.graphql"
+import { LocaleOptions } from "luxon"
+import { extractNodes } from "v2/Utils/extractNodes"
+import { appendCurrencySymbol } from "v2/Apps/Order/Utils/currencyUtils"
+import { PlaceholderEntityHeader } from "v2/Components/EntityHeaders/PlaceholderEntityHeader"
+
+const ORDER_LABELS = {
+  APPROVED: "Confirmed",
+  CANCELED: "Canceled",
+  FULFILLED: "Delivered",
+  IN_TRANSIT: "In transit",
+  PROCESSING: "Processing",
+  REFUNDED: "Refunded",
+  SUBMITTED: "Pending",
+} as const
+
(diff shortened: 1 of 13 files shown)


src/v2/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -0,0 +1,351 @@
+import {
```
dzucconi, line 1: All this code is largely the same, just cleaned up a bit. There's some issues with it we can tackle separately or ignore.

src/v2/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -0,0 +1,351 @@
...
+  Image,
+  PendingCircleIcon,
+  Separator,
+  Skeleton,
+  SkeletonBox,
+  SkeletonText,
+  Text,
+  XCircleIcon,
+} from "@artsy/palette"
+import { DateTime } from "luxon"
```
dzucconi, line 18: We should enrich all the stitched date fields rather than having to import luxon for something like formatting.

src/v2/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -0,0 +1,351 @@
...
+              </RouterLink>
+            </>
+          )}
+        </Flex>
+      </Flex>
+
+      <GridColumns m={2}>
+        <Column span={6} display="flex" alignItems="center">
+          {image ? (
+            <Image
```
dzucconi, line 117: Realizing it'd be useful to have a non-circular entity header.

src/v2/Apps/Settings/SettingsApp.tsx
```diff
@@ -41,6 +41,11 @@ const tabs = [
 ]
 
 const SettingsApp: React.FC<SettingsAppProps> = ({ me, children }) => {
+  if (!me) {
+    // TODO: `@principalField` isn't working here and can't get `onServerSideRender` to work
```
dzucconi, line 45: So this is weird. `@principalField` works in a raw query but for some reason it's not working to 403 in the application.

I also couldn't get `onServerSideRender` to execute at all.

src/v2/Components/Pagination/CommercePagination.tsx
```diff
@@ -0,0 +1,31 @@
+import { Pagination } from "./index"
+import { createFragmentContainer, graphql } from "react-relay"
+
+export const CommercePaginationFragmentContainer = createFragmentContainer(
```
dzucconi, line 4: I need to circle back and transform the stitched schema to avoid this duplicated fragment container. Works for the time being.

src/v2/Components/Pagination/index.tsx
```diff
@@ -1,37 +1,40 @@
...
   hasNextPage: boolean
-  pageCursors: Pagination_pageCursors
+  // TODO: Hacks around stitching. See if we can transform the schema to make this unnecessary.
+  pageCursors: Pagination_pageCursors | CommercePagination_pageCursors
   scrollTo?: string
   offset?: number
-  getHref?: PaginationProps["getHref"]
+  getHref?: BasePaginationProps["getHref"]
   onClick?: (cursor: string, page: number) => void
   onNext?: (page: number) => void
```
dzucconi, line 25: This whole component is kinda odd. We should have a common connection interface that we have a fragment on rather than a fragment on pageCursors and then having to cargo cult the rest of the params.

damassi:
👍

PR #8991, opened 2021-12-02
fix: Update order summary 'My Collection' copy

rajsam003:
This PR includes the change request from this ticket - [CX-2193](https://artsyproduct.atlassian.net/browse/CX-2193)

image

image


src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -85,10 +86,14 @@ export class TransactionDetailsSummaryItem extends React.Component<
...
                 Congratulations! This artwork will be added to your Collection
-                once the gallery ships the order.
+                once the gallery confirms the order.
               </Text>
               <Text variant="sm">
-                View and manage your Collection in the Artsy App.
+                View and manage all artworks in your Collection{" "}
+                {!isEigen ? "on the Artsy app." : "through your "}
+                {isEigen && (
+                  <RouterLink to={"artsy:///my-profile"}>profile</RouterLink>
```
rajsam003, line 95: I am not sure if this is the way to navigate in Eigen. If there is any other way let me know I will change it.

olerichter00:
```suggestion
                  <RouterLink to={"artsy:///my-profile"}>profile.</RouterLink>
```

olerichter00: We use deep linking for iOS and Android and the web routes usually match the [routes in the app](https://github.com/artsy/eigen/blob/0f5bc8a844cccea4fc1eccee81814af30df97068/src/lib/navigation/routes.tsx#L118).

olerichter00 (changes requested):
Looks good so far. I've added a comment about a typo and the link to `/my-profile`

olerichter00 (approved):
🚀

PR #8994, opened 2021-12-02
feat(NX-3011): new Offer Submission inputs and calculation

tam-kis:
New Offer Submission feature : [NX-3011](https://artsyproduct.atlassian.net/browse/NX-3011) , [NX-3012](https://artsyproduct.atlassian.net/browse/NX-3012) 

- Added new radio options
- Calculation for Exact and Range offers
- Input toggle when Different amount is selected
- Style and text changes for Add note and Info sections

`PriceOptions.tsx` is the component with all the radios and price calculations. It's imported in `Order/Routes/Offer/index.tsx`

Exact price:
Screenshot 2021-12-20 at 17 57 43

Range price:
Screenshot 2021-12-20 at 17 57 06

Different amount input:

Screenshot 2021-12-20 at 17 57 52

Screenshot 2021-12-20 at 17 58 06


src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,109 @@
...
+            label={asCurrency(value!)}
+            onSelect={() => setValue(value!)}
+            key={`price-option-${value}`}
+            data-test="price-options"
+          >
+            <Text variant="sm" color="black60">
+              {description}
+            </Text>
+          </BorderedRadio>
+        )) as any /* FIXME */
```
jo-rs, line 87: There's a TS error that happens here (visible when removing the `as any`)

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,145 @@
...
+                />
+              </Flex>
+            )}
+          </BorderedRadio>
+        )}
+    </RadioGroup>
+  )
+}
+
+export const PriceOptionsFragmentContainer = createFragmentContainer(
```
araujobarret, line 122: 👍

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+
+import { PriceOptions_artwork } from "v2/__generated__/PriceOptions_artwork.graphql"
+import { appendCurrencySymbol } from "../Utils/currencyUtils"
+
+export interface PriceOptionsProps {
+  setValue: (value: number) => void
+  onFocus: () => void
+  showError?: boolean
+  artwork: PriceOptions_artwork | null | undefined
```
araujobarret, line 14: You don't need to specify `null | undefined` here, the `PriceOptions_artwork`, artwork in this context will be always defined(you can't offer to an undefined artwork) but the properties might be null see auto-generated graphql file

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,145 @@
+import { BorderedRadio, Flex, RadioGroup, Text } from "@artsy/palette"
+import { useEffect, useState } from "react"
+import { OfferInput } from "v2/Apps/Order/Components/OfferInput"
+import { compact } from "lodash"
+import { createFragmentContainer, graphql } from "react-relay"
+
+import { PriceOptions_artwork } from "v2/__generated__/PriceOptions_artwork.graphql"
+import { appendCurrencySymbol } from "../Utils/currencyUtils"
```
araujobarret, line 8: @jo-rs can we try tackling this Gravity tech debt after the end of January? it's only increasing in Force/Eigen

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,145 @@
...
+            }}
+            data-test="price-option"
+            key="price-option-custom"
+          >
+            {toggle && (
+              <Flex flexDirection="column" marginTop={2}>
+                <OfferInput
+                  id="OfferForm_offerValue"
+                  showError={showError}
+                  onChange={value => {
```
araujobarret, line 108: nit:
```suggestion
                  onChange={setCustomValue}
```

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -0,0 +1,145 @@
...
+        ))
+        .concat(
+          <BorderedRadio
+            value="custom"
+            label="Different amount"
+            onSelect={() => {
+              customValue && setValue(customValue)
+              setToggle(true)
+            }}
+            data-test="price-option"
```
araujobarret, line 100: I guess you don't need to specify the `data-test` here to be able to query the radio elements, you can you `screen.getAllByRole("radio")`, also is a good approach avoiding queries by `testId` as much as we can, has the lowest priority from the library
more info on this [link](https://testing-library.com/docs/queries/about/#priority)

src/v2/Apps/Order/Components/__tests__/PriceOptions.jest.tsx
```diff
@@ -0,0 +1,103 @@
...
+  })
+  it("fires click event with correct value", () => {
+    fireEvent.click(radios[1])
+    expect(radios[1]).toBeChecked()
+    expect(setValue).toHaveBeenLastCalledWith(150)
+  })
+  it("conditionally displays input field", async () => {
+    expect(within(radios[3]).queryByRole("textbox")).not.toBeInTheDocument()
+    fireEvent.click(radios[3])
+    const input = await within(radios[3]).findByRole("textbox")
```
araujobarret, line 68: good `within` usage 👏

src/v2/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -234,22 +252,47 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
...
-                  onChange={offerValue => this.setState({ offerValue })}
-                  onFocus={this.onOfferInputFocus.bind(this)}
-                />
-              </Flex>
-              {Boolean(offerItem?.price) && (
-                <Text my={1} variant="xs" color="black60">
-                  List price:{" "}
-                  {appendCurrencySymbol(offerItem?.price, order.currencyCode)}
-                </Text>
+              {(!newOfferSubmissionEnabled || isInquiryCheckout) && (
```
araujobarret, line 255: the name here `isInquiryCheckout` is a bit dubious to me, I was reading thinking that's is the offer inquiry but then I re-checked and the variable meaning is if the artwork has no price/priceRange, can you please change it to a proper explanatory name?
I don't have a strong suggestion but would stick to something related to the meaning, `hasNoPrice`, etc

araujobarret (commented):
Overall looks good, thanks for using RTL in the tests. I left some comments but none of them are major issues

src/v2/Apps/Order/Routes/Offer/index.tsx
jo-rs: Yeah, this is a tricky one because, sure `hasNoPrice`  makes sense, but context-wise, the only reason that the user would be seeing this component and a price wouldn’t be present is if the offer was started through conversations and hence: `isInquiryCheckout`.

src/v2/Apps/Order/Components/PriceOptions.tsx
jo-rs: What is this referring to, @araujobarret ? 🤔

araujobarret: Isn't this `appendCurrySymbol` the USD thing from the other ticket?

src/v2/Apps/Order/Routes/Offer/index.tsx
araujobarret: Ah okay, so when we're in the `Offer` submission flow given no price nor listPrice we consider it `InquiryCheckout`, right?
Can you clarify the difference between Inquiry and Offer, please? 🙏

araujobarret (approved):

src/v2/Apps/Order/Components/PriceOptions.tsx
jo-rs: Oh, yes, sorry. Defo something to tackle in MP, but I think it's already somewhere on the board. Right, @ashleyjelks ? 🤔

src/v2/Apps/Order/Routes/Offer/index.tsx
jo-rs: Of course!
We consider an "Inquiry" when the collector clicks on a "Contact Gallery" button on the artwork page. If the user then clicks on "Make Offer" from within that inquiry/conversation, then we consider that offer to be "inquiry checkout", which you'll also see it referred as "MOI" => "Make Offer on Inquiry".

A regular offer, or a non-MOI, is one where the conversation tab wasn't involved to start the offer; the user clicked on "Make Offer" directly on the artwork page.

PR #8995, opened 2021-12-02
(FX-3617): total count of filters

tr-ann:
Jira: [FX-3617](https://artsyproduct.atlassian.net/browse/FX-3617)

### Description
Display total count of filters at top of artwork grid in line with the current app experience

### Changes
- add helper to count filters
- add tests for that

### Demo

https://user-images.githubusercontent.com/56556580/144445982-b7583d05-29e0-42c9-bc06-b6221c9eddce.mov

### DIFF
--- a/src/v2/Components/ArtworkFilter/Utils/getSelectedFIltersCounts.ts
+++ b/src/v2/Components/ArtworkFilter/Utils/getSelectedFIltersCounts.ts
@@ -1,0 +1,70 @@
+import { isArray, isEmpty } from "lodash"
+import {
+  ArtworkFilters,
+  FilterParamName,
+  initialArtworkFilterState,
+  SelectedFiltersCounts,
+} from "../ArtworkFilterContext"
+
+const waysToBuyFilterNames = [
+  FilterParamName.waysToBuyBuy,
+  FilterParamName.waysToBuyMakeOffer,
+  FilterParamName.waysToBuyBid,
+  FilterParamName.waysToBuyInquire,
+]
+
+const rangeFilterNames = [
+  FilterParamName.width,
+  FilterParamName.height,
+  FilterParamName.priceRange,
+]
+
+export const getSelectedFiltersCounts = (selectedFilters: ArtworkFilters) => {
+  const counts: Partial<SelectedFiltersCounts> = {}
+  const filtersParams = Object.values(FilterParamName)
+
+  Object.entries(selectedFilters).forEach(([paramName, paramValue]) => {
+    if (!filtersParams.includes(paramName as any)) {
+      return
+    }
+
+    switch (true) {
+      case waysToBuyFilterNames.includes(paramName as FilterParamName): {
+        counts.waysToBuy = (counts.waysToBuy ?? 0) + 1
+        break
+      }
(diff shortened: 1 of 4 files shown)


src/v2/Components/ArtworkFilter/Utils/getSelectedFIltersCounts.ts
```diff
@@ -0,0 +1,72 @@
...
+  FilterParamName.waysToBuyInquire,
+]
+
+const rangeFilterNames = [
+  FilterParamName.width,
+  FilterParamName.height,
+  FilterParamName.priceRange,
+]
+
+export const getSelectedFiltersCounts = (selectedFilters: ArtworkFilters) => {
```
dimatretyak, line 22:
```javascript
export const getSelectedFiltersCounts = (selectedFilters: ArtworkFilters = {}) => {...}
```

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -169,6 +170,12 @@ export const BaseArtworkFilter: React.FC<
   const previousFilters = usePrevious(filterContext.filters)
   const { user, isLoggedIn } = useSystemContext()
   const { pills = [], setPills } = useFilterPillsContext()
+  const currentlySelectedFilters = getSelectedFiltersCounts(
+    filterContext.currentlySelectedFilters?.() || {}
+  )
```
dimatretyak, line 175: **suggestion**

```javascript
export const getSelectedFiltersCounts = (selectedFilters: ArtworkFilters = {}) => {...}
...
const currentlySelectedFilters = getSelectedFiltersCounts(
    filterContext.currentlySelectedFilters?.()
)
```

dblandin (dismissed):
Looks great!! 💯

src/v2/Components/ArtworkFilter/Utils/getSelectedFIltersCounts.ts
```diff
@@ -0,0 +1,72 @@
...
+        counts.artistIDs = (counts.artistIDs ?? 0) + 1
+        break
+      }
+      case isArray(paramValue): {
+        if (isEmpty(paramValue)) {
+          break
+        }
+
+        const isArtistsFilter = paramName === FilterParamName.artistIDs
+        const countToAdd = isArtistsFilter ? counts.artistIDs ?? 0 : 0
```
dimatretyak, line 54: **question**: can't we just use `const countToAdd = counts.artistIDs ?? 0` here ?

tr-ann: We check if it is artists filter because we have filter `artistsIFollow` that is processed separately. So if we have `artistsIFollow: true` we should add all artist filter options to `counts.artistsID`.

I think we can replace it with:
```
if (paramName === FilterParamName.artistIDs) {
  counts.artistIDs = paramValue.length + counts.artistIDs ?? 0
  break
}

counts[paramName] = paramValue.length
break
```

dimatretyak (approved):

tr-ann:
approved by Lois

PR #8999, opened 2021-12-02
GRO-544: Hover state added for Home Page Article Cards

The-Beez-Kneez:
New behavior:
https://user-images.githubusercontent.com/30025439/144517193-e14dd232-6805-430f-9cbb-1a4fb55578b4.mov

### DIFF
--- a/src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
+++ b/src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
@@ -20,6 +20,7 @@ import {
 import { compact, take } from "lodash"
 import * as React from "react";
 import { createFragmentContainer, graphql } from "react-relay"
+import { StyledLink } from "v2/Components/Links/StyledLink";
 import { Masonry } from "v2/Components/Masonry"
 import { useSystemContext, useTracking } from "v2/System"
 import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
@@ -27,6 +28,7 @@ import { RouterLink } from "v2/System/Router/RouterLink"
 import { Media } from "v2/Utils/Responsive"
 import { HomeFeaturedMarketNewsQuery } from "v2/__generated__/HomeFeaturedMarketNewsQuery.graphql"
 import { HomeFeaturedMarketNews_articles } from "v2/__generated__/HomeFeaturedMarketNews_articles.graphql"
+import styled from "styled-components"
 
 const ARTICLE_COUNT = 6
 
@@ -34,8 +36,15 @@ interface HomeFeaturedMarketNewsProps {
   articles: HomeFeaturedMarketNews_articles
 }
 
+const StyledTitle = styled(StyledLink)`
+  text-decoration: none;
+  &:hover {
+    color: ("blue100");
+  }
+`
+
 const HomeFeaturedMarketNews: React.FC<HomeFeaturedMarketNewsProps> = ({
-  articles,
+  articles
 }) => {
   const { trackEvent } = useTracking()
   const [firstArticle, ...restOfArticles] = articles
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
```diff
@@ -20,22 +20,31 @@ import {
 import { compact, take } from "lodash"
 import * as React from "react";
 import { createFragmentContainer, graphql } from "react-relay"
+import { StyledLink } from "v2/Components/Links/StyledLink";
```
dzucconi, line 23: We don't need to import this. What you want to style is actually the parent anchor tag, which in this case is a `RouterLink`

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
```diff
@@ -20,22 +20,31 @@ import {
...
 const ARTICLE_COUNT = 6
 
 interface HomeFeaturedMarketNewsProps {
   articles: HomeFeaturedMarketNews_articles
 }
 
+const StyledTitle = styled(StyledLink)`
+  text-decoration: none;
+  &:hover {
+    color: ("blue100");
```
dzucconi, line 42: To set something as our themed blue you need to use `themeGet`

```
```

Which is a function that is bound to our underlying theme and let's us access properties from it.

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
```diff
@@ -82,8 +91,10 @@ const HomeFeaturedMarketNews: React.FC<HomeFeaturedMarketNewsProps> = ({
             <Text variant="xs" textTransform="uppercase" my={1}>
               {firstArticle.vertical}
             </Text>
-
+          
+          <StyledTitle to={firstArticle.href ?? ""}>
```
dzucconi, line 95: So right here you'd have an anchor tag nested within another anchor tag, which is invalid.

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
```diff
@@ -136,9 +147,11 @@ const HomeFeaturedMarketNews: React.FC<HomeFeaturedMarketNewsProps> = ({
...
                       {article.vertical}
                     </Text>
-
-                    <Text variant="lg">{article.title}</Text>
-
+                    
+                    <StyledTitle to={article.href ?? ""}>
+                      <Text variant="lg">{article.title}</Text>
+                    </StyledTitle>
+
```
dzucconi, line 154: Ditto, just style the parent one and then if you want you can target specific sub-nodes.

https://styled-components.com/docs/advanced#referring-to-other-components

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
The-Beez-Kneez: What do you pass in here for the `to` prop then? It will not allow it to exist without something being passed for that prop?

dzucconi: See above — you're won't be using a link here at all. There can only be a single link; which is the parent `RouterLink`

dzucconi (approved):

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
The-Beez-Kneez: Incorporated the `RouterLink` instead.

src/v2/Apps/Home/Components/HomeFeaturedMarketNews.tsx
The-Beez-Kneez: Incorporated the `RouterLink`

jonallured (approved):
LGTM!! 🎉

PR #9009, opened 2021-12-06
feature: Track Page Cache Metrics

icirellik:
Due to some limitations in how Datadog determines what route to use for a trace, the cached page data needs to be returned in the final route. This allows the trace to select the correct URL.


ovasdi (approved):
LGTM ❗

PR #9010, opened 2021-12-06
GRO-548: [Mobile] Trending Artists on Artsy - CTA link is wrapping

The-Beez-Kneez:
Current Behavior:
Screen Shot 2021-12-06 at 2 00 43 PM

New Behavior:
https://user-images.githubusercontent.com/30025439/145096403-cea9fa79-f3bf-456a-80a9-866d00eda3f3.mov

### DIFF
--- a/src/v2/Apps/Home/Components/HomeCurrentFairs.tsx
+++ b/src/v2/Apps/Home/Components/HomeCurrentFairs.tsx
@@ -103,6 +103,7 @@ const HomeCurrentFairsContainer: React.FC = ({ children }) => {
 
         <Text
           variant="sm"
+          textAlign="right"
           as={RouterLink}
           // @ts-ignore
           to="/art-fairs"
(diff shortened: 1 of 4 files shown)


The-Beez-Kneez:
@dzucconi would love to clarify if this was what Jonathan was referring to.

dzucconi (changes requested):
So you're going to want to find the `viewAllLabel` text inside of the `RailHeader` and text align that. Replicate by viewing the component at a mobile viewport.

dzucconi:
You want to replicate by forcing that line of text to break. So either artificially increase the title to check or really size down your viewport.

The-Beez-Kneez:
Updated and also included a screen recording of the functionality for the `View All` text wrapping with right alignment.

dzucconi:
It's still wrapping with left alignment — let's pair on the solution.

dzucconi (commented):
`HomeFeaturedMarketNewsContainer` also has an `Explore Editorial` string that needs the alignment updated

The-Beez-Kneez:
@dzucconi updated following our pairing and discussion.

dzucconi (approved):

PR #9032, opened 2021-12-09
[WIP] fix(TX-67): html showing in messages 

rquartararo:
[TX-67](https://artsyproduct.atlassian.net/browse/TX-67)

This PR is attempting to fix the issue where links to some messages from Artsy (specifically from Zendesk) are being rendered as HTML code. 

**I'm really not sure of the best solution for this one, so other opinions and ideas would be greatly appreciated!**

My first idea was to use `react-html-parser`, however, there was an issue where links are not showing on new lines and I wasn’t sure how to fix it:

Screenshot 2021-11-05 at 3 08 45 PM (1)

The second idea (and what I'm using in this PR) is to use `dangerouslySetInnerHTML`. To make this work with Linkify, we need to replace `react-linkify` with `linkify-html`. Using `linkify-html` also removes the need for the `linkTargetDecorator()` workaround. 

We could use [sanitize-html](https://www.npmjs.com/package/sanitize-html) to remove the danger of using `dangerouslySetInnerHTML`. But now I feel we would be adding a lot of dependencies for a seemingly simple issue. 

### Before and After

Screenshot 2021-12-09 at 11 38 40 AM

Screenshot 2021-12-09 at 10 20 32 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,6 +172,8 @@
     "jsdom": "11.6.2",
     "jwt-decode": "2.2.0",
     "knox": "0.9.2",
+    "linkify-html": "^3.0.4",
+    "linkifyjs": "^3.0.4",
     "lodash": "4.17.21",
     "mailcheck": "1.1.1",
     "memoize-one": "^5.2.1",
--- a/src/v2/Apps/Conversation/Components/Message.tsx
+++ b/src/v2/Apps/Conversation/Components/Message.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import { createFragmentContainer } from "react-relay"
 import { graphql } from "relay-runtime"
-import Linkify from "react-linkify"
+import linkifyHtml from "linkify-html"
 import {
   Box,
   BoxProps,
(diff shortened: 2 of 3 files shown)


dzucconi:
These messages come from both us and third parties?

dzucconi:
This is a lot of dependency to add for such a small thing, as you say, already. I'd also question the need for `react-linkify` to begin with.

dzucconi:
If messages come from third parties, where are we sanitizing inputs? Can we trust messages from the server regardless?

dzucconi:
One simple option would be to trust messages from Artsy and just render the HTML while not doing so for messages not from us.

rquartararo:
> One simple option would be to trust messages from Artsy and just render the HTML while not doing so for messages not from us.

This is a good option, however, we would run into the issue of needing both `linkify-html` for the html rendered messages and `react-linkify` for the other messages. Removing linkify could also be an option, but then users would need to copy-paste the urls as there is no way to create a link in the text input.

dzucconi:
I mean, I think the real solution lies in not doing this client-side at all. Metaphysics should return clean HTML. You could sanitize it there then linkify any links.

You could optimistically render a non-linkified version client-side then just shunt the HTML in in it's place once that returns.

PR #9034, opened 2021-12-09
fix(FX-3636): Do not send initial values as saved search criteria

anastasiapyzhik:
[FX-3636]

We should not send default filter values in attributes for create alert mutation

### Demo

See sent criteria for given alert

Screenshot 2021-12-09 at 16 01 15

#### Before

Screenshot 2021-12-09 at 16 02 45

#### After

Screenshot 2021-12-09 at 16 01 44

[FX-3636]: https://artsyproduct.atlassian.net/browse/FX-3636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/ArtworkFilter/SavedSearch/Utils/savedSearchCriteria.tsx
+++ b/src/v2/Components/ArtworkFilter/SavedSearch/Utils/savedSearchCriteria.tsx
@@ -1,5 +1,8 @@
 import { SearchCriteriaAttributes } from "v2/__generated__/createSavedSearchAlertMutation.graphql"
-import { ArtworkFilters } from "../../ArtworkFilterContext"
+import {
+  ArtworkFilters,
+  initialArtworkFilterState,
+} from "../../ArtworkFilterContext"
 import { allowedSearchCriteriaKeys } from "./constants"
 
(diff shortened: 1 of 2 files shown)


src/v2/Components/ArtworkFilter/SavedSearch/Utils/savedSearchCriteria.tsx
```diff
@@ -16,11 +19,29 @@ export const getAllowedFiltersForSavedSearchInput = (
   return allowedFilters
 }
 
+const isDefaultValue = (paramName, paramValue) => {
+  return (
+    (Array.isArray(paramValue) && !paramValue.length) ||
+    initialArtworkFilterState[paramName] === paramValue
+  )
+}
+
```
dimatretyak, line 28: **suggestion**

```javascript
const isDefaultValue = (paramName, paramValue) => {
  if (Array.isArray(paramValue)) {
    return paramValue.length === 0
  }

  return initialArtworkFilterState[paramName] === paramValue
}
```

dimatretyak (dismissed):
Works great 👍

mutation

PR #9042, opened 2021-12-10
fix(SWA-161): The contact information step fields should be limited to 256 characters

Serge0n:
Jira Ticket:ticket: : [SWA-161](https://artsyproduct.atlassian.net/browse/SWA-161)

This PR adds length limits for the inputs on the Contact information step. Also, I noticed that there were a couple of inputs in ArtworkDetails step without that limit. I added it too


iskounen (approved):
Thank you for adding `maxLength` to all of the text fields in the submission flow!

PR #9048, opened 2021-12-10
fix(GRO-688): moves/migrates 2fa components

dzucconi:
Closes: [GRO-688](https://artsyproduct.atlassian.net/browse/GRO-688), [GRO-689](https://artsyproduct.atlassian.net/browse/GRO-689)

This PR moves the `TwoFactorAuthentication` components. No attempt was made at any refactors aside from utilizing the new design system and fixing how they refetch. Will leave some comments below. 

These do look passable in V2 and continue to work on the existing production settings page.


src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
```diff
@@ -59,8 +68,8 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = props => {
       setShowCompleteRedirectModal(true)
     }
 
-    if (props.me.hasSecondFactorEnabled) {
-      relayRefetch?.refetch({}, {}, showCompleteModalCallback)
+    if (me.hasSecondFactorEnabled) {
+      relay.refetch({}, {}, showCompleteModalCallback)
```
dzucconi, line 72: One minor thing I changed: previously these would accept a prop from the parent component to refetch _that_ component. I switched them to just be local refetches, which makes more sense. What would actually make sense though is not refetching at all and returning a `me` in the fields so you can simply return the correct data along with the mutation. It seems like this is not a commonly known thing that you can do?

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
```diff
@@ -121,99 +130,114 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = props => {
...
+              loading={isCreating}
+              disabled={isCreating}
+            >
+              Set Up
+            </Button>
           )}
         </Flex>
       </Flex>
+
+      {/* Modals */}
```
dzucconi, line 218: Really, there should just be a single modal.

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/BackupSecondFactor/BackupSecondFactorModalContent.tsx
```diff
@@ -1,3 +1,5 @@
+// TODO: Delete this file once settings2 is launched
```
dzucconi, line 1: These component we actually migrated.

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/README.md
```diff
@@ -0,0 +1,6 @@
+## Migration Notes
+
+- Moved from `src/v2/Components/UserSettings/TwoFactorAuthentication`
+- No files were renamed and no attempts at reorganization were made.
+- Clicking App Authenticator "Edit", confirming password, then closing the resulting pop up, prevents the user from repeating this action without a page refresh. Since the state in AppSecondFactor is governed by the confluence of a number of boolean flags, they aren't reset correctly. Should be remodelled as a single state.
```
dzucconi, line 5: One bug I noticed. Figured I'd log it here. But could backlog a ticket too.

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
damassi: I think it's pretty well known? But i could be wrong.

PR #9049, opened 2021-12-10
Web - Display offer/order information in conversation side drawer

lilyfromseattle:
# [NX-3018]

__Story__

As a collector, I want to see a full breakdown of my offer when submitting through conversations.

__Scenarios__

Make Offer in Inquiry:Given I have submitted an offer within an inquiry,When I open the side drawer on the conversation,Then I see the artwork information, the order information (order number number, order breakdown, shipping address, payment method) and any attachments to the conversation,And the icon to open the drawer is replaced and reads “Details“ alongside it on larger screens.

Inquiry:Given I am on an inquiry without an offer submitted,When I open the side drawer on the conversation,Then I see the artwork information and any attachments to the conversation,And the icon to open the drawer is replaced and reads “Details“ alongside it on larger screens.

__Before:__
Screen Shot 2021-12-10 at 3 28 13 PM

__After:__
_With an active order:_
Screen Shot 2021-12-10 at 3 20 56 PM
Screen Shot 2021-12-10 at 3 21 02 PM

_Without an active order:_
Screen Shot 2021-12-10 at 3 22 22 PM

__After All Refactoring and Updates:__
Screen Shot 2022-01-06 at 12 57 55 PM

[NX-3018]: https://artsyproduct.atlassian.net/browse/NX-3018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lilyfromseattle:
I don't know what happened with the generated files or how to address them now, if anyone else has any insight please let me know.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -301,7 +301,7 @@ export const ConversationPaginationContainer = createPaginationContainer(
         unread
         orderConnection(
           first: 10
-          states: [APPROVED, FULFILLED, SUBMITTED, REFUNDED, CANCELED]
+          states: [APPROVED, FULFILLED, SUBMITTED]
           participantType: BUYER
```
araujobarret, line 305: we have some cases where we also query from the seller, for example when the seller sends you a counteroffer. In this case(Buyer only) the `lastOffer` won't reflect the real offer sent back by the Partner. [reference](https://github.com/artsy/eigen/blob/c0f23dff4b92c8a1f10f13500b87f1de2859044a/src/lib/Scenes/Inbox/Components/Conversations/OrderInformation.tsx#L31) also the order page contains the same logic as this reference

src/v2/Apps/Order/Components/ShippingAddress.tsx
```diff
@@ -19,19 +20,26 @@ export const ShippingAddress = ({
     country,
     phoneNumber,
   },
+  textColor,
```
araujobarret, line 23: nit you can define the default here saving the ternary operators down below, improves readability

src/v2/Apps/Order/Components/CreditCardDetails.tsx
```diff
@@ -6,18 +6,20 @@ export const CreditCardDetails = ({
...
   expirationYear,
   responsive = true,
+  textColor,
 }: {
   brand: string
   lastDigits: string
   expirationMonth: number
   expirationYear: number
   responsive?: boolean
+  textColor?: string
```
araujobarret, line 16: ditto regarding default value

araujobarret (changes requested):
Overall looks good but as I mentioned some cases are missing, in our case we can't cover only BUYER from OrderConnection and also the message box depending on the status of the offer flow.

src/v2/Apps/Conversation/__tests__/DetailsSidebar.jest.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { setupTestWrapperTL } from "v2/DevTools/setupTestWrapper"
+import { screen } from "@testing-library/react"
+import { DetailsSidebarFragmentContainer } from "../Components/DetailsSidebar"
+import { DetailsSidebar_Test_Query } from "v2/__generated__/DetailsSidebar_Test_Query.graphql"
+
+jest.unmock("react-relay")
+
+const { renderWithRelay } = setupTestWrapperTL<DetailsSidebar_Test_Query>({
+  Component: props => {
+    if (!props.node) return null
```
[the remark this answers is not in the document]
lilyfromseattle: I get this error when I do that ```Type 'Container<DetailsProps>' is not assignable to type 'ComponentType<DetailsSidebar_Test_QueryResponse>'.

jo-rs:
Alignment looks off, no? 🤔 
Screenshot 2021-12-24 at 02 28 24

src/v2/Apps/Order/Utils/getStatusCopy.tsx
```diff
@@ -0,0 +1,320 @@
...
+      break
+    case "Standard":
+      daysToDeliver = "3-5 business days"
+  }
+  return daysToDeliver
+    ? `Your order will be delivered in ${daysToDeliver} once shipped, plus up to 7 days processing time. `
+    : null
+}
+
+export const trackingInfo = (
```
araujobarret, line 303: nit: we don't need the export here

araujobarret (dismissed):
Looks good, thanks for the changes!

lean8086 (changes requested):
When I choose "Arrange for pickup" on the shipping step of the checkout while making an offer, then I see a comma on the Ship To section of the sidebar. It can be related to the ShippingAddress component always expecting to receive an address?

Update: [this is the use case I believe is missing](https://github.com/artsy/force/pull/9049#discussion_r778799604).

src/v2/Apps/Conversation/Components/DetailsSidebar.tsx
```diff
@@ -197,82 +211,164 @@ export const Details: FC<DetailsProps> = ({
...
-          <Box px={2}>
-            <Text variant="xs" mb={2}>
+          <TransactionDetailsSummaryItemFragmentContainer
+            order={activeOrder}
+            showOrderNumberHeader
+          />
+          <StackableBorderBox>
+            <Box>
+              <Text variant="md" mb={2}>
+                Ship To
```
lean8086, line 226: Question: what should happen to this entire block if the user chooses pick up instead of providing a shipping address? I couldn't find the use case on the figma file.

Also, a small detail based on the design. Would it make sense to have it this way?:
```suggestion
                Ship to
```

lean8086: I spoke with Kseniya to find the use case in the figma file, and this is the one that I think it's missing here:

image

araujobarret: Good catch!
I was about to say it, we use the same idea in Eigen. To double-check you can visit the orders page, which should be like how it is there.

lean8086 (approved):
Looks great. Thanks for the changes! 🚀

PR #9053, opened 2021-12-13
dep(jest): Upgrade jest

araujobarret:
This PR first intention was to fix the missing autocomplete from `jest` assertions but was not possible unless `jest` was upgraded too. That being said, this upgrade has broken a lot of tests that had to be fixed and a few ones refactored with RTL.

List of changes
- Fixes autocomplete from `expect` assertions
- Refactors a few tests to use RTL because they were broken after the upgrade
- Introduces an idea of how to test Relay refetches(based on official docs)
- Removes jest-dom hack
- Other minor adjusts to make things work


__mocks__/react-relay.js
```diff
@@ -1,4 +1,4 @@
-let React = require.requireActual("react")
+let React = jest.requireActual("react")
```
araujobarret, line 1: `require.requireActual` is no longer available anymore

jest.config.v2.js
```diff
@@ -14,8 +14,7 @@ module.exports = {
       },
       reporters: ["default", "jest-junit"],
       setupFilesAfterEnv: ["<rootDir>/src/v2/jest.envSetup.ts"],
-      // JSDom pinned for React Testing Library -- see HACKS.md
-      testEnvironment: "jest-environment-jsdom-thirteen",
+      testEnvironment: "jest-environment-jsdom",
```
araujobarret, line 17: we don't need the hacked environment anymore

src/desktop/apps/armory_week/fixture.ts
```diff
@@ -0,0 +1,103 @@
+export default {
```
araujobarret, line 1: had to change the fixture from json to ts, for some reason wasn't working, was an easy-go solution

src/desktop/apps/auction/components/layout/active_bids/__tests__/MyActiveBids.jest.js
```diff
@@ -5,7 +5,7 @@ const { Auction } = require("desktop/models/auction")
 const CurrentUser = require("desktop/models/current_user")
 
 // FIXME: Add required props
-xdescribe("apps/auction/components/layout/active_bids/MyActiveBids", () => {
+describe.skip("apps/auction/components/layout/active_bids/MyActiveBids", () => {
```
araujobarret, line 8: xdescribe, xit should be changed to describe.skip and it.skip

src/desktop/apps/consign/__tests__/fixture.ts
```diff
@@ -0,0 +1,157 @@
+export default {
```
araujobarret, line 1: ditto about the fixture change from json to ts

src/v2/Apps/Artist/Routes/AuctionResults/__tests__/ArtistAuctionResults.jest.tsx
```diff
@@ -1,14 +1,11 @@
-/* eslint-disable jest/no-done-callback */
```
araujobarret, line 1: Converted to RTL since became broken with newer jest

src/v2/Apps/Artist/Routes/AuctionResults/__tests__/ArtistAuctionResults.jest.tsx
```diff
@@ -75,84 +55,121 @@ describe("AuctionResults", () => {
...
-      checkboxes.at(2).simulate("click")
-      setTimeout(() => {
-        // expect no new call
-        expect(openAuthModal).toHaveBeenCalledTimes(1)
-        done()
+      expect(operationVariables.categories).toContain("Work on Paper")
+      expect(openAuthModal).toHaveBeenCalledTimes(1)
+
+      fireEvent.click(checkboxes[2])
+      act(() => {
```
araujobarret, line 126: instead of spying relay.refetch prop this is the way we should observe for new relay operations, if we don't receive a second one will throw an error. With this approach, we're able to check changes from the relay query, in this case if the query was updated with the selected filter.

src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarArtists.jest.tsx
```diff
@@ -24,51 +23,51 @@ describe("ArtworkSidebarArtists", () => {
...
-    beforeEach(async () => {
-      wrapper = await getWrapper()
-    })
-
     it("displays artist name for single artist", () => {
-      expect(wrapper.html()).toContain("Josef Albers")
-      expect(
-        wrapper.find({ href: "/artist/josef-albers" }).hostNodes().length
-      ).toBe(1)
+      renderWithRelay({
```
araujobarret, line 45: Upgraded to RTL too

src/v2/Apps/Auction/Routes/__tests__/ConfirmBid.jest.tsx
```diff
@@ -182,17 +182,14 @@ describe("Routes/ConfirmBid", () => {
...
-        expect(mockBidderPositionQuery.mock.calls[0][1]).toEqual({
-          bidderPositionID: "pending-bidder-position-id-from-polling",
-        })
-
-        expect(window.location.assign).toHaveBeenCalledWith(
-          `/artwork/${ConfirmBidQueryResponseFixture.artwork?.slug}`
-        )
-        done()
-      }, 1001)
+      await new Promise(resolve => setTimeout(resolve, 1001))
```
araujobarret, line 185: I don't like using timers in this case, I tried to use jest timers fn but wasn't working and was consuming so much time. Refactored this part because by default jest doesn't allow anymore done unless you disable the linting rule

src/v2/Apps/FeatureAKG/__tests__/FeatureAKGRoute.jest.tsx
```diff
@@ -79,6 +79,7 @@ describe("FeatureAKG", () => {
         trackEvent,
       }
     })
+    jest.setTimeout(10000)
```
araujobarret, line 82: This test became flaky and timed out a fewer times, I increased the timeout of it but in the future would be better to refactor it, but more than 1k lines in a single test file and I have no context about this AKG guy

src/v2/Apps/Order/Components/__tests__/PaymentPicker.jest.tsx
```diff
@@ -194,6 +194,13 @@ describe("PaymentPickerFragmentContainer", () => {
     _mockStripe().createToken.mockImplementation(() =>
       Promise.resolve({ error: "bad error" })
     )
+    env.clearErrors()
```
araujobarret, line 197: jest blocked using his hooks outside of test files, the easy refactor was to shift this control to the tests

src/v2/Components/ArtworkFilter/__tests__/ArtworkFilter.jest.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { useTracking } from "v2/System/Analytics/useTracking"
 import { ArtworkFilter, BaseArtworkFilter } from "v2/Components/ArtworkFilter"
-import { MockBoot, renderRelayTree } from "v2/DevTools"
+import { MockBoot } from "v2/DevTools"
 import { renderToString } from "v2/DevTools/__tests__/MockRelayRendererFixtures"
 import { ArtworkQueryFilter } from "../ArtworkQueryFilter"
 import { ArtworkFilterFixture } from "./fixtures/ArtworkFilter.fixture"
-import { Pagination } from "v2/Components/Pagination"
 import { initialArtworkFilterState } from "../ArtworkFilterContext"
+import { setupTestWrapperTL } from "v2/DevTools/setupTestWrapper"
```
araujobarret, line 9: another test that became broken, so refactored to use RTL

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -116,7 +116,12 @@ const FiltersWithScrollIntoView: React.FC<{
   const scrollRef = useRef<HTMLDivElement | null>(null)
 
   return (
-    <Box ref={scrollRef as any} overflowY="scroll" height="100%">
+    <Box
+      ref={scrollRef as any}
+      overflowY="scroll"
+      height="100%"
+      data-testid="FiltersWithScrollIntoView"
```
araujobarret, line 123: was impossible to query it from the tests with RTL without the testid, it's the only non-dev change

src/v2/DevTools/__tests__/renderUntil.jest.tsx
```diff
@@ -1,5 +1,6 @@
 import { mount } from "enzyme"
 import * as React from "react"
+import { setImmediate } from "timers"
```
araujobarret, line 3: setImmediate doesn't come as global anymore in jest 26 onwards

src/v2/DevTools/createTestEnv.tsx
```diff
@@ -78,26 +78,6 @@ class TestEnv<MutationNames extends string, TestPage extends RootTestPage> {
       {} as any
     )
 
-    beforeEach(() => {
-      this.errors = []
-    })
-
-    afterEach(() => {
```
araujobarret, line 85: not possible to use out of test files anymore, shifted the responsibility to the callers to control from the available methods

src/v2/DevTools/setupTestWrapper.tsx
```diff
@@ -68,14 +72,22 @@ type SetupTestWrapper<T extends OperationType> = {
...
+type RenderWithRelay = RTLRenderResult & { env: RelayMockEnvironment }
 export const setupTestWrapperTL = <T extends OperationType>({
   Component,
   query,
   variables = {},
 }: SetupTestWrapper<T>) => {
-  const renderWithRelay = (mockResolvers: MockResolvers = {}) => {
+  const renderWithRelay = (
+    mockResolvers: MockResolvers = {},
+    manualEnvControl?: boolean
```
araujobarret, line 88: I added the option to control the env from the caller, when we need to check for refetches we can't in the current structure and exporting the env alone runs into racing conditions to mock the operations.

src/v2/jest.envSetup.ts
```diff
@@ -83,50 +83,56 @@ if (process.env.ALLOW_CONSOLE_LOGS !== "true") {
...
-        }
 
-        jest.spyOn(console, type).mockImplementation(handler)
-      }
-    })
-    done() // it is important to call this here or every test will timeout
+          jest.spyOn(console, type).mockImplementation(handler)
+        }
+      })
+      resolve(null)
```
araujobarret, line 135: transformed into a promise to remove the done control

__mocks__/react-relay.js
damassi: Interesting that they removed this! I wonder why.

src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarArtists.jest.tsx
damassi: Sweet 🙏

src/v2/DevTools/setupTestWrapper.tsx
damassi: @araujobarret - if there's an example of this somewhere in the codebase min posting back here in a comment? That way there's a bit of an example / papertrail that other devs can refer to

src/v2/DevTools/setupTestWrapper.tsx
```diff
@@ -94,11 +106,13 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
-      return MockPayloadGenerator.generate(operation, mockResolvers)
-    })
+    if (!manualEnvControl) {
+      env.mock.resolveMostRecentOperation(operation => {
+        return MockPayloadGenerator.generate(operation, mockResolvers)
+      })
+    }
 
-    return view
+    return { ...view, env }
```
damassi, line 115: Surprised this return change didn't break other tests. I guess this is because RTL uses global state via imports to assert against screen content, versus enzyme. where it returns an instance of the wrapper?

damassi (approved):
So psyched on this! Thanks @araujobarret 💯 

(Also curious to see how this might impact my Relay upgrade work. Mergin')

package.json
```diff
@@ -353,9 +354,9 @@
     "imagesloaded": "4.1.2",
     "immer": "9.0.7",
     "inquirer": "1.2.3",
-    "jest": "24.9.0",
+    "jest": "27.3.1",
```
pvinis, line 357: nice

src/desktop/apps/auction/components/layout/active_bids/__tests__/MyActiveBids.jest.js
pvinis: weird, is that a jest thing? they stopped supporting xit?

src/desktop/components/clock/test/view.test.js
```diff
@@ -124,7 +124,7 @@ describe("ClockView", function () {
   })
 
   // TODO: components/clock/test/view.coffee
-  return xdescribe("#stateCallback", function () {
+  return describe.skip("#stateCallback", function () {
```
pvinis, line 127: wow ive never seen a `describe` on a return!

src/v2/DevTools/renderUntil.tsx
```diff
@@ -1,5 +1,6 @@
 import { ReactWrapper, RenderUntilPredicate, mount } from "enzyme"
-import * as React from "react";
+import * as React from "react"
```
pvinis, line 2: is this `* as` needed anymore?

src/v2/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -151,6 +151,9 @@ describe("Shipping", () => {
     TestPage: ShippingTestPage,
   })
 
+  beforeEach(hooks.clearErrors)
+  afterEach(hooks.clearMocksAndErrors)
```
pvinis, line 155: these are great!

pvinis (commented):
wow very nice!

src/v2/DevTools/renderUntil.tsx
damassi: This was a codemod addition when we added auto react imports. For some reason it included a bunch of these (and semis) in certain cases and so we've been getting diffs where it removes the semi.

araujobarret:
@damassi did you squash & merge? I'm asking bc of 
image
if you did maybe I should fix the commit tree that became messy after a couple of stale time of my local branch 😓

src/desktop/components/clock/test/view.test.js
araujobarret: me neither, I only fixed what jest complained, had to spend time in the other refactors 😅

src/desktop/apps/auction/components/layout/active_bids/__tests__/MyActiveBids.jest.js
araujobarret: [they still support](https://jestjs.io/docs/api#describeskipname-fn) but now jest breaks complaining to use skip instead, so the fix was to change instead of suppressing the warning

src/v2/DevTools/setupTestWrapper.tsx
araujobarret: @damassi this was the first idea introduced here, before with enzyme we were doing the strange [spy thing](https://github.com/artsy/force/blob/main/src/v2/Apps/Fair/Components/__tests__/FairBooths.jest.tsx#L62-L69)

The approach I did here I based on the [docs](https://relay.dev/docs/guides/testing-relay-components/#refetch-component) how to test refetch case, but in my case here I only cared about the `operation.request`, I didn't remock the fixture based on the filter change, maybe we can refactor another test(without a big fixture 🙏 ) with RTL + refetch test idea and put in the docs as reference, I don't know a good one to do it, maybe this `FairBooths`

damassi:
I did not squash and merge! I just hit the green merge button. I think its ok! Squash and merge is mostly a guideline to follow.

HACKS.md
```diff
@@ -14,22 +14,6 @@ Tell us when we can remove this hack.
 
 > 👀 When adding a new item, see comment on top of file for template.
 
-## Jest's JSDom pinned to v13
```
anandaroop, line 17: 🤩 Nice to see a `HACKS` document actually shrink in size!

Thanks for cleaning this up and getting us over this hump.

PR #9055, opened 2021-12-13
feat(FX-3641): Show Create Alert button when user both logged in and out

anastasiapyzhik:
[FX-3641]

At the moment Create an Alert button is visible on artworks grid page for logged in users and hidden for logged out users. Instead, the button should be visible for both logged in and logged out users.

### Acceptance Criteria

- Create an Alert button is visible regardless of whether user is logged in or logged out
- If a user that is logged in clicks the button they see the Create an Alert modal
- If a user that is logged out clicks the button they see the standard login screen pop up where they are prompted to login/create account

Also, after successful auth Create Alert modal pops up without additional button click.

### Demo

https://user-images.githubusercontent.com/44819355/146196258-5e4aef51-c913-4f71-997a-27453c97fdcc.mov

[FX-3641]: https://artsyproduct.atlassian.net/browse/FX-3641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
+++ b/src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
@@ -10,7 +10,7 @@ import {
 } from "@artsy/cohesion"
 import { SavedSearchAlertModal } from "v2/Components/SavedSearchAlert/SavedSearchAlertModal"
 import { SavedSearchAlertMutationResult } from "v2/Components/SavedSearchAlert/SavedSearchAlertModel"
-import { openAuthToFollowSaveCreate } from "v2/Utils/openAuthModal"
+import { openAuthToSatisfyIntent } from "v2/Utils/openAuthModal"
 import { mediator } from "lib/mediator"
 
 interface CreateAlertButtonProps extends ButtonProps {
@@ -45,7 +45,7 @@ export const CreateAlertButton: React.FC<CreateAlertButtonProps> = ({
     if (isLoggedIn) {
       handleOpenForm()
     } else {
-      openAuthToFollowSaveCreate(mediator, {
+      openAuthToSatisfyIntent(mediator, {
         entity: {
           name: savedSearchAttributes.name,
           slug: savedSearchAttributes.slug,
--- a/src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
+++ b/src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
@@ -4,14 +4,14 @@ import { SavedSearchAttributes } from "../types"
 import { useSystemContext, useTracking } from "v2/System"
 import {
   ActionType,
+  AuthIntent,
   ContextModule,
-  Intent,
   PageOwnerType,
 } from "@artsy/cohesion"
 import { SavedSearchAlertModal } from "v2/Components/SavedSearchAlert/SavedSearchAlertModal"
 import { SavedSearchAlertMutationResult } from "v2/Components/SavedSearchAlert/SavedSearchAlertModel"
-import { openAuthModal } from "v2/Utils/openAuthModal"
-import { ModalType } from "v2/Components/Authentication/Types"
+import { openAuthToFollowSaveCreate } from "v2/Utils/openAuthModal"
+import { mediator } from "lib/mediator"
 
 interface CreateAlertButtonProps extends ButtonProps {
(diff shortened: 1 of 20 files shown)


src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
```diff
@@ -27,6 +41,21 @@ export const CreateAlertButton: React.FC<CreateAlertButtonProps> = ({
...
+    if (isLoggedIn) {
+      handleOpenForm()
+    } else {
+      openAuthToFollowSaveCreate(mediator, {
+        entity: {
+          name: savedSearchAttributes.name,
+          slug: savedSearchAttributes.slug,
+        },
+        contextModule: ContextModule.artworkGrid,
+        intent: "createAlert" as AuthIntent,
```
anastasiapyzhik, line 54: the intent for this action is not added to cohesion yet, left comment for @louislecluse in jira ticket about this

dblandin: Just created a PR for that here: https://github.com/artsy/cohesion/pull/274

src/v2/Utils/openAuthModal.ts
```diff
@@ -13,24 +13,28 @@ export interface AuthModalOptions extends ModalOptions {
...
     }
   }, 100)
 }
 
-const authModalReady = (mediator: Mediator): boolean => {
-  return mediator.ready("open:auth")
-}
-
-export const openAuthToFollowSave = (
+export const openAuthToFollowSaveCreate = (
```
dblandin, line 37: **question**: I'm struggling to understand this function. Does this handle the open auth behavior for following, saving, and creating alerts? Is there a way to extract the create alert tracking or maybe we just use `openAuthModal` directly?

dblandin (commented):
Looking good! Left a comment about the `openAuthToFollowSaveCreate` function.

src/v2/Components/ArtworkFilter/SavedSearch/Components/CreateAlertButton.tsx
anastasiapyzhik: @dblandin thanks for prompt action!

src/v2/Utils/openAuthModal.ts
anastasiapyzhik: It handles the open auth behavior for following artist/partner/gene, saving artwork, and creating an alert by customizing options for follow up actions depending on passed intent. We can pass necessary options directly to `openAuthModal`, but I think this logic is similar to two others, + it handles redirecting user to _/login_ in case opening auth modal is not handled properly for some reason.

What do you think of this?

src/v2/Components/ArtworkFilter/__tests__/CreateAlertButton.jest.tsx
```diff
@@ -0,0 +1,121 @@
...
+      expect(openAuthToFollowSaveCreate).toHaveBeenCalledWith(mediator, {
+        entity: {
+          name: "test-artist-name",
+          slug: "example-slug",
+        },
+        contextModule: "artworkGrid",
+        intent: "createAlert",
+      })
+    })
+  })
```
dimatretyak, line 120: 👏

dimatretyak (dismissed):
LGTM 👍

src/v2/Utils/openAuthModal.ts
dblandin: Maybe in that case we can rename the function to something more generic like `openAuthToSatisfyIntent` if the function is able to figure out which intent is active and respond accordingly?

anastasiapyzhik: Good idea!

dblandin (dismissed):
👏🏼

dimatretyak (approved):

src/v2/Utils/Hooks/useAuthIntent/index.ts
```diff
@@ -49,6 +52,11 @@ export const runAuthIntent = async (
   try {
     await (() => {
       switch (value.action) {
+        case "createAlert":
+          if (value.kind === "artist") {
+            triggerEvent(mediator, "auth:login:success")
```
dimatretyak, line 57: maybe it's better to rename it to `alert:create` or `alert:create_modal:open`?

PR #9060, opened 2021-12-14
fix: unselecting price filter option

tr-ann:
### Description
When user unselects price filter option the `null` value is passed to filter state and it causes error in `extractFiltersPills` helper

### Demo
Before

https://user-images.githubusercontent.com/56556580/146011862-6b5888b9-0ad5-4b96-a46c-5e0788d4f6a6.mov

After

https://user-images.githubusercontent.com/56556580/146012269-891afa6e-8d84-4cd4-8ef6-e0219861a69a.mov


dimatretyak (approved):
LGTM. Thank you for this changes 👍

dblandin (approved):
Awesome. Thank you for the before/after videos!!

PR #9064, opened 2021-12-14
fix(partner): improves image quality

dzucconi:
We recently added a `normalized` version to these images; but for whatever reason these were mostly favoring `large`. I should make an update in Force to do this by default but for now this fixes the quality issues for these.

(Also really need to prioritize the V3 migration of this app)


dzucconi:
A note that the `Merge On Green` label did not, in fact, merge on green.

PR #9065, opened 2021-12-14
feature: Prevent Server Side Code from Loading Client Side

icirellik:
**Description**

This change relocates one of the server-side middlewares to prevent legacy
files and their dependencies from being accidentally shipped to the client. This 
one middleware includes over 100 legacy files and their corresponding 
dependencies that are not needed on the client. 

In addition, it may be worth further investigation into how we can automatically 
isolate all modules that would be included when using the new `onServerSideRender` 
functionality as this can accidentally include all server-side dependencies on the client.

The bundle size decreases from 2.88 MiB (gzip) to 2.5 MiB (gzip)


damassi (changes requested):
The pattern we've migrated to is using the `onServerSideRender` hook within our routes. Looking in `handleArtworkImageDownload` all that we're importing is `superagent` and the artwork model, two common libs throughout our client-side code. It's an acceptable tradeoff for colocation, and if we can move those to the common bundle if its that big of a deal or find a way of factoring them out via `isomorphic-fetch`.

damassi:
Another way to fix without replacing superagent with fetch api is to simply move that import down into the server-side function:
```tsx
  if (artwork.isDownloadable(req.user)) {
    const request = require('superagent') 
    ...
  }
```

icirellik:
The colocation debate is very much a matter of perspective and not really worth debating, in many cases such as this moving server-side code in this manner makes it harder to find during development. That aside, as a team, we need to be far more careful when reaching into the legacy stack from the new application. In this case, pulling in this single file includes over 100 legacy files and all of their dependencies.

If it were only one dependency and a single model this would be an acceptable tradeoff. Additionally, given what this middleware does and the large amount of legacy code that comes with it, it should be easy and quick enough for someone with more React experience than myself to rebuild it.

We should either move it back or rewrite it.

damassi:
If you're not willing to make the suggested change then i'll open a separate PR and we can close this one.

icirellik:
PR #9066 doesn't prevent the modules from being bundled or loaded client-side.

damassi:
Then we'll follow-up with replacing superagent with isomorphic fetch.

PR #9068, opened 2021-12-14
feat(GRO-664): styles one trust cookie consent banner 

dzucconi:
Draft for a bit until we resolve a few questions; but basically: this does allow us to style the banner using our themed values.

I left a comment on the actual code but yeah, this thing would be funny if it wasn't so sad.

Since we're outside of react we don't actually have access to the 'current theme' so we have to pull values out of the static theme object. Not a huge deal.

The main question I have is: do we merge this or do we just use this to 'develop' the style sheet then extract that and paste it into the CSS override in OneTrust's web interface. The main reason I ask this is that this seems like a big ugly chunk of CSS to load when it won't apply most of the time.

https://user-images.githubusercontent.com/112297/146075365-fad6ebd5-f434-43db-9101-35ca194fac46.mov

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
@@ -27,6 +27,7 @@ export const SIZES = [
   { displayName: "Medium (40 – 100cm)", name: "MEDIUM" },
   { displayName: "Large (over 100cm)", name: "LARGE" },
 ]
+const ONE_IN_TO_CM = 2.54
 
 type CustomRange = (number | "*")[]
 
@@ -35,7 +36,9 @@ type CustomSize = {
   width: CustomRange
 }
 
-const convertToCentimeters = (element: number) => Math.round(element * 2.54)
+const convertToCentimeters = (element: number) => {
+  return Math.round(element * ONE_IN_TO_CM)
+}
 
 export const parseRange = (range?: string) => {
   return range?.split("-").map(s => {
@@ -49,7 +52,7 @@ const convertRangeElementToInches = (element: number | "*") => {
     return element
   }
 
-  return Math.round((element / 2.54) * 100) / 100
+  return element / ONE_IN_TO_CM
 }
 
 const convertRangeToInches = (range: CustomRange) => {
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter.jest.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter.jest.tsx
@@ -100,8 +100,12 @@ describe("SizeFilter", () => {
     userEvent.click(screen.getByText("Set size"))
 
     expect(currentContext().filters?.sizes).toEqual([])
-    expect(currentContext().filters?.height).toEqual("4.72-6.3")
-    expect(currentContext().filters?.width).toEqual("4.72-6.3")
+    expect(currentContext().filters?.height).toEqual(
+      "4.724409448818897-6.299212598425196"
+    )
+    expect(currentContext().filters?.width).toEqual(
+      "4.724409448818897-6.299212598425196"
+    )
   })
 
   it("updates the filter values when only one dimension is added", () => {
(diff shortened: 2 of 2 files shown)


src/v2/System/Router/Boot.tsx
```diff
@@ -83,6 +84,7 @@ export const Boot = track(undefined, {
                     <ToastsProvider>
                       <Grid fluid maxWidth="100%">
                         <GlobalStyles />
+                        <OneTrustConsentStyles />
```
damassi, line 87: I wonder how consent works. Is it something that we set on the user model? Or is it all 3rd party processed? Because if there's a place we check then this can be conditionally loaded via a dynamic import on a need-to basis.

dzucconi: Right, which is why I'm thinking we should just copy out the compiled styles and not actually render this component in prod. It's partially geo, and then presumably there's a cookie (which I guess we could check that)

src/v2/Components/OneTrustConsentStyles.ts
```diff
@@ -0,0 +1,189 @@
...
+import { BUTTON_SIZES, BUTTON_VARIANTS, DROP_SHADOW } from "@artsy/palette"
+
+const toStyle = (style: Record<string, string | number | undefined>) => {
+  return Object.entries(style)
+    .map(([key, value]) => {
+      const property = key.replace(/[A-Z]/g, m => `-${m.toLowerCase()}`)
+      return `${property}: ${value} !important;`
+    })
+    .join("")
+}
```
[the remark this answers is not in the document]
jonallured: LOL I thought when you posted a screenshot of this that it was from someone else's code. 😝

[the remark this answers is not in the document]
dzucconi: We're fighting fire with fire here

src/v2/System/Router/Boot.tsx
jonallured: I would assume _some_ need to change this stuff but not a lot of changes. Meaning I wouldn't want to "throw away" the code that generates this stylesheet even if that's what we end up with in production.

dzucconi:
Now with modal: 

I left the toggle switches alone, which are pretty ugly, but probably more trouble than they are worth.

https://user-images.githubusercontent.com/112297/146452184-ecca3655-6af9-49c0-a6f3-916eba8d30d5.mov

dzucconi:
So once you accept they add a cookie `OptanonConsent` which contains some things like a timestamp and a `consentId`.

I suspect reverse engineering this to conditionally include the styles would be error-prone/brittle.

I'm kind of thinking we just merge this for now and maybe come back to it to extract the styles once it's launched and 'settled'.

jonallured:
> merge this for now

yeah seems legit 👍

src/v2/Components/OneTrustConsentStyles.ts
```diff
@@ -0,0 +1,472 @@
...
+          }
+
+          &:last-child {
+            ${toStyle({ marginTop: THEME.space[1], marginLeft: 0 })},
+          }
+        }
+      }
+    }
+  }
+`
```
damassi, line 472: dude lol

damassi: I'm so sorry 😦

damassi (approved):
🚀

dzucconi:
Woops didn't realize this was failing

dzucconi:
Very odd, this is causing the specs to time out. Maybe if I add a global mock.

src/v2/DevTools/MockBoot.tsx
```diff
@@ -1,8 +1,10 @@
 import { Boot } from "v2/System/Router"
-import * as React from "react";
+import * as React from "react"
 import { Breakpoint } from "v2/Utils/Responsive"
 import { buildClientAppContext } from "desktop/lib/buildClientAppContext"
 
+jest.mock("v2/Components/OneTrustConsentStyles")
```
damassi, line 6: ❇️

dzucconi:
I... did not merge this. Not sure what happened.

PR #9069, opened 2021-12-14
chore: replace superagent with fetch

damassi:
Follow-up to https://github.com/artsy/force/pull/9066 

Replaces use of superagent with fetch-api which we use all over.


damassi:
@icirellik - Feel free to extract [this model reference](https://github.com/artsy/force/blob/0f4afbfb1df31aaccef0db10faaa2a0458038c92/src/desktop/models/artwork.js#L71-L77) or file a ticket with grow and we can handle it.

PR #9079, opened 2021-12-15
chore: Wrap lot image and title in a link back to the artwork page

annacarey:
Collab with @nspinazz89 on [BID-121](https://artsyproduct.atlassian.net/browse/BID-121)

Wraps the lot image and lot #/title/date/artists in a link that redirects to the artwork page.

I wanted to link the Artsy logo back to the artwork page too, but it seemed to be used in Jade templates (!?!?!?!) so I think it's OK to leave as is?
Screen Shot 2021-12-15 at 5 19 19 PM

### DIFF
--- a/src/v2/Apps/Auction/Components/LotInfo.tsx
+++ b/src/v2/Apps/Auction/Components/LotInfo.tsx
@@ -3,6 +3,7 @@ import { LotInfo_artwork } from "v2/__generated__/LotInfo_artwork.graphql"
 import { LotInfo_saleArtwork } from "v2/__generated__/LotInfo_saleArtwork.graphql"
 import * as React from "react"
 import { RelayProp, createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Auction/Components/LotInfo.tsx
```diff
@@ -16,28 +17,32 @@ export const LotInfo: React.FC<Props> = ({ artwork, saleArtwork }) => {
     counts: { bidderPositions: bidCount },
   } = saleArtwork
   return (
-    <Flex py={4}>
+    <Flex id="our-parent-flex-box" py={4}>
```
damassi, line 20: Why add an `id` here?

annacarey: Oh whoops that was so we could see it in the UI. thanks for catching that. will remove

PR #9083, opened 2021-12-16
fix(SWA-147): Clickable area for Back button is too large

Serge0n:
Jira Ticket:ticket: : [SWA-147](https://artsyproduct.atlassian.net/browse/SWA-147)

This PR fixes the clickable area for the Back button in the Submission flow. Also, [noticed that there are new rules for the use `sharify`](https://github.com/artsy/force/blob/e5c72b0bfb6dfca04f8273e89dbe52c8e958c6bb/.eslintrc.js#L114), so i fixed it too

**VIDEO:clapper: :**

https://user-images.githubusercontent.com/55637696/146342597-aa238744-01ba-48e2-93d9-e42d9c0f830d.mov

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
@@ -14,7 +14,7 @@ import { useSubmission } from "../Utils/useSubmission"
 import { contactInformationValidationSchema } from "../Utils/validation"
 import { BackLink } from "v2/Components/Links/BackLink"
 import { useErrorModal } from "../Utils/useErrorModal"
-import { data as sd } from "sharify"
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
```diff
@@ -82,7 +82,7 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
           relayEnvironment,
           submission,
           user?.id,
-          !isLoggedIn ? sd.SESSION_ID : undefined
+          !isLoggedIn ? getENV("SESSION_ID") : undefined
```
iskounen, line 85: **question (non-blocking):** Why was this change needed? (Are we deprecating sharify?)

iskounen (approved):

PR #9084, opened 2021-12-16
fix(FX-3603): remove extra offset for exhibitors nav

dimatretyak:
Type: **Fix**
Jira ticket: [FX-3603]

### Before - Web
https://user-images.githubusercontent.com/3513494/146403988-61d45df9-f969-4602-989c-91a4b880537b.mp4

### After - Web
https://user-images.githubusercontent.com/3513494/146404016-4e17ad07-2556-42f7-9e2f-2934acc0d241.mp4

### Before - mWeb
https://user-images.githubusercontent.com/3513494/146404395-7d7724f0-bcf0-4e56-adb1-ff32d95ff59a.mp4

### After - mWeb
https://user-images.githubusercontent.com/3513494/146404423-14f7de89-ed3f-4890-a146-be49d6af702b.mp4

[FX-3603]: https://artsyproduct.atlassian.net/browse/FX-3603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Fair/Routes/FairExhibitors.tsx
+++ b/src/v2/Apps/Fair/Routes/FairExhibitors.tsx
@@ -1,12 +1,8 @@
 import React, { useEffect } from "react"
-import { Box, DROP_SHADOW, FullBleed, Spacer, Text } from "@artsy/palette"
+import { Box, Spacer, Text } from "@artsy/palette"
 import { graphql, createFragmentContainer } from "react-relay"
 import { FairExhibitors_fair } from "v2/__generated__/FairExhibitors_fair.graphql"
 import { FairExhibitorsGroupFragmentContainer as FairExhibitorsGroup } from "../Components/FairExhibitors"
-import { ExhibitorsLetterNavFragmentContainer as ExhibitorsLetterNav } from "../Components/ExhibitorsLetterNav"
-import { Sticky } from "v2/Components/Sticky"
-import { AppContainer } from "v2/Apps/Components/AppContainer"
-import { HorizontalPadding } from "v2/Apps/Components/HorizontalPadding"
 import { getExhibitorSectionId } from "../Utils/getExhibitorSectionId"
 import { useRouter } from "v2/System/Router/useRouter"
 import { useExhibitorsTabOffset } from "../Utils/useExhibitorsTabOffset"
@@ -34,18 +30,6 @@ const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair }) => {
   return (
     <>
       <Spacer mt={4} />
-
-      <Sticky>
-        {({ stuck }) => (
-          <FullBleed style={stuck ? { boxShadow: DROP_SHADOW } : undefined}>
-            <AppContainer>
-              <HorizontalPadding>
-                <ExhibitorsLetterNav fair={fair} />
-              </HorizontalPadding>
-            </AppContainer>
-          </FullBleed>
-        )}
-      </Sticky>
 
       {fair.exhibitorsGroupedByName?.map(exhibitorsGroup => {
         const { letter } = exhibitorsGroup!
@@ -78,7 +62,6 @@ export const FairExhibitorsFragmentContainer = createFragmentContainer(
           }
           ...FairExhibitorsGroup_exhibitorsGroup
         }
-        ...ExhibitorsLetterNav_fair
       }
     `,
   }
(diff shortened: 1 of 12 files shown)


dzucconi:
Why is this manually offset at all? There shouldn't be a need for that

dimatretyak:
> Why is this manually offset at all? There shouldn't be a need for that

If you mean [this part of the code](https://github.com/artsy/force/blob/3d005d2b152a7c46f4a72ee03e9b0d6642e769f2/src/v2/Apps/Fair/Utils/useExhibitorsTabOffset.ts#L11), then without this offset, when clicking on the letter in "Exhibits AZ", it will scroll with the wrong offset

dzucconi:
https://github.com/artsy/force/blob/main/src/v2/Utils/Hooks/useScrollTo.ts is sticky-aware

dimatretyak:
> https://github.com/artsy/force/blob/main/src/v2/Utils/Hooks/useScrollTo.ts is sticky-aware

@dzucconi refactored the code (now use `useScrollTo` hook instead of `scrollIntoView`)

dimatretyak:
FIY an `offset = 205` is used for "Exhibitors A-Z" in order to avoid such a problem: when clicking on a letter, it will scroll to the wrong position

https://user-images.githubusercontent.com/3513494/146422038-8953f964-5054-4d3f-836a-b74292f2e25e.mp4

src/v2/Apps/Fair/Routes/FairExhibitors.tsx
```diff
@@ -19,34 +14,22 @@ interface FairExhibitorsProps {
 const FairExhibitors: React.FC<FairExhibitorsProps> = ({ fair }) => {
   const { match } = useRouter()
   const { focused_exhibitor: focusedExhibitorID } = match.location.query
-  const offset = useExhibitorsTabOffset()
+  const { scrollTo } = useScrollTo({
+    selectorOrRef: `#jump--${focusedExhibitorID}`,
+    behavior: "smooth",
+    offset: 205,
```
dimatretyak, line 20: FIY an offset = 205 is used for "Exhibitors A-Z" in order to avoid such a problem (when clicking on a letter, it will scroll to the wrong position)


https://user-images.githubusercontent.com/3513494/146422227-b99fcc14-50e6-45b6-95ba-c5042e6454c8.mp4

dzucconi: Though that offset is more a 'margin' since it should scroll exactly to the top of the desired element.

dzucconi: If it's not then something is wrong.

dzucconi: And if something is wrong we should fix the underlying problem here because if there's a bug and it gets fixed, this 'offset' is gonna be incorrect.

dimatretyak: > Though that offset is more a 'margin' since it should scroll exactly to the top of the desired element.

Yes, it will scroll to the desired element. We just have a fair header component (highlighted in red in the screenshot) and it overlaps the element we need (`offset 205` fixes this and for this reason we use such an offset)

one x Artsy | Artsy 2021-12-16 20-59-57

dzucconi: Oh you're not using the StickyContext — that's why.

If you refactor this to use it you'll be fine.

dzucconi: To be clear: once you wrap the page in a StickyContext you can also stack stickies. The useScrollTo hook is aware of the total offset of these.

dimatretyak: hmm, although I see that the Fair screen is already wrapped in `StickyProvider` [here](https://github.com/artsy/force/blob/3140ce4ed12c740231540511ff9b8e6188075dbe/src/v2/Apps/Fair/FairApp.tsx#L72). Or is it different?

dzucconi: Ah, you're right — lemme check out the code and take a look.

dimatretyak: Discussed this moment in the DM

dblandin:
@dimatretyak Trying to understand where things ended up between the comment threads and the current diff. Would you mind recapping what you discussed over DM and where this shook out? I still see a manual offset and some calls to `scrollIntoView` vs the `useScrollTo` hook

dimatretyak:
> @dimatretyak Trying to understand where things ended up between the comment threads and the current diff. Would you mind recapping what you discussed over DM and where this shook out? I still see a manual offset and some calls to `scrollIntoView` vs the `useScrollTo` hook

There are multiple problems that prevent using only `useScrollTo` without specifying a fixed offset. Damon suggested going back to the old implementation for now and using `scrollIntoView` with specific offset

dblandin (dismissed):
Thanks for the update! LGTM 👍🏼

PR #9085, opened 2021-12-16
fix: Prevent duplicate pageview event in search GRO-653

jonallured:
This PR disables the underlying `RouterLink` click behavior in favor of letting the `SearchBar` navigate:

https://github.com/artsy/force/blob/6baaeb7283df84d1fd04c852d978acf83cb1343e/src/v2/Components/Search/SearchBar.tsx#L268-L296

Steps to reproduce this bug are in the Jira ticket!

https://artsyproduct.atlassian.net/browse/GRO-653

/cc @artsy/grow-devs


dzucconi (approved):
Ah, OK so clicks still trigger suggestion selection. 👍

jonallured:
Oh and here's the segment console that shows a LACK of duplicate pageview events:
Screen Shot 2021-12-16 at 4 01 13 PM

PR #9086, opened 2021-12-17
fix visual bugs in inbox details

erikdstock:
This PR refactors the inbox, particularly the details panel and conversation list, to address some bugs, both visual and functional.
Addresses [TX-74]

There are a few other fixes mixed in:
- addressing deprecations
- inbox header displays incorrectly at 768-900px (the sm-md breakpoint) 
- a case where the height is too small, causing a border to not extend its container
- some window size watching code was unnecessarily complex thanks to new hooks
- removed most usage of the `<Media greaterThan="md" />` component, which was the source of some flex item bugs
- removed most usage of the `media.md` style helper, which I found counterintuitive when working with complex styles
- **not fixed:** When resizing the window transitions fire off which cause the details panel to appear and immediately hide itself. After talking with @dzucconi we elected to _not_ fix this because meticulously resizing a window is not a realistic user interaction. Here is a [codesandbox with a minimal reproduction of the issue, which again, we are not fixing in this pr](https://codesandbox.io/s/responsive-css-transitions-kipy2).

Because of the complexity of this code, I moved some responsibility around - the `ConversationList` which can appear in multiple contexts now depends on its container for its positioning and other constraints.

## After

[TX-74]: https://artsyproduct.atlassian.net/browse/TX-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi:
If you set the transition on the `.show` state; and then `transition: none` for the resting state; you won't see the pop in. You'll get transitions IN but not OUT.

You could probably add the transition property on the click and then remove it on resize.

Also — I wouldn't get overly concerned with what happens when resizing the browser window. This is not something that people generally do a lot of IRL while using an application and a moment of UI oddness during the resize isn't the worst thing that could happen.

erikdstock:
Leaving this as a WIP for the holiday season. Where it's at:
- Tried setting `transition:none` only while the user is resizing the window. A basic approach to start/end debouncing in hook form (which we don't have in force) is [here](https://codesandbox.io/s/callback-forked-8ksip?file=/src/index.js) but I have failed to implement it locally - see the commit https://github.com/artsy/force/pull/9086/commits/29c58d156a98c11cd01eadda2e83eb86a737b8e2 (which is immediately reverted) for my attempt. After talking with @dzucconi we agreed this is not a must-have because real users don't typically resize the window willy-nilly.
- refactored details and DetailsContainer styled component to put all transitions in one single line for visual clarity. I can probably remove the media queries
- Still to do:
- [ ] Update the fresnel media component to use v3 breakpoints
- [ ] Maybe go back to `media.xs'styles'` instead of media queries in styled component
- [ ] double check things are displaying correctly, especially that the details header doesn't disappear at small breakpoints.

erikdstock:
I thought I had fixed this but i'm getting it again: the arrow to go back to inbox is disappearing at the higher full-screen breakpoint.
image

src/v2/Apps/Conversation/Components/ConversationHeader.tsx
```diff
@@ -1,61 +1,57 @@
...
-        />
-      </Media>
-      <Media greaterThan="sm">
-        <LargeConversationHeader
-          showDetails={showDetails}
-          setShowDetails={setShowDetails}
-          partnerName={partnerName}
-        />
-      </Media>
+      <SmallConversationHeader
```
araujobarret, line 45: I guess for this case the ideal usage would be
```jsx
<Media lessThan="md">
  <SmallConversationHeader />
</Media>
<Media greaterThanOrEqual="md">
  <LargeConversationHeader /> 
</Media>
```
then you don't need that breakpoints control inside the styled-components and I think that won't cause any visual issues like the one we had in the routes file.
Based on https://artsy.github.io/blog/2019/05/24/server-rendering-responsively/

araujobarret (approved):
Good clean-up! Left just one comment regarding responsive handling in one component, haven't found anything critical here

src/v2/Apps/Conversation/Components/ConversationHeader.tsx
erikdstock: I _think_ that this was required because of some quirk in how our `Media` component works. I'll double check before merging!

erikdstock: Ok, so I checked this today and thought it seemed fine with the Media components but now I think i remember the source of the bug- these are flex items (their parent is a `Flex`) and the styles in the actual `_ConversationHeader` components styles are not being applied as expected. I'll triple check this tomorrow before merging!

erikdstock:
Alright, tested with the media component and it seems to work fine in concert with the other styles i added. #mergeongreen.

src/v2/Apps/Conversation/Components/ConversationHeader.tsx
araujobarret: Thanks for the effort here, wasn't a major thing but good to keep consistency with responsiveness usage around!

PR #9090, opened 2021-12-17
fix(settings): accommodates users that do not have passwords setup

dzucconi:
A minor change here: users who sign up through social auth don't have passwords set initially. This adapts the password form to accommodate those users. This code is currently untested because we were unable to get social auth working at all locally. Though I believe this should work correctly, I'm going to just check this in staging and go from there.


dzucconi:
This does indeed work correctly.

PR #9091, opened 2021-12-17
fix: incorrect import causing 500

jo-rs:
Tiny fix for a `500` when trying to access the `/inquiry/:id/user_outcome` route reported by @bhoggard .
The actual error was: `TypeError: Artwork is not a constructor`, which seems to have been caused by decaffeination.
cc: @artsy/negotiate-devs

### DIFF
--- a/src/desktop/models/mixins/relations/inquiry_outcome.js
+++ b/src/desktop/models/mixins/relations/inquiry_outcome.js
@@ -5,13 +5,13 @@
  */
 export const InquiryOutcomeRelations = {
   related() {
+    const { Artwork } = require("../../artwork")
 
-    const Artwork = require('../../artwork');
+    const inquireable =
+      this.get("inquireable_type") === "Artwork"
+        ? new Artwork(this.get("inquireable"))
+        : undefined
 
-    const inquireable = this.get('inquireable_type')  === 'Artwork' ?
-      new Artwork(this.get('inquireable')) : undefined;
-
-    return this.__related__ =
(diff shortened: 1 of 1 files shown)


src/desktop/models/mixins/relations/inquiry_outcome.js
```diff
@@ -5,13 +5,13 @@
  */
 export const InquiryOutcomeRelations = {
   related() {
+    const { Artwork } = require("../../artwork")
```
jo-rs, line 8: ☝️ the actual fix.
The rest is linter.

dzucconi:
Question: how did someone get to this route at all?

jo-rs:
> Question: how did someone get to this route at all?

@dzucconi there's an email that goes out routing to this after an inquiry, I think.

dzucconi:
Ah, OK so we'll have to support this using the rebuilt inquiry flow. — There's no way just linking to the artwork page there would work the same?

jo-rs:
> Ah, OK so we'll have to support this using the rebuilt inquiry flow. — There's no way just linking to the artwork page there would work the same?

There's still some functionality there...which hopefully someone is using for something?
Screenshot 2021-12-17 at 18 07 33


Screenshot 2021-12-17 at 18 10 11

dzucconi:
Ah, fun haven't seen this before. OK thanks!

bhoggard (approved):
Thanks!

PR #9093, opened 2021-12-17
fix: validates authentication and signs out if invalid

dzucconi:
So this replicates what https://github.com/artsy/force/blob/main/src/lib/syncAuth.ts does but for the new app shell.

I do think we should actually refresh the session without signing out the user, if possible? But I'm not entirely sure what that entails.

### DIFF
--- a/src/v2/Utils/Hooks/useAuthValidation.ts
+++ b/src/v2/Utils/Hooks/useAuthValidation.ts
@@ -1,0 +1,31 @@
+import { useEffect } from "react"
+import { fetchQuery, graphql } from "react-relay"
+import { useSystemContext } from "v2/System"
+import { logout } from "../auth"
+import { useAuthValidationQuery } from "v2/__generated__/useAuthValidationQuery.graphql"
+
+export const useAuthValidation = () => {
+  const { relayEnvironment } = useSystemContext()
+
+  useEffect(() => {
+    const exec = async () => {
+      const { authenticationStatus } = await fetchQuery<useAuthValidationQuery>(
+        relayEnvironment!,
+        graphql`
+          query useAuthValidationQuery {
+            authenticationStatus
+          }
+        `,
+        {}
+      )
+
+      if (authenticationStatus === "INVALID") {
+        await logout()
+        window.location.reload()
+        return
+      }
+    }
+
+    exec()
+  }, [relayEnvironment])
+}
--- a/src/v2/Apps/Authentication/__tests__/authenticationRoutes.jest.tsx
+++ b/src/v2/Apps/Authentication/__tests__/authenticationRoutes.jest.tsx
@@ -21,6 +21,7 @@ jest.mock("../Utils/helpers", () => ({
   ...jest.requireActual("../Utils/helpers"),
   setCookies: jest.fn(),
 }))
+jest.mock("v2/Utils/Hooks/useAuthValidation")
 
 describe("authenticationRoutes", () => {
   const mockCheckForRedirect = checkForRedirect as jest.Mock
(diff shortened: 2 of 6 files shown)


src/v2/Utils/auth.ts
```diff
@@ -2,9 +2,12 @@
 // - Wraps auth related functions in consistent Promise interface
 // - Automatically handles CSRF token, session ID, ReCaptcha token
 
+// TODO: replace with getENV
```
damassi, line 5: I see `getENV` below -- comment still relevant?

src/v2/Utils/auth.ts
```diff
@@ -170,3 +173,41 @@ export const logout = async () => {
...
+ * If user is logged in with invalid authentication (401); `Promise` resolves to `Status.Invalid`
+ * All other status codes will also resolve to `Status.LoggedIn` because we don't know
+ * whether or not the authentication is valid (error could be something else)
+ */
+export const validate = async (): Promise<Status> => {
+  const currentUser = getENV("CURRENT_USER")
+
+  if (!currentUser) return Status.LoggedOut
+
+  const response = await fetch(`${getENV("API_URL")}/api/v1/me`, {
```
damassi, line 197: To simplify this, can we just go through MP? 
```tsx
const me = await fetchQuery(graphql`
  query MeQuery {
    me {
      ...
    }
  }
`)

if (!me) {
  return Status.Invalid
}
```

damassi (dismissed):
Looks 👌 

One q about leaning on MP versus a rest endpoint. Also, can we add some test coverage here? Once that's in, feel free to merge.

src/v2/Utils/auth.ts
dzucconi: Yeah, good call

src/v2/Utils/auth.ts
```diff
@@ -170,3 +173,41 @@ export const logout = async () => {
...
+    credentials: "same-origin",
+    headers: {
+      Accept: "application/json",
+      "Content-Type": "application/json",
+      "X-Requested-With": "XMLHttpRequest",
+      "X-XAPP-TOKEN": getENV("ARTSY_XAPP_TOKEN"),
+      "X-ACCESS-TOKEN": currentUser.accessToken,
+    },
+  })
```
jonallured, line 206: Not trying to be difficult here but I'm genuinely confused about when `fetch` should be used - especially given this comment:

> yah at this point [code that uses fetch] should be considered something of a code smell

https://artsy.slack.com/archives/C01ADJNCS5D/p1638998715486800?thread_ts=1638997191.482800&cid=C01ADJNCS5D

I mean I actually find this code quite readable and love it but it conflicts with that statement from @damassi - how do yall think through these issues? 🤔 

thanks!

damassi: By code-smell I meant: in v2, we get 99% of all data need from MP. If for some reason we're _not_ using MP, there should be an exceptional reason to promote that use-case to the 1%.

dzucconi: `fetch` should be used if you need to make some request outside of GraphQL. Here I could just use GraphQL, as @damassi points out; and avoid the header set up.

src/v2/Utils/auth.ts
dzucconi: Actually, in this case I don't think we're returning the status code. We could wrap up a proper check for this in Metaphysics though.

damassi: Would be worth it.

src/v2/Utils/Hooks/useAuthValidation.ts
```diff
@@ -0,0 +1,31 @@
...
+      if (authenticationStatus === "INVALID") {
+        await logout()
+        window.location.reload()
+        return
+      }
+    }
+
+    exec()
+  }, [relayEnvironment])
+}
```
dzucconi, line 31: Updated to use query

PR #9096, opened 2021-12-20
chore: Skipping rounding for custom size values

dimatretyak:
Type: **Chore**
Jira ticket: [FX-3667]

### Context
If the user specifies the same custom size values for Force and Eigen, they will be converted to inches differently. In turn, this can add problems when working with saved search alerts (for example, a user can create an alert with the same criteria)

#### Force
* width: 100 - 199 cm -> 39.37 - 78.35 in
* height: 200 - 299 cm -> 78.74 - 117.72 in

#### Eigen
* width: 100 - 199 cm -> 39.37007874015748 - 78.34645669291338 in
* height: 200 - 299 cm -> 78.74015748031496 - 117.71653543307086 in

#### Acceptance Criteria
* Values from cm to in should be converted equally to Force and to Eigen

[FX-3667]: https://artsyproduct.atlassian.net/browse/FX-3667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
@@ -27,6 +27,7 @@ export const SIZES = [
   { displayName: "Medium (40 – 100cm)", name: "MEDIUM" },
   { displayName: "Large (over 100cm)", name: "LARGE" },
 ]
+const ONE_IN_TO_CM = 2.54
 
 type CustomRange = (number | "*")[]
 
@@ -35,7 +36,9 @@ type CustomSize = {
   width: CustomRange
 }
 
-const convertToCentimeters = (element: number) => Math.round(element * 2.54)
+const convertToCentimeters = (element: number) => {
+  return Math.round(element * ONE_IN_TO_CM)
+}
 
 export const parseRange = (range?: string) => {
   return range?.split("-").map(s => {
@@ -49,7 +52,7 @@ const convertRangeElementToInches = (element: number | "*") => {
     return element
   }
 
-  return Math.round((element / 2.54) * 100) / 100
+  return element / ONE_IN_TO_CM
 }
 
 const convertRangeToInches = (range: CustomRange) => {
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter.jest.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter.jest.tsx
@@ -100,8 +100,12 @@ describe("SizeFilter", () => {
     userEvent.click(screen.getByText("Set size"))
 
     expect(currentContext().filters?.sizes).toEqual([])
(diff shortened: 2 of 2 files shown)


dzucconi:
How is this preferable? Eigen should round to two decimal places instead.

dimatretyak:
> Eigen should round to two decimal places instead

We can do rounding on the Eigen side (I did it initially), but then we will have to round to an integer when converting from in to cm (at the moment, you can enter a float value for the size filter on Eigen). And also now Force doesn't convert from in to cm quite correctly

For example, consider this example:
1. Convert 1 cm to in: `Math.round((1 / 2.54) * 100) / 100 === 0.39`
2. Convert from in to cm: `0.39 * 2.54 === 0.9906`

dzucconi:
I wouldn't expect us to convert back to the original representation so I'm not sure your example makes sense?

I'm also a little disturbed we aren't handling this in Metaphysics where we can just have an agreed upon solution for both platforms.

Also we need to be rounding to two decimal points, regardless of the conversion. Anything else is going to look like a bug in the UI.

dimatretyak:
> I wouldn't expect us to convert back to the original representation so I'm not sure your example makes sense?

For example, such logic is performed for [SavedSearchAlertModal](https://github.com/artsy/force/blob/418cc745db6fcb896ffb5ee5f6101c8297fdb1ce/src/v2/Components/SavedSearchAlert/SavedSearchAlertModal.tsx#L56) (since the values in `filterContext` are stored in `in` and we need to display it in `cm`)

Banksy - 833 Artworks for Sale on Artsy 2021-12-20 18-25-57

dzucconi:
OH: this isn't for display? This is just for filtering? So the user won't see a crazy number like `117.71653543307086`?

dimatretyak:
> OH: this isn't for display? This is just for filtering? So the user won't see a crazy number like `117.71653543307086`?

No, the user will not see "crazy" numbers. The user will continue to see integer values as they are displayed now. For the most part, the current changes are necessary in order for the backend to be able to find duplicates of previously created saved search alerts. 

Because now there is such a problem:
* User created a saved search alert with `min width = 100cm ` via Force. 100 cm on Force is **39.37 in**
* User specifies `min width = 100 cm` in Eigen and presses the "Create Alert" button. 100 cm on Eigen is **39.37007874015748 in**
* Query is executed and checks whether the user has previously created a saved search alert with such criteria
* Since in the saved search alert created via Force, **39.37** is stored as min width (and not **39.37007874015748** as we have it on Eigen), this will not be considered a duplicate and the user can create an alert with the same criteria

dzucconi:
Ah, I get it now. Thanks for bearing with. Makes sense.

dzucconi (approved):

dimatretyak:
> Ah, I get it now. Thanks for bearing with. Makes sense.

Sorry if I initially confused you and did not explain in detail. And thank you for your patience!

PR #9099, opened 2021-12-20
feat(GRO-640): Sell With Artsy banner

dzucconi:
Re: [GRO-640](https://artsyproduct.atlassian.net/browse/GRO-640)

* This doesn't implement a fallback image yet (so just black background if there's no image, though really, most artists are going to have an image).
* This shows up on all artist pages.
* <del>I was a little unclear about the linking. For P1/P2 we're supposed to link directly to the "submission flow" — what does that mean? Linking to this page: https://www.artsy.net/consign/submission/artwork-details</del>
  * <del>Does that mean we can delete the artist-specific consign pages like https://www.artsy.net/artist/banksy/consign — which will no longer be linked anywhere (I believe)?</del>
* I copied both the analytics and linking from the existing consign button.

https://user-images.githubusercontent.com/112297/146825339-a49579eb-3237-4e50-b3f6-f3896ce989e9.mov


src/v2/Apps/Artist/Routes/Overview/Components/ArtistSellWithArtsy.tsx
```diff
@@ -0,0 +1,181 @@
...
+        internalID
+        slug
+        href
+        targetSupply {
+          isInMicrofunnel
+        }
+        image {
+          resized(width: 640) {
+            src
+          }
```
dzucconi, line 136: As it happens; we kind of do want a low quality image that we'll just blur. If we were to use a non-blurred image it'd have to be enormous since it's essentially full-screen since it is both full width and is progressively revealed by scrolling. So like ~4000px wide.

PR #9106, opened 2021-12-21
chore: add tracking for price options

jo-rs:
Addresses [NX-3026]

Implements tracking events for new offer submission flow.

**Click-event values:**
- `10% below the list price`
- `15% below the list price`
- `20% below the list price`
- `Low-end of range`
- `Midpoint`
- `Top-end of range`
- `Different amount`
- `We recommend changing your offer`

_**NOTE:** there's two events, one of them commented out, which will need to be better implemented, or commented back in where relevant, while @tam-kis is working on [NX-3013]._

cc: @artsy/negotiate-devs @daytavares 

[NX-3026]: https://artsyproduct.atlassian.net/browse/NX-3026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NX-3013]: https://artsyproduct.atlassian.net/browse/NX-3013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -30,6 +56,14 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
...
   useEffect(() => {
     customValue && setValue(customValue)
+    // TODO: move this call if necessary once the feature is implemented
+    if (toggle && customValue && customValue < priceOptions[0]?.value!) {
+      tracking.trackEvent({
+        action_type: AnalyticsSchema.ActionType.ViewedOfferTooLow,
+        flow: AnalyticsSchema.Flow.MakeOffer,
+        order_id: order.internalID,
+      })
+    }
```
jo-rs, line 66: This should trigger when the "Galleries usually accept offers within 20% of the listed price; any lower is likely to be rejected." notice is implemented on [NX-3013].

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -56,7 +90,7 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
       if (listPrice?.major) {
         return {
           value: listPrice.major * (1 - pricePercentage),
-          description: `${pricePercentage * 100}% below list price`,
+          description: `${pricePercentage * 100}% below the list price`,
```
jo-rs, line 93: Copy fix requested by design

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -67,6 +101,9 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
     ? getRangeOptions()
     : getPercentageOptions()
 
+  // TODO: add call bellow when the feature is implemented
+  // trackClick("We recommend changing your offer", priceOptions[0]?.value!)
```
jo-rs, line 105: This should trigger when the "We recommend changing your offer to..." cta is implemented on [NX-3013].

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -137,5 +176,10 @@ export const PriceOptionsFragmentContainer = createFragmentContainer(
         }
       }
     `,
+    order: graphql`
+      fragment PriceOptions_order on CommerceOrder {
+        internalID
+      }
```
jo-rs, line 182: Added since `AnalyticsSchema` requires it for these click events.

src/v2/Apps/Order/Components/PriceOptions.tsx
araujobarret: if we're going to fix it soon I don't mind it here, otherwise, separate in a different useEffect call
```jsx
useEffect(() => {
    customValue && setValue(customValue)
}, [customValue, setValue])

useEffect(() => {
  // TODO: move this call if necessary once the feature is implemented
    if (toggle && customValue && customValue < priceOptions[0]?.value!) {
      tracking.trackEvent({
        action_type: AnalyticsSchema.ActionType.ViewedOfferTooLow,
        flow: AnalyticsSchema.Flow.MakeOffer,
        order_id: order.internalID,
      })
    }
}, [toggle, customValue, priceOptions, order])
```

araujobarret: Sorry, I missed the other review but I just realized a naming Component design issue here.

The `setValue` prop should be called `onChange` to be more agnostic from the caller side in what to do when a change happens. `setValue` is too biased to what `OfferRoute` wants to do when a change happens.
 Can you change it, please? 💟

araujobarret (approved):

src/v2/Apps/Order/Components/PriceOptions.tsx
jo-rs: Yeah, the `ViewedOfferTooLow` tracking event will be moved when that functionality is implemented.
Otherwise, added the dep update for `setValue` 👍

araujobarret: thanks for doing it!
and about the other kind request? 😂 

>Sorry, I missed the other review but I just realized a naming Component design issue here.

>The setValue prop should be called onChange to be more agnostic from the caller side in what to do when a change happens. setValue is too biased to what OfferRoute wants to do when a change happens.
Can you change it, please? 💟

jo-rs: Oops, sorry, I missed that. Fixed!

jo-rs:
Merging this since all concerns have been addressed and it's under a feature flag, so we can start testing.

PR #9107, opened 2021-12-21
feat: expose Make Offer button on eligible artworks

ashleyjelks:
This PR adds the Make Offer button on the artwork page when a work is offerable from inquiry _and_ the price is either listed as a range or exact number. On these works when the price is not listed we continue to show the "Contact Gallery" button as the only CTA.

**On an `isOfferableFromInquiry` work with a price range:**
Screen Shot 2021-12-21 at 4 18 39 PM

**On an `isOfferableFromInquiry` work with an exact price:**
Screen Shot 2021-12-21 at 4 27 46 PM

**On an `isOfferableFromInquiry` work with a hidden price:**
Screen Shot 2021-12-21 at 4 18 50 PM

cc @artsy/negotiate-devs 

https://artsyproduct.atlassian.net/browse/NX-3042


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -357,21 +357,48 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
-    )
+    const artworkEcommerceAvailable = !!(isAcquireable || isOfferable)
 
-    if (!artwork.sale_message && !artwork.is_inquireable) {
+    if (!artwork.sale_message && !isInquireable) {
       return <Separator />
     }
 
+    const shouldDisplayMakeOfferButton = (): boolean => {
+      if (isOfferable || (isPriceListed && isOfferableFromInquiry)) {
```
jo-rs, line 379: I don't think we have to be checking for `isPriceListed` here, no? 🤔 
Also, I think here should be the lab flag check; something like `if (isOfferable || (isOfferableFromInquiry && hasFeatureFlag("Make Offer On All Eligible Artworks"))) {` (or however we check for those 😅 ).

jo-rs: Also, we can probably simplify this to directly return that condition.

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -357,21 +357,48 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
+
+    const shouldDisplayMakeOfferAsPrimary = (): boolean => {
+      if (
+        shouldDisplayMakeOfferButton() &&
+        shouldDisplayContactGalleryButton()
+      ) {
+        return true
+      }
+      return false
+    }
```
jo-rs, line 393: Similarly:
`const shouldDisplayMakeOfferAsPrimary = () => shouldDisplayMakeOfferButton() && shouldDisplayContactGalleryButton()`
...I think 🤔

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -357,21 +357,48 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
+
+    const shouldDisplayContactGalleryButton = (): boolean => {
+      if (isInquireable && !isAcquireable && !isOfferable) {
+        return true
+      }
+      return false
+    }
```
jo-rs, line 400: And same here, about directly returning the condition.

jo-rs (changes requested):
Generally looks good, we just need to be mindful of hiding this behind the feature flag for now.
Also a couple of nitpicks 😬

src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercial.jest.tsx
```diff
@@ -151,6 +154,47 @@ describe("ArtworkSidebarCommercial", () => {
     expect(wrapper.text()).toContain("Make offer")
   })
 
+  it("displays both Make Offer and Contact Gallery CTAs when offerable from inquiry and exact price listed", () => {
```
araujobarret, line 157: Please don't create any new tests with Enzyme, we're migrating bit by bit to RTL which is reliable and more efficient to the tests, if it is too much refactoring the test file at least the newer ones use RTL, see more on [link](https://github.com/artsy/force/blob/main/docs/best_practices.md#write-unit-tests-for-new-components)

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -440,33 +465,37 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
+              width="100%"
+              size="medium"
+              onClick={this.handleInquiry.bind(this)}
+              mt={shouldDisplayMakeOfferAsPrimary() ? 2 : 0}
+              variant={
+                shouldDisplayMakeOfferAsPrimary()
+                  ? "secondaryOutline"
+                  : "primaryBlack"
+              }
+              data-test-id="contactGalleryButton"
```
araujobarret, line 494: I don't think we need a test-id here to select this button in the tests.
You can use(RTL): 
```javascript
screen.getAllByRoles("button")[desiredIndex]
screen.getByText("Contact Gallery")
```
both cases are better than test ids, test ids is the last resort if we can't query the component in any possible way

ashleyjelks: Noted about RTL but I would push back about grabbing components by their index rather than a specific property on the component, e.g. a test id bc if the UI changes, e.g. a new button, then we have to update the tests to reflect what could be a moving target (new UI elements on web page)

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
ashleyjelks: > I don't think we have to be checking for isPriceListed here, no? 🤔

Looks like we could use either field, checking if listPrice.price is null or for the value of this boolean. I think it's a case of 2 +2 = 4 as is 3 + 1 here unless there's some reason why one check over the other is valid?
https://ibb.co/XpxYT90

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
ashleyjelks: Ah I see what you mean. This was just a refactor to [clean this up ](https://github.com/artsy/force/blob/a1d6c27a6b8a29eaeb9778c19cd4686e95d3e7d3/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx#L459)a bit.

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
ashleyjelks: Same here, trying to avoid TS errors with an explicit return if the condition isn't met. https://ibb.co/k13CPKt

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
lidimayra: Interesting discussion here... I also thought that using test-ids would be preferable, for the same reasons that @ashleyjelks mentioned... I would be curious to learn more about the downsides of this approach, as there's probably something that I'm not seeing...

araujobarret: @lidimayra test-ids are only recommended when you have no other choice, that's why it's in the lower level of priority in the [React testing library documentation](https://testing-library.com/docs/queries/about/#priority)

@ashleyjelks I agree about the index, but the getByText sill be useful in this case and has a higher priority over test-id queries

lidimayra: @araujobarret Cool, many thanks for the clarification and the reference 🙌

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
ashleyjelks: I asked about this in Slack [here](https://artsy.slack.com/archives/C02JHHHKP5K/p1640269454255200?thread_ts=1640206269.250700&cid=C02JHHHKP5K) and now I think checking if the price is listed might be the way to go?

jo-rs: Hmm I'm not sure if we care. At least, I think, on Inquiry Checkout we're not checking if the artwork has a listed price, only for `isOfferableFromInquiry`, and the logic should be the same, no? Maybe I'm missing something 🤔

ashleyjelks: I was going off of the acceptance criteria in the [ticket](https://artsyproduct.atlassian.net/browse/NX-3042):

`MO button shows on artwork page of artworks currently available through MOI as long as price display is exact price or price range.`

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
lean8086: @ashleyjelks seems that `isInquireable` needs to be checked as boolean like the other two:
```
const shouldDisplayContactGalleryButton = (): boolean => !!isInquireable && !isAcquireable && !isOfferable
```

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -357,21 +358,51 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
+      if (
+        isOfferable ||
+        (isPriceListed && isOfferableFromInquiry && labFeatureEnabled)
+      ) {
+        return true
+      }
+      return false
+    }
+
+    const shouldDisplayMakeOfferAsPrimary = (): boolean => {
```
lean8086, line 393: I think all the "shouldDisplayX" functions could simply be constants instead, right?

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -456,17 +483,21 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
-              >
-                Contact Gallery
-              </Button>
-            )}
+          {shouldDisplayContactGalleryButton() && (
+            <Button
+              width="100%"
+              size="medium"
+              onClick={this.handleInquiry.bind(this)}
+              mt={shouldDisplayMakeOfferAsPrimary() ? 2 : 0}
```
lean8086, line 491: Just curious: I'm still not sure when we use `mt` on a Spacer vs. on the component itself. For consistency sake, should we use `mt` also on the Make offer button, or use a Spacer here, or is it ok to have both?

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -357,21 +358,51 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
 
   render() {
     const { artwork, inquiryComponent } = this.props
+    const isPriceListed = !artwork?.isPriceHidden
+    const isOfferableFromInquiry = artwork?.isOfferableFromInquiry
+    const isOfferable = artwork?.is_offerable
+    const isAcquireable = artwork?.is_acquireable
+    const isInquireable = artwork?.is_inquireable
```
jo-rs, line 365: We can probably remove the aliases on the graphql query, and just destruct the artwork object, no?
`const { isOfferable, isAcquireable, isInquireable } = artwok`

jo-rs (commented):
Other than what I commented re. graphql aliases, I agree with @lean8086's comments. I think once that is taken care of, we're good to go 🚀

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
ashleyjelks: This comp is pretty old...was there a reason they were aliased in the first place? I was hesitant to make this code change without understanding the reason for it to begin with so I just limited changes to the scope of the ticket.

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -510,10 +533,19 @@ export const ArtworkSidebarCommercialFragmentContainer = createFragmentContainer
...
+          ...ArtworkSidebarSizeInfo_piece
+        }
         internalID
-        is_for_sale: isForSale
+        isOfferableFromInquiry
+        isPriceHidden
         is_acquireable: isAcquireable
+        is_for_sale: isForSale
         is_inquireable: isInquireable
         is_offerable: isOfferable
```
jo-rs, line 550: Sorry, I wasn't clear on my previous comment 😓 
We can remove these aliases and just have:
```suggestion
        isAcquireable
        is_for_sale: isForSale
        isInquireable
        isOfferable
```
I think we were using the aliases because of some outdated MP issue, so they're no longer needed.

ashleyjelks: Got it. In that case I'm going to merge this and move that to another ticket. 1) Its outside of the scope 2) its not a blocker to this work 3) I would have to look into how this component's children consume its props to ensure no regressions are introduced and that too is outside of the scope of this work.

PR #9113, opened 2021-12-22
feat(FX-3648, FX-3646): Add a search results demo to /search2 route

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3648], [FX-3646]

### Acceptance Criteria
* As an Artsy engineer
* When I visit `/search2/results?query=abc`
* Then I see artist results from Algolia

### Demo links
* Algolia results for `abc` - [LINK](https://algolia-results.artsy.net/search2/results?query=abc)
* Algolia results for `banksy` - [LINK](https://algolia-results.artsy.net/search2/results?query=banksy)
* Algolia results for `banksy` with pagination - [LINK](https://algolia-results.artsy.net/search2/results?query=banksy&page=5)

### Demo video
https://user-images.githubusercontent.com/3513494/147341234-7140e87b-b545-47cc-b8d2-562054cd8d62.mp4

[FX-3648]: https://artsyproduct.atlassian.net/browse/FX-3648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FX-3646]: https://artsyproduct.atlassian.net/browse/FX-3646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e4c4e75f1cc34714aeef61b4452e13d554decea7
</p>

src/v2/Apps/Algolia/algoliaRoutes.tsx
```diff
@@ -0,0 +1,60 @@
+import loadable from "@loadable/component"
+import { AppRouteConfig } from "v2/System/Router/Route"
+import { graphql } from "react-relay"
+
+const AlgoliaApp = loadable(
```
damassi, line 5: I imagine that after this is promoted out of a demo state we'll rename these components (and the route) to just search, yah?

src/v2/Apps/Algolia/routes/AlgoliaResults/AlgoliaResults.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const onSearchStateChange = updatedSearchState => {
+    if (debouncedSetStateRef.current) {
+      clearTimeout(debouncedSetStateRef.current)
+    }
+
+    debouncedSetStateRef.current = setTimeout(() => {
+      const params = searchStateToUrl(updatedSearchState)
+      const url = `${match.location.pathname}${params}`
+
+      window.history.pushState(null, "", url)
```
damassi, line 58: We'll want to use our router here -- `const router = useRouter()` (from v2/System/Router`)

src/v2/Apps/Algolia/routes/AlgoliaResults/Components/AlgoliaPagination.tsx
```diff
@@ -0,0 +1,40 @@
+import { Box, Pagination } from "@artsy/palette"
+import { useScrollTo } from "v2/Utils/Hooks/useScrollTo"
+import { ANCHOR_CONTAINER_ID } from "v2/Apps/Algolia/constants"
+import { createPageCursors } from "v2/Apps/Algolia/Utils/pagination"
+
+export const AlgoliaPagination = props => {
+  const { nbPages, currentRefinement, refine } = props
```
damassi, line 7: What does `nb` mean? what does `refine` mean? If possible we should make these variables a bit more explicit.

src/v2/Apps/Algolia/routes/AlgoliaResults/Components/__tests__/AlgoliaResultsList.jest.tsx
```diff
@@ -0,0 +1,161 @@
...
+      name: "article",
+    },
+    {
+      displayName: "Artist",
+      name: "artist",
+    },
+  ],
+}
+
+const SearchResultsFixture: SearchResults<AlgoliaHit> = {
```
damassi, line 123: We should avoid fixtures if possible as they become a burden to maintain in the future. That's not to say that we should never use them, only that if we do, we should only supply fixture data for the props that we're asserting against in the test, or that which is required in order to get the component to render. (Relay test tools auto-fixture everything else). If these conditions are required here then ignore this comment!

damassi:
This is looking great @dimatretyak 👍 No real big changes over here! 

Adding @dzucconi as he should give this a look.

package.json
```diff
@@ -204,6 +205,8 @@
     "react-dom": "17.0.2",
     "react-dropzone": "^11.4.2",
     "react-head": "3.0.2",
+    "react-instantsearch-core": "^6.18.0",
+    "react-instantsearch-dom": "^6.18.0",
```
dzucconi, line 209: I mean — I would not have used this library and would much rather us write our own code. Though there's the question of whether or not we have the time to do this?




This is a lot of code to add to nearly every major page on the site.

cc @icirellik

yarn.lock
```diff
@@ -2,6 +2,115 @@
...
+
+"@algolia/transporter@4.11.0":
+  version "4.11.0"
+  resolved "https://registry.yarnpkg.com/@algolia/transporter/-/transporter-4.11.0.tgz#a8de3c173093ceceb02b26b577395ce3b3d4b96f"
+  integrity sha512-k4dyxiaEfYpw4UqybK9q7lrFzehygo6KV3OCYJMMdX0IMWV0m4DXdU27c1zYRYtthaFYaBzGF4Kjcl8p8vxCKw==
+  dependencies:
+    "@algolia/cache-common" "4.11.0"
+    "@algolia/logger-common" "4.11.0"
+    "@algolia/requester-common" "4.11.0"
+
```
dzucconi, line 113: 😱

src/v2/Apps/Algolia/algoliaRoutes.tsx
dimatretyak: Yes, that's right

src/v2/Apps/Algolia/routes/AlgoliaResults/Components/AlgoliaPagination.tsx
dimatretyak: * `nbPages` - The total number of existing pages.
* `refine` - Selects a specific page

You can read more about these [here](https://www.algolia.com/doc/api-reference/widgets/pagination/react/#create-a-react-component)

src/v2/Apps/Algolia/routes/AlgoliaResults/Components/__tests__/AlgoliaResultsList.jest.tsx
dimatretyak: > If these conditions are required here then ignore this comment!

Yes, all these fixtures are necessary here in order for the component to render without errors and then to cover it with tests correctly

src/v2/Apps/Algolia/routes/AlgoliaResults/AlgoliaResults.tsx
dimatretyak: I initially tried using `router.push`, but unfortunately I got the wrong behavior: when using pagination, a new page always opens from the top and there is no smooth scrolling animation (see attached video)

https://user-images.githubusercontent.com/3513494/147920161-9a023d6d-7061-4798-8b64-a53068b5326e.mp4

package.json
dimatretyak: > I mean — I would not have used this library and would much rather us write our own code. Though there's the question of whether or not we have the time to do this?

Good question. I'll see what can be done and if there is any way to reduce the size

dimatretyak: Made some changes, now `react-instantsearch-dom` is not used

dblandin (commented):
Thanks for kicking off this process @dimatretyak !

Some high-level feedback:

- I think we should trim out some of the Algolia-naming. For example, we might consider naming this new app `Search2` with the expectation that some version of it will eventually replace the existing Search app.
- Similarly, we can update the route from /algolia to /search2 to further minimize the implementation details
- If size of the packages is a concern, we should look into ways of bundling those dependencies so that they're only sent to the client when necessary. Perhaps we could defer some importing until a search interface is active?

I think this PR does a great job of mimicking the current search results page design, but I wonder if there might be simpler diff available that leans more on Algolia-provided react components to start with. We don't yet have new designs available internally and the next iteration might differ quite a bit from what we have right now. We just don't know yet.

For this exploration, I think the more we can lean on Algolia as a platform, the better - which would include using Algolia react components and writing a minimal amount of supporting code. This is just a first step and we'll have discussions with design to ensure that any new implementation matches the look and feel of the wider ecosystem. We can defer whether we accomplish that by "theming" Algolia components, or completely customizing components using our own Palette design system (as shown in this PR).

dimatretyak:
> I think this PR does a great job of mimicking the current search results page design, but I wonder if there might be simpler diff available that leans more on Algolia-provided react components to start with. We don't yet have new designs available internally and the next iteration might differ quite a bit from what we have right now. We just don't know yet.

For this exploration, I think the more we can lean on Algolia as a platform, the better - which would include using Algolia react components and writing a minimal amount of supporting code. This is just a first step and we'll have discussions with design to ensure that any new implementation matches the look and feel of the wider ecosystem. We can defer whether we accomplish that by "theming" Algolia components, or completely customizing components using our own Palette design system (as shown in this PR).

If we decide to use Algolia components, then in this case we will have to return `react-instantsearch-dom` dependency back, which will affect the final bundle size. And also we will have to add styles for these components, since out of the box there is just a layout.

It seems to me that the current implementation (through custom components and Algolia connectors) is more scalable and will give us more flexibility in the future, since we will independently control the behavior and styling of these components ourselves.

dzucconi:
> It seems to me that the current implementation (through custom components and Algolia connectors) is more scalable and will give us more flexibility in the future, since we will independently control the behavior and styling of these components ourselves.

Strongly concur with this — Algolia's business isn't building UI. Any UI package is just going to get in our way and bloat the application with parallel implementations of the vast majority of our components.

I also have a strong suspicion that adding it 'as a first step' would just doom us to using it unless there is a strong firewall between a throwaway prototype and the rest of the codebase.

package.json
dimatretyak: @dzucconi Can I mark this thread as resolved?

dzucconi: 👍

dblandin:
> It seems to me that the current implementation (through custom components and Algolia connectors) is more scalable and will give us more flexibility in the future, since we will independently control the behavior and styling of these components ourselves.

> Strongly concur with this — Algolia's business isn't building UI. Any UI package is just going to get in our way and bloat the application with parallel implementations of the vast majority of our components.

That's fair! If we want to jump right into to Algolia's "connector" support considering that we already have a mature library of our own design system components (which we can extend as needed) - that sounds good to me. It's definitely the more flexible option. We'll have to see if it's also the scalable option.

dimatretyak:
> I think we should trim out some of the Algolia-naming. For example, we might consider naming this new app Search2 with the expectation that some version of it will eventually replace the existing Search app.

> Similarly, we can update the route from /algolia to /search2 to further minimize the implementation details

Done

dblandin (dismissed):
This looks like a great start to me! 💯 

Might wait to merge until we get another look from @damassi and @dzucconi.

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
```diff
@@ -0,0 +1,128 @@
+import { useState, useMemo, useRef, useEffect } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { Search2Results_system } from "v2/__generated__/Search2Results_system.graphql"
+import algoliasearch from "algoliasearch"
+import {
+  InstantSearch,
+  Configure,
+  connectPagination,
+  connectStateResults,
+} from "react-instantsearch-core"
```
anandaroop, line 10: **praise:** Ahh, nice. So these were all just re-exported by `react-instantsearch-dom`, which is why we're able to instead use these directly from `react-instantsearch-core` if we're creating our own DOM, yes?

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
```diff
@@ -0,0 +1,128 @@
...
+    setSearchState(urlToSearchState(match.location.search))
+  }, [match.location])
+
+  if (searchClient && selectedIndice) {
+    return (
+      <InstantSearch
+        searchClient={searchClient}
+        indexName={selectedIndice.name}
+        searchState={searchState}
+        onSearchStateChange={onSearchStateChange}
```
anandaroop, line 87: **question (non-blocking):** I was trying to understand where this debounced callback fires — would we see this now, or only after the search input and/or facets are hooked up?

anandaroop (commented):
Amazing start! 👏🏽

src/v2/Apps/Search2/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const HITS_PER_PAGE = 10
+export const ANCHOR_CONTAINER_ID = "jump--searchResults"
+export const DEBOUNCE_TIME = 400
+export const DENIED_SEACH_STATE_KEYS = ["configure"]
```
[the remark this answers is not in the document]
dimatretyak: Oh, I didn't notice that. Thanks for the comment. Fixed

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
dimatretyak: Yes, you are absolutely right. After looking at [the source code](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/index.js#L14-L43) of `react-instantsearch-dom`, you can see that all the connectors we need are imported from `react-instantsearch-core`. And for this reason, we can safely abandon the use of `react-instantsearch-dom`

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
dimatretyak: We can see now if we use the pagination from below

https://user-images.githubusercontent.com/3513494/148361830-bbf15925-c077-4198-9fbb-4f9cf9f4e9d8.mp4

cypress/integration/search2.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { visitWithStatusRetries } from "../helpers/visitWithStatusRetries"
+
+describe("Search", () => {
+  it("/search2/results?:query", () => {
+    visitWithStatusRetries("/search2/results?query=andy")
```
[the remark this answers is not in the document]
dimatretyak: Yes, we can make the url in the following format: `search2?term=abc`.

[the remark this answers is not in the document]
dimatretyak: @dblandin which url format should I do: as Roop (1) suggests or as described in the task in Jira (2)?

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
```diff
@@ -0,0 +1,128 @@
...
+const ConnectedSearch2Pagination = connectPagination(Search2Pagination)
+const ConnectedSearch2Results = connectStateResults(
+  Search2ResultsListFragmentContainer
+)
+
+export const Search2Results: React.FC<Search2ResultsProps> = ({ system }) => {
+  const { algolia } = system
+  const { match } = useRouter()
+  const [searchState, setSearchState] = useState(
+    urlToSearchState(match.location.search)
```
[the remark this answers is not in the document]
dimatretyak: Hmm, good idea. Checked everything again and we can use just `match.location.query`

cypress/integration/search2.spec.ts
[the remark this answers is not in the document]
dblandin: @dimatretyak Let's go with @anandaroop's suggestion!

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
dzucconi: What exactly does this package provide? It's not immediately clear to me?

dimatretyak: `react-instantsearch-core ` contains all the basic logic for working with Algolia in React

dzucconi: But like: what's that actually mean? We're just hitting Metaphysics ultimately, right?

cypress/integration/search2.spec.ts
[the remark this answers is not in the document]
dimatretyak: If we want to support url in `search2?term=banksy` format, then we need to convert `term` query param to `query`, since the `query` key is expected in the `searchState` for [Algolia InstantSearch](https://www.algolia.com/doc/api-reference/widgets/ui-state/react/) (see the attached screenshot)

By this reason, these converters are used:
* converting `term` query param to `query` for Algolia Instantsearch [[link to the source code](https://github.com/artsy/force/blob/02f7229e1a7b523d1cdf6bef8314c4823626769b/src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx#L58-L59)]
* reverse conversion of query to term when the state of the algolia search changes (for example, during pagination) in order to synchronize the changes with the url [[link to the source code](https://github.com/artsy/force/blob/02f7229e1a7b523d1cdf6bef8314c4823626769b/src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx#L58-L59)]

@dblandin @anandaroop maybe it's still better to support the url in `search2?query=banksy` format and make redirects for old search links (from external links, bookmarks etc)?

searchState | React InstantSearch widget | Algolia 2022-01-06 19-31-14

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
dimatretyak: > We're just hitting Metaphysics ultimately, right?

Not really. Metaphysics on the new search page is only [needed](https://github.com/artsy/force/blob/02f7229e1a7b523d1cdf6bef8314c4823626769b/src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx#L45) to get the app id and api key for Algolia. All search results will be displayed **from Algolia**

dimatretyak: For example, for "Artist" pill the results from [this Algolia index](https://www.algolia.com/apps/CQWT6HEHVH/explorer/browse/Artist_staging?query=banksy&searchMode=search) will be displayed

dzucconi: Ah, OK. I continue to think it is a very bad idea to not encapsulate an external service within Metaphysics — essentially working around the entire problem it solves... but I'm sure your team has thought through all the problems if it's gotten to this point.

cypress/integration/search2.spec.ts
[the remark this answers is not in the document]
anandaroop: Thanks for looking into that. I think it's also fine to punt on this for now and leave as-is. We can add a redirect later if it makes sense. (We discussed the possibility of a re-imagined search experience so maybe a simple redirect won't make sense either, but I'll put a ticket on the backlog so we can track it)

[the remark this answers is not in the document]
dimatretyak: We discussed it with @anandaroop and decided not to make changes yet and leave `search2/results?query=banksy`

src/v2/Apps/Search2/routes/Search2Results/Search2Results.tsx
dblandin: > Ah, OK. I continue to think it is a very bad idea to not encapsulate an external service within Metaphysics — essentially working around the entire problem it solves... but I'm sure your team has thought through all the problems if it's gotten to this point.

I think the most compelling reasons _not_ to encapsulate and abstract Algolia within Metaphysics have been:
- To support Algolia's automatic routing of the search request to the closest data center - minimizing latency
- Not needing to spend time building out a potentially complex proxying layer

Definitely not too late to experiment with other approaches though. Do you think we should explore using the InstantSearch library with a custom backend proxy and/or expose a new GraphQL field that abstracts most of the Algolia bits away? 

For example, we could explore an approach that would:
- Expose a new API endpoint in Gravity that proxies Algolia-formatted requests to Algolia
- Expose new Metaphysics fields to power searching/filtering via Algolia

https://www.algolia.com/doc/guides/building-search-ui/going-further/backend-search/in-depth/backend-instantsearch/js/#searching-for-results
https://github.com/charlypoly/algolia-graphql-schema

During very early [conversations](https://www.notion.so/artsy/Integration-with-Search-Vendor-2bb4f7dd196a456394f26477d97d4b58) - there was quite a bit of concern against maintaining too much of our own search infrastructure but perhaps we swung too far in that direction. We can still successfully leverage a backend system like Algolia while owning more of the glue that connects it with our apps via Metaphysics or other APIs.

dzucconi: I'm mostly concerned as to how this plays with the existing schema — if Algolia makes it into all filtered artwork sets then the vast majority of the UI depends on it. I guess I just don't understand the mechanics of how we would use our existing UI and fragments inside of Algolia without just simply hitting Metaphysics again anyway.

How much performance really are we actually saving by hitting Algolia directly instead of stitching it into Metaphysics?

dimatretyak:
@dblandin @anandaroop @dzucconi can I merge this PR or do I need to add some changes/improvements?

dblandin:
> can I merge this PR or do I need to add some changes/improvements?

I think we should temporarily hold off on this PR pending some investigation to integrate Algolia-query support into Metaphysics. @dzucconi made some good points that we should follow up on before commiting to a particular approach.

PR #9115, opened 2021-12-23
Consolidate address forms

lean8086:
Use one address form component to create and edit addresses on the platform. Use [Formik](https://formik.org/docs/overview) for form validations. Use [Yup](https://www.npmjs.com/package/yup?activeTab=readme) for validation schemas.

Ticket https://artsyproduct.atlassian.net/browse/NX-2993
More context https://www.notion.so/artsy/Consolidate-Address-Forms-be80599c5893416cad5cd0d6f3191bf1

- Create component BillingInfoFormContext using formik context. **Is there a better name?**
- Make it work with FC and CC.
- Implement it on BidForm, RegistrationForm, AddressModal, PaymentModal, Shipping/index, PaymentPicker.
- Use AddressForm component for address inputs.
- Implement Yup for validation schemas.

Co-authored-by: @laurabeth @rquartararo @araujobarret 

cc: @artsy/negotiate-devs


src/v2/Apps/Auction/Components/RegistrationForm.tsx
```diff
@@ -1,124 +1,91 @@
...
   needsIdentityVerification: boolean
 }
+export const RegistrationForm: React.FC<RegistrationFormProps> = ({
+  onSubmit,
+  trackSubmissionErrors,
+  needsIdentityVerification,
+}) => {
+  return (
+    <BillingInfoFormContext
+      formKeys={["addressWithPhone", "agreeToTerms", "creditCard"]}
```
araujobarret, line 33: isn't possible to use the keys here as the formik type from the `BillingInfoFormContext`?
I'm not sure but I guess we can override the `FormikProps<Pick<BillingInfoFormValues, key in FormKeys>>` so in the render method below you won't need to specify the type and import any extra type.
```jsx
{formik => ()}
```

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -113,147 +159,26 @@ export const AddressModal: React.FC<Props> = ({
...
-            }
-          }}
+        <BillingInfoFormContext
+          onSubmit={onSubmit}
+          initialValues={{ address: { ...address } }} // Spread to avoid immutability issues on changes
+          formKeys={["addressWithPhone"]}
         >
-          {(formik: FormikProps<SavedAddressType>) => (
-            <form onSubmit={formik.handleSubmit}>
+          {(formik: FormikProps<BillingInfoPicked>) => (
```
araujobarret, line 167: ditto here about type extends from `formKeys` prop

src/v2/Apps/Auction/Components/Form/BillingInfoFormContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+      onSubmit={onSubmit}
+      validationSchema={schema}
+      validateOnMount
+    >
+      {children}
+    </Formik>
+  )
+}
+
+export const BillingInfoFormContext = forwardRef<BillingInfoFormProps>(
```
araujobarret, line 40: for consistency's sake use the suffix `...ContextProvider`

src/v2/Apps/Order/Components/CreditCardInput.tsx
```diff
@@ -53,12 +52,12 @@ const StyledBorderBox = styled(BorderBox)<BorderProps>`
 `
 
 interface CreditCardInputProps {
-  error?: StripeError
   onChange?: (response: StripeCardNumberElementChangeEvent) => void
 }
 
 export const CreditCardInput: React.FC<CreditCardInputProps> = props => {
   const [focused, setFocused] = useState(false)
+  const { errors, setFieldError } = useFormikContext<BillingInfoWithTerms>()
```
araujobarret, line 60: This approach is fine but I guess we could export specific typed context hooks from the ContextProvider and use the exported hook from there instead of using `useFormikContext<BillingInfoWithTerms>()`

From the context provider file, you could have
```jsx
const useBillingInfoWithTermsContext = () => {
  const formikContext = useFormikContext<BillingInfoWithTerms>()
  
  return formikContext
}
```
in this way we make sure we use the right hook attached to the FormContext we want, saves from misusage and typing boilerplate.

src/v2/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -42,6 +40,8 @@ import {
 } from "v2/System/SystemContext"
 import { createStripeWrapper } from "v2/Utils/createStripeWrapper"
 import { isNull, mergeWith } from "lodash"
+import { BillingInfoFormContext } from "v2/Apps/Auction/Components/Form"
```
araujobarret, line 43: I guess the Form Context is in the wrong place, if it's a shared thing across different apps should not be inside Auction

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -83,6 +74,8 @@ import {
   ContextModule,
   OwnerType,
 } from "@artsy/cohesion"
+import { BillingInfoFormContext } from "v2/Apps/Auction/Components/Form"
```
araujobarret, line 77: ditto about the Context location

src/v2/Apps/Order/Utils/formValidators.tsx
```diff
@@ -1,48 +1,22 @@
-import { Address } from "v2/Components/AddressForm"
 import { isEmpty } from "lodash"
 
-export const validatePresence = (value: string): string | null => {
+export const validatePresence = (value: string | null): string | null => {
   if (value == undefined || isEmpty(value.trim())) {
```
araujobarret, line 4: then here you can => `if (!value || isEmpty(value.trim())) {`

src/v2/Components/AddressForm.tsx
```diff
@@ -170,82 +148,130 @@ export const AddressForm: React.FC<AddressFormProps> = ({
...
+                !billing ? "Only used for shipping purposes" : undefined
+              }
+              options={countries}
+              style={{
+                letterSpacing: "1px",
+                borderRight: "none",
+              }}
+              // TODO: Replace data-testid with queryByDisplayValue() on tests
+              // when a `selected` property is implemented
+              data-testid="phoneNumberCountry"
```
araujobarret, line 250: you don't need this testid here, you can `screen.getByRole("combobox")` from the tests to query this guy

araujobarret (changes requested):
I've requested some changes in this PR, major typings and organizational stuff

src/v2/Apps/Auction/Components/Form/initialValues.ts
```diff
@@ -1,34 +1,52 @@
...
+export const initAddressWithPhone: Address | SavedAddressType = {
+  ...initAddress,
   phoneNumber: "",
 }
 
-export const initialValuesForRegistration: FormValuesForRegistration = {
-  address,
+export const initBillingInfoWithTerms: Pick<
+  BillingInfoFormValues,
+  "address" | "agreeToTerms" | "creditCard"
```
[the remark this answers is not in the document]
lean8086: Although `addressWithPhone` exists as a key of the type `BillingInfoFormValues`, we never use the initial values as `addressWithPhone` but always as `address`. As far as I understand, we should never Pick `addressWithPhone`, right? I can be missing something, though.

src/v2/Apps/Auction/Components/RegistrationForm.tsx
lean8086: I get the concept but I guess `addressWithPhone` would be a problem here bc of what I mentioned [here](https://github.com/artsy/force/pull/9115#discussion_r777584355)?

araujobarret: I guess you can just ignore my request, it's going too much into this mapping thing, but has no relation with your previous comment though.

src/v2/Apps/Auction/Components/Form/initialValues.ts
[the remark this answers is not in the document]
laurabeth: @lean8086 That's correct, because we'll never call addressWithPhone out of the formik context. It's always sent to us or back to the API under the umbrella of address. I think the intention here was to make it work, then (in a follow up PR or ticket) standardize the forms and their data sources to use 'phone number' as a key external to the address.

@araujobarret It is confusing! 😆 Most of these alternate cases stem from the data formatting inconsistencies we talked about. Hopefully in some follow up PRs we'll be able to standardize and clean up some of these different "initial values" objects.

src/v2/Apps/Auction/Components/OnSubmitValidationError.tsx
```diff
@@ -18,15 +21,15 @@ export type TrackErrors = (errors: string[]) => void
...
-}> = props => {
-  const { cb, formikProps } = props
+export const OnSubmitValidationError: React.FC<{ cb: TrackErrors }> = ({
+  cb,
+}) => {
+  const formikProps = useFormikContext<
+    BillingInfoWithTerms | BillingInfoWithBid
+  >()
 
   const effect = () => {
```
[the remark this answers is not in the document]
lean8086: This seems to be [copied directly from a suggested temporary solution](https://github.com/jaredpalmer/formik/issues/1484#issuecomment-490558973) for a missing property on Formik to pass a callback for when the submission doesn't happen (in our case, for error tracking).

I rewrote the component, keeping the same dependencies. Adding `isValid` and `errors` makes the callback execute one extra time, and I don't think it's worthwhile spending time on fixing it since **the entire component must be removed when Formik implements the feature.**

src/v2/Apps/Order/Components/CreditCardInput.tsx
lean8086: 👍  Similar as Laura mentioned in [another comment](https://github.com/artsy/force/pull/9115#discussion_r778233177), the idea is to hopefully standardize and simplify these datatypes.

src/v2/Apps/Auction/Components/BidForm.tsx
```diff
@@ -81,68 +70,58 @@ export const BidForm: React.FC<Props> = ({
   const {
     requiresCheckbox,
     requiresPaymentInformation,
-    // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-  } = determineDisplayRequirements(saleArtwork.sale.registrationStatus, me)
-  const validationSchema = requiresCheckbox
+  } = determineDisplayRequirements(
+    saleArtwork.sale?.registrationStatus!,
+    me as { hasQualifiedCreditCards: boolean }
```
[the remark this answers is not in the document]
lean8086: Being this is not part of this PR, I'd prefer not to apply changes here.

araujobarret (dismissed):
I guess we're ready to go, I would say stills a bit unclear some typing of the new forms but I don't wanna block this PR, feel free to merge as soon as you need 🚀

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -84,9 +86,14 @@ export const AddressModal: React.FC<Props> = ({
   const createMutation =
     modalDetails?.addressModalAction === "createUserAddress"
 
+  const handleClearAddress = (): void => {
+    console.log("handleClearAddress")
```
lean8086, line 90: Needs to be removed

src/v2/Apps/Order/Components/AddressModal.tsx
```diff
@@ -222,16 +229,21 @@ export const AddressModal: React.FC<Props> = ({
...
           action: () => {
             setShowDialog(false)
+            handleClearAddress()
           },
           text: "Cancel",
         }}
-        onClose={() => setShowDialog(false)}
+        onClose={() => {
+          setShowDialog(false)
+          handleClearAddress()
```
lean8086, line 245: I see `handleClearAddress` only does execute `onClearAddress`... could we use `onClearAddress()` directly here?

laurabeth: There's a pattern in force that suggests "onAction" a passed prop and "handleAction" is used within the component. I'm just following that paradigm for the sake of consistency.

src/v2/Apps/Order/Components/AddressModal.tsx
laurabeth: Good catch, thanks!

PR #9087, opened 2021-12-17
refactor: Update /auction/id to support MPv2

damassi:
This updates all references to MPv1 to point at MPv2 in /desktop/apps/auction/id. 

Doing a scan in force, it looks like this is the last hold out. 

Spinning up a review-app here: 
- https://github.com/artsy/force/pull/9095

Handing this over to #product-bidder-experience to merge. Clicking around, everything works as expected but more formal verification via review-app would be good.

PR #9141, opened 2022-01-05
fix: Finish auction/:id MPV2 support

annacarey:
Branched off of the [review app PR](https://github.com/artsy/force/pull/9095) though probably should have branched off of @damassi's [actual PR](https://github.com/artsy/force/pull/9087).

I think this branch can be merged into the latter, which is what this PR is doing.

Also I didn't add tests for this but it could be a good idea...

This PR:

- [x] fixes the info on the artwork brick
- [x] fixes the artist filter

### DIFF
--- a/src/desktop/apps/auction/actions/artworkBrowser.js
+++ b/src/desktop/apps/auction/actions/artworkBrowser.js
@@ -374,6 +374,7 @@ function formatDataForMPv2(saleArtworksConnection) {
     .map(saleArtwork => {
       const props = {
         ...saleArtwork,
+        ...saleArtwork.saleArtwork,
         ...saleArtwork.sale_artwork,
       }
       return {
--- a/src/desktop/apps/auction/components/artwork_browser/sidebar/ArtistFilter.js
+++ b/src/desktop/apps/auction/components/artwork_browser/sidebar/ArtistFilter.js
@@ -85,7 +85,7 @@ const mapStateToProps = state => {
     filterParams.include_artworks_by_followed_artists
   const allArtists = { id: "artists-all", name: "All" }
   const allArtistsSelected =
-    artistIds.length === 0 && !filterParams.include_artworks_by_followed_artists
+    artistIds.length === 0 && !includeArtworksByFollowedArtists
   const countDisplay = numArtistsYouFollow > 0 ? numArtistsYouFollow : undefined
   const artistsYouFollow = {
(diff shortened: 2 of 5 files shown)


src/desktop/apps/auction/actions/artworkBrowser.js
```diff
@@ -374,6 +374,7 @@ function formatDataForMPv2(saleArtworksConnection) {
     .map(saleArtwork => {
       const props = {
         ...saleArtwork,
+        ...saleArtwork.saleArtwork,
         ...saleArtwork.sale_artwork,
```
annacarey, line 378: Is this spreading confusing to people? Values from all three are needed in the component, so it's the easiest way to make sure `saleArtwork` has everything we need.

src/desktop/apps/auction/queries/filter.js
```diff
@@ -24,7 +24,7 @@ export const filterQuery = `
       aggregations {
         slice
         counts {
-          value
+          id: value
```
annacarey, line 27: The `Checkbox` component is looking at `id` instead of `value` so that was the source of the bug inside of `ArtistFilter.js`.

src/desktop/apps/auction/components/artwork_browser/sidebar/ArtistFilter.js
```diff
@@ -85,7 +85,7 @@ const mapStateToProps = state => {
     filterParams.include_artworks_by_followed_artists
   const allArtists = { id: "artists-all", name: "All" }
   const allArtistsSelected =
-    artistIds.length === 0 && !filterParams.include_artworks_by_followed_artists
+    artistIds.length === 0 && !includeArtworksByFollowedArtists
```
annacarey, line 88: Super minor, just using the variable defined on line 84.

src/desktop/apps/auction/actions/artworkBrowser.js
damassi: not confusing, its mainly just flattening things out

damassi (approved):
Very nice 👌 I knew it would be pretty simple, just a bit of digging. Glad to see that was indeed the case.

annacarey:
Thanks for taking a look @damassi. i'll merge this into your branch and do a last little QA before merging yours in, unless you or anyone has objections

bhoggard (approved):
Thank you!

PR #9146, opened 2022-01-06
feat(Article2): Add /article2 and /articles2

damassi:
WIP

This stubs out the new /article2 routes (with queries). Note that i'm having to fix the issue with anonymous types over in positron before Relay will work. Running into some friction over there with a joiql upgrade.


damassi:
superseded

PR #9149, opened 2022-01-06
[NX-3039] Enable edition sets selection for inquireable artworks

lidimayra:
[NX-3039]

Even though the artwork is listed as `Contact Gallery` only, we still want to display Edition Sets select radio buttons (in case `Make Offer On All Eligible Artworks` lab feature is enabled)

**Before:**
Screenshot 2022-01-05 at 15 52 27

**After:**
Screenshot 2022-01-05 at 15 53 19

Important to have in mind that these changes are focused exclusively on **displaying** the radio buttons.
Actual usage of this data will be implemented during [NX-3051]

[NX-3039]: https://artsyproduct.atlassian.net/browse/NX-3039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NX-3051]: https://artsyproduct.atlassian.net/browse/NX-3051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -98,9 +98,15 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
     )
   }
 
-  renderEditionSet(editionSet: EditionSet, includeSelectOption: boolean) {
+  renderEditionSet(
+    editionSet: EditionSet,
+    includeSelectOption: boolean,
+    editionSelectableOnInquireable: boolean
+  ) {
     const editionEcommerceAvailable =
-      editionSet?.is_acquireable || editionSet?.is_offerable
+      editionSet?.is_acquireable ||
+      editionSet?.is_offerable ||
+      editionSelectableOnInquireable
 
     const editionFragment = (
       <Flex justifyContent="space-between" flex={1}>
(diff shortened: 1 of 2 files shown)


lidimayra:
Acceptance tests build [is failing](https://circleci.com/gh/artsy/force/181475?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) because it's raising 404 when trying to access data for the mock auctions that are weekly seeded on staging env.

This is happening due to an issue on Gravity that's preventing the seeding rake task from succeeding.
https://github.com/artsy/gravity/pull/14827 was opened to fix this issue.

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -377,7 +390,14 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
       selectedEditionSet,
     } = this.state
 
-    const artworkEcommerceAvailable = !!(isAcquireable || isOfferable)
+    const editionSelectableOnInquireable = !!(
+      artwork.is_inquireable && labFeatureEnabled
```
araujobarret, line 394: If the ff is on shouldn't we display the radios independently if it's "Buy now", "Make Offer" or "Contact Gallery"? (I'm considering that have edition sets)

araujobarret (commented):
Thanks for tackling the changes here!

I have one question, we added the radios to the artwork page, but as soon as we select one option, are we sure the modal with the edition that appears next will have the selected set that we selected previously from the radio?

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
lidimayra: I would say yes if we are completely sure that those are the only states possible (`acquirable, `offerable` and `inquireable`). I just tried to be a bit more conservative here because it is not yet completely clear to me if there is any other possibility that I'm not seeing (and the scenario that we are sure that we're focusing on the MOOAEA project would be the inquireable scenario).  I expect to have a better picture of all possible scenarios during [NX-3057](https://artsyproduct.atlassian.net/jira/software/c/projects/NX/boards/41?modal=detail&selectedIssue=NX-3057).

araujobarret (approved):

PR #9153, opened 2022-01-06
chore(GRO-687): Consolidate error handling

dzucconi:
### Client-side errors
- [x] Client-side routing (404-style) error
    * Gets handled `RenderStatus` => `RenderError` — here we can render the `ErrorPage` within the `AppShell` without issue. Links work but there's a flash of empty screen before the route loads.
- [x] thrown runtime error (500-style) within the React application - client-side
    * Caught by `ErrorBoundary` — here we can render the `ErrorPage`. Links do not work because the app is in an error state. Renders without `AppShell`.

### Server-side errors
- [x] 4xx errors outside of React routes
    * Gets handled in the `errorHandlerMiddleware` — Renders `ErrorPage` without `AppShell` and does not mount the client-side.
- [x] 4xx errors within React routes.
    * Gets handled in the `RenderStatus` => `RenderError` — Renders `ErrorPage` with `AppShell`. This one is weird though: It renders with a `200` status code & I'm only able to get this to render in the `AppShell` with 7126dd64648f532589b61d7e217c8049a8d732c5. If I leave the build alone the throw there makes it such that the `errorHandlerMiddleware` picks it up and ultimately uses the correct status code from the intial error. As such it seems like we only have the correct status code by accident?
- [x] thrown runtime error within the React application - SSR
    * Gets handled in the `errorHandlerMiddleware` — Renders `ErrorPage` without `AppShell` and does not mount the client-side.
- [x] Errors raised in `onServerSideRender`
    * Gets handled in the `errorHandlerMiddleware` — Renders `ErrorPage` without `AppShell` and does not mount the client-side.


src/v2/Apps/Components/AppShell.tsx
```diff
@@ -23,27 +23,27 @@ const logger = createLogger("Apps/Components/AppShell")
...
+  const showNav = !routeConfig?.hideNav
+  const showFooter = !isEigen && !routeConfig?.hideFooter
+  const appContainerMaxWidth = routeConfig?.displayFullPage ? "100%" : null
 
   // Check to see if a route has a onServerSideRender key; if so call it. Used
   // typically to preload bundle-split components (import()) while the route is
   // fetching data in the background.
   useEffect(() => {
-    if (isFunction(routeConfig.onClientSideRender)) {
+    if (isFunction(routeConfig?.onClientSideRender) && match) {
```
damassi, line 44: Curious why all of this stuff got a little more safe? (its fine, but wondering why we needed changes here)

src/v2/System/Router/ErrorBoundary.tsx
```diff
@@ -43,59 +51,59 @@ export class ErrorBoundary extends React.Component<Props, State> {
...
+    const {
+      asyncChunkLoadError,
+      detail,
+      genericError,
+      isError,
+      message,
+    } = this.state
 
-    switch (true) {
-      case asyncChunkLoadError: {
```
damassi, line 61: why are we removing this critical check?

src/v2/System/Router/ErrorBoundary.tsx
```diff
@@ -43,59 +51,59 @@ export class ErrorBoundary extends React.Component<Props, State> {
...
-        return (
-          <ErrorModalWithReload
-            message="Please check your network connection and try again."
-            show={asyncChunkLoadError}
-          />
```
damassi, line 66: This error modal is really nice in different contexts, particularly on mobile. As long as the reload functionality is there and very clear to the end user we can go with something more generic (`<ErrorPage>`)

damassi (approved):
I just gotta say that this PR is awesome, even though there are the noted issues. Big 👍 . We can continue iterating.

src/v2/server.ts
```diff
@@ -35,70 +20,22 @@ app.get(
...
-        },
-        meta: {
-          appleItunesApp: `${APP_URL}${CURRENT_PATH}`,
-        },
-        // TODO: Post-release review that sharify is still used in the template.
-        sd: sharify.data,
-      }
-
-      res.render(`${PUBLIC_DIR}/index.ejs`, options)
+      renderServerApp({ req, res, ...rest })
```
damassi, line 38: it might be nice if this function returned a value (html) that we then handle  here, and elsewhere.

src/v2/System/Router/ErrorBoundary.tsx
damassi: To elaborate further with a better example, the modal with reload is helpful for when there are page timeouts (often with mobile), and this modal appears, saying things have timed out, and then there's a reload button that the user can click

src/v2/System/Router/buildServerApp.tsx
```diff
@@ -255,7 +255,12 @@ export function buildServerApp(
         // extractor.getLinkTags()
 
         // Get serializable Relay data for rehydration on the client
-        const _relayData = await (relayEnvironment.relaySSRMiddleware as RelayServerSSR).getCache()
+        let _relayData = [] as any[]
+        try {
+          _relayData = await (relayEnvironment.relaySSRMiddleware as RelayServerSSR).getCache()
+        } catch (error) {
+          console.error(error)
```
dzucconi, line 262: So what was happening is that:

404 errors within the routing framework would hit `RenderStatus => RenderError`, and since this was handled it'd be rendered normally, so on the server-side that means `buildServerApp`. But since we 404'd I guess there's nothing in the cache so this `getCache` call was throwing.

When this threw, that meant that ultimately the error would get propagated to the `errorHandlerMiddleware`, and since initially there was a 404 error which started this process, that status code was getting set correctly.

By catching this error here the `ErrorPage` renders normally within the `AppShell` — BUT we're not setting the status code anywhere so it's rendering with a `200`.

src/v2/System/Router/RenderStatus.tsx
```diff
@@ -77,14 +77,15 @@ export const RenderError: React.FC<{
   }
 
   const message =
-    (process.env.NODE_ENV || sd.NODE_ENV) === "development"
+    getENV("NODE_ENV") === "development"
       ? String(props.error.data)
-      : "Internal error"
+      : "Internal Error"
```
damassi, line 82: Adding right after this line 
```tsx
if (typeof window === "undefined") {
    throw new HttpError(props.error.status, message)
  }
```
will yield proper status and pass things along to errorMiddleware

damassi: (In any case, this `RenderStatus` is really more of a client-side thing, so i don't find it to be too weird, as long as the gotcha here has a comment)

src/v2/server.ts
damassi, line 38: 👍

damassi (commented):
Before this PR is merged we should document the control flow for error handling somewhere while it is fresh and top-of-mind. Maybe inside of `errorHandler`? Or inside a markdown file in /docs. Wherever it might be most discoverable for someone digging later (docs folder, strangely, is often less discoverable than code -- you get my drift) 

The other thought is we should add a minimal nav bar and footer to the server-side error page just so things are filled out a bit visually. 

Amazing work wrangling this into a much better state 👍

src/v2/System/Router/ErrorBoundary.tsx
damassi: Ah, i see we still have the network timeout [here](https://github.com/artsy/force/pull/9153/files#diff-45adeffd54fe16decbf8e730d662c0a6569d769b0566d2016b3b22f894bd76ccR8). This error is a bit less worrisome, though a 'reload page' button would still be nice UX if we see exactly the `asyncChunkLoadError` error since a page reload tends to fix it.

src/v2/Apps/Components/AppShell.tsx
dzucconi: This was so that we can use it sans-routing. In this PR I wound up not using the `AppShell` but in thinking about rendering a 'minimal' SSR only shell — we can actually completely re-use it with a few small changes (will follow up)

damassi: sweet

dzucconi: There's some issues with breakpoints in the footer and with login/out status apparently. Going with the minimal header style we use in a few other places, for now.

src/lib/middleware/errorHandler.tsx
```diff
@@ -0,0 +1,112 @@
...
+ * - 4xx errors outside of React routes
+ *   - Handled by `errorHandlerMiddleware`
+ * - 4xx errors within React routes
+ *   - Initially caught by `RenderStatus#RenderError`, wherein we throw
+ *     a new HTTPError with the status code, which ultimately handled by `errorHandlerMiddleware`
+ * - Thrown runtime error within the React application
+ *   - Handled by `errorHandlerMiddleware`
+ * - Errors raised in a route's `onServerSideRender`
+ *   - Handled by `errorHandlerMiddleware`
+ */
```
damassi, line 22: Thank you

damassi (approved):
A+ man

PR #9157, opened 2022-01-07
fix: Mark deprecated component as such

jonallured:
I did this:

```
$ ag -l ToastComponent | xargs sed -i '' "s/Toast/StaleToast/g"
```

And then updated the folder/filename and added that little breadcrumb comment for Future Us in case we stumble upon this again. A better solution would be to migrate that order page that uses it over to the Palette version of Toast but it's Friday I'm doing the best I can people.

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Components/Toast/ToastComponent.tsx
+++ b/src/v2/Components/Toast/ToastComponent.tsx
@@ -12,6 +12,10 @@ interface ToastComponentProps {
   onCloseToast: () => void
 }
 
+/**
+ * @deprecated in favor of our Toast components in @artsy/palette
+ * https://palette-storybook.artsy.net/?path=/story/components-toast--default
+ */
 const ToastComponent: React.FC<ToastComponentProps> = ({
   notificationAction,
   showNotification,


jacobherrington (commented):
CI is angry about a type declaration!

```
yarn run v1.22.5
$ tsc
src/v2/Apps/Settings/Routes/Shipping/Components/UserSettingsAddresses.tsx:59:9 - error TS2322: Type '{ commitMutation: CommitMutation; isCommittingMutation: boolean; isEigen?: boolean | undefined; mediator?: Mediator | undefined; notificationCount?: number | undefined; ... 14 more ...; inCollectorProfile: true; }' is not assignable to type 'IntrinsicAttributes & MappedFragmentProps<Pick<SavedAddressesProps, "onSelect" | "me" | "commitMutation" | "inCollectorProfile" | "onChangeAddressCount" | ... 5 more ... | "onShowToast">> & { ...; } & { ...; }'.
  Property 'onShowStaleToast' does not exist on type 'IntrinsicAttributes & MappedFragmentProps<Pick<SavedAddressesProps, "onSelect" | "me" | "commitMutation" | "inCollectorProfile" | "onChangeAddressCount" | ... 5 more ... | "onShowToast">> & { ...; } & { ...; }'.

59         onShowStaleToast={onShowStaleToast}
           ~~~~~~~~~~~~~~~~


Found 1 error.
```

jonallured:
Ok the Right Way to do this is with a `@deprecated` comment so switched over to that but I still think that `StaleToastComponent` is a fun way to deprecate something. 😝

dzucconi (approved):

PR #9162, opened 2022-01-09
security(colors.js): Lock version to 1.4 due to corruption

damassi:
See https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/ 

We don't import this lib directly, but it is an ephemeral dep, which still runs the risk of an automatic update. 

Screen Shot 2022-01-09 at 12 59 03 PM


damassi:
I think this is safe to merge - weekend staging copy is blowing up the acceptance test.

PR #9168, opened 2022-01-10
feat(articles): promotes /articles2 => /articles

dzucconi:
Promotes the articles index route. The existing Cypress specs still work. Deletes some old code (though, of course, I'm sure there's some more that can be deleted but is difficult to ascertain).


src/desktop/apps/articles/index.js
```diff
@@ -12,9 +13,6 @@ app.locals.resize = resize
 app.locals.crop = crop
 app.locals.toSentence = toSentence
 
-app.get("/posts", routes.redirectMagazine)
-app.get("/magazine", routes.redirectMagazine)
-app.get("/articles", routes.articles)
```
damassi, line 17: so much baggage everywhere

damassi (approved):
👏

PR #9172, opened 2022-01-11
fix: prevent open redirects

sonecabr:
This is a fix for a known open redirect bug. 
Case:
 - If you add '//example.coM' ( capital is required ), than you get redirected to example.com
 
This PR:
 - Apply a regex to block this redirect, so only valid internal uri are accepted.


ovasdi (approved):
LGTM.

Verified locally.

https://user-images.githubusercontent.com/29984068/150033959-e4abc3c3-505f-4650-9bbe-3afc581508ba.mov

ovasdi:
Looked at the acceptance failure and it seems unrelated.

PR #9176, opened 2022-01-11
chore: port content from /about to /about2

ovasdi:
This ports over content from `/about` to `/about2` with **very loose** styling and without responsive layout. Leaving the latter parts to the experts.

https://user-images.githubusercontent.com/29984068/149032546-ce57008e-3e4d-446d-aa9c-f88a35e14784.mp4


damassi:
Thank you!

PR #9186, opened 2022-01-12
File Structure and Hard Coded Banner

The-Beez-Kneez:
This is just to get us a general component outline as well as our dummy data in and fresh.

### DIFF
--- a/src/desktop/apps/diverselyOwnedGalleries/data.ts
+++ b/src/desktop/apps/diverselyOwnedGalleries/data.ts
@@ -1,97 +1,332 @@
 export const diverseGalleries = {
-  "blackOwnedGalleries": {
+  blackOwnedGalleries: {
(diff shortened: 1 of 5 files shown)


src/desktop/apps/diverselyOwnedGalleries/data.ts
```diff
@@ -0,0 +1,332 @@
...
+      id: "5ff756f6c41b29001206784f",
+      slug: "art-lead-her",
+    },
+    "fountain-house-gallery": {
+      name: "Fountain House Gallery",
+      id: "55a3ed79726169063d0000b7",
+      slug: "fountain-house-gallery",
+    },
+  },
+}
```
dzucconi, line 332: I do think we should follow this up by just exposing a `badges` field on the partner. Wouldn't be difficult and would save us having to ship this data source.

PR #9187, opened 2022-01-12
Allow for redirecting outside of artsy.net (within reason) after auth

sweir27:
## Problem
While our `/signup` and `/login` pages accept a `redirectTo` parameter so a user can be navigated to a specific page post-login/sign up, currently that `redirectTo` only works for relative paths within artsy.net.

This has worked OK for our current purposes. However, we'd like to use force's login page instead of the gravity-backed login page when signing up via oauth on non-artsy.net sites (like when signing in via live.artsy.net). This restriction makes that currently not possible, as if a user navigates to artsy.net/login from a non-artsy.net page, they cannot be redirected back.

## Solution
Add an environment variable, `AUTH_REDIRECT_ALLOWED_HOSTS`, where we can configure hosts that we'd like to allow redirects to after auth. This will likely include a combination of `api.artsy.net`, `artsy.net`, `www.artsy.net`, `staging.artsy.net`, ... etc. depending on the environment.

Some more comments inline!

**Update 1/13/22**
Right now, redirection happens on the client, which means we would have to expose our array of allowed hosts on the client. I added a commit to instead expose a route `/auth-redirect?redirectTo=`. After auth-ing with api.artsy.net, if you're not redirecting to an artsy.net URL, we send you to that internal route, which will decide whether or not to redirect you.


src/v2/Apps/Authentication/Utils/helpers.ts
```diff
@@ -143,9 +143,22 @@ export async function apiAuthWithRedirectUrl(
...
+
+  const allowedHosts = sd.AUTH_REDIRECT_ALLOWED_HOSTS?.split(",") || []
+
+  if (!allowedHosts.includes(redirectPath.hostname)) {
+    // If it's not, redirect to the homepage
+    redirectURL = new URL("/", sd.APP_URL)
+  }
+
+  // const strippedRedirectURL =
+  //   redirectURL.origin + redirectURL.pathname + redirectURL.search
```
sweir27, line 156: I'm not sure why the current behavior re-creates the URL via concatenation vs. just using the provided `URL`. I removed it, and tests pass/behavior looks OK.

If we want to keep this concatenation behavior, I can uncomment this line (and pass that `strippedRedirectURL` through).

src/v2/Apps/Authentication/Utils/helpers.ts
```diff
@@ -143,9 +143,22 @@ export async function apiAuthWithRedirectUrl(
   response: Response,
   redirectPath: URL
 ): Promise<URL> {
-  const redirectUrl = sd.APP_URL + redirectPath.pathname + redirectPath.search
+  let redirectURL = redirectPath
+
+  const allowedHosts = sd.AUTH_REDIRECT_ALLOWED_HOSTS?.split(",") || []
```
sweir27, line 148: With this behavior, all hosts need to be configured (including the existing `APP_URL`). Alternatively, we could add logic to allow for both `AUTH_REDIRECT_ALLOWED_HOSTS` and `APP_URL`, which will also allow for a less risky deploy.

src/lib/setup_sharify.js
```diff
@@ -34,6 +34,7 @@ sharify.data = _.extend(
     "APPLICATION_NAME",
     "ARTSY_EDITORIAL_CHANNEL",
     "AUCTION_ZENDESK_KEY",
+    "AUTH_REDIRECT_ALLOWED_HOSTS",
```
joeyAghion, line 37: Just confirming: this list is _not_ sent to the client?

src/v2/Apps/Authentication/Utils/helpers.ts
joeyAghion: It's reasonable to apply _different_ logic when redirecting internally vs. externally, so I think it would be nice if `AUTH_REDIRECT_ALLOWED_HOSTS` included only the external hosts, and `APP_URL` was accepted independently. (That also makes the empty default safer to deploy and reasonable for development.)

src/config.ts
```diff
@@ -29,6 +29,7 @@ export const APPLICATION_NAME: any = "force-staging"
 export const APPLY_URL: any = "http://apply.artsy.net"
 export const ARTSY_EDITORIAL_CHANNEL: any = "5759e3efb5989e6f98f77993"
 export const AUCTION_ZENDESK_KEY: any = null
+export const AUTH_REDIRECT_ALLOWED_HOSTS: any = "localhost"
```
joeyAghion, line 32: I might call this `ALLOWED_REDIRECT_HOSTS` for readability.

joeyAghion (approved):

src/v2/Apps/Authentication/Server/redirectPostAuth.ts
```diff
@@ -0,0 +1,24 @@
+export function redirectPostAuth({ req, res }) {
```
sweir27, line 1: This function handles the logic to determine whether or not a redirect from artsy.net is "safe".

First, we ensure that all of the configured allowed hosts are *.artsy.net domains (or localhost is OK, if you're in local development).

Then, we redirect to either the artsy.net homepage (if the URL is not allowed), or to the requested external URL.

src/v2/Apps/Authentication/Utils/helpers.ts
```diff
@@ -103,7 +103,13 @@ export const handleSubmit = async (
 
       const result = await apiAuthWithRedirectUrl(res, afterAuthURL)
 
-      window.location.assign(result.href)
+      if (result.origin === sd.APP_URL) {
```
sweir27, line 106: If you're trying to redirect to an internal URL, we allow that. Otherwise, you get sent to the server.

src/v2/Apps/Authentication/Utils/helpers.ts
sweir27: @joeyAghion what do you think about this?

src/v2/Apps/Authentication/Server/redirectPostAuth.ts
```diff
@@ -0,0 +1,24 @@
...
+  const redirectTo = req.query["redirectTo"]
+
+  const configuredAllowedHosts =
+    process.env.ALLOWED_REDIRECT_HOSTS?.split(",") || []
```
sweir27, line 5: @damassi what is our preferred convention for accessing env vars that are not in sharify (i.e. server-side only)? I'm not sure if this is it.

damassi: we should use `getENV`

src/v2/Apps/Authentication/Utils/helpers.ts
damassi: `getENV('APP_URL')`

src/v2/Apps/Authentication/Server/redirectPostAuth.ts
damassi: Though this also works!

damassi (dismissed):
This looks great. If there's a relevant test that can be written in [here](https://github.com/artsy/force/blob/5d9fb3bf3e820d823ccc3a4ac205825686c4e3bd/src/v2/Apps/Authentication/__tests__/authenticationRoutes.jest.tsx) we should, but otherwise LGTM!

joeyAghion (dismissed):

sweir27:
Updated based on comments! Planning to merge this in on Monday so I have some time to test all of the various flows.

src/v2/Apps/Authentication/__tests__/authenticationRoutes.jest.tsx
```diff
@@ -22,10 +23,23 @@ jest.mock("../Utils/helpers", () => ({
...
+  const mockGetENV = getENV as jest.Mock
+
+  mockGetENV.mockImplementation(key => {
+    switch (key) {
+      case "APP_URL":
+        return "https://artsy.net"
+      case "ALLOWED_REDIRECT_HOSTS":
+        return "off.artsy.net"
+    }
+  })
```
damassi, line 42: unless strictly necessary being here I'd move this block down into the test where its relevant, just so future tests don't accidentally collide with another's expectation

PR #9188, opened 2022-01-12
feat: add item for segment/onetrust integration

artsyjian:
Follows https://github.com/artsy/force/pull/8853 which integrated Segment/OneTrust. The integration requires us to assign cookie categories to relevant Segment Destinations. Update Best Practices doc about how to do that.


damassi:
Thank you for the update 👍

PR #9191, opened 2022-01-12
fix: Render dimensions when digits are present

jacobherrington:
JIRA Card: https://artsyproduct.atlassian.net/browse/GRO-710

In cases where a dimension field contains no digits ([production example](https://www.artsy.net/artwork/larva-labs-sealed-cryptopunk-number-934)), we shouldn't be rendering blank dimensions.

<details>
<summary>Broken Example - Before (local)</summary>

</details>
<details>
<summary>Broken Example - After (local)</summary>

</details>
<details>
<summary>Artwork with Dimensions and Editions - After (local)</summary>
image

</details>

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
@@ -1,5 +1,5 @@
 import { Box, Text } from "@artsy/palette"
-import { Component } from "react";
+import { Component } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 
 import { ArtworkSidebarSizeInfo_piece } from "v2/__generated__/ArtworkSidebarSizeInfo_piece.graphql"
@@ -11,16 +11,22 @@ export interface ArtworkSidebarSizeInfoProps {
 export class ArtworkSidebarSizeInfo extends Component<
   ArtworkSidebarSizeInfoProps
 > {
+  sizeInfoMissing = (dimensions, edition_of) => {
+    const dimensionsPresent =
+      /\d/.test(dimensions?.in) || /\d/.test(dimensions?.cm)
+
+    return !edition_of?.length && !dimensionsPresent
+  }
+
   render() {
     const {
       piece: { dimensions, edition_of },
     } = this.props
-    if (
-      !(edition_of && edition_of.length) &&
-      !(dimensions && (dimensions.in || dimensions.cm))
-    ) {
+
+    if (this.sizeInfoMissing(dimensions, edition_of)) {
       return null
     }
+
     return (
       <Box color="black60">
         {dimensions?.in && <Text variant="md">{dimensions.in}</Text>}
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
@@ -6,6 +6,7 @@ import {
   FilledOutMetadataMultipleEditionSets,
   FilledOutMetadataNoEditions,
   FilledOutMetadataOneEditionSet,
+  FilledOutMetadataNoSizeInfo,
   MetadataForAuctionWork,
 } from "../../../../__tests__/Fixtures/Artwork/ArtworkSidebar/ArtworkSidebarMetadata"
 import { ArtworkSidebarClassification } from "../../ArtworkSidebar/ArtworkSidebarClassification"
@@ -55,7 +56,7 @@ describe("ArtworkSidebarMetadata", () => {
       )
     })
 
-    it("displays dimentions", () => {
+    it("displays dimensions", () => {
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
       const html = wrapper.html()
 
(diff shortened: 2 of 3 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
```diff
@@ -17,7 +17,10 @@ export class ArtworkSidebarSizeInfo extends Component<
     } = this.props
     if (
       !(edition_of && edition_of.length) &&
-      !(dimensions && (dimensions.in || dimensions.cm))
+      !(
+        dimensions &&
+        (dimensions?.in?.match(/\d/) || dimensions?.cm?.match(/\d/))
```
dzucconi, line 22: What's the value here that's manifesting the behavior we're seeing?

dzucconi: Also I wouldn't be too sure that we shouldn't display non digit string values. If these are strings there's probably a pretty reasonable reason for that. Though, I can't think of one — I feel like @anandaroop would know if there's something weird we put in here that is none-the-less valid?

damassi:
Can we add a link to the JIRA ticket as well as a screenshot to the description field?

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
damassi: FWIW https://github.com/artsy/gravity/blob/main/app/models/util/dimensions.rb

anandaroop: CMS does currently enforce numeric-only input through validations like [these](https://github.com/artsy/volt/blob/9b6cbbde427d9ce712edc9a17fda340bc43a392a/app/javascript/apps/ArtworkForm/validation/width.ts)

But I'm pretty certain this was not always the case, and that we could very well have some non-numeric chars in older artworks.

(Also `\d` might be too strict, we'd need `\.` too at least, even in the most optimistic case)

anandaroop: Sure enough…

```
pp Artwork.find('4d8b92fc4eb68a1b2c000b3c').attributes.slice(:_id, :width, :height, :depth, :diameter, :created_at, :published)

{"_id"=>BSON::ObjectId('4d8b92fc4eb68a1b2c000b3c'),
 "width"=>"22 1/4",
 "height"=>"20 1/2",
 "depth"=>"",
 "created_at"=>2010-12-20 19:48:55 UTC,
 "published"=>true}
```

Not sure if we have anything more exotic than virgules though.

jacobherrington: Not sure I even meant to PR this - lol

I updated this with a more forgiving rule: the dimensions should include a digit somewhere.

dzucconi: 👍 could we add a spec too?

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
```diff
@@ -11,16 +11,22 @@ export interface ArtworkSidebarSizeInfoProps {
 export class ArtworkSidebarSizeInfo extends Component<
   ArtworkSidebarSizeInfoProps
 > {
+  sizeInfoMissing = (dimensions, edition_of) => {
+    const dimensionsPresent =
+      /\d/.test(dimensions?.in) || /\d/.test(dimensions?.cm)
+
+    return !(edition_of && edition_of.length) && !dimensionsPresent
```
damassi, line 18: Note you can use the existential operator to slightly simplify:

```ts
return !edition_of?.length && !dimensionsPresent
```

damassi (approved):
LGTM 👍

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
jacobherrington: Ooh, good call.

Pushed!

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarSizeInfo.tsx
damassi: @dzucconi - I think [this](https://github.com/artsy/force/pull/9191/files#diff-2657a287f0af87d5a20febe783f6ffb886fbd2c2897a83c6cba7fac96de7574bR122) should cover it, right @jacobherrington?

dzucconi: It apparently wasn't covering it before though?

jacobherrington:
@dzucconi Added a test! Thanks for the callout

src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
```diff
@@ -150,6 +151,17 @@ describe("ArtworkSidebarMetadata", () => {
     })
   })
 
+  describe("for artwork with no size info", () => {
+    it("does not render any size info", async () => {
+      // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
```
damassi, line 156: @jacobherrington - note that we shouldn't be using these flags. There's background info on this [here](https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-copying-and-try-to-fix--ts-expect-error-please_fix_me_strict_null_check_migration-flags)

PR #9195, opened 2022-01-14
fix(FX-3679): Issues with saving alert modal on mobile devices

dimatretyak:
Type: **Fix**
Jira ticket: [FX-3679]

### Description
the "save alert" button is not always available for pressing in create alert modal (the problem is especially relevant for Safari on iOS)

### Steps to Reproduce
* Go to artist page
* Select some filters
* Click “Create an alert” button
* Try to click on the "Save Alert" button

### Current Outcome
In some cases, the "Save Alert" button gets obscured by Safari’s bottom navigation bar

### Expected Outcome
The "Save Alert" button is always available for pressing and does not overlap with anything


dimatretyak:
There is a [Safari-specific](https://artsy.slack.com/archives/C9SATFLUU/p1642192136109400?thread_ts=1642169105.102300&cid=C9SATFLUU) "issue", but this is the standard behavior of the browser

dblandin (approved):
Nice! Glad the `ModalDialog` component fit our use case here 👌🏼

src/v2/Components/SavedSearchAlert/SavedSearchAlertModal.tsx
```diff
@@ -127,17 +127,18 @@ export const SavedSearchAlertModal: React.FC<SavedSearchAlertFormProps> = ({
 
   return (
     <FormikProvider value={formik}>
-      <Form onSubmit={formik.handleSubmit} role="form">
-        <Modal
-          show={visible}
+      {visible && (
```
mdole, line 130: hm wait do we still need the `<Form>` here? Or does it just work because it's wrapped in `<FormikProvider>` and has the `formik.handleSubmit()` call?

mdole (approved):
one thing to confirm, otherwise looking good!

src/v2/Components/SavedSearchAlert/SavedSearchAlertModal.tsx
dimatretyak: > Or does it just work because it's wrapped in <FormikProvider> and has the formik.handleSubmit() call

Yes, you are absolutely right. We don't need to use the `Form` component, since we manually call `formik.handleSubmit`

PR #9201, opened 2022-01-17
fix(FX-3679): Previously entered data for saved search alert remains after the modal is closed

dimatretyak:
Type: **Fix**
Jira ticket: [FX-3679]

### Steps to Reproduce
* Go to artist page
* Specify some filters
* Click on “Create Alert” button
* Specify name for alert in create alert modal
* Close modal
* Click on “Create Alert” button again

### Current Outcome
Previously specified name for the alert is displayed

### Expected Outcome
* Should clear the name input when create alert modal is closed
* Should return the “Email alerts” checkbox to checked state when create alert modal is closed
* Should return the “Push alerts” checkbox to not checked state when create alert modal is closed

#### Web
https://user-images.githubusercontent.com/3513494/149781314-16e80026-9afa-4aa7-99cf-8562bb427456.mp4

### mWeb
https://user-images.githubusercontent.com/3513494/149781372-20953762-78fa-4a87-a6b1-31fc68421b60.mp4


src/v2/Components/SavedSearchAlert/SavedSearchAlertModal.tsx
```diff
@@ -125,69 +86,121 @@ export const SavedSearchAlertModal: React.FC<SavedSearchAlertFormProps> = ({
...
+                      selected={values.email}
+                    />
+                  </Box>
+                  <Spacer mt={4} />
+                  <Box display="flex" justifyContent="space-between">
+                    <Text>Mobile Alerts</Text>
+                    <Checkbox
+                      onSelect={selected => setFieldValue("push", selected)}
+                      selected={values.push}
+                    />
```
dzucconi, line 190: Checkbox accepts children which it will use as the label: `<Checkbox {...}>Mobile Alerts</Checkbox>`

dimatretyak: If we use this approach, we will get a slightly different result that is expected by design

image

dzucconi: Ah, I see; in that case make sure you're always using a v3 variant with `Text` in this case; probably `md`

dzucconi: Though — I do wonder how we wound up with a layout that's not specified in the design system. :\ — I'd pose that question to design.

dimatretyak: > Ah, I see; in that case make sure you're always using a v3 variant with `Text` in this case; probably `md`

Done

dimatretyak: > Though — I do wonder how we wound up with a layout that's not specified in the design system. :\ — I'd pose that question to design.

Asked this question [here](https://www.figma.com/file/zbEFZqjKgKquZwtU8U44j9?node-id=3970:9533#141304787)

PR #9215, opened 2022-01-18
fix(payment): Buyer can't accept seller's offer when payment requires authentication

laurabeth:
Addresses [TX-72] Pair w/ @starsirius on this fix.

The bug originally demonstrated in this ticket was resolved by the time I got to it, but there was still a secondary bug to be addressed: after the user went through the card verification process they were redirected back to the payment screen.

The issue was solved by removing a redirect to the `/status` route and instead resubmitting the form for the user. 

[TX-72]: https://artsyproduct.atlassian.net/browse/TX-72?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rquartararo (approved):
LGTM 👍

starsirius (approved):
It took a lot longer to reproduce than the actual fix. Thanks for the persistence!! 💪 

Separately I wonder if there is a good strategy to test this kind of behavior.

PR #9216, opened 2022-01-18
feat(NX-3014): Scroll to focus functionality when entering custom offer price

dariakoko:
Added scroll to focus functionality when entering custom offer price.

Screenshot before toggle or text input was pressed:
image

Screenshot after toggle or text input was pressed:
image

### DIFF
--- a/src/v2/Apps/Order/Components/PriceOptions.tsx
+++ b/src/v2/Apps/Order/Components/PriceOptions.tsx
@@ -106,7 +106,7 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
   // trackClick("We recommend changing your offer", priceOptions[0]?.value!)
 
   const { scrollTo } = useScrollTo({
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -143,7 +150,10 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
                   id="OfferForm_offerValue"
                   showError={showError}
                   onChange={setCustomValue}
-                  onFocus={onFocus}
+                  onFocus={() => {
+                    onFocus
```
damassi, line 154: I think this is now missing an invocation (`onFocus()`)

dariakoko: Right, thanks for spotting. Fixed it

jo-rs (approved):
Looks and works great! Nice, @dariakoko! 🥇

PR #9223, opened 2022-01-19
feat(works-for-you2): Rebuild page 

damassi:
Addresses https://artsyproduct.atlassian.net/browse/GRO-740 

This rebuilds the works for you page and brings it into the new framework. The new FilterSelect palette component worked well here and will follow up with the other spots in the ArtworkFilter. 

wfy

Seriously struggling to figure out how to properly test using RTL (and also relay test data custom fixtures aren't working as expected anymore??), but will follow up in a separate PR. 

cc @artsy/grow-devs


src/v2/Apps/_WorksForYou2/WorksForYou2App.tsx
```diff
@@ -1,9 +1,160 @@
...
+    } else {
+      router.push("/works-for-you2")
+    }
+  }
 
-export const WorksForYou2App: React.FC = () => {
   return (
     <>
-      <Text variant="xl">Works For You 2 App</Text>
+      <MetaTags title="Works For You | Artsy" description="TODO" />
```
damassi, line 57: TODO

package.json
```diff
@@ -90,7 +90,7 @@
     "@artsy/fresnel": "3.2.1",
     "@artsy/gemup": "0.1.0",
     "@artsy/multienv": "^1.2.0",
-    "@artsy/palette": "18.12.2",
+    "@artsy/palette": "18.15.1",
```
dzucconi, line 93: Needs a rebase with main

src/v2/Apps/_WorksForYou2/WorksForYou2App.tsx
dzucconi: The title should be `"Works by Artists You Follow | Artsy"`, and I don't think there is a description.

src/v2/Apps/_WorksForYou2/Components/WorksForYou2ArtistFeed.tsx
```diff
@@ -0,0 +1,156 @@
...
+      setLoading(false)
+    })
+  }
+
+  return (
+    <>
+      <EntityHeader
+        name={artist.name!}
+        meta={meta}
+        imageUrl={avatarImageUrl}
```
dzucconi, line 49: Instead of giving this `imageUrl` pass it `image` with an object containing a `src`, `srcSet`, and `lazyLoad`

dzucconi (approved):
Looks good. Minor comments. We also are getting rid of this CTA?

damassi:
Ah, good catch about the CTA, will fix 👍

damassi:
Updated: 

Screen Shot 2022-01-20 at 2 56 05 PM

PR #9226, opened 2022-01-19
feat(new-routes): Added in the Preferences2 route

The-Beez-Kneez:
This is the scaffolding for the upcoming Interest-based preference center work. Set up our development route under `/preferences2`

### DIFF
--- a/src/v2/Apps/_Preferences2/preferencesRoutes.tsx
+++ b/src/v2/Apps/_Preferences2/preferencesRoutes.tsx
@@ -2,7 +2,7 @@ import loadable from "@loadable/component"
 import { AppRouteConfig } from "v2/System/Router/Route"
 
 const PreferencesApp = loadable(
-  () => import(/* webpackChunkName: "aboutBundle" */ "./PreferencesApp"),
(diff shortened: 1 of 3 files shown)


src/v2/Apps/_Preferences2/preferencesRoutes.tsx
```diff
@@ -0,0 +1,16 @@
+import loadable from "@loadable/component"
+import { AppRouteConfig } from "v2/System/Router/Route"
+
+const PreferencesApp = loadable(
+  () => import(/* webpackChunkName: "aboutBundle" */ "./PreferencesApp"),
```
damassi, line 5: We'll want to update the `webpackChunkName` prop to `preferencesBundle -- this is how webpack knows to split the js here into a new file that _only_ contains code related to the preference center

damassi (approved):
One minor fix, otherwise LGTM 👍

src/v2/Apps/_Preferences2/preferencesRoutes.tsx
The-Beez-Kneez: Updated.

PR #9228, opened 2022-01-19
fix: Show old offer submission for editioned artworks for now

jo-rs:
Quick fix to display the "old" offer submission for editioned artworks for now. We'll have a quick follow-up to make it work better with the new flow.


lilyfromseattle (approved):
👍

PR #9231, opened 2022-01-19
chore: Remove unnecessary TS annotation

jacobherrington:
This PR is a follow-up to #9191. It's not directly related to the functionality, but @damassi let me know we should be trying not to use this magic comment.

In the test file, there was one instance I couldn't figure out how to remove, happy to get some help there or punt it for now!

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
@@ -1,6 +1,7 @@
 import { ArtworkSidebarMetadata_Test_QueryRawResponse } from "v2/__generated__/ArtworkSidebarMetadata_Test_Query.graphql"
 import { renderRelayTree } from "v2/DevTools"
 import { graphql } from "react-relay"
+import { ReactWrapper } from "enzyme"
 import {
   EmptyMetadataNoEditions,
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
```diff
@@ -40,173 +41,146 @@ describe("ArtworkSidebarMetadata", () => {
...
+      expect(wrapper?.html()).toContain("Lot 210")
     })
 
     it("does not display lot number when present if work is not biddable(auction closed)", async () => {
       const closedAuctionArtwork = {
         ...MetadataForAuctionWork,
         is_biddable: false,
       }
+
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
```
jacobherrington, line 160: This is the one I left in!

damassi (approved):
Thanks for the follow-up (and cleaning up the rest of the spots, too) 👍 lets ship it

src/v2/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarMetadata.jest.tsx
damassi: No worries 👍

PR #9241, opened 2022-01-21
Redirect user after checkout, based on order source

lean8086:
Redirect user to artwork page, status page, or conversation based on the order source.

[NX-3040](https://artsyproduct.atlassian.net/browse/NX-3040)

This PR introduces the proposal of opening the congrats modal window on the artwork page [using the state on the router](https://github.com/artsy/force/compare/lean8086/NX-3040/change-confirmation-page-based-on-order-source#diff-1aaa2033a9ae43ffb3da1495105eac4324ed37ff31bbf19c13da79955ac4da47R153). To be challenged on [NX-3041](https://artsyproduct.atlassian.net/browse/NX-3041)


src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -122,13 +124,34 @@ export class ReviewRoute extends Component<ReviewProps> {
...
+          )
+        }
+        // Make offer from artwork page redirects to the artwork page
+        const artworkId = get(
+          order,
+          o => o.lineItems?.edges?.[0]?.node?.artwork?.slug
+        )
+        return router.push({
+          pathname: `/artwork/${artworkId}`,
+          state: { offerOrderHasBeenSubmitted: true },
```
lean8086, line 153: This is a proposal to open the congrats modal window on the artwork page using the state on the router. To be challenged on [NX-3041](https://artsyproduct.atlassian.net/browse/NX-3041)

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -122,13 +124,34 @@ export class ReviewRoute extends Component<ReviewProps> {
...
+                `/user/conversations/${order.conversation.internalID}`
+              )
+            : router.push(`/orders/${order.internalID}/status`)
+        }
+        // Buy-mode order redirects to the status page
+        if (order.mode !== "OFFER") {
+          return router.push(`/orders/${order.internalID}/status`)
+        }
+        // Make offer in inquiry redirects to the conversation page
+        if (order.source === "inquiry" && !isEigen) {
```
lean8086, line 141: Looking again at it, should Eigen redirect to /status here?

src/v2/Apps/Order/Routes/__tests__/Review.jest.tsx
```diff
@@ -69,14 +72,19 @@ class ReviewTestPage extends OrderAppTestPage {
 
 describe("Review", () => {
   beforeAll(() => {
-    window.sd = { STRIPE_PUBLISHABLE_KEY: "" } as GlobalData
+    window.sd = { STRIPE_PUBLISHABLE_KEY: "" }
     ;(useTracking as jest.Mock).mockImplementation(() => ({
       trackEvent: jest.fn(),
     }))
+    ;(useSystemContext as jest.Mock).mockImplementation(() => ({
```
damassi, line 79: Rather than having to parenthetically typecast each spot where this is used would suggest creating a reusable variable 

```tsx
const mockUseSystemContext = useSystemContext as jest.Mock
```

src/v2/Apps/Order/Routes/Review/index.tsx
lean8086: Update: I tried MOI on Eigen and it goes to /status

lean8086: The app’s checkout flow works on a webview inside a modal on Eigen. There seems to be difficult (or impossible?) to close the modal window from inside a webview, so that’s why we currently take the user to the congrats page (/status) instead of the conversation on MOI. See https://github.com/artsy/force/pull/7974.

jo-rs (approved):
Nice! Looks great, @lean8086! 💯

lidimayra (approved):

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -52,6 +52,7 @@ export interface ReviewProps extends SystemContextProps {
   dialog: Dialog
   commitMutation: CommitMutation
   isCommittingMutation: boolean
+  user?: User
```
araujobarret, line 55: you don't need this,`SystemContextProps` already provides it and it's not possible to be undefined from there

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -123,13 +124,34 @@ export class ReviewRoute extends Component<ReviewProps> {
...
-        this.props.order.conversation && !this.props.isEigen
-          ? this.props.router.push(
-              `/user/conversations/${this.props.order.conversation.internalID}`
-            )
-          : this.props.router.push(
-              `/orders/${this.props.order.internalID}/status`
-            )
+        const { order, router, user, isEigen } = this.props
+
+        if (!userHasLabFeature(user, "Make Offer On All Eligible Artworks")) {
```
araujobarret, line 129: we can improve the if's here once the feature flag is retired, the isEigen check can be done before all and make everything easier to read

araujobarret (approved):
all non-blocking comments, the interface one is easy to fix, good job!

src/v2/Apps/Order/Routes/Review/index.tsx
lean8086: Good to know! Will remove this in the incoming PR.

PR #9242, opened 2022-01-21
fix(settings2): delete q/a

dzucconi:
This just picks up from some of the comments I left on the original PR

* Fixes the layout to match other tabs
* Adds loading button feedback
* Log out after successful delete + before redirecting
* Refactors to use Formik instead of state


src/v2/Apps/Settings/Routes/DeleteAccount/DeleteAccountRoute.tsx
```diff
@@ -1,83 +1,116 @@
...
+        <Formik
+          validateOnMount
+          initialValues={{
+            explanation: "",
+            confirmation: false,
+          }}
+          validationSchema={Yup.object().shape({
+            explanation: Yup.string().min(1).required(),
+            confirmation: Yup.boolean().oneOf([true]),
+          })}
```
dzucconi, line 38: Rather than manually managing state we can just leverage the validators.

src/v2/Apps/Settings/Routes/DeleteAccount/DeleteAccountRoute.tsx
```diff
@@ -1,83 +1,116 @@
...
+                }
+              )
+
+              sendToast({
+                variant: "success",
+                message: "Profile deleted successfully",
+                description: "Redirecting you to the Artsy homepage",
+              })
+
+              await logout()
```
dzucconi, line 57: Previously this would redirect to the homepage and the recent bugfix to check for valid logins would then log you out and refresh. This avoids the extraneous refresh.

src/v2/Apps/Settings/Routes/DeleteAccount/DeleteAccountRoute.tsx
```diff
@@ -1,83 +1,116 @@
...
+
+              window.location.href = "/"
+            } catch (err) {
+              sendToast({
+                variant: "error",
+                message: "There was a problem",
+                description: err?.[0]?.message ?? err.message,
```
dzucconi, line 64: No need to parse JSON, not sure what was going on with that.

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsEmailPreferences/SettingsEditSettingsEmailPreferences.tsx
```diff
@@ -28,7 +28,7 @@ export const SettingsEditSettingsEmailPreferences: React.FC<SettingEditSettingsE
 
   const handleSelect = async newEmailFrequency => {
     const variables = { emailFrequency: newEmailFrequency }
-    try{
+    try {
```
dzucconi, line 31: How is code without prettier making it into the codebase?

damassi: A lot of folks `--no-verify` and don't use VSCode (or hate prettier). But this is the minority, so its a bit heavy handed to fail CI on a prettier check.

PR #9243, opened 2022-01-22
Allow for auth-ing from the API if logged into force

sweir27:
Still on the quest to allow us to use force's login form to auth into api.artsy.net! (Which will allow us to deprecate gravity's /log_in form, and we'll have one source of truth).

This PR fixes an issue that occurs when a user is logged into force, but _not_ into api.artsy.net.

Although we automatically log you into api.artsy.net after you log into force, it's possible to log out of api.artsy.net while still being logged into force. This can happen by:
- Clicking the "Sign out" button from api.artsy.net 
- If you were previously logged in but now the app you're logging into requires 2FA
- ... other pathways!

When that happens, we enter into an infinite loop:
- When you attempt to authorize, Gravity will automatically redirect you to the login form (in this scenario, force's), if you're not logged into the API
- When you visit force's login page, if [you're already logged in](https://github.com/artsy/force/blob/main/src/v2/Apps/Authentication/Server/redirectIfLoggedIn.ts#L5-L7) (to _force_), it'll automatically redirect you to the supplied `redirectTo` URL, in this case Gravity's authorize endpoint
- Then Gravity will see that you're still not logged in, and redirect you to force's login page...
- Which will redirect you back...
- And on and on

This PR fixes that ☝️ issue by having the `/login` and `/signup` routes accept a query parameter, `api_login`. If passed in, we allow you to access the form, even if you're already logged into force.

I'm interested in feedback on this. It's a blunt tool. It assumes that you won't access the `/login` route with `api_login=true` directly, only through Gravity's oauth flow (which only sends you there if you're not logged in). You could be logged into api.artsy.net, visit `/login?api_login=true`, and still see a login form.

### DIFF
--- a/src/v2/Apps/Authentication/Utils/useAuthForm.ts
+++ b/src/v2/Apps/Authentication/Utils/useAuthForm.ts
@@ -17,12 +17,12 @@ export function useAuthForm({ canonical, pageTitle, type }: UseAuthFormProps) {
   const {
     action,
     afterSignUpAction,
-    api_login,
     contextModule,
     copy: copyQueryParam,
     destination,
     intent,
     kind,
+    oauthLogin,
     objectId,
   } = match.location.query as ModalOptions
 
@@ -55,12 +55,12 @@ export function useAuthForm({ canonical, pageTitle, type }: UseAuthFormProps) {
   const options: ModalOptions = {
     action,
     afterSignUpAction,
-    api_login,
     contextModule,
     copy: copyQueryParam,
     destination,
     intent,
     kind,
+    oauthLogin,
     objectId,
     redirectTo,
     signupReferer,
--- a/src/v2/Apps/Authentication/Utils/useAuthForm.ts
+++ b/src/v2/Apps/Authentication/Utils/useAuthForm.ts
@@ -17,6 +17,7 @@ export function useAuthForm({ canonical, pageTitle, type }: UseAuthFormProps) {
   const {
     action,
     afterSignUpAction,
+    api_login,
     contextModule,
     copy: copyQueryParam,
     destination,
@@ -54,6 +55,7 @@ export function useAuthForm({ canonical, pageTitle, type }: UseAuthFormProps) {
   const options: ModalOptions = {
     action,
     afterSignUpAction,
+    api_login,
     contextModule,
     copy: copyQueryParam,
     destination,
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Authentication/Utils/useAuthForm.ts
```diff
@@ -54,6 +55,7 @@ export function useAuthForm({ canonical, pageTitle, type }: UseAuthFormProps) {
   const options: ModalOptions = {
     action,
     afterSignUpAction,
+    api_login,
```
sweir27, line 58: This is necessary to add here so that if you navigate between `/login` and `/signup` via the links on the form, the query parameter is passed through.

Also-- I'm not sure whether to prefer camelcase or snakecase here. The parameter originates from gravity which is why I chose snake_case, but I'm not tied to it.

damassi: Let's do camel case -- the loose reason being, after a parameter is passed into a system the new system owns it; it can then get passed around further without knowledge of the originating system and we'd want to stay closest to the idioms within the new owner.

damassi (commented):
This looks good to me, but the question remains as to whether infinite loops are still possible via some other future unaccounted for situation? If so, a comment might be good somewhere.

src/v2/Apps/Authentication/Utils/useAuthForm.ts
joeyAghion: +1 to consistent camel case. If querystring parameters are the "API" exposed by Force, Gravity should consume it according to Force's conventions.

src/v2/Components/Authentication/Types.ts
```diff
@@ -66,7 +66,10 @@ export interface ModalOptions {
    * }
    */
   afterSignUpAction?: AfterSignUpAction
-
+  /*
+   * Whether or not the user is using the form via gravity's oauth flow.
+   */
+  api_login?: boolean
```
joeyAghion, line 72: "API" doesn't have much to do with this (it just happens that the API host is responsible for oauth). Could this be renamed `oauthLogin` or something that makes its intent obvious like `resetLogin`?

joeyAghion (approved):
It's conventional, in my experience, for login pages to redirect users who are already logged in back to the application. I'm not sure if this is for security reasons, or simplicity (it's confusing and potentially buggy to start accepting new credentials and updating the session of an already-authenticated session). Is there a way we can _log out_ the existing session before presenting the form in these cases?

joeyAghion (approved):

PR #9245, opened 2022-01-24
chore: deprecate clickedArtistSeriesGroup helper

gkartalis:
# Description

This PR resolves [FX-3345](https://artsyproduct.atlassian.net/browse/FX-3345)

Deprecation of clickedArtistSeriesGroup helper

[This](https://github.com/artsy/cohesion/pull/298) is the cohesion related PR, should be reviewed and merged after this.


src/v2/Apps/Collect/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.jest.tsx
```diff
@@ -88,19 +88,23 @@ describe.skip("ArtistSeriesRail", () => {
...
-        horizontal_slide_position: 2,
-        type: "thumbnail",
-      })
+      expect(trackEvent.mock.calls[0]).toMatchInlineSnapshot(`
+            Array [
+              Object {
+                "action": "clickedArtistSeriesGroup",
+                "context_module": "context-module",
+                "context_page_owner_id": "context-page-owner-id",
+                "context_page_owner_slug": "context-page-owner-slug",
```
mdole, line 97: hm, this is surprising to me - shouldn't the snapshot have matched the old test expectations (line 91 - 102 on the left)? Especially since we explicitly set the ID, slug, and page owner type on lines 60 - 62 above. Is there something I'm missing here @gkartalis ?

src/v2/Components/ArtistSeriesRail/__tests__/ArtistSeriesItem.jest.tsx
```diff
@@ -42,19 +42,23 @@ describe("Artist Series Rail Item", () => {
...
+                "context_page_owner_type": "artist",
+                "curation_boost": true,
+                "destination_page_owner_id": "internal-id",
+                "destination_page_owner_slug": "slug",
+                "destination_page_owner_type": "artistSeries",
+                "horizontal_slide_position": 2,
+                "type": "thumbnail",
+              },
+            ]
+          `)
```
mdole, line 61: same questions here as above

mdole (commented):
props for chipping away at these tickets @gkartalis! I gotta pick up a couple as well, keep things moving forward :)

Left a question about expected test results, let me know if you want to discuss or pair!

src/v2/Apps/Collect/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.jest.tsx
gkartalis: great catch! So in this case maybe I should revert everything since the tests are marked with a `skip` thus this test file will never run.

src/v2/Components/ArtistSeriesRail/__tests__/ArtistSeriesItem.jest.tsx
gkartalis: in this case I just changed the way we assert on analytics from `toHaveBeenCalledWith` to -> `toMatchInlineSnapshot`.

The reason for that is that we discussed it a while ago with @pvinis and decided (as a result of [this rfc](https://github.com/artsy/cohesion/issues/212)) that it would be better to assert on analytics with `toMatchInlineSnapshot`

mdole: oop yep you're totally right! sorry I missed this the first time around 😅

src/v2/Apps/Collect/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.jest.tsx
mdole: oh yeah...huh. I mean, seems like maybe we should either a) remove the test suite or b) fix it, since skipped tests are so icky. definitely would expand the scope of this PR though if we choose to fix it. I'll leave it up to you!

mdole (approved):

src/v2/Apps/Collect/Routes/Collection/Components/CollectionsHubRails/ArtistSeriesRail/__tests__/ArtistSeriesRail.jest.tsx
gkartalis: Cool, I will completely delete them and will take it on me to re-write the tests in another pr (goes straight into my #ff list)

gkartalis:
#squashOnGreen

pvinis:
awesome! thanks George!

PR #9248, opened 2022-01-24
feat(settings2): rebuilds shipping tab

dzucconi:
Just a quick little rebuild of this tab since the existing form was messy, had other concerns, had UX/UI issues. Was quicker just to rebuild it rather than attempt surgery on that.

<del>I need to make some updates to these mutations: All mutations really should return the parent objects (!!!) — so many mutations have this issue, particularly with Gravity mutations, and it makes it super annoying to update the UI after the mutation.</del> Done.

https://user-images.githubusercontent.com/112297/150845261-584732fb-af47-4ec8-8be3-12470094c549.mov

cc @jo-rs and @lean8086 — I see https://github.com/artsy/force/pull/9115 — but I don't think it's a great idea to consolidate, necessarily. At least when it comes to settings. I'm not sure what else the old form was used for. (Though, with additional fragment spreads in the mutation, this modal could work anywhere. I could extract this to give you a cleaner slate to work with.)


src/v2/Apps/Shipping/Routes/Shipping/ShippingApp.tsx
```diff
@@ -1,11 +1,12 @@
...
+import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { Title } from "react-head"
 import { ShippingApp_me } from "v2/__generated__/ShippingApp_me.graphql"
+// eslint-disable-next-line no-restricted-imports
 import { data as sd } from "sharify"
 import { Box } from "@artsy/palette"
 import { UserSettingsTabs } from "v2/Components/UserSettings/UserSettingsTabs"
-import { UserSettingsAddressesFragmentContainer as UserSettingsAddresses } from "v2/Apps/Settings/Routes/Shipping/Components/UserSettingsAddresses"
+import { UserSettingsAddressesFragmentContainer as UserSettingsAddresses } from "./UserSettingsAddresses"
```
dzucconi, line 9: Moved the old components back here.

damassi:
So @damon, this is a rebuild that lives under this route, and [this PR](https://github.com/artsy/force/pull/9115) is effectively separate? Is there any way we can share this stuff? It seems like there's a lot of good possible shared stuff in here.

dzucconi:
That PR is separate. The modal form itself could be separate. One would have to spread any relevant fragments into the mutations so that the UI updates though.

dzucconi:
Now with: https://github.com/artsy/force/pull/9248/commits/067f1ea7b045dcdfc2945923fd5c78c04b64490d

We get the UI updating on mutation:

https://user-images.githubusercontent.com/112297/150872046-8f4c6ab0-69db-4f80-a198-c61a70933f97.mov

Instead of manually manipulating the Relay store!

PR #9253, opened 2022-01-24
GRO-703: User sees preference center options matching comp 

The-Beez-Kneez:
This is the WIP update UI for the new incoming `Preferences Center`, including the functionality for the allowing selecting and unselecting checkboxes.

This encompasses the following JIRA tickets: 
[GRO-699]
[GRO-700]
[GRO-701]
[GRO-702]
[GRO-703]

New UI and Checkbox behavior:

https://user-images.githubusercontent.com/30025439/151579491-63edf338-f906-40c8-b040-440bfb81f57c.mov

[GRO-699]: https://artsyproduct.atlassian.net/browse/GRO-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-700]: https://artsyproduct.atlassian.net/browse/GRO-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-701]: https://artsyproduct.atlassian.net/browse/GRO-701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-703]: https://artsyproduct.atlassian.net/browse/GRO-703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GRO-702]: https://artsyproduct.atlassian.net/browse/GRO-702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/_Preferences2/PreferencesApp.tsx
```diff
@@ -1,9 +1,176 @@
...
-      <Text variant="xl">Preferences App</Text>
+      <Text variant="xl" mt={4} mb={6}>
+        Preferences Center
+      </Text>
+
+      <Separator mt={2} />
+
+      <GridColumns>
+        <Column
+          span={12}
```
dzucconi, line 27: So all this is doing is just creating one big column. What you want is two columns right? Like the first column contains the description and maybe spans 8. The second column spans the remaining, 4, and contains the checkbox.

src/v2/Apps/_Preferences2/PreferencesApp.tsx
```diff
@@ -1,9 +1,176 @@
...
+          pr={12}
+          mt={2}
+          display="flex"
+          justifyContent="space-between"
```
dzucconi, line 31: All these props can be deleted; just use `Column` alone with the `span` prop.

src/v2/Apps/_Preferences2/PreferencesApp.tsx
The-Beez-Kneez: Updated.

src/v2/Apps/_Preferences2/PreferencesApp.tsx
The-Beez-Kneez: Updated.

src/v2/Apps/_Preferences2/PreferencesApp.tsx
```diff
@@ -1,9 +1,208 @@
...
+            <Flex p={2}>
+              <Box>
+                <Text variant="md">Custom Alerts</Text>
+                <Text variant="sm" color="black60">
+                  A round up of updates on your favorite artists, chosen by you
+                </Text>
+              </Box>
+            </Flex>
+
+            <Separator mt={2}/>
```
The-Beez-Kneez, line 76: @dzucconi I am having issues utilizing the `<Join>` element to apply a `<Separator>` between the email options and the unsubscribe from all options. How would I successfully apply them between the `<Column>` elements?

damassi (commented):
Latest changes look great 👌

src/v2/Apps/_Preferences2/PreferencesApp.tsx
```diff
@@ -1,9 +1,164 @@
...
+      <Formik
+        initialValues={{
+          first: false,
+          second: false,
+          third: false,
+          fourth: false,
+          fifth: false,
+        }}
+        onSubmit={values => {
+          console.log(values)
```
damassi, line 29: (I'm assuming that a follow-up PR will handle this bit; this is just a lil bit of prework)

damassi (approved):
👌  🚢

src/v2/Apps/_Preferences2/PreferencesApp.tsx
The-Beez-Kneez: Yup, hoping to work on this chunk once we successfully update MP.

src/v2/Apps/_Preferences2/__tests__/PreferencesApp.jest.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PreferencesApp } from "../PreferencesApp"
+
+describe("PreferencesApp", () => {
+  it("renders the preference center", () => {
+    render(<PreferencesApp></PreferencesApp>)
+
+    expect(screen.getByText("Preferences Center")).toBeInTheDocument()
+  })
+
+  it.skip("allows user to uncheck all boxes with unsubscribe all", async () => {
```
damassi, line 13: Any reason we're skipping the test? Also, if there's any other behavior we can test here that would be good. (Totally fine with merging this and returning to tests later tho; don't want to hold up momentum)

dzucconi: I looked at this with @The-Beez-Kneez and it seemed like the component wasn't updating (the props looked fine). Not entirely sure what the underlying issue is.

damassi: Don't want to rehash this right now, but just use enzyme, we get better holistically better test coverage and its easier to write and read and more.

I'm very close to the opinion that we should let devs choose which testing system they want to use. Some might be better for react testing library, but we shouldn't be fighting with tests. We haven't had to in the past, aside from minor stuff like `flushPromiseQueue` hacks, which is entirely minor.

The-Beez-Kneez: I am skipping this for now, because @jacobherrington and I have plans to look at it and write something more covering as we progress. But I did not want to remove the test entirely, so that we have a launch base when we return to it.

If y'all are interested in telling @jacobherrington and I the best path forward for testing react components in Force, we are open to it. Out plan was to add in some Cypress tests.

damassi: We can add Cypress tests, but we should def be writing unit tests as well for this kind of functionality. If you can cover all behavior using react testing library, feel free, but also feel free to use what works best for you (as long as there's coverage).

jacobherrington: I will revisit this test and figure out what's going on.

The intention was to cover the functionality described in the cards with unit tests, then add an integration test at the end to catch regressions, etc.

jacobherrington: For posterity: https://github.com/artsy/force/pull/9291

PR #9254, opened 2022-01-25
refactor(artwork-filter): Swap in FilterSelect for artwork filters with search

damassi:
This is follow-up work to https://github.com/artsy/palette/pull/1099.

Swaps in the new `<FilterSelect>` component and deletes the old code. 

One spot that doesn't have a filterable section but should is Artists. That code is pretty crazy right now so not gonna touch it but will maybe look later. 

cc @artsy/grow-devs


src/v2/Components/ArtworkFilter/ArtworkFilters/ArtistNationalityFilter.tsx
```diff
@@ -11,12 +11,12 @@ export const ArtistNationalityFilter: React.FC<ArtistNationalityFilterProps> = (
 }) => {
   return (
     <ResultsFilter
+      expanded={expanded}
```
damassi, line 14: We're at the point where all of these filter components are generic enough to be generated from an array of config objects. Not sure if thats worth it or not, cause this is pretty readable and discoverable

src/v2/Components/ArtworkFilter/ArtworkFilters/Utils/useFilterSelectResults.ts
```diff
@@ -0,0 +1,54 @@
...
+    setFilter(facetName, selectedNationalities)
+  }
+
+  return {
+    handleFilterSelectChange,
+    items,
+    labelWithCount,
+    selectedItems,
+  }
+}
```
damassi, line 54: TODO: add a test here

src/v2/Components/ArtworkFilter/ArtworkFilters/Utils/sortResults.ts
```diff
@@ -0,0 +1,19 @@
...
+  const selectedFacets = allResults.filter(({ value }) => {
+    return selectedValues.includes(value)
+  })
+
+  return selectedFacets.concat(
+    allResults.filter(({ value }) => {
+      return !selectedValues.includes(value)
+    })
+  )
+}
```
damassi, line 19: i can export this from palette

dzucconi (dismissed):

src/v2/Components/ArtworkFilter/ArtworkFilters/Utils/useFilterSelectResults.ts
```diff
@@ -0,0 +1,54 @@
...
+  SelectedFiltersCountsLabels,
+  Slice,
+  useArtworkFilterContext,
+} from "v2/Components/ArtworkFilter/ArtworkFilterContext"
+import { useFilterLabelCountByKey } from "../../Utils/useFilterLabelCountByKey"
+
+export interface UseFilterSelectResultsProps {
+  facetName: keyof MultiSelectArtworkFilters
+  filtersCountKey: SelectedFiltersCountsLabels
+  label
```
damassi, line 13: oops

PR #9256, opened 2022-01-25
[Feat]: Display error msg when email is not valid in cons submissions

AlicanAkyuz:
### Description 

This PR updates the Contact Information step of the consignment submission flow in order to display an errror message under the email input when user types in an invalid email address. 

#### Video 

https://user-images.githubusercontent.com/42584148/150965202-a119daee-ff98-4dce-a12c-f1da67bfa7d4.mov

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
@@ -64,7 +64,7 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
       />
 
       {/* render email error separately to have its design aligned with PhoneNumberInput */}
-      {errors.email && values.email && (
+      {errors.email && (values.email || touched.email) && (
         <Text variant="xs" mt={0.5} color="red100">
           {errors.email}
         </Text>
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
@@ -62,9 +62,7 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
         onChange={handleChange}
         onBlur={handleBlur}
         error={touched.email && errors.email}
-        data-testid="emaill"
       />
-
       <PhoneNumberInput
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
```diff
@@ -63,6 +63,13 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
         onBlur={handleBlur}
       />
 
+      {/* render email error separately to have its design aligned with PhoneNumberInput */}
```
kiryl-zubarau, line 66: Have you tried to use input's `error` prop? if PhoneNumberInput error and Input error different let's update PhoneNumberInput error.
`error={touched.email && errors.email}`

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
```diff
@@ -63,6 +63,13 @@ export const ContactInformationForm: React.FC<ContactInformationFormProps> = ({
...
+      {errors.email && values.email && (
```
kiryl-zubarau, line 67: I suppose we also want to show the error if the field was touched(after input loses focus)

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/Components/ContactInformationForm.tsx
AlicanAkyuz: Yes, I've tried. if we use the error prop and want to have email and phone number input styles to be the same, we would need to:

- update the styles of Select component (for red border) in PhoneNumberInput
- add logic to dynamically update those styles when the Input next to Select is hovered on 

which is an overkill, in my opinion. but I'm fine with your suggestion if you think going through styles updates makes sense.

AlicanAkyuz: ok I'll give it a try

AlicanAkyuz: this works! thanks

kiryl-zubarau (approved):

iskounen (approved):

kiryl-zubarau (approved):

PR #9263, opened 2022-01-25
chore(passport): Move files from artsy-passport repo into Force

damassi:
Addresses https://artsyproduct.atlassian.net/browse/PLATFORM-3687 

This moves @artsy/passport into force. It was as simple as copying files over, and then updating main middleware import path from `@artsy/passport` to `lib/passport`. Tests all pass, login and so on works. 

cc @artsy/grow-devs


jacobherrington (approved):
I agree with this decision!

I think it makes sense to merge this if tests are passing and we will definitely notice any weird behaviors in staging.

mc-jones:
Looks good. Thanks for doing this! I would think we will also want to go ahead and bring @artsy/passport-local-with-otp in also, as a follow-up.

damassi:
I'll take care of the `passport-local-with-otp` now

PR #9265, opened 2022-01-25
feat(settings2): rebuilds payments tab

dzucconi:
Same deal as the shipping tab. Just easier to a quick and clean rebuild. Will take this off draft when the MP schema updates in main.

https://user-images.githubusercontent.com/112297/151063400-fb8886a9-2424-4885-82b1-bb28194b047a.mov

### DIFF
--- a/src/v2/Apps/Settings/Routes/Payments/Components/SettingsPaymentsMethod.tsx
+++ b/src/v2/Apps/Settings/Routes/Payments/Components/SettingsPaymentsMethod.tsx
@@ -1,0 +1,114 @@
+import {
+  Clickable,
+  CreditCardIcon,
+  Flex,
+  Text,
+  useToasts,
+} from "@artsy/palette"
+import { FC, useState } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { SettingsPaymentsMethod_method } from "v2/__generated__/SettingsPaymentsMethod_method.graphql"
+import { useDeleteCreditCard } from "../useDeleteCreditCard"
+
+enum Mode {
+  Pending,
+  Deleting,
+}
+
+interface SettingsPaymentsMethodProps {
+  method: SettingsPaymentsMethod_method
+}
+
+const SettingsPaymentsMethod: FC<SettingsPaymentsMethodProps> = ({
+  method,
+}) => {
+  const [mode, setMode] = useState(Mode.Pending)
+
+  const { sendToast } = useToasts()
+  const { submitMutation } = useDeleteCreditCard()
+
+  const handleClick = async () => {
+    setMode(Mode.Deleting)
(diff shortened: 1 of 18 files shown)


src/v2/Apps/Settings/Routes/Payments/Components/SettingsPaymentsMethodForm.tsx
```diff
@@ -0,0 +1,269 @@
...
+                    <Message variant="error">
+                      {status.message ||
+                        "Something went wrong. Please try again."}
+                    </Message>
+                  </Column>
+                )}
+              </GridColumns>
+
+              {/* TODO: Disable -1 tabindex in focus lock */}
+              {/* Modal footer button is outside the form element. Hidden button supports <enter> */}
```
dzucconi, line 257: Going to make this update in Palette now.

src/v2/Apps/Settings/Routes/Payments/Components/SettingsPaymentsMethod.tsx
```diff
@@ -0,0 +1,114 @@
...
+}
+
+interface SettingsPaymentsMethodProps {
+  method: SettingsPaymentsMethod_method
+}
+
+const SettingsPaymentsMethod: FC<SettingsPaymentsMethodProps> = ({
+  method,
+}) => {
+  const [mode, setMode] = useState(Mode.Pending)
```
damassi, line 25: We use this pattern in so many spots now -- its a good pattern -- but starting to wonder if we might not want to migrate it to a hook? 

```tsx
const [mode, setMode, Pending, Deleting] = useMode(['pending', 'deleting']) 
```

dzucconi: I was also wondering about this but couldn't come up with an API that was more succinct than just typing it out manually. Lemme see if I can make this example API work.

damassi: every PR with this pattern I wonder the same thing, and then assume ^, but i keep on wondering about it so... maybe there's something there

dzucconi: Hm, yeah; really there's not a great way to do it type-wise without casting that array `as const`.

Really the most compact expression of this is simply to use strings, but specify the valid type:

```tsx
const [mode, setMode] = useState<'Pending' | 'Saving'>('Pending')
```

damassi: Maybe I can look at this today real quick

PR #9268, opened 2022-01-25
chore(lib): Move passport-local-with-otp into force

damassi:
Follow up to https://github.com/artsy/force/pull/9263, which moves `@artsy/passport-local-with-otp` directly into force, so we don't have to manage more libs. 

Tests all pass: 

Screen Shot 2022-01-25 at 2 02 41 PM


src/lib/passport/lib/passport/index.js
```diff
@@ -2,54 +2,62 @@
...
-const FacebookStrategy = require('passport-facebook').Strategy;
-const AppleStrategy = require('passport-apple');
-const LocalWithOtpStrategy = require('@artsy/passport-local-with-otp').Strategy;
-const callbacks = require('./callbacks');
-const { serialize, deserialize } = require('./serializers');
-const opts = require('../options');
+const passport = require("passport")
+const FacebookStrategy = require("passport-facebook").Strategy
+const AppleStrategy = require("passport-apple")
+const LocalWithOtpStrategy = require("lib/passport-local-with-otp/lib").Strategy
```
damassi, line 8: ☝️ only actual change

joeyAghion (dismissed):

damassi:
No clue why tests aren't passing -- all good locally. Will dig in a bit.

joeyAghion:
Just following up about this @damassi. I tried rerunning the build but the test failure looks repeatable and related.

damassi:
@joeyAghion - sorry, just now seeing this! Will return to this PR on Fri.

damassi:
Tests keep failing in CI and not sure why. Going to hand this off to #product-velocity

anandaroop:
🟢 but worth double-checking via review app https://relocate-passport-with-otp.artsy.net

joeyAghion:
Thanks @anandaroop. After some initial difficulties (relating to cross-app caches), I was able to log in with 2FA-enrolled users and without, and tested a few error scenarios as well.

joeyAghion (approved):

PR #9272, opened 2022-01-26
[NX-3057] Remove unneeded conditionals to hide Contact Gallery button

lidimayra:
[NX-3077]

We had a conditional to hide the Contact Gallery button on the artwork page in case the artwork is offerable/acquireable.
This condition doesn't seem to make sense as an artwork can't be selected as `offerable` **and** `inquireable` at the same time in CMS. However, it wouldn't also make any difference in the flow either.

Given we're changing the `offerable` scenario in MOOAEA, artworks that are `inquireable` are also `offerable`, therefore, we do not want to hide the button in these cases.

Besides this, we're handling these conditionals in Gravity at this point (see https://github.com/artsy/gravity/pull/14971). Gravity is ensuring Contact Gallery button is hidden in case `offer` field is `true` (and not if `offerable?` method is `true` as it was happening here).

Force shouldn't be responsible for these rules. In order to identify if `Contact Gallery` must be displayed or not, asking if an artwork is `inquireable` should be enough.

[NX-3077]: https://artsyproduct.atlassian.net/browse/NX-3077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -403,15 +403,12 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
       return <Separator />
     }
 
-    const shouldDisplayContactGalleryButton: boolean | null =
-      isInquireable && !isAcquireable && !isOfferable
-
     const shouldDisplayMakeOfferButton: boolean | null =
       isOfferable ||
       (isPriceListed && isOfferableFromInquiry && labFeatureEnabled)
 
     const shouldDisplayMakeOfferAsPrimary: boolean | null =
-      shouldDisplayMakeOfferButton && shouldDisplayContactGalleryButton
+      shouldDisplayMakeOfferButton && isInquireable
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -403,15 +403,12 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
...
-    const shouldDisplayContactGalleryButton: boolean | null =
-      isInquireable && !isAcquireable && !isOfferable
-
     const shouldDisplayMakeOfferButton: boolean | null =
       isOfferable ||
       (isPriceListed && isOfferableFromInquiry && labFeatureEnabled)
 
     const shouldDisplayMakeOfferAsPrimary: boolean | null =
-      shouldDisplayMakeOfferButton && shouldDisplayContactGalleryButton
+      shouldDisplayMakeOfferButton && isInquireable
```
araujobarret, line 411: easier to understand 🎖️

araujobarret (approved):

PR #9275, opened 2022-01-26
feat(hooks): Add useMode hook and update refs

damassi:
We've been using the `enum Mode` pattern for a little while now and i'm always thinking we could eliminate the extra code around enums. 

What this does is gives us a new `useMode` hook which is an alias to `useState`. While its not much, it does formalize the fact that we're using the mode pattern, and instead of enums uses string unions. Enums are a weird feature of TypeScript that can lead to problems in different contexts. 

cc @artsy/grow-devs

### DIFF
--- a/src/v2/Utils/Hooks/useMode.ts
+++ b/src/v2/Utils/Hooks/useMode.ts
@@ -1,0 +1,16 @@
+import { useState } from "react"
+
+/**
+ * This is basically just a wrapper around `useState` for intent signaling.
+ *
+ * @example
+ *
+ * type Mode = "Start" | "End"
+ *
+ * const [mode, setMode] = useMode<Mode>('Start')
+ *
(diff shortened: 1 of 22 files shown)


src/v2/Utils/Hooks/useMode.ts
```diff
@@ -0,0 +1,14 @@
+import { useState } from "react"
+
+/**
+ * This is basically just a wrapper around `useState` for intent signaling.
```
dzucconi, line 4: Smart

dzucconi (approved):

src/v2/Utils/Hooks/useMode.ts
damassi: 🤷 Its slightly silly, but it does feel-better code wise now

PR #9143, opened 2022-01-05
feat(GRO-644): Add sign up via Google to Force

jacobherrington:
This PR allows people to create an account with Google via Force (which we already do in Eigen).

There will be a follow-up task to update the sign-up and sign-in visuals after this in [GRO-639](https://artsyproduct.atlassian.net/browse/GRO-639).

<details>
<summary>Signing up with a Google account</summary>
<a href="https://www.loom.com/share/baf63346515d4dcd8533448f9ac9790d">
    <p>Watch Video</p>
    
</a>
</details>

Thanks, @damassi and @jonallured, for helping me get this done!

JIRA link: https://artsyproduct.atlassian.net/browse/GRO-644

PR #9282, opened 2022-01-27
Deploy

artsyit:
This is an automatically generated release PR!


jacobherrington:
The keys for https://github.com/artsy/force/pull/9143 are set in production.

PR #9285, opened 2022-01-27
chore: remove security page from v2 not needed anymore

TMcMeans:
@damassi before you merge, I just wanted to make sure it was the security page that was supposed to be deleted? On jira we noted that it was the education page:
https://artsyproduct.atlassian.net/browse/GRO-732?focusedCommentId=45496

I might have mixed up which pages should stay or go. Let me know before merging!


damassi:
👍

damassi:
Oh lol, i hit merge on one PR but meant the other, and now i see the description. But yeah! this is correct. The security page is actually covered by the `StaticPage` app, which renders things like `security`, `terms-of-service`, etc. I created this little app (the code deleted here) not knowing how we did the static page stuff, and once i realized that we could delete this now-unused code.

TMcMeans:
Right, I forgot we noted on the jira epic that the security and terms and conditions page were covered by the StaticPage app. Thanks for reminding me!

PR #9287, opened 2022-01-28
bugfix(artist-app): Fix missing overview tab

damassi:
Fixes an unreported bug I noticed on a few of these low-content artists where some overview content was appearing (and thus showing the overview page); however since we weren't checking for the condition we were hiding the overview tab. 

Screen Shot 2022-01-27 at 10 13 24 PM

cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Artist/Routes/Overview/Utils/hasOverviewContent.ts
+++ b/src/v2/Apps/Artist/Routes/Overview/Utils/hasOverviewContent.ts
@@ -4,16 +4,19 @@ export const hasOverviewContent = ({
   statuses,
   related,
   biographyBlurb,
+  insights,
 }: Pick<
   ArtistApp_sharedMetadata,
-  "statuses" | "related" | "biographyBlurb"
+  "biographyBlurb" | "insights" | "related" | "statuses"
 >): boolean => {
   const showArtistBio = biographyBlurb && Boolean(biographyBlurb.text)
   const showRelatedCategories = related?.genes?.edges?.length! > 0
+  const showCareerHighlights = insights?.length
 
   return Boolean(
     showArtistBio ||
       showRelatedCategories ||
+      showCareerHighlights ||
       statuses?.articles ||
       statuses?.cv ||
       statuses?.shows
--- a/src/v2/Apps/Artist/__tests__/ArtistApp.jest.tsx
+++ b/src/v2/Apps/Artist/__tests__/ArtistApp.jest.tsx
@@ -83,6 +83,7 @@ describe("ArtistApp", () => {
           Artist: () => ({
             biographyBlurb: null,
             related: null,
+            insights: null,
             statuses: {
               articles: false,
               cv: false,


damassi:
More discussion [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1643351582972189)

src/v2/Apps/Artist/Routes/Overview/Utils/hasOverviewContent.ts
```diff
@@ -4,16 +4,19 @@ export const hasOverviewContent = ({
...
+  "biographyBlurb" | "insights" | "related" | "statuses"
 >): boolean => {
   const showArtistBio = biographyBlurb && Boolean(biographyBlurb.text)
   const showRelatedCategories = related?.genes?.edges?.length! > 0
+  const showCareerHighlights = insights?.length
 
   return Boolean(
     showArtistBio ||
       showRelatedCategories ||
+      showCareerHighlights ||
```
damassi, line 19: As the artist page changes through time this approach is always gonna be a game of chase. Don't have any better ideas ATM but at some point we should re-review.

src/v2/Apps/Artist/ArtistApp.tsx
```diff
@@ -45,8 +45,26 @@ const ArtistApp: React.FC<ArtistAppProps> = ({ artist, children, match }) => {
   }
 
   const showOverviewTab = hasOverviewContent(artist)
-  const showArtworksTab = artist?.statuses?.artworks
-  const showAuctionResultsTab = artist?.statuses?.auctionLots
+  const showAuctionResultsTab =
+    artist?.statuses?.auctionLots || artist?.auctionResultsStatus?.edges?.length
```
joeyAghion, line 49: (Discussed a little [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1643376118098399?thread_ts=1643351582.972189&cid=C01ADJNCS5D).) I think we should fix the API's `display_auction_link` property, since its whole purpose is to help with this logic. It also respects the `Artist#hide_auction_link` property which presumably is important.

When I tested one of the artists referred to in [the ticket](https://artsyproduct.atlassian.net/browse/GRO-682), it already had a positive count and a visible tab. The other did not, which is possible when the artist's record has never been "touched" via follows, artworks, sales, or shows. We should probably refresh these counts periodically, rather than depending on those events. Even better, Diffusion could publish an event when there are new [published] auction results for an artist, and Gravity could subscribe to it and refresh then.

src/v2/Apps/Artist/ArtistApp.tsx
```diff
@@ -45,8 +45,26 @@ const ArtistApp: React.FC<ArtistAppProps> = ({ artist, children, match }) => {
...
+    // past auction results, and in that case, we show the artworks tab because
+    // we need to also show the auction results tab.
+  } else if (!showOverviewTab && showAuctionResultsTab) {
+    showArtworksTab = true
+
+    // By default, if there's no content, show artworks page with zerostate but
+    // also hide the tab.
+  } else if (!showOverviewTab && !showArtworksTab && !showAuctionResultsTab) {
+    showArtworksTab = false
+  }
```
dzucconi, line 67: `non-blocking`: I find assigning booleans like this kinda weird. Presumably this is equivalent to:

```js
const showArtworksTab =
  hasArtworks ||
  // We use a view of the artworks tab zero state even if there's no artworks
  // and hide the tab. However, often an artist has no artworks but a lot of
  // past auction results, and in that case, we show the artworks tab because
  // we need to also show the auction results tab.
  (!showOverviewTab && showAuctionResultsTab) ||
  // By default, if there's no content, show artworks page with zerostate but
  // also hide the tab.
  !(!showOverviewTab && !showArtworksTab && !showAuctionResultsTab);
```

But if the former is more clear; by all means.

dzucconi (approved):

src/v2/Apps/Artist/ArtistApp.tsx
damassi: Going with the former 😄. Verbose and old school, if ugly

PR #9292, opened 2022-01-31
Add a “See details.” text link that opens up the conversation sidebar

lean8086:
[NX-3069](https://artsyproduct.atlassian.net/browse/NX-3069)

Because users will be pointed to the inbox after making an offer (see [NX-3041](https://artsyproduct.atlassian.net/browse/NX-3041)), we need to make it clear for them where to see the details of the offer.

Twin PR on Eigen: https://github.com/artsy/eigen/pull/6139

image

[Figma file](https://www.figma.com/file/PgmxHwBnoOUS4ybwpK2mTN/Make-Offer-on-all-Eligible-Artworks?node-id=525%3A3575) for reference.

### DIFF
--- a/src/v2/Apps/Conversation/Components/ConversationMessages.tsx
+++ b/src/v2/Apps/Conversation/Components/ConversationMessages.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useEffect, useState } from "react";
+import { Fragment, useEffect, useState } from "react"
 import { createFragmentContainer } from "react-relay"
 import { graphql } from "relay-runtime"
 import { Spacer } from "@artsy/palette"
@@ -20,6 +20,7 @@ interface ConversationMessageProps {
   messages: ConversationMessages_messages
   events: ConversationMessages_events | null
   lastViewedMessageID?: string | null
+  setShowDetails: (showDetails: boolean) => void
 }
 type Order = NonNullable<
   NonNullable<
@@ -32,6 +33,7 @@ export const ConversationMessages = ({
   messages,
   events,
   lastViewedMessageID,
+  setShowDetails,
 }: ConversationMessageProps) => {
   let [messagesAndEvents, setMessagesAndEvents] = useState<MessageType[][]>([])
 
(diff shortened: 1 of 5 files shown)


src/v2/Apps/Conversation/Components/ConversationMessages.tsx
```diff
@@ -72,80 +74,81 @@ export const ConversationMessages = ({
...
+              />
+            )}
+
+            {[...messageGroup].reverse().map((message, messageIndex) => {
+              if (isRelevantEvent(message)) {
+                return (
+                  <OrderUpdateFragmentContainer
+                    key={`event-${messageIndex}`}
+                    event={message as any}
+                    setShowDetails={setShowDetails}
```
lean8086, line 105: This is the only line that changed in the return statement. Everything else is prettified.

src/v2/Apps/Conversation/Components/OrderUpdate.tsx
```diff
@@ -15,12 +16,17 @@ import { TimeSince } from "./TimeSince"
 import { OrderUpdate_event } from "../../../__generated__/OrderUpdate_event.graphql"
 export interface OrderUpdateProps {
   event: OrderUpdate_event
+  setShowDetails: (showDetails: boolean) => void
 }
 
-export const OrderUpdate: React.FC<OrderUpdateProps> = ({ event }) => {
+export const OrderUpdate: React.FC<OrderUpdateProps> = ({
+  event,
+  setShowDetails,
```
araujobarret, line 24: IMO this level of prop drilling makes the maintenance complicated, when we have to pass it down to the grandchildren I think we should provide a more elegant solution, like a shared hook among them, context API, store, etc.

araujobarret: I know this prop drilling was here before, so I don't wanna block the ticket, just flagging bc I've seen so many components we have to touch in order to call a common callback among conversation components.

@jo-rs can we have a refactor ticket for it? 💟 

another good example(from the pov that needs to refactor) is this [guy](https://github.com/artsy/force/blob/dd1088ea9a638fa5fd2e24bb7c2e4ec1b606a27b/src/v2/Apps/Conversation/Components/Conversation.tsx#L33) ideally we would just have something like a callback and the pagination/refetch container would handle when we load more, instead we're prop drilling `relay.environment` when we have hooks that you can use to import it, just closing pointing out that we need a smart way to handle shared callbacks and states(rn not needed) in the conversation context instead of keeping this prop drilling here.

src/v2/Apps/Conversation/Components/__tests__/ConversationMessages.jest.tsx
```diff
@@ -11,6 +11,7 @@ const { renderWithRelay } = setupTestWrapperTL({
       <ConversationMessagesFragmentContainer
         messages={me.conversation.messagesConnection}
         events={me.conversation.orderConnection}
+        setShowDetails={() => {}}
```
jo-rs, line 14: Would it make sense to set this to not only test if "See Details" is there, but what is triggering?

src/v2/Apps/Conversation/Components/OrderUpdate.tsx
```diff
@@ -30,6 +36,9 @@ export const OrderUpdate: React.FC<OrderUpdateProps> = ({ event }) => {
       message = `You sent ${isCounter ? "a counteroffer" : "an offer"} for ${
         event.offer.amount
       }`
+      if (!isCounter) {
+        action = { label: "See details", onClick: () => setShowDetails(true) }
```
jo-rs, line 40: 🥇 for making it modular.

src/v2/Apps/Conversation/Components/OrderUpdate.tsx
jo-rs: Oh, sorry, I missed these messages last week 😕 
Agreed @araujobarret , a hook or context might make sense here. I'll create a ticket for a refactor. Thank you! 🙏

jo-rs (approved):
Looks good to me! 💪 
I just have one small non-blocking question about the test.

src/v2/Apps/Conversation/Components/__tests__/ConversationMessages.jest.tsx
lean8086: Good call. I think it makes sense to test it 👍

jo-rs (approved):
Lovely. Thank you for making the changes! 💯

PR #9293, opened 2022-01-31
[feat]: AlicanAkyuz/SWA-182/Analytics for aftermath of an artwork submissions

AlicanAkyuz:
### Description 

Ticket: [SWA-182](https://artsyproduct.atlassian.net/browse/SWA-182)

This PR adds 2 tracking events, specifically for: 

- when user clicks on `Submit Another Artwork` 
- when user clicks on a `FAQ` 

after completing an artwork submission. 

Note that the above-mentioned ticket specifies another tracking event for `clickedArtworkGroup`, but this event[ already exists](https://github.com/artsy/force/blob/main/src/v2/Apps/Consign/Routes/MarketingLanding/Components/SoldRecently.tsx#L42) in the codebase. 

#### Video

https://user-images.githubusercontent.com/42584148/152376933-ea8bef4c-9e5d-4b99-914a-9eed6c34ced6.mov


patrinoua (approved):
Looks good!

src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
```diff
@@ -2,8 +2,25 @@ import { Button, Flex, Text, Spacer, Box } from "@artsy/palette"
...
+  const { match } = useRouter()
+  const { trackEvent } = useTracking()
+
+  const trackSubmitAnotherWorkClick = () =>
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.SubmitAnotherArtwork,
+      context_module: ContextModule.consignSubmissionFlow,
+      context_owner_type: OwnerType.consignmentSubmission,
+      submission_id: match.params.id,
+      user_email: isLoggedIn ? user?.email : "",
```
iskounen, line 20: 1. **question:** The schema that was requested from Data did not include this field. What will it be used for?
2. **suggestion:** We have access to an anonymous submission's email address at this point. Should we record the email address that was provided on the contact information step for both authenticated and anonymous submissions?

src/v2/System/Analytics/Schema/Values.ts
```diff
@@ -146,6 +146,12 @@ export enum ActionType {
    */
   ClickedArtworkGroup = "clickedArtworkGroup",
   ClickedBuyViewingGroup = "clickedBuyViewingRoom",
+
+  /**
+   * Artwork Submission
+   */
+  SubmitAnotherArtwork = "submitAnotherArtwork",
+  ClickedFAQ = "clickedFAQ",
```
iskounen, line 154: **question:** I was expecting to also see the `clickedArtworkGroup` event represented in this PR. Is that no longer needed?

iskounen: Ah! I read the PR description after sending in my review... 🤦‍♂️

src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
AlicanAkyuz: 1 - The expected schema specified in [this sheet](https://docs.google.com/spreadsheets/d/1QtlyK1XGi1QWSqqqT4mMmZUOYMfPxHlOjupNYZ6LCGo/edit#gid=0) from Data includes the fields `user_email` and `user_id` if user is logged in. I also see it in the body of this event in [Cohesion](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/Consignments.ts#L136). 

2- that's a good suggestion. But the data team might be expecting `user_email` when there is also a `user_id` sent. I'll raise this question in the ticket.

iskounen: We are looking at the same spreadsheet, but different sheets. The sheet titled "Web Submission Confirmation Page" doesn't have the `user_email` field, but the sheet titled "App submission flow" does.

I assume that we'll want it for both, but I'll just confirm in the same ticket.

src/v2/Apps/Consign/Routes/MarketingLanding/Components/FAQ.tsx
```diff
@@ -1,119 +1,155 @@
...
 export const FAQ: React.FC = () => {
+  const { trackEvent } = useTracking()
+
+  const trackClickedFAQ = (label: string) => {
+    trackEvent({
+      action_type: AnalyticsSchema.ActionType.ClickedFAQ,
+      context_module: ContextModule.consignSubmissionFlow,
+      context_owner_type: OwnerType.consignmentSubmission,
+      subject: label,
+    })
```
kiryl-zubarau, line 14: 2 pages use the component: /consign and /consign/submission/:id/thank-you, should we track the event for both of them?

AlicanAkyuz: thanks for this comment 🙏  I clarified with data team: we are tracking a FAQ click only after a submission is made now.

src/v2/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
```diff
@@ -81,7 +83,7 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
           action: ActionType.consignmentSubmitted,
           submission_id: submission.id,
           user_id: me?.internalID,
-          user_email: me?.email ?? undefined,
+          user_email: isLoggedIn && me?.email ? me.email : submissionEmail,
```
AlicanAkyuz, line 86: note: I changed this after abhiti's feedback: if a user is logged in we grab their email we have, and if they're not logged in then we grab the one they entered.

src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
```diff
@@ -57,3 +81,11 @@ export const ThankYou: React.FC = () => {
...
   )
 }
+
+export const ThankYouFragmentContainer = createFragmentContainer(ThankYou, {
+  submission: graphql`
+    fragment ThankYou_submission on ConsignmentSubmission {
+      userEmail
+    }
+  `,
+})
```
kiryl-zubarau, line 91: Are you sure that it works correctly? This query should fail since the user shouldn't be able to load submissions with the `submitted` state. I suppose it may work if the information comes from the Relay store, but if information comes from the server we should fix that.

AlicanAkyuz: it looks like it works, I've added a video in the description.

kiryl-zubarau: very interesting, I'll check it locally tomorrow.

kiryl-zubarau: I suppose it happens because you have admin rights.

AlicanAkyuz: thank you. do we not allow querying 'submitted' submissions, by the way? I wonder where this restriction is set.

kiryl-zubarau: Yes, we allow loading or updating the submission only if the submission has draft status. the logic here: https://github.com/artsy/convection/blob/4fd180fc614b67089eb3f203ffff329525481793/app/graphql/resolvers/submission_resolver.rb#L9

AlicanAkyuz: thank you for this, I had no idea. I'll test it with a non-admin user tomorrow morning.

AlicanAkyuz: tried with non-admin user: correct, no submission is received.

kiryl-zubarau (approved):
LGTM!

PR #9300, opened 2022-01-31
fix(GRO-725): prevents exposure of token to 3rd parties

dzucconi:
Closes [GRO-725](https://artsyproduct.atlassian.net/browse/GRO-725)

Main change in https://github.com/artsy/force/commit/451c3b9c47184da80e6dcd54e72c9dd2b0891e8b

Included a minor refactor to use Formik & improve client-side validation.

### DIFF
--- a/src/v2/Apps/Authentication/Routes/ResetPasswordRoute.tsx
+++ b/src/v2/Apps/Authentication/Routes/ResetPasswordRoute.tsx
@@ -8,12 +8,15 @@ import {
   PasswordInput,
   useToasts,
 } from "@artsy/palette"
-import React, { useState } from "react"
+import { Formik } from "formik"
+import React from "react"
 import { MetaTags } from "v2/Components/MetaTags"
 import { useRouter } from "v2/System/Router/useRouter"
 import { resetPassword } from "v2/Utils/auth"
 import { getENV } from "v2/Utils/getENV"
 import { useMode } from "v2/Utils/Hooks/useMode"
+import * as Yup from "yup"
+import { password } from "v2/Components/Authentication/Validators"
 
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Authentication/Routes/ResetPasswordRoute.tsx
```diff
@@ -86,56 +45,121 @@ export const ResetPasswordRoute: React.FC<ResetPasswordRouteProps> = () => {
...
-        alignItems="center"
-        flexDirection="column"
-        minHeight="100vh"
-        justifyContent="center"
-        mx="auto"
-        py={4}
-        maxWidth={440}
-        // @ts-ignore
-        onSubmit={handleSubmit}
+      <Formik
```
damassi, line 48: all in

src/v2/System/Server/sharifyHelpers.ts
```diff
@@ -64,7 +67,7 @@ function generateCustomSharifyScript(sharifyData) {
 
 export function getV2SharifyScript(sharifyData) {
   const script = generateCustomSharifyScript(
-    pick(sharifyData, V2_SHARIFY_ALLOWLIST)
+    pick(sharifyData, [...V2_SHARIFY_ALLOWLIST])
```
damassi, line 70: why spread here?

damassi (approved):

src/v2/System/Server/sharifyHelpers.ts
dzucconi: Since I changed it to a `const`, the lodash type complaining and not accepting me casting it `as string[]` for whatever reason.

PR #9302, opened 2022-02-01
docs: update docs to add an ENV var

dblandin:
- As of #8062, the desktop/mobile-specific config files have been consolidated
  into the src/config.ts file
- Add note to type a new sharify addition

cc/ @sonecabr @dzucconi


damassi:
Thank you for the doc update!

PR #9305, opened 2022-02-01
fix: update when the price too low notice shows up

jo-rs:
Based on a request from design to change the "Offer too low" notice always to be visible until the user has entered an amount higher than 80% of the listed price.

https://user-images.githubusercontent.com/7117670/152004741-628f0c74-9c0a-4bc6-be97-34d9e28c66c6.mov

cc @artsy/negotiate-devs

### DIFF
--- a/src/v2/Apps/Order/Components/__tests__/PriceOptions.jest.tsx
+++ b/src/v2/Apps/Order/Components/__tests__/PriceOptions.jest.tsx
@@ -124,15 +124,13 @@ describe("PriceOptions", () => {
         )
       )
       fireEvent.click(radios[3])
-      expect(trackEvent).toHaveBeenLastCalledWith(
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Components/__tests__/PriceOptions.jest.tsx
```diff
@@ -124,15 +124,13 @@ describe("PriceOptions", () => {
         )
       )
       fireEvent.click(radios[3])
-      expect(trackEvent).toHaveBeenLastCalledWith(
+      expect(trackEvent).toHaveBeenCalledWith(
```
jo-rs, line 127: I switched this since the click event, and the "viewed offer too low" event are triggered simultaneously.

lidimayra (approved):

lean8086 (approved):
LGTM!

PR #9306, opened 2022-02-01
fix: Allow users to sign in via Google

jacobherrington:
This took like 10 minutes, I'm PRing it just so users can log in with Google while I'm working [on the updated visuals](https://artsyproduct.atlassian.net/browse/GRO-639?atlOrigin=eyJpIjoiYjRkMWI4NDIwMzAyNGVmOTkzZDViNGY1Zjk2YjI5ZmYiLCJwIjoiaiJ9).

Before:
image

After:
image

### DIFF
--- a/src/v2/Components/Authentication/Views/LoginForm.tsx
+++ b/src/v2/Components/Authentication/Views/LoginForm.tsx
@@ -172,6 +172,11 @@ export class LoginForm extends Component<FormProps, LoginFormState> {
                       this.props.onFacebookLogin?.(e)
                     })
                   }}
+                  onGoogleLogin={e => {
+                    this.setState({ isSocialSignUp: true }, () => {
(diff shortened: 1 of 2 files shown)


src/v2/Components/Authentication/Views/LoginForm.tsx
```diff
@@ -172,6 +172,11 @@ export class LoginForm extends Component<FormProps, LoginFormState> {
                       this.props.onFacebookLogin?.(e)
                     })
                   }}
+                  onGoogleLogin={e => {
+                    this.setState({ isSocialSignUp: true }, () => {
+                      this.props.onGoogleLogin?.(e)
+                    })
+                  }}
```
damassi, line 179: @jacobherrington - since this is a new behavior/feature, we should add a test before merge

PR #9308, opened 2022-02-01
User on gallery detail page for black owned gallery sees ribbon

lilyfromseattle:
[GRO-771]

There might be a less repetitive solution and I'm open to suggestions, but everything I tried seemed overly complex or confusing so I just did the straightforward thing. The "Black Owned" category doesn't actually exist yet, so I used a different category to get the screenshots. 

Screen Shot 2022-02-01 at 1 36 16 PM
Screen Shot 2022-02-01 at 1 36 08 PM

[GRO-771]: https://artsyproduct.atlassian.net/browse/GRO-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Partner/PartnerApp.tsx
+++ b/src/v2/Apps/Partner/PartnerApp.tsx
@@ -9,6 +9,7 @@ import { PartnerMetaFragmentContainer } from "./Components/PartnerMeta"
 import { StickyProvider } from "v2/Components/Sticky"
 import { PartnerArtistsLoadingContextProvider } from "./Utils/PartnerArtistsLoadingContext"
 import { HttpError } from "found"
+import { Ribbon } from "v2/Components/Ribbon"
 
 export interface PartnerAppProps {
   partner: PartnerApp_partner
@@ -24,11 +25,15 @@ export const PartnerApp: React.FC<PartnerAppProps> = ({
     displayFullPartnerPage,
     isDefaultProfilePublic,
     partnerPageEligible,
+    categories,
   } = partner
 
   if (!isDefaultProfilePublic || !partnerPageEligible) {
     throw new HttpError(404)
   }
+
+  const isBlackOwned =
+    categories!.filter(c => c && c.name === "Black Owned").length > 0
 
   return (
     <PartnerArtistsLoadingContextProvider>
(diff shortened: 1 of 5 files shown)


The-Beez-Kneez:
@dzucconi do we have a less repetitive way to incorporate a new component such as this `Ribbon`? We do not really have a set precedent or previous example on how we would want to incorporate something like this.

dzucconi:
So, yeah:

* First we'd want to get a spec in the design system Figma first. Presumably there'd be variants on this; different configurable colors, sizes, and sanctioned labels.
* Once we have a spec we could implement this in Palette.

src/v2/Components/Ribbon.tsx
```diff
@@ -0,0 +1,85 @@
...
+          <Text variant="md" color="white">
+            {ribbonDivider}
+          </Text>
+          <Text variant="xs" color="white">
+            {ribbonText}
+          </Text>
+        </Flex>
+      </Media>
+    </>
+  )
```
dzucconi, line 84: Yeah, we can get rid of both the replication and the media component usage. Along with reducing the number of DOM nodes. If you want we could pair on an implementation in Palette?

src/v2/Apps/Partner/PartnerApp.tsx
```diff
@@ -24,12 +25,16 @@ export const PartnerApp: React.FC<PartnerAppProps> = ({
...
     partnerPageEligible,
+    categories,
   } = partner
 
   if (!isDefaultProfilePublic || !partnerPageEligible) {
     throw new HttpError(404)
   }
 
+  const isBlackOwned =
+    categories!.filter(c => c && c.name === "Black Owned").length > 0
```
dzucconi, line 36: I think it'd be useful to expose some field on Metaphysics for this. Like presumably there could be other ribbon labels for different kinds of galleries? Could there be a gallery that has more than one?

The-Beez-Kneez:
@lilyfromseattle sounds like we get to build a brand new Palette component! Woooo!

You should sync up with @dzucconi

src/v2/Apps/Partner/PartnerApp.tsx
jonallured: In the spirit of impact over perfection we are embracing the single category haha. Its certainly possible in the future that we'll want to apply multiple categories and have them show up but we have no comps or concrete plans for this!

PR #9309, opened 2022-02-01
fix: Reaction Import in V2

icirellik:
Clean up an import that is erroneously causing `@arsty/reaction` to be
imported into v2.


erikdstock (approved):
Well i wanted it to be a surprise... but ... i'm bringing reaction back! I can only assume this was an unfortunate error, perhaps an auto-import that came up in rebasing.

erikdstock:
!squashongreen

PR #9313, opened 2022-02-01
feat[WIP-dont merge]: GRO-734 migrate about2 to palette v3

TMcMeans:
### The type of this PR is: feat

This PR solves [GRO-734]

### Description

As part of the v3 static page migration, we are migrating a handful of static pages over to the new palette v3 framework. Routing is already set up, but content and styling based on figma designs handed over from design team needs to be implemented. 

### Screenshots
-will add responsive desktop and mobile images once page content is fully updated

### DIFF
--- a/src/v2/Apps/_About2/AboutApp.tsx
+++ b/src/v2/Apps/_About2/AboutApp.tsx
@@ -1,14 +1,6 @@
-import {
-  Text,
-  Box,
-  Image,
-  GridColumns,
-  Column,
-  Button,
-  BorderBox,
-} from "@artsy/palette"
+import { Box, Column, GridColumns, Image, Text } from "@artsy/palette"
 import { MetaTags } from "v2/Components/MetaTags"
-import { RouterLink } from "v2/System/Router/RouterLink"
+import { Rail } from "v2/Components/Rail"
 
 export const AboutApp: React.FC = () => {
   return (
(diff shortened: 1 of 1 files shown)


src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +25,232 @@ export const AboutApp: React.FC = () => {
...
-                  height="480px"
-                  src="https://artsy-media-uploads.s3.amazonaws.com/8uU7UcN_OJ8Go3LUjwQ3FQ/about_cheimAndRead.jpg"
-                />
-                <Text>Cheim &amp; Read - NEW YORK</Text>
-              </Box>
+      <Rail
+        title="Discover Artworks Just for You"
+        subTitle="On Artsy"
+        getItems={() => {
+          const items: RailComponentProps[] = [
```
dzucconi, line 51: One suggestion here: Rather than replicating the `ShelfArtwork` component by hand and including this data. You could just grab all the IDs for these artworks then write a query `{ artworks(ids: ["caleb-hahne-carry-me-home", ...etc]) { ...` which would let you use the real `ShelfArtwork` componentry. (Including things like the save button, correct metadata, etc.)

damassi: @dzucconi is right here and this would def simplify things but will also require a bit of relay work. Happy to pair on this (or better yet we can do a follow-up PR implementing this suggestion)

TMcMeans:
This PR is ready for review. A few things I feel like I could do better: 
- the ios/android app links are kinda simple links wrapped around transparent svgs that I pulled from vanity that link to app stores which I'm not sure is totally ok for responsiveness. I should have them open up to the actual app store but could use some support on implementing that 

- would like to do a query for the artwork rails but can't make heads or tails of the graphql/relay stuff. Could use pairing support, but can still open this PR to review regardless

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
-      </Box>
-    </>
+                <RouterLink to={`mailto:${specialist.email}`}>
+                  {specialist.email}
+                </RouterLink>
+              </Flex>
+            </Flex>
+            <Spacer my={2} />
+            <Text borderBottom="1px solid" borderBottomColor="black15" />
+            <Spacer my={2} />
```
dzucconi, line 394: This last spacer would count as an external margin — golden rule of margins = always use internal margins never outer margins.

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+          </Box>
+        ))}
+      </Column>
+    </GridColumns>
   )
 }
+
+const AdvisorySpecialists = [
```
dzucconi, line 402: For variables like this either use either normal camel casing or CONSTANT_CASE. Save PascalCase for components.

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+          works that you love—with your price range, taste, and current
+          collection in mind.
+        </Text>
+        <Text variant="sm" mt={4}>
+          More questions about collecting?
+        </Text>
+        <Flex>
+          <Button
+            variant="primaryBlack"
+            width="40%"
```
dzucconi, line 354: I find any kind of explicit width or height suspicious — we could likely achieve the desired layout using `GridColumns` so that we adhere to the underlying grid system.

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+          galleries, auction houses, fairs, and institutions. To connect
+          collectors to the art they care about, we feature the largest
+          selection of international emerging, established, and blue-chip
+          artists.
+        </Text>
+      </Column>
+      <Column span={12}>
+        <Spacer my={6} />
+        <Text borderBottom="1px solid" borderBottomColor="black15" />
+        <Spacer my={6} />
```
dzucconi, line 325: Outer margin ☝️

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+          The Art Market at Your Fingertips
+        </Text>
+        <Text variant="sm" mt={1} mb={4}>
+          Discover artworks just for you, get market insights, and buy and sell
+          with confidence—all on the Artsy app.
+        </Text>
+        <Flex>
+          <RouterLink to="https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080">
+            <Image
+              src="http://files.artsy.net/images/download-ios-app-transparent.svg"
```
dzucconi, line 246: Yeah it's totally fine to use SVGs like this. Whenever we inline them we're including them in the code bundle and sending them over the wire with the HTML page response. This is frequently desirable (like in the case of Palette icons) because it lets us control their fill color. In the case of a one-off like this, just using it like an image is great. One of the benefits is that the browser can cache them like normal images (in this case Vanity doesn't send the correct cache-control headers which... is a separate problem). And ideally we'd be able to also use `lazyLoad` here, which we can't because Palette's image component makes assumptions about the bg color (another separate problem).

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+        </Text>
+        <Text variant="xl" mt={1}>
+          The Art Market at Your Fingertips
+        </Text>
+        <Text variant="sm" mt={1} mb={4}>
+          Discover artworks just for you, get market insights, and buy and sell
+          with confidence—all on the Artsy app.
+        </Text>
+        <Flex>
+          <RouterLink to="https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080">
```
dzucconi, line 244: For an external link we don't really need to use `RouterLink` — though it's fine — and can use a normal `a` tag. Here we'd want to also add `target="_blank" rel="noopener noreferrer"`

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+            return <RailComponent {...item} />
+          })
+        }}
+      />
+
+      <SellWithArtsySection />
+      <Spacer mt={4} />
+      <AppDownloadSection />
+      <CollectorInfoSection />
+      <ArtsySpecialistsSection />
```
dzucconi, line 103: In lieu of the outer margins on all these components you'd control spacing here with spacers.

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -18,235 +27,406 @@ export const AboutApp: React.FC = () => {
...
+}
+
+const SellWithArtsySection: React.FC = () => {
+  const image = resized(
+    "http://files.artsy.net/images/molly_green_original.jpeg",
+    { width: 640 }
+  )
+
+  return (
+    <Container bg="black100" color="white100" mt={100}>
```
damassi, line 144: As a good habit, we should avoid putting padding or margin on the inner containers of components themselves and instead move that a bit higher up. Meaning:

Avoid: 
```tsx
const Foo = () => {
  return <Box mb={1}>Hi</Box>
}
const Bar = () => {
  return <Box mb={2}>Hello</Box>
}
const Baz = () => {
  return <Box>There</Box>
}
const App = () => {
  return (
    <>
      <Foo />
      <Bar />
      <Baz />
    </>
  )
}
```

Better: 
```tsx
const Foo = () => {
  return <Box>Hi</Box>
}
const Bar = () => {
  return <Box>Hello</Box>
}
const Baz = () => {
  return <Box>There</Box>
}
const App = () => {
  return (
    <>
      <Foo />
      <Spacer mb={2} />
      <Bar />
       <Spacer mb={2} />
      <Baz />
    </>
  )
}
```

The latter example is a little less clean code-wise, but its more maintainable in the long run.

src/v2/Apps/_About2/AboutApp.tsx
dzucconi: I haven't looked at the comps but it _looks_ to me like we'd want the buttons to just occupy the full space with a `space.2` gap in between.

So just removing the widths and using `flex={1}` on each button and sticking a `<Spacer ml={2} />` in between should accomplish this.

src/v2/Apps/_About2/AboutApp.tsx
damassi: Ah jinx 😄 - commented on that here: https://github.com/artsy/force/pull/9313#discussion_r803127183

damassi (approved):
A couple minor issues around container margins, but once those are taken care of looks good to merge from me 👍

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -269,7 +282,7 @@ const AppDownloadSection: React.FC = () => {
   )
 }
 
-const CollectorInfoSection: React.FC = () => {
+const CollectorInfoComponent: React.FC = () => {
```
damassi, line 285: Just in general, with react stuff, no need to encode the word 'component' in the variable name here and elsewhere, since that's implied by the type -- React.FC (FC = "Function Component")

TMcMeans: ah shoot I was trying to be uniform with naming, but thats an easy fix thanks for catching that

damassi:
@TMcMeans - one thing that is helpful during PR review is after changes are made and specific comments addressed, thumbs-up the comment (or comment in reply) and go through and resolve the comments so that they collapse. That gives the team a fresh look at the comments that are still relevant. 

We'll also want to update the PR title so that `WIP` is no longer in the title name.

src/v2/Apps/_About2/AboutApp.tsx
TMcMeans: this isnt resolved yet but Damon and I are gonna pair on it in a future PR

src/v2/Apps/_About2/AboutApp.tsx
TMcMeans: gonna refactor in a future PR and remove the legacy code in the /about page and do any code cleanup addressed here

PR #9315, opened 2022-02-02
feat(FX-3714): Parse old format of values for size filter

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3714]

### Description
Previously, when creating alerts in Eigen, old format of values for the size filter was used, where the user could specify only single size option. A little later, this logic was updated to match Force (user can specify multiple options). But since the user could have previously created a saved search alert with old format, it is necessary to correctly convert old format to the new one in order for the selected filters to be displayed correctly as pills on edit alert screen.

### Acceptance Criteria
* Old format of values for the size filter should be converted to the new format
  * `dimensionRange: "*-16.0"` must be converted to `sizes: ["SMALL"]`
  * `dimensionRange: "16.0-40.0"` must be converted to `sizes: ["MEDIUM"]`
  * `dimensionRange: "40.0-*"` must be converted to `sizes: ["LARGE"]`
* Correctly display old format of values as pills
  * `dimensionRange: "*-16.0"` must be converted to `Small (under 40cm)`
  * `dimensionRange: "16.0-40.0"` must be converted to `Medium (40 – 100cm)`
  * `dimensionRange: "40.0-*"` must be converted to `Large (over 100cm)`

### Demo
For example, we have created alerts with old format of values for the size filter
old format

and here is the result of their conversion to a new format

https://user-images.githubusercontent.com/3513494/152148981-0409023e-05f3-4a66-ac80-6a4391104f62.mp4

[FX-3714]: https://artsyproduct.atlassian.net/browse/FX-3714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/SavedSearchAlert/Utils/__tests__/extractPills.jest.ts
+++ b/src/v2/Components/SavedSearchAlert/Utils/__tests__/extractPills.jest.ts
@@ -1,0 +1,84 @@
+import { extractSizeLabelForOldFormat, extractSizeLabel } from "../extractPills"
+
+describe("extractSizeLabel", () => {
+  it("returns only small size filter value", () => {
+    expect(extractSizeLabel(["SMALL"])).toEqual([
+      {
+        filterName: "sizes",
+        name: "SMALL",
+        displayName: "Small (under 40cm)",
+      },
+    ])
+  })
+
+  it("returns only medium size filter value", () => {
+    expect(extractSizeLabel(["MEDIUM"])).toEqual([
+      {
+        filterName: "sizes",
+        name: "MEDIUM",
+        displayName: "Medium (40 – 100cm)",
+      },
+    ])
+  })
+
+  it("returns only large size filter value", () => {
+    expect(extractSizeLabel(["LARGE"])).toEqual([
+      {
+        filterName: "sizes",
+        name: "LARGE",
+        displayName: "Large (over 100cm)",
+      },
+    ])
+  })
+
+  it("returns small and large size filter values", () => {
+    expect(extractSizeLabel(["SMALL", "LARGE"])).toEqual([
+      {
+        filterName: "sizes",
+        name: "SMALL",
+        displayName: "Small (under 40cm)",
+      },
+      {
+        filterName: "sizes",
+        name: "LARGE",
+        displayName: "Large (over 100cm)",
+      },
+    ])
+  })
+})
+
+describe("extractSizeLabelForOldFormat", () => {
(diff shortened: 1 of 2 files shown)


gkartalis (approved):

mdole (commented):
ooo interesting! good thought @dimatretyak.

My two cents here: I would suggest that instead of adding logic to handle this outdated format to our frontend clients, we instead just migrate those filters to have the correct "new" values in our database.

So that would be something like (in pseudocode in Gravity):
```
For all search_criteria with a dimensionRange filter
If the value of dimensionRange is *-16.0
  Add sizes: ['SMALL']
  Delete dimensionRange
...
[and continue for medium and large]
```

I think that way we end up with less tech debt and less confusing search_criteria to maintain. What do you think @dimatretyak @gkartalis ?

dimatretyak:
> ooo interesting! good thought @dimatretyak.
> 
> My two cents here: I would suggest that instead of adding logic to handle this outdated format to our frontend clients, we instead just migrate those filters to have the correct "new" values in our database.
> 
> So that would be something like (in pseudocode in Gravity):
> 
> ```
> For all search_criteria with a dimensionRange filter
> If the value of dimensionRange is *-16.0
>   Add sizes: ['SMALL']
>   Delete dimensionRange
> ...
> [and continue for medium and large]
> ```
> 
> I think that way we end up with less tech debt and less confusing search_criteria to maintain. What do you think @dimatretyak @gkartalis ?

I agree with you. I asked this question [here](https://artsy.slack.com/archives/C9SATFLUU/p1643737542293489) yesterday

mdole:
ah dang - sorry I missed that @dimatretyak! I can ticket that migration now and follow up on it tomorrow :) Please always feel free to @ me if you haven't gotten a response on something - if I can't answer it I'm happy to find someone who can!

dimatretyak:
> ah dang - sorry I missed that @dimatretyak! I can ticket that migration now and follow up on it tomorrow :) Please always feel free to @ me if you haven't gotten a response on something - if I can't answer it I'm happy to find someone who can!

don't apologize and don't worry, everything is ok 😅. If it's not difficult for you, please create a ticket for this task. Thank you!

dimatretyak:
Сhanges will be on the Gravity side

mdole:
Ticketed here! https://artsyproduct.atlassian.net/browse/FX-3716

PR #9319, opened 2022-02-02
[PX-4857] Billing address isn't saved to the credit card for arta shipping

fladson:
This fixes the case where orders with Arta enabled and with "Billing and shipping address are the same" option enabled, weren't saving the address on CreditCard records.

_Before and after_
Screenshot 2022-02-02 at 14 54 37


src/v2/Apps/Order/Components/PaymentPicker.tsx
```diff
@@ -520,6 +520,15 @@ export const PaymentPickerFragmentContainer = createFragmentContainer(
...
           }
+          ... on CommerceShipArta {
+            name
+            addressLine1
+            addressLine2
+            city
+            region
+            country
+            postalCode
+          }
```
pepopowitz, line 531: I was initially unsure how adding fields to the query fragment would resolve this issue, but I think what is happening is the values that come back in the query [get added to the address that's sent to stripe](https://github.com/artsy/force/blob/8a1b817e2b85add7038e5776f0df57bbd4d53138/src/v2/Apps/Order/Components/PaymentPicker.tsx#L375-L380) -- do I understand that correctly?

pepopowitz (changes requested):
Nice, I think it makes sense! My only blocking request is to either match .tool-versions to .nvmrc or remove it.

src/v2/Apps/Order/Components/PaymentPicker.tsx
jpotts244: Fascinating find!

.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 12.20.2
```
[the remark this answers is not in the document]
fladson: 🤔  I based that on the `package.json`, will remove it.

src/v2/Apps/Order/Components/PaymentPicker.tsx
fladson: yeah, that's pretty much it.

pepopowitz (approved):

PR #9320, opened 2022-02-02
feat(TX-90): new checkout order breakdown for Avalara phase 2

rquartararo:
[TX-90](https://artsyproduct.atlassian.net/browse/TX-90)

This PR updates a lot of small details regarding the transaction breakdown steps. See the [Figma](https://www.figma.com/file/hDihSIzPU2DYDPye12iF0y/VAT%2FTaxes?node-id=1252%3A1472) to view all changes. 

This is behind the "Avalara Phase 2" lab flag. 

### UC1

**Buy Now**
Screenshot 2022-02-02 at 10 10 31 AM

**Make Offer**
Screenshot 2022-02-02 at 10 13 56 AM

### UC4 

Screenshot 2022-02-02 at 10 35 41 AM

### Order History

Screenshot 2022-02-02 at 10 07 37 AM


src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -41,6 +46,12 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
     textOverflow: "ellipsis",
   } as any
 
+  const artworkPrice = getOfferItemFromOrder(lineItems)
```
rquartararo, line 49: The "List Price" elsewhere is using the last offer price, so I did the same here, but I'm wondering if this should not be the case. Shouldn't the list price be the first given price regardless of the offer amount changes?

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -23,10 +24,11 @@ export interface TransactionDetailsSummaryItemProps
   useLastSubmittedOffer?: boolean
   offerContextPrice?: "LIST_PRICE" | "LAST_OFFER"
   showOfferNote?: boolean
-  placeholderOverride?: string | null
+  transactionStep?: string | null
```
rquartararo, line 27: I replaced 'placeholderOverride' with 'transactionStep' so that we can conditionally change the transaction details summary more easily.

rquartararo:
@damassi Do you know if we have tests for `ArtworkSummaryItem.tsx`, and if so, where they would be located?

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -67,6 +78,15 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
         <Text variant="sm" color="black60">
           {shippingOrigin}
         </Text>
+        {user && userHasLabFeature(user, "Avalara Phase 2") && (
+          <Text variant="sm">
+            {artworkPrice &&
+              `${priceLabel} ${appendCurrencySymbol(
+                artworkPrice.price,
+                currencyCode
```
starsirius, line 86: #non-blocking. Eventually we want to move the currency formatting logic to Metaphysics, instead of handling it on every client. @ashleyjelks has an [open PR](https://github.com/artsy/metaphysics/pull/3692) which can be a good pattern to follow.

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
starsirius: Makes sense to me!

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -116,54 +131,89 @@ export class TransactionDetailsSummaryItem extends React.Component<
...
+        ? "Waiting for final costs"
+        : "To be confirmed*"
+    }
+
+    if (transactionStep === "payment") {
+      return this.avalaraPhase2enabled ? "Waiting for final costs" : "—"
+    }
+
+    return this.avalaraPhase2enabled ? "Calculated in the next steps" : "—"
+  }
```
starsirius, line 148: #nit-picking. My experience to make removing the feature flag easier in the future is to wrap the condition as outside as possible so we can simply delete a few lines. For example (not tested),
```typescript
amountPlaceholder = () => {
  const { transactionStep } = this.props

  if (this.avalaraPhase2enabled) {
    return ["review", "payment"].includes(transactionStep) ? "Waiting for final costs" : "—"
  }

  return transactionStep === "review" ? "To be confirmed*" : "—"
}
```

starsirius (approved):
Had a few non-blocking comments and looks great! 💯  Given the complexity of the checkout flow, I suggest we do a bit more thorough QA and see if we cover all the steps.

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
rquartararo: This is so much cleaner thanks 🙏

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
rquartararo: Great, I will take a look!

rquartararo:
> Had a few non-blocking comments and looks great! 💯 Given the complexity of the checkout flow, I suggest we do a bit more thorough QA and see if we cover all the steps.

@starsirius Thank you for such a thorough review!

PR #9327, opened 2022-02-03
feat(auction2): First pass rebuild /auction/id 

damassi:
This is a **first-pass** at rebuilding/refactoring the /auction/id page. Most of this work involved taking apart the old page to the point where I understood it well enough to work on it, and then went through and tried to to simplify the control flow, file count and organization as much as humanly possible while still preserving some of the logic that existed previously. Most of the hard work here was untangling things since it crossed numerous sides of force, from old force apps, to new force apps, to old server-side logic to new. All this said, there's some copy 🍝 in here from old (meaning new v2/Auction) app that could likely be cleaned up further. 

A couple things to note:
- I've only worked against the happy path. Meaning:
- Staging auction was created in https://joe.artsy.net/job/gravity-staging-create-auction-on-demand/ 
- Bid on those artworks, they always succeeded
- Haven't tested any of the numerous failing paths that the logic accounts for 
- Still need to write tests 
- Still need to verify identity verification flows still work 
- Need to wait on MyBids component (ask out to design)
- Still need to add promoted sale rail 

- Example with header. NO Bidding (old auction engine / staging copy): http://localhost:5000/auction2/heritage-urban-art-companion
- Example that you **can** bid on; created from jenkins: http://localhost:5000/auction2/chris-moction

Now that this thing is together to the point where it's actually possible to develop for I think a good next task would be to write up control flow document for all possible auction states and verify each case executes properly. 

Also, sorry for lack of git history here. This thing was honestly too insane to think about apart from just smearing paint all around and being patient.


src/v2/Apps/Auction2/Components/Form/initialValues.ts
```diff
@@ -0,0 +1,34 @@
...
+  address,
+  creditCard: undefined,
+  agreeToTerms: false,
+}
+
+export const initialValuesForBidding: FormValuesForBidding = {
+  ...initialValuesForRegistration,
+  // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
+  selectedBid: undefined,
+}
```
damassi, line 34: All of this form stuff is copied over from old auction registration app. Just making it all work, and will then refactor

src/v2/Apps/Auction2/Routes/Bid/useSubmitBid.ts
```diff
@@ -0,0 +1,204 @@
...
+      artwork,
+    })
+
+    if (requiresPaymentInformation) {
+      try {
+        await createToken(values, helpers)
+      } catch (error) {
+        logger.error("Error creating token", error)
+      }
+    }
```
damassi, line 63: Since each of these can throw unique errors each needs to be try / caught and in the next PR this will be hooked up to an error handler for directing this stuff back to the UI

damassi:
@mzikherman / @bhoggard - Needless to say, this has not been QA'd yet -- _however_, I don't expect the structure and whatnot to change much and feel fairly confident that folks could work in parallel in this `v2/Apps/Auction2` folder on the timed bidding stuff (if you need to get started soonish). Otherwise, just going to moving getting this closer to completion and will let you know when its done. Things should come together pretty fast once this is merged.

src/v2/Apps/Auction2/Components/Form/OnSubmitValidationError.tsx
```diff
@@ -0,0 +1,49 @@
...
+  - IF the form has been submitted at least once
+  - AND the form is not submitting at this moment
+  - AND the form is invalid
+  - AND (in useEffect dependencies array) the form submitting state just
+    changed (because it is false, it must have been true above)
+  - THEN run the callback prop.
+  Background:
+    https://github.com/jaredpalmer/formik/issues/1484#issuecomment-490558973
+ */
+
```
dzucconi, line 20: Yeah, I find this odd — _why_ would you ever want to call `onSubmit` if there's a validation error.

src/v2/Apps/Auction2/Components/Form/Utils/formatError.ts
```diff
@@ -0,0 +1,17 @@
...
+        return [error]
+      }
+      case error?.message: {
+        return [error.message]
+      }
+    }
+  })()
+
+  return errorMessages
+}
```
dzucconi, line 17: An amusing amount of code is required to deal with our wildly inconsistent API errors.

src/v2/Apps/Auction2/Routes/AuctionFAQRoute.tsx
```diff
@@ -0,0 +1,224 @@
...
+import styled from "styled-components"
+import { createFragmentContainer, graphql } from "react-relay"
+import { AuctionFAQRoute_viewer } from "v2/__generated__/AuctionFAQRoute_viewer.graphql"
+import { MetaTags } from "v2/Components/MetaTags"
+
+interface AuctionFAQRouteProps {
+  viewer: AuctionFAQRoute_viewer
+}
+
+const AuctionFAQRoute: React.FC<AuctionFAQRouteProps> = ({ viewer }) => {
```
dzucconi, line 19: I could have sworn we had this already for the artwork page as a modal.

dzucconi: Ah, it's a real page and different enough. 👍

src/v2/Apps/Auction2/Routes/Auction2RegistrationRoute.tsx
```diff
@@ -0,0 +1,137 @@
...
+  sale: Auction2RegistrationRoute_sale
+}
+
+const Auction2RegistrationRoute: React.FC<Auction2RegistrationRouteProps> = ({
+  me,
+  sale,
+}) => {
+  const { tracking } = useAuctionTracking()
+  const { router } = useRouter()
+  const { needsIdentityVerification } = computeProps({ sale, me })
```
dzucconi, line 33: I don't see how extracting a function here improves things

src/v2/Apps/Auction2/Routes/Auction2RegistrationRoute.tsx
```diff
@@ -0,0 +1,137 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  return (
+    <ModalDialog
+      title="Register to Bid on Artsy"
+      onClose={handleModalClose}
+      width="50%"
+      // TODO: PR fix for this to palette ModalBase component
+      margin="auto"
```
dzucconi, line 57: What's the issue? We set the width of `ModalDialog` without there being a centering problem

src/v2/Apps/Auction2/Routes/Bid/Auction2BidRoute.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  const {
+    artworkSlug,
+    bidderID,
+    displayIncrements,
+    requiresCheckbox,
+    requiresPaymentInformation,
+    selectedBid,
+    validationSchema,
+  } = computeProps({ artwork, match, me })
```
dzucconi, line 48: `non-blocking`: This pattern just highlights inadequacies in the GraphQL schema. Either to do with nullability or to do with computed fields which should be located there. I realize you've got other fish to fry right now; but just a comment.

src/v2/Apps/Auction2/Components/Form/ErrorStatus.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <>
+      {title && <Text variant="md">{title}</Text>}
+
+      <Text variant="xs" color="red100">
+        {message}
+      </Text>
+    </>
+  )
```
dzucconi, line 81: I wonder if a toast is more appropriate?

jonallured: Damon loves toast 🍞

dzucconi: I do love consistency in the UI 😝 

IMO: inline error messages relate to specific fields. Toasts for everything else — with the exception being modal dialogs where we have to inline any feedback.

src/v2/Apps/Auction2/Routes/Bid/useSubmitBid.ts
```diff
@@ -0,0 +1,204 @@
...
+        helpers.setStatus(result.status)
+        helpers.setSubmitting(false)
+      }
+    }
+  }
+
+  return {
+    checkBidStatus,
+  }
+}
```
dzucconi, line 204: Fun

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
```diff
@@ -96,8 +99,15 @@ export class ArtworkSidebarBidAction extends React.Component<
   redirectToBid(firstIncrement: number) {
     const { slug, sale } = this.props.artwork
     const bid = this.state.selectedMaxBidCents || firstIncrement
+
     const href = `/auction/${sale?.slug}/bid/${slug}?bid=${bid}`
-    window.location.href = href
+
+    // FIXME: Remove this after Auction2 launches
+    if (getENV("ENABLE_AUCTION_V2")) {
```
dzucconi, line 106: Wondering if we could use the new feature flag service to roll this out? 🤔

dzucconi (dismissed):
Looks good to me! If a bit difficult to wrap my head around this much code without working through it.

src/v2/Apps/Auction2/Components/Form/ErrorStatus.tsx
damassi: I was actually wondering about the toasts here, lets discuss in Slack

src/v2/Apps/Auction2/Routes/Auction2RegistrationRoute.tsx
```diff
@@ -0,0 +1,137 @@
...
+    },
+  })
+
+  const handleModalClose = () => {
+    router.push(`/auction2/${sale.slug}`)
+  }
+
+  useEffect(() => {
+    tracking.registrationPageView()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
damassi: I was going to ask about this. If we want something to only fire once and only once, don't we have to use `[]`? How does this work with the dependency array? (I guess, if yr certain the dep is stable its ok, but still.)

src/v2/Apps/Auction2/Routes/AuctionFAQRoute.tsx
```diff
@@ -0,0 +1,224 @@
...
+  return {
+    biddingContent,
+    buyersPremiumTaxesAndFeesContent,
+    paymentsAndShippingContent,
+    emailsAndAlertsContent,
+    conditionsOfSaleContent,
+  }
+}
+
+const HTML: FC<{ children: string }> = ({ children, ...rest }) => {
```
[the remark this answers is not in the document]
damassi: +1

src/v2/Apps/Auction2/Components/Form/OnSubmitValidationError.tsx
damassi: I'm tempted to just delete this because i honestly don't understand why it was added initially and we don't see this kind of thing elsewhere. I guess i will and we can work from there.

src/v2/Apps/Auction2/Routes/Auction2RegistrationRoute.tsx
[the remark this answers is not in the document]
dzucconi: Oh, I see, it's the page view. Yeah, in this case that's fine.

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarBidAction.tsx
damassi: I think that service is still being worked on

src/v2/Apps/Auction2/Routes/Auction2RegistrationRoute.tsx
damassi: Thinking about your other comment -- in both here and in the file above, the bulk of this is for `needsIdentityVerification`. Def a case for moving to MP. Will do.

src/v2/Apps/Auction2/Routes/Bid/Components/AuctionLotInfo.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const artwork = saleArtwork.artwork
+  const bidCount = saleArtwork.counts?.bidderPositions!
+
+  if (!artwork) {
+    return null
+  }
+
+  return (
+    <Flex alignItems="center">
+      <RouterLink to={`/artwork/${artwork?.slug}`}>
```
damassi, line 22: fixme, check for image existy

PR #9337, opened 2022-02-04
feat(PX-4842): Use process with Arta shipping field

leamotta:
Completes https://artsyproduct.atlassian.net/browse/PX-4842

Using brand new field to decide whether an Artwork can be shipped with Arta or not


jpotts244 (approved):
+1 thanks for this! LGTM + lets test on staging

PR #9341, opened 2022-02-04
feat: Remove admin flag for NFT medium filter and move above fold

annacarey:
Removing the admin flag for the NFT medium type and moving it up above the fold in the medium filter on artsy.net/collect

Thanks to @SamRozen's [ping in Slack](https://artsy.slack.com/archives/C02STV6RDT4/p1643834792432649), now noticing that we have ~15 works populated [here](https://www.artsy.net/collect/nft) that partners added themselves in the last couple weeks. This is even before we've done the migration that we're planning for next week: manually switching the medium type on ~230 works to "NFT" so they populate here too. I think that's enough to roll out to production.

How it looks above the fold:
Screen Shot 2022-02-04 at 4 29 40 PM


anandaroop (approved):
👍🏽

PR #9351, opened 2022-02-07
feat: Remove exuberant greeting GRO-800

jonallured:
FWIW I like exuberance!

https://artsyproduct.atlassian.net/browse/GRO-800

/cc @artsy/grow-devs


jacobherrington (approved):
No exuberance allowed!

PR #9352, opened 2022-02-07
feat: Fallback to GDPR-compliant version GRO-802

jonallured:
Rather than falling back to the version of our signup form that does not have the checkboxes split out, this PR takes the opposite approach - when in doubt comply with this rule. This should not happen often but we had some trouble with ip lookups recently and that meant for visitors in GDPR countries we were not in compliance - yikes!

In terms of the code changes all I'm doing here is taking advantage of the fact that empty string is falsy in Javascript. 🤷 

https://artsyproduct.atlassian.net/browse/GRO-802

/cc @artsy/grow-devs


The-Beez-Kneez:
LGTM, glad we went with the less legal ramifications option lol

PR #9355, opened 2022-02-07
feat: Use article byline on homepage GRO-721

jonallured:
All I did here was switch from the `article.name` to the `byline` field that was already added by @dzucconi - here's some visual proof:

### before

Screen Shot 2022-02-07 at 3 41 47 PM

### after

Screen Shot 2022-02-07 at 3 41 37 PM

I was expecting to have to deal with some test fallout but...I think it was all correctly regenerated by relay? Pretty magical! ✨ 

https://artsyproduct.atlassian.net/browse/GRO-721

/cc @artsy/grow-devs


damassi:
> I was expecting to have to deal with some test fallout but...I think it was all correctly regenerated by relay? Pretty magical!

No doubt, the engineers at facebook have the best ideas

damassi (approved):

PR #9364, opened 2022-02-08
feat(offer): use submitOfferOrderWithConversation mutation

araujobarret:
[NX-3044]
@artsy/negotiate-devs 

Uses submitOfferOrderWithConversation mutation in order to create conversation and like to the offer and make an offer.
Also creates a note to be used as the first message of the conversation when no note is provided.

[NX-3044]: https://artsyproduct.atlassian.net/browse/NX-3044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
+++ b/src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
@@ -27,6 +27,10 @@ jest.mock("v2/Utils/Events", () => ({
 jest.mock("v2/Utils/Hooks/useMatchMedia", () => ({
   __internal__useMatchMedia: () => ({}),
 }))
+jest.mock("v2/Utils/user", () => ({
+  getUser: jest.fn(),
+  userHasLabFeature: jest.fn().mockReturnValue(false),
+}))
 
 const mockPostEvent = require("v2/Utils/Events").postEvent as jest.Mock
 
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
```diff
@@ -202,6 +207,10 @@ describe("Offer InitialMutation", () => {
       expect(page.offerInput.text()).toMatch("Offer amount missing or invalid.")
     })
 
+    it("adds a custom note given no note present", async () => {
```
araujobarret, line 210: when retiring the feature flag we must add the test case here, right now is too complicated with this test file to mock some stuff only in a single test case

src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
```diff
@@ -24,6 +24,11 @@ jest.mock("v2/Utils/Events", () => ({
   postEvent: jest.fn(),
 }))
 
+jest.mock("v2/Utils/user", () => ({
```
lean8086, line 27: should we remove this user mock since we will write the test case after retiring the feature flag?

lean8086 (approved):
LGTM and works like a charm! I left a question to know if an unused mock should be removed.

src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
araujobarret: Thanks for the review!
without it some tests will break 🙃

jo-rs (approved):
Looking good and working! 💪

src/v2/Apps/Order/Routes/__tests__/Offer.jest.tsx
lidimayra: Impact Over Perfection 🚀

PR #9367, opened 2022-02-08
feat(article): implements social embeds

dzucconi:
Closes [GRO-751](https://artsyproduct.atlassian.net/browse/GRO-751)

Will rebase once schema changes are in main before merging.


src/v2/Apps/Article/Components/Sections/ArticleSectionSocialEmbed.tsx
```diff
@@ -0,0 +1,51 @@
...
+  section,
+}) => {
+  const ref = useRef<HTMLDivElement | null>(null)
+
+  useEffect(() => {
+    if (!ref.current || !section.embed) return
+
+    // React doesn't load the script tags when using `dangerouslySetInnerHTML`
+    // So just inject the HTML manually after mount.
+    ref.current.innerHTML = section.embed
```
dzucconi, line 21: Previously we were using this in Reaction: https://github.com/humanmade/react-oembed-container —

They specifically say in their readme that this shouldn't work but it ... appears to work fine. Will just revisit if needed.

PR #9372, opened 2022-02-09
refactor(hook): Update useMutation hook

damassi:
This updates the implementation of `useMutation`'s `submitMutation` fn to be a little clearer and somewhat more verbose. New API: 

```tsx
const { submitMutation } = useMutation(...)

await submitMutation({ 
  variables: { foo },
  rejectIf: res => res.field.hasError
})
```


dzucconi (dismissed):
🙏

PR #9373, opened 2022-02-09
docs: Add more recent example for --watch [ci skip]

jacobherrington:
I think it might be a little more helpful to show the `--watch` flag with Jest since most folks probably won't be working on the legacy CoffeScript and Mocha code.


damassi:
whew! yes this is very very old. Thank you 👍

PR #9385, opened 2022-02-10
fix: platform-3750: Remove embedly as it's not in use anymore

rumyantseva:
[Embedly](https://embed.ly/) is no longer being used and so the integration should be removed from force.

## How to release?

As embedly is not in use anymore, there are no configs that need to be changed to make the changes available.
However, embedly-related configuration is provided in config maps - `EMBEDLY_ENDPOINT` and `EMBEDLY_KEY`. These variables need to be unset via

```
hokusai [staging|production] env unset EMBEDLY_ENDPOINT
hokusai [staging|production] env unset EMBEDLY_KEY
```


sonecabr (approved):
LGTM

mc-jones:
@dzucconi - Just flagging in case you didn't see it on Jira. We thought this would be a good first ticket for Elena's rotation, so have closed the ticket I previously assigned to you.

@rumyantseva - Thanks for picking this up! Let me know if you want to pair on taking a look at those failing tests!

dzucconi (dismissed):
🙏

sonecabr (approved):
Looks good to me.

rumyantseva:
@joeyAghion good point, thank you! Yes, env vars unset is needed, I'll update the description.

PR #9388, opened 2022-02-10
feat(GRO-817): Allow user to link a Google account

jacobherrington:
I noticed that it's not currently possible to link a Google account to an already existing account.

I'm having a bit of trouble testing this locally (cors issues with ngrok), but I think it should work. @dzucconi would love your feedback since you're blamed for this feature in the first place!


jacobherrington:
Tested in staging! Works as intended. :)

PR #9392, opened 2022-02-10
fix(nos): refactor nos to get listprice on edition sets

tam-kis:
[NX-3064](https://artsyproduct.atlassian.net/browse/NX-3064)

Fix New Offer Submission to work on Edition sets that have multiple size options.
Now we are getting the `listPrice` and checking the price range with `displayPriceRange` both from `CommerceOrder` . 

Edition Set:
Screenshot 2022-02-10 at 18 53 02


src/v2/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -258,7 +258,11 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
 
     const artwork = this.props.order.lineItems?.edges?.[0]?.node?.artwork
     const isInquiryCheckout = !artwork?.isPriceRange && !artwork?.price
-    const isEdition = artwork?.isEdition
+
+    // TODO: fix typescript error for displayPriceRange
+    const displayRange =
+      this.props.order.lineItems?.edges?.[0]?.node?.artworkOrEditionSet
+        ?.displayPriceRange || false
```
tam-kis, line 265: Typescript error for `displayPriceRange`

araujobarret: you can just `const displayRange = !!offerItem?.displayPriceRange` and then it's done

araujobarret: or if you don't need this variable somewhere else I would get rid of it and use only `!!offerItem?.displayPriceRange` where you need

tam-kis: Thanks! I thought there must be a shorter way to do it.

src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -70,20 +68,21 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
...
+  const listPrice = offerItem?.listPrice
 
   const getRangeOptions = () => {
-    const minPriceRange = listPrice?.minPrice?.major
-    const maxPriceRange = listPrice?.maxPrice?.major
-    const midPriceRange = Math.round(
-      (Number(minPriceRange) + Number(maxPriceRange)) / 2
-    )
-
+    const listPriceRange = listPrice as PriceRange
```
araujobarret, line 82: optional: taking a look again makes sense to me to move those functions like `getRangeOptions` and `getPercentageOptions` out of the component, passing as argument Price or PriceRange variable. You wouldn't need any typecast also when calling them, for example:

```jsx
const getRangeOptions = (price: PriceRange) => {
...
}
const getPercentageOptions = (price: Price) => {
...
}

export const PriceOptions: React.FC<PriceOptionsProps> = ({
onChange,
  onFocus,
  showError,
  order,
}) => {
  ...
  const priceOptions = isPriceRange ? getRangeOptions(listPrice as PriceRange) : getPercentageOptions(listPrice as Price)
}
```
we could go further around `displayPriceRange` property, right now we check if it's true/false, we could use typeguards on `listPrice` to check if it's Price/List price, this would make us more type assertive avoiding typecast to it; not saying we should do it rn, but we should discuss more about the idea 👍

araujobarret (approved):
Looks good 👍 
left a small comment, optional to do it

damassi:
@tam-kis - can this be closed?

jo-rs:
> @tam-kis - can this be closed?

@damassi I'm using it as a base for a new PR. I'll turn it into a draft for now, and will close after that one is done.

PR #9393, opened 2022-02-10
feat: implement unleash into Force

mc-jones:
The PR implements unleash into Force, allowing developers to use feature flags. Co-authored by @icirellik, @ovasdi, @sonecabr and @mc-jones

### Usage
1. Import the `useFeatureFlag` into the module
2. Set the component to conditionally render with the  `useFeatureFlag(<name_of_test>)`
3. Set a strategy in the unleash console and enable the desired environment.

### Next Steps
- [x] Add new env variables to shared.env
- [x] Set new env variables in staging and prod
- [x] Add Force as a project in Unleash console

### Follow-ups 
- Confirm that we will have `session_id` at this point server side. 
- Evaluate `unleash-express` to possibly replace current middleware.

### Demo
https://user-images.githubusercontent.com/38149304/153483727-35e0ee4f-96d3-4643-9213-707204fb24a3.mov

[PLATFORM-3921]

[PLATFORM-3921]: https://artsyproduct.atlassian.net/browse/PLATFORM-3921?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/featureFlags/featureFlagProvider.ts
```diff
@@ -0,0 +1,12 @@
+export type FeatureFlag = string
+export type FeatureFlags = Array<string>
+
+export interface FeatureFlagConfig extends Iterable<string> {}
+
+export type FeatureFlagContext = Record<string, any>
+
+export interface FeatureFlagProvider {
```
damassi, line 8: As this file is just a collection of types and not the actual provider, that should be noted in the filename -- `featureFlagTypes.ts`.

src/lib/featureFlags/unleashProvider.ts
```diff
@@ -0,0 +1,83 @@
...
+    private _appName: string,
+    private _instanceId: string,
+    private _serverKey: string
+  ) {}
+
+  [Symbol.iterator] = function* (): Iterator<string> {
+    yield this.url
+    yield this.appName
+    yield this.instanceId
+    yield this.serverKey
```
damassi, line 24: What is the reasoning behind using such advanced JS features as generators here?

src/v2/System/Router/UnleashProvider.tsx
```diff
@@ -0,0 +1,33 @@
+import { FeatureFlags } from "lib/featureFlags/featureFlagProvider"
```
damassi, line 1: This entire file should live in `lib/unleash`, not in our router

src/v2/System/Router/UnleashProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+function FeatureFlagProvider({ featureFlags, children }) {
+  return (
+    <FeatureFlagContext.Provider value={featureFlags}>
+      {children}
+    </FeatureFlagContext.Provider>
+  )
+}
+
+function useFlag(flagName) {
```
damassi, line 20: `export function useFlag ...`

src/v2/System/Router/UnleashProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  if (!flagName) {
+    return null
+  }
+
+  return Object.keys(context).indexOf(flagName) !== -1 && context[flagName]
+}
+
+export { FeatureFlagProvider, useFlag }
```
damassi, line 33: This can be deleted per the comments above

src/v2/System/Router/UnleashProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const FeatureFlagContext = React.createContext<FeatureFlagProviderInterface>({
+  featureFlags: [],
+})
+
+interface FeatureFlagProviderInterface {
+  featureFlags: FeatureFlags
+}
+
+function FeatureFlagProvider({ featureFlags, children }) {
```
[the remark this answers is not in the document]
damassi: Additionally, in general, file names should match the "main" export 1:1. The point of the file is `FeatureFlagProvider`, but the file name is `UnleashProvider`, which leads to some indirection.

src/lib/featureFlags/featureFlags.ts
```diff
@@ -0,0 +1,30 @@
+import { FeatureFlagConfig, FeatureFlagProvider } from "./featureFlagProvider"
+
+const registeredProviders = new Map<symbol, any>()
+
+export async function createFlagProvider<T extends FeatureFlagConfig>(
```
damassi, line 5: `createFlagProvider` -> `createFeatureFlagProvider`

src/lib/featureFlags/featureFlags.ts
```diff
@@ -0,0 +1,30 @@
...
+  type: symbol,
+  config?: T
+): Promise<FeatureFlagProvider> {
+  if (!registeredProviders.has(type)) {
```
damassi, line 9: ditto here. all of the variable names should be consistent if possible. There's now `FeatureFlagProvider`, flagProvider`, registeredProvider`, `registerFlagProvider`. We should stick with one idiom throughout

src/lib/featureFlags/featureFlags.ts
```diff
@@ -0,0 +1,30 @@
...
+    throw new Error(`The type provider ${type.toString()} is not registered.`)
+  }
+
+  const clz = registeredProviders.get(type)
```
damassi, line 13: what is `clz`? We should use full words for variable names to avoid confusion.

src/middleware.ts
```diff
@@ -68,6 +68,12 @@ import { serverTimingHeaders } from "./lib/middleware/serverTimingHeaders"
 import { splitTestMiddleware } from "./desktop/components/split_test/splitTestMiddleware"
 import { IGNORED_ERRORS } from "./lib/analytics/sentryFilters"
 import { sharifyToCookie } from "./lib/middleware/sharifyToCookie"
+import { featureFlagMiddleware } from "lib/middleware/featureFlagMiddleware"
```
damassi, line 71: since our feature flags is basically a lib, and we have `lib/featureFlags`, why not co-locate everything there so that related files aren't spread throughout the codebase?

src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -0,0 +1,47 @@
...
+        provider = await createFlagProvider(providerType, config)
+        resolve(provider)
+      } catch {
+        reject("An unknown error occurred while creating the flag provider.")
+      }
+    }).then(provider => {
+      // Create feature flag context per request
+      const featureFlagContext = {
+        userId: res.locals.user ? res.locals.user.id : null,
+        sessionsId: res.locals.sd.SESSION_ID,
```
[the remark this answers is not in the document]
joeyAghion: 0. Typo: ~`sessionsId`~ `sessionId`

1. My understanding is that the _client-side_ page will always have an "anonymous id" generated by Segment. That isn't available here. I _think_ Force always tracks a session ID so this might be a reasonable replacement for it (but someone else should confirm).

2. These basic attributes should support the built-in "stickiness" behaviors necessary for per-user toggles or gradual roll-outs or split tests. Additional attributes are possible, but I believe they must be [supported by clients](https://docs.getunleash.io/advanced/custom_activation_strategy) before they can be used, so no rush there.

damassi (commented):
This looks like a good first pass, but I have some concerns about the overall complexity in this PR for what we need to do. There's a whole lot going on that can be quite dramatically simplified and also a lot of stuff that should be cleaned up for consistency / comprehension sake before merge. Stuff like unnecessary generators, classes, variable name overloading, multiple files spread all around and more. 

After things are made consistent, what I would suggest is:
- All unleash / feature flag related files live in `/lib/featureFlags`
- Three files total: `setupFeatureFlags.ts`, `featureFlagMiddleware.ts`, `FeatureFlagProvider.ts` (react)

mc-jones:
Thanks for the quick review @damassi! 

One thing I should have included in the PR description, though, is our thinking around the approach you see in this PR. We had an initial approach that was slightly simpler, but ultimately we felt like this was a great case for dependency injection; the simpler implementation was proving difficult to test and we felt would have required significantly more work for a developer to swap in a different SDK in the future. While I definitely think the DI approach has added a bit of complexity, in comparison, we felt the trade-off was worth it to improve extensibility and testing.


I have to run a bit early today and still need to take a closer look at your comments in more detail but realized that the PR should have been more clear around the decision-making and trade-offs. Hopefully, this makes sense, and will get back to you on the specific points tomorrow!

damassi:
If you were running into issues with the simpler approach and testing, I'm happy to pair on it. I don't want to block work, but at a glance I think this does deserve a second pass, because all we're doing is instantiating an instance of unleash, injecting it as middleware, and then providing a react hook context in order to access it from our frontend code. We shouldn't ever let testing get in the way of a known simpler approach to an implementation because it's likely that the testing problems can be solved by reaching out. 

But aside from that, things like generators and the like are mostly a red flag in JavaScript, unless they're truly needed, and when that is so you'll know it. For something like this where the main aim is getting at the unleash instance, it's simply not needed and serves to complexify something that is fundamentally straightforward.

More than happy to pair on this tomorrow if you're interested; I think it would be fruitful. And if you'd like to merge and iterate that's also fine (after the cleanup and co-location concerns are addressed).

damassi (approved):
One last minor request around a react hook name -- `useFlag` -> `useFeatureFlag`, but otherwise the organizational and naming updates make this feel good to merge over here and then iterate 👍

src/lib/featureFlags/featureFlags.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  FeatureFlagConfig,
+  FeatureFlagProvider,
+} from "./featureFlagProviderShared"
+
+const registeredFeatureFlagProviders = new Map<symbol, any>()
+
+export async function createFeatureFlagProvider<T extends FeatureFlagConfig>(
```
damassi, line 8: Another naming thing is: in React, there's a standard context "Provider" naming pattern, and doubling up between there and here is slightly confusing. 

Is there another term that we can use use so that we're not sharing conflicting conceptual vocabulary? `featureFlagManager`, `featureFlagController`, or something like that?

damassi:
One other (future) thought is it'd be nice to be able to access these feature flags directly from the system context, and accessible via `useSystemContext`. That's where we get other system-level things, like our relayEnvironment, the user, our router, etc. Having our feature flags accessible there seems like a good central place to gain access, even if that data is simply proxied from the `FeatureFlagProvider`.

src/lib/featureFlags/featureFlagService.ts
```diff
@@ -1,29 +1,32 @@
-import {
-  FeatureFlagConfig,
-  FeatureFlagProvider,
-} from "./featureFlagProviderShared"
+export type FeatureFlag = string
+export type FeatureFlags = Array<string>
+
+export type FeatureFlagContext = Record<string, any>
+
+export interface FeatureFlagProvider {
```
damassi, line 6: Can we update this type name to `FeatureFlagService`

src/lib/featureFlags/featureFlagService.ts
```diff
@@ -1,29 +1,32 @@
...
+  init: () => void
+  getFeatures: () => FeatureFlags
+  enabled: (name: string, context?: FeatureFlagContext) => boolean
+}
 
 const registeredFeatureFlagProviders = new Map<symbol, any>()
```
damassi, line 12: `registeredFeatureFlagServices`

src/lib/featureFlags/featureFlagService.ts
```diff
@@ -1,29 +1,32 @@
...
   if (typeof featureFlagClass !== undefined) {
     /*
      ** Instantiate a new feature flag provider instance. Uses a custom
      ** configuration if provided, otherwise use default config from passed in
      ** featureFlagProviderClass.
      */
-    const featureFlagProvider = config
-      ? new featureFlagClass!(...config)
-      : new featureFlagClass!()
+    const featureFlagProvider = new featureFlagClass!()
```
damassi, line 29: `featureFlagService`

src/lib/featureFlags/unleashProvider.ts
```diff
@@ -5,47 +5,11 @@ import {
   UNLEASH_INSTANCE_ID,
   UNLEASH_SERVER_KEY,
 } from "../../config"
-import {
-  FeatureFlagConfig,
-  FeatureFlagProvider,
-} from "./featureFlagProviderShared"
+import { FeatureFlagProvider } from "./featureFlagService"
```
damassi, line 8: `FeatureFlagService`

src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -1,23 +1,19 @@
...
   FeatureFlagProvider,
-} from "lib/featureFlags/featureFlagProviderShared"
+} from "lib/featureFlags/featureFlagService"
 
-export function featureFlagMiddleware<T extends FeatureFlagConfig>(
-  providerType: symbol,
-  config?: T
-) {
+export function featureFlagMiddleware(providerType: symbol) {
   let provider
```
damassi, line 9: `service`

src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -1,23 +1,19 @@
...
-import { createFeatureFlagProvider } from "lib/featureFlags/featureFlags"
 import {
-  FeatureFlagConfig,
+  createFeatureFlagService,
   FeatureFlagProvider,
-} from "lib/featureFlags/featureFlagProviderShared"
+} from "lib/featureFlags/featureFlagService"
 
-export function featureFlagMiddleware<T extends FeatureFlagConfig>(
-  providerType: symbol,
```
damassi: `serviceType`

src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -1,23 +1,19 @@
...
+export function featureFlagMiddleware(providerType: symbol) {
   let provider
   return (req: ArtsyRequest, res: ArtsyResponse, next: NextFunction) => {
     new Promise<FeatureFlagProvider>(async (resolve, reject) => {
       if (provider) {
         resolve(provider)
       }
       try {
-        provider = await createFeatureFlagProvider(providerType, config)
+        provider = await createFeatureFlagService(providerType)
```
damassi, line 16: `service = await`

src/lib/featureFlags/unleashProvider.ts
mc-jones: Have removed the custom config and generator.

src/lib/featureFlags/featureFlags.ts
mc-jones: Good point. Changed to `featureFlagService`

joeyAghion (approved):

src/lib/featureFlags/unleashService.ts
```diff
@@ -5,13 +5,13 @@ import {
   UNLEASH_INSTANCE_ID,
   UNLEASH_SERVER_KEY,
 } from "../../config"
-import { FeatureFlagProvider } from "./featureFlagService"
+import { FeatureFlagService } from "./featureFlagService"
 
 // Pass in as argument to registerFeatureFlagProvideder() when using unleash as feature flag service
-export const UnleashProvider = Symbol("UnleashProvider")
+export const UnleashService = Symbol("UnleashService")
```
damassi, line 11: 👍

damassi:
Thanks for the naming updates 👍

PR #9399, opened 2022-02-11
Minor passport cleanup

jacobherrington:
Following behind #9263, I realized we weren't running the passport tests in CI.

This PR should fix that and I also DRYed up a few really repetitive functions.

I tested the refactor locally with Google and things seemed to still be functional. I'll give it a shot in staging before this is deployed, but pretty confident it's all good.


damassi (approved):
Not sure if there will be follow-up around the noted stubbed / removed tests, but the cleanup looks A+ 👍

jacobherrington:
@damassi A todo comment felt more honest/useful than disabling the linter warnings! 🤷

PR #9405, opened 2022-02-14
feat(TX-90): transaction details updates

rquartararo:
[TX-90](https://artsyproduct.atlassian.net/browse/TX-90)

This is a follow-up PR to https://github.com/artsy/force/pull/9320 to make some changes after a QA on staging.

Changes include: 

- Add a link to "Additional duties and taxes may apply at import".
- Remove the "Taxes may apply at checkout." message from sold artworks.
- Redid `shippingNotCaluclated` and `amountPlaceholder` logic to instead check if the transaction step is on "offer" or "shipping" as it makes more sense to check the only two steps where shipping has not been added. This prevents "calculated in next steps" showing on other screens like order history, which was previously not accounted for. 
- Updated transaction details messaging text size for mobile.

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -450,7 +450,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
               <Spacer mt={1} />
             )}
 
-          {avalaraPhase2Enabled && !isSold && (
+          {avalaraPhase2Enabled && artworkEcommerceAvailable && !isSold && (
             <Text variant="xs" color="black60">
               Taxes may apply at checkout.{" "}
               <a


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -449,7 +450,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
               <Spacer mt={1} />
             )}
 
-          {avalaraPhase2Enabled && (
+          {avalaraPhase2Enabled && !isSold && (
```
starsirius, line 453: Maybe we can reuse [`artworkEcommerceAvailable`](https://github.com/artsy/force/blob/55baef7ada76d6a3072a1bfd242d644869d5f649/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx#L400-L404) which takes into account [only for sale](https://github.com/artsy/gravity/blob/1cd2ea38e35729e38fe98b34bbc84363d047b872/app/models/util/for_sale.rb#L285-L300) artworks.

src/v2/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -135,9 +142,9 @@ export class TransactionDetailsSummaryItem extends React.Component<
     const { transactionStep } = this.props
 
     if (this.avalaraPhase2enabled) {
-      return ["review", "payment"].includes(transactionStep!)
-        ? "Waiting for final costs"
-        : "Calculated in the next steps"
+      return ["shipping", "offer"].includes(transactionStep!)
+        ? "Calculated in next steps"
+        : "Waiting for final costs"
```
starsirius, line 147: Makes sense to me! Thanks for catching this!

starsirius (dismissed):
Had a comment about reusing the `artworkEcommerceAvailable` logic if possible. Otherwise looks great to me! 🎉

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
rquartararo: Yes, this makes sense. Would be weird to have the tax message on artworks that aren't even for sale. Thanks!

PR #9407, opened 2022-02-14
feat(NX-3041): Implement new congrats for make offer on eligible artworks

lean8086:
[NX-3041](https://artsyproduct.atlassian.net/browse/NX-3041)

**On Force**
Created modal dialog on the artwork page to redirect to inbox after offer submission.

How it works:
1. When the user submits an order on the Review step, the order reference is sent to the artwork page (router state)
2. The order is passed as a property to the Artwork component, and removed from the state (modal should open once)
3. Modal is shown open and can be closed, or redirects the user to the inbox

image

**On Eigen**
Updated Status (congrats) page: webviews must be closed by the user, therefore we can't redirect them to the artwork page.

image

[Figma file](https://www.figma.com/file/PgmxHwBnoOUS4ybwpK2mTN/Make-Offer-on-all-Eligible-Artworks?node-id=525%3A3575) for reference.

cc @artsy/negotiate-devs


src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import {
+  getStatusCopy,
+  continueToInboxText,
+} from "v2/Apps/Order/Utils/getStatusCopy"
+import { RouterLink } from "v2/System/Router/RouterLink"
+import { SubmittedOrderModal_order } from "v2/__generated__/SubmittedOrderModal_order.graphql"
+
+interface SubmittedOrderModalProps {
+  order: Pick<SubmittedOrderModal_order, "mode" | "stateExpiresAt">
```
lean8086, line 12: I'm using `Pick` because I have this error with `$fragmentReft`, but it should not be necessary:
```
Property '' $fragmentRefs'' is missing in type 'SubmittedOrderModal_order' but required in type '_FragmentRefs<"SubmittedOrderModal_order">'.ts(2741)
index.d.ts(215, 5): '' $fragmentRefs'' is declared here.
SubmittedOrderModal.tsx(12, 3): The expected type comes from property 'order' which is declared here on type 'IntrinsicAttributes & MappedFragmentProps<Pick<SubmittedOrderModalProps, "order">> & { componentRef?: ((ref: any) => void) | undefined; } & { ...; }'
```

src/v2/Apps/Artwork/Components/SubmittedOrderModal/__tests__/SubmittedOrderModal.jest.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const mockedResolver = {
+  CommerceOrderConnectionWithTotalCount: () => ({
+    edges: [{ node: { mode: "OFFER", stateExpiresAt: "Feb 28" } }],
+  }),
+}
+
+describe("SubmittedOrderModal", () => {
+  const { renderWithRelay } = setupTestWrapperTL({
+    Component: ({ me }) => (
```
lean8086, line 16: I don't understand why it's expecting to find `children` here:
```
Property 'me' does not exist on type '{ children?: ReactNode; }'
```
Afaik it's [sending all the props as props](https://github.com/artsy/force/blob/main/src/v2/DevTools/setupTestWrapper.tsx#L96-L103). Am I looking on the right place, though?

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
araujobarret: You should stick to the relay generated type from the fragment query, the prop, in this case, should contain `" $refType": "SubmittedOrderModal_order"` and not only the 2 other props.
Probably is something with your VSCode, consider restarting it and should solve the issue. Another thing you can also run is `yarn relay:clean` to clean some weird Relay stuff from here.

src/v2/Apps/Artwork/Components/SubmittedOrderModal/__tests__/SubmittedOrderModal.jest.tsx
araujobarret: this should do the work
```jsx
const { renderWithRelay } = setupTestWrapperTL<
    SubmittedOrderModal_Test_Query
  >({
    Component: props => (
      <SubmittedOrderModalFragmentContainer
        order={props!.me!.conversation!.orderConnection!.edges![0]!.node!}
      />
    ),
    ...
```
or if you don't want to be so type-specific, you could type the props as any after Component, will be less verbose and less noisy from TS workarounds

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
araujobarret: I tried here I no error popped out.
There is no relation to another fragment type such as `Review_order`, in your case you're creating the fragment `SubmittedOrderModal_order` from graphql call, and relay compiler generates that type which should be used in the relay fragment prop of the component, no other way around.

src/v2/Apps/Order/Utils/getStatusCopy.tsx
```diff
@@ -276,6 +276,15 @@ export const covidNote = (): React.ReactNode => {
   )
 }
 
+export const continueToInboxText = (): React.ReactNode => {
```
araujobarret, line 279: I don't think putting the spacer here it's a good idea, despite my previous comment this function should return only the text and the component which uses it should place the Spacer there.

src/v2/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -238,6 +248,7 @@ const TrackingWrappedArtworkApp: React.FC<Props> = props => {
         routerPathname={pathname}
         referrer={referrer}
         shouldTrackPageView={isComplete}
+        submittedOrder={submittedOrder}
```
araujobarret, line 251: this must come from relay fragment query and not from router state

araujobarret (changes requested):
Thanks for the work but I believe we have some issues in the concept of dealing with relay props and storing them in the router to pass to other components.

src/v2/Apps/Order/Routes/Status/index.tsx
```diff
@@ -94,13 +102,15 @@ export class StatusRoute extends Component<StatusProps> {
                       {description && (
                         <Message p={[2, 4]}>{description}</Message>
                       )}
-                      {showTransactionSummary ? (
+                      {shouldContinueToInbox ? (
+                        <Text>{continueToInboxText()}</Text>
```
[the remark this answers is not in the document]
lean8086: very insightful! I didn't know how badly this could potentially mess things up 🔥

src/v2/Apps/Order/Utils/getStatusCopy.tsx
lean8086: Agreed.

src/v2/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -223,6 +230,9 @@ const TrackingWrappedArtworkApp: React.FC<Props> = props => {
   // Check to see if referrer comes from link interception.
   // @see interceptLinks.ts
   const referrer = state && state.previousHref
+  const submittedOrder = state && state.submittedOrder
+  // Discard the submitted order from the state after passing it the first time
+  if (submittedOrder) state.submittedOrder = null
```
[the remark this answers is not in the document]
araujobarret: >Would it be that the case if line 233 is a boolean coming from the review page? I guess the piece I'm missing is how can the artwork page know if the user comes from a specific use case (MOOAEA)
See [these lines on a previous PR](https://github.com/artsy/force/pull/9241/files#diff-1aaa2033a9ae43ffb3da1495105eac4324ed37ff31bbf19c13da79955ac4da47R151-R154) as an example.

Yes, I guess it makes sense, I would also just checking returning to the specific artwork after the redirect to the conversations to make sure that state is not persisted causing issues when collectors coming back to artworks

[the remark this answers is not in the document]
lean8086: > I would also just checking returning to the specific artwork after the redirect to the conversations to make sure that state is not persisted causing issues when collectors coming back to artworks

That's what emptying the state variable was achieving. I get it's an anti-pattern, though. Any ideas on how to achieve the same? Maybe passing a variable to the router state is not the right option? Maybe the same approach but breaking fewer rules would be passing a query string on the artwork page URL instead. wdyt?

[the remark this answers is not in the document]
lean8086: I guess another problem is how the artwork page can know for sure which order belongs to it. I can query `me.orders` and check the `artworks` there to match. Not sure what happens if, for example, a collector makes more than an offer on the same artwork, though (maybe on different editions).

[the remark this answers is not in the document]
araujobarret: I'm fine with the url param to be very honest, I prefer it since we don't need to persist anything

>I guess another problem is how the artwork page can know for sure which order belongs to it. I can query me.orders and check the artworks there to match. Not sure what happens if, for example, a collector makes more than an offer on the same artwork, though (maybe on different editions).

I guess you have something like myLastOffer to query that grabs only the last one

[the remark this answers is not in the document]
araujobarret: for uh-oh edition sets if it's easy we can handle it here, otherwise, we ticket a follow up on it

src/v2/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -49,6 +53,11 @@ export class ArtworkApp extends React.Component<Props> {
    * data that remains consistent with the rest of the app.
    */
   componentDidMount() {
+    if (this.shouldRenderSubmittedOrderModal()) {
+      // TODO: Look into using router push
+      // this.props.router.replace(this.props.match.location.pathname)
+      window.history.pushState({}, null, this.props.match.location.pathname)
```
lean8086, line 59: I wanted to use the router for removing the query param but found no way to keep the page from rerendering. Ended up copying the solution we're using all over Force.

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+    </ModalDialog>
+  )
+}
+
+export const SubmittedOrderModalFragmentContainer = createFragmentContainer<
+  SubmittedOrderModalProps
+>(SubmittedOrderModal, {
+  me: graphql`
+    fragment SubmittedOrderModal_me on Me {
+      orders(states: [SUBMITTED], mode: OFFER) {
```
lean8086, line 51: Tried to get `myLastOffer` but seems limited to a few properties that are not useful for this case. I'm getting user's orders instead...

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+        edges {
+          node {
+            stateExpiresAt(format: "MMM D")
+            lineItems {
```
lean8086, line 55: Only `stateExpiresAt` is needed on this modal dialog. Added this artwork slug check to be sure the latest order is also the one being displayed on the artwork page.

src/v2/Apps/Artwork/ArtworkApp.tsx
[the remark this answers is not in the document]
lean8086: Follow-up on this conversation: I pushed changes moments ago that I'll keep working on, but wanted to quickly check with you @araujobarret. It includes:

1. Review redirects to the artwork page with a query param
2. Artwork page picks up the query param, renders modal component, and removes query param:
3. Modal component picks up latest offer order
4. Modal component shows up if the latest order's artwork matches with the one on the artwork page.

More comments in the code itself.

src/v2/Apps/Order/Utils/getStatusCopy.tsx
```diff
@@ -276,6 +276,9 @@ export const covidNote = (): React.ReactNode => {
   )
 }
 
+export const continueToInboxText =
```
araujobarret, line 279: Q, why do we need it here and not directly in the component?

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
araujobarret: Humm I thought we had something like `myLastOffer` around, sorry to point you in the wrong direction.
It's logical but we can't assure that `const submittedOrder = me!.orders!.edges![0]!.node!` is gonna get exactly the order related to this artwork but I guess we can live with it
@jo-rs thoughts, objections?

araujobarret (approved):
Thanks for the changes here, I feel ok but one extra pair of eyes as a reviewer here would be nice too. Since, I'm the assignee here gonna wait someone else's review

src/v2/Apps/Artwork/Components/SubmittedOrderModal/SubmittedOrderModal.tsx
jo-rs: We do have that but that's related to a specific order object, so I don't think it'd apply here.
Taking a look at `exchange` and `metphysics` really quick. We had this issue at some point with bids where we were pulling the last bid in this way, but it arrived not sorted correctly. We might need to add sorting to the query just in case, and might as well limit it to 1, since as far as I understand we only need one, no?

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -148,10 +147,7 @@ export class ReviewRoute extends Component<ReviewProps> {
           order,
           o => o.lineItems?.edges?.[0]?.node?.artwork?.slug
         )
-        return router.push({
-          pathname: `/artwork/${artworkId}`,
-          state: { offerOrderHasBeenSubmitted: true },
-        })
+        return router.push(`/artwork/${artworkId}?orderHasBeenSubmitted=true`)
```
jo-rs, line 150: Are we using camel-cased params generally? 🤔

lean8086: True!

src/v2/Apps/Order/Utils/getStatusCopy.tsx
lean8086: We use it in two different places (congrats modal and Eigen's status page), always with other texts/components from this same file. I'd say the reasoning behind it was for centralizing it. Thoughts?

araujobarret: Nah, it's okay, just wondering if the tradeoff was paid, but I can live with it

PR #9412, opened 2022-02-14
feat(GRO-806): article classic layout

dzucconi:
Pretty simple; keeping the existing layout since that looks fine/consistent with v3 and just swapping out the related articles rail for a new channel articles rail.

https://user-images.githubusercontent.com/112297/153931825-a215b7f2-ce12-4117-a5d5-c1446443e92c.mov


damassi:
Our jobs page looking modern 🤯

damassi:
At some point we should loop around to this: https://www.artsy.net/jobs (though not sure how its built)

dzucconi:
@damassi 😈 JSON page...

damassi:
So this is the "cell"? Will keep this in mind -- we've got a few spots that can be updated. 

Screen Shot 2022-02-14 at 12 36 47 PM

PR #9415, opened 2022-02-14
feat[GRO-734]: Tmcmeans/about2 palette v3 refactor

TMcMeans:
This PR is for doing a little bit of code cleanup and adding tests for the new `/about2` page. 

I guess I should have split up the small naming convention changes/adding alt tag stuff and the smoke test but the test is so small that I didn't feel it needed a new PR. Hoping the smoke test will pass in CircleCI, though because it's not passing on my local machine.

### DIFF
--- a/cypress/integration/about2.spec.ts
+++ b/cypress/integration/about2.spec.ts
@@ -1,0 +1,6 @@
+describe("About", () => {
+  it("renders page content", () => {
+    cy.get("h1").should("contain", "For the Love of Art")
+    cy.title().should("eq", "About | Artsy")
+  })
+})
--- a/cypress/integration/about2.spec.ts
+++ b/cypress/integration/about2.spec.ts
@@ -1,5 +1,8 @@
+import { visitWithStatusRetries } from "../helpers/visitWithStatusRetries"
+
 describe("About", () => {
   it("renders page content", () => {
+    visitWithStatusRetries("about")
     cy.get("h1").should("contain", "For the Love of Art")
     cy.title().should("eq", "About | Artsy")
   })
--- a/src/v2/Apps/_About2/AboutApp.tsx
+++ b/src/v2/Apps/_About2/AboutApp.tsx
@@ -134,7 +134,7 @@ const RailComponent: React.FC<RailComponentProps> = props => {
         width={props.width}
         height={props.height}
         src={props.imgUrl}
-        alt=""
+        alt="carousel of artwork images with titles and gallery names"
       />
       <Text variant="md">{props.name}</Text>
       <Text variant="md" color="black60">
@@ -177,7 +177,7 @@ const SellWithArtsy: React.FC = () => {
             style={{
               objectFit: "cover",
             }}
-            alt=""
+            alt="a Molly Green art piece titled Cached"
           />
         </Box>
       )}
@@ -282,7 +282,7 @@ const AppDownload: React.FC = () => {
             height="100%"
             srcSet={image.srcSet}
             lazyLoad
-            alt=""
+            alt="painting of two iphones displaying the artsy mobile app"
           />
         </ResponsiveBox>
       </Column>
@@ -313,11 +313,11 @@ const CollectorInfo: React.FC = () => {
             height="100%"
             srcSet={image.srcSet}
             lazyLoad
-            alt=""
+            alt="picture of the Collection for Carole Server by Emily Johnston"
           />
         </ResponsiveBox>
         <Text variant="xs" textColor="black60" mt={0.5}>
-          The Collection fo Carole Server by Emily Johnston for Artsy 2015.
+          The Collection for Carole Server by Emily Johnston for Artsy 2015.
           Courtesy of Carole Server.
         </Text>
       </Column>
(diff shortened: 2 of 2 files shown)


damassi (dismissed):
@TMcMeans - If this is ready to go be sure to [use the `assignee` field](https://github.com/artsy/README/blob/main/playbooks/engineer-workflow.md#assignees-and-reviewers) to assign to someone to merge. But all in all looks great 👍

cypress/integration/about2.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { visitWithStatusRetries } from "../helpers/visitWithStatusRetries"
+
+describe("About", () => {
+  it("renders page content", () => {
+    visitWithStatusRetries("about")
```
damassi, line 5: /about2

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -282,15 +282,15 @@ const AppDownloadComponent: React.FC = () => {
             height="100%"
             srcSet={image.srcSet}
             lazyLoad
-            alt=""
+            alt="painting of two iphones displaying the artsy mobile app"
```
dzucconi, line 285: This isn't a painting of iPhones but rather it is two iPhones sitting on top of a painting.

However I'd also consider this a 'decorative image': https://webaim.org/techniques/alttext/#decorative in which case the blank alt is desireable.

It also looks like we'll have to replace this image: it's low quality and the credit is cut off (the credit shouldn't even be in the image IMO, but rather located below in actual text).

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -134,7 +134,7 @@ const RailComponent: React.FC<RailComponentProps> = props => {
         width={props.width}
         height={props.height}
         src={props.imgUrl}
-        alt=""
+        alt="carousel of artwork images with titles and gallery names"
```
dzucconi, line 137: We'll be replacing these with the ShelfArtworks so you can leave this alone.

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -177,7 +177,7 @@ const SellWithArtsyComponent: React.FC = () => {
             style={{
               objectFit: "cover",
             }}
-            alt=""
+            alt="a Molly Green art piece titled Cached"
```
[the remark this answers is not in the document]
TMcMeans: but wouldn't it be desirable to have an alt tag regardless?

src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -313,11 +313,11 @@ const CollectorInfoComponent: React.FC = () => {
             height="100%"
             srcSet={image.srcSet}
             lazyLoad
-            alt=""
+            alt="picture of the Collection for Carole Server by Emily Johnston"
```
dzucconi, line 316: This is redundant information, take a look at https://webaim.org/techniques/alttext/#context

src/v2/Apps/_About2/AboutApp.tsx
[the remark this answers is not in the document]
dzucconi: No, actually: https://webaim.org/techniques/alttext/#decorative (though you can ignore the comment about making things background images)

src/v2/Apps/_About2/AboutApp.tsx
TMcMeans: I pulled that image directly from the comp, and the credit line is part of that image?

dzucconi: Yeah, the comps frequently (mostly, really) don't have usable image assets — I'd ask design to provide an asset that isn't so low quality. Since the image displays at max 910x652, we'd need a minimum resolution of 1820x1304. And really the credit just looks more like a mistake to my eye than something intentional so I'd push back on that. We shouldn't be locking text inside of images anyway (if it _was_ we'd use that text verbatim in the `alt`).

src/v2/Apps/_About2/AboutApp.tsx
TMcMeans: gonna read through that article and handle these review notes for sure

src/v2/Apps/_About2/AboutApp.tsx
TMcMeans: would you say the same about the other image in the section below in the `collecting` section?? I believe it was pulled from the figma comp as well.

dzucconi: It looks _fine_. It looks like it was enlarged using nearest neighbor interpolation; but there's enough underlying resolution there that it doesn't matter much when we downsize it.

I think you need a plugin to actually download the original Figma image, which is crazy. It's really rare that we are actually pulling static image assets. When this does happen I try to hunt down the original from the designer or the source (for instance the artwork on Artsy).

damassi: Are these image concerns a blocker @dzucconi? Or can we merge and follow-up?

dzucconi: Oh sure; swapping images is totally a follow up thing.

TMcMeans:
this PR has been merged but the comments haven't been addressed yet. I'll push another PR linking this one and resolve the issues after.

PR #9416, opened 2022-02-14
feat(router): Add useCurrentRoute

damassi:
This is just an idiomatic React proxy to `getCurrentRoute` (which we use in some class components).


dzucconi (approved):

damassi:
Messed something up, reverting, and in any case i found a more duh way to do this via `useRouter`

PR #9419, opened 2022-02-15
[NX-3084] Hide ShippingInfo for inquireable artworks

lidimayra:
[NX-3084]

This is a quick fix just because MOOAEA project was adding the shipping info to the artwork page in artworks that didn't contain this information until now. This is leading to an unexpected message being displayed (like `Shipping null`). This was happening because this message is displayed in offerable artworks.

Screenshot 2022-02-15 at 12 07 09

Changes here just aim to prevent this from being displayed in a scenario that was affected by MOOAEA project.

The actual fix might consist of understanding why the message is being returned like this from metaphysics.

```graphql
{
  artwork(id: "mark-kostabi-the-bad-poetry-of-the-grid") {
    shippingInfo
  }
}

```

```
{
  "data": {
    "artwork": {
      "shippingInfo": "Shipping: null domestic, $1,200 rest of world"
    }
  },
  ...
}
```

When querying gravity we see this artwork has an international shipping fee set, but it has no domestic shipping fee:
```
gravity:staging> artwork.domestic_shipping_fee_cents
=> nil
gravity:staging> artwork.international_shipping_fee_cents
=> 120000
```

Once we fix this, we can even discuss if we want to undo the changes from this PR.

For more details see this slack thread:
https://artsy.slack.com/archives/C9YNS4X32/p1644914386896509

[NX-3084]: https://artsyproduct.atlassian.net/browse/NX-3084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -445,9 +445,8 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
           )}
 
           {artworkEcommerceAvailable &&
-            (artwork.shippingOrigin || artwork.shippingInfo) && (
-              <Spacer mt={1} />
-            )}
+            (artwork.shippingOrigin || artwork.shippingInfo) &&
+            !isInquireable && <Spacer mt={1} />}
 
           {avalaraPhase2Enabled && (
             <Text variant="xs" color="black60">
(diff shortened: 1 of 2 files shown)


joeyAghion (commented):
This seems like a bug in Metaphysics' [upstream `shippingInfo` property](https://github.com/artsy/metaphysics/blob/df89ed4fbd8dcb7d1d14aeecc44121a0e160a4f1/src/schema/v2/artwork/index.ts#L758-L818) that could affect multiple clients. (It handles a missing international fee, but not a missing domestic fee.)

If the point of that property is to describe shipping options consistently across the platform, I'd recommend fixing it there rather than working around it in Force.

lidimayra:
Thanks @joeyAghion!!
I just opened https://github.com/artsy/metaphysics/pull/3771 to handle the missing scenario in Metaphysics (when domestic shipping is null and international shipping is present)

Still, we're not sure if we want to display this message in case of artworks marked as inquireable.
We might decide to display something like: `Calculated at checkout` or even not displaying the message at all. Joh will check this to find out if galleries might have any objections here.

Either way, this PR doesn't seem to make sense at this point, so I'm closing it.

PR #9420, opened 2022-02-15
feat(FX-3713): Ability for user to select size filters in Imperial/or Metric 

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3713]


src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -63,37 +86,51 @@ const convertRangeToInches = (range: CustomRange) => {
...
+export const getCustomSizeRangesInInches = (
+  customSize: CustomSize,
+  sourceMetric: Metric
+) => {
+  let sizes = customSize
+
+  if (sourceMetric === "cm") {
+    sizes = {
+      width: convertRangeToInches(customSize.width),
+      height: convertRangeToInches(customSize.height),
```
dblandin, line 114: **suggestion**: Noticed some very long converted values in the query params when selecting custom ranges. Should we use `Math.round` in `convertRangeToInches` as well?

src/v2/Components/ArtworkFilter/ArtworkFilters/__tests__/SizeFilter.jest.tsx
```diff
@@ -185,3 +227,116 @@ describe("SizeFilter", () => {
...
+  })
+
+  it("should return correct custom size ranges in centimeters when default values are passed", () => {
+    const range: CustomSize = {
+      width: ["*", 10],
+      height: [15, "*"],
+    }
+    const result = {
+      width: "*-3.937007874015748",
+      height: "5.905511811023622-*",
```
dblandin, line 337: **suggestion**: Related to the above comment - I think we can trim off some precision here.

dblandin (commented):
Lookin good!

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
dimatretyak: Yes, we can do this, but there may be problems with the reverse conversion. For example:
* user specified `100 cm`
* convert from `cm` to `in` (`100cm` -> `39.37007874015748 in`)
* Apply `Math.round` (get `39 in`)
* convert from `in` to `cm` (`39 in` -> `99 cm`)

dblandin: How about rounding to two decimal places?

```typescript
const cm = 100
const inches = Math.round((cm / 2.54) * 100) / 100 // => 39.37
Math.round((inches * 2.54) * 100) / 100 // => 100
```

dimatretyak: I also considered this option, but there is still a problem with rounding

```javascript
const cm = 1
const inches = Math.round((cm / 2.54) * 100) / 100 // => 0.39
Math.round((inches * 2.54) * 100) / 100 // => 0.99
```

dimatretyak: But this option will work (when converting from `in` to `cm` and rounding occurs to an integer)

```javascript
const cm = 1
const inches = Math.round((cm / 2.54) * 100) / 100 // => 0.39
Math.round(inches * 2.54) // 1
```

dblandin: Looks good to me 👍🏼

dimatretyak: I remembered [some details](https://github.com/artsy/force/pull/9096) about why the custom values were not rounded on Force. If we return the rounding to Force, then we will need to add changes to Eigen as well.

dblandin: Ahhh, I see. Uff...

In that case let's skip the rounding and keep the long conversions. They'll only be visible in the URL query params anyway. Shouldn't be too noticeable.

But I think we should also prioritize pushing some of these concerns back from the UI, whether that's in [Metaphysics](https://github.com/artsy/metaphysics) or [Gravity](https://github.com/artsy/gravity) 🔒. From the UI's perspective, it should just request filtering in centimeters and let the backend handle _how_ that happens.

dimatretyak: > But I think we should also prioritize pushing some of these concerns back from the UI, whether that's in [Metaphysics](https://github.com/artsy/metaphysics) or [Gravity](https://github.com/artsy/gravity) 🔒. From the UI's perspective, it should just request filtering in centimeters and let the backend handle how that happens.

Good suggestion!

PR #9426, opened 2022-02-15
GRO-770: User on gallery listing page sees label on black owned gallery card

The-Beez-Kneez:
This is to support the newly created "Black Owned" gallery badges

New Functionality:
Screen Shot 2022-02-15 at 11 51 54 AM


dzucconi:
So to recap what I said over chat:

To figure out where that type error was coming from I: rebased with `git rebase main` to clean up the history, then just pulled off those 3 commits that were left. `git reset HEAD~1` will pull off a single commit. Then it was clear that there was a `ts-ignore` that was removed in the process:

.

So I checked out that file to restore it. And then you can recommit only the relevant changes in a single commit from there.

The-Beez-Kneez:
Ended up rebasing and opening a second PR @dzucconi

PR #9428, opened 2022-02-15
GRO-770: Reverted and Added: EntityHeader and Badging to PartnerCell

The-Beez-Kneez:
Ended up having revert issues, so I closed and reopened. This is to support the newly created "Black Owned" gallery badges.

New Functionality:
Screen Shot 2022-02-16 at 9 29 24 AM

### DIFF
--- a/src/v2/Components/Cells/PartnerCell.tsx
+++ b/src/v2/Components/Cells/PartnerCell.tsx
@@ -41,6 +41,9 @@ const PartnerCell: React.FC<PartnerCellProps> = ({
   const locations = extractNodes(partner.locationsConnection)
   const meta = uniq(locations.map(location => location.city?.trim())).join(", ")
   const image = partner.profile?.image?.cropped
+  const badgedPartners = partner.categories
+    ?.filter(category => category?.slug === "black-owned")
+    .map(category => ({ children: category?.name }))
 
   if (!partner.profile) {
     return null
@@ -70,6 +73,7 @@ const PartnerCell: React.FC<PartnerCellProps> = ({
         name={partner.name!}
         meta={meta}
         smallVariant
+        badges={badgedPartners}
         mb={1}
         FollowButton={
           <FollowProfileButtonFragmentContainer
@@ -159,6 +163,10 @@ export const PartnerCellFragmentContainer = createFragmentContainer(
               city
             }
           }
+        }
+        categories {
+          name
+          slug
         }
         profile {
           ...FollowProfileButton_profile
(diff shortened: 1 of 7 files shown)


src/v2/Components/Cells/PartnerCell.tsx
```diff
@@ -41,6 +41,9 @@ const PartnerCell: React.FC<PartnerCellProps> = ({
   const locations = extractNodes(partner.locationsConnection)
   const meta = uniq(locations.map(location => location.city?.trim())).join(", ")
   const image = partner.profile?.image?.cropped
+  const badgedPartners = partner.categories
+    ?.filter(category => category?.slug === "black-owned")
+    .map(category => ({ children: category?.name }))
```
dzucconi, line 46: A nice follow up would be to add a `displayable` (or something) argument to the Metaphysics field and do the filtering there.

dzucconi (dismissed):

src/v2/Components/Cells/PartnerCell.tsx
jonallured, line 46: **suggestion [non-blocking]:** use more intermediate variables

I think what this is doing is taking the categories array, finding whether it contains the black owned PartnerCategory and transforming it into an object that is suitable for the `EntityHeader` component, right?

I might suggest some intermediate vars - something more like this:

```ts
// coerce away the case of nil by falling back to an empty array
const categories = partner.categories || []
// supporting a new category badge would mean updating this list of slugs:
const badgedCategorySlugs = ["black-owned"]
const matchingCategories = categories.filter(category => badgedCategorySlugs.includes(category.slug))
const badges = matchingCategories.map(category => ({ children: category.name }))
```

And I guess typing this out has me wondering if there should be a function extracted here with tests.

dzucconi: If you were going to go the route of testing this as a more formal thing, I'd definitely just do this in Metaphysics; since this is just some temporary logic if it's in Force.

The-Beez-Kneez: Should we cut out a separate ticket for a potential MP breakout? Or should the above int variable breakdown go in on its own?

I personally think both should be done.

The-Beez-Kneez: Added in the suggested variables and also created the following JIRA ticket for a MP breakout of the filtering logic:
https://artsyproduct.atlassian.net/browse/GRO-829

The-Beez-Kneez: Created a JIRA ticket to discuss the following proposed MP transition.

jonallured (approved):
Let's do this!

PR #9432, opened 2022-02-16
feat: update NFTs nav link to /nft

dblandin:
This commit updates the NFTs nav link to a relative `/nft` href. This will match a currently [registered vanity shortcut](https://artsy.net/nft) to `/gene/nft`.

The gene is currently unpublished so the page a user is redirected to throws a Forbidden 403 error.

We can either:
- Publish the gene, which will make the gene page visible
- Update the vanity shortcut to another visible page. The collect page filtered
  to artworks of the NFT medium type, for example.

Either way, changing the nav link to not point directly to a specific subdomain gives us more flexibility and a UI to change where this link redirects a user to.

Link to a related [Slack thread](https://artsy.slack.com/archives/CHVFQ06DV/p1645020706482189) 🔒

### DIFF
--- a/src/v2/Components/NavBar/NavBar.tsx
+++ b/src/v2/Components/NavBar/NavBar.tsx
@@ -382,10 +382,7 @@ export const NavBar: React.FC = track(
                   Museums
                 </NavBarItemLink>
 
-                <NavBarItemLink
-                  href="https://nft.artsy.net"
-                  onClick={handleClick}
-                >
+                <NavBarItemLink href="/nft" onClick={handleClick}>
                   NFTs
                 </NavBarItemLink>
               </Flex>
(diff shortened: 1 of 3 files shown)


sweir27:
For the next NFT sale, I believe the team is going to build a proper landing page for all NFT sales! Once that's up, I think we can just leave it pointing to https://nft.artsy.net. To be more explicit, at that point we could change the name in the nav bar to `NFT Auctions`? (cc @GuillaumeDelg)

I don't mind having this be configurable, but I think it might be a little weird for our users if the link switches back and forth with some frequency.

src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -382,10 +382,7 @@ export const NavBar: React.FC = track(
                   Museums
                 </NavBarItemLink>
 
-                <NavBarItemLink
-                  href="https://nft.artsy.net"
-                  onClick={handleClick}
-                >
+                <NavBarItemLink href="/nft" onClick={handleClick}>
```
sweir27, line 385: Small thing-- I don't love using a vanity URL here, since it'll be confusing if we ever add an `/nft` path internal to Artsy.

Could we instead make this an ENV var?

PR #9433, opened 2022-02-16
chore(example-app): Simplify how we demo ArtworkFilter

damassi:
Addresses https://artsyproduct.atlassian.net/browse/WP-92 

Looking at our example app artwork filter, I realized we could simplify the bare bones use-case quite a bit.


damassi:
Going to merge this on 🍏 as there's something I need from it in my other PR! Can address any feedback / follow-up tomorrow.

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -51,40 +46,42 @@ import {
...
 import { extractPills } from "../SavedSearchAlert/Utils/extractPills"
 import { useFilterPillsContext } from "./SavedSearch/Utils/FilterPillsContext"
 import { getTotalSelectedFiltersCount } from "./Utils/getTotalSelectedFiltersCount"
 
+interface ArtworkFilterProps extends SharedArtworkFilterContextProps, BoxProps {
+  enableCreateAlert?: boolean
+  Filters?: JSX.Element
+  offset?: number
+  // Input variables passed to FilterArtworkConnection `input` argument
+  relayRefetchInputVariables?: object
```
damassi, line 61: Realized we had no way to control how refetches work after the initial filter render, when we call `prepareVariables` from the routers file, so added that here and cleaned up some types below.

src/v2/Apps/Example/exampleRoutes.tsx
```diff
@@ -96,30 +96,34 @@ export const exampleRoutes: AppRouteConfig[] = [
         },
         query: graphql`
           query exampleRoutes_ArtworkQuery($slug: String!) {
-            artwork(id: $slug) {
+            artwork(id: $slug) @principalField {
```
araujobarret, line 99: 🎖️

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -51,40 +46,42 @@ import {
...
+  // Root-level variables passed to Relay query
+  relayVariables?: object
```
araujobarret, line 63: can't we use the relay-generated types for variables here instead of just object?

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -269,21 +225,16 @@ export const BaseArtworkFilter: React.FC<
...
-    const relayRefetchVariables = {
-      first: 30,
-      ...allowedFilters(filterContext.filters),
-      keyword: filterContext.filters!.term,
-    }
-
     const refetchVariables = {
-      input: relayRefetchVariables,
+      input: {
+        first: 30,
```
araujobarret, line 233: I didn't get the change of 10 from the route to 30 here, was it because of the issue you mentioned here?

araujobarret (commented):
Good to see that everything worked and 0 touches in tests, super cool PR! 🚀

src/v2/Components/ArtworkFilter/index.tsx
damassi: We could, but that would then require us to pass in a union type here of all the various artwork filters that we're using throughout Force. Its not _so_ bad, but in practice it can get confusing since the filter is meant to be generic but is tied kinda lightly to the structure of the query. This is something that we can and should revisit when we refactor the Filter to be less dependent on GraphQL.

PR #9439, opened 2022-02-17
docs: Update troubleshooting doc [ci skip]

jacobherrington:
Figured I would do this while the thought was fresh.

Should we also rephrase the error message @damassi?


damassi:
I always forget about this troubleshooting doc! For an error like this though I think rephrasing the error message itself should be sufficient. However it looks like that work has now been reverted: https://github.com/artsy/force/pull/9437 so we should hold on this for the time being.

PR #9449, opened 2022-02-18
[FF] Setup React intl

lidimayra:
This is part of the result of an internationalization project that was worked on during the hackathon (hack9).

Changes in this PR consist of a very basic structure to enable localization on Force.
Once we move forward with the initial setup, we could split subsequent work - internationalizing other strings contained in the project - maybe we could mob on it on another Future Friday (?)

Slack channel: [#hack9-localizations](https://artsy.slack.com/archives/C02T87BHDAT)
[Notion page](https://www.notion.so/artsy/Internationalization-e4afd66b69804d9293c706df2111f75a)

---

_We still have some more changes that were excluded from this PR but can be found [here](https://github.com/artsy/force/commits/hack9-localizations/react-intl)_ like:
- Internationalization of some components (This PR is already big enough)
- Non-English language translations (we won't support any other language yet)
- Dropdown containing language selector on footer (for the same reason above)


src/v2/Apps/Components/AppShell.tsx
```diff
@@ -75,48 +77,58 @@ export const AppShell: React.FC<AppShellProps> = props => {
...
-      // Prevents horizontal scrollbars from `FullBleed` + persistent vertical scrollbars
-      overflowX="hidden"
-      // Implements "Sticky footer" pattern
-      // https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/
-      minHeight="100vh"
-      flexDirection="column"
+    <IntlProvider
+      messages={localizations["en_US"]}
+      locale="en-us"
+      defaultLocale="en-us"
```
dzucconi, line 88: We should put this provider in `Boot` with the other providers.

jest.config.legacy.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
         "\\.(gql|graphql)$": "jest-transform-graphql",
         "^.+\\.coffee$":
           "<rootDir>/node_modules/jest-coffee-preprocessor/index.js",
-        ".(ts|tsx|js|jsx)": "babel-jest",
+        "^.+\\.(ts|tsx|js|jsx)$": "babel-jest",
```
pvinis, line 21: was this causing issues before?

src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -149,7 +150,7 @@ export const NavBar: React.FC = track(
                   flex={1}
                   size="small"
                 >
-                  Sign Up
+                  <FormattedMessage id="navbar.signup" />
```
dzucconi, line 153: Is there type support (like do these ids autocomplete)?

src/v2/compiled-lang/en-us.json
```diff
@@ -0,0 +1,98 @@
+{
+  "navbar.artists": [
```
pvinis, line 2: whats the diff of the `lang` and `compiled-lang` files? do we need both?

src/v2/Apps/lang/en-us.json
```diff
@@ -0,0 +1,17 @@
+{
```
pvinis, line 1: should we merge this file with the other one outside `Apps/`?

src/v2/Components/NavBar/NavBar.tsx
pvinis, line 153: i still feel very strongly that this should not look like `<FormattedMessage id="navbar.signup" />`, and instead should look like `t("navbar.signup")` or something very very minimal.

pvinis:
added some comments, none of them blocking, mostly suggestions and questions.

package.json
```diff
@@ -20,6 +20,7 @@
     "clean": "scripts/clean.sh",
     "clean:relay": "rm -rf '$TMPDIR'/RelayFindGraphQLTags-*",
     "compile": "babel src/v2 --out-dir dist/v2 -s --source-map --extensions '.js,.jsx,.ts,.tsx' --ignore src/v2/**/__tests__",
+    "compile-formatjs": "formatjs compile",
```
damassi, line 23: Following other idioms in package.json `compile:formatjs`

damassi: or better: `formatjs:compile`

src/v2/Components/NavBar/NavBar.tsx
dzucconi: Yeah, this is kinda confusing. There's a hook that can be used which seems nicer.

pvinis: if this doesnt exist, having a hook or a global func could totally solve this with TS.
```
const t = useTranslations()
return (
  <Text>
    t("navbar.signup")
  </Text>
)
```

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -18,6 +18,8 @@ import { AppToasts } from "./AppToasts"
 import { useNavBarHeight } from "v2/Components/NavBar/useNavBarHeight"
 import { useProductionEnvironmentWarning } from "v2/Utils/Hooks/useProductionEnvironmentWarning"
 import { useAuthValidation } from "v2/Utils/Hooks/useAuthValidation"
+import { IntlProvider } from "react-intl"
+import locale_en_us from "../../compiled-lang/en-us.json"
```
damassi, line 22: Follow-up: I'm curious what including all of the different languages would amount to file-size-wise. In the future we should probably code-split via `import()` and `react-loadable` as we do with our other split components [as seen here](https://github.com/artsy/force/blob/main/src/v2/Apps/Artist/artistRoutes.tsx#L12-L17).

src/v2/Components/NavBar/NavBar.tsx
dzucconi: https://formatjs.io/docs/react-intl/api/

https://formatjs.io/docs/react-intl/components#why-components

It seems like this lib does a lot and tries to be a do-all solution for text formatting. Which 🤷 

I'm not sure why the hook wouldn't "Provide TypeScript type definitions" though?

pvinis: m indeed. we can at least do what we do with palette tokens, and wrap their func in our func, and add types based on our json translation files 🤔

src/v2/Apps/Components/AppShell.tsx
damassi: We can move this to boot, but better would be to write our own component that wraps this and encapsulates the other setup functionality in one location, eg 

 ```tsx
 // v2/System/i18n/InternationalizationProvider.tsx 
 
import { IntlProvider } from "react-intl"
import locale_en_us from "../../compiled-lang/en-us.json"
 
 const localizations = {
    en_US: locale_en_us,
  }
  
export const InternationalizationProvider: React.FC = ({ children }) => {
  return (
    <IntlProvider
      messages={localizations["en_US"]}
      locale="en-us"
      defaultLocale="en-us"
    >
      {children}
    </IntlProvider>
  )
}
 ```
 
 And then import this into `Boot` as suggested above.

damassi: Also, note the output of the json file above: `__i18n__` -- this matches idioms for our other compiled target output `v2/__generated__`, for relay.

src/v2/Apps/lang/en-us.json
damassi: I suggested above that we output to `v2/__i18n__`, like we do for relay generated fragments (`v2/__generated__`)

damassi: I think this should live in `v2/System/i18n/lang`

src/v2/Components/NavBar/NavBar.tsx
damassi: Agree with @pvinis! Being locked into an react component for this can be limiting here since we're often using logic in plain js -- ` foo ? "hello" : "world". That kind of thing. 

But that said, i could be convinced that a react component like this is indeed better. Worth a bit of discussion.

src/v2/System/SystemContext.tsx
```diff
@@ -111,10 +113,21 @@ export const SystemContextProvider: FC<SystemContextProps> = ({
...
+
+    <IntlProvider
+      messages={localizations["en_US"]}
+      locale="en-us"
+      defaultLocale="en-us"
+    >
+      <SystemContext.Provider value={providerValues}>
+        {children}
+      </SystemContext.Provider>
+    </IntlProvider>
```
damassi, line 130: This can be removed in favor of the Boot suggestion up above

damassi: Although we _will_ be using system context for triggering updates to the language. 

```tsx
const { setLanguage } = useSystemContext()
... 

<Select values={languages} onChange={lang => setLanguage(lang) />
```

That kind of thing.

src/v2/lang/en-us.json
```diff
@@ -0,0 +1,50 @@
+{
```
damassi, line 1: Wondering what other's think, but should we keep all per-language translations in one single file? I can imagine it getting messy, but at the same time I can also imagine it being functionally easier to use add to and revise, all things considered. 

Another option would be to follow the pattern that we've got setup for tests and which everyone is familiar with and add folder-level `__i18n__`. So like:

```bash
├── Bar.tsx
├── Foo.tsx
├── __i18n__
│   ├── Bar.i18n.json
│   └── Foo.i18n.json
└── __tests__
    ├── Bar.test.tsx
    └── Foo.test.tsx
```

damassi (commented):
PR is looking awesome @lidimayra 💯 

Can we add a testing example to this? I'm still uncertain about how that side of things will work, if we were to add intel and then run our existing tests.

damassi (commented):
Also echoing @dzucconi, I too am curious about alternative libs and so on. I know React-intl has been around the longest and has a lot users, but what else is out there? We might be able to get away with something simpler that doesn't require a compiler, `json` files and so on. 

_Totally naive implementation without understand any of the underlying gotchas_: 

```tsx
const content = {
  en_us: {
    nav: {
      about: "About"
      ... 
    }
  }
}

const useContent = () => {
  const { lang } = useSystemContext()
  
  return {
    content: content[lang]
  }
}
```

```tsx
const MyComponent = () => {
  const { content } = useContent() 
  
  return (
    <>
      <Text>{content.nav.about}</Text>
    </>
  )
}
```

We'd get instant typescript support with no effort, and everything is pretty straightforward. 

Again, I'm not an i18n expert, but I do wonder what's strictly needed for the scope of our site.

dzucconi:
@damassi Right, this is exactly what I was wondering. Seems like total overkill for our purposes but maybe I'm missing something?

dzucconi:
@damassi we'd be able to chose the bundle at runtime though? I suppose you could tie this to routing (e.g. `https://artsy.net/en-us/artist/andy-warhol`) — though that's a whole other convo.

damassi:
Another thing I woke up thinking about that I can't quite shake is this need for multiple language files, which will be hard to keep in sync (maybe thats what formatjs does? validate and scan all lang content files and evaluate subtle differences?). 

But that got me thinking of other possible structures if we were to go with a lightweight internal i18n lib:

```tsx
const content = {
  nav: {
    home: {
      en_us: "Home",
      es_es: "Casa", 
      ab_cd: "AAfuenf"
    }
  }
}
```

So in this case, there's only one single master graph for everything which eliminates any potential for drift, and if we need to add a new translation we just update the leaves of each word branch with a new language key / translation. We could then use a js proxy so that we wouldn't have to manually access the current language each time:

```
content.nav.home
```
then proxies to 
```
content.nav.home[lang]
```
Trying to think about how to achieve what we need to achieve with the least amount of moving parts. Having to manually maintain all of our old data fixtures through lib upgrades and so on triggers a bit of code ptsd in me... That's why if we can do simple, we should. Our site isn't _so_ big, all said.

damassi:
(Wanted to /cc @jonallured on this. Thoughts on i18n from past experience?)

jonallured:
👋 lots of good stuff to ponder in this PR thanks for getting us going on this @lidimayra!! ❤️ 

My only experience with i18n is in the context of a Rails app and a third party translation service. We worked with them to settle on a file format for the translations and then make a rake task to import them into the project. I don't remember all the details but I'll just note that whatever we do here should probably connect with how we'll work with a third party. I'd hate to see us make tech choices without taking into consideration the larger loop of working with a service ya know?

jest.config.legacy.js
lidimayra: Yes. This setting was leading our json files to be misinterpreted by the transpiler (build would fail trying to read json files as js files).

Example of failure when running `yarn test:ci:jest:v2`:
```
FAIL  src/v2/Apps/BuyerGuarantee/Routes/__tests__/BuyerGuaranteeIndex.jest.tsx
  ● Test suite failed to run

    SyntaxError: /app/src/v2/compiled-lang/en-us.json: Missing semicolon. (2:18)

      1 | {
    > 2 |   "navbar.artists": [
        |                   ^
      3 |     {
      4 |       "type": 0,
      5 |       "value": "Artists"

      32 |   useAuthIntent()
      33 |   useAuthValidation()
    > 34 |
         | ^
      35 |   const { children, match } = props
      36 |   const routeConfig = match ? findCurrentRoute(match) : null
      37 |   const { isEigen } = useSystemContext()
```

src/v2/compiled-lang/en-us.json
lidimayra: The ones in `lang` are the ones added by us.
In case we also adopt a third-party translation tool, these are the ones that are synced between our project and the tool.

The ones in `compiled-lang` are the ones that are actually consumed by the application.

More on this can be found here:
- [Message extraction](https://formatjs.io/docs/getting-started/message-extraction)
- [Message distribution](https://formatjs.io/docs/getting-started/message-distribution)

lidimayra:
Thank you so much for for the comments and the careful review! 🙌❤️  
I'm trying to go through everything today.

Something that might answer some questions that were raised by @damassi , @dzucconi and @jonallured : 
react-intl works with other third-party services (as it can be seen on [this page](https://formatjs.io/docs/getting-started/message-extraction))

At some point, we want to adopt one of these services (it would be useful for the product design team, and Erin has been talking about this [on #hack9-localizations channel](https://artsy.slack.com/archives/C02T87BHDAT/p1643025939006000)). A bit more on this is also available on [this notion page](https://www.notion.so/artsy/Internationalization-e4afd66b69804d9293c706df2111f75a).

Besides this, react-intl also provides everything we need to handle pluralization, date formatting, numbers, etc. And it's widely used lib, with great support from the community.

package.json
lidimayra: Much better ❤️ 
I made these changes and squashed them into the existing commits (and I updated the commit messages to reflect the instructions using these commands).

package.json
```diff
@@ -29,6 +30,7 @@
     "docker:electron": "yarn docker --target electron-runner -t force:electron",
     "docker:force": "yarn docker -t force",
     "docker:clean": "docker rmi force force:builder force:electron",
+    "extract": "formatjs extract",
```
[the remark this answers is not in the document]
lidimayra: Same as https://github.com/artsy/force/pull/9449#discussion_r819574939

src/v2/System/SystemContext.tsx
lidimayra: I made the changes to create an `InternationalizationProvider` and I'm importing it in `Boot`, but it's still not really clear to me how can I remove this from `SystemContext` without affecting the tests... 🤔

src/v2/compiled-lang/en-us.json
```diff
@@ -0,0 +1,98 @@
+{
```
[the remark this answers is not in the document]
lidimayra: Applied and squashed 😊

src/v2/Apps/lang/en-us.json
lidimayra: I moved this to `v2/System/i18n/lang` for now, but I can change this in case we decide we really want to have multiple locale files

src/v2/System/SystemContext.tsx
damassi: It might impact the tests, but in general all of our react context providers live in boot since that's the root of our application. For tests, if we need access to certain providers, we can wrap the test in `<MockBoot>`

lidimayra: Cool, thank you!!
I wrapped NavBar tests in `<MockBoot>` and those are passing now! 🙌 

Lack of `IntlProvider` in `SystemContext` is leading to issues on `/terms`, though... is this  one an exception that doesn't benefit from `Boot`?

Screenshot 2022-03-11 at 18 36 57

lidimayra:
There is another PR with a react-i18next implementation.
See #9797

src/v2/System/SystemContext.tsx
lidimayra: This started to happen after removing `IntlProvider` from `SystemContext` because `/terms` does not rely on `Boot`. As we can see [in the `.jade` files](https://github.com/artsy/force/blob/9e9af2114ff29b0565e10b88a80e9e081afec954/src/desktop/components/main_layout/header/templates/index.jade), we use some stich components, which replicate parts of the components using on static pages like these.

In order to make this work, we should also add the `IntlProvider` to the components in the [stitch_components ](https://github.com/artsy/force/tree/main/src/desktop/components/react/stitch_components) directory (components that can't benefit from `Boot`)

src/v2/Components/NavBar/NavBar.tsx
lidimayra: An implementation like this doesn't seem to be this straightforward with `react-intl`, but it's the natural way to go in `react-i18next`.

There is this other PR here with an attempt of using `react-i18next`: https://github.com/artsy/force/pull/9797

damassi:
👋 Heads up! A major change is incoming in Force, where we flatten out the `src/v2` directory into `src`. This will take place **next Thursday** and be merged in, which could impact PRs with rebase conflicts (since all of our import paths will soon change). See [this thread](https://artsy.slack.com/archives/C02BC3HEJ/p1657821372559579) for more info.

PR #9450, opened 2022-02-18
fix: palette breaking visual changes

dzucconi:
Nothing too exciting here just that https://github.com/artsy/palette/pull/1130 will introduce a breaking visual change. Since dropdowns are now `fixed` they are outside of their anchor's stacking context, which is also `fixed` with a z-index of `100`.

This fixes that and introduces a z-index map into the picture for all of our fixed elements. Modal isn't included here since Palette controls this and sets some weird super high value. (We may wish to bring that in here in the future though).


dzucconi:
(Will rebase once Palette is released)

PR #9451, opened 2022-02-18
[GRO-734](feat): small design fixes for /about2

TMcMeans:
I replaced the app store image in the App section to one without a caption and replaced the transparent app store buttons with ones with color. The big image is still low res, but this can get shipped without it for now.

### DIFF
--- a/src/v2/Apps/_About2/AboutApp.tsx
+++ b/src/v2/Apps/_About2/AboutApp.tsx
@@ -98,7 +98,7 @@ const SellWithArtsy: React.FC = () => {
             style={{
               objectFit: "cover",
             }}
-            alt="a Molly Green art piece titled Cached"
+            alt=""
           />
         </Box>
       )}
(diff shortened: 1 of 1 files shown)


src/v2/Apps/_About2/AboutApp.tsx
```diff
@@ -181,7 +184,7 @@ const AppDownload: React.FC = () => {
             rel="noopener noreferrer"
           >
             <Image
-              src="http://files.artsy.net/images/download-ios-app-transparent.svg"
+              src="http://files.artsy.net/images/download-ios-app.svg"
               mr={4}
```
The-Beez-Kneez, line 188: Non-blocking: Just curious @dzucconi I thought in general we do not apply margin spacing directly on palette <image>?

dzucconi: It's fine to do so. Just sometimes `Spacer` components are more clear. Kind of depends on the context, I guess.

dzucconi (approved):

src/v2/Apps/_About2/AboutApp.tsx
damassi: @The-Beez-Kneez - You might be thinking of the rule around how we don't apply margin spacing on outer-most container components?

PR #9452, opened 2022-02-18
chore: Remove artsy-passport example code

jacobherrington:
I'm fairly sure this is totally unused, so for the sake of future devs
maybe we should remove the example code.

cc @damassi Does this make sense? Is there any legacy reason you can think of not to remove all of this?

We don't really need the context now that we aren't trying to set up new applications with this package.


damassi:
Totally fine to remove (and agree that its confusing that its there) 👍

damassi (approved):

PR #9457, opened 2022-02-18
fix: Move checkbox higher in signup form GRO-835

jonallured:
In order to improve the signup process slightly this PR does two things:

* moves the checkboxes up under the form fields
* removes lying error messages

The way I did this was: copy/paste the checkbox code, call `preventDefault` on the click events firing from the third party auth buttons.

Looks like this:
Screen Shot 2022-02-18 at 2 12 13 PM
Screen Shot 2022-02-18 at 2 12 27 PM
Screen Shot 2022-02-18 at 2 33 49 PM

https://artsyproduct.atlassian.net/browse/GRO-835

/cc @artsy/grow-devs


jacobherrington (approved):
I like this more, for sure!

Definitely more improvements we can make after we've dissected the task a little better, but this should be a good way to address folks being confused by the very odd required field.

PR #9458, opened 2022-02-18
Remove references to removed auth providers

jacobherrington:
Twitter and LinkedIn authentication support [was previously removed](https://github.com/artsy/artsy-passport/pull/131), so I think we should delete these references to that code too.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -85,8 +85,6 @@ export const GOOGLE_SECRET: any = null
 export const GOOGLE_MAPS_API_KEY: any = null
 export const IMAGE_PROXY: any = "GEMINI"
 export const IP_DENYLIST: any = ""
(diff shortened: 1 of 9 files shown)


src/config.ts
```diff
@@ -85,8 +85,6 @@ export const GOOGLE_SECRET: any = null
 export const GOOGLE_MAPS_API_KEY: any = null
 export const IMAGE_PROXY: any = "GEMINI"
 export const IP_DENYLIST: any = ""
-export const LINKEDIN_KEY: any = null
-export const LINKEDIN_SECRET: any = null
```
jacobherrington, line 89: These keys are not set in production or staging according to hokusai.

damassi:
The best kind of PRs

PR #9459, opened 2022-02-18
feat(article): implements entity tooltips

dzucconi:
Closes: [GRO-749](https://artsyproduct.atlassian.net/browse/GRO-749)

https://user-images.githubusercontent.com/112297/154770922-9f9ba7c7-4bf7-4339-818e-15beb1be3982.mov

### DIFF
--- a/src/v2/Apps/Article/Components/ArticleHTML.tsx
+++ b/src/v2/Apps/Article/Components/ArticleHTML.tsx
@@ -1,13 +1,43 @@
-import { Box, BoxProps, THEME_V3 as THEME } from "@artsy/palette"
+import { Box, BoxProps, THEME_V3 as THEME, useDidMount } from "@artsy/palette"
 import { themeGet } from "@styled-system/theme-get"
 import { FC } from "react"
 import styled from "styled-components"
+import reactHtmlParser from "@artsy/react-html-parser"
+import { ArticleTooltip, isSupportedArticleTooltip } from "./ArticleTooltip"
 
 interface ArticleHTMLProps extends BoxProps {
   children: string
 }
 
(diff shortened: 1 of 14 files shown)


src/v2/Apps/Article/Components/ArticleHTML.tsx
```diff
@@ -1,13 +1,43 @@
-import { Box, BoxProps, THEME_V3 as THEME } from "@artsy/palette"
+import { Box, BoxProps, THEME_V3 as THEME, useDidMount } from "@artsy/palette"
 import { themeGet } from "@styled-system/theme-get"
 import { FC } from "react"
 import styled from "styled-components"
+import reactHtmlParser from "@artsy/react-html-parser"
```
dzucconi, line 5: Note to self to add this to the proper deps. I think it's currently a transient dep via reaction.

src/v2/Apps/Article/Components/ArticleHTML.tsx
```diff
@@ -1,13 +1,43 @@
...
+      <ArticleTooltip entity={entity} id={id} href={href}>
+        {node.textContent}
+      </ArticleTooltip>
+    )
+  }
+
+  if (isMounted) {
+    return <Container>{reactHtmlParser(children, { transform })}</Container>
+  }
+
```
dzucconi, line 40: Here we use our fork of `reactHtmlParser` — which "replaces the `htmlparser2` dependency with the `DOMParser` API provided by modern browsers" — this is great because the package's file size is _dramatically_ reduced and we only need to run this on the client-side anyway.

So on the server side: this returns the normal HTML, on the client-side it returns an identical looking body but with the tooltips inserted.

dzucconi:
Smoke test failure is unrelated: https://github.com/artsy/force/pull/9463 — this should be fine to merge

PR #9462, opened 2022-02-22
GRO-585: Final Copy update from legal to Preference Center checkboxes 

The-Beez-Kneez:
Incorporated the final copy changes from legal. 

Approved copy can be found in the following design docs:

https://www.figma.com/file/90ug4QmzA8cr1YKjagjFw1/Preference-Center?node-id=317%3A3484


jonallured (approved):
LGTM!! 🎉

PR #9463, opened 2022-02-22
fix(artists): prevents 500s when missing data

dzucconi:
This page was done prior to strict-mode being enabled which would have dealt with these errors. This page being down is causing a smoke test to fail as well.


damassi:
@dzucconi - How was this caught in relation to React.Strict mode? 

This was added carelessly in another PR without any wider site testing which frustrated me greatly. I understand its value, but also recognize the need to do be careful with these app wide settings. 

If it's bringing down pages what should be our next steps? Audit, or remove until we're ready to deliberately add in which case we can then properly audit before merge?

dzucconi:
Typescript strict mode, rather.

@damassi I don't think this has anything to do with feature flags? Just a long standing bug where we weren't guarding on null data.

damassi:
Oh, ok whew. I thought this was a React.Strict mode issue

PR #9471, opened 2022-02-23
feat(FX-3767): Use user’s preferred metric setting for size filter and saved search alerts

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3767]
Review app: https://user-length-preferences.artsy.net/artist/banksy/works-for-sale

### Acceptance Criteria
* Apply user’s preferred metric for size filter
* Display saved search pills in user’s preferred metric

#### IN is user’s preferred metric
https://user-images.githubusercontent.com/3513494/155298508-be1c7256-d768-4441-8446-6cfb522ce126.mp4

#### CM is user’s preferred metric
https://user-images.githubusercontent.com/3513494/155298547-95fff22c-197c-4ab6-89bb-b7bd4b8730a2.mp4

[FX-3767]: https://artsyproduct.atlassian.net/browse/FX-3767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -37,12 +41,24 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
     slug: artist.slug,
   }
 
+  const metric = getSupportedMetric(me?.lengthUnitPreference)
+  const filters: ArtworkFiltersState = {
+    metric,
+    ...match.location.query,
+  }
+
```
dimatretyak, line 49: Does anyone have an idea how best to implement this? Because now it is necessary to add these changes in all places where `ArtworkFilterContextProvider` is used

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -37,12 +41,24 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
...
       aggregations={aggregations}
       counts={artist.counts as Counts}
-      filters={match.location.query}
-      onChange={updateUrl}
+      filters={filters}
+      onChange={filterState => {
+        updateUrl(filterState, {
+          defaultValues: {
+            metric,
+          },
```
dimatretyak, line 59: These changes are necessary to ensure that the user’s preferred metric is not displayed in the url by default if it differs from the default metric.

### Example
 * [Default metric](https://github.com/artsy/force/blob/9dc6d876e8c7073aba6f3742585faebea2460bd5/src/v2/Components/ArtworkFilter/Utils/metrics.ts#L3) is `CM`
 * User’s preferred metric is `IN`
 * If there were no such changes, the url would contain `metric=IN` (something like this: `artsy.net/artist/banksy/works-for-sale?metric=IN`)

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -66,11 +65,11 @@ export const parseRange = (range: string = "", metric: Metric): Numeric[] => {
     if (s === "*") return s
     const value = parseFloat(s)
 
-    if (metric === "cm") {
+    if (metric === "CM") {
       return convertToCentimeters(value)
     }
 
-    return value
+    return Math.round(value)
```
dimatretyak, line 72: FIY: See discussion [here](https://artsy.slack.com/archives/C9SATFLUU/p1645464514742379?thread_ts=1645106514.729079&cid=C9SATFLUU)

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
dimatretyak: What do you think about making user's preferred metric available in [User](https://github.com/artsy/force/blob/9dc6d876e8c7073aba6f3742585faebea2460bd5/src/v2/System/SystemContext.tsx#L94) from `SystemContext`?

mdole: I think that makes sense! We would still have to update the different artwork grids, but I think it would just be a matter of adding `const { user } = useSystemContext(); const metric = getSupportedMetric(user.lengthUnitPreference)` which is definitely less steps.

I'm not enough of a Force expert to say for sure if `useSystemContext` is the preferred way of setting and fetching a value like this, but it seems to be widely used at the moment and looks reasonable to me.

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
mdole: I think it would probably be fine to display the metric in the URL 🤷 I don't feel strongly about it, but if this adds to the amount of code we have to add to all of our artwork grids and adds a bit of complexity for future developers, I might suggest we just show the metric in the address bar.

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -66,7 +65,7 @@ export const parseRange = (range: string = "", metric: Metric): Numeric[] => {
     if (s === "*") return s
     const value = parseFloat(s)
 
-    if (metric === "cm") {
+    if (metric === "CM") {
```
mdole, line 68: This makes me think - is there a reason we're doing these values as `CM` instead of `cm`? I think it's fine either way, we should just make sure that Gravity can handle both uppercase and lowercase 🤔

mdole (approved):
nice work @dimatretyak ! Agreed with the idea to make it easier to store a user's preferred metric somewhere central and accessible, otherwise I think this is looking good.

I know we're going to work on storing metrics on SearchCriteria and fetching them, but my feeling is that we could release this work and iterate on it without having to rip out a lot of code. Let me know if you agree and we can get it merged!

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
dimatretyak: > is there a reason we're doing these values as CM instead of cm

I decided to store values in Force in the same format as they are stored on Gravity side

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
damassi: Similar to @mdole's suggestion, would it be possible to add this as a setting on the settings page? and then that result would be available on the user object. _How_ we get that result is up to us. For example, a hook in the ArtworkFilter:

```tsx
const useUserMetricSetting = () => {
  const metric = getSupportedMetric(user.lengthUnitPreference)
  return metric 
}
```

> the user will need to log in again to get a fresh

This is a good question. When the app first boots (meaning, when a user first arrives) we populate the `user` object via `getServerApp` from [here](https://github.com/artsy/force/blob/main/src/v2/System/Router/buildServerApp.tsx#L106) (and corresponding `getClientApp` on the client), which comes from express.js's `req.user`. So once that's loaded on a page request its set, and if anything were to change we'd have to find a way to refresh this object. This is a good problem to solve for that we'll need to eventually get to as force's framework code continues to grow more sophisticated. 

@dzucconi - have we done anything like this before? I can't recall.

damassi: Another thing about getting the user via `useSystemContext` is I think it would be ok to expand the system context with a `me` prop that makes a `fetchQuery` request out to MP for the `me` field and returns all associated data. Correspondingly, there could be an 
```
const { updateMe } = useSystemContext()
```
which calls a mutation to update `me`. 


I'm not sure exactly what the best architecture for this is off the top of my head, but we use `me` enough within the app to arguably be able to access it through the system context. Note the difference between `me` and `user` -- me has a lot more info, and we've built a lot of UI features on top of it. This wouldn't be a replacement for proper relay-based components that call `me` via a fragment container (we don't want to break Relay's data-masking). 

But that said, the info on me is valuable, and being able to access it globally in this way could be useful.  

This might be a larger discussion worth raising in Slack, however.

damassi: Slack thread for posterity: https://artsy.slack.com/archives/CP9P4KR35/p1646066609727149

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
mdole: Worth noting that we also want users to be able to toggle back and forth on-page, and we plan to save that unit (`in/cm`) on individual saved alerts as well. 

Here's how I'm thinking of the actions we need to support:
- Preference is stored on user and fetched on page load
- User can still toggle to the other metric that is not their preference
- User can save an alert with either unit

I don't have a strong opinion about in vs. out of URL as long as all of this works 🤷‍♂️

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
dimatretyak: @damassi how do you feel about keeping the current implementation of getting the user's preferred metric?

I tested solution that was [discussed yesterday](https://artsy.slack.com/archives/CP9P4KR35/p1646070286694939?thread_ts=1646066609.727149&cid=CP9P4KR35). The main disadvantage is that it will be executed _asynchronously_ and at the time of rendering filters there are no guarantees that the request will be fully executed (we can manually synchronize the values, but is it worth doing so?)
```javascript
export const useUserPreferences = () => {
  const [metric, setMetric] = useState<Metric>(DEFAULT_METRIC)

  useEffect(() => {
    const exec = async () => {
      try {
        const response = await fetchQuery<useUserPreferencesQuery>(
          relayEnvironment!,
          graphql`
            query useUserPreferencesQuery {
              me {
                lengthUnitPreference
              }
            }
          `,
          {}
        )

        setMetric(response.me?.lengthUnitPreference)
      } catch (error) {
        console.log("[debug] error", error)
      }
    }

    exec()
  }, [relayEnvironment])

  return { metric }
}
```

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
dimatretyak: I agree with @mdole that metric should be in the URL

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
damassi: After our discussion yesterday this occurred to me! That we won't be able to get everything during the SSR pass and then it could cause a lot of staggered UI shifting later. So yeah, totally fine with the current approach if we need it to be synchronous 👍

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
damassi: Sounds good!

src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
dimatretyak: Nice. Thank you!

PR #9476, opened 2022-02-23
feat(auction2): Add promoted sale rail

damassi:
This adds the promoted sale "Buy Now" rail to the page:

Screen Shot 2022-02-23 at 11 03 01 AM

cc @artsy/grow-devs


damassi:
Ah @dzucconi's internet is down, assigning to myself but can follow up feedback-wise.

PR #9486, opened 2022-02-24
feat(FX-3759): add slider for price filter

dimatretyak:
Type: **Feature**
Jira ticket: [FX-3759]
Review app: https://slider-for-price-filter.artsy.net/artist/banksy/works-for-sale

### Some notes
* If custom price inputs are empty, then artworks without price restrictions are displayed (`priceRange=*-*`)
* If the user manually specified values in custom size inputs, then artworks will be displayed from the specified price range
  * if the user specified `min price = 15000`, `max price = 25000`, then artworks from **$15,000-$25,000** price range will be displayed (`priceRange=15000-25000`) 
  * if the user specified `min price = 35000`, `max price = 65000`, then artworks from **$35,000-$65,000** price range will be displayed (`priceRange=35000-65000`)
  * if the user specified only `min price = 25000`, then artworks from **$25,000 and above** price range will be displayed (`priceRange=25000-*`)
  * if the user specified only `max price = 45000`, then artworks from **$0 to $45.000** price range will be displayed (`priceRange=*-45000`)
* When the max slider is all the way to the right it effectively means no upper bound on the price (`priceRange=*-*`)

### Demo
https://user-images.githubusercontent.com/3513494/155761323-e852937e-e936-4b6e-817d-3c5ea7022e54.mp4

[FX-3759]: https://artsyproduct.atlassian.net/browse/FX-3759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -0,0 +1,211 @@
...
+  return range.map((value, index) => {
+    if (value === "*") {
+      return DEFAULT_RANGE[index]
+    }
+
+    return value as number
+  })
+}
+
+const convertToArtworkFilterFormatRange = (range: number[]) => {
```
gkartalis, line 63: **suggestion(non-blocking):** 
```suggestion
const convertToFilterFormatRange = (range: number[]) => {
```

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -0,0 +1,211 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [shouldStageFilterChanges]
+  )
+
+  useEffect(() => {
+    // if price filter or filters state is being reset, then also clear local input state
+    if (reset || priceRange === DEFAULT_PRICE_RANGE) {
+      setRange(DEFAULT_CUSTOM_RANGE)
+    }
+  }, [reset, priceRange])
```
gkartalis, line 113: 🤘😍

gkartalis (approved):
Thanks for the great test cases 💯

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
dimatretyak: Done

gkartalis (approved):

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -0,0 +1,211 @@
...
+import styled from "styled-components"
+import { themeGet } from "@styled-system/theme-get"
+import { FilterExpandable } from "./FilterExpandable"
+import { isCustomValue } from "./Utils/isCustomValue"
+import { useFilterLabelCountByKey } from "../Utils/useFilterLabelCountByKey"
+import { Range, RANGE_DOT_SIZE } from "v2/Components/Range"
+import { debounce } from "lodash"
+
+// Disables arrows in numeric inputs
+export const NumericInput = styled(LabeledInput).attrs({ type: "number" })`
```
damassi, line 16: I'd suggest moving this NumericInput component below the main `PriceFilterNewComponent` since its mostly just CSS and is secondary to the file's main concern.

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -0,0 +1,211 @@
...
+      return "*"
+    }
+
+    return value
+  })
+}
+
+export const getValue = (value: CustomRange[number]) => {
+  return value === "*" || value === 0 ? "" : value
+}
```
damassi, line 75: (also non-blocking): All of these other support functions can also live at the bottom of the component.

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const [range, setRange] = useState(parseRange(priceRange))
+  const sliderRange = parseSliderRange(range)
+  const [minValue, maxValue] = range
+  const [defaultMinValue, defaultMaxValue] = DEFAULT_RANGE
+
+  const filtersCount = useFilterLabelCountByKey(
+    SelectedFiltersCountsLabels.priceRange
+  )
+  const label = `Price${filtersCount}`
+  const selection = currentlySelectedFilters?.().priceRange
```
damassi, line 99: Since we're already calling this above do we need to call it again? 

(side-thought: It might be worth quickly refactoring this `currentlySelectedFilters` behind a `useCurrentlySelectedFilters` hook so we don't always have to call `useArtworkFilterContext` to get at this commonly used function)

mdole (approved):
nice! let's address Chris' feedback and get this shipped 🚢 . Great work!

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
dimatretyak: @damassi do you mind if these changes are added in a separate PR?

src/v2/Components/Range.tsx
```diff
@@ -0,0 +1,132 @@
+import { themeGet } from "@styled-system/theme-get"
+import { Range as RCRange, RangeProps } from "rc-slider"
+import styled from "styled-components"
+
+const BORDER_RADIUS_BASE = 6
+export const RANGE_DOT_SIZE = 16
```
anandaroop, line 6: Sorry for late feedback, but should `RANGE_DOT_SIZE` should be 32 to match Figma?

Screen Shot 2022-03-13 at 12 10 50 PM

dimatretyak: If we set the size to 32px, it will look too big. I discussed this point with Lois and showed this problem. She agreed that it was better to specify a smaller size

image

package.json
```diff
@@ -209,7 +209,7 @@
     "query-string": "5.1.1",
     "raf": "3.4.0",
     "raven-js": "3.27.0",
-    "rc-slider": "7.0.8",
+    "rc-slider": "^9.7.5",
```
dzucconi, line 212: So we actually want to remove this dependency. There's no need for it.

This range component is a great example of something that should be in the design system. I'll port a dependency free version over to Palette then replace this.

dzucconi: Reason being:



This library somehow manages to include it's own tooltip engine; among other crazy dependencies.

PR #9487, opened 2022-02-24
feat(NX-3082): change CTA copies

araujobarret:
[NX-3082]

We're changing the copies around `Buy now`, `Make offer` and `Contact for Price` but only the copies, we are not touching in any `saleMessage` coming from Gravity or other services, nor the language elsewhere, example we still call Buy now and everything related to it, only the copy has changed to Purchase.

Artwork page (Price on Request)

Search screen (Price on Request)

Artwork page (Purchase & Make an Offer)

Conversation (Make an Offer)

[NX-3082]: https://artsyproduct.atlassian.net/browse/NX-3082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -91,9 +91,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
   renderSaleMessage(saleMessage: string) {
     return (
       <Text variant="lg" data-test="SaleMessage">
-        {saleMessage === "Contact For Price"
-          ? "Contact for Price"
-          : saleMessage}
+        {saleMessage === "Contact For Price" ? "Price on Request" : saleMessage}
       </Text>
     )
   }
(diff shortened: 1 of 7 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -509,7 +507,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
                 loading={isCommittingCreateOfferOrderMutation}
                 onClick={this.handleCreateOfferOrder.bind(this)}
               >
-                Make offer
+                Make an Offer
```
jo-rs, line 510: Should this be capitalized like this? 🤔 
I'm seeing on the ticket that it's marked that way, but might be worth confirming with PM and design; I think we moved away from that way of capitalizing.

jo-rs: Same with "Price on request"

jo-rs (approved):
Nice nice 💯

PR #9493, opened 2022-02-25
feat: remove high speed bump modal from offer

araujobarret:
[NX-3074]

- Removes high-speed bump modal when an offer is higher than the list price
- Fixes an issue when the price was hidden we showed the low-speed bump modal informing the offer was X% under the list price

@artsy/negotiate-devs 

[NX-3074]: https://artsyproduct.atlassian.net/browse/NX-3074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Routes/Offer/index.tsx
+++ b/src/v2/Apps/Order/Routes/Offer/index.tsx
@@ -49,7 +49,6 @@ export interface OfferState {
   offerNoteValue: TextAreaChange
   formIsDirty: boolean
   lowSpeedBumpEncountered: boolean
-  highSpeedBumpEncountered: boolean
   isToggleRadio: boolean
 }
 
@@ -59,7 +58,6 @@ const logger = createLogger("Order/Routes/Offer/index.tsx")
 export class OfferRoute extends Component<OfferProps, OfferState> {
   state: OfferState = {
     formIsDirty: false,
-    highSpeedBumpEncountered: false,
     lowSpeedBumpEncountered: false,
     offerNoteValue: { exceedsCharacterLimit: false, value: "" },
     offerValue: 0,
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -90,20 +88,6 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
     })
   }
 
-  @track<OfferProps>(props => ({
```
araujobarret, line 93: data team is in the loop about the event removal

src/v2/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -181,6 +160,7 @@ export class OfferRoute extends Component<OfferProps, OfferState> {
       ?.displayPriceRange
 
     if (
+      !isPriceHidden &&
```
araujobarret, line 163: ditto here, I'm informing them we're going to reduce the number of events because of the bug fix

src/v2/Apps/Order/Routes/__tests__/Offer2.jest.tsx
```diff
@@ -47,7 +47,7 @@ describe("an offer on the work with price hidden", () => {
   })
 
   it("shows no modal warning when an offer made on work with hidden price", async () => {
-    await page.setOfferAmount(2000000)
+    await page.setOfferAmount(2)
```
araujobarret, line 50: now we're doing for the lower value scenario, before we were not doing it

src/v2/Apps/Order/Routes/Offer/index.tsx
jo-rs: Thank you for adding that! 🙏

jo-rs (approved):
Lovely. Looks great! 💪

PR #9496, opened 2022-02-25
chore: Price DB copy update

rajsam003:
This PR resolves [CX-2346](https://artsyproduct.atlassian.net/browse/CX-2346) 

|Before|
|---|
|image|
|After|
|image|

### DIFF
--- a/src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
+++ b/src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
@@ -19,7 +19,7 @@ export const PriceDatabaseBenefits: React.FC = () => {
       <GridColumns mt={4} gridRowGap={[2, 0]}>
         <Column span={12}>
           <Text as="h1" variant={["xl", "xxl"]}>
-            Auction records from 340,000
+            Auction records from 300,000
             <br />
             artists—and counting
           </Text>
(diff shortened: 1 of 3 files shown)


src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
```diff
@@ -19,24 +19,34 @@ export const PriceDatabaseBenefits: React.FC = () => {
...
 
       <Section
-        title="The largest publicly available art market database"
-        text="With millions of auction records, our growing database encompasses over 900 auction houses with pricing that dates back to 1986. Get real-time records from history-making auction houses like Christie's, Sotheby's, Phillips, Bonhams, and more."
+        title="Get in-depth art market data"
+        description={
+          <Text variant="sm">
+            Browse millions of current and historical results from leading
+            auction houses across the globe.
+          </Text>
```
olerichter00, line 35: What is the reason for this change?

olerichter00 (commented):
Thanks for updating the copy!

Will there be a follow-up to add a section for Value Prop 4?

src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
rajsam003: @MounirDhahri 👀?

olerichter00 (approved):

src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
MounirDhahri: we just spoke of this during the standup.

MounirDhahri: @olerichter00 yes, we will add the value prop 4 once we get the image for value prop 4 from the Brand team

MounirDhahri (approved):

src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
```diff
@@ -70,10 +86,10 @@ export const PriceDatabaseBenefits: React.FC = () => {
 
 const Section: React.FC<{
   title: string
-  text: string
+  description: ReactElement
```
MounirDhahri, line 89: can we revert to using text as a prop here until we get the value prop 4

MounirDhahri (changes requested):

src/v2/Apps/PriceDatabase/Components/PriceDatabaseBenefits.tsx
MounirDhahri: But I agree, it makes sense to revert this until we need it

PR #9497, opened 2022-02-25
feat[FX-3718]: update color palette

mdole:
Updates Force's color filter palette to reflect the changes from https://github.com/artsy/gravity/pull/15040 and https://github.com/artsy/gravity/pull/15051/files.

This should not be merged until all the Gravity PRs have been merged and artworks are reindexed to have the new color values!

Screen Shot 2022-02-25 at 6 37 50 PM

Jira ticket: https://artsyproduct.atlassian.net/browse/FX-3718

### DIFF
--- a/src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
+++ b/src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
@@ -26,8 +26,8 @@ export type Color =
   | "yellow"
 
 interface Props {
-  medium: Medium | null
-  color: Color | null
+  medium: Medium | undefined
+  color: Color | undefined
 }
 
 export function getMetadata({ medium, color }: Props) {
--- a/src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
+++ b/src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
@@ -15,21 +15,19 @@ export type Medium =
 
 export type Color =
   | "black-and-white"
-  | "darkblue"
-  | "darkgreen"
-  | "darkviolet"
+  | "blue"
   | "gold"
-  | "lightblue"
-  | "lightgreen"
+  | "gray"
+  | "green"
   | "orange"
   | "pink"
+  | "purple"
   | "red"
-  | "violet"
   | "yellow"
 
 interface Props {
-  medium: Medium
-  color: Color
+  medium: Medium | null
+  color: Color | null
 }
 
 export function getMetadata({ medium, color }: Props) {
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
```diff
@@ -15,21 +15,19 @@ export type Medium =
 
 export type Color =
   | "black-and-white"
```
The-Beez-Kneez, line 17: Nonblocking:

Why do we have `Black and White` as a combined filtered option? I know that we specifically have up right now a highlight for the "Black" show, which primarily features all black pieces. So why would these not be their own individual color options?

damassi: @The-Beez-Kneez - think B&W photography. Monochrome is a whole sub-world inside the art world.

src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
```diff
@@ -15,21 +15,19 @@ export type Medium =
...
   | "pink"
+  | "purple"
   | "red"
-  | "violet"
   | "yellow"
 
 interface Props {
-  medium: Medium
-  color: Color
+  medium: Medium | null
```
nickskalkin, line 29: @mdole could you elaborate on this change? Why are we doing them optional?

nickskalkin (approved):
👍

src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
mdole: Yeah, what @damassi said is correct. That said, it's an interesting question, and we certainly could consider having black & white as separate filters (because there are plenty of works that have lots of black or white in them). 

The way this filter works right now is [it checks](https://github.com/artsy/gravity/blob/1baa64541f1f2f75c6005a38688928ad4e445636/app/models/util/color_distance.rb#L63) if a work is _only_ black and white. So a work shouldn't be able to return both "black & white" and "red", for example.

Worth considering if we should split out black/white filters that could be included with other colors. I might suggest we do so for `black` but not for `white` because a white filter would return lots of works that have white backgrounds. There are lots of works that are shot against black backgrounds or have white matting/framing; it's hard to know if that would be enough to really skew the filter balance (i.e. showing lots of colorful works that happen to be shot against white backgrounds, rendering the filter somewhat useless) but I think it's possible.

src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
mdole: good question @nickskalkin! I originally added the `| null` because that's how they were being [passed in the tests](https://github.com/artsy/force/blob/b149001f55aefda4e311d252a5b67dcb8e8be1c3/src/v2/Apps/Collect/Routes/Collect/Utils/__tests__/getMetadata.jest.ts#L8).

Your comment encouraged me to make sure I understood what was happening here. It turns out that the color/medium values defined in `getMetadata` are used for these "alternate routes" for the /collect page. For example, you can hit https://www.artsy.net/collect/painting or https://www.artsy.net/collect/color/yellow.

I'm guessing those exist for SEO reasons. Either way, it turns out that if the user _doesn't_ hit one of those routes but just hits the collect page normally (in other words, https://www.artsy.net/collect/ or https://www.artsy.net/collect/painting?colors%5B0%5D=yellow), then `getMetadata()` will return `undefined` for both color and medium.

In summary: these types should be `| undefined` because they will often _be_ undefined. I just pushed up another commit that changes this to `| undefined` and updates the test as well.

Thanks for flagging! Would have been misleading to introduce `| null` here :)

anandaroop (approved):

src/v2/Apps/Collect/Routes/Collect/Utils/getMetadata.ts
mdole: 👍 Posed the question to FX's PM and designer [in Slack](https://artsy.slack.com/archives/C9SATFLUU/p1646146804105089), will see what they say!

PR #9498, opened 2022-02-25
mc-jones/fix unleash implementation

mc-jones:
This PR is a follow-up to the initial implementation of Unleash in Force, which led to an incident due to incorrect logic in `featureFlagMiddleware`. 

### Issue
In `featureflagMiddleware`, we checked [here](https://github.com/artsy/force/blob/2706f4a2edd6b55ebd06ca9c765e8f68fccbbea1/src/lib/middleware/featureFlagMiddleware.ts#L12-L20) if there is an existing unleash client, and create one if there isn’t. We had the logic wrong and were creating a new instance of the unleash client on every request here, which caused the huge [spike of requests, leading to high latency and timeouts](https://app.datadoghq.com/apm/service/force/express.request?env=production&panel_end=1645656270957&panel_paused=true&panel_start=1645652670957&topGraphs=latency%3Alatency%2CbreakdownAs%3Apercentage%2Cerrors%3Acount%2Chits%3Acount&start=1645074000000&end=1645160340000&paused=true).  There is a review app up which simply [wraps the try/catch block](https://github.com/artsy/force/blob/review-app-unleash-test/src/lib/middleware/featureFlagMiddleware.ts#L14-L21) in an else statement, which seems to have it working as intended ([polling every 15 seconds](https://my.papertrailapp.com/events?q=unleash+ingress+UA%3D%22review-app-unleash-test%22)).

Full Incident thread [here](https://artsy.slack.com/archives/C033592FNBZ)

### Original PR
https://github.com/artsy/force/pull/9393

### Next Steps
- Remove custom react context implementation in favor of using existing `SystemContext`. [PLATFORM-3918]
- Allow for experimentation by implementing `useVariant` and analytics tracking. [PLATFORM-3922]

[PLATFORM-3918]

[PLATFORM-3918]: https://artsyproduct.atlassian.net/browse/PLATFORM-3918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/middleware/featureFlagMiddleware.ts
+++ b/src/lib/middleware/featureFlagMiddleware.ts
@@ -15,7 +15,7 @@ export function featureFlagMiddleware(serviceType: symbol) {
         try {
           service = await createFeatureFlagService(serviceType)
           resolve(service)
-        } catch {
+        } catch (error) {
           reject("An unknown error occurred while creating the flag service.")
         }
       }
@@ -41,7 +41,7 @@ export function featureFlagMiddleware(serviceType: symbol) {
         next()
       })
       .catch(error => {
-        console.error(error)
+        console.error("[Force] Error creating feature flag service:", error)
         next()
       })
   }
--- a/src/lib/middleware/featureFlagMiddleware.ts
+++ b/src/lib/middleware/featureFlagMiddleware.ts
@@ -14,7 +14,6 @@ export function featureFlagMiddleware(serviceType: symbol) {
       } else {
         try {
           service = await createFeatureFlagService(serviceType)
-          console.log("resolved Unleash")
           resolve(service)
         } catch {
           reject("An unknown error occurred while creating the flag service.")
--- a/src/lib/middleware/featureFlagMiddleware.ts
+++ b/src/lib/middleware/featureFlagMiddleware.ts
@@ -1,0 +1,48 @@
+import { NextFunction } from "express"
+import { ArtsyRequest, ArtsyResponse } from "./artsyExpress"
+import {
+  createFeatureFlagService,
+  FeatureFlagService,
+} from "lib/featureFlags/featureFlagService"
+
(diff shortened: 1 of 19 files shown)


src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -0,0 +1,48 @@
...
+  return (req: ArtsyRequest, res: ArtsyResponse, next: NextFunction) => {
+    new Promise<FeatureFlagService>(async (resolve, reject) => {
+      if (service) {
+        resolve(service)
+      } else {
+        try {
+          service = await createFeatureFlagService(serviceType)
+          resolve(service)
+        } catch {
+          reject("An unknown error occurred while creating the flag service.")
```
damassi, line 19: Why not reject the actual error (which is known, because if it were not then the `catch` block wouldn't exectute)? 

```
try {
  ...
} catch (error) {
  reject(error)
}
```
This is then handled below at line 44 (see comment)

src/lib/middleware/featureFlagMiddleware.ts
```diff
@@ -0,0 +1,48 @@
...
+            res.locals.sd.featureFlags[flag] = service.enabled(
+              flag,
+              featureFlagContext
+            )
+          }
+        }
+        next()
+      })
+      .catch(error => {
+        console.error(error)
```
damassi, line 44:
```tsx
console.error("[Force] Error creating feature flag service:", error)
```

damassi: As a side note, this mixture of `async / await` and vanilla promises (`.then/.catch`) is a bit confusing. Best to stick with `async / await` everywhere per current JS standards.

joeyAghion (approved):

PR #9499, opened 2022-02-25
feat(GRO-757): article infinite scroll

dzucconi:
Closes: [GRO-757](https://artsyproduct.atlassian.net/browse/GRO-757)

https://user-images.githubusercontent.com/112297/155785043-fc994a9c-0e1e-489a-a750-c3056dbf675c.mov


dzucconi:
I'm wondering if we should put a limit on how many articles can be loaded in this fashion — since obviously performance will degrade quite a bit if you get deep.

damassi (approved):
Looks perfect 💯

PR #9512, opened 2022-02-28
Make Marquee static with Cultural Attributes

The-Beez-Kneez:
This is a quick PR to make the banner static for the Cultural Attributes banner.

Current Behavior:
Uploading Screen Recording 2022-02-28 at 7.15.32 AM.mov…

New Behavior:
Screen Shot 2022-02-28 at 7 16 25 AM

### DIFF
--- a/src/v2/Apps/Partner/PartnerApp.tsx
+++ b/src/v2/Apps/Partner/PartnerApp.tsx
@@ -46,7 +46,11 @@ export const PartnerApp: React.FC<PartnerAppProps> = ({
         <PartnerHeader partner={partner} />
 
         <FullBleed mb={[2, 4]}>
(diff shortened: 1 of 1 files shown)


dzucconi (approved):
Technically invalid but this effectively works to stop the animation by 'breaking' it.

The-Beez-Kneez:
Lolz @damon I noticed we were not passing back any kind of static.

Should I do the whole palette change and then this one?

dzucconi:
This is fine for now if you want to get the quick change in. I suspect we'll want to actually change the props API completely. Since `speed` is really `duration` and we'll want to have the speeds under control of the system / handled responsively better

The-Beez-Kneez:
The speeds, may be destroyed completely since design is now walking that back. So I am fine with just this change for now so we can at least move forward, till they have settled on a decision.

PR #9513, opened 2022-02-28
refactor: add featureFlags to SystemContext

mc-jones:
This PR simplifies the Unleash implementation by removing the previously used custom React `Context.Provider` in favor of `Force`'s `SystemContext`. 

## Related PRs
- https://github.com/artsy/force/pull/9498
- https://github.com/artsy/force/pull/9393

## Follow-ups
- Allow for experimentation by implementing `useVariant` and analytics tracking. [PLATFORM-3922]

[PLATFORM-3922]: https://artsyproduct.atlassian.net/browse/PLATFORM-3922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/featureFlags/FeatureFlagProvider.tsx
+++ b/src/lib/featureFlags/FeatureFlagProvider.tsx
@@ -1,26 +1,10 @@
-import * as React from "react"
-
-const FeatureFlagContext = React.createContext<EnabledFeatureFlags>({})
+import { useSystemContext } from "v2/System"
 
 export type EnabledFeatureFlags = Record<string, boolean>
 
-export function FeatureFlagProvider({
-  featureFlags,
-  children,
-}: {
-  featureFlags: EnabledFeatureFlags
-  children: any
-}) {
-  return (
-    <FeatureFlagContext.Provider value={featureFlags}>
-      {children}
-    </FeatureFlagContext.Provider>
-  )
-}
-
 export function useFeatureFlag(flagName) {
-  const context = React.useContext(FeatureFlagContext)
-  if (context === undefined) {
+  const { featureFlags } = useSystemContext()
+  if (featureFlags === undefined) {
     throw new Error("useFeatureFlag must be used within a FeatureFlagProvider")
   }
 
@@ -28,5 +12,5 @@ export function useFeatureFlag(flagName) {
     return null
   }
 
-  return Object.keys(context).indexOf(flagName) !== -1 && context[flagName]
+  return Object.keys(featureFlags).indexOf(flagName) !== -1 && featureFlags[flagName]
 }
--- a/src/lib/featureFlags/FeatureFlagProvider.tsx
+++ b/src/lib/featureFlags/FeatureFlagProvider.tsx
@@ -5,12 +5,16 @@ export type EnabledFeatureFlags = Record<string, boolean>
 export function useFeatureFlag(flagName) {
   const { featureFlags } = useSystemContext()
   if (featureFlags === undefined) {
-    throw new Error("useFeatureFlag must be used within a FeatureFlagProvider")
-  }
-
-  if (!flagName) {
+    console.error("[Force] Error: featureFlags is undefined in SystemContext")
     return null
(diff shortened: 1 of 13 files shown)


src/lib/featureFlags/FeatureFlagProvider.tsx
```diff
@@ -1,32 +1,20 @@
...
-      {children}
-    </FeatureFlagContext.Provider>
-  )
-}
-
 export function useFeatureFlag(flagName) {
-  const context = React.useContext(FeatureFlagContext)
-  if (context === undefined) {
-    throw new Error("useFeatureFlag must be used within a FeatureFlagProvider")
+  const { featureFlags } = useSystemContext()
```
damassi, line 6: This file can be moved to `v2/System`

src/lib/featureFlags/FeatureFlagProvider.tsx
```diff
@@ -1,32 +1,20 @@
...
   }
 
   if (!flagName) {
+    console.error("[Force] Error: cannot find flagName in featureFlags")
     return null
   }
 
-  return Object.keys(context).indexOf(flagName) !== -1 && context[flagName]
+  return (
+    Object.keys(featureFlags).indexOf(flagName) !== -1 && featureFlags[flagName]
```
damassi, line 18: I think this logic might be slightly simplified, and we can lean on typescript here as well for additional protection. 

Would suggest something like:

```tsx
export function useFeatureFlag(flagName: string): string | null {
  const { featureFlags } = useSystemContext()
  
  if (!featureFlags) {
    console.error("[Force] Error: featureFlags is undefined in SystemContext")
    return null
  }
  
  const flagName = featureFlags[flagName]
  
  if (!flagName) {
    console.error("[Force] Error: cannot find flagName in featureFlags: " Object.keys(featureFlags))
  }
  
  return flagName
}
```

damassi (dismissed):
Looks great @mc-jones 👍 . One suggestion / relocation request around `useFeatureFlag.tsx`, and it'd be nice to see a basic test for the hook, but otherwise LGTM 👍

src/v2/System/Router/buildClientApp.tsx
```diff
@@ -113,7 +113,7 @@ export function buildClientApp(config: RouterConfig): Promise<Resolve> {
             user={user}
             relayEnvironment={relayEnvironment}
             routes={routes}
-            featureFlags={featureFlags}
+            featureFlags={getENV('featureFlags')}
```
damassi, line 116: this is minor and non-blocking, but all of our other ENV variables, even set from sharify, are `FEATURE_FLAGS`, and it would be good to remain consistent here across the board.

src/lib/featureFlags/FeatureFlagProvider.tsx
mc-jones: Have added the typing, but going to hold off on changing this too much, as I think it may change not insignificantly as we add in the `useVariant` [functionality](https://docs.getunleash.io/advanced/toggle_variants).

mc-jones:
> Looks great @mc-jones 👍 . One suggestion / relocation request around `useFeatureFlag.tsx`, and it'd be nice to see a basic test for the hook, but otherwise LGTM 👍

@damassi - Totally makes sense. ✔️ .

src/v2/System/SystemContext.tsx
```diff
@@ -77,6 +77,9 @@ export interface SystemContextProps extends SystemContextState {
 
   /** Is there a logged in user? */
   isLoggedIn?: boolean
+
+  /** FeatureFlags */
+  featureFlags?: Record<string, boolean>
```
damassi, line 82: @mc-jones - didn't catch this before merge, but one tiny little nbd follow up is this could use the `EnabledFeatureFlags` type that was added below and shared with `Boot`

PR #9519, opened 2022-03-01
feat(GRO-755): implements ads

dzucconi:
This doesn't quite close out the ticket. I'm going to go over the logic tomorrow with Rose to deal with the inter-article ads. I'll write down a real spec based on what I find.

I also need to add in targeting data and click + impression tracking. So a follow up tomorrow.

------

Initially I ported over what we had in Reaction which used: https://github.com/nfl/react-gpt but this library is unmaintained and makes use of an API that Google is removing completely in exactly a month.

So I then tried to just use the [GPT library directly](https://developers.google.com/publisher-tag/guides/get-started). Long story but, turns out: it's very difficult to manage within a SPA. I couldn't get typings to work and when things aren't just so the ads simply don't load with relatively opaque error messages, which are unexplained.

Then I randomly stumbled into https://github.com/jaanauati/react-dfp — which I've included here. This works fine. The key part is that it spins up a manager that sits inside of context and gives us a reasonable API for doing impression tracking.

------

One minor innovation here is that I made these responsive obviating the need for the smaller 'mobile' size. Previously ads would just break the layout of the page at certain dimensions, now they responsively scale.


src/v2/Components/Ad/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  "Desktop_RightRail1",
+  "Desktop_TopLeaderboard",
+  "Mobile_InContentLB1",
+  "Mobile_InContentLB2",
+  "Mobile_InContentLBRepeat",
+  "Mobile_InContentMR1",
+  "Mobile_InContentMR2",
+  "Mobile_InContentMRRepeat",
+  "Mobile_TopLeaderboard",
+] as const
```
dzucconi, line 16: Oddly, I still don't know what these strings do or mean. We only show a single ad at a time site-wide (in the 3 sizes). They have no bearing on how the ads are displayed. I'll just copy how Reaction uses them and add a comment explaining my confusion tomorrow if my session with Rose doesn't turn up any clues.

src/v2/Components/Ad/Ad.tsx
```diff
@@ -0,0 +1,56 @@
+import { FC, memo } from "react"
```
damassi, line 1: Is it premature to add this to `v2/Components`?

dzucconi: Yeah; I suppose there's no intention to show ads anywhere else.

damassi (dismissed):
Straightforward 👍

jonallured (commented):
Heroic! 🏆

PR #9526, opened 2022-03-02
feat(FX-3780): Display filter criteria for each alert in list view

gkartalis:
# Description

- Utilize the [new human-friendly criteria labels](https://github.com/artsy/metaphysics/pull/3762) to display pills for filters applied for each saved alert
- add ability to expand/collapse filters 
- remove `getNamePlaceholder` function and create default titles in alerts only with the artist name

Jira: [FX-3780](https://artsyproduct.atlassian.net/browse/FX-3780)
Design: [Figma specs](https://www.figma.com/file/xsnfIWE7H67ZndGoEPVtN0/Alerts-2022?node-id=1703%3A26768)

|Screenshots|-|
|--|--|
|Screenshot 2022-03-09 at 12 08 07|Screenshot 2022-03-09 at 12 06 04|

### Video demo

https://user-images.githubusercontent.com/21178754/157430139-b0187eb0-829f-4139-bdd9-5c5b94a908b8.mov


src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
```diff
@@ -1,11 +1,17 @@
...
+import styled from "styled-components"
 import { RouterLink } from "v2/System/Router/RouterLink"
 import { SavedSearchAlertListItem_item } from "v2/__generated__/SavedSearchAlertListItem_item.graphql"
 import { EditAlertEntity } from "../types"
 
 export type SavedSearchAlertListItemVariant = "active" | "inactive"
 
+const AlertPill = styled(Pill)`
+  pointer-events: none;
+  border-color: #707070; /* black60 */
```
gkartalis, line 12: Used styled components for this one since we don't have a property in pills for changing the border color.

dimatretyak, line 12: **suggestion**: use `themeGet`
```javascript
const AlertPill = styled(Pill)`
  pointer-events: none;
  border-color: ${themeGet("colors.black60")}
`
```

damassi: @dzucconi - thoughts about adding this to Palette?

gkartalis: @dimatretyak also mentioned [this approach](https://github.com/artsy/force/pull/9526#discussion_r817853889), so if we decide not adding this to palette will go ahead with Dima's suggestion

dzucconi: Yeah: Either we add this with design guidance as a real variant or state. Or we don't do it.

damassi: That's now two examples where we're exiting the design system, though. If we're going to be changing the border color of the pill semi-regularly, we should just add a prop over there, right?

dzucconi: It should be a real variant, if design really wants this.

damassi:
```
<Pill borderColor="black60" />
```

gkartalis: I like this one 👆  and I think that makes sense for cases like this.

damassi: @gkartalis - here's an example of how we use variants, which could be applied to `Pill`: https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Banner/Banner.tsx#L64-L85

dzucconi: Honestly I'm not convinced this isn't a mistake on design's part. This just looks like the normal square variant. Can we confirm with design that this is intended as a new variant?

damassi: Oh interesting, didn't look in the Pill component and didn't realize we were already using variants there 👍

gkartalis: @dzucconi will check with them and will post an update here 🤘

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
```diff
@@ -59,6 +91,24 @@ export const SavedSearchAlertListItem: React.FC<SavedSearchAlertListItemProps> =
...
       </Flex>
+      <Spacer my={2} />
+      {labels.map(label => (
+        <AlertPill key={label.value} variant="textSquare" active mr={1} mb={1}>
+          {label.value}
+        </AlertPill>
+      ))}
+      {item.labels.length > DISPLAYED_PILLS_COUNT && (
+        <Clickable
+          aria-label="yolo"
```
dzucconi, line 102: ☝️

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
```diff
@@ -18,12 +36,26 @@ export const SavedSearchAlertListItem: React.FC<SavedSearchAlertListItemProps> =
...
+  const toggleExpandFiltersText = isExpanded
+    ? "Close all filters"
+    : "Show all filters"
 
   return (
     <Box
       key={item.internalID}
       px={[2, 4]}
-      py={4}
+      pb={30}
```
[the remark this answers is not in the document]
gkartalis: wanted to ask about that, so what I wanted to do here is to give padding top 40 and padding bottom 30, how do we do that?

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
gkartalis: woops removing now

gkartalis: typical stray "yolo" situation

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
dzucconi: 👍 

> although it is not that straightforwards that this is it.

What do you mean by this?

gkartalis: While having a quick first look in the component, didn't understand that all I had to do is to pass the `active` prop, and I realized that only after going through figma but it's okay I guess :)

dzucconi: Ah yeah; sorry. I also didn't put that together. It looks like the docs for this component could use some attention.

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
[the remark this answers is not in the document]
damassi: Also note that we have a lot of new designers starting who might not be that familiar with our design system. Devs, if they feel comfortable, should be the last guard for adhering to the design system `space` values so whenever stray values like this show up in PR review do call it out. Rounding to the closest value (within reason) is sufficient action.

anandaroop (dismissed):

dimatretyak (approved):
LGTM. Thanks for the great test cases 🥇

PR #9530, opened 2022-03-02
feat: add server error logging in middleware

ovasdi:
https://artsyproduct.atlassian.net/browse/PLATFORM-4003

This change is an action item following [#inc-141](https://artsy.app.opsgenie.com/incident/detail/0f3fd6e1-9e57-40a5-a41b-92bc54a86356). During investigations responders noticed that the observed 504 errors were _not_ recorded in papertrail, making troubleshooting more difficult.

This adds server error logging to force middleware _errorHandler_ when in production mode.

### DIFF
--- a/src/lib/middleware/errorHandler.tsx
+++ b/src/lib/middleware/errorHandler.tsx
@@ -36,6 +36,7 @@ import {
   injectGlobalStyles,
   ThemeProviderV3,
 } from "@artsy/palette"
+import createLogger from "v2/Utils/logger"
 
 const { GlobalStyles } = injectGlobalStyles()
 
@@ -67,6 +68,12 @@ export const errorHandlerMiddleware = async (
 
   if (enableLogging && err.status !== 404) {
     console.log(detail)
+  }
+
+  // Log server errors (code, stack trace) when in production mode
+  if (NODE_ENV === "production" && code >= 500) {
+    const logger = createLogger("lib/middleware/errorHandlerMiddleware")
+    logger.error(code, detail)
   }
 
   try {
--- a/src/lib/middleware/errorHandler.tsx
+++ b/src/lib/middleware/errorHandler.tsx
@@ -39,6 +39,7 @@ import {
 import createLogger from "v2/Utils/logger"
 
 const { GlobalStyles } = injectGlobalStyles()
+const logger = createLogger("lib/middleware/errorHandlerMiddleware")
 
 export const errorHandlerMiddleware = async (
   err: any,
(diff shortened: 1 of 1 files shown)


src/lib/middleware/errorHandler.tsx
```diff
@@ -69,6 +70,12 @@ export const errorHandlerMiddleware = async (
     console.log(detail)
   }
 
+  // Log server errors (code, stack trace) when in production mode
+  if (NODE_ENV === "production" && code >= 500) {
+    const logger = createLogger("lib/middleware/errorHandlerMiddleware")
```
damassi, line 75: might as well move this line up after line 41 in case we want to use it elsewhere later

src/lib/middleware/errorHandler.tsx
```diff
@@ -69,6 +70,12 @@ export const errorHandlerMiddleware = async (
     console.log(detail)
   }
 
+  // Log server errors (code, stack trace) when in production mode
+  if (NODE_ENV === "production" && code >= 500) {
```
damassi, line 74: There's a guard in [`logger.ts`](https://github.com/artsy/force/blob/main/src/v2/Utils/logger.ts#L20) that will only dispatch if in prod, so no need to check ENV here

damassi (dismissed):
Two minor comments, otherwise LGTM 👍

ovasdi:
> Two minor comments, otherwise LGTM 👍

Thanks for the quick review @damassi. I've moved the second `if` into an `else if`. Seemed more neat and it wont double log the same error if running in verbose mode.

src/lib/middleware/errorHandler.tsx
```diff
@@ -67,6 +69,9 @@ export const errorHandlerMiddleware = async (
 
   if (enableLogging && err.status !== 404) {
     console.log(detail)
```
damassi, line 71: this can be logger.log(detail)

damassi:
Sounds good! one last little note about replacing the existing console.log in favor of our new logger, but after that looks good feel free to merge on 🍏

src/lib/middleware/errorHandler.tsx
```diff
@@ -106,7 +111,7 @@ export const errorHandlerMiddleware = async (
       mount: false, // Does not mount the client-side
     })
   } catch (err) {
-    console.error(err)
+    logger.error(err)
```
ovasdi, line 114: @damassi I also switched the bottom catch to use logger. Is that OK?

mc-jones (approved):

PR #9535, opened 2022-03-02
fix: TX-66 - gracefully handle orders with unpublished artworks.

erikdstock:
This PR updates our graphql queries and adds some type safety to avoid 500 errors when trying to load an order for an artwork that has become unpublished. 

Fixes [TX-66]

TODO: 
- [x] more manual testing
- [x] refine clumsily-fixed test fixtures
- [ ] test with both published and deleted artworks (clarify if we need to do the latter)

[TX-66]: https://artsyproduct.atlassian.net/browse/TX-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
+++ b/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
@@ -34,9 +34,6 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
   const firstLineItem = get({}, () => lineItems?.edges?.[0]?.node!)
   const { artwork, artworkVersion } = firstLineItem!
 
-  if (!artwork) console.warn("artwork missing")
-  if (!artworkVersion) console.warn("artwork version missing")
-
   const { artistNames, title, image } = artworkVersion || {}
   const { date, shippingOrigin } = artwork || {}
 
--- a/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
+++ b/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
@@ -31,9 +31,14 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
   },
   ...others
 }) => {
-  const artwork = get({}, () => lineItems?.edges?.[0]?.node?.artwork)
+  const firstLineItem = get({}, () => lineItems?.edges?.[0]?.node!)
+  const { artwork, artworkVersion } = firstLineItem!
 
-  const { artistNames, title, date, shippingOrigin, image } = artwork!
+  if (!artwork) console.warn("artwork missing")
+  if (!artworkVersion) console.warn("artwork version missing")
+
+  const { artistNames, title, image } = artworkVersion || {}
+  const { date, shippingOrigin } = artwork || {}
 
   const imageURL =
     image &&
@@ -117,10 +122,12 @@ export const ArtworkSummaryItemFragmentContainer = createFragmentContainer(
                 }
               }
               artwork {
-                artistNames
-                title
                 date
                 shippingOrigin
+              }
+              artworkVersion {
+                artistNames
+                title
                 image {
                   resized_ArtworkSummaryItem: resized(width: 55) {
                     url
--- a/src/v2/Apps/Order/OrderApp.tsx
+++ b/src/v2/Apps/Order/OrderApp.tsx
@@ -6,7 +6,7 @@ import { findCurrentRoute } from "v2/System/Router/Utils/findCurrentRoute"
 import { ErrorPage } from "v2/Components/ErrorPage"
 import { MinimalNavBar } from "v2/Components/NavBar/MinimalNavBar"
 import { RouterState, withRouter } from "found"
-import { Component } from "react";
+import { Component } from "react"
 import { Meta, Title } from "react-head"
 import { graphql } from "react-relay"
 import { loadStripe } from "@stripe/stripe-js"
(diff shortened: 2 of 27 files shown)


src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -31,9 +31,14 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
...
-  const artwork = get({}, () => lineItems?.edges?.[0]?.node?.artwork)
+  const firstLineItem = get({}, () => lineItems?.edges?.[0]?.node!)
+  const { artwork, artworkVersion } = firstLineItem!
 
-  const { artistNames, title, date, shippingOrigin, image } = artwork!
+  if (!artwork) console.warn("artwork missing")
+  if (!artworkVersion) console.warn("artwork version missing")
+
+  const { artistNames, title, image } = artworkVersion || {}
+  const { date, shippingOrigin } = artwork || {}
```
rquartararo, line 41: I'm a bit confused by this, could you clarify the difference between `artwork` and `artworkVersion`?

rquartararo:
This looks good so far! Do we want to give a visible warning to the user when an artwork has been unpublished?

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -117,10 +122,12 @@ export const ArtworkSummaryItemFragmentContainer = createFragmentContainer(
...
               }
               artwork {
-                artistNames
-                title
                 date
                 shippingOrigin
+              }
+              artworkVersion {
+                artistNames
+                title
```
starsirius, line 130: 👍  switching to use artwork versions.

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
starsirius, line 41: Looks like there is information that's not available on artwork version. I think a good first step is to not 500 the entire page. 👍

erikdstock: When an artwork is updated we sometimes [create a new `ArtworkVersion` record](https://github.com/artsy/gravity/blob/main/app/models/domain/artwork.rb#L1542), which is like a snapshot of some core fields of the artwork at that time. Both are attached to the exchange order, so we can access those fields on the artworkVersion which remains visible after the artwork is unpublished.

The [actual logic where we do this](https://github.com/artsy/gravity/blob/main/app/models/domain/artwork.rb#L1361-L1365) only gets [called sometimes](https://github.com/artsy/gravity/search?q=.versioned) which makes things a little more confusing - it's not just a rails `after_save` hook or something like that.

erikdstock:
Now i'm wondering if this is a potential source of a bad experience. what if the old version is wrong and the partner updates it? their order might still hold the stale artwork version. I'd like to check if this is the ideal experience - another option would be to just say `const { title = "some fallback text", artistNames = ["fallback name"], date, etc } = artwork || {}`, or even to not use fallback text at all and settle for that being blank.

starsirius:
> Now i'm wondering if this is a potential source of a bad experience. what if the old version is wrong and the partner updates it? their order might still hold the stale artwork version.

I actually think it's good to show the exact version and snapshot that buyer purchased. It's sort of like an immutable receipt/email. Re the fallback version to display in case some data is missing: I think we can come up with our engineering design™️ and get approval from product!

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
rquartararo: Thank you this makes a lot more sense now!

erikdstock:
> > Now i'm wondering if this is a potential source of a bad experience. what if the old version is wrong and the partner updates it? their order might still hold the stale artwork version.
> 
> 
> 
> I actually think it's good to show the exact version and snapshot that buyer purchased. It's sort of like an immutable receipt/email. Re the fallback version to display in case some data is missing: I think we can come up with our engineering design™️ and get approval from product!

What if the order is not yet fulfilled, like in the offer phase the gallery edits it to clarify the item?

erikdstock:
Attached are images of the artwork as it appears when unpublished in the order page and the user's _orders_ page, which was not changed. Note that this artwork was created by the staging seeding job, which due to a quirk in the script does not have its initial artist names or title attached, causing bad output - but this would not happen in production.

If we are ok with this it is ready for merge.

image
image

starsirius:
> What if the order is not yet fulfilled, like in the offer phase the gallery edits it to clarify the item?

This is a great question! Basically what should we display (e.g. live data or a snapshot data from an artwork version) from the point a buyer starts checking out? I think trying to display live data and falling back to snapshot when live data is no longer available makes sense for some metadata (potentially we can display message like "the artwork metadata has been updated since you started the checkout.").

We do have stricter checks to make sure when a buyer submits a [BN](https://github.com/artsy/exchange/blob/479f3d258bf2a9fc21c921254acb9eda601a466c/app/services/order_service.rb#L53-L56) or [MO](https://github.com/artsy/exchange/blob/479f3d258bf2a9fc21c921254acb9eda601a466c/app/services/offer_service.rb#L40) oder, the latest artwork version is the same (i.e., partner hasn't changed some key data), to avoid e.g., a price bump without buyer's awareness.

starsirius:
Merging! 🎉

PR #9536, opened 2022-03-03
fix: SWA-263 Resources: label is missing spaces at artsy.net/consign

evaschilken:
- change syntax from {" "} to `&nbsp;` because then the space is not removed when using flexbox


AlicanAkyuz (approved):
looks good. thank you for catching and fixing this bug 🙏

PR #9537, opened 2022-03-03
feat(article): places additional ads; documents placement

dzucconi:
Continuing on [GRO-755](https://artsyproduct.atlassian.net/browse/GRO-755)

This places the remaining ads and ads some documentation regarding the `units` and placement.

Still to do are analytics.

### DIFF
--- a/src/v2/Apps/Article/Components/ArticleSectionAd.tsx
+++ b/src/v2/Apps/Article/Components/ArticleSectionAd.tsx
@@ -1,0 +1,56 @@
+import { FullBleed } from "@artsy/palette"
+import { FC } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { ArticleAd } from "./ArticleAd"
+import { ArticleSectionAd_article } from "v2/__generated__/ArticleSectionAd_article.graphql"
+
+interface ArticleSectionAdProps {
+  article: ArticleSectionAd_article
+  i: number
+}
+
+const ArticleSectionAd: FC<ArticleSectionAdProps> = ({ article, i }) => {
+  switch (article.layout) {
+    case "FEATURE": {
+      if (i === 1) {
+        return (
+          <FullBleed bg="black5" p={1}>
+            <ArticleAd unit="Desktop_Leaderboard1" size="970x250" />
+          </FullBleed>
+        )
+      }
+
+      if (i === 6) {
+        return (
+          <FullBleed bg="black5" p={1}>
+            <ArticleAd unit="Desktop_Leaderboard2" size="970x250" />
+          </FullBleed>
+        )
+      }
+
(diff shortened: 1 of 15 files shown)


src/v2/Apps/Article/Components/ArticleSectionAd.tsx
```diff
@@ -0,0 +1,56 @@
...
+    case "FEATURE": {
+      if (i === 1) {
+        return (
+          <FullBleed bg="black5" p={1}>
+            <ArticleAd unit="Desktop_Leaderboard1" size="970x250" />
+          </FullBleed>
+        )
+      }
+
+      if (i === 6) {
```
damassi, line 23: whats `i === 1`? `i === 6`? I imagine its some kind of positional flag but is there some way we can let our future selves know this?

dzucconi: I can rename to `index` — just the index of the section. I feel like `i` is the one-letter variable we can get away with 😄

src/v2/Apps/Articles/Components/NewsIndexArticles.tsx
```diff
@@ -42,10 +43,21 @@ export const NewsIndexArticles: FC<NewsIndexArticlesProps> = ({
...
+            <Fragment key={article.internalID}>
+              <ArticleBodyFragmentContainer article={article} />
+
+              {/* Insert an ad after every 6th, article; beginning with 3rd */}
+              {(i - 2) % 6 === 0 && (
+                <ArticleAd
+                  size="970x250"
+                  unit={(() => {
+                    if (i === 2) return "Desktop_Leaderboard1"
+                    if (i === 8) return "Desktop_Leaderboard2"
```
damassi, line 55: ditto with these numbers. The comment and code makes sense, and I don't know if it can be more straightforward, but variable names (might) help

src/v2/Apps/Article/Components/ArticleSectionAd.tsx
damassi: Oh i mean the 6, 8, etc

damassi: all said, this likely can't be improved, but just wanted to call it out in case

dzucconi: oh: https://github.com/artsy/force/pull/9537/files#diff-afe1a72a48bafcf27d37e846452b047bc39a82e51a59afe9d75c6f4fb84922e7

dzucconi: Yeah; maybe just leaving a comment to the README on the placement.

PR #9538, opened 2022-03-03
feat(FX-3774): Setup QA merchandising A/B test

dimatretyak:
Type: **Chore**
Jira ticket: [FX-3774]

### Acceptance Criteria
* A collector visits an artist page (e.g. `/artist/damien-hirst`)
* Force checks the artist slug against a list of targeted artist slugs
* If the artist slug is within the targeted set, check the A/B test variant.
  * If variant is`experiment`, set the default sort to `-decayed_merch`
  * If variant is `control`, leave the default sort as `-decayed_merch_uncurated`
* If the artist slug is **NOT** within the targeted set, then set the default sort to `-decayed_merch`

### Notes
Helper text was added for debug purposes

screenshot

### Demo
* Targeted artist, SplitTest variant `Experiment` - [link](https://qa-merchandising.artsy.net/artist/banksy/works-for-sale?split_test[artist_grid_manual_curation_trial_qa]=experiment)
* Targeted artist, SplitTest variant `Control` - [link](https://qa-merchandising.artsy.net/artist/banksy/works-for-sale?split_test[artist_grid_manual_curation_trial_qa]=control)
* NOT Targeted artist, SplitTest variant `Experiment` - [link](https://qa-merchandising.artsy.net/artist/lilian-martinez/works-for-sale?split_test[artist_grid_manual_curation_trial_qa]=experiment)
* NOT Targeted artist, SplitTest variant `Control` - [link](https://qa-merchandising.artsy.net/artist/lilian-martinez/works-for-sale?split_test[artist_grid_manual_curation_trial_qa]=control)

### Useful links
* [Technical Plan](https://www.notion.so/artsy/Merchandising-artist-artwork-grids-2dd791bb0a4845fe9d582c0d7fa47e86)
* [Curation AB test doc](https://www.notion.so/artsy/Curation-AB-test-b12aae656d1e48409fc7b6b29741f837)
* [Curating Artist Pages Trial](https://docs.google.com/spreadsheets/d/11g0CGH9b1kF0XPs4l3vo32bYz7y__ngK64cHyh9qBPo/edit#gid=78273406)
* [Project in Unleash](https://unleash.artsy.net/projects/default/features2/artist_grid_manual_curation_trial)

[FX-3774]: https://artsyproduct.atlassian.net/browse/FX-3774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/desktop/lib/buildServerAppContext.ts
```diff
@@ -25,6 +25,7 @@ export const buildServerAppContext = (
       contextPageOwnerType: pageType,
       contextPageOwnerSlug: pageSlug,
     },
+    featureFlags: res.locals.sd.FEATURE_FLAGS,
```
dimatretyak, line 28: I pass `featureFlags` here to be able to get the flag value from `context` in routing (for example, in `prepareVariables` like [here](https://github.com/artsy/force/blob/92c4c4580bdc432f586e05ba88dc96f7239c4bbd/src/v2/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts#L7))

src/v2/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
```diff
@@ -2,16 +2,27 @@ import { paramsToCamelCase } from "v2/Components/ArtworkFilter/Utils/urlBuilder"
 import { initialArtworkFilterState } from "v2/Components/ArtworkFilter/ArtworkFilterContext"
 import { allowedFilters } from "v2/Components/ArtworkFilter/Utils/allowedFilters"
 
-export function getWorksForSaleRouteVariables({ artistID }, props) {
+export function getWorksForSaleRouteVariables(
+  { artistID },
+  { location, context }
+) {
+  const isMerchandisingEnabled = !!context.featureFlags?.["merchandising-test"]
```
dimatretyak, line 9: If we try to use `getENV("FEATURE_FLAGS")`, we will get `undefined` during server rendering

damassi:
@dimatretyak - i think you might be behind the `main` branch, as this was merged (i think) yesterday? https://github.com/artsy/force/pull/9513 and covers your approach

dimatretyak:
> @dimatretyak - i think you might be behind the `main` branch, as this was merged (i think) yesterday? #9513 and covers your approach. Or did you find a but in this PR that is then fixed here?

Not quite. Yesterday's changes add support only for system context

damassi:
I see! I should have read closer. Would you mind removing [this code](https://github.com/artsy/force/blob/main/src/v2/System/Router/Boot.tsx#L66) and [this code](https://github.com/artsy/force/blob/main/src/v2/System/Router/Boot.tsx#L71) then, if possible? Because now with your changes everything should now be coming in via [`context`](https://github.com/artsy/force/blob/main/src/v2/System/Router/Boot.tsx#L63) which is spread in [below](https://github.com/artsy/force/blob/main/src/v2/System/Router/Boot.tsx#L73).

src/v2/Apps/Artist/Routes/WorksForSale/Utils/__tests__/merchandisingTrial.jest.ts
```diff
@@ -0,0 +1,55 @@
...
+jest.mock("v2/System/useSystemContext")
+
+describe("useCheckManualGridCuration", () => {
+  const mockUseSystemContext = useSystemContext as jest.Mock
+
+  describe("when the feature flag is enabled", () => {
+    beforeEach(() => {
+      mockUseSystemContext.mockImplementation(() => ({
+        featureFlags: {
+          [MANUAL_GRID_CURATION_FEATURE_FLAG_KEY]: true,
```
mc-jones, line 16: Doesn't affect your implementation, as we will always return out a boolean for the `useFeatureFlag` hook, but want to flag that with the implementation of `useFeatureVariant` the shape of the`featureFlags` value in `systemContext` was changed, which you can see [here](https://github.com/artsy/force/blob/main/src%2Fv2%2FSystem%2F__tests__%2FuseFeatureFlag.jest.tsx#L11).

mc-jones: Actually, take this back. See comment below.

src/v2/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
```diff
@@ -1,17 +1,32 @@
...
 
-export function getWorksForSaleRouteVariables({ artistID }, props) {
+export function getWorksForSaleRouteVariables(
+  { artistID },
+  { location, context }
+) {
+  const featureFlags = context.featureFlags ?? {}
+  const isManualGridCurationEnabled = !!featureFlags[
+    MANUAL_GRID_CURATION_FEATURE_FLAG_KEY
+  ]
```
mc-jones, line 16: The new shape of `featureFlags` mentioned above will break this. But would also be helpful to know if there is a reason we aren't using the `useFeatureFlags` hook here, rather than checking it from the `context` which I am assuming is the `systemContext` getting passed down? Understanding where the `useFeatureFlags` hook isn't enough for developers use cases will help with its development.

damassi: @mc-jones - I can quickly answer this! So react hooks can only be used in react components, so they wouldn't work here. Part of the work in [this PR](https://github.com/artsy/force/pull/9545) was ensuring that the flags get passed into our context setup further down, so that when we instantiate the framework router we can pass said context in [here](https://github.com/artsy/force/blob/main/src/v2/System/Router/buildClientApp.tsx#L93) both in `buildClientApp` and `buildServerApp`, so that this data is accessible in all parts of the app, and most notably at the router level, which this file directly touches. 

So the flow is:
- grab feature flags from express
- create context
- inject context into router
- inject context into boot

And then at the router level, we can pluck context values out both in `prepareVariables` [as seen here](https://github.com/artsy/force/blob/main/src/v2/Apps/Auction2/auction2Routes.tsx#L74-L87). Or alternatively in react class components (which can't use hooks) we can access via `this.props.match.context.featureFlags` (which is fortunately rare since most of our app has been migrated to modern react at this point). 

Hope this makes sense!

src/v2/Apps/Artist/Routes/WorksForSale/Utils/__tests__/merchandisingTrial.jest.ts
dimatretyak: Updated

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -183,6 +183,8 @@ export interface ArtworkFilterContextProps {
   /** The current artwork filter state (which determines the network request and the url querystring) */
   filters?: ArtworkFiltersState
 
+  defaultFilters?: ArtworkFiltersState
```
dimatretyak, line 186: This is necessary in order to correctly restore the `sort` value depending on the A/B test when the "Clear All" button was pressed on mobile device

### Before
https://user-images.githubusercontent.com/3513494/157660295-3613af21-0a47-41c5-a2ea-1d9324ec3bf4.mp4

### After
https://user-images.githubusercontent.com/3513494/157660376-9bb51f2b-1b63-4b3b-838c-83a0975f96ca.mp4

damassi, line 186: 🙏 for the PR comment, but could we add a comment above line 186 that describes what this does? 

Also wondering if after the A/B test is done we can remove this prop?

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -585,7 +591,8 @@ const artworkFilterReducer = (
 }
 
 export const getSelectedFiltersCounts = (
```
damassi, line 593: Can we add a comment that describes `defaultFilters` works and maybe a `TODO: remove once A/B test is over` if this applies

src/v2/System/Analytics/trackingMiddleware.ts
```diff
@@ -109,6 +116,19 @@ export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
...
+            routes.some(route => {
+              const matcher = match(route, { decode: decodeURIComponent })
+              const foundMatch = !!matcher(pathname)
+
+              if (foundMatch) {
+                trackExperimentViewed(abTest)
+                return true
+              }
+            })
+          })
```
damassi, line 131: 👍

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
dimatretyak: > if after the A/B test is done we can remove this prop

I think yes

damassi: Since this prop adds a somewhat large amount of complexity to the artwork filter we should comment around the key areas that can be removed after the test is over

damassi (approved):
One comment about adding `TODO`s around the areas that can be undone once the test is completed, but otherwise A+ work on this @dimatretyak! Looks great.

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -258,6 +260,7 @@ export type SharedArtworkFilterContextProps = Pick<
   | "sortOptions"
   | "onFilterClick"
   | "ZeroState"
+  | "defaultFilters"
```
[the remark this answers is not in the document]
dimatretyak: Fixed

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
dimatretyak: Fixed

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
dimatretyak: Fixed

src/v2/Components/ArtworkFilter/ArtworkFilterContext.tsx
```diff
@@ -183,6 +183,8 @@ export interface ArtworkFilterContextProps {
   /** The current artwork filter state (which determines the network request and the url querystring) */
   filters?: ArtworkFiltersState
 
+  // TODO: remove once A/B test is over
```
damassi, line 186: Can we add the name of the a/b test

damassi (dismissed):

src/v2/System/Analytics/__tests__/trackingMiddleware.jest.ts
```diff
@@ -142,4 +146,56 @@ describe("trackingMiddleware", () => {
       })
     })
   })
+
+  describe("triggering AB test experiment viewed events", () => {
```
mzikherman, line 150: nice specs!

mzikherman (dismissed):
LGTM! Very nice.

src/v2/Utils/merchandisingTrial.ts
```diff
@@ -0,0 +1,71 @@
...
+]
+export const TARGETED_ARTIST_ROUTES = TARGETED_ARTIST_SLUGS.map(slug => {
+  return `/artist/${slug}/works-for-sale`
+})
+
+export const getDefaultSortValueByVariant = (artistSlug: string) => {
+  const variant = getENV("ARTIST_GRID_MANUAL_CURATION_TRIAL_QA")
+  const isTargetedArtistSlug = TARGETED_ARTIST_SLUGS.includes(artistSlug)
+
+  if (isTargetedArtistSlug && variant === "control") {
```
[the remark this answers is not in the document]
anandaroop: Ah good q! This does look fishy, but in this experiment:

- the control group = let default sort do its thing, **except for** pinning artworks to featured slots, so they see an _uncurated_ grid

- the experiment group = default sort with the usual pinning of artworks works, so they see a curated grid.

The experiment here really is: what is the measurable effect of doing of lot of manual curation?

@dimatretyak feel free to chime in if I've misstated anything.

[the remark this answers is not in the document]
dblandin: Yep, it's a bit unusual but Roop's breakdown is accurate.

We want to test if artist artworks grids which are manually curated perform better than our default sort algorithm. Since we don't want to interfere with artist pages outside of the scope of this A/B test, we decided to define a new sort option to use for the control, and let the experiment trigger normal behavior including currently enabled curation, knowing that we'll also be seeding those `Artwork#featured_slot` values separately.

The A/B test is web only but these curations which also trickle down to the mobile apps during the test.

dblandin (dismissed):
Nice work! 💯

src/v2/Utils/merchandisingTrial.ts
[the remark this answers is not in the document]
dimatretyak: @anandaroop you have described everything correctly. Thank you for it!

src/v2/System/Analytics/__tests__/trackingMiddleware.jest.ts
dimatretyak: Thanks!

dimatretyak:
### Some notes
* Merch Artwork Grid Test QA will be performed on [review app](https://qa-merchandising.artsy.net/)
* If the QA session is successful, then
  * We remove the `_qa` suffix from the experiment name
  * This PR will be merged and released on Monday

src/lib/middleware/redisPageCache.ts
```diff
@@ -55,6 +56,12 @@ export function redisPageCacheMiddleware(
     return next()
   }
 
+  // TODO: remove once "artist_grid_manual_curation_trial" A/B test is over
+  // Skip targeted artist routes
+  if (TARGETED_ARTIST_ROUTES.includes(req._parsedUrl.pathname)) {
+    return next()
+  }
+
```
dimatretyak, line 64: There was a problem with A/B testing (see [this thread](https://artsy.slack.com/archives/C9SATFLUU/p1647011695022429) for more details) due to page caching. For this reason, caching has been disabled for the target artist pages.

@damassi @mzikherman  when you have a free minute, please take a look if everything is OK

src/lib/middleware/__tests__/redisPageCache.jest.ts
```diff
@@ -72,6 +78,10 @@ describe("pageCacheMiddleware", () => {
       method: "GET",
       path: "/artist-series/test-artist",
       url: "https://artsy.net/artist-series/test-artist",
+      // TODO: remove once "artist_grid_manual_curation_trial" A/B test is over
+      _parsedUrl: {
+        pathname: "/artist-series/test-artist",
+      },
```
dblandin, line 84: **question**: Is the artist-series page affected by this test (and cache) as well?

src/lib/middleware/redisPageCache.ts
```diff
@@ -55,6 +56,12 @@ export function redisPageCacheMiddleware(
     return next()
   }
 
+  // TODO: remove once "artist_grid_manual_curation_trial" A/B test is over
+  // Skip targeted artist routes
+  if (TARGETED_ARTIST_ROUTES.includes(req._parsedUrl.pathname)) {
+    return next()
+  }
```
dblandin, line 63: **praise**: I just QA'd the review app again and I can no longer reproduce the mismatch so this seems to have worked! 🙌🏼

dblandin (approved):
Awesome sleuthing on that cache issue Dima! 🎉 🔎

src/lib/middleware/__tests__/redisPageCache.jest.ts
dimatretyak: Yep (If I understood your question correctly)

FYI These changes are necessary here in order to avoid such an error for [these lines](https://github.com/artsy/force/blob/2a803c67a5cd9547a3a07c0c548d3ab9755c90b7/src/lib/middleware/redisPageCache.ts#L61)(cannot get `pathname` from `req._parsedUrl`)
image

PR #9541, opened 2022-03-03
ci(notifications): Update slack message to be more clear

damassi:
This updates our #practice-web slack notification to be a bit more clear, and provide a path forward to fix.


ovasdi (approved):
LGTM. Thank you for cleaning up the message and adding clear steps to take.

PR #9545, opened 2022-03-04
refactor: pass feature flags to client and server contexts

dimatretyak:
Related PR: https://github.com/artsy/force/pull/9513

As mentioned [here](https://github.com/artsy/force/pull/9538#issuecomment-1058368742), this PR contains small changes for [this PR](https://github.com/artsy/force/pull/9513) (we pass feature flags not only to `SystemContext`, but also to `ClientContext` and `ServerContext`)


src/v2/System/useFeatureFlag.tsx
```diff
@@ -1,7 +1,5 @@
 import { useSystemContext } from "v2/System"
 
-export type EnabledFeatureFlags = Record<string, boolean>
```
dimatretyak, line 3: it may be worth moving this type to `v2/Typings/common.ts` (to avoid cyclical imports) and use it [here](https://github.com/artsy/force/blob/5fa3a5cac134c666611ec135f138d18ff76f0cdd/src/desktop/lib/buildClientAppContext.ts#L10) and [here](https://github.com/artsy/force/blob/5fa3a5cac134c666611ec135f138d18ff76f0cdd/src/v2/System/SystemContext.tsx#L82)?

damassi (approved):
Really nice simplification @dimatretyak 🙏 

cc @mc-jones

PR #9546, opened 2022-03-04
[DRAFT] cannot set headers after they are sent to the client

ovasdi:
## Background

As a followup to #[inc-141](https://artsy.app.opsgenie.com/incident/detail/0f3fd6e1-9e57-40a5-a41b-92bc54a86356) (_Force 504 errors and unleash request spike_) the velocity team is looking to update the [sentryFilters](https://github.com/artsy/force/blob/main/src/lib/analytics/sentryFilters.ts#L18) removing the _[response timeout](https://artsyproduct.atlassian.net/browse/PLATFORM-4004)_ from the list of ignored errors (which currently does not end up in sentry). As of today 10AM ET, the sentry statistics over the last 24h are:

Screen Shot 2022-03-04 at 10 07 56 AM

The decision to do this was a result of not being able to trace the 504s seen during the mentioned incident. One of the concerns is that removing this type of error from filters may contribute to us hitting sentry rate limits. As I started to look at sentry (and papertrail) I noticed that the following error is occurring many times throughout the day: _Cannot set headers after they are sent to the client_.

Screen Shot 2022-03-04 at 9 58 24 AM

Going by sentry issue title, this affects multiple endpoints:

Screen Shot 2022-03-04 at 10 29 05 AM

According to sentry, the error was first seen on Jan 31st.

### Problem

_During the request lifecycle (server side), the application is doing some redirecting and there seems to be logic missing that tells the server to stop processing a request once the necessary redirects have been triggered._

I decided to look into this specific error (and other errors) before updating the sentry filters and introducing another type of error being sent to sentry.

As far as the specific exception is concerned, _Cannot set headers after they are sent to the client_, the frequency in sentry is:

```
~90/h (vs 2nd most seen error ~30)
~2k/day (vs 2nd most seen error ~500)
~65k/mo (vs 2nd most seen error 16k)
```

During our weekly mob @artsy/velocity-engineers spent some time discussing this issue. We were able to reproduce the issue and introduce a _hacky_ fix, mostly for the sake of demonstration to start a discussion about whether a more appropriate fix should be introduced.

### Demonstration

Running force (locally) in production mode and then signing in results in the following:

```
GET /auth-redirect?redirectTo=https%3A%2F%2Fstagingapi.artsy.net%2Fusers%2Fsign_in%3Fredirect_uri%3Dhttp%253A%252F%252Flocalhost%253A5000%252F%26trust_token... 302 1388.480ms 127.0.0.1 "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36"
lib/middleware/errorHandlerMiddleware | 500 Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:561:11)
    at ServerResponse.header (/Users/ozzievasdi/code/force/node_modules/express/lib/response.js:767:10)
    at ServerResponse.send (/Users/ozzievasdi/code/force/node_modules/express/lib/response.js:170:12)
    at done (/Users/ozzievasdi/code/force/node_modules/express/lib/response.js:1004:10)
    at tryHandleCache (/Users/ozzievasdi/code/force/node_modules/ejs/lib/ejs.js:278:5)
    at View.exports.renderFile [as engine] (/Users/ozzievasdi/code/force/node_modules/ejs/lib/ejs.js:489:10)
    at View.render (/Users/ozzievasdi/code/force/node_modules/express/lib/view.js:135:8)
    at tryRender (/Users/ozzievasdi/code/force/node_modules/express/lib/application.js:640:10)
    at Function.render (/Users/ozzievasdi/code/force/node_modules/express/lib/application.js:592:3)
    at ServerResponse.render (/Users/ozzievasdi/code/force/node_modules/express/lib/response.js:1008:7)
    at n.renderServerApp (/Users/ozzievasdi/code/force/server.dist.js:1:574465)
    at /Users/ozzievasdi/code/force/server.dist.js:1:3114205
    at tryCatch (/Users/ozzievasdi/code/force/node_modules/regenerator-runtime/runtime.js:45:40)
    at Generator.invoke [as _invoke] (/Users/ozzievasdi/code/force/node_modules/regenerator-runtime/runtime.js:274:22)
    at Generator.prototype.<computed> [as next] (/Users/ozzievasdi/code/force/node_modules/regenerator-runtime/runtime.js:97:21)
    at asyncGeneratorStep (/Users/ozzievasdi/code/force/node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
```

This is one path, that was obvious from looking at papertrail, to trigger this error. There are likely others. The committed change prevents this one error from being thrown but as mentioned previously, there are likely other redirect related conditions that would trigger it.

### Open Questions

- Is it worth pursuing a more appropriate fix for this issue?
- What would be an appropriate fix for something like this?
- If the fix is deemed worth it, who is in the best position to tackle the fix?

> I tried to capture the problem as best as I could. Hope this is enough to start a meaningful discussion.

### DIFF
--- a/src/v2/Apps/Artist/artistRoutes.tsx
+++ b/src/v2/Apps/Artist/artistRoutes.tsx
@@ -8,6 +8,7 @@ import { paramsToCamelCase } from "v2/Components/ArtworkFilter/Utils/urlBuilder"
 import { initialAuctionResultsFilterState } from "./Routes/AuctionResults/AuctionResultsFilterContext"
 import { allowedAuctionResultFilters } from "./Utils/allowedAuctionResultFilters"
 import { enableArtistPageCTA } from "./Server/enableArtistPageCTA"
+import { redirect } from "v2/System/Server/redirectHelper"
 
 const ArtistApp = loadable(
   () => import(/* webpackChunkName: "artistBundle" */ "./ArtistApp"),
@@ -87,7 +88,12 @@ export const artistRoutes: AppRouteConfig[] = [
     path: "/artist/:artistID",
     ignoreScrollBehaviorBetweenChildren: true,
     getComponent: () => ArtistApp,
-    onServerSideRender: enableArtistPageCTA,
+    onServerSideRender: props => {
+      redirect({
+        onRedirect: () => enableArtistPageCTA(props),
+        ...props,
+      })
+    },
     onClientSideRender: () => {
       ArtistApp.preload()
       OverviewRoute.preload()
@@ -107,7 +113,12 @@ export const artistRoutes: AppRouteConfig[] = [
       {
         path: "/",
         getComponent: () => OverviewRoute,
-        onServerSideRender: enableArtistPageCTA,
+        onServerSideRender: props => {
+          redirect({
+            onRedirect: () => enableArtistPageCTA(props),
+            ...props,
+          })
+        },
         onClientSideRender: () => {
           OverviewRoute.preload()
         },
--- a/src/v2/Apps/Authentication/authenticationRoutes.tsx
+++ b/src/v2/Apps/Authentication/authenticationRoutes.tsx
@@ -152,6 +152,7 @@ export const authenticationRoutes: AppRouteConfig[] = [
     path: "/auth-redirect",
     onServerSideRender: props => {
       redirectPostAuth(props)
+      props.res.locals.hasRedirected = true
     },
   },
 ]
--- a/src/v2/Apps/Authentication/authenticationRoutes.tsx
+++ b/src/v2/Apps/Authentication/authenticationRoutes.tsx
@@ -8,6 +8,7 @@ import { redirectIfLoggedIn } from "./Server/redirectIfLoggedIn"
 import { setCookies } from "./Utils/helpers"
 import { redirectPostAuth } from "./Server/redirectPostAuth"
 import { stringify } from "qs"
+import { redirect } from "v2/System/Server/redirectHelper"
 
 const ForgotPasswordRoute = loadable(
   () =>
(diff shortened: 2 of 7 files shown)


damassi:
I have indeed seen this error from `onServerSideRender` from time-to-time but so far it's been harmless compared to the benefit. That said its been on the list. I'd suggest removing the `responseTimeout` from sentryFilters and adding `Cannot set headers after they are sent to the client` instead, for the time being.

damassi:
Here's a quick sketch:  

```tsx
interface RedirectProps {
  path?: string
  statusCode? string 
  onRedirect?: () => void 
  req: ArtsyReq
  res: ArtsyRes 
}

export function redirect({ path, statusCode, onRedirect, req, res }: RedirectProps) {
  res.locals.hasRedirected = true 
  
  if (onRedirect) {
    onRedirect() 
  } else {
    res.redirectTo(path, statusCode) 
  }
}
```

Which can then be used like so:

```tsx
import { redirect } from 'v2/System/Server'

onServerSideRender: props => {
  redirect({
    onRedirect: () => redirectPostAuth(props)
  }
},
```

Or if we don't have a function that has a lot of its own behavior, `redirect` can be used more simply like so:

```tsx
onServerSideRender: props => {
  redirect({ to: "/foo", status: 304 })
}
```

ovasdi:
> I have indeed seen this error from `onServerSideRender` from time-to-time but so far it's been harmless compared to the benefit. That said it's been on the list to fix, and the implementation around `onServerSideRender` is def a first pass.
> 
> For now, I'd suggest removing the `responseTimeout` from sentryFilters and adding `Cannot set headers after they are sent to the client` instead, for the time being, or creating a generic redirect function that will set `props.res.locals.hasRedirected = true` internally, so that we don't have to litter them throughout the code, and instead update each redirect that we're using in `onServerSideRender` to use that redirect function instead.
> 
> The addition to `v2/server` seems fine!

@damassi thank you for the explanation and for providing suggestions and path forward. Seeing as the issue is on your radar I am happy to close this up and let the experts handle solving this.

Since you took the time to provide a nice example, I wanted to give it a try tho and see what the test suite has to say about it.

ovasdi:
Spent some time looking into this and am not able resolve the issues. As this isn't strictly necessary and y'all are aware, I shall close this.

PR #9547, opened 2022-03-04
feat: add useVariant to Unleash Implementation

mc-jones:
This PR adds the ability to use [variants](https://docs.getunleash.io/advanced/toggle_variants) Unleash implementation, which is necessary for A/B testing.

### Considerations
This approach aims to mimic the approach used by the [unleash-react-proxy](https://docs.getunleash.io/sdks/proxy-react):

```tsx
import { useVariant } from "v2/System/useFeatureFlag";

const TestComponent = () => {
  const variant = useVariant('travel.landing');

  if (variant.enabled && variant.name === "SomeComponent") {
    return <SomeComponent />
  } else if (variant.enabled && variant.name === "AnotherComponent") {
    return <AnotherComponent />
  }
  return <DefaultComponent />
};
```

### Related PRs
- #9513
- #9498
- #9393

### Follow-Ups
- Add analytics support (i.e. call 'Experiment Viewed') on initial  view
- Add support for cached pages (i.e. unauthenticated pages)

### DIFF
--- a/src/typings/sharify.d.ts
+++ b/src/typings/sharify.d.ts
@@ -1,3 +1,5 @@
+import { FeatureFlags } from 'v2/System/useFeatureFlag'
+
 declare module "sharify" {
   function sharify(): void
   export = sharify
@@ -35,7 +37,7 @@ declare module "sharify" {
       readonly ENABLE_YOUR_ALERTS_PAGE: boolean
       readonly FACEBOOK_APP_NAMESPACE: string
       readonly FACEBOOK_ID: string
-      readonly FEATURE_FLAGS: Record<string, boolean>
+      readonly FEATURE_FLAGS: FeatureFlags
       readonly GEMINI_CLOUDFRONT_URL: string
       readonly GENOME_URL: string
       readonly GOOGLE_ADWORDS_ID: string
--- a/src/typings/sharify.d.ts
+++ b/src/typings/sharify.d.ts
@@ -1,5 +1,3 @@
-import { FeatureFlags } from 'v2/System/useFeatureFlag'
-
 declare module "sharify" {
   function sharify(): void
   export = sharify
@@ -37,7 +35,7 @@ declare module "sharify" {
       readonly ENABLE_YOUR_ALERTS_PAGE: boolean
       readonly FACEBOOK_APP_NAMESPACE: string
       readonly FACEBOOK_ID: string
-      readonly FEATURE_FLAGS: FeatureFlags
+      readonly FEATURE_FLAGS: any
       readonly GEMINI_CLOUDFRONT_URL: string
       readonly GENOME_URL: string
       readonly GOOGLE_ADWORDS_ID: string
--- a/src/v2/System/useFeatureFlag.tsx
+++ b/src/v2/System/useFeatureFlag.tsx
@@ -27,7 +27,7 @@ export function useFeatureVariant(featureName: string): Variant | null {
   const { featureFlags } = useSystemContext()
   const variant = featureFlags?.[featureName]?.variant
 
-  if (variant === undefined) {
+  if (!variant) {
     console.error(
       "[Force] Error: cannot find variant on featureFlags: ",
       featureFlags
--- a/src/v2/System/useFeatureFlag.tsx
+++ b/src/v2/System/useFeatureFlag.tsx
@@ -1,7 +1,9 @@
 import { useSystemContext } from "v2/System"
+import { Variant } from "unleash-client"
 
 export function useFeatureFlag(flagName: string): boolean | null {
   const { featureFlags } = useSystemContext()
+
   if (featureFlags === undefined) {
     console.error("[Force] Error: featureFlags is undefined in SystemContext")
     return null
@@ -15,4 +17,31 @@ export function useFeatureFlag(flagName: string): boolean | null {
   return (
     Object.keys(featureFlags).indexOf(flagName) !== -1 && featureFlags[flagName]
   )
+}
+
+export function useFeatureVariant(featureName: string): Variant | null {
+  const { featureVariants } = useSystemContext()
+  const errorObject = {
+    enabled: false,
+    name: "false",
+  }
+
+  if (!featureVariants) {
+    console.error("[Force] Error: no argument passed into useFeatureVariant")
+    return errorObject
+  }
+
+  if (!featureName) {
+    console.error("[Force] Error: no argument passed into useFeatureVariant")
+    return errorObject
+  }
+
+  if (!featureVariants[featureName]) {
+    console.error(
+      "[Force] Error: the variant can't be found on featureVariants"
+    )
+    return errorObject
+  }
+
+  return featureVariants[featureName]
 }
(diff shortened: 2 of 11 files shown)


src/v2/System/useFeatureFlag.tsx
```diff
@@ -13,6 +22,34 @@ export function useFeatureFlag(flagName: string): boolean | null {
...
+
+  if (!featureName) {
+    console.error("[Force] Error: no argument passed into useFeatureVariant")
+    return errorObject
+  }
+
+  if (!featureFlags[featureName]) {
+    console.error("[Force] Error: the variant can't be found on featureFlags")
+    return errorObject
+  }
```
damassi, line 52: Similar to feedback on a previous PR, each of these error cases could be a bit of overkill:

- `featureFlags` existy check not necessary as the default can be an empty object within system context. `featureFlags` will always exist
- `featureName` existy will 100% of the time be caught by TypeScript
- `!featureFlags[featureName]` is a valid check

Really all we need (both here and in `useFeatureFlag` is  

```tsx
const useVariant(featureName:string) => {
  const { featureFlags } = useSystemContext()
  const variant = featureFlags[featureName]?.variant
  
  if (!variant) {
    console.error("[Force]: Error: the variant can't be found on featureFlags", featurefFlags) 
    return null 
  }
  
  return variant
}
```

We should avoid stubbing out mock flag responses for error states as the consumer of the flag should be the one checking for `null`, and then handling it there.

damassi: (Referring to bullet point #1, [this line](https://github.com/artsy/force/blob/main/src/desktop/lib/buildServerAppContext.ts#L28) (and corresponding line in buildClientAppContext) can be:

```tsx
featureFlags: res.locals.sd.FEATURE_FLAGS ?? {}
```

or even better, and more lower down so we don't have to update these context files,
```tsx
res.locals.sd.FEATURE_FLAGS = <some logic to set> ?? {} 
```

src/v2/System/useFeatureFlag.tsx
```diff
@@ -1,18 +1,39 @@
...
+  variant: Variant
+}
+
+export function useFeatureFlag(featureName: string): boolean | null {
   const { featureFlags } = useSystemContext()
-  if (featureFlags === undefined) {
-    console.error("[Force] Error: featureFlags is undefined in SystemContext")
+  const flagEnabled = featureFlags?.[featureName]?.flagEnabled
+
+  if (flagEnabled === undefined) {
```
damassi, line 15: `!flagEnabled` will check for `false, undefined, null, 0`

src/v2/System/useFeatureFlag.tsx
```diff
@@ -1,18 +1,39 @@
...
-  if (!flagName) {
-    console.error("[Force] Error: no argument passed into useFeatureFlag")
+  return flagEnabled
+}
+
+export function useFeatureVariant(featureName: string): Variant | null {
+  const { featureFlags } = useSystemContext()
+  const variant = featureFlags?.[featureName]?.variant
+
+  if (variant === undefined) {
```
damassi, line 30: ditto - here -- `!variant` is a wider match and more commonly seen around code

damassi (approved):
Looks good with the latest changes. One last suggestion around `!variant` but otherwise 🚀

src/v2/System/useFeatureFlag.tsx
mc-jones: Correct. And we want to return out false, when `featureFlags?.[featureName]?.flagEnabled = false`.

src/v2/System/useFeatureFlag.tsx
mc-jones: >or even better, and more lower down so we don't have to update these context files

This should be handled here: https://github.com/artsy/force/blob/main/src/lib/middleware/featureFlagMiddleware.ts#L32

damassi: oh sweet 🙏

src/v2/System/useFeatureFlag.tsx
damassi: Cool 👍  to be clear, just meaning if we were to do `flagEnabled === undefined` we might miss some case, and `!thing` is a quick way to match a variety of falsy cases.

mc-jones:
@damassi - Quick Q, do you know why this might be failing on [these type-checks](https://app.circleci.com/pipelines/github/artsy/force/27258/workflows/87f65902-4fd1-4a8d-bbbf-5082942594f0/jobs/212013) for files I didn't touch?

src/v2/System/useFeatureFlag.tsx
mc-jones: For sure! My thinking was the safe navigational operator will catch and return out `undefined` for anything nullish, allowing us to rely on `undefined` for the subset of falsy cases we're concerned with catching.

src/v2/System/useFeatureFlag.tsx
mc-jones: Ah, for sure. Makes sense here. I group change the logic where I didn't need to.

damassi:
> Quick Q, do you know why this might be failing on [these type-checks](https://app.circleci.com/pipelines/github/artsy/force/27258/workflows/87f65902-4fd1-4a8d-bbbf-5082942594f0/jobs/212013) for files I didn't touch?

Interesting! i'm not too sure, i can pull down branch real quick tho.

src/typings/sharify.d.ts
```diff
@@ -35,7 +37,7 @@ declare module "sharify" {
       readonly ENABLE_YOUR_ALERTS_PAGE: boolean
       readonly FACEBOOK_APP_NAMESPACE: string
       readonly FACEBOOK_ID: string
-      readonly FEATURE_FLAGS: Record<string, boolean>
+      readonly FEATURE_FLAGS: FeatureFlags
```
damassi, line 40: @mc-jones - It looks like this is the culprit -- env vars can only be core types (string, bool, number, etc) and its not liking that record with complex types embedded in, leading to cascading failures. 

If we can get away with `Record<string, boolean>` and live with a bit of duplication that should be fine.

damassi (approved):

src/typings/sharify.d.ts
mc-jones: Interestingly, the issue ultimately was the import statement, which I never would have guessed given the error message - so thanks for narrowing it down! This seems related to how TS handles [namespaces](https://stackoverflow.com/questions/39040108/import-class-in-definition-file-d-ts); moving the import within the namespace declaration works but doesn't seem like a great solution. I don't love having this typed as an `any` but this seems consistent with how we handle some of the other more complex data and allows us to move this forward. I will look into improving this in subsequent iterations.

One option would be similar to what @dimatretyak mentioned [here](https://github.com/artsy/force/pull/9545#discussion_r819496567) and move the interface to a shared typing file.

damassi: > One option would be similar to what @dimatretyak mentioned https://github.com/artsy/force/pull/9545#discussion_r819496567 and move the interface to a shared typing file.

It's probably a bit overkill IMO, since once we have the data and it's in the `SystemContext`, we have the types. Reading things at the sharify level outside of the react hook will likely be pretty rare (I _think_).

PR #9560, opened 2022-03-07
fix(colors): revert recent changes to artwork color filtering

anandaroop:
In order to un-break Eigen's color filtering we'd like to revert all artworks to index the old color palette _only_, and thus we'd need to  also roll back Force's color filtering UI, i.e. revert https://github.com/artsy/force/pull/9497

(This should get merged and deployed before https://github.com/artsy/gravity/pull/15099)

See the following Slack threads:

- Diagnosis: https://artsy.slack.com/archives/C9SATFLUU/p1646667810727089 
- Fix: https://artsy.slack.com/archives/C9SATFLUU/p1646683881730749


anandaroop:
Based on the discussion [here](https://artsy.slack.com/archives/C9SATFLUU/p1646749668383349) I'm going to proceed with merging and deploying this.

PR #9563, opened 2022-03-07
[GRO-834] add Trove Rail to the Artsy homepage

TMcMeans:
This PR solves: [GRO-834]

To get more artwork on the homepage and available for users to check out!

### logged out 
Screen Shot 2022-03-08 at 12 43 57 PM

### logged in 
Screen Shot 2022-03-08 at 12 44 33 PM

Note: We're gonna want to remove Trove as an item on the carousel for `logged in` / `logged out` states which is something an admin does. 

Gotta coordinate with Jenna before pushing this live. 

[GRO-834]: https://artsyproduct.atlassian.net/browse/GRO-834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
@@ -22,13 +22,15 @@ export const HomeTroveArtworksRail: React.FC<HomeTroveArtworksRailProps> = ({
   viewer,
 }) => {
   const artworks = extractNodes(viewer.saleArtworksConnection)
+
   if (artworks.length === 0) {
     return null
   }
+
   return (
     <Rail
       title="Trove"
-      subTitle="A weekly curated selection of the best works on Artsy by emerging and sought after Artists."
+      subTitle="A weekly curated selection of the best works on Artsy by emerging and sought after artists."
       getItems={() => {
         return artworks.map(artwork => (
           <ShelfArtworkFragmentContainer
@@ -98,7 +100,7 @@ const PLACEHOLDER = (
   <Skeleton>
     <Rail
       title="Trove"
-      subTitle="A weekly curated selection of the best works on Artsy by emerging and sought after Artists."
+      subTitle="A weekly curated selection of the best works on Artsy by emerging and sought after artists."
       getItems={() => {
         return [...new Array(8)].map((_, i) => {
           return (
--- a/src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
+++ b/src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
@@ -1,0 +1,118 @@
+import { ContextModule } from "@artsy/cohesion"
+import {
+  Box,
+  Skeleton,
+  SkeletonBox,
+  SkeletonText,
+  Spacer,
+} from "@artsy/palette"
+import { createFragmentContainer, graphql } from "react-relay"
+import { ShelfArtworkFragmentContainer } from "v2/Components/Artwork/ShelfArtwork"
+import { Rail } from "v2/Components/Rail/Rail"
+import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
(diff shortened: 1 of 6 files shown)


TMcMeans:
My graphql query is supposed to pull the first 12 artworks but I only see a few, should I switch `isClosed` (for bidding I think) to false??

src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
```diff
@@ -0,0 +1,118 @@
...
+  viewer: HomeTroveArtworksRail_viewer
+}
+
+export const HomeTroveArtworksRail: React.FC<HomeTroveArtworksRailProps> = ({
+  viewer,
+}) => {
+  const artworks = extractNodes(viewer.saleArtworksConnection)
+  if (artworks.length === 0) {
+    return null
+  }
```
damassi, line 27: this is so #superminor as to ignore, but still a tip 😄 

As you start writing more and more code, a good code writing habit is to add some linebreaks for readability. Like... think of each statement as a paragraph of sorts. You've got your variables declared on top (new paragraph), then you're checking a condition (new paragraph) then you're returning some JSX:

```tsx 
const artworks = extractNodes(viewer.saleArtworksConnection)

if (artworks.length === 0) {
  return null
}

return (
  ... 
)
```

When everything is compressed together it can often feel like a run-on sentence, or an extra long paragraph -- hard to understand the gist of at a glance.

The-Beez-Kneez: Ditto on this advice.

src/v2/Apps/Home/__tests__/HomeTroveArtworksRail.jest.tsx
```diff
@@ -0,0 +1,47 @@
...
+      }),
+    })
+
+    expect(wrapper.text()).toContain("Trove")
+    expect(wrapper.text()).toContain(
+      "A weekly curated selection of the best works on Artsy"
+    )
+    expect(wrapper.html()).toContain("test-href")
+  })
+})
```
The-Beez-Kneez, line 47: Thank you for adding a spec file :)

jonallured: Plus one 🏆

src/v2/Apps/Home/HomeApp.tsx
```diff
@@ -63,6 +64,8 @@ export const HomeApp: React.FC<HomeAppProps> = ({
         <HomeFeaturedGalleriesRailQueryRenderer />
 
         <HomeTrendingArtistsRailQueryRenderer />
+
+        <HomeTroveArtworksRailQueryRenderer />
```
jonallured, line 68: Is this the right place? Shouldn't this be up higher in the list like between works for you and market news? 🤔

The-Beez-Kneez: Idk if there is a specification on where it is supposed to be on the Homepage. 

The design docs show it at the top under the main carousel in both the signed in and signed out flow.

src/v2/Apps/Home/Components/HomeTroveArtworksRail.tsx
```diff
@@ -0,0 +1,118 @@
...
+    />
+  )
+}
+
+export const HomeTroveArtworksRailFragmentContainer = createFragmentContainer(
+  HomeTroveArtworksRail,
+  {
+    viewer: graphql`
+      fragment HomeTroveArtworksRail_viewer on Viewer {
+        saleArtworksConnection(first: 12, geneIDs: "trove") {
```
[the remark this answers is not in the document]
TMcMeans: im definitely okay with pulling more art pieces to standardize the look but could be a design/Jenna question.

damassi (approved):
Looks great @TMcMeans 💯

PR #9564, opened 2022-03-07
Feat: analytics support for useVariant

mc-jones:
This PR adds analytics to the `useVariant` hook by sending an `Experiment Viewed` event to segment on the first instance the user is shown the flag's variation. 

### Considerations
To track if a user has seen the experiment before we are simply storing an array on `localStorage` with the name of the feature flag and the variant e.g., `${featureName}+${variantName}`.  On each `useFeatureVariant` we check the array to see if the experiment has been viewed before, and push it onto the array if not. This is a simple approach, but has the disadvantage of potentially sending multiple events for a single user in cases where they have cleared application storage from their browser. I'm not sure we can get around this issue by sticking with client-side tracking of repeat views. 

### Related PRs
- #9513
- #9498
- #9393
- #9547
- https://github.com/artsy/cohesion/pull/304

### Follow-ups
- Add support for caching (don't cache views with experiments)
- Confirm client-side tacking of repeat views is the approach we want to stick with.

### DIFF
--- a/src/v2/System/__tests__/useFeatureFlag.jest.tsx
+++ b/src/v2/System/__tests__/useFeatureFlag.jest.tsx
@@ -1,12 +1,21 @@
 import { useFeatureFlag, useFeatureVariant } from "../useFeatureFlag"
 import { useSystemContext } from "v2/System/useSystemContext"
+import { useTracking } from "react-tracking"
 
 jest.mock("v2/System/useSystemContext")
 
 const mockUseSystemContext = useSystemContext as jest.Mock
 let mockFeatureFlags
+let trackEvent
 
 beforeAll(() => {
+  trackEvent = jest.fn()
+  ;(useTracking as jest.Mock).mockImplementation(() => {
+    return {
+      trackEvent,
+    }
+  })
+
   mockFeatureFlags = {
     featureFlags: {
       "feature-a": {
--- a/src/v2/System/useFeatureFlag.tsx
+++ b/src/v2/System/useFeatureFlag.tsx
@@ -57,6 +57,7 @@ export function useFeatureVariant(featureName: string): Variant | null {
 }
 
 function shouldTrack(featureName: string, variantName: string): boolean {
+  // Value to set and read from the experimentsViewed key in localStorage.
   const experimentName = `${featureName}+${variantName}`
   const viewedExperiments = getExperimentsViewed()
 
--- a/src/v2/System/useFeatureFlag.tsx
+++ b/src/v2/System/useFeatureFlag.tsx
@@ -1,5 +1,7 @@
+import * as Schema from "v2/System/Analytics"
 import { useSystemContext } from "v2/System"
 import { Variant } from "unleash-client"
+import { useTracking } from "react-tracking"
 
 export type FeatureFlags = Record<string, FeatureFlagDetails>
 
@@ -25,6 +27,7 @@ export function useFeatureFlag(featureName: string): boolean | null {
 
 export function useFeatureVariant(featureName: string): Variant | null {
   const { featureFlags } = useSystemContext()
+  const { trackEvent } = useTracking()
   const variant = featureFlags?.[featureName]?.variant
 
   if (!variant) {
@@ -35,5 +38,47 @@ export function useFeatureVariant(featureName: string): Variant | null {
     return null
   }
 
+  // detremine if we should track the experimentView
+  if (typeof window !== "undefined") {
+    let trackFeatureView = maybeTrackFeatureView(featureName, variant.name)
+
+    if (trackFeatureView) {
+      trackEvent({
+        action: Schema.ActionType.ExperimentViewed,
+        service: "unleash",
+        experiment_name: featureName,
+        variant_name: variant.name,
+        payload: variant.payload,
+      })
+    }
+  }
+
   return variant
+}
+
+function maybeTrackFeatureView(featureName, variantName) {
+  const experimentName = `${featureName}+${variantName}`
+  let viewedExperiments = getExperimentsViewed()
+
+  if (viewedExperiments.includes(experimentName)) {
+    return false
+  }
+
+  viewedExperiments.push(experimentName)
+  setExperimentsViewed(viewedExperiments)
+
+  return true
+}
+
+function getExperimentsViewed() {
+  let experimentsViewed = window.localStorage.getItem("experimentsViewed")
+
+  return experimentsViewed === null ? [] : JSON.parse(experimentsViewed)
+}
+
+function setExperimentsViewed(_viewedExperiments) {
+  window.localStorage.setItem(
+    "experimentsViewed",
+    JSON.stringify(_viewedExperiments)
+  )
 }
--- a/src/v2/System/useFeatureFlag.tsx
+++ b/src/v2/System/useFeatureFlag.tsx
@@ -27,7 +27,6 @@ export function useFeatureFlag(featureName: string): boolean | null {
 
 export function useFeatureVariant(featureName: string): Variant | null {
   const { featureFlags } = useSystemContext()
-  const { trackEvent } = useTracking()
   const variant = featureFlags?.[featureName]?.variant
 
   if (!variant) {
@@ -38,22 +37,35 @@ export function useFeatureVariant(featureName: string): Variant | null {
     return null
   }
 
-  // detremine if we should track the experimentView
+  return variant
+}
+
+export function useTrackVariantView({
+  experiment_name,
+  variant_name,
+  payload,
+  context_owner_type,
+  context_owner_id,
+  context_owner_slug,
+}) {
+  const { trackEvent } = useTracking()
+
   if (typeof window !== "undefined") {
-    const trackFeatureView = shouldTrack(featureName, variant.name)
+    const trackFeatureView = shouldTrack(experiment_name, variant_name)
 
     if (trackFeatureView) {
       trackEvent({
         action: Schema.ActionType.ExperimentViewed,
         service: "unleash",
-        experiment_name: featureName,
-        variant_name: variant.name,
-        payload: variant.payload,
+        experiment_name,
+        variant_name,
+        payload,
+        context_owner_type,
+        context_owner_id,
+        context_owner_slug,
       })
     }
   }
-
-  return variant
 }
 
 function shouldTrack(featureName: string, variantName: string): boolean {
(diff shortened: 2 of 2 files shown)


src/v2/System/useFeatureFlag.tsx
```diff
@@ -35,5 +38,50 @@ export function useFeatureVariant(featureName: string): Variant | null {
...
+  return true
+}
+
+function getExperimentsViewed(): string[] {
+  let experimentsViewed = window.localStorage.getItem("experimentsViewed")
+
+  return experimentsViewed === null ? [] : JSON.parse(experimentsViewed)
+}
+
+function setExperimentsViewed(_viewedExperiments: string[]) {
```
damassi, line 82: `_someVariableName` (the underscore) signifies that a variable is unused

src/v2/System/useFeatureFlag.tsx
```diff
@@ -35,5 +38,50 @@ export function useFeatureVariant(featureName: string): Variant | null {
     return null
   }
 
+  // detremine if we should track the experimentView
+  if (typeof window !== "undefined") {
+    let trackFeatureView = maybeTrackFeatureView(featureName, variant.name)
```
damassi, line 43: let -> const

src/v2/System/__tests__/useFeatureFlag.jest.tsx
```diff
@@ -1,12 +1,21 @@
...
+import { useTracking } from "react-tracking"
 
 jest.mock("v2/System/useSystemContext")
 
 const mockUseSystemContext = useSystemContext as jest.Mock
 let mockFeatureFlags
+let trackEvent
 
 beforeAll(() => {
+  trackEvent = jest.fn()
```
pvinis, line 12: probably not for this PR, but this would be great in the "setup jest" part 👌

src/v2/System/useFeatureFlag.tsx
pvinis: probably the linter complained for shadowing and that's the reason for the underscore. I'd just ignore the rule 🤔

src/v2/System/useFeatureFlag.tsx
```diff
@@ -35,5 +38,50 @@ export function useFeatureVariant(featureName: string): Variant | null {
...
+        variant_name: variant.name,
+        payload: variant.payload,
+      })
+    }
+  }
+
   return variant
 }
+
+function maybeTrackFeatureView(
```
pvinis, line 59: I hate commenting on names, but I will do it 😅.
I think the convention with `maybe*` is that the function will either do or will not do the thing. so here, sometime it will track and sometimes it won't. but the function here actually never does the tracking. it's only checking if it should.

I think it's clearer if the func is called `shouldTrack` and returns a bool, then the if does the tracking, or the func is called `maybeTrack` and it does the check but includes the if with the tracking as well, and returns nothing.

does that make sense?

pvinis (approved):
nice! I added some suggestions. nothing is blocking though.

pvinis:
regarding "don't cache views with experiments", is there anything that could be done to cache per-user per-experimentvariant? 🤔 or would that be of little benefit?

src/v2/System/useFeatureFlag.tsx
damassi: better to just name it something else than to bypass strong built-in conventions

src/v2/System/useFeatureFlag.tsx
mc-jones: Totally right to call this out. This name was a holdover from the first pass, which had the `trackEvent` call within `maybeTrackFeatureView`. Agree that changing to `shouldTrack` is more clear.

joeyAghion (dismissed):
The risk you describe of sending multiple events seems totally acceptable to me. It would even be "correct" to track each instance of the experiment being "viewed," although that might be a little wasteful of Segment events. The denominator in any analytics will be the unique user/anonymous IDs rather than the raw events.

Tracking a trivial experiment (that affects both authenticated and unauthenticated visitors) should be part of this or a quick follow-up, so we can validate that the downstream data looks correct.

Q: How often does Force's `SESSION_ID` change compared to [Segment's `anonymousId`](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/)? For consistency of experience (even for unauthenticated visitors), I wonder if we should prefer using the `anonymousId`.

src/v2/System/useFeatureFlag.tsx
```diff
@@ -35,5 +38,50 @@ export function useFeatureVariant(featureName: string): Variant | null {
...
   }
 
+  // detremine if we should track the experimentView
+  if (typeof window !== "undefined") {
+    let trackFeatureView = maybeTrackFeatureView(featureName, variant.name)
+
+    if (trackFeatureView) {
+      trackEvent({
+        action: Schema.ActionType.ExperimentViewed,
+        service: "unleash",
```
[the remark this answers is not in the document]
pvinis: its more for future-proofing. if at some point we have another service, or we could even send something like `"echo"` for knowing its an old school flag if we want.

damassi (dismissed):
Looks good to me 👍

mc-jones:
> regarding "don't cache views with experiments", is there anything that could be done to cache per-user per-experimentvariant? 🤔 or would that be of little benefit?

One thing to note is we don't cache when we have an [authenticated user](https://github.com/artsy/force/blob/main/src%2Flib%2Fmiddleware%2FredisPageCache.ts#L53). But my initial feeling is that the value wouldn't be there compared with the complexity it might introduce.

mc-jones:
> Q: How often does Force's SESSION_ID change compared to [Segment's anonymousId](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/)? For consistency of experience (even for unauthenticated visitors), I wonder if we should prefer using the anonymousId.

Looks like we set the session here with `SESSION_COOKIE_MAX_AGE` here, which is `SESSION_COOKIE_MAX_AGE=5184000000` (60 days) in production. 

Based on that article you sent, segment's `anonymouID` [persists as long as the user doesn't clear their localStorage](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#segment-id-persistence). So in this sense, I think you may be right to to use that instead. This will allow [gradual rollouts](https://docs.getunleash.io/user_guide/activation_strategy) that use `session_id` as the stickiness factor to move together.

joeyAghion:
I was thinking (incorrectly) of "session" in more of the analytics sense of 30-60 minutes of activity. If the session ID actually persists for 60 days (meaning an anonymous visitor would retain the same split-test treatment for that long), my concern isn't anything to worry about. Up to you if you think we should change anything about the current implementation.

dimatretyak:
**thoughts**: maybe it's worth adding the ability to disable automatic sending of an `Experiment Viewed` event when calling `useFeatureVariant`? 

There may be situations in which we need to send an `Experiment Viewed` event only on certain pages (for example, as in [this task](https://artsyproduct.atlassian.net/browse/FX-3774))

src/v2/System/useFeatureFlag.tsx
```diff
@@ -35,5 +38,44 @@ export function useFeatureVariant(featureName: string): Variant | null {
...
+        experiment_name: featureName,
+        variant_name: variant.name,
+        payload: variant.payload,
+      })
+    }
+  }
```
dimatretyak, line 54: **thought**: maybe we should place this logic in `useEffect`?

mc-jones:
> thoughts: maybe it's worth adding the ability to disable automatic sending of an Experiment Viewed event when calling useFeatureVariant?

I'm wondering if this might be a good use for the option payload. that can be included with the variant? 
``` TSX
{
   ...
   "variant": {
    "name": "with-payload-string",
    "payload": {
      "type": "string", // can be json, csv, or string
      "value": "this string is the variant's payload"
    },
}
```

dimatretyak:
> @dimatretyak
> 
> > thoughts: maybe it's worth adding the ability to disable automatic sending of an Experiment Viewed event when calling useFeatureVariant?
> 
> I'm wondering if this might be a good use for the option payload. that can be included with the variant?
> 
> ```tsx
> {
>    ...
>    "variant": {
>     "name": "with-payload-string",
>     "payload": {
>       "type": "string", // can be json, csv, or string
>       "value": "this string is the variant's payload"
>     },
> }
> ```

sorry, I didn't quite get your idea 😅. If it's not too much trouble, can you please describe it in a little more detail?

src/v2/System/useFeatureFlag.tsx
damassi: Ah yeah, nice catch -- that's a must 👍

damassi:
> thoughts: maybe it's worth adding the ability to disable automatic sending of an Experiment Viewed event when calling useFeatureVariant?

My 2c: we should cross this bridge when we get to it and not overly complicate things at this stage, since the issue has not yet come up.

mc-jones:
> sorry, I didn't quite get your idea 😅. If it's not too much trouble, can you please describe it in a little more detail? Do you want to add the ability to specify pages in the payload on which an `Experiment Viewed` event should be triggered?

No worries. Was suggesting this based on a misreading of your need here (so can be safely ignored). But the general idea was to conditionally take action based on the payload values, yes. But, agree that parsing through that large of a payload would be difficult, and not particularly good DX.

mc-jones:
Have broken out the tracking call into its own function, as discussed [here](https://artsy.slack.com/archives/C02NAJ2CGLW/p1646848833002799?thread_ts=1645527881.920729&cid=C02NAJ2CGLW).  @dimatretyak and @damassi , I've also wrapped it in `useEffect`.

src/v2/System/useFeatureFlag.tsx
```diff
@@ -37,3 +39,56 @@ export function useFeatureVariant(featureName: string): Variant | null {
...
+    const trackFeatureView = shouldTrack(experiment_name, variant_name)
+
+    if (trackFeatureView) {
+      trackEvent({
+        action: Schema.ActionType.ExperimentViewed,
+        service: "unleash",
+        experiment_name,
+        variant_name,
+        payload,
+        context_owner_type,
```
mc-jones: @damassi - once we enforce the [cohesion schema](https://github.com/artsy/cohesion/pull/304), this is going to be required. What is the best way to grab this data?

damassi: If i'm understanding correctly, right now this isn't a dynamic value that we can "get at" -- it's generally supplied as a static value: 

Screen Shot 2022-03-09 at 12 57 08 PM

So if this will soon be enforced as a requirement, we might want to expand the hook arguments a bit:

```tsx
const useFeatureFlag = ({ flagName, analyticsOwnerType }) => { ... }
```

Something like that.

PR #9565, opened 2022-03-08
chore(auction2): Stub out tests 

damassi:
This stubs out the tests for /auction2 and defines all of the test-case behavior. I did this by opening each file and going line by line and looking at behavior and then stubbing out a test case to be filled in in a follow-up PR. 

#### A note about process here (entirely from my own experience):

When working on something entirely new, some folks prefer the TDD approach by defining test cases before writing any code and then making the tests pass, but I've found (personally) that with complicated UI code this is a bit rigid and can slow things down, especially when dealing with non-trivial flows that might frequently change during the design process as things are becoming more well understood. 

Additionally, there's so much test setup code that can't quite be known ahead of time (short of some kind of :galaxybrain:-like scenario), and attempting to define it ahead of time can lead to an enormous amount of churn in terms of design. 

With this in mind I always write tests at the very end when working on a new project. It keeps the initial development process free flowing and unblocked, and then when tests start to get written there's a clear snowball effect; as more of the test scaffolding gets written, the more it can be used in other tests and by the 40-50% point usually all of the pieces are in place and test writing becomes more of a production job with very little thought required. 

Once the project is out the door, however, and additional features are added / changed, PRs should always contain tests. The above only applies to greenfield-esque-like situations. 

cc @artsy/grow-devs


src/v2/DevTools/setupTestWrapper.tsx
```diff
@@ -119,7 +119,7 @@ export const setupTestWrapperTL = <T extends OperationType>({
 }
 
 /**
- * @deprecated Use `setupTestWrapperTL`, which uses `@testing-library/react`
+ * @note See `setupTestWrapperTL`, which uses `@testing-library/react`
```
damassi, line 122: If devs want to use RTL they can, but if they also want to use Enzyme that should be fine too. They're different things.

PR #9572, opened 2022-03-08
chore: add MNTN tracking pixels

laurabeth:
Addresses [GRO-775]

Add tracking pixel to footer for ubiquitous cover.
Add conversion pixel to app shell, rendering conditionally when a conversion route is found. Also added tracking pixel to the sign up modal, which doesn't change routes before displaying.

[GRO-775]: https://artsyproduct.atlassian.net/browse/GRO-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/index.ejs
```diff
@@ -76,6 +76,10 @@
...
+    <script type="text/javascript">(function(){"use strict";var e=null,b="4.0.0",n="32548",additional="term=value",t,r,i;try{t=top.document.referer!==""?encodeURIComponent(top.document.referrer.substring(0,2048)):""}catch(o){t=document.referrer!==null?document.referrer.toString().substring(0,2048):""}try{r=window&&window.top&&document.location&&window.top.location===document.location?document.location:window&&window.top&&window.top.location&&""!==window.top.location?window.top.location:document.location}catch(u){r=document.location}try{i=parent.location.href!==""?encodeURIComponent(parent.location.href.toString().substring(0,2048)):""}catch(a){try{i=r!==null?encodeURIComponent(r.toString().substring(0,2048)):""}catch(f){i=""}}var l,c=document.createElement("script"),h=null,p=document.getElementsByTagName("script"),d=Number(p.length)-1,v=document.getElementsByTagName("script")[d];if(typeof l==="undefined"){l=Math.floor(Math.random()*1e17)}h="dx.mountain.com/spx?"+"dxver="+b+"&shaid="+n+"&tdr="+t+"&plh="+i+"&cb="+l+additional;c.type="text/javascript";c.src=("https:"===document.location.protocol?"https://":"http://")+h;v.parentNode.insertBefore(c,v)})()</script>
+  <%% } %>
```
dzucconi, line 81: If we're loading it via the hook, why would we also need the script tag?

laurabeth: There are two scripts: The tracking pixel, which is meant to be ubiquitous, and the conversion pixel, which is the one we would load via hook. This is the former.

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -57,6 +57,25 @@ export const AppShell: React.FC<AppShellProps> = props => {
...
+      script.type = "javascript/text"
+
+      document.body.appendChild(script)
+    }
+    return () => {
+      if (document.getElementById("mntn_conversion")) {
+        document.body.removeChild(script)
+      }
+    }
+  }, [match])
```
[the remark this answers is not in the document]
laurabeth: Nice! I actually had a `useScript` hook written to the same effect... then began to wonder if I could use a hook inside a useEffect, because the dependency array isn't linked to the `src` changing, but the `match`. 

I'll check out the hook you've mentioned and see if this question even remains relevant.

src/v2/System/Analytics/MNTN/conversionPixelScript.js
```diff
@@ -0,0 +1,95 @@
+export default (function () {
```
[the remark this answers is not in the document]
laurabeth: Really good point. Will do some digging.

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -151,18 +151,12 @@ function idConversionRoute(pathname: string | undefined) {
...
-    // "/personalize/artists", // account creation
-    // "/personalize/budget", // account creation
-    // `/auction/${saleName}/bid/${artworkID}?bid=${bidID}`, // bid placed
-    // `/auction-registration/${saleName}`, // auction registration
-    // `/orders/7ef042fe-1897-4076-8f15-0d336db880c6/status`, // offer made, purchase
+    ["/signup"], // account creation
+    ["/personalize", "/artists"], // account creation
+    ["/personalize", "/budget"], // account creation
+    ["/auction", "/bid"], // bid placed
+    ["/auction-registration"], // auction registration
```
damassi, line 158: Ah! this has actually changed in the new app: `/auction/:id/register`

src/desktop/components/main_layout/templates/scripts.jade
```diff
@@ -121,10 +121,16 @@ if !sd.THIRD_PARTIES_DISABLED
...
+  script( type="text/javascript" ).
+    ;(function(){"use strict";var e=null,b="4.0.0",n="32548",additional="term=value",t,r,i;try{t=top.document.referer!==""?encodeURIComponent(top.document.referrer.substring(0,2048)):""}catch(o){t=document.referrer!==null?document.referrer.toString().substring(0,2048):""}try{r=window&&window.top&&document.location&&window.top.location===document.location?document.location:window&&window.top&&window.top.location&&""!==window.top.location?window.top.location:document.location}catch(u){r=document.location}try{i=parent.location.href!==""?encodeURIComponent(parent.location.href.toString().substring(0,2048)):""}catch(a){try{i=r!==null?encodeURIComponent(r.toString().substring(0,2048)):""}catch(f){i=""}}var l,c=document.createElement("script"),h=null,p=document.getElementsByTagName("script"),d=Number(p.length)-1,v=document.getElementsByTagName("script")[d];if(typeof l==="undefined"){l=Math.floor(Math.random()*1e17)}h="dx.mountain.com/spx?"+"dxver="+b+"&shaid="+n+"&tdr="+t+"&plh="+i+"&cb="+l+additional;c.type="text/javascript";c.src=("https:"===document.location.protocol?"https://":"http://")+h;v.parentNode.insertBefore(c,v)})()
+
```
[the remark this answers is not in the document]
damassi: Page launched, so this code can be removed

src/v2/Apps/Components/AppShell.tsx
laurabeth: Will update, thanks!

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -152,32 +152,18 @@ function idConversionRoute(pathname: string | null) {
...
-    // `/auction/${saleName}/bid/${artworkID}?bid=${bidID}`, // bid placed
-    // `/auction-registration/${saleName}`, // auction registration
-    // `/orders/7ef042fe-1897-4076-8f15-0d336db880c6/status`, // offer made, purchase
-    //  no route for 'gallery contacted',
+  console.log("AppShell::idConversionRoute: pathname", pathname)
+  const conversionRoutes: RegExp[] = [
+    /\/signup/g,
+    /(\/auction\/)([a-z]|[0-9]|\-)+(\/bid\/)([a-z]|[0-9]|\-)+\?(sort=)([a-z]|\_)+\&(bid\=)([0-9]+)/g,
+    /\/auction\/registration\/([a-z]|[0-9]|\-)+/g,
+    /(\/orders\/)([0-9]|[a-z]|\-)+(\/status)/g,
```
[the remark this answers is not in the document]
dzucconi: ☝️ — https://github.com/component/path-to-regexp

damassi:
This is a good first pass! A couple suggestions before merge that would make this simpler:
- We should be using `path-to-regex`, as suggested above. We already have it in the app and use it elsewhere. Using it, we no longer need to mess with custom regexes and the logic is simplified 
- We no longer need to add the script to the .jade template now that the auctions page is merged; that can be removed
- Any follow-up on the question about using a CDN script?  
- We should remove all of the MTN script logic from `AppShell` so that that that file can remain clear and focused. 

While this is significantly different, as a design suggestion I would suggest playing with something like the below (and would be happy to pair on this!):

1. We should create a new component in `v2/Components/MTNTrackingPixel.tsx`
2. In it, we can conditionally load the script based on path name, and if it doesn't match, and return `null`. eg, 

```tsx
import { match as matchPath } from 'path-to-regex'
import { useRouter } from 'v2/System/Router/useRouter"
import { useLoadScript } from "v2/Utils/Hooks" 

export const MTNTrackingPixel = () => {
  const router = useRouter()
  
  const routePatterns = [
    "/signup",
    "/auction/:slug/bid/:artworkSlug",
    "/auction/:slug/register",
    "/order/:id/status"
  ]
  
  const matcher =  matchPath(routePatterns, { decode: decodeURIComponent })
  const foundMatch = !!matcher(router.match?.location?.pathname)
  
  if (foundMatch) {
    return <MTNScript /> 
  }
  
  return null
}

const MTNScript = () => {
  useLoadScript({ id: "mtnTrackingID", src: "path/to/CDN/or/script")
  return null 
}
```

Then later, in `AppShell`, we can load it like so:

```
return (
 ... other app shell components 
 <MTNTrackingPixel /> 
)
```

Also a couple other misc notes:
- As feedback is resolved in comments in PRs, its good to make a comment noting the resolution or add a thumbs up, and then mark it as resolved so that the conversation thread collapses and all that remains on the PR is actionable items
- If there's a comment with a suggestion and / or actionable item, best to address it rather than [leave folks hanging](https://github.com/artsy/force/pull/9572#discussion_r833434992)
- We should be using semantic commits and PR titles per [this accepted RFC](https://github.com/artsy/README/issues/327))

src/v2/System/Analytics/MNTN/conversionPixelScript.js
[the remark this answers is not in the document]
damassi: 👍

damassi:
> I usually do a first pass where I just get the thing working to my satisfaction, then go back to the PR to read over comments, make changes, respond to or resolve convos.

Its all good! Might be helpful to explain a bit more from a reviewers perspective tho: I can't speak for everyone, but I know Jon, Damon and I tend to pretty instantly respond to PRs and other comms, and so if something has been opened and assigned, but previous comments haven't been addressed (even if one does plan on looping back later), that creates some confusion / inefficiency on the reviewers side. That was why i reached out on slack yesterday to confirm it was ready to look at; i saw that it was opened, but saw a pretty crucial piece of PR review unaddressed / unacknowledged, which was confusing and so i wasn't sure if that was a mistake. So best to button that stuff up before final review, if possible 🙏  Even a tiny comment like "will look at this after so and so" is enough to let a reviewer know that the comment isn't left hanging during a review cycle.

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -19,6 +19,7 @@ import { useNavBarHeight } from "v2/Components/NavBar/useNavBarHeight"
 import { useProductionEnvironmentWarning } from "v2/Utils/Hooks/useProductionEnvironmentWarning"
 import { useAuthValidation } from "v2/Utils/Hooks/useAuthValidation"
 import { Z } from "./constants"
+import { MntnConversionPixel } from "../../System/Analytics/MNTN"
```
damassi, line 22: This is super minor and no real need to fix (but def something to keep an eye on with auto-imports from VSCode). For anything that's more that one level deep, its best to replace with absolute paths since it makes refactoring so much easier later, eg, 

```tsx
import { MntnConversionPixel } from "v2/System/Analytics/MNTN"
```

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -99,6 +77,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
   useNetworkOfflineMonitor()
   useProductionEnvironmentWarning()
 
+  console.log("AppShell::preRender", pathname)
```
damassi, line 80: We should be sure to remove console logs

src/v2/Apps/Components/AppShell.tsx
```diff
@@ -141,26 +120,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
           </Flex>
         )}
       </Theme>
+      <MntnConversionPixel path={pathname} />
```
damassi, line 123: No need to pass the `path` prop as we can pluck it from the hook (see below)

src/v2/Components/Authentication/Views/SignUpForm.tsx
```diff
@@ -18,6 +18,7 @@ import { AuthenticationPasswordInput } from "../Components/AuthenticationPasswor
 import { AuthenticationCheckbox } from "../Components/AuthenticationCheckbox"
 import { AuthenticationFooter } from "../Components/AuthenticationFooter"
 import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
+import { MntnConversionPixel } from "../../../System/Analytics/MNTN"
```
damassi, line 21: ditto on abs path

damassi: Also, since we'll probably not expand the Analytics folder as most of our tracking / analytics stuff is dynamically embedded via segment, it would be a bit simpler to move this to `v2/Components/MTNTrackingPixel.tsx` and treat it like a normal component.

src/v2/Components/Authentication/Views/SignUpForm.tsx
```diff
@@ -305,6 +288,7 @@ export class SignUpForm extends Component<SignUpFormProps, SignUpFormState> {
                   showRecaptchaDisclaimer={this.props.showRecaptchaDisclaimer}
                 />
               </Join>
+              <MntnConversionPixel path="/signup" />
```
damassi, line 291: Ditto, no need to pass path

damassi (commented):
Looking good @laurabeth - just a couple more changes and we should be good to go.

src/v2/System/Analytics/MNTN/MntnConversionPixel.tsx
```diff
@@ -12,7 +12,7 @@ const MntnConversionPixel: FC<MntnConversionPixelProps> = ({ path }) => {
     if (matchFound && !document.getElementById("mntn_conversion")) {
       script = document.createElement("script")
       script.id = "mntn_conversion"
-      script.src = src
+      script.text = text
```
damassi, line 15: Interesting, perhaps we could update `useLoadScript` to check for `src` _or_ `text` (if one passes in `src`, use that, if one passes in `text`, use that. I bet that would solve the issue.

src/v2/Apps/Components/AppShell.tsx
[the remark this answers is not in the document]
laurabeth: added

src/v2/Components/Authentication/Views/SignUpForm.tsx
laurabeth: In this case there is no path for the component rendering the pixel, so it has to be passed. This is the case where we need to track a conversion event inside the signup modal.

src/v2/System/Analytics/MNTN/MntnConversionPixel.tsx
```diff
@@ -0,0 +1,72 @@
+import { match } from "path-to-regexp"
+import React, { FC, useEffect } from "react"
+
+interface MntnConversionPixelProps {
+  path: string | null
+}
+
+const MntnConversionPixel: FC<MntnConversionPixelProps> = ({ path }) => {
```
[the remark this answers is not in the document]
laurabeth: done, and ditto re: path prop.

damassi (approved):
Just did another review pass and looks good @laurabeth 👍 Thanks for working through all the various feedback on this. Feel free to merge when ready 🚢

src/v2/Components/Authentication/Views/SignUpForm.tsx
laurabeth: Would also be useful to use in the case of gallery inquiries but a double render there is removing the pixel immediately after it's added to the DOM. May work on this in a follow up.

laurabeth:
@damassi looked back over these changes this morning and I can't figure out why they're not working vs the previous implementation. If you'd like to pair on this later I'm up for it, otherwise I'm inclined to just revert to  eec6651 and call it a day.

damassi:
I can pull down the branch in a moment and give it some  👀

laurabeth:
@damassi Circling back to this after way too long. Just wanted to let you know I hear you on this:

>  if something has been opened and assigned, but previous comments haven't been addressed (even if one does plan on > looping back later), that creates some confusion / inefficiency on the reviewers side.

I was amped after the transfer to grow and definitely tagged people too early in my process. I'll be more aware next time. Also, thanks again for all the help on this!

damassi:
Know that it's always good to get early feedback, just be sure to keep WIP in the title or the PR in draft status 👍

PR #9574, opened 2022-03-09
feat(TX-231): release new order breakdown

rquartararo:
[TX-231](https://artsyproduct.atlassian.net/browse/TX-231)

This PR releases the [new order breakdown](https://artsyproduct.atlassian.net/browse/TX-90) by taking it out from behind the Avalara Phase 2 feature flag.

### DIFF
--- a/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
+++ b/src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
@@ -13,8 +13,6 @@ import {
   Text,
   StackableBorderBox,
 } from "@artsy/palette"
-import { userHasLabFeature } from "v2/Utils/user"
-import { useSystemContext } from "v2/System/SystemContext"
 
 export interface ArtworkSummaryItemProps extends Omit<FlexProps, "order"> {
   order: ArtworkSummaryItem_order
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -80,15 +76,13 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
         <Text variant="sm" color="black60">
           {shippingOrigin}
         </Text>
-        {user && userHasLabFeature(user, "Avalara Phase 2") && (
```
starsirius, line 83: Is the lab feature still used after this PR? If not, we can retire it!

starsirius (approved):
Exciting!! 🎉

src/v2/Apps/Order/Components/ArtworkSummaryItem.tsx
rquartararo: @starsirius Confirmed with Roberta this will be staying behind the flag.

starsirius: Ohhh yeah you're right. Let's keep it! Thanks for double checking!

PR #9575, opened 2022-03-09
chore: Removed usage of Make Offer On All Eligible Artworks flag

jo-rs:
`force` side of [NX-3048]

Releases Make Offer On All Eligible Artworks.
cc @artsy/negotiate-devs 

[NX-3048]: https://artsyproduct.atlassian.net/browse/NX-3048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -81,8 +81,8 @@ export class ReviewRoute extends Component<ReviewProps> {
       const orderOrError =
         this.props.order.mode === "BUY"
           ? (await this.submitBuyOrder()).commerceSubmitOrder?.orderOrError
-          : (await this.submitOffer(setupIntentId)).commerceSubmitOrderWithOffer
-              ?.orderOrError
+          : (await this.submitOffer(setupIntentId))
```
araujobarret, line 84: not 100% sure, but I guess we haven't touched anything here, should stay with the Exchange mutation instead of the new one

araujobarret (changes requested):

araujobarret (dismissed):
Nvm my last comment, wasn't seeing clearly the block 😢

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -407,8 +401,7 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
     }
 
     const shouldDisplayMakeOfferButton: boolean | null =
-      isOfferable ||
-      (isPriceListed && isOfferableFromInquiry && makeOfferEnabled)
+      isOfferable || (isPriceListed && isOfferableFromInquiry)
```
lidimayra, line 404: We shouldn't have to check for `(isPriceListed && isOfferableFromInquiry)` anymore. `ForSale#offerable?` is handling these:
https://github.com/artsy/gravity/blob/e69d1154a6371249d791ede0d765c15ef8671473/app/models/util/for_sale.rb#L290-L294
https://github.com/artsy/gravity/blob/e69d1154a6371249d791ede0d765c15ef8671473/app/models/util/for_sale.rb#L383-L384

There wouldn't be much problem about being redundant, but the issue here is that this check is not taking partner `inquiry_order_enabled` flag into account, which means that we display the `Make Offer` button for any artwork that's offerable from inquiry in case the price is not hidden (even if `inquiry_order_enabled` is off).

I would suggest removing this entirely.
If we want to be extra careful for some reason and we don't want to remove it yet, I guess it would be better to also add the inquiry_order_enabled check then.

jo-rs: Makes sense. Removing the other two now, and will ticket a cleanup for this file which is a bit off, but it's out of this scope.

lidimayra (approved):

PR #9594, opened 2022-03-10
feat: pass in node env to unleash

mc-jones:
This PR passes in the environment to the Unleash client upon initialization. This will enable us to check for the environment in `context` from a [custom strategy.](https://docs.getunleash.io/advanced/custom_activation_strategy). Or, if we upgrade to a paid version, we can use [strategy constraints](https://docs.getunleash.io/advanced/strategy_constraints).

### DIFF
--- a/src/lib/featureFlags/unleashService.ts
+++ b/src/lib/featureFlags/unleashService.ts
@@ -12,7 +12,9 @@ export class UnleashFeatureFlagService implements FeatureFlagService {
   constructor(
     private url = UNLEASH_API,
     private appName = UNLEASH_APP_NAME,
-    private serverKey = UNLEASH_SERVER_KEY
+    private serverKey = UNLEASH_SERVER_KEY,
+    private environment = process.env.NODE_ENV,
+    private refreshInterval = 60000
   ) {}
 
   async init(): Promise<void> {
@@ -23,8 +25,8 @@ export class UnleashFeatureFlagService implements FeatureFlagService {
     this._unleash = await startUnleash({
       url: this.url,
       appName: this.appName,
-      refreshInterval: 60000,
-      environment: process.env.NODE_ENV,
+      refreshInterval: this.refreshInterval,
+      environment: this.environment,
       customHeaders: {
         Authorization: this.serverKey,
       },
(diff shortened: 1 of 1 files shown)


damassi:
Mind updating the PR title to be semantic-commit friendly?

src/lib/featureFlags/unleashService.ts
```diff
@@ -12,7 +17,9 @@ export class UnleashFeatureFlagService implements FeatureFlagService {
   constructor(
     private url = UNLEASH_API,
     private appName = UNLEASH_APP_NAME,
-    private serverKey = UNLEASH_SERVER_KEY
+    private serverKey = UNLEASH_SERVER_KEY,
+    private environment = NODE_ENV,
```
dblandin, line 21: Unlike `Rails.env` which can resolve to either `development`, `staging`, or `production`, I believe `NODE_ENV` is usually either `development` or `production`. Both our staging and production environments currently have `NODE_ENV` set to `production`.

We could spin up a new `UNLEASH_ENV` environment variable to thread through here.

Also noting that the Unleash documentation has marked the strategy constraint approach [as deprecated](https://docs.getunleash.io/advanced/strategy_constraints#constrain-on-a-specific-environment) but I don't see much of an alternative if we're only given two named environments, even with Pro. Enterprise is the next plan that would allow to use create new environments.

If we move forward with an open-source compatible custom environment strategy, there's an example [here](https://github.com/Unleash/unleash-client-node/blob/bbf99bcacfc67c516b9c81d80494141e17b5e8d5/test/unleash.test.js#L11-L19).

mc-jones: Yeah, you're correct, I believe we're always running staging in production mode.

For a custom strategy, we do already get `appName` which is going to be  `force` for `production`, `forcestg` for `staging`, and `forcedev` for `developement`. This is returned out in the context so we could still use a similar approach as to what you suggest, just evaluate the `appName` instead of the environment.

Do believe it still makes sense to pass in the env, as this may be useful for different cases.

mc-jones: Also, writing this out makes me realize we should change the app names to better match other conventions, so I will change `forcestg` to `force-staging`, and `forcedev` to `force-development`.

joeyAghion (approved):

PR #9599, opened 2022-03-11
feat: add simple unleash experiment in artist filters

gkartalis:
# Description

- Created `filters-expanded-experiment` flag in Unleash dashboard
- Activation strategy:
  - gradual rollout 100%
  - stickiness: `userID`
- Variants:
  - we created the [variants](https://docs.getunleash.io/advanced/toggle_variants) that contain the actual experiment. Here we define a new variant as well as a control variant. It is worth noticing that all the variants will be weighted equally. You are able to define many variants – in this example, we have created 2. “control” is the control variant, while “experiment” is the alternative we want to test.
- add tracking `ExperimentViewed` event
- updated types and imports for analytics.

### Thoughts

Noticed that there is no way of having two levels of `stickiness` 🤔 

- If we pick `sessionID` stickiness we result in having a consistent result during a session BUT if the user starts a new session they might have a different experience.
- If we pick `userID` stickiness we might have some inconsistencies like user before logging in has a different experience than after logging in.

In the end we decided to go with `userID`

Jira: [FX-3766](https://artsyproduct.atlassian.net/browse/FX-3766)

### DIFF
--- a/docs/unleash_a-b_testing.md
+++ b/docs/unleash_a-b_testing.md
@@ -1,0 +1,53 @@
+# A/B Testing with Unleash
+
+## Creating a test
+
+- Log into unleash.artsy.net
+- Press Create feature toggle
+  - add a name, description and select the toggle type as "Experiment"
+- Go to Variants Tab and add the variant names (e.g. "control" and "experiment").
+- Go to Strategies Tab and pick your strategy, usually this will be Gradual Rollout for experiments.
+- select a stickiness parameter (userId will make sure that a user will have the same experience every time they use the website)
+- Go to Overview and enable it in development
+
+Inside the component that you want to access the variants use the `useFeatureVariant` hook.
+
+```diff
++   const variant = useFeatureVariant("my-awesome-experiment")
+
+```
+
+variant will be an object with this structure
+
+```js
+{
+  name: "my-variant-name",
+  enabled: true
+}
+```
+
+### Adding tracking data and confirming that it is received
+
+In the component that you added the experiment you just need to add the following:
+
+```diff
++  const {
++    contextPageOwnerId,
++    contextPageOwnerSlug,
++    contextPageOwnerType,
++  } = useAnalyticsContext()
+
++  useTrackVariantView({
++    experiment_name: "my-awesome-experiment",
++    variant_name: variant?.name!,
++    context_owner_id: contextPageOwnerId,
++    context_owner_slug: contextPageOwnerSlug,
++    context_owner_type: contextPageOwnerType,
++  })
+```
+
+In [segment live debugger](https://app.segment.com/artsy-engineering/sources/force-staging/debugger) you can see a live stream of all the events on staging. You can use the searchbar to filter out `Experiment Viewed`, click on the one that is yours and review that everything is in order.
+
+### Useful links
+
+[Setup an ab experiment in 3 simple steps](https://www.getunleash.io/blog/a-b-n-experiments-in-3-simple-steps)
--- a/docs/unleash_a-b_testing.md
+++ b/docs/unleash_a-b_testing.md
@@ -2,7 +2,7 @@
 
 ## Creating a test
 
-- Log into unleash.artsy.net
+- Log into [Unleash](unleash.artsy.net)
 - Press Create feature toggle
   - add a name, description and select the toggle type as "Experiment"
 - Go to Variants Tab and add the variant names (e.g. "control" and "experiment").
--- a/docs/unleash_a-b_testing.md
+++ b/docs/unleash_a-b_testing.md
@@ -38,11 +38,12 @@ In the component that you added the experiment you just need to add the followin
 +  } = useAnalyticsContext()
 
 +  useTrackVariantView({
-+    experiment_name: "my-awesome-experiment",
-+    variant_name: variant?.name!,
-+    context_owner_id: contextPageOwnerId,
-+    context_owner_slug: contextPageOwnerSlug,
-+    context_owner_type: contextPageOwnerType,
++    experimentName: "my-awesome-experiment",
++    variantName: variant?.name!,
++    contextOwnerId: contextPageOwnerId,
++    contextOwnerSlug: contextPageOwnerSlug,
++    contextOwnerType: contextPageOwnerType,
++    shouldTrackExperiment: true/false/condition
 +  })
 ```
 
--- a/src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
@@ -17,6 +17,7 @@ import {
   useTrackVariantView,
 } from "v2/System/useFeatureFlag"
 import { useAnalyticsContext } from "v2/System"
+import { OwnerType } from "@artsy/cohesion"
 
 interface ArtworkFiltersProps {
   user?: User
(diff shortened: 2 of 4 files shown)


src/v2/Components/ArtworkFilter/__tests__/ArtworkFilterMobileActionSheet.jest.tsx
```diff
@@ -10,10 +11,20 @@ import { ArtworkFilters } from "../ArtworkFilters"
 jest.mock("v2/Utils/Hooks/useMatchMedia", () => ({
   __internal__useMatchMedia: () => ({ sm: true }),
 }))
+jest.mock("v2/System/Analytics/useTracking")
```
gkartalis, line 14: mocks `useTracking` since the tests were failing without this (after the addition of `useTrackVariantView` hook in `ArtworkFilters/index.ts`

damassi: (Note this is a common pattern in our tests where tracking intersects)

docs/unleash_a-b_testing.md
```diff
@@ -0,0 +1,53 @@
+# A/B Testing with Unleash
+
+## Creating a test
+
+- Log into unleash.artsy.net
```
damassi, line 5: can we make this a proper markdown link?

dimatretyak (approved):
LGTM. I only have one question: should we expand filters only on desktop?

damassi (approved):
👌 👌

gkartalis:
> LGTM. I only have one question: should we expand filters for experiment only on desktop?

@dimatretyak Exactly, it is supposed to affect only desktop filters

dimatretyak (approved):

dblandin (commented):
Looking good!

Are we limiting the A/B test to only filters on the artist page grids? I'm not seeing a condition for that as is.

> gradual rollout 100%

What does this mean in an A/B test context?

gkartalis:
> Are we limiting the A/B test to only filters on the artist page grids? I'm not seeing a condition for that as is.

We don't right now, we are using this component almost everywhere, was thinking to use `OwnerType.artist` to differentiate between screens, what do you think about this?

One other thing that got me thinking was the following: 

We are using the `useTrackVariantView` in order to keep track of analytics which is a hook but how can we conditionally use it in cases like ours (running an experiment on a reusable component only for when it appears on artist screen)? 🤔 

> gradual rollout 100%

Gradual rollout 100% means making the flag enabled to 100% of our users. By selecting that we can also make use of the stickiness to userID (which is not possible if we don't specify that gradual rollout strategy via unleash)

Unleash then takes into account the variants and when having two by default 50% of users will get 
`control` variant and the other 50% will get the `experiment` variant.

mc-jones:
This is awesome. Thanks for kicking off the first experiment and creating docs!

>We don't right now, we are using this component almost everywhere, was thinking to use OwnerType.artist to differentiate between screens, what do you think about this?
>
>One other thing that got me thinking was the following:
>
>We are using the useTrackVariantView in order to keep track of analytics which is a hook but how can we conditionally use it in cases like ours (running an experiment on a reusable component only for when it appears on artist screen)? 🤔

@gkartalis - This is a good call out, and will be important given that we've [limited the number of events sent to the segment](https://github.com/artsy/force/blob/c2546164bbbe37004ec2ff3b6e1b6d5b5ae34d66/src/v2/System/useFeatureFlag.tsx#L74-L97). For widely reused components, seems likely we would have scenarios where the tracking event would fire and the user possibly didn't actually see the experiment or control. I'll leave it to you for how you want to handle this specific experiment, and will look at what we may be able to add in the `useTrackVariantView` to handle these cases.

mc-jones (dismissed):

src/v2/System/useFeatureFlag.tsx
```diff
@@ -11,6 +11,15 @@ interface FeatureFlagDetails {
   variant: Variant
 }
 
+interface VariantTrackingProperties {
+  experiment_name: string
+  variant_name: string
+  payload?: string
+  context_owner_type?: OwnerType
```
mc-jones, line 18: `context_owner_type` is required by the [cohesion schema](https://github.com/artsy/cohesion/blob/85bdd8b2eb4334497f9c9856e99cdd70424c6af6/src/Schema/Events/ExperimentViewed.ts#L38), so should be required here, also.

src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
```diff
@@ -21,6 +26,24 @@ interface ArtworkFiltersProps {
...
   const { user, relayEnvironment } = props
+  const {
+    contextPageOwnerId,
+    contextPageOwnerSlug,
+    contextPageOwnerType,
+  } = useAnalyticsContext()
+
+  const variant = useFeatureVariant("filters-expanded-experiment")
+
+  useTrackVariantView({
```
dblandin, line 37: **question**: I feel like this hook needs to be deconstructed into functions that can be called conditionally or it needs to support a prop that signals whether the analytics call should be sent.

As is, we can't supply any conditions here which is tricky when we want to use the hook within shared components.

@gkartalis Maybe for now we can call `trackEvent` explicitly and not rely on this new hook?

gkartalis: Will call `trackEvent` explicitly for now 💯

src/v2/System/useFeatureFlag.tsx
```diff
@@ -11,6 +11,16 @@ interface FeatureFlagDetails {
...
 }
 
+interface VariantTrackingProperties {
+  experiment_name: string
+  variant_name: string
+  payload?: string
+  context_owner_type: OwnerType
+  context_owner_id?: string
+  context_owner_slug?: string
+  should_track?: boolean
```
gkartalis, line 21: not sure if the name is the best since we have a should_track function but with this way making it optional and defaulting to true we will be able to reuse this. @dblandin @mc-jones what do you think?

dblandin: I like that approach!

On a separate non-blocking note: it feels a bit odd to have snake_case keys in the interface here. If this is meant to be an abstracted API, would it make sense to support camelCase keys and map/convert before the internal trackEvent call?

dblandin (dismissed):
LGTM!

src/v2/System/useFeatureFlag.tsx
gkartalis: > On a separate non-blocking note: it feels a bit odd to have snake_case keys in the interface here. If this is meant to be an abstracted API, would it make sense to support camelCase keys and map/convert before the internal trackEvent call?

gave it a go I think it is ready 🤩 when you have some time just have another look 🎉

src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
```diff
@@ -21,6 +27,28 @@ interface ArtworkFiltersProps {
...
+    experimentName: "filters-expanded-experiment",
+    variantName: variant?.name!,
+    contextOwnerId: contextPageOwnerId,
+    contextOwnerSlug: contextPageOwnerSlug,
+    contextOwnerType: contextPageOwnerType!,
+    shouldTrackExperiment: isArtistPage,
```
dblandin, line 46: **praise**: 🤗

dblandin (approved):
Awesome! Looks great!! 💯

PR #9607, opened 2022-03-14
fix [CX-2424]: Hide your gallery intro from the collector profile

MrSltun:
This PR Resolves [CX-2424]

This PR hides the Gallery Intro from the collector profile since it is not being used anymore in the CMS

| Before | After |
|--|--|
| with gallery intro | without gallery intro |

### Side note 
I left the component as reference when we will be adding the new collector

[CX-2424]: https://artsyproduct.atlassian.net/browse/CX-2424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olerichter00 (approved):
🌟

PR #9610, opened 2022-03-14
Mc jones/unleash cache key

mc-jones:
This PR adds caching support for our feature flag and experimentation strategy. Following the same approach as our current split test implementation, we generate a list of running feature flags per request and add that to the cache key.

### Considerations
We currently have an [active A/B test running](https://github.com/artsy/force/pull/9538), so the new approach in this PR will be running alongside the old approach until it is completed. Once the final A/B test is complete, we will remove the legacy code and just use the `unleash` based implementation.

### Related PRs
- https://github.com/artsy/force/pull/9513
- https://github.com/artsy/force/pull/9498
- https://github.com/artsy/force/pull/9393
- https://github.com/artsy/force/pull/9547
- https://github.com/artsy/cohesion/pull/304

### Follow-ups
- Run a trivial experiment with new implementation
- Remove legacy split test code

[PLATFORM-4024]

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -127,6 +127,10 @@ export function initializeMiddleware(app) {
     app.use(splitTestMiddleware)
   }
 
+  // Need sharify for unleash
+  registerFeatureFlagService(UnleashService, UnleashFeatureFlagService)
+  app.use(featureFlagMiddleware(UnleashService))
+
   // Initialize caches
   applyCacheMiddleware(app)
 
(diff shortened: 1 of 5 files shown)


src/middleware.ts
```diff
@@ -127,6 +127,10 @@ export function initializeMiddleware(app) {
     app.use(splitTestMiddleware)
   }
 
+  // Need sharify for unleash
+  registerFeatureFlagService(UnleashService, UnleashFeatureFlagService)
+  app.use(featureFlagMiddleware(UnleashService))
```
mc-jones, line 132: Need to setup unleash before we `applyCacheMiddleware` since it relies on `res.locals.FEATURE_FLAGS'

damassi (dismissed):
👌 looks great

PR #9615, opened 2022-03-14
fix(GRO-880): Fleshes out article meta tags

dzucconi:
Closes: [GRO-880](https://artsyproduct.atlassian.net/browse/GRO-880)


jonallured (dismissed):
nice one 😍

PR #9616, opened 2022-03-14
fix: Add scope parameter to Google strategy

jacobherrington:
Based on the documentation, I think we should be doing this anyway. I'm hoping it will resolve https://artsyproduct.atlassian.net/browse/GRO-879 as well.

### DIFF
--- a/src/lib/passport/lib/passport/index.js
+++ b/src/lib/passport/lib/passport/index.js
@@ -49,6 +49,7 @@ module.exports = function () {
           clientSecret: opts.GOOGLE_SECRET,
           passReqToCallback: true,
           callbackURL: `${opts.APP_URL}${opts.googleCallbackPath}`,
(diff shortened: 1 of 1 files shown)


jonallured:
I'd like to know more - can you drop a link to the docs that caused you to want to add this? Thanks!

jacobherrington:
@jonallured Well, the [README](https://github.com/jaredhanson/passport-google-oauth2#configure-strategy) has this included when setting up the strategy. 😄

I also found an [open issue](https://github.com/jaredhanson/passport-google-oauth2/issues/76) with the same error with a suggestion to try this, which we should probably do anyway. Since we can't replicate the issue we saw in the video on Jira, I'm sort of hoping I just missed this in the docs and it will resolve that issue too!

jonallured (approved):
Hope is an important part of engineering!!

PR #9618, opened 2022-03-15
chore(auction2): Wire up and test auction tracking

damassi:
This wires up /auction2 tracking and adds some test around the code. 

Next pr will contain last little bit of qa, then a launch PR. 

cc @artsy/grow-devs


damassi:
mergin' to continue

PR #9619, opened 2022-03-15
chore(auction2): Last QA

damassi:
More last minute QA, mostly related to the bid form and ensuring validation works as well as guarding against multiple credit cards or addresses being submitted when the reserve price isn't met, but the initial credit card / registration is successful. 

cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Auction2/Queries/useRefreshUserData.ts
+++ b/src/v2/Apps/Auction2/Queries/useRefreshUserData.ts
@@ -1,0 +1,27 @@
+import { Environment, fetchQuery, graphql } from "relay-runtime"
+import { useSystemContext } from "v2/System"
+
+/**
+ * After certain operations (like adding a credit card during a bid), we need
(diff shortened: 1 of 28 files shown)


src/v2/Apps/Auction2/Queries/useRefreshUserData.ts
```diff
@@ -0,0 +1,27 @@
...
+export const useRefreshUserData = () => {
+  const { relayEnvironment } = useSystemContext()
+
+  const refreshUserData = () => {
+    return fetchQuery(
+      relayEnvironment as Environment,
+      graphql`
+        query useRefreshUserDataQuery {
+          me {
+            ...Auction2BidRoute_me
```
damassi, line 18: In both mutations as well as queries, this works great for refreshing the relay store / UI -- just pass the fragment in that you want to refresh, and it will update.

PR #9628, opened 2022-03-15
feat: Artsy shipping international

leamotta:
https://artsyproduct.atlassian.net/browse/PX-4937

Request Artsy Shipping quotes when the artwork has Artsy Shipping International activated and the collector is outside of the US.


src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -192,8 +194,8 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
     return (
       this.state.shippingOption === "SHIP" &&
-      processWithArtaShipping &&
-      shippingCountry === "US"
+      ((processWithArtaShipping && shippingCountry === "US") ||
+        (artsyShippingInternational && shippingCountry !== "US"))
```
jpotts244, line 198: Just for my own learning 🧠 :

Am I correct in saying will be relying on validations on the CMS side of things to handle checks of where the artwork is located? Assuming the protection to only allow US exports live there?

jpotts244, line 198: Just for my own learning 🧠 :

Am I correct in saying will be relying on validations on the CMS side of things to handle checks of where the artwork is located? Assuming the protection to only allow US exports live there?

jpotts244 (approved):

src/v2/Apps/Order/Routes/Shipping/index.tsx
leamotta: Yes!
This logic will have to be changed when new countries are included into Artsy Shipping, we will have to check artwork location against user address...
Though nothing blocks us to do it now, I think I'll give it a try

jpotts244: No presh! we can handle that when we get to it

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -192,8 +194,9 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
```
leamotta, line 198: the remaining Q is if we should rename artsyShippingInternational to processWithArtsyShippingInternational cc: @oxaudo @jpotts244 @ansor4

oxaudo: That was not a rename suggestion. I was suggesting to add a parallel method to `processWithArtaShipping` that for now will only check for `artsyShippingInternational` in gravity. But again - don't feel too strongly about it. I'm just very skeptical that those requirements to only ever check for this one variable will not change soon....

leamotta: Thanks for further explaining! Since there were no strong opinions and we don't face a situation where we're "forcing" partners to adopt Artsy Shipping for international, I can't think of a case in the near future that will require a more complex logic than the partner just opting them in manually. I'll merge this as is

leamotta: Thanks for further explaining! Since there were no strong opinions and we don't face a situation where we're "forcing" partners to adopt Artsy Shipping for international, I can't think of a case in the near future that will require a more complex logic than the partner just opting them in manually. I'll merge this as is

PR #9629, opened 2022-03-15
feat(article): roughs in analytics

dzucconi:
This PR ports over the tracking that exists in Reaction, as is.


src/v2/Apps/Article/useArticleTracking.ts
```diff
@@ -0,0 +1,68 @@
+import { useTracking } from "react-tracking"
```
dzucconi, line 1: @abhitip most of the tracking is here. See comments below.

src/v2/Apps/Article/useArticleTracking.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+
+    clickedPlayVideo: () => {
+      trackEvent({
+        action: "Click",
+        label: "Play video",
+      })
+    },
+  }
+}
```
dzucconi, line 68: (Turns out there's really not much.)

src/v2/Apps/Article/useArticleTracking.ts
abhitip: looks good!

PR #9630, opened 2022-03-15
fix(auction): Fix image resolution

damassi:
Fixes the image resolution on the main header image, and other weirdness with the bid modal image / active bids.
 
 

Screen Shot 2022-03-15 at 12 40 19 PM

### DIFF
--- a/src/v2/Apps/Auction/AuctionApp.tsx
+++ b/src/v2/Apps/Auction/AuctionApp.tsx
@@ -50,8 +50,8 @@ export const AuctionApp: React.FC<AuctionAppProps> = ({
         <AuctionMetaFragmentContainer sale={sale} />
 
         <Join separator={<Spacer my={4} />}>
-          {sale.coverImage?.cropped ? (
-            <FullBleedHeader src={sale.coverImage.cropped.src} />
+          {sale.coverImage?.url ? (
+            <FullBleedHeader src={sale.coverImage.url} />
           ) : (
             <Spacer my={2} />
           )}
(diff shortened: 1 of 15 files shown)


src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -113,9 +113,7 @@ export const AuctionAppFragmentContainer = createFragmentContainer(AuctionApp, {
 
       internalID
       coverImage {
-        cropped(width: 1800, height: 600, version: "normalized") {
-          src
-        }
+        url
```
dzucconi, line 116: This should be fine but I'd probably figure out what versions this image has and then specify 2 or 3 in size order

damassi: fixed 👍

src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -113,9 +113,7 @@ export const AuctionAppFragmentContainer = createFragmentContainer(AuctionApp, {
 
       internalID
       coverImage {
-        cropped(width: 1800, height: 600, version: "normalized") {
-          src
-        }
+        url(version: ["larger", "large"])
```
dzucconi, line 116: So I checked Gemini to see what versions we actually have:

```
gemini:staging> Template.where(key: 'sale').first.versions.map { |x| [x.key, x.width, x.height] }
=> [["large_rectangle", 640, 480], ["wide", 2000, nil], ["square", 230, 230], ["source", 1920, 1920]]
```

So looks like you want this to be `version: ["wide", "source", "large_rectangle"]`

dzucconi: (Also it's worth noting that none of these are really adequate; we'd probably want to add a normalized to this.)

damassi: Updated 👍

PR #9633, opened 2022-03-15
chore:update time stamp for auctions detail and add cascading interval label

nspinazz89:
image
This PR updates the copy around the start/end time time stamp on the auctions detail page as well as add a label for the cascading interval per [BID-207](https://artsyproduct.atlassian.net/browse/BID-207)

 It works in conjunction with this previously merged [Metaphysics PR.](https://github.com/artsy/metaphysics/pull/3862)


src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -74,6 +74,11 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
           )}
         </Flex>
       </Flex>
+      <Flex alignItems="center" justifyContent="space-between">
+        <Text variant="md" pr={2}>
+          {sale.auctionsDetailCascadingIntervalLabel}
```
bhoggard, line 79: Does it really make sense to generate this text in Metaphysics, when it's very simple and just refers to the time? Or are we doing this in case we change the units to minutes?

mzikherman (approved):

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
annacarey: We'll need it for Eigen, so I think it's nice to have the logic somewhere central. There's also [a little more logic](https://github.com/artsy/metaphysics/blob/02246518290d75ee3a6aa92585aebe119b279ba4/src/schema/v2/sale/index.ts#L251) than just showing the string so I'm a fan of keeping it in MP

src/v2/Apps/Auction/Components/AuctionDetails/__tests__/AuctionDetails.jest.tsx
```diff
@@ -57,10 +57,20 @@ describe("AuctionDetails", () => {
...
+  })
+
+  it("show the interval label", () => {
+    const wrapper = getWrapper({
+      Sale: () => ({
+        auctionsDetailCascadingIntervalLabel:
+          "Lots close in 1 minute intervals",
+      }),
+    })
+    expect(wrapper.text()).toContain("Lots close in 1 minute intervals")
```
annacarey, line 73: #minor but just realized this should read `1-minute intervals` with the `-`. no need to fix now in MP but maybe add a quick ticket so we don't forget?

annacarey (approved):
This looks great! We *may* want to have a designer give a quick once-over but only if they want to. I'll ask in channel but otherwise LGTM!

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
nspinazz89: yeah that's what I was going to say as well, there is logic for determining when to show it in metaphysics

src/v2/Apps/Auction/Components/AuctionDetails/__tests__/AuctionDetails.jest.tsx
nspinazz89: sounds good will do

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -99,7 +104,8 @@ export const AuctionDetailsFragmentContainer = createFragmentContainer(
 
         name
         slug
-        formattedStartDateTime
+        auctionsDetailFormattedStartDateTime
+        auctionsDetailCascadingIntervalLabel
```
dzucconi, line 108: Just a note: if you find yourself prefixing multiple fields like this it's probably worth nesting them inside of their own object?

Also this is very suspiciously close to naming a field after a UI element which is absolutely something we'd want to avoid. I'd be curious if there's a broader more accurate name for this concept?

dzucconi: Like: what's the difference between `formattedStartDateTime` and `auctionsDetailFormattedStartDateTime` — it's really not comprehensible if I'm just trying to read this code.

dzucconi: Do we want to deprecate the old field and this is the new one that should be used?

PR #9637, opened 2022-03-16
feat: Hide unknown medium text

rajsam003:
This PR resolves [CX-2413](https://artsyproduct.atlassian.net/browse/CX-2413)

Before
image

After
image

### DIFF
--- a/src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
+++ b/src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
@@ -179,9 +179,11 @@ const LargeAuctionItem: FC<Props> = props => {
               {title && date_text && ", "}
               {date_text}
             </Text>
-            <Text variant="xs" color="black60">
-              <Capitalize>{mediumText}</Capitalize>
-            </Text>
+            {mediumText !== "Unknown" && (
+              <Text variant="xs" color="black60">
+                <Capitalize>{mediumText}</Capitalize>
+              </Text>
+            )}
           </Box>
         </Flex>
       </Col>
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -616,15 +622,21 @@ const renderSmallCollapse = (props, user, mediator, filtersAtDefault) => {
         <Row mb={2}>
           <Col xs={4}>
             <Text variant="xs" fontWeight="bold">
-              Artwork Info
+              {categoryText !== "Unknown" && mediumText !== "Unknown"
+                ? "Artwork Info"
+                : "Artwork Dimension"}
```
rajsam003, line 627: Added the condition for the title as well. As if `categoryText` and `mediumText` both are unknown we will have only `dimension_text`

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -542,12 +544,16 @@ const renderLargeCollapse = (props, user, mediator, filtersAtDefault) => {
         <Row>
           <Col sm={2}>
             <Text variant="xs" fontWeight="bold">
-              Artwork Info
+              {categoryText !== "Unknown"
+                ? "Artwork Info"
+                : "Artwork Dimension"}
```
rajsam003, line 549: Added the condition for the title as well. As if `categoryText` is unknown we will have only dimension_text

MrSltun (approved):
LGTM! Great work Sam! 👏 👏

src/v2/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
MounirDhahri: I think this should be `dimensions` instead of `dimension`

MounirDhahri (commented):
Would you mind adding a test for this to make sure we don't mistakenly bring this back in the future

PR #9645, opened 2022-03-16
feat: Handle malformed unsubscribe requests GRO-864

jonallured:
This PR rounds out our support for the unsubscribe page by addressing malformed requests. There are two types:

* no auth token and not signed in
* params that look like this: `?authentication_token=valid-token?utm=oops`

For the former [I learned](https://artsy.slack.com/archives/C01ADJNCS5D/p1647378599865709?thread_ts=1647378480.341719&cid=C01ADJNCS5D) from @dzucconi that I should remove the `principalField` decoration from the query and then I could detect when the viewer isn't populated and run the prompt to sign in. This broke tests - more on that in a sec.

For the latter I copy/pasted and then adapted [the existing helper](https://github.com/artsy/force/blob/main/src/v2/Apps/Unsubscribe/UnsubscribeApp.tsx#L14-L17) function from the `UnsubscribeApp`. I also copy/pasted the tests.

Back to the tests that I broke - I spent about an hour trying and failing to properly mock the relay tests using this as a guide:

https://github.com/artsy/force/tree/0b291f005763e7c2600a5077786c9510bf655079/src/v2/DevTools

There's a deadline for this work so I made the call to mark the failing tests with `skip` and I promise I will come back to this when I'm not feeling pressure to ship. ❤️ 

<details>
<summary>some visual proof</summary>

Screen Shot 2022-03-16 at 11 16 14 AM
Screen Shot 2022-03-16 at 11 16 18 AM
Screen Shot 2022-03-16 at 11 16 21 AM
Screen Shot 2022-03-16 at 11 16 23 AM
Screen Shot 2022-03-16 at 11 16 26 AM
Screen Shot 2022-03-16 at 11 16 28 AM

</details>

https://artsyproduct.atlassian.net/browse/GRO-864

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/_Preferences2/PreferencesApp.tsx
+++ b/src/v2/Apps/_Preferences2/PreferencesApp.tsx
@@ -3,6 +3,7 @@ import {
   Checkbox,
   Column,
   GridColumns,
+  Message,
   Separator,
   Text,
   useToasts,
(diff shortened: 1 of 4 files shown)


jacobherrington (approved):
I can take a look at these tests too!

PR #9648, opened 2022-03-16
fix(auction): corrects display of tz in production

dzucconi:
I'm pretty surprised this is never come up since any SSR rendered page that relies on the X-TIMEZONE header is going to display UTC.

I suppose one way to fix this would be to set a cookie and defer to that on subsequent requests but the first request will always be UTC.


damassi (approved):
Looks good! Thanks for the quick fix and hopping on this @dzucconi  🙏

dzucconi:
Looks like your force push blew away my changes @damassi

PR #9655, opened 2022-03-16
fix(auction): Use artwork grid

damassi:
This adds grid support if we're on an auction/id page:

Screen Shot 2022-03-16 at 3 04 57 PM


src/v2/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -174,10 +174,10 @@ export class ArtworkGridContainer extends React.Component<
     const nodes = extractNodes(artworks)
 
     return (
-      <GridColumns>
-        {nodes.map((artwork, index) => {
+      <GridColumns width="100%">
```
dzucconi, line 177: As far as layout being compressed; this was your only problem; since it was a flex-child.

src/v2/Components/Artwork/FlatGridItem.tsx
```diff
@@ -0,0 +1,94 @@
...
+  FlatGridItem,
+  {
+    artwork: graphql`
+      fragment FlatGridItem_artwork on Artwork {
+        ...Metadata_artwork
+        ...SaveButton_artwork
+
+        internalID
+        title
+        image_title: imageTitle
```
damassi, line 79: all our random fixtures around force were complaining since this fragment is spread in to a parent that has fixture-based tests

PR #9664, opened 2022-03-17
feat(CX-2409): When the user has the SWA<>MYCollection feature enabled - show updated copy

dariakoko:

This PR solves [CX-2409]

### Description
Updated copy on SWA confirmation page to mention that submission is in MyC
When FF "swa_my_collection" is disabled   | When FF "swa_my_collection" is enabled
------------- | -------------
IMAGE 2022-03-17 14:02:19 | IMAGE 2022-03-17 14:39:54

Updated copy on SWA confirmation page to mention that submission is in MyC -daria

[CX-2408]: https://artsyproduct.atlassian.net/browse/CX-2408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CX-2409]: https://artsyproduct.atlassian.net/browse/CX-2409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
+++ b/src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
@@ -30,11 +30,12 @@ export const ThankYou: React.FC = () => {
       <Box maxWidth="720px" mt={4}>
         <Text variant="lg" color="black60">
           We will email you within 1-3 days to confirm if your artwork has been
-          accepted or not.
+          accepted or not. In the meantime your submission will appear in the
+          feature, My Collection, on the Artsy app.
         </Text>
         <Text variant="lg" mt={2} color="black60">
-          In the meantime your submission will appear in the feature, My
-          Collection, on the Artsy app.
+          With low fees, informed pricing, and multiple sales options, why not
+          submit another piece with Artsy.
         </Text>
       </Box>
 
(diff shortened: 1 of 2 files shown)


MounirDhahri (approved):
Great one, thanks for adding the tests

src/v2/Apps/Consign/Routes/SubmissionFlow/ThankYou/ThankYou.tsx
```diff
@@ -24,19 +26,40 @@ export const ThankYou: React.FC = () => {
...
-        <Text variant="lg">
-          We’ll email you within 1–3 business days to let you know the status of
-          your submission.
-        </Text>
-        <Text variant="lg" mt={2}>
-          In the meantime, feel free to submit another work—and benefit from
-          Artsy’s low fees, informed pricing, and multiple selling options.
-        </Text>
-      </Box>
+      {isSWAMCEnabled ? (
```
MounirDhahri, line 29: @dariakoko we should also be checking if the user is logged in since non logged in users have no collection

dariakoko: I don't think so. If the user is not logged in, the FF check is returning null

dariakoko: I'll double check

MounirDhahri: yeah, that's true actually. I didn't think about that. No action is needed then. I would recommend though for future work to check if the user is signed in or not explicitly. That's because once we would like to remove this feature flag, we will have to be careful about the user login status

PR #9667, opened 2022-03-17
fix(auction): adjusts pagination count

dzucconi:
Closes: [GRO-893](https://artsyproduct.atlassian.net/browse/GRO-893)


dzucconi:
There's a spec update in here too but apparently the UI isn't reflecting it.

src/v2/Apps/Auction/Components/AuctionArtworkFilter.tsx
```diff
@@ -99,7 +99,7 @@ export const getArtworkFilterInputArgs = (user?: User) => {
   // Shared with auctionRoutes
   return {
     aggregations,
-    first: 10,
+    first: 39,
```
jonallured, line 102: Ah darn somehow I overlooked this - nice one 👍

PR #9671, opened 2022-03-17
fix(GRO-892): Redirect from the old /sale path to /auction

jacobherrington:
The type of this PR is: **Bugfix** (kinda)

This PR solves [GRO-892]

### Description

I'm not 100% certain this is the preferred way to accomplish this with our router, but it does seem to work! 😆 

cc @damon is this fine to do?

[GRO-892]: https://artsyproduct.atlassian.net/browse/GRO-892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Auction/auctionRoutes.tsx
+++ b/src/v2/Apps/Auction/auctionRoutes.tsx
@@ -1,4 +1,5 @@
 import loadable from "@loadable/component"
+import { Redirect } from "found"
 import { graphql } from "relay-runtime"
 import { getInitialFilterState } from "v2/Components/ArtworkFilter/Utils/getInitialFilterState"
 import { AppRouteConfig } from "v2/System/Router/Route"
@@ -165,6 +166,16 @@ export const auctionRoutes: AppRouteConfig[] = [
         }
       }
     `,
+  },
+  {
+    // Redirect from the old route to the new one
+    path: "/sale/:slug?",
+    children: [
+      new Redirect({
+        from: "/",
+        to: "/auction/:slug?",
+      }) as any,
+    ],
   },
 ]
 


src/v2/Apps/Auction/auctionRoutes.tsx
```diff
@@ -48,6 +49,16 @@ const AuctionFAQRoute = loadable(
...
 
 export const auctionRoutes: AppRouteConfig[] = [
+  {
+    // Redirect from the old route to the new one
+    path: "/sale/:slug?",
+    children: [
+      new Redirect({
+        from: "/",
+        to: "/auction/:slug?",
+      }) as any,
```
dzucconi, line 59: https://4catalyzer.github.io/found/advanced/redirects

Shouldn't this be `from: "/sale/:slug?",`

damassi: I tried this yesterday before Jenna gave us the go ahead to just remove /sale and it didn't quite work. I bet this does tho...

src/v2/Apps/Auction/auctionRoutes.tsx
```diff
@@ -48,6 +49,16 @@ const AuctionFAQRoute = loadable(
...
+    ],
+  },
```
damassi, line 61: Can we move this new code to the bottom of the route array since it will rarely be touched

src/v2/Apps/Auction/auctionRoutes.tsx
jacobherrington: Based on how I'm reading it, no, but maybe! I haven't worked with the router before, but glad to check that and change it if it works.

src/v2/Apps/Auction/auctionRoutes.tsx
jacobherrington: Done!

src/v2/Apps/Auction/auctionRoutes.tsx
dzucconi: I'm confused why would be redirecting from the root.

jacobherrington: I saw this method of redirecting in orderRoutes and consignRoutes:

https://github.com/artsy/force/blob/9e9af2114ff29b0565e10b88a80e9e081afec954/src/v2/Apps/Order/orderRoutes.tsx#L305-L309

It seems a little odd to me, but it was the only way I found to achieve this with my limited knowledge of Force internals.

I don't think it's redirecting from the root, but more like the root of the path that it's nested under, but I might be wrong!

dzucconi: So does specifying it as `sale/:slug` simply not work? What about `*`?

dzucconi: But sure, I get what you're saying.

dzucconi (approved):

jacobherrington:
Rebased to turn CI green

PR #141, opened 2016-09-13
@broskoski => Responsive content differentiation pages

kanaabe:
Closes https://github.com/artsy/force/issues/116

Full width (looks the same):
image

Medium width:
responsive

Mobile width:
image
responsive-mobile

PR #9675, opened 2022-03-17
notify #dev on failed prod deploys

ovasdi:

### Description

Action item from incident review (#141). Notify engineering when a force deploy fails.

Looking at #practice-web channel, I do not see any post-steps logging, only the slack notifications from the [validate_production_schema](https://github.com/artsy/force/blob/main/.circleci/config.yml#L33). I wonder if using channel IDs would help as the configuration seems correct to me.

The slack app _should_ be able to send notifications to #dev as it is authorized to do so:

Screen Shot 2022-03-17 at 3 32 49 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -414,9 +414,10 @@ workflows:
             - horizon/block
             - validate_production_schema
           post-steps:
+          # notify: practice-web, dev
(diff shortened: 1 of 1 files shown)


damassi:
ty!

PR #9679, opened 2022-03-17
fix: unleash artist filter experiment bug

mc-jones:

### Description
This PR fixes a bug in the filter expansion experiment by ensuring we are checking against the current page type, which in this use case should be `Artists`. ~~**It is meant to be explanatory, and not merged**, as it demonstrates how to fix the bug, but doesn't address the issue of if we are sending the right data with the analytics (I didn't get time to look into this and am still not sure if the observed behavior with `useAnalyticsContext` is a bug, or it is not being implemented properly).~~ 

Under the previous implementation, I was seeing `contextOwnerType` was always set to the entry point page for the user. For example. the value of `contextPageOwner` was `home` if you entered from `artsy.net/`, even when navigating to other pages.  This resulted in us only seeing the expected behavior of an expanded filter when the `user` (including unauthenticated users) hard reloaded the artist page. 

```js 
const isArtistPage = contextPageOwnerType === OwnerType.artist // only worked when user entered app at the Artist page
```

### Related issue 

[In the conversation that prompted us to look at this](https://artsy.slack.com/archives/CA8SANW3W/p1647537193559959), it was also mentioned that it seemed like `Unleash` was not sending the correct data back to `Force`. This seems to have actually been related to a combination of the strategy configuration and how stickiness works in `Unleash`. After removing the Hosts strategy, and using the `default` stickiness for the gradual rollout, we saw the expected behavior. To test: 

1. Navigate to `staging.artsy.net` in an incognito window and DON'T sign in (if you test this while authenticated, it will use your userId as the stickiness factor since that gets priority when set to `default`, and serve you the same outcome each time. If you want to test with an authenticated user, you can set the stickiness value to `random`.)
2. Inspect `sd.FEATURE_FLAGS["filters-expanded-experiment"]` in the dev tools console to see which group you're in (I.e control or experiment)
3. Close all incognito windows. (If you just open a new incognito window without closing the session your stickiness factor may not change and you will see the same outcome. multiple incognito windows will use the same session id).
4. Repeat until you feel comfortable its about a 50/50 split (know this isn't great)

Here is why I think you were seeing such odd results: 

Its not really clear, and [hidden in the docs](https://docs.getunleash.io/user_guide/activation_strategy#gradual-rollout), how `Unleash` handles stickiness when it is set to `UserId`. If you set `stickiness` by `userId` then any unauthenticated session will always come back false since the userId will be `null`. This is likely why you were seeing the flag and the variant come back as `false` until you added the `Host` strategy. When you added the `Host` strategy, you were essentially overriding the gradual rollout strategy because the Host strategy condition would always be true, and wasn't constrained by `stickiness`.


src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
```diff
@@ -28,23 +27,21 @@ interface ArtworkFiltersProps {
...
-    contextPageOwnerType,
-  } = useAnalyticsContext()
-  const { match } = useRouter()
+  let currentPage
 
-  const isArtistPage = match.location.pathname.slice(1, 7) === OwnerType.artist
+  if (typeof window !== "undefined") {
+    const { pageType } = getContextPageFromClient()
+    currentPage = pageType
+  }
```
dzucconi, line 35: If we're going to leave this here then I'd at least add a `TODO` or `HACK` comment regarding the why of it and maybe a note to switch to checking the route.

Mocking useRouter is easy to do however if that's really the blocker.

gkartalis: @mc-jones exactly! Also this is the [sample unleash doc](https://github.com/artsy/force/blob/main/docs/unleash_a-b_testing.md) file that I added that we can update with our learnings etc

src/v2/System/useFeatureFlag.tsx
```diff
@@ -60,15 +60,14 @@ export function useTrackVariantView({
...
+    if (
+      typeof window !== "undefined" &&
+      typeof window.analytics !== "undefined"
+    ) {
       const trackFeatureView = shouldTrack(experimentName, variantName)
-
       if (trackFeatureView && shouldTrackExperiment) {
-        trackEvent({
-          action: ActionType.experimentViewed,
+        window.analytics.track(ActionType.experimentViewed, {
```
gkartalis, line 70: After a quick QA round we found out that `trackEvent` wasn't behaving as expected and was almost never called. Replacing that with `window.analytics.track` fixed the issue and now analytics are being tracked 100% of the time.

Let me know what you think about this

dzucconi: Would very much be interested in knowing why?

dzucconi: cc @damassi

gkartalis: I would be really interested in knowing that as well not 100% sure the why but it wasn't called when expected, happy to pair on this if you have any other suggestion.

gkartalis: When pulling the trackEvent on a place like a button and we pressed it it was tracking as expected, but inside the `useEffect` there it is not working and I really don't know the why

dzucconi: I'd at least leave a comment explaining this. This is the kind of thing I'd immediately refactor to use trackEvent if I came across it.

gkartalis: done! 👍

mc-jones: I've [ticketed out the follow-up](https://artsyproduct.atlassian.net/browse/PLATFORM-4059) to take another look at the analytics approach. Do we want to move forward with this approach to continue testing, while we continue to look at the analytics question separately?

src/v2/System/useFeatureFlag.tsx
```diff
@@ -60,15 +60,18 @@ export function useTrackVariantView({
...
   useEffect(() => {
-    if (typeof window !== "undefined") {
-      const trackFeatureView = shouldTrack(experimentName, variantName)
-
+    if (
+      typeof window !== "undefined" &&
+      typeof window.analytics !== "undefined"
+    ) {
+      const trackFeatureView =
+        shouldTrackExperiment && shouldTrack(experimentName, variantName)
```
gkartalis, line 69: Added this extra check here to prevent saving the experiment in localstorage when the user is in a page that is not showing the experiment (eg. in our case we want the experiment to be displayed in the artist page although without this check if a user visited a page with an artwork grid the tracking event wasnt dispatched but we saved the experiment in the local storage which caused the inability to dispatch the tracking event when the user viewed the page with the experiment)

gkartalis (dismissed):

src/v2/Components/ArtworkFilter/ArtworkFilters/index.tsx
```diff
@@ -21,6 +27,26 @@ interface ArtworkFiltersProps {
 // Some filters will be rendered only if there is the necessary data in aggregations (for example, ArtistsFilter)
 export const ArtworkFilters: React.FC<ArtworkFiltersProps> = props => {
   const { user, relayEnvironment } = props
+  const { match } = useRouter()
+
+  const isArtistPage = match.location.pathname.includes("/artist/")
+
+  const variant = useFeatureVariant("filters-expanded-experiment")
+
+  const { trackVariantView } = useTrackVariantView({
```
damassi, line 36: (blocking): If we have a hook called `useFeatureVariant` we should keep the language 1:1 with other related functions -- `useTrackVariantView` -> `useTrackFeatureVariant`. 

"View" also presumes a few different things when really what we want here is something generic.

PR #9684, opened 2022-03-18
fix(analytics): prevents double pageview events

dzucconi:

### Description

* https://github.com/artsy/palette/pull/1143 - fixes the `/works-for-you` issue
* 0dc5b57028b2a4acd8aebe932b0679cf210020cd - fixes the `/art-fairs` issue by just directly linking to the correct route.

I removed the redirect here since we already have one here: https://github.com/artsy/force/blob/bb2c444e9f555fc58817782782df2eb6232d7ea4/src/lib/middleware/hardcodedRedirects.ts#L29 — also it was a `302` not a `301` which it should be.


jonallured:
tyyyy 🏆

PR #9687, opened 2022-03-18
fix:Update active bids display to respect closed sales or cascading lots

bhoggard:
The type of this PR is: fix

The Active Bids view on the new auction page didn't respect closed sales, and also needed to be updated for cascading lots.

This PR solves [BID-256].

## Normal sale before -- this PR hides the Active Bids display when the sale is closed.

Screen Shot 2022-03-18 at 3 23 47 PM

## Sale with cascading end times

Screen Shot 2022-03-18 at 3 17 22 PM

[BID-256]: https://artsyproduct.atlassian.net/browse/BID-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -29,14 +29,14 @@ export const AuctionApp: React.FC<AuctionAppProps> = ({
 }) => {
   const { contextPageOwnerType, contextPageOwnerSlug } = useAnalyticsContext()
 
+  const showLotStandings = me?.showLotStandingsTab?.length && !sale.isClosed
+
   const tabBar = {
     isVisible:
-      sale.showAssociatedSale ||
```
damassi, line 34: Why remove this line?

src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -29,14 +29,14 @@ export const AuctionApp: React.FC<AuctionAppProps> = ({
...
-      me?.showLotStandingsTab?.length ||
+      showLotStandings ||
       viewer.showFollowedArtistsTab?.edges?.length ||
       sale.showBuyNowTab,
-    showAssociatedSale: sale.showAssociatedSale,
```
damassi, line 38: And this?

damassi (changes requested):
Noticed that `associatedSales` tab logic has been (accidentally?) removed

src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -29,14 +29,14 @@ export const AuctionApp: React.FC<AuctionAppProps> = ({
 }) => {
   const { contextPageOwnerType, contextPageOwnerSlug } = useAnalyticsContext()
 
+  const showLotStandings = me?.showLotStandingsTab?.length && !sale.isClosed
+
   const tabBar = {
     isVisible:
-      sale.showAssociatedSale ||
-      me?.showLotStandingsTab?.length ||
```
damassi, line 35: If we're no longer using this aliased field, it can be removed from the fragment query

src/v2/Apps/Auction/AuctionApp.tsx
```diff
@@ -29,14 +29,14 @@ export const AuctionApp: React.FC<AuctionAppProps> = ({
 }) => {
   const { contextPageOwnerType, contextPageOwnerSlug } = useAnalyticsContext()
 
+  const showLotStandings = me?.showLotStandingsTab?.length && !sale.isClosed
```
damassi, line 32: We can rename the aliased field below in the query from `showLotStandings` to `showActiveBids` to better reflect the UI.

And we'll also want to update [the test](https://github.com/artsy/force/blob/main/src/v2/Apps/Auction/__tests__/AuctionApp.jest.tsx#L143) to check for this condition.

src/v2/Apps/Auction/Components/AuctionActiveBids.tsx
```diff
@@ -236,7 +237,7 @@ const BidButton: React.FC<
         router.push(href)
       }}
     >
-      Increase Bid
+      {lotStanding.saleArtwork?.endedAt ? "Bidding closed" : "Increase Bid"}
```
damassi, line 240: We'll want to update the [`AuctionActiveBids` test](src/v2/Apps/Auction/Components/__tests__/AuctionActiveBids.jest.tsx) to reflect this new behavior.

src/v2/Apps/Auction/AuctionApp.tsx
annacarey: I'd also rename the variable to `showActiveBids` then

annacarey: Confirming that `isClosed` will only be false on `sale` once the last lot is closed (if the sale has cascading end times). Pretty sure yes, but just making sure `isClosed` doesnt check if sale's `end_at` field is in the past because the lots could still be open if that's the case.

src/v2/Apps/Auction/Components/AuctionActiveBids.tsx
annacarey: You also may want to check with design. From the screenshots it looks like the two buttons are slightly different sizes based on which text is displaced which is slightly awkward. Also maybe it's odd to have "Bidding closed" displayed on a button as it's not really a CTA?

annacarey (changes requested):
Looks like Chris is interested in some tests and I think you may want to check w/ design/Joh if you haven't on the button.

src/v2/Apps/Auction/AuctionApp.tsx
bhoggard: Sorry, rebase problems. I will fix.

damassi:
@annacarey - note that the design here was kinda "wing it in the middle of the night" style -- the idea was to get things out and then iterate. So wouldn't hold the PR up here for that, but rather get the logical / naming things addressed and then follow-up with design. (Note: the design team has already started pondering what the next visual iteration of this page will be.)

bhoggard:
Updated button sizes

Screen Shot 2022-03-18 at 5 04 54 PM

src/v2/Apps/Auction/Components/AuctionActiveBids.tsx
bhoggard: I added a test

src/v2/Apps/Auction/Components/__tests__/AuctionActiveBids.jest.tsx
```diff
@@ -109,6 +109,36 @@ describe("AuctionActiveBids", () => {
...
+                  },
+                  endedAt: null,
+                },
+              },
+            ],
+          }),
+        })
+
+        expect(wrapper.text()).toContain("Bidding Closed")
+        expect(wrapper.text()).toContain("Increase Bid")
```
nspinazz89, line 138: a single view is supposed to have Bidding closed and increase bid?

nspinazz89 (commented):
Looks good just one question

damassi (approved):
LGTM! One very nitpicky comment but totally free to ignore

src/v2/Apps/Auction/Components/__tests__/AuctionActiveBids.jest.tsx
bhoggard: It has 2 lots in the view.

PR #9704, opened 2022-03-22
chore: empty commit

dzucconi:
🕳️


jonallured:
Our theory is that causing assets to rebuild will fix a problem we're only seeing in staging:

https://artsy.slack.com/archives/C01ADJNCS5D/p1647962153021399?thread_ts=1647960456.291119&cid=C01ADJNCS5D

PR #9708, opened 2022-03-22
refactor(GRO-827): entity header fragment containers

dzucconi:
Closes [GRO-827](https://artsyproduct.atlassian.net/browse/GRO-827)

### Description

Here I begin the process of deprecating `EntityHeader` from Palette, in favor of replacing it with a number of fragment container-based entity headers for individual entities.

This PR should cover most of them. I started the process of replacing existing code with the new fragment-based components though there's a lot of ground to cover. Will follow up by deprecating the Palette component.

https://user-images.githubusercontent.com/112297/159542977-ad69a191-f732-4bbb-a68a-b77199399122.mov


.storybook/preview.js
```diff
@@ -1,17 +1,26 @@
...
 import { MediaContextProvider } from "v2/Utils/Responsive"
+import { createRelaySSREnvironment } from "v2/System/Relay/createRelaySSREnvironment"
+import { SystemContextProvider } from "v2/System"
 
 const { GlobalStyles } = injectGlobalStyles()
 
+const relayEnvironment = createRelaySSREnvironment({
+  metaphysicsEndpoint: "https://metaphysics-staging.artsy.net/v2",
+  cache: JSON.parse(window.__RELAY_BOOTSTRAP__ || "{}"),
+})
```
dzucconi, line 11: cc @damassi

.storybook/preview.js
```diff
@@ -1,17 +1,26 @@
...
 export const decorators = [
   Story => {
     return (
-      <MediaContextProvider>
-        <ThemeProviderV3>
-          <GlobalStyles />
-          <Story />
-        </ThemeProviderV3>
-      </MediaContextProvider>
+      <SystemContextProvider relayEnvironment={relayEnvironment}>
```
damassi, line 16: 👍

.storybook/preview.js
damassi: ahh _continues reading_

damassi (approved):
This looks great, and I think the variance in these headers warrants bringing them into force, but please add some tests light display tests to lock these things down. Would take all of 20 minutes to add coverage and would prevent things from unintentionally breaking in the future.

dzucconi:
In an ideal world we'd just have visual diffing specs on the stories I wrote; which cover every usecase.

damassi:
> In an ideal world we'd just have visual diffing specs on the stories I wrote; which cover every usecase.

I agree this is procedurally simpler, but would it work at scale in an application codebase that isn't a library and on a large team? Without a mechanism to enforce this sort of thing and the willingness to advocate for a process change such as this, we'd end up with more tooling to accommodate and patchy coverage as some devs would be aware and others not. Those that are not might make a change and then just batch-accept the visual change.

Simple display-based unit tests are baseline, easy to write, have durability through time, point to changes made in the code via errors that generally emit stack traces, and more. 

In practice, with tests, everyday product devs will not wantonly delete a test to accomodate a change they made to the source code. They will look at their change and the test and think about what it is that's going on. They're forced to. Not so with other models loosely applied.

dzucconi:
Not just simpler: better.

These components are simple with barely any logic. Aside from some minor display logic in the partner one; there's really no functionality. A couple of switches. The tests I'll write really don't offer much in the way of safety. Visual diffing specs that request data from the API would on the other hand.

The tests won't offer any guidance when refactoring or working on these components. The stories will.

(I'm writing specs for these of course. I'm just saying: given two options I'd take the stories any day.)

damassi:
The PR _is_ approved tho, so if you do think that tests aren't valuable here from a wider team vantage point in the future that's also totally fine! I'm just constantly trying to think from the perspective of devs who are stumbling by, versus super sr engineers, and haven't been able to come up with any better kind of insurance. (I know we go back and forth on this, its just tests are super easy to write and just take a sec, so why not)

src/v2/Components/EntityHeaders/__tests__/EntityHeaderPartner.jest.tsx
```diff
@@ -0,0 +1,55 @@
...
+        Partner: () => ({
+          name: "Example Partner",
+          initials: "EP",
+        }),
+      })
+
+      expect(screen.getByText("EP")).toBeInTheDocument()
+    })
+  })
+})
```
damassi, line 55: (ty)

PR #9710, opened 2022-03-22
fix(types): fixes type-errors

dzucconi:
Feels like bailing out a sinking ship at this point.


dzucconi:
https://github.com/artsy/force/pull/9711 fixes this

PR #9713, opened 2022-03-22
feat:adds cascading end times timer to sale detail page

nspinazz89:
[BID-207] The type of this PR is: feature

This PR adds a new timer component to the sale detail view per [BID-207]. It also has some small copy updates to Sale Detail page. 

This is still somewhat a WIP as I need to swap out moment for luxon according to some warnings from our commit hooks, but I wanted to start gathering feedback.

[BID-207]: https://artsyproduct.atlassian.net/browse/BID-207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BID-207]: https://artsyproduct.atlassian.net/browse/BID-207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Before Sale Starts
image

Sale has Started
image

Less than 24 Hours till end 
image

Less than 1 Hour till end
image

Lots are closing
image

All Lots have closed
image


src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,110 @@
...
+)
+
+export const getTimerCopy = (time, startAt, endAt, endedAt) => {
+  const { days, hours, minutes, seconds } = time
+
+  const parsedDays = parseInt(days, 10)
+  const parsedHours = parseInt(hours, 10)
+  const parsedMinutes = parseInt(minutes, 10)
+  const parsedSeconds = parseInt(seconds, 10)
+  const thisMoment = moment()
```
annacarey, line 65: Did you try using the `useTimer` hook for all of this logic? I don't think you need to use the `moment` library at all here. `useTimer` can provide a value for `hasStarted` that you can use instead. The hook handles the "current time" concept for you.

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // 1-24 hours until close
+    else if (parsedDays < 1 && parsedHours >= 1) {
+      copy = `${parsedHours}h ${parsedMinutes}m Until Bidding Ends`
+      color = "red100"
+    }
+  }
+  console.log(copy)
+  console.log(color)
```
annacarey, line 107: Should be deleted :)

src/v2/Components/__tests__/SaleDetailTimer.jest.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+require("v2/Utils/getCurrentTimeAsIsoString").__setCurrentTime(
+  "2022-03-21T20:22:32.000Z"
+)
+
+describe("getTimerCopy", () => {
+  describe("when the sale is open", () => {
+    describe("when the close date/time is more than 24 hrs before closing", () => {
+      const time = { days: "1", hours: "4", minutes: "00", seconds: "00" }
+      const startAt = moment().subtract(1, "hours")
```
annacarey, line 16: I find these a lot more readable if you feed it a specific date and not a moment calculated date. But maybe that's just me?

src/v2/Components/__tests__/SaleDetailTimer.jest.tsx
```diff
@@ -0,0 +1,116 @@
+import { getTimerCopy } from "../SaleDetailTimer"
+import moment from "moment"
+
+jest.useFakeTimers()
```
annacarey, line 4: Do you need this?

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const startAt = sale?.startAt
+
+  const endedAt = sale?.endedAt
+
+  const { time } = useTimer(endAt!, startAt!)
+
+  if (!endAt) {
+    return null
+  }
```
annacarey, line 32: #minor but you can move up to line 23

src/v2/Components/__tests__/SaleDetailTimer.jest.tsx
```diff
@@ -0,0 +1,116 @@
...
+          "Bidding Starts Today"
+        )
+        expect(getTimerCopy(time, startAt, endAt, endedAt).color).toEqual(
+          "blue100"
+        )
+      })
+    })
+  })
+
+  describe("sale has closed", () => {
```
annacarey, line 107: We want to display an empty string when the sale as closed? Does it work to just not render the component if the sale is closed?

annacarey (changes requested):
Left some questions/comments. Mainly interested in dropping the use of `moment` for the `useTimer` hook logic. Otherwise this is super close! Nice work.

src/v2/Components/SaleDetailTimer.tsx
nspinazz89: I tried that early on and struggled with it, but I see what I did wrong and will be able to refactor without using those. Good call!

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,110 @@
...
+    // 1-24 hours until close
+    else if (parsedDays < 1 && parsedHours >= 1) {
+      copy = `${parsedHours}h ${parsedMinutes}m Until Bidding Ends`
+      color = "red100"
+    }
+  }
+  console.log(copy)
+  console.log(color)
+
+  return { copy, color }
```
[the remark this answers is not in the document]
nspinazz89: Will do!

src/v2/Components/SaleDetailTimer.tsx
nspinazz89: You can't actually because of the hook to useTimer, you'll get a linting error if you try and move it up ` React Hook "useTimer" is called conditionally.`

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { createFragmentContainer, graphql } from "react-relay"
+import * as React from "react"
+import { Flex, Text } from "@artsy/palette"
+import { useTimer } from "v2/Utils/Hooks/useTimer"
+
+export interface SaleDetailTimerProps {
+  sale: SaleDetailTimer_sale
+}
+
+function relativeDaysUntil(parsedDays, suffix) {
```
mzikherman, line 11: Let's move this helper lower down in the file. It's more customary for the exported objects near the top of the file to be the actual React component.

annacarey (dismissed):

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,101 @@
...
+    return `${suffix} Today`
+  } else if (parsedDays === 1) {
+    return `${parsedDays} Day Until ${suffix}`
+  } else {
+    return `${parsedDays} Days Until ${suffix}`
+  }
+}
+
+export const SaleDetailTimer: React.FC<SaleDetailTimerProps> = ({ sale }) => {
+  const endAt = sale?.endAt
```
mzikherman, line 22: #minor lots of newlines around these

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -44,7 +45,11 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
       </GridColumns>
 
       <Spacer my={4} />
-
+      <Flex alignItems="center" justifyContent="space-between">
+        <Flex alignItems="center">
+          <SaleDetailTimerFragmentContainer sale={sale} />
```
mzikherman, line 50: Is it right to render this on all sales (even w/o cascading)? It feels like it might be odd to render this above the below `AuctionDetailsStartTimeQueryRenderer` (two timers)?

Relatedly- pls update the PR description w/ screenshots of the changes!

src/v2/Components/SaleDetailTimer.tsx
```diff
@@ -0,0 +1,101 @@
...
+  let copy = ""
+  let color = "blue100"
+
+  // Sale has not yet started
+  if (!hasStarted) {
+    copy = relativeDaysUntil(parsedDays, "Bidding Starts")
+  }
+  // Lots are closing
+  else if (hasEnded) {
+    copy = "Lots are closing"
```
mzikherman, line 75: This feels possibly inaccurate, when the sale is already over we display 'Lots are closing'? I would have expected to maybe not have this rendered at all and it just displays 'Sale Closed ....' w/ the date (or something like that).

src/v2/Components/__tests__/SaleDetailTimer.jest.tsx
```diff
@@ -0,0 +1,85 @@
...
+        expect(getTimerCopy(time, hasStarted, hasEnded).copy).toEqual(
+          "Lots are closing"
+        )
+        expect(getTimerCopy(time, hasStarted, hasEnded).color).toEqual("red100")
+      })
+    })
+  })
+
+  describe("when the sale is not yet open", () => {
+    describe("when the sale has is more than 1 day away", () => {
```
mzikherman, line 57: #minor "when the sale has is" worded awkwardly

src/v2/Components/__tests__/SaleDetailTimer.jest.tsx
```diff
@@ -0,0 +1,85 @@
...
+        expect(getTimerCopy(time, hasStarted, hasEnded).copy).toEqual(
+          "3 Days Until Bidding Starts"
+        )
+        expect(getTimerCopy(time, hasStarted, hasEnded).color).toEqual(
+          "blue100"
+        )
+      })
+    })
+
+    describe("when the sale has is less than 1 day away", () => {
```
mzikherman, line 71: similar, "has is" awkward wording

mzikherman (commented):
I had a few clarifying questions + request for screenshots for the different states!

src/v2/Components/SaleDetailTimer.tsx
nspinazz89: so 'hasEnded' in this context is the sales endAt which corresponds to the beginning of closing, not the sale's endedAt, that is used on line 39 to not render the text if we have a value for endedAt.

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -15,6 +15,7 @@ import { AuctionDetails_sale } from "v2/__generated__/AuctionDetails_sale.graphq
 import { AuctionDetails_me } from "v2/__generated__/AuctionDetails_me.graphql"
 import { AuctionInfoSidebarFragmentContainer } from "./AuctionInfoSidebar"
 import { RegisterButtonFragmentContainer } from "../RegisterButton"
+import { SaleDetailTimerFragmentContainer } from "v2/Components/SaleDetailTimer"
```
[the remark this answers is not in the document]
nspinazz89: I created a ticket to consolidate the SaleDetailTimer with @annacarey's LotTimer, so I'd advocate for keeping it separate for now with the expectation that both will be refactored in to a single component in the near future.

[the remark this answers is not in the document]
annacarey: I didn't mean the entire `LotTimer` component, I meant the `getTimerCopy` helper. I agree with @mzikherman, that this should be left in the same component.

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -65,6 +70,12 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
         </Flex>
       </Flex>
 
+      <Flex alignItems="center" justifyContent="space-between">
+        <Text variant="md" pr={2}>
+          {sale.cascadingEndTime?.intervalLabel}
```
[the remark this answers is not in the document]
annacarey: This is also related to @mzikherman's above comment. Shouldnt this not show if the sale doesn't have a `cascadingEndTime` set?

src/v2/Components/SaleDetailTimer.tsx
annacarey: Maybe `lotsAreClosing` should be the name of the argument?

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -44,7 +45,13 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
...
 
       <Spacer my={4} />
-
+      <Flex alignItems="center" justifyContent="space-between">
+        <Flex alignItems="center">
+          {!!sale.cascadingEndTimeInterval && (
+            <SaleDetailTimerFragmentContainer sale={sale} />
+          )}
+        </Flex>
+      </Flex>
```
damassi, line 54: Can we get away with just 

```tsx
{!!sale.cascadingEndTimeInterval && (
  <SaleDetailTimerFragmentContainer sale={sale} />
```

with none of the flex containers? They don't seem necessary, or at least the second one can be omitted.

damassi: Also, we'd want to put the guard on the outside so that if `cascadingEndTimeInterval` is falsy then the outer container don't render and embed empty divs

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -65,6 +72,14 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
         </Flex>
       </Flex>
 
+      <Flex alignItems="center" justifyContent="space-between">
+        <Text variant="md" pr={2}>
+          {!!sale.cascadingEndTimeInterval &&
+            !sale.endedAt &&
+            sale.cascadingEndTime?.intervalLabel}
+        </Text>
+      </Flex>
```
damassi, line 81: Same here, don't think the flex container is necessary since all we're doing is rendering out one text label. Can we omit it completely? And if not, we'll want to 
- store the label in a variable so that its a bit more readable for the future
```
const cascadingEndTimeIntervalLabel = !!sale.cascadingEndTimeInterval && !sale.endedAt && sale.cascadingEndTime?.intervalLabel
```
and then later, use it as the outermost bit to guard against empty divs:

```tsx
{cascadingEndTimeIntervalLabel && 
  <Text variant="md" pr={2}>{cascadingEndTimeIntervalLabel}</Text?
}
```

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
nspinazz89: things start getting a little wonky without the flex boxes 
image
I'm sure there is a way around it, that being said front end is definitely not my forte.

nspinazz89: actually was able to get away with removing one.

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
nspinazz89: that works! Good call!

src/v2/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
damassi: It might be worth adding `textAlign='left'` on the `<Text>` component itself, but if that doesn't work then all good 👍

mzikherman (approved):
Looks good!

I think we can probably skip a review app, but let's mocktion (a cascading and a regular) on staging soon after merging.

PR #9714, opened 2022-03-23
feat: change copies from input field in Offer

araujobarret:

This PR solves [NX-3111]

### Description

Changes the copies from the low amount warning in the Offer. Slightly adjust the design removing underline decoration and not breaking the line anymore/

Price
Screenshot 2022-03-23 at 10 40 09

Price Range
Screenshot 2022-03-23 at 10 51 51

@artsy/negotiate-devs 

[NX-3111]: https://artsyproduct.atlassian.net/browse/NX-3111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Components/MinPriceWarning.tsx
+++ b/src/v2/Apps/Order/Components/MinPriceWarning.tsx
@@ -22,19 +22,15 @@ export const MinPriceWarning: React.FC<MinPriceWarningProps> = ({
       flow: AnalyticsSchema.Flow.MakeOffer,
       order_id: orderID,
     })
-  }, [])
+  }, [orderID, tracking])
 
   return (
     <Clickable cursor="pointer" onClick={onClick} mt={2}>
       <Message variant="warning" p={2}>
         <Text>
-          Galleries usually accept offers within
           {isPriceRange
-            ? " the displayed price range"
-            : " 20% of the listed price"}
-          ; any lower is likely to be rejected.
-        </Text>
-        <Text style={{ textDecoration: "underline" }}>
+            ? "Offers lower than the displayed price range are often declined. "
+            : "Offers less than 20% of the list price are often declined. "}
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Order/Components/MinPriceWarning.tsx
```diff
@@ -22,19 +22,15 @@ export const MinPriceWarning: React.FC<MinPriceWarningProps> = ({
       flow: AnalyticsSchema.Flow.MakeOffer,
       order_id: orderID,
     })
-  }, [])
+  }, [orderID, tracking])
```
araujobarret, line 25: nit: added missing deps from the hook, checked and it's not triggered more than once

src/v2/Apps/Order/Components/MinPriceWarning.tsx
```diff
@@ -22,19 +22,15 @@ export const MinPriceWarning: React.FC<MinPriceWarningProps> = ({
...
 
   return (
     <Clickable cursor="pointer" onClick={onClick} mt={2}>
       <Message variant="warning" p={2}>
         <Text>
-          Galleries usually accept offers within
           {isPriceRange
```
lean8086, line 31: Strange that we don't test both of these use cases.

lean8086 (dismissed):
LGTM!

src/v2/Apps/Order/Components/MinPriceWarning.tsx
araujobarret: Actually, they were in a different component, fixed and improved in this [commit](https://github.com/artsy/force/pull/9714/commits/f5053a9051b2afb65493b665acdef18564cde3ca)

PR #9715, opened 2022-03-23
feat(NOSF2): change order of options starting with the exact price

tam-kis:
This PR solves: [NX-3109](https://artsyproduct.atlassian.net/browse/NX-3109)

Reversed order of price options for exact price. First option is now replaced with the exact price.

Screenshot 2022-03-23 at 12 13 53

### DIFF
--- a/src/v2/Apps/Order/Components/PriceOptions.tsx
+++ b/src/v2/Apps/Order/Components/PriceOptions.tsx
@@ -96,17 +96,26 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
     }))
   }
 
+  const fixedPrice = Number(listPrice?.major)
+
+  const percentagePrices = [
+    { value: fixedPrice, description: "Exact price" },
+    {
+      value: Number([Math.round(fixedPrice * (1 - 0.1))]),
+      description: "10% below the list price",
+    },
+    {
+      value: Number([Math.round(fixedPrice * (1 - 0.2))]),
+      description: "20% below the list price",
+    },
+  ]
+
   const getPercentageOptions = () => {
-    return [0.2, 0.15, 0.1].map((pricePercentage, idx) => {
-      if (listPrice?.major) {
-        return {
-          key: `price-option-${idx}`,
-          value: Math.round(listPrice.major * (1 - pricePercentage)),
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -97,12 +97,14 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
...
 
   const getPercentageOptions = () => {
-    return [0.2, 0.15, 0.1].map((pricePercentage, idx) => {
+    return [0, 0.1, 0.2].map((pricePercentage, idx) => {
       if (listPrice?.major) {
         return {
           key: `price-option-${idx}`,
           value: Math.round(listPrice.major * (1 - pricePercentage)),
-          description: `${pricePercentage * 100}% below the list price`,
+          description: !!pricePercentage
```
araujobarret, line 105: non-blocking: I would not convert a number to boolean, I think it's clearer if compare `pricePercentage !== 0`

araujobarret (approved):
All good too 🚀 

Left a small request for readability's sake and we need to check Integrity if it's going to fail.

src/v2/Apps/Order/Components/__tests__/PriceOptions.jest.tsx
```diff
@@ -178,30 +178,31 @@ describe("PriceOptions", () => {
       expect(radios).toHaveLength(4)
     })
     it("correctly formats values", () => {
-      expect(radios[0]).toHaveTextContent("€80.00")
-      expect(radios[1]).toHaveTextContent("€85.00")
-      expect(radios[2]).toHaveTextContent("€90.00")
+      expect(radios[0]).toHaveTextContent("€100")
```
jo-rs, line 181: Shouldn't these have decimal points? 🤔

tam-kis: Right, good catch!

src/v2/Apps/Order/Components/PriceOptions.tsx
tam-kis: Updated, thanks for the suggestion!

PR #9716, opened 2022-03-23
feat(NOSF2): reverse order of options for range price

tam-kis:

This PR solves [NX-3110](https://artsyproduct.atlassian.net/browse/NX-3110)

Reversed order for options with range price, from high to low price.
Screenshot 2022-03-23 at 13 12 08


araujobarret (approved):
All good 🚀

PR #9718, opened 2022-03-23
feat: remove clickable from low offer notice text

araujobarret:

This PR solves [NX-3112]

### Description

Removes clickable from low offer warning text

Going to inform the data team we don't have anymore the event "We recommend changing your offer" being dispatched

@artsy/negotiate-devs 

[NX-3112]: https://artsyproduct.atlassian.net/browse/NX-3112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Components/MinPriceWarning.tsx
+++ b/src/v2/Apps/Order/Components/MinPriceWarning.tsx
@@ -1,16 +1,14 @@
-import { Clickable, Message, Text } from "@artsy/palette"
+import { Message, Text } from "@artsy/palette"
 import { useEffect } from "react"
 import { AnalyticsSchema, useTracking } from "v2/System/Analytics"
 
 interface MinPriceWarningProps {
   isPriceRange: boolean
-  onClick: React.MouseEventHandler<HTMLButtonElement>
   minPrice: string
   orderID: string
 }
 
 export const MinPriceWarning: React.FC<MinPriceWarningProps> = ({
-  onClick,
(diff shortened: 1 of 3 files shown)


jo-rs (changes requested):
Looks good! 🦾 
Other than the conflict: 
I don't fully remember, so I might be wrong, but I think on the `RadioGroup` component, the `onSelect` and `defaultValue` are only there for the autoselect functionality. So you might be able to remove them without causing any issues, and it would be better to do so to keep things clean. Just check that nothing breaks 🤷

jo-rs:
> Looks good! 🦾 Other than the conflict: I don't fully remember, so I might be wrong, but I think on the `RadioGroup` component, the `onSelect` and `defaultValue` are only there for the autoselect functionality. So you might be able to remove them without causing any issues, and it would be better to do so to keep things clean. Just check that nothing breaks 🤷

Nope. I was wrong. Really sorry. I forgot about [this](https://github.com/artsy/force/pull/9718/files#diff-e3c631e6b08c80b65bdfee8975885955c11cf9850f8fb85bfc7e38657b1e10a9L235) use case. Thank you for pointing that out, @tam-kis! 🙏

jo-rs (approved):
Sorry about the mixup before. Ship it! 🚀

PR #9720, opened 2022-03-23
chore: remove unused artist_id/gene_id query

mdole:

### Description

While looking at https://artsyproduct.atlassian.net/browse/FX-3798, we noticed that `artistID` and `geneID` were being referenced in Force and nowhere else. I took a look and believe they are not actually being used anywhere (combination of cmd + f and removing them and confirming that I didn't break anything locally).

This means that we should only need to support `artistIDs`, `geneIDs`, `keyword`, and `tagID` in our Gravity `MarketingCollection { query { ... } }` type!


src/v2/Apps/Collect/Routes/Collection/index.tsx
```diff
@@ -49,7 +49,7 @@ export const CollectionApp: React.FC<CollectionAppProps> = props => {
     descending_artworks,
     ascending_artworks,
   } = collection
-  const collectionHref = `${sd.APP_URL}/collection/${slug}`
+  const collectionHref = `${getENV("APP_URL")}/collection/${slug}`
```
nickskalkin, line 52: just wondering, where is this change coming from? @mdole

nickskalkin (approved):
👍

Mitch-Henson (approved):

PR #9726, opened 2022-03-23
fix(GRO-882): prevents stretched thumbnail images

dzucconi:

This PR solves [GRO-882]

### Description

Prevents images from stretching by switching over to cropped. (We could also just use the image.width and image.height if we wanted to keep proportionally resized images, which would fix the stretching.)

Also just fixes the collapsing anchor tag. (I see this _alot_, if you're gonna wrap something in a link it and it's not a flex child should be `display: block` (or `flex` or whatever. Just not `inline`). It's less a problem for pointer device users and more a problem for when you're tabbing around (or trying to inspect the DOM).

[GRO-882]: https://artsyproduct.atlassian.net/browse/GRO-882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi:
Might as well bump the res on those headers while I'm in here. One moment.

dzucconi:
Sharper now:

src/v2/Apps/Auctions/Components/MyBids/MyBidsBidHeader.tsx
```diff
@@ -17,6 +17,8 @@ export const MyBidsBidHeader: React.FC<MyBidsBidHeaderProps> = ({ sale }) => {
 
   const image = sale.coverImage?.cropped
 
+  console.log(sale.coverImage)
```
damassi, line 20: stray

damassi (approved):

PR #9735, opened 2022-03-24
fix: cell consistency

dzucconi:

### Description

I noticed some inconsistencies in the artist rail while looking at the artist page today. So I pulled out an artist `Cell` and fixed some consistency issues with the other cells. Still TODO is to pull out a Show cell. That should be the bulk of them. Then it's just finding all the uses here. The whole site is pretty much composed entirely of these things so there's a lot of wins here for code reuse and layout consistency. Surprised we didn't do this sooner.

https://user-images.githubusercontent.com/112297/160009414-89c45466-3cd3-46b7-a17c-0482176fe94f.mov


damassi (approved):
Nice, good idea 👍

PR #9736, opened 2022-03-24
fix(featureflags): return empty object regardless when setting up feature flags

dzucconi:

### Description

If you don't have Unleash set up locally or if there's an error, anything using a feature flag is going to error:


mc-jones:
🙏

PR #9741, opened 2022-03-25
feat(FX-3846): display alternate metadata when hovering on artwork grid item

dimatretyak:
This PR solves [FX-3846]

### Demo
https://user-images.githubusercontent.com/3513494/160117036-60cce5ce-78a6-4190-938d-e610f6bc54a5.mp4


nickskalkin (approved):
Looks good to me 👍  

The only thing unclear to me is why [this test](https://github.com/artsy/force/pull/9741/files#diff-76ed794e874c32ba3bf803a6bf256bca2eb3820865ad83f3a254988ae6c4a82cR73) is changed. Could you explain it to me @dimatretyak?

dimatretyak:
> The only thing unclear to me is why [this test](https://github.com/artsy/force/pull/9741/files#diff-76ed794e874c32ba3bf803a6bf256bca2eb3820865ad83f3a254988ae6c4a82cR73) is changed. Could you explain it to me @dimatretyak?

Yep, sure. This is due to the fact that we are requesting a little more information [here](https://github.com/artsy/force/blob/b13385fada3dbee658e671efba2ea5071f3b840f/src/v2/Components/Artwork/Details.tsx#L337) ([attributionClass](https://github.com/artsy/force/blob/b13385fada3dbee658e671efba2ea5071f3b840f/src/v2/Components/Artwork/HoverDetails.tsx#L66-L68), [mediumType](https://github.com/artsy/force/blob/b13385fada3dbee658e671efba2ea5071f3b840f/src/v2/Components/Artwork/HoverDetails.tsx#L69-L71)) than it was originally

gkartalis (approved):

src/v2/Components/Artwork/HoverDetails.tsx
```diff
@@ -0,0 +1,82 @@
...
+  artwork: HoverDetails_artwork
+}
+
+const HoverDetails: FC<HoverDetailsProps> = ({ artwork }) => {
+  const labels = compact([
+    artwork.attributionClass?.name,
+    artwork.mediumType?.name,
+  ])
+  const pills = labels.map((label, index) => ({
+    id: `${artwork.internalID}-pill-${index}`,
```
dblandin, line 19: **question / non-blocking**: Is it important to use the artwork ID and pass it down to the NonClickablePill as its key or could we rely just on indexes instead? Do having the artwork ID encoded in the component key help with testing? I don't see it currently used for that purpose.

src/v2/Components/Artwork/HoverDetails.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }))
+
+  if (pills.length === 0) {
+    return null
+  }
+
+  return (
+    <HoverContainer>
+      {pills.map((pill, index) => {
+        const isLastPill = pills.length - 1 === index
```
dblandin, line 30: **question / non-blocking**: Might not be an appropriate use case for this, but whenever I see logic for determining if an item is last to adjust styling/spacing I think of our [Join component](https://palette.artsy.net/elements/join/) in Palette. Could that and a spacer component with right margin be useful here?

dblandin (commented):
Looks excellent. Nice work! 💯

dblandin (approved):

src/v2/Components/Artwork/HoverDetails.tsx
dimatretyak: Good catch! I once again forgot about the existence of this component

src/v2/Components/Artwork/HoverDetails.tsx
dimatretyak: > Is it important to use the artwork ID and pass it down to the NonClickablePill as its key or could we rely just on indexes instead?

It is better to always try to pass a unique value for key (this is considered good practice in [React](https://reactjs.org/docs/lists-and-keys.html#keys)). For this situation, we can also pass index as a key, but I decided to follow the recommended rules + this will help us to be safe and avoid problems that may arise if we use index for key (such problems are very difficult to debug)

dblandin: Gotcha. Yeah, seems like keys need to be unique among siblings and not globally but even then, it looks like React still recommends using IDs if available. Sounds good and thanks for clarifying! 👍🏼

PR #9742, opened 2022-03-25
chore: remove unused CollectionsRail component

mdole:
The type of this PR is: **Cleanup**

### Description

While working on moving MarketingCollections from KAWS to Gravity, we were wondering if we needed to support `randomizationSeed` and `showOnEditorial`, which [led to](https://github.com/artsy/gravity/pull/15152/files#r835208868) looking at these components, which led to me realizing that I don't think they're referenced anywhere. 

Cleaning these up should allow us to deprecate `randomizationSeed` and `showOnEditorial`, continuing to clean up and simplify the collections schema.

Let me know if you have questions or I missed something!

cc @dblandin


damassi:
✌️

PR #9748, opened 2022-03-25
refactor(cells): show cell

dzucconi:
Another cell type refactor. This time shows.


damassi (dismissed):
Looks good! One non-blocking (and very minor) bit of feedback for the cells in general is its more common to see naming that looks something like 
```
ThingNameCategory
```
vs 
```
CategoryThingName
```
So:
```
HomeCell
ArtistCell
ArtworkCell
AuctionCell
EtcCell
```

dzucconi:
I prefer broader => smaller. Helps when reading to see the a particular class of object followed by its refinement. 🤷

damassi:
Yeah i was just thinking too that when dealing with something that's more framework-y general, its nice to be able to scan and see the broad category and know that it's not a component that lives in say `Apps/Auction` (where the name prefix is the idiom)

PR #9757, opened 2022-03-28
fix(auction): prevents overflowing text from breaking out of column in Firefox

dzucconi:
Some interaction between `white-space: nowrap` and CSS grid: https://bugzilla.mozilla.org/show_bug.cgi?id=1528030


src/v2/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -178,7 +178,7 @@ export class ArtworkGridContainer extends React.Component<
       <GridColumns width="100%">
         {nodes.map(artwork => {
           return (
-            <Column span={[6, 4]} key={artwork.internalID}>
+            <Column span={[6, 4]} key={artwork.internalID} minWidth={0}>
```
damassi, line 181: 😓

PR #9767, opened 2022-03-29
feat(FX-3847): Move position of save button on artwork grid

gkartalis:

This PR solves [FX-3847]

### Description

WIP


src/v2/Components/Artwork/Details.tsx
```diff
@@ -294,6 +310,7 @@ export const DetailsFragmentContainer = createFragmentContainer(Details, {
       href
       title
       date
+      is_saved
```
gkartalis, line 313: **relay-question:** Do we need to include `is_saved` here to make it accessible in this component? Asking cause down below we are spreading the `NewSaveButton_artwork` which made me thing that since it includes the `is_saved` it would make it accessible in the component but it didn't 🤔

dimatretyak: if `is_saved` should be used in `Details` component (as far as I understand, we should display the "save" button without a hover effect if the user saved artwork earlier), then you need to query it here

dimatretyak: > Asking cause down below we are spreading the NewSaveButton_artwork which made me thing that since it includes the is_saved it would make it accessible in the component but it didn't

in `Details` component, only the data that you explicitly requested will be available (data from fragments will not be available). You can check the available data in an automatically [generated GraphQL file](https://github.com/artsy/force/blob/e03d6679830be82c9458bdbda77fe23b7535631d/src/v2/__generated__/Details_artwork.graphql.ts#L7)

gkartalis:
> @gkartalis I think this PR might need to be rebased off of origin/main so that the commits from https://github.com/artsy/force/pull/9741 aren't included in the diff. Would you mind pushing up the result of the rebase and re-ping for review?
Screenshot 2022-03-30 at 16 51 52

It seems like the diff is ok since I rebased of main after https://github.com/artsy/force/pull/9741 got merged, I only see the commits in the history 🤔

src/v2/Components/Artwork/Details.tsx
```diff
@@ -230,14 +233,20 @@ export const Details: React.FC<DetailsProps> = ({
             )}
         </Flex>
       )}
+      {(!!rest.artwork.is_saved || isHovered) && (
```
dimatretyak, line 236: **question**: I think it is necessary to check `isTouch` and if it is `true`, then don't display the "save" button (this check is [performed](https://github.com/artsy/force/blob/main/src/v2/Components/Artwork/SaveButton/useSaveButton.tsx#L30) for `isSaveButtonVisible`). what do you think about it?

dimatretyak:
"Save" button overlaps long titles

save

src/v2/Components/Artwork/GridItem.tsx
```diff
@@ -94,13 +94,6 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
         </Link>
 
         <Badge artwork={artwork} />
-
-        {isSaveButtonVisible && (
-          <SaveButtonFragmentContainer
-            contextModule={contextModule || ContextModule.artworkGrid}
-            artwork={artwork}
-          />
-        )}
```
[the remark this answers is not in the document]
gkartalis: 😬  woops, great point @dimatretyak

src/v2/Components/Artwork/Details.tsx
```diff
@@ -301,6 +310,7 @@ export const DetailsFragmentContainer = createFragmentContainer(Details, {
       href
       title
       date
+      is_saved: isSaved
```
dblandin, line 313: **suggestion**: These aliases were useful when we were migrating from Metaphysics's V1 to V2 schema but for new additions, we should stick with camelCased attribute names.

dblandin (commented):
Looking good! 👍🏼

src/v2/Components/Artwork/Details.tsx
gkartalis: Added that as well! 🤘

mdole (commented):
great work @gkartalis and @dimatretyak !! Couple quick notes:
- Hovering over the heart on a work you have not saved shows "Unsave", we should fix that
Screen Shot 2022-04-04 at 12 10 49 PM

- The icon in the review app is smaller than in the Figma file, though there's also this note in Figma that is confusing me:

> black 60 / decrease the icon size but keep the hover/interact space larger

Not sure if you've already talked with Lois about that or not, but wanted to flag!

gkartalis:
@mdole thanks! 

> Hovering over the heart on a work you have not saved shows "Unsave", we should fix that
addressed that just now 🤘
> The icon in the review app is smaller than in the Figma file, though there's also this note in Figma that is confusing me:
Already checked with design about this and it looks good for them 🤘

src/v2/Components/Artwork/SaveButton/NewSaveButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+const BTN_HEIGHT = 20
+const BTN_WIDTH = 20
+
+export const NewSaveButton: React.FC<SaveButtonProps> = ({
+  artwork,
+  contextModule,
+}) => {
+  const tracking = useTracking()
+  const [isHovered, setIsHovered] = useState(false)
```
dimatretyak, line 27: **question**: maybe we can use [useSaveButton](https://github.com/artsy/force/blob/7004af4c956c28111ef4910695cd9c0f82683ca9/src/v2/Components/Artwork/SaveButton/useSaveButton.tsx#L10) hook here?

dimatretyak (approved):
Nice work 👍

nickskalkin (approved):
great work 👍

src/v2/Components/Artwork/Details.tsx
```diff
@@ -230,14 +239,22 @@ export const Details: React.FC<DetailsProps> = ({
...
       )}
-      {!hideArtistName && <ArtistLine {...rest} />}
+      <Flex flexDirection="row" justifyContent="space-between">
+        {!hideArtistName && <ArtistLine {...rest} />}
+        {shouldShowHoverSaveButton && (
+          <NewSaveButtonFragmentContainer
+            contextModule={ContextModule.artworkGrid}
+            artwork={rest.artwork}
+          />
+        )}
```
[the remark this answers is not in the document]
gkartalis: In `<GridItem />` we never pass the `hideArtistName` (see [here](https://github.com/artsy/force/blob/f022c3d8d0c26aa345b848e7f036e846137e2305/src/v2/Components/Artwork/GridItem.tsx#L112)). So this is why I didn't consider that.

Am I missing any case were this would cause a problem?

src/v2/Components/Artwork/SaveButton/NewSaveButton.tsx
gkartalis: Why would we want to do that though? 🤔

src/v2/Components/Artwork/Details.tsx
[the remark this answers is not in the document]
dimatretyak: > In <GridItem /> we never pass the hideArtistName (see [here](https://github.com/artsy/force/blob/f022c3d8d0c26aa345b848e7f036e846137e2305/src/v2/Components/Artwork/GridItem.tsx#L112)). So this is why I didn't consider that.

 ahh. good point! I forgot about the fact that the changes are only available in `GridItem` component

src/v2/Components/Artwork/Details.tsx
```diff
@@ -212,9 +217,13 @@ export const Details: React.FC<DetailsProps> = ({
...
 }) => {
   const { isAuctionArtwork } = useArtworkGridContext()
+  const { isHoverEffectEnabled } = useHoverMetadata()
+
+  const shouldShowHoverSaveButton =
+    isHoverEffectEnabled && (!!rest.artwork.is_saved || (isHovered && !isTouch))
 
   return (
-    <>
+    <Box position="relative">
```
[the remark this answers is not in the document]
gkartalis: great point removing that now

src/v2/Components/Artwork/Details.tsx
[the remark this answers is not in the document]
dimatretyak: Thanks for checking

src/v2/Components/Artwork/SaveButton/NewSaveButton.tsx
dimatretyak: To be able to reuse [this logic](https://github.com/artsy/force/blob/7004af4c956c28111ef4910695cd9c0f82683ca9/src/v2/Components/Artwork/SaveButton/useSaveButton.tsx#L17-L27) [here](https://github.com/artsy/force/blob/7004af4c956c28111ef4910695cd9c0f82683ca9/src/v2/Components/Artwork/SaveButton/NewSaveButton.tsx#L52-L58) and [here](https://github.com/artsy/force/blob/7004af4c956c28111ef4910695cd9c0f82683ca9/src/v2/Components/Artwork/SaveButton/NewSaveButton.tsx#L27). It's just a suggestion, if you don't agree with it – just ignore it. Maybe I get too hung up on such things😅

gkartalis: Cool, gave it a quick try and i felt it complicated things 🙈  will ignore it for now but thanks for raising that!

dimatretyak: Got it. Thanks!

gkartalis:
#squashOnGreen

PR #9772, opened 2022-03-30
feat(NX-3053): Only display edition set selection modal when the artwork is non-unique

lean8086:

This PR solves [NX-3053]

### Description

- make tracking event on the confirm button optional: only modal window triggers tracking
- make the content of the confirm button dynamic
- convert OpenInquiryModalCTA into MakeOfferOnInquiryButton
    - make offer button can either open the modal or create an offer directly now
    - creates an offer using the same button component as in the modal window

cc @artsy/negotiate-devs 

[NX-3053]: https://artsyproduct.atlassian.net/browse/NX-3053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Conversation/Components/MakeOfferOnInquiryButton.tsx
```diff
@@ -0,0 +1,105 @@
...
+            <Text color="black60" variant="xs" mb={1}>
+              Only purchases completed with our secure checkout are protected by{" "}
+              <RouterLink to="/buyer-guarantee" target="_blank">
+                The Artsy Guarantee
+              </RouterLink>
+              .
+            </Text>
+          </Flex>
+        </Flex>
+        {artwork.editionSets && artwork.editionSets.length > 1 ? (
```
lean8086, line 61: Working on [NX-3052](https://artsyproduct.atlassian.net/browse/NX-3052) I saw we use `artwork.isEdition`. Would that be also the case here?

src/v2/Apps/Conversation/Components/ConfirmArtworkButton.tsx
```diff
@@ -77,7 +84,7 @@ export const ConfirmArtworkButton: React.FC<ConfirmArtworkButtonProps> = props =
       disabled={props.disabled}
       flexGrow={1}
     >
-      Confirm
+      {props.children || "Confirm"}
```
lean8086, line 87: I use children to make the text part of the content of the button. As a counter-example, this is a property called `buttonText` [on Eigen](https://github.com/artsy/eigen/blob/main/src/app/Scenes/Inbox/Components/Conversations/InquiryMakeOfferButton.tsx#L140) and makes it more difficult to read when defininig both buttons as part of a conditional like [in here](https://github.com/artsy/force/pull/9772/files#diff-6d5dd60e277443342ce1f269fd86dfb8d81ceaae975b653fc20d65b88b1d3075R62-R80).

jo-rs (dismissed):
LGTM! 💪 
I just left one small question about the tracking implementation.

PR #9773, opened 2022-03-30
[SWA-297]: submissionId param

AlicanAkyuz:

This PR solves [[SWA-297](https://artsyproduct.atlassian.net/browse/SWA-297)]

### Description
This PR aims at sending a request to Convection via MP, after a visitor/user coming from a submission receipt email signs up/logs in to the site. 

Whether a visitor/user comes from a submission email is determined by checking the existence of a URL param named `submissionId`. 

When `submissionId` is found in the URL after a sign up/log in action, `submissionId` is set as a cookie and used to make a request to MP upon sign up/log in.


src/v2/Apps/Components/AppShell.tsx
```diff
@@ -30,6 +31,7 @@ interface AppShellProps {
 export const AppShell: React.FC<AppShellProps> = props => {
   useAuthIntent()
   useAuthValidation()
+  useConnectUserToSubmission()
```
damassi, line 34: My first impression is that user-submitted consignments (which are extremely rare) are too specific to be checking as a condition in the AppShell (which is more general). 

Can this go somewhere else more specific to consignments? 

Def non-blocking! Just a thought. And it might be better to move this hook into `Apps/Consign` until it is needed in more than one location in the app.

fladson (approved):

src/v2/Apps/Components/AppShell.tsx
kiryl-zubarau: I think you can check [AuthIntent](https://github.com/artsy/force/blob/7ce1a753ac9c2ee81e70890697c1c766572aaa35/src/v2/Utils/Hooks/useAuthIntent/index.ts#L41). I used it some time ago to save a submission after user logged in(it was before anonymous submissions).

AlicanAkyuz: @damassi do you think kiril's above suggestion would be a good fit for the hook?

fladson (dismissed):

src/v2/Apps/Components/AppShell.tsx
AlicanAkyuz: thank you both. 

I moved the hook to auth intent for now and would like to merge this. please let me know if you can think of a better strategy for this and I'll make the changes in another PR.

PR #9703, opened 2022-03-22
fix: Switched isOfferableFromInquiry check to isOfferable

jo-rs:
This PR solves [NX-3114]

The type of this PR is: Bugfix.

Since we moved all offers to go through conversations, we switched the check for the counteroffer modal from `isOfferableFromInquiry` to `isOfferable`.
cc @artsy/negotiate-devs 

Co-authored-by: Carlos Barreto <araujobarret@gmail.com>
Co-authored-by: Lidiane Taquehara <lidiane.taquehara@artsymail.com>

[NX-3114]: https://artsyproduct.atlassian.net/browse/NX-3114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9777, opened 2022-03-30
fix: Adjust Conversation's CTA modals to consider offerable and offerableFromInquiry

jo-rs:

This PR **actually** solves [NX-3114]

### Description

Caught by @lidimayra.
[This](https://github.com/artsy/force/pull/9703) PR fixed an issue but introduced another problem. This PR should fix that by taking both properties into consideration.
cc @artsy/negotiate-devs


chr-tatu:
I moved the ticket back to `In Review` to reflect the current state.

araujobarret (approved):

src/v2/Apps/Conversation/Components/__tests__/Conversation.jest.tsx
```diff
@@ -64,9 +64,19 @@ describe("Conversation", () => {
...
+      })
+
+      expect(
+        screen.getByText("Be Protected by The Artsy Guarantee")
+      ).toBeInTheDocument()
+    })
+
+    it("shows the buyer guarantee message if offerable from inquiry", async () => {
+      renderWithRelay({
+        Artwork: () => ({ isOfferableFromInquiry: true, isOfferable: false }),
```
chr-tatu, line 79: Is there a test with both `isOfferable` and `isOfferableFromInquiry` `true`?

jo-rs: It's an `||` so I didn't think it was needed, but can quickly add it to be foolproof 👍

jo-rs: Added. Resolving this 📕

PR #9778, opened 2022-03-30
Prefix Google maps key with PUBLIC_ to indicate that it's intentionally exposed in HTML and not a vulnerability

joeyAghion:
The type of this PR is: Enhancement

This PR solves [PLATFORM-3925] (not really _solves_ but that's what the template demands...)

### Description

We receive many bounty reports about keys "leaked" in public HTML. This particular key is restricted to particular domains and operations, so this is intentional and not to be considered a vulnerability. We get continuing reports anyway, so this PR tries to signal in a more obvious way that the "key" can be public.

[PLATFORM-3925]: https://artsyproduct.atlassian.net/browse/PLATFORM-3925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi (dismissed):
Good idea

joeyAghion:
Almost forgot to mention: I've set the new value in staging and production:

```
hokusai [staging|production] env set PUBLIC_GOOGLE_MAPS_API_KEY=...
```

Once released, the old key can be removed from both.

PR #9780, opened 2022-03-30
slack notify dev when validate_production_schema fails

ovasdi:

### Description

Specify slack channels in the `validate_production_schema` job instead of _post-steps_ to deliver notifications to both channels.

In my [testing](https://github.com/artsy/force/commit/5fa8cd717dd8e9e9d04215c635429d5002f13377), the post-step would not trigger a slack/notify if one of the required jobs fails and sends a slack notification.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,14 @@ jobs:
 
   validate_production_schema:
     executor: node/build
+    parameters:
+      # find channel ID at the botton of channel details view (slack app) or by right clicking on channel, and copying the link. The ID will be visible at the end of that URL.
+      practice_web:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -25,13 +25,21 @@ jobs:
 
   validate_production_schema:
     executor: node/build
+    parameters:
+      practice_web:
+        type: string
+        default: C2TQ4PT8R
```
damassi, line 31: Mind adding a comment here pointing to a link where this channel string was found? Will help future devs if this needs to be moved around.

damassi (approved):

PR #9797, opened 2022-04-01
Setup react-i18next

lidimayra:
Thanks to @araujobarret, now we also have an example of i18n on Force using [react-i18next](https://react.i18next.com/) - in parallel to #9449, based on [react-intl](https://formatjs.io/docs/react-intl/).

This implementation seems much simpler than the other one and it would be great to see what other engineers think about this.


src/v2/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -77,9 +77,6 @@ const SuggestionItemLink = styled(RouterLink).attrs({
   min-height: 60px;
 `
 
-export const PLACEHOLDER = "Search by artist, gallery, style, theme, tag, etc."
-export const PLACEHOLDER_XS = "Search Artsy"
```
lidimayra, line 81: These weren't being used anywhere besides inside the `SearchBar` component, so we deleted it from here.

src/v2/System/i18n/i18n.ts
```diff
@@ -0,0 +1,78 @@
...
+const locales = ["en-US", "de-DE"]
+
+i18n
+  .use(initReactI18next)
+  .use(LanguageDetector)
+  .init({
+    debug: true,
+    // backend: {
+    //   loadPath: "assets/locales/{{lng}}/{{ns}}.json",
+    // },
```
lidimayra, line 14: Ideally, we would like the JSON files to be loaded like this, but we couldn't make it work (serving the static assets using webpacker... any ideas?).

src/desktop/components/react/stitch_components/ReactionTooltipQuestion.tsx
```diff
@@ -1,5 +1,6 @@
 import { HelpIcon } from "@artsy/palette"
 import { Tooltip } from "v2/Components/Tooltip"
+import "v2/System/i18n/i18n"
```
lidimayra, line 3: We actually wanted that this `import` would happen only [on the `Boot` component](https://github.com/artsy/force/blob/981fa970763491e8b4204a8a0cf20648101c703d/src/v2/System/Router/Boot.tsx#L31) and it shouldn't be needed anywhere else. However, the components inside the `stitch_components` directory are used in the `.jade` files for rendering some of our static pages (like `/terms`, for example), which don't rely on `Boot` - this is why we need to replicate this `import` in some of these files.

See also [this comment](https://github.com/artsy/force/pull/9449#discussion_r824933305)

src/v2/System/i18n/i18n.ts
araujobarret: we tried a few options but none worked, basically, we need to serve the locale JSON files somewhere, we tried under /assets but got lost in the setup of how to do it, any suggestions? cc @damassi 😬

damassi: Happy to pair on this when y'all have a moment to pick this back up!

lidimayra: Thanks @damassi, that would be great! When would be a good moment for you?

Also: is there something that should be prepared beforehand?

src/v2/Components/NavBar/NavBar.tsx
```diff
@@ -230,7 +236,7 @@ export const NavBar: React.FC = track(
                         })
                       }}
                     >
-                      Log In
+                      {t("navbar.login")}
```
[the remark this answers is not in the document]
lidimayra: Thanksss @pvinis , applied these changes and squashed it!! ❤️

src/desktop/components/react/stitch_components/ReactionTooltipQuestion.tsx
damassi: I wouldn't worry about anything that exists inside of `/desktop`; this folder is soon going to be deleted completely.

src/v2/System/i18n/i18n.ts
damassi: I'm back from OOO; please reach out in slack!

src/desktop/components/react/stitch_components/ReactionTooltipQuestion.tsx
lidimayra: hmm, but it's still being used right? The issue here is that if we don't make this change here, some of our existing pages crash...

damassi: Oh i see! ok, np 👍

damassi: Once we delete desktop this will just go away safely

damassi:
👋 Heads up! A major change is incoming in Force, where we flatten out the `src/v2` directory into `src`. This will take place **next Thursday** and be merged in, which could impact PRs with rebase conflicts (since all of our import paths will soon change). If your PR can be merged in, please do so by next Weds. (See [this thread](https://artsy.slack.com/archives/C02BC3HEJ/p1657821372559579) for more info.)

jest.config.v2.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
       testURL: "http://localhost",
       transform: {
         "\\.(gql|graphql)$": "jest-transform-graphql",
-        ".(ts|tsx|js|jsx)": "babel-jest",
+        "^.+\\.(ts|tsx|js|jsx)$": "babel-jest",
```
pvinis, line 22: this regex feels weird.. the `$` at the end makes sense, and should be enough to exclude the json files. the beginning part seems unnecessary?

src/v2/System/i18n/i18n.ts
```diff
@@ -0,0 +1,24 @@
+import i18n from "i18next"
+import { initReactI18next } from "react-i18next"
+
+const translations = require("v2/System/locales/en-US/translation.json")
+
+i18n.use(initReactI18next).init({
+  debug: true,
```
pvinis, line 7: do we need this `true`? 🤔

src/v2/Components/NavBar/__tests__/NavBar.jest.tsx
```diff
@@ -62,17 +62,17 @@ describe("NavBar", () => {
   describe("desktop", () => {
     it("renders logged out items", () => {
       const wrapper = getWrapper()
-      expect(wrapper.html()).toContain("Log In")
-      expect(wrapper.html()).toContain("Sign Up")
+      expect(wrapper.html()).toContain("navbar.login")
```
pvinis, line 65: this is strange.. should it contain the key? feels like the test should be testing if the actual text is there.

something like
```
it("works in English", () => {
  expect(getWrapper().html()).toContain("Log In")
})
it("works in Greek", () => {
  someFuncToMockBrowserLang("gr")
  expect(getWrapper().html()).toContain("Σύνδεση")
})
```

src/v2/Components/Search/SearchBar.tsx
```diff
@@ -28,10 +26,11 @@ import { useDidMount } from "v2/Utils/Hooks/useDidMount"
 import createLogger from "v2/Utils/logger"
 import { Media } from "v2/Utils/Responsive"
 import { SearchInputContainer } from "./SearchInputContainer"
+import { withTranslation, WithTranslation } from "react-i18next"
 
 const logger = createLogger("Components/Search/SearchBar")
 
-export interface Props extends SystemContextProps {
+export interface Props extends SystemContextProps, WithTranslation {
```
pvinis, line 33: why would we do this here this way, and not with the same `useTranslation` hook?

pvinis: oh i see.. this is a class component.

even though it makes sense, I would much prefer it if we used something like https://github.com/artsy/eigen/blob/26a98c408c223efefbb0b8309448c3bb79256ab3/src/app/Scenes/Search/components/CityGuideCTA.tsx#L12-L13 (docs here at the bottom https://github.com/artsy/eigen/blob/b4484db7fb7ddfbad4e903c993628bbd90a909b3/docs/palette_v3_migration_guide.md)
so we can keep everything nice and centralized, with a `t` exposed.

pvinis (changes requested):
some questions and some suggestions 🤔

src/v2/System/i18n/i18n.ts
lidimayra: I thought it would be better to keep it `true` as it is helpful in case we have issues when translations can't be loaded (default value is false). Would it be better to remove it?

src/v2/Components/NavBar/__tests__/NavBar.jest.tsx
lidimayra: Yeah, this is something that we also discussed during the implementation... but if the goal here is that at some point the language experts should be capable of using a third-party service to update the translations as they wish, do we want the tests to fail when the content of text changes?

jest.config.v2.js
lidimayra: I don't know, but I'm trying to apply this here: [b5a4584](https://github.com/artsy/force/pull/9797/commits/b5a458433ba7f97eb767004e8969028bb471031c) 🤞

src/v2/Components/NavBar/__tests__/NavBar.jest.tsx
pvinis: i would say yes 🤔.

i guess we should support both? it would be great to be able to test the actual text we expand to, to test that languages are actually applied correctly, but also we should allow for key strings to be checked, though that can just be the `testID` 🤔. Im not set on one or the other.

Maybe we start with something and see how it goes. I just feel like checking just for the key might take us to a place where a translation is missing, and we never know, because our test is green, and our website says `navbar.signup` instead of `Sign Up`.

src/v2/System/i18n/i18n.ts
pvinis: as long as it doesnt break anything for real user on prod, then lets keep it here. I'm just not sure what this `true` entails, so I was being cautious.

pvinis (dismissed):
approving this, since the comments are non-blocking. happy to pair or make a PR on top of this for a `<ClassI18n>` component to simplify the class comp case.

src/v2/Components/Search/SearchBar.tsx
lidimayra: I completely agree that your suggestion would be much cleaner, if it's possible to open a PR on top of this one, that would be great. I just don't want to put any pressure on you either, I'm already super grateful for the careful code review. ❤️

src/v2/System/i18n/i18n.ts
lidimayra: Good call!
This is the official definition: https://www.i18next.com/overview/configuration-options

> logs info level to console output. Helps finding issues with loading not working

As I understand, it's just about logs, so I don't think it can break anything. Do you think I could be missing something?

src/v2/Components/NavBar/__tests__/NavBar.jest.tsx
lidimayra: Yeah, this is a super valid point
I added [1b51258](https://github.com/artsy/force/pull/9797/commits/1b512587d6c40577e8ad1fab8a578314a5b30237) to change the test to look for the strings instead of keys

src/v2/System/i18n/i18n.ts
pvinis: makes sense. all good then :)

src/v2/Components/Search/SearchBar.tsx
pvinis: I'll take a quick stab at it.

src/v2/System/i18n/i18n.ts
damassi: We should only enable this in development, which is as easy as checking `debug: getENV("NODE_ENV") === "development"

src/v2/System/i18n/ClassI18n.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from "react"
+import { TFunction, useTranslation } from "react-i18next"
+
+interface ClassI18nProps {
+  children:
+    | React.ReactNode
+    | (({ t }: { t: TFunction<"translation", undefined> }) => React.ReactNode)
+}
+
+export const ClassI18n = ({ children }: ClassI18nProps) => {
```
damassi, line 10: @lidimayra - mind adding some docs in this file around what this file does and why its needed?

damassi:
🎉

src/v2/System/locales/en/translation.json
```diff
@@ -0,0 +1,22 @@
...
+    "artists":"Artists",
+    "artworks":"Artworks",
+    "galleries":"Galleries",
+    "auctions":"Auctions",
+    "fairs":"Fairs",
+    "museums":"Museums",
+    "viewingRooms":"Viewing Rooms",
+    "nfts":"NFTs",
+    "signup":"Sign up",
+    "login":"Log In",
```
anandaroop, line 15: Two things:

- Wondering about difference in casing between these two phrases. Intentional? (Looks they were both title-case before, but now "Sign up" is sentence-case.)

- Noting [the fallout](https://artsy.slack.com/archives/C9SATFLUU/p1658171312694339) in Integrity specs that we're seeing in CI now, related to this change

damassi: Yep, should be title case

damassi: Fixed over here: https://github.com/artsy/force/pull/10550

PR #9799, opened 2022-04-04
[WIP] Evaschilken/fix/px 4953 checkout continue button is inactive default shipping

evaschilken:
The type of this PR is: fix

This fix selected the first shipping quote by default and the user does not have to actively click the first quote. 

[PX-4953](https://artsyproduct.atlassian.net/browse/PX-4953) Checkout: Continue button is inactive even with default shipping option selected

This PR solves [PX-4953]

### Description

This PR is marked as [WIP] as it still should be verified that the selected shipping quote is tracked in google analytics correctly. By removing the necessity to click a shipping quote the event clickedSelectShippingOption is not triggered for all orders.

Link to Slack thread: https://artsy.slack.com/archives/C9ZJYFDNF/p1648566785382469


src/v2/Apps/Order/Utils/shippingUtils.ts
```diff
@@ -147,6 +147,18 @@ export const getShippingOption = (requestedFulfillmentType?: string) => {
   return result
 }
 
+export const getDefaultShippingQuoteId = (order: Shipping_order) => {
```
oxaudo, line 150: Interesting. I believe almost the same thing - if not exactly the same with sorting is done in `ShippingQuotes.tsx`. It's important I believe that the same logic for sorting is used in both places. We probably should refactor to reuse this code rather than repeat it.

oxaudo (approved):
If I remember correctly the consensus was  that this click tracking is not too important here so solution proposed here is a good one to merge.

I have one comment on implementation/refactor. Take it or leave it:)

src/v2/Apps/Order/Utils/shippingUtils.ts
jpotts244: + can adjust!

src/v2/Apps/__tests__/Fixtures/Order.ts
```diff
@@ -395,26 +395,14 @@ export const UntouchedBuyOrderWithShippingQuotes = {
             edges: [
```
jpotts244, line 395: Below are changes the mocked data to reflect the new order we should receive quotes from Exchange API
The response should be ordered by `price_cents`:: https://github.com/artsy/exchange/pull/1156

damassi:
@evaschilken, before merge can we update the PR title to follow our recommended format [per this RFC?](https://github.com/artsy/README/issues/327)

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -609,13 +611,27 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
 
     const showSavedAddresses =
       shippingSelected && addressList && addressList.length > 0
+
     const isArtaShipping = this.isArtaShipping()
+
+    //  add this around
+    const isArtsyShippingAbleToProceed = isArtaShipping
```
oxaudo, line 618: maybe `shippingProcessable` or `shippingActionable`

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -609,13 +611,27 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
...
+      ? !!(shippingQuotes && shippingQuotes.length > 0)
+      : true
+
+    //  ****** rename
+    const shouldArtsyShippingDefault = isArtaShipping
```
oxaudo, line 623: `useDefaultArtaShippingQuote` ?

oxaudo: Also realizing we need to rename to `isArtsyShipping` :)

src/v2/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -786,7 +793,6 @@ export class ShippingRoute extends Component<ShippingProps, ShippingState> {
                 <Button
                   onClick={this.onContinueButtonPressed}
                   loading={isCommittingMutation}
-                  disabled={isContinueButtonDisabled}
                   variant="primaryBlack"
```
jpotts244, line 796: TO DO: we need to ticket better error handling and disabled state, in this PR we are removing the disabled state for the `continue` button due to the new auto quote selection logic. We need to disable in the correct spot where there are actually no quotes returned from arta OR the request to arta failed.

oxaudo: To clarify - that would be a new functionality that is why we are not tying it to this bug. In current production version 'continue' button is NOT disabled where arta returns no quotes or errors.

src/v2/Apps/Order/Components/ShippingQuotes.tsx
```diff
@@ -58,9 +58,7 @@ export const ShippingQuotes: React.FC<ShippingQuotesProps> = ({
   selectedShippingQuoteId,
   ...rest
 }) => {
-  const quotes = compact(
-    shippingQuotes?.map(quote => quote.node)
-  ).sort((a, b) => (a && b ? a.priceCents - b.priceCents : 0))
+  const quotes = compact(shippingQuotes?.map(quote => quote.node))
```
oxaudo, line 61: 🎉 Love that we do not do it inline any more

PR #9804, opened 2022-04-04
chore(images): adds compression taste-test story

dzucconi:
I'm going to do some informal polling on determining appropriate compression levels. Re: [GRO-822](https://artsyproduct.atlassian.net/browse/GRO-822). So this adds a story that runs through some images applying compression and allowing to you compare easily.


jonallured (commented):
Smart - good idea to use storybooks for this. 🏆

PR #9830, opened 2022-04-07
chore: clean up dead code

dzucconi:
This PR just removes 3 apps.

* `style_guide` just had a redirect to Palette. No one is hitting this route. Deleted the app.
* `apply` is just a set of redirects to apply.artsy.net. Migrated to the `hardcodedRedirects` file
* `inquiry` was, if memory serves, used in Martsy where we just linked to an inquiry page rather than opening it in a modal. No longer needed. Deleted the app.


damassi:
🎉

PR #9839, opened 2022-04-07
fix: Suppress lots that are closed from trending rail GRO-891

jonallured:
I expect this PR will fail on CI because I can't get one of the jest specs to behave. I did a quick call with @dzucconi and it was not clear to us what was happening so I'm going to assign you @damassi and maybe you can see something we did not?

Either way the actual point of this PR is to remove closed lots from this rail:

Screen Shot 2022-04-07 at 2 54 08 PM

I assumed there was some problem with the way we were making the query to MP and ultimately Gravity but after some chatting with @mzikherman I think this is actually something else, likely a missed indexing opportunity when a sale closes. More chatter here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1649262707508329

It's worth noting his advice towards the end of that thread:

> in general tho, i think its known and accepted that you may see elasticsearch staleness, even if you ‘correctly’ have everything hooked up, unfortunately.

So I could follow up here and see about improving the indexing and I should but ultimately this code change ensures the correctness of the rail so it seems worth doing to me. 🤷 

https://artsyproduct.atlassian.net/browse/GRO-891

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Auctions/Components/__tests__/TrendingLotsRail.jest.tsx
+++ b/src/v2/Apps/Auctions/Components/__tests__/TrendingLotsRail.jest.tsx
@@ -66,12 +66,15 @@ describe("TrendingLotsRail", () => {
 
   it("skips closed lots", () => {
     const wrapper = getWrapper({
-      SaleArtworksConnection: () => ({
-        edges: [{ node: { sale: { isClosed: true } } }],
-      }),
+      Boolean: () => true,
+      SaleArtworksConnection: () => {
+        return {
+          edges: [{ node: { sale: { isClosed: true } } }],
+        }
+      },
     })
 
-    expect(wrapper.html()).toContain("No Works To Show")
+    expect(wrapper.text()).toContain("No Works To Show")
   })
 
   it("tracks clicks", () => {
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Auctions/Components/__tests__/TrendingLotsRail.jest.tsx
```diff
@@ -64,6 +64,16 @@ describe("TrendingLotsRail", () => {
...
   })
 
+  it("skips closed lots", () => {
+    const wrapper = getWrapper({
+      SaleArtworksConnection: () => ({
+        edges: [{ node: { sale: { isClosed: true } } }],
+      }),
+    })
+
+    expect(wrapper.html()).toContain("No Works To Show")
```
jonallured, line 74: This errors and is very confusing! When I log out the node inside the component it does NOT match the mock data there in the `getWrapper` call - why is that? 🤔

dzucconi: And @damassi note that we also tried `Sale: () => ({ isClosed: true })` which is what I'd typically do and didn't work either

PR #9846, opened 2022-04-08
fix: no shipping info when price on request

araujobarret:

This PR solves [NX-3106]

### Description

It Fixes one issue given price on request, we were displaying shipping info when we don't want in this case.

Before
Screenshot 2022-03-11 at 16 34 23

After
Screenshot 2022-04-08 at 11 00 14

[NX-3106]: https://artsyproduct.atlassian.net/browse/NX-3106?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -95,15 +95,14 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
     )
   }
 
-  renderEditionSet(
-    editionSet: EditionSet,
-    includeSelectOption: boolean,
-    editionSelectableOnInquireable: boolean
-  ) {
+  renderEditionSet(editionSet: EditionSet) {
+    const {
+      is_offerable: isOfferable,
+      is_acquireable: isAcquireable,
+      is_inquireable: isInquireable,
+    } = this.props.artwork
     const editionEcommerceAvailable =
-      editionSet?.is_acquireable ||
-      editionSet?.is_offerable ||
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -414,8 +411,8 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
           ) : (
             <>
               {this.renderEditionSets(
-                artworkEcommerceAvailable,
-                editionSelectableOnInquireable
+                artworkEcommerceAvailable || !!isInquireable,
+                !!isInquireable
```
tam-kis, line 415: Why are we checking twice if `!!isInquireable` , is the first one strictly related to `artworkEcommerceAvailable` and doesn't work with just `!!isInquireable`?

araujobarret: looks like a checking but I only kept the same behavior to the `renderEditionSets` function, which expected the first param to also check `isInquireable`, the second parameter was also `isInquireable`. I can refactor the function if you ask 😬

araujobarret: I refactored a bit to pick artwork prop inside the renderEditionSet function in [dc6953f](https://github.com/artsy/force/pull/9846/commits/dc6953f74d5bef6419467ecd9879483dd00720b3) 👍

PR #9849, opened 2022-04-08
chore: remove `getAttributionClassValueByLabel` helper

dimatretyak:
The type of this PR is: **Chore/Refactor**

### Description
Small refactor for "Create Alert" button on artwork page


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateAlertButton.tsx
```diff
@@ -90,7 +87,7 @@ export const ArtworkSidebarCreateAlertButtonFragmentContainer = createFragmentCo
           slug
         }
         attributionClass {
-          name
+          internalID
```
gkartalis, line 90: I had to run the query to make sure what internalID is giving us in this case because when I first saw that I was a bit confused.

Querying for this in insomnia 
```graphql
{
   artwork(id: "anthony-cudahy-ian-coil") {
		attributionClass {
			internalID
			name
		}
	}
}
```

I am getting the following response:

```graphql
{
	"data": {
		"artwork": {
			"attributionClass": {
				"internalID": "unique",
				"name": "Unique"
			}
		}
	}
}
```

which makes sense, leaving that here for other reviewers (if any)

gkartalis (dismissed):

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateAlertButton.tsx
dimatretyak: Thanks!

PR #9859, opened 2022-04-11
fix: broken conversations given no Collector name present

araujobarret:

This PR solves [NX-3119]

### Description

Removes one unused(`name`) prop from Conversation query that was breaking the client given some abnormal Collectors without a name, that fixes the broken state when opening conversations.


src/v2/Apps/Conversation/Components/Reply.tsx
```diff
@@ -1,6 +1,6 @@
 import { Environment, RelayRefetchProp } from "react-relay"
-import { useRef, useState } from "react";
-import * as React from "react";
+import { useRef, useState } from "react"
```
araujobarret, line 2: sorry about that, prettier strikes back 👎

lidimayra (approved):
👏 👏 👏

PR #9864, opened 2022-04-11
ci: Remove orb dependency from circleci config

lidimayra:

### Description

This was supposed to be shared, but it was only being used in Force, so it makes more sense to reference it directly here.


dblandin:
Going to go ahead and merge this early since this is diff only touches CI config and it's preventing the release of a production bug fix.

PR #9880, opened 2022-04-12
fix(auction): Restore old urls for older versions of Eigen

damassi:

### Description

Following the thread [here](https://artsy.slack.com/archives/C01ADJNCS5D/p1649788140435499) it turns out some old Eigen versions are relying on urls that no longer exist after the /auction/id rebuild. This restores those URLs:

- /auction-registration/<sale-slug>: redirects to /sale/id/register
- /auction-faq: promotes /auction-faq2 to /auction-faq
- /auction/<sale-slug>/buyers-premium: redirects to /auction-faq (since that page contains buyer premium information

### DIFF
--- a/src/v2/Apps/Auction/auctionRoutes.tsx
+++ b/src/v2/Apps/Auction/auctionRoutes.tsx
@@ -159,7 +159,7 @@ export const auctionRoutes: AppRouteConfig[] = [
     ],
   },
   {
-    path: "/auction-faq2",
+    path: "/auction-faq",
     getComponent: () => AuctionFAQRoute,
     query: graphql`
       query auctionRoutes_AuctionFAQRouteQuery {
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Auction/auctionRoutes.tsx
```diff
@@ -169,6 +169,27 @@ export const auctionRoutes: AppRouteConfig[] = [
...
+      new Redirect({
+        from: "/",
+        to: "/auction/:slug/register",
+      }) as any,
+    ],
+  },
+
+  {
+    // Legacy redirect for old Eigen clients
+    path: "/auction/:slug/buyers-premium",
```
brainbicycle, line 185: 🙏 thanks!

brainbicycle (approved):
thanks!!!

PR #9881, opened 2022-04-12
fix(show): higher quality installation shots; top context bar consistency

dzucconi:
It was brought up in `#gallery-help` that the install shots looked low quality. They were using the 'larger' source size for resizing. So this PR Fixes that. I also fixed some consistency things on the page and wound up addressing consistency among the `TopContextBar`s around the site. Now the component handles all the common use-cases. Added corresponding stories.


src/v2/Apps/ArtistSeries/Components/ArtistSeriesHeader.tsx
```diff
@@ -34,47 +27,19 @@ const ArtistSeriesHeader: React.FC<ArtistSeriesHeaderProps> = ({
...
-              />
-            }
-          />
-        </Flex>
+      <TopContextBar
+        href={artist.href}
+        displayBackArrow
+        src={artist.image?.url}
+      >
+        {artist.name}
```
damassi, line 42: 👍

PR #9897, opened 2022-04-13
chore: removes unused dependencies

dzucconi:
This only looks at dependencies and not devDependencies. I went about this in a naive way: project searching each dependency and only removing it if nothing in Force is referencing it. (I suppose it's possible that Reaction is depending on one of these?)

The only benefit we'd get from this is a slightly shorter `yarn install` — but it's nice to get rid of the noise.

There's certainly some dead model/view code that is requiring libraries but not actually in use, but it's too difficult to untangle that. So much more to come once we drop the old apps, obviously.


damassi (approved):
A+

PR #9900, opened 2022-04-14
feat(TX-255): add ACH bank debit as option to payment flow

rquartararo:

This PR solves [TX-255]

### Description
This PR is the start of the new payment flow with Stripe ACH bank debit as an option. 

**This PR includes:**
- Refactoring of `Routes/Payment/index.tsx` to a functional component. 
- Two-step payment process where the user can choose which payment method to select, and then enter the details in the next step. (behind the “stripe_ACH” feature flag)
- Fetching of the SetupIntent from Gravity(in the process of moving to Exchange) to load the Stripe Payment Element for users to connect/enter bank information. (behind the “stripe_ACH” feature flag)
- Attaches bank info to SetupIntent after the user continues to the next transaction step. (behind the “stripe_ACH” feature flag)

**This PR does NOT include:**
- Optimisation for mobile view --- I still need to fix the sidebar submit button. 
- Redirect to the “Review” step (will complete in a separate PR) 

### Stripe Docs used
[Accept ACH Payment](https://stripe.com/docs/payments/ach-debit/accept-a-payment)
[Payment Element](https://stripe.com/docs/payments/payment-element)
[SetupIntent API](https://stripe.com/docs/payments/setup-intents)

### Recording 

https://user-images.githubusercontent.com/50849237/163352324-8c9ad3dc-d515-491a-9ade-491360402d2f.mov

[TX-255]: https://artsyproduct.atlassian.net/browse/TX-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -50,36 +51,35 @@ export interface PaymentProps {
...
 }
 
-interface PaymentState {
-  isGettingCreditCardId: boolean
-}
+type Props = PaymentProps & StripeProps
 
 const logger = createLogger("Order/Routes/Payment/index.tsx")
 
-@track((props: PaymentProps) => ({
```
rquartararo, line 59: Not sure if this is the correct way to add @track to a functional component. Any ideas on this?

src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    if (error) {
+      console.log("error", error)
+    }
+  }
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <LoadingArea isLoading={isPaymentElementLoading}>
+        {isPaymentElementLoading && <Box height={300}></Box>}
```
rquartararo, line 41: I added this to the loading area so that there was space where the element would be appearing. Is there a better way to achieve this?

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -211,6 +155,66 @@ export class PaymentRoute extends Component<
...
+  return (
+    <Box data-test="orderPayment">
+      <OrderStepper
+        currentStep="Payment"
+        steps={order.mode === "OFFER" ? offerFlowSteps : buyNowFlowSteps}
+      />
+      <TwoColumnLayout
+        Content={
+          isACHEnabled ? (
+            <PaymentContent
```
rquartararo, line 168: The new two-step payment flow is in the `PaymentContent` component.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { relayEnvironment } = useSystemContext()
+  const [clientSecret, setClientSecret] = useState("")
+
+  const stripePromise = loadStripe(getENV("STRIPE_PUBLISHABLE_KEY"), {
+    betas: ["us_bank_account_beta_2"],
+  })
+
+  useEffect(() => {
+    // Create SetupIntent as soon as the page loads
+    commitMutation<createBankDebitSetupMutation>(relayEnvironment!, {
```
damassi, line 20: we have the `useMutation` hook you can use here. Lots of examples throughout Force, most notably in the new `Apps/Auction` folder

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -50,36 +51,35 @@ export interface PaymentProps {
...
       : AnalyticsSchema.Flow.MakeOffer,
 }))
-export class PaymentRoute extends Component<
-  PaymentProps & StripeProps,
-  PaymentState
-> {
-  state: PaymentState = { isGettingCreditCardId: false }
-  paymentPicker = createRef<PaymentPicker>()
-  onContinue = async () => {
+export const PaymentRoute: FC<Props> = props => {
```
starsirius, line 64: Thank you for refactoring! 💯

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -0,0 +1,127 @@
...
+  } = props
+  const [stepOneComplete, setStepOneComplete] = useState(false)
+  const [paymentMethodSelection, setPaymentMethodSelection] = useState("")
+  const [paymentMethod, setPaymentMethod] = useState("")
+
+  const paymentMethodLabel = (paymentMethod: string) => {
+    switch (paymentMethod) {
+      case "credit_card":
+        return "Credit Card"
+      case "bank_transfer":
```
starsirius, line 49: Ideally the engineering-facing variables and user-facing labels would match, but in reality they can diverge. I wonder if it's better to be consistent for those engineering-facing variables which we can control. For example, maybe we can use `bank_debit` as the payment method to be consistent with Exchange, and user-facing label can say "Bank Transfer" (or whatever copy desired).

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -0,0 +1,127 @@
...
+    setStepOneComplete(true)
+  }
+
+  return (
+    <div>
+      <Flex
+        flexDirection="column"
+        style={isLoading ? { pointerEvents: "none" } : {}}
+      >
+        {/* Step One Payment Menthod */}
```
starsirius, line 65: 👍

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -0,0 +1,127 @@
...
+              order={order}
+              innerRef={paymentPicker}
+            />
+            <Spacer mb={4} />
+            <Media greaterThan="xs">
+              <ContinueButton onClick={onContinue} loading={isLoading} />
+            </Media>
+          </Collapse>
+        )}
+        {paymentMethod && paymentMethod === "bank_transfer" && (
```
starsirius, line 119: Very clear!

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -50,36 +51,35 @@ export interface PaymentProps {
...
+  const [isGettingCreditCardId, setIsGettingCreditCardId] = useState(false)
+  const { order, isCommittingMutation } = props
+  const isLoading = isGettingCreditCardId || isCommittingMutation
+  const paymentPicker = createRef<PaymentPicker>()
+  const isACHEnabled = useFeatureFlag("stripe_ACH")
```
starsirius, line 69: Yes!

src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [isPaymentElementLoading, setIsPaymentElementLoading] = useState(true)
+
+  const handleSubmit = async event => {
+    event.preventDefault()
+
+    if (!stripe || !elements || !user) {
+      return
+    }
+
+    const { error } = await stripe.confirmSetup({
```
starsirius, line 20: I'm a bit confused by `stripe.confirmSetup` vs. ACH-specific `stripe.collectBankAccountForSetup` and `stripe.confirmUsBankAccountSetup`. Will also do some more research to get a better understanding.

src/v2/Components/BankDebitForm/BankDebitForm.tsx
starsirius: Seems reasonable to me and the recording looks great!

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+          $input: CreateBankDebitSetupInput!
+        ) {
+          createBankDebitSetup(input: $input) {
+            clientSecret
+          }
+        }
+      `,
+      variables: {
+        input: {
+          clientMutationId: "create-bank-debit-setup",
```
starsirius, line 32: I think this can be omitted because Relay client generates this?

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+          paymentMethodTypes: ["us_bank_account"],
```
starsirius, line 33: With the updated [Exchange mutation](https://github.com/artsy/exchange/pull/1129), clients won't need to pass the payment method types!

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { createBankDebitSetupMutation } from "v2/__generated__/createBankDebitSetupMutation.graphql"
+import { useSystemContext } from "v2/System"
+import { BankDebitForm } from "./BankDebitForm"
+
+export const BankDebitProvider: FC = () => {
+  const { relayEnvironment } = useSystemContext()
+  const [clientSecret, setClientSecret] = useState("")
+
+  const stripePromise = loadStripe(getENV("STRIPE_PUBLISHABLE_KEY"), {
+    betas: ["us_bank_account_beta_2"],
```
starsirius, line 15: Maybe the beta flag can be dropped now the product is public.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { commitMutation, graphql } from "relay-runtime"
+import { createBankDebitSetupMutation } from "v2/__generated__/createBankDebitSetupMutation.graphql"
+import { useSystemContext } from "v2/System"
+import { BankDebitForm } from "./BankDebitForm"
+
+export const BankDebitProvider: FC = () => {
+  const { relayEnvironment } = useSystemContext()
+  const [clientSecret, setClientSecret] = useState("")
+
+  const stripePromise = loadStripe(getENV("STRIPE_PUBLISHABLE_KEY"), {
```
starsirius, line 14: [Here](https://stripe.com/docs/stripe-js/react#elements-provider) mentioned that we might want to move `loadStripe` outside to avoid recreating the `Stripe` object on every render.

starsirius (commented):
Looking great! Very excited!

src/v2/Components/BankDebitForm/BankDebitForm.tsx
rquartararo: It is my understanding that `stripe.confirmUsBankAccountSetup` is done during the Payment Element account selection process. The stripe docs for implementing ACH do not include using the Payment Element, so I had to go back and forth between the ACH-specific doc and the Payment Element implementation doc to get this flow.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Thanks, I think you're right!

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Yay! Thanks!

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Good catch let me test it without.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: I tried this and it broke the `mockStripe` tests for some reason. I will look into it more.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Thanks, I will use that instead!

src/v2/Components/BankDebitForm/BankDebitForm.tsx
rquartararo: but yes I'm also a bit confused by this as well

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
rquartararo: Ah okay, yeah that makes sense thank you for the suggestion!

src/v2/Apps/Order/Routes/Payment/index.tsx
starsirius: I think for tracking functional components, we can use the [`useTracking`](https://github.com/NYTimes/react-tracking#usage-with-react-hooks) hook. So the decorator here for class components can just be removed.

src/v2/Components/BankDebitForm/BankDebitForm.tsx
starsirius: Still curious the difference here between using `stripe.confirmSetup` and `stripe.confirmUsBankAccountSetup`. Definitely non-blocking for this PR, and we can test in the followup.

src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <form onSubmit={handleSubmit}>
+      <LoadingArea isLoading={isPaymentElementLoading}>
+        {isPaymentElementLoading && <Box height={300}></Box>}
+        <PaymentElement
+          onReady={() => setIsPaymentElementLoading(false)}
+          options={{
+            fields: {
+              billingDetails: {
+                name: "never",
+                email: "never",
```
starsirius, line 48: #non-blocking. I wonder if there is a case where buyers want to use a different billing information than the logged-in user. 🤔

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const data = await submitMutation({
+        variables: { input: { id: order.internalID } },
+      })
+
+      if (
+        data.commerceCreateBankDebitSetupForOrder?.actionOrError.__typename ===
+        "CommerceOrderRequiresAction"
```
starsirius, line 25: Created a ticket for error handling here: https://artsyproduct.atlassian.net/browse/TX-285

starsirius (approved):
Amazing! 💯  👏  👏  👏  Can't wait to try it out on staging!

src/v2/Apps/Order/Routes/Payment/index.tsx
rquartararo: Okay so we can remove this and we don't need to add the `useTracking` hook in it's place? I don't see the `@track` being used in the component anywhere, so I assume that would be the case.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Ah, I forgot to look into this, mind if I make this a task for the next PR?

src/v2/Components/BankDebitForm/BankDebitForm.tsx
rquartararo: I thought of that as well... I can bring it up in the next ACH meeting.

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
rquartararo: Thanks!

PR #9901, opened 2022-04-14
Return country code only if required

kiryl-zubarau:

### Description

This PR fixes issue disscussed here: https://artsy.slack.com/archives/C9YNS4X32/p1649897338091979

### DIFF
--- a/src/v2/Components/LocationAutocompleteInput.tsx
+++ b/src/v2/Components/LocationAutocompleteInput.tsx
@@ -174,7 +174,10 @@ export type Location = {
   coordinates?: number[]
 }
 
-export const normalizePlace = (place?: Place): Location => {
+export const normalizePlace = (
+  place?: Place,
+  withCountryCode: boolean = false
+): Location => {
   if (!place) return { city: "" }
 
   if (!("address_components" in place)) {
@@ -220,7 +223,7 @@ export const normalizePlace = (place?: Place): Location => {
     stateCode: components.state?.short_name,
     postalCode: components.postalCode?.long_name,
     country: components.country?.long_name,
-    countryCode: components.country?.short_name,
+    countryCode: withCountryCode ? components.country?.short_name : undefined,
   }
 }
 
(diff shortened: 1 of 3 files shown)


patrinoua (approved):

AlicanAkyuz (approved):

src/v2/Components/LocationAutocompleteInput.tsx
```diff
@@ -220,7 +223,7 @@ export const normalizePlace = (place?: Place): Location => {
     stateCode: components.state?.short_name,
     postalCode: components.postalCode?.long_name,
     country: components.country?.long_name,
-    countryCode: components.country?.short_name,
+    countryCode: withCountryCode ? components.country?.short_name : undefined,
```
jacobherrington, line 226: **question (non-blocking):** I'm curious why were are using `undefined` rather than `null`.

My understanding is that it's preferable to not use `undefined` manually and bias to `null` for values that represent an empty value.

kiryl-zubarau: I used `undefined` here to prevent the `countryCode` field from being sent to the server, if I use `null` here the Relay will think that I want to send the field to the server with `null` value, but if I'll send `undefined` the Relay will ignore the field.

iskounen: That's interesting to know! However, I worry that your intention behind using `undefined` won't be communicated to future developers who encounter this code (and we'll end up with the same bug if they change it). This fix is urgent so I also think that this is (non-blocking) but it would be good to take this object out of the `return` statement and conditionally add `countryCode` to it so that the intention is clearer.

iskounen (approved):
LGTM. I think that some discussion about how this change fixes the issue discussed in the link would be helpful to have in the description, but given the urgency let's merge.

kiryl-zubarau:
Going to merge! Feel free to left any comments. Will be updated in follow up PR

src/v2/Components/LocationAutocompleteInput.tsx
kiryl-zubarau: Generally I want to back to the previous version of the `normalizePlace` and always return the `countryCode`, but update places where `normalizePlace` uses to use only needed fields. That should save code from the same problem in future(adding new field).

iskounen: Cool. I'm looking forward to checking that out in the follow-up PR 🙌

PR #9910, opened 2022-04-14
feat: migrates article channel 'blogs'

dzucconi:
* https://github.com/artsy/metaphysics/pull/3972 and https://github.com/artsy/metaphysics/pull/3973 adds support for article channels in metaphysics

This PR:

* Rebuilds `/life-at-artsy`, `/artsy-education`, `/buying-with-artsy`, as a "Channel" page within the articles apps.
* Deletes old "team_channel" route and redirects former top-level routes to `/channel/:id`
* Deletes remaining Parsely code.

### Old / New
32w2l1mN


dzucconi:
From what I can tell Parsley is already removed from the integrations. There was just some things left around from it.

damassi:
it would be great to kill that top coffeescript image but i know that's not part of this code

dzucconi:
lol wanna art direct some code? I think this can be swapped in Writer.

PR #9912, opened 2022-04-14
ci(slack-notifications): remove practice-web

damassi:
The type of this PR is: **CI**

### Description

This removes practice-web from out-of-sync MP slack pings since we now have that notification going directly to #dev


ovasdi (approved):
🦖

PR #9914, opened 2022-04-15
chore: deletes personalize app (onboarding)

dzucconi:
This PR deletes the personalize app + routes. I've included redirects to the homepage on the off chance something else is linking to them.

I suspect some integrity specs will fail here as well. Will convert off draft after I have a PR fixing those.


dzucconi:
There's some peripheral clean up here to do too. Like how Cohesion requires an `onboarding` flag for it's `CreatedAccount` event:

dzucconi:
https://github.com/artsy/integrity/pull/323

The-Beez-Kneez:
https://media.giphy.com/media/gg2VBE0zVrMGUutcwy/giphy.gif

jonallured:
Pretty funny to think of this as a chore haha - let's do it!

PR #9920, opened 2022-04-18
refactor: migrates shortcuts

dzucconi:
Simple PR, but I ran into a confusing issue that I need a second set of eyes for when you get back @damassi.

So firstly: we do have the need for non-React "modern" Express apps. This doesn't really make sense as a middleware because we want it to only match `/:short`. There's some other routes that will fall under a similar constraint, sitemaps, for example. We'll also probably want to refactor [this mess](https://github.com/artsy/force/blob/30d1b5706b027164767c89ee08848a6e3d54648d/src/desktop/index.ts#L61-L66) of what are now just a set of redirects for similar top-level routes.

It also doesn't make a ton of sense to have to go through the GraphQL layer for this. Additionally we're losing what caching we were doing here previously. This likely isn't a huge deal and we could cache if need be but figured I'd PR without.


src/desktop/index.ts
```diff
@@ -4,6 +4,10 @@ import * as globalReactModules from "desktop/components/react/stitch_components"
 
 const app = express()
 
+// FIXME: Move this
+import { app as shortcuts } from "v2/System/Server/shortcuts"
+app.use(shortcuts)
```
dzucconi, line 9: So this shouldn't be here but bizarrely I was unable to actually mount it anywhere else and it still work. This is one of those things that will require a rubber duck pair because if I try to explain it just doesn't make sense.

src/v2/System/Server/shortcuts.ts
```diff
@@ -0,0 +1,23 @@
...
+    const short = req.params.short.toLowerCase()
+
+    fetch(`${getENV("API_URL")}/api/v1/shortcut/${short}`, {
+      headers: {
+        "X-XAPP-TOKEN": getENV("ARTSY_XAPP_TOKEN"),
+      },
+    })
+      .then(response => response.json())
+      .then(data => {
+        res.redirect(301, data.long)
```
dzucconi, line 19: Previously we weren't 301ing. I'm not if this is desirable since you do technically have the ability to update a redirect. cc @jonallured

src/v2/System/Server/index.ts
```diff
@@ -0,0 +1,6 @@
+import { Express } from "express"
+import { app as shortcuts } from "./shortcuts"
+
+export const initializeExpressApps = (app: Express) => {
+  app.use(shortcuts)
+}
```
dzucconi, line 6: Was going to use this. It doesn't mount the routes correctly (!?)

dzucconi:
It feels like we should handle the top-level profile redirects and shortcuts in a single express app/route handler, actually now that I look at this.

src/v2/System/Server/shortcuts.ts
jonallured: Did some poking around in the db here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1650317555473629

My take is: let's 301. 😃

src/desktop/index.ts
```diff
@@ -42,8 +43,8 @@ app.use(require("./apps/eoy_2016"))
 app.use(require("./apps/sitemaps"))
 app.use(require("./apps/rss"))
 
-// Non-profile dynamic vanity url apps
-app.use(require("./apps/shortcuts"))
+// FIXME: Move to v2/server.ts once remaining legacy apps are retired (the order of this app matters)
+app.use(shortcutsServerRoutes)
```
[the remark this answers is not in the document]
dzucconi: Ah, rather than throwing a 404 on missing shortcut; we should just `next` onto the next handler. Falling thru completely is a 404.

PR #9932, opened 2022-04-20
fix(NX-3092): constraint the height of the conversation list container

lean8086:

This PR solves [NX-3092]

### Description

The calculations and scroll event were fine. The problem was that the layout of the conversation list wasn't limited to the height of the viewport, as it is limited on the main layout.

cc @artsy/negotiate-devs 

[NX-3092]: https://artsyproduct.atlassian.net/browse/NX-3092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Conversation/Components/ConversationList.tsx
+++ b/src/v2/Apps/Conversation/Components/ConversationList.tsx
@@ -8,7 +8,7 @@ import {
 import styled from "styled-components"
 import compact from "lodash/compact"
 import { Box, Flex, Spinner } from "@artsy/palette"
-
+import { themeGet } from "@styled-system/theme-get"
 import { ConversationSnippetFragmentContainer as ConversationSnippet } from "./ConversationSnippet"
 import { ConversationListHeader } from "./ConversationListHeader"
 
@@ -24,6 +24,16 @@ const SpinnerContainer = styled.div`
   width: 100%;
   height: 100px;
   position: relative;
+`
+
+const ConstrainedHeightContainer = styled(Flex)`
+  height: calc(100vh - 103px);
+  width: 100%;
+  overflow: hidden;
+  flex-direction: column;
+  @media ${themeGet("mediaQueries.xs")} {
+    height: calc(100vh - 60px);
+  }
 `
 
 export const PAGE_SIZE: number = 15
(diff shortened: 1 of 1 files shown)


src/v2/Apps/Conversation/Components/ConversationList.tsx
```diff
@@ -26,6 +26,16 @@ const SpinnerContainer = styled.div`
   position: relative;
 `
 
+const ConstrainedHeightContainer = styled(Flex)`
+  height: calc(100vh - 103px);
```
lean8086, line 30: I wanted to avoid repeating this line of code [from a parent component](https://github.com/artsy/force/blob/main/src/v2/Apps/Conversation/Routes/Conversation/index.tsx#L28-L33), especially having a number in `px` in it. I tried extending the component in styled-components (example below), and the result is two different classes applied to the same element, being `height` not inherited. Happy to hear alternatives to this!

```
import { ConstrainedHeightContainer } from "./Routes/Conversation"

const ConstrainedContainer = styled(ConstrainedHeightContainer)`
  ...
`
```

lidimayra: I have no clue how can I be helpful here... but the changes in this PR are enough to fix a bug, so unless anyone else has other ideas, I wouldn't hold this PR because of this.

I'm approving it for now and if we come up with ideas for improvements here, I would be fine having that submitted in a new PR

lidimayra (approved):
LGTM!

src/v2/Apps/Conversation/Components/ConversationList.tsx
araujobarret: I know was here before but I also don't like those magic numbers, I guess the right API in order to load more if we reach a certain part of the screen is the [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) instead of crazy math with pixels, but the fix seems reasonable, gonna create a ticket to spike the Intersection Observer attempt 👍

araujobarret (approved):

src/v2/Apps/Conversation/Components/ConversationList.tsx
araujobarret: We have something similar in [Palette](https://github.com/artsy/palette/blob/main/packages/palette/src/utils/useSentinelVisibility.tsx) but the approach is different there, maybe we need a hook similar to this to load more or not if a sentinel div is reached out

lean8086: Thanks for the feedback and all the info!

PR #9940, opened 2022-04-21
fix: fill width/height for MagnifyImage container

dimatretyak:

#### Before
before

#### After
after


dimatretyak:
I merged changes, since a QA session is scheduled for this today

PR #9947, opened 2022-04-21
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)

artsyjian:
feat: use latest version of artsy managed orbs, and track app version in k8s. (PLATFORM-4103)


ovasdi (approved):
🦖

PR #9952, opened 2022-04-22
fix[cookie-consent]: add bottom margin to checkout flow

mdole:
The type of this PR is: **bugfix**

### Description

Currently, if a user has not accepted cookies and views an order page, they will be unable to see the "continue" button at the bottom of the screen and are unable to scroll it into view.

While we figure out a better long-term solution, this PR adds more margin to the container so that users can at least scroll the button into view and not get stuck.

Before:
before

After:
after

After + user has accepted cookies:
after-accepting-cookies


dimatretyak (approved):
LGTM 👍

evaschilken:
Thanks Matt! I have checked out the branch and tested locally but it seems like there are some areas where the padding might still not be enough:
1. Register to bid: Cookie consent is hiding the checkbox and continue button
2. Sign up: Cookie consent is hiding Facebook Sign-on button

Screenshot 2022-04-22 at 14 49 50
Screenshot 2022-04-22 at 14 54 54

Can the padding be increased even more?

PR #9954, opened 2022-04-22
[GRO-861] turn on query param flag for fallback artists 

TMcMeans:
The type of this PR is: feat

This PR solves [GRO-861]

### Description 
Screen Shot 2022-04-25 at 4 15 16 PM

[GRO-861]: https://artsyproduct.atlassian.net/browse/GRO-861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonallured:
> The feature flag isn't being recognized

How are you seeing this error? The fact that CI is green on `relay-check` leads me to believe you got this right? How do you know it's not working?

TMcMeans:
> 
im having this issue when testing in localhost:
Screen Shot 2022-04-22 at 1 23 17 PM

jonallured:
I think there might be a terminology misunderstanding here. When it comes to Force a `feature flag` controls what a user can do. We use them for a few different things like only allowing admins to see things during the development process and then flip it to more people when we're ready to take something live to production.

The `includeFallbackArtists` flag you've added to the query for related artists is not related to this. My guess is that you're getting that error because you local copy of Force is missing...something, I'm not sure what exactly haha. Next week we can get you squared away but what I'd recommend is going back to latest main and get that working before adding in the changes that this PR is for.

TMcMeans:
so I think the screenshot I posted was an unrelated issue, but I am still confused as to what to call flags for APIs that turn on/off data payloads? Also, still stuck on this issue and not sure how to resolve.

jonallured:
Nice one let's do this!

PR #9956, opened 2022-04-25
feat: new international shipping quotes

leamotta:
Part of https://artsyproduct.atlassian.net/browse/PX-4961

Added new quotes and descriptions for international shipping, plus using the mapping based on the quote key instead of the display name

Screenshot 2022-04-27 at 14 23 37


ansor4 (approved):
clean!

PR #9962, opened 2022-04-25
fix: updated logic of displaying "Create Alert" button on artwork page

dimatretyak:

### Description
[Notion card](https://www.notion.so/artsy/Create-alert-CTA-is-duplicated-on-bidding-closed-artwork-the-main-CTA-should-remain-the-one-belo-67e24932b88b4cbab0084909ac9b7710)

#### Before
before

#### After
after


src/v2/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -37,13 +38,17 @@ export const ArtworkSidebar: React.FC<ArtworkSidebarProps> = ({
   )
 
   // If we have info about the lot end time (cascading), use that.
-  const { sale, saleArtwork } = artwork
+  const { sale, saleArtwork, is_sold, is_in_auction } = artwork
   const endAt = saleArtwork?.endAt
   const startAt = sale?.startAt
   const { hasEnded } = useTimer(endAt!, startAt!)
 
+  const shouldSkipCreateAlert =
```
dblandin, line 46: **suggestion**: I wonder if this might be misinterpreted in the future because we still have a create alert CTA under these conditions, it's just rendered somewhere else. Maybe something like `shouldHideDetailsCreateAlertCTA` might be more clear/verbose?

dblandin (approved):
LGTM! Just left a small naming suggestion.

src/v2/Apps/Artwork/Components/ArtworkSidebar/index.tsx
dimatretyak: Yeah, I like this name 👍

PR #9968, opened 2022-04-25
fix: revert yarn/node orb version

artsyjian:
bumped these orbs version to latest in https://github.com/artsy/force/pull/9947, not realizing they carry node version. undo.


ovasdi (approved):
🦖

PR #9982, opened 2022-04-26
[WIP] feat(ArtworkFilter): Make filter URL driven so that browser back buttons preserve previous selections

damassi:

This PR solves [GRO-824]

### Description

This updates our Artwork filter so that we can drive via the back / forward buttons on the browser. Previously when a user made a selection and then hit the back button it would take them away from their previous selection, versus pushing it into the browsers history stack. This fixes that. 

Also fixes a bunch of random UX issues like the page loader post filter select being stuck at the top of the page rather than in the viewport, and the zerostate doing the same. Now they will stay within the users scroll position eliminating confusion. 

[GRO-824]: https://artsyproduct.atlassian.net/browse/GRO-824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Collect/Routes/Collect/index.tsx
+++ b/src/v2/Apps/Collect/Routes/Collect/index.tsx
@@ -38,7 +38,7 @@ export const CollectApp: React.FC<CollectAppProps> = ({
   match: { location, params },
   viewer,
 }) => {
-  const { router } = useRouter()
+  const { router, match } = useRouter()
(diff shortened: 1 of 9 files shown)


damassi:
Closing in favor of pass 2: https://github.com/artsy/force/pull/10069

PR #9988, opened 2022-04-27
refactor: Consolidate logic and functionality on sale and lot timer and on web/app

annacarey:

This PR solves [PX-355](https://artsyproduct.atlassian.net/browse/BID-355), [BID-297](https://artsyproduct.atlassian.net/browse/BID-297), [BID-298](https://artsyproduct.atlassian.net/browse/BID-298), and [BID-302](https://artsyproduct.atlassian.net/browse/BID-302).

### Description

Makes logic to display sale and lot countdown timer on web the same as on the app (which should all be correct). This included small changes to copy and color and also refactoring the two helpers for the sale and lot timers to instead use one helper that is tested with all cases in one file. 

The changes include:
- The number of days counting down in sale detail page in app and web were 1 off, so now that's fixed
- Should say "Bidding starts today" when sale is <1 day from opening
- Lot timer should be red when countdown is less than one hour (previously was only red when < 2 mins)


mzikherman:
Ah you need a rebase!

src/v2/Utils/getSaleOrLotTimerInfo.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+interface TimerInfo {
+  copy: string
+  color: string
+}
+
+// If `extendedBiddingEndAt` is passed in and non-null, the `time`
+// parameter refers to the time left in the extended period.
+export const getSaleOrLotTimerInfo = (
```
mzikherman, line 15: I would change the signature of this method and pass these in via a `params` or `options` object. Right now it's a list of things, which includes a mix of optional and not-optional arguments w/ defaults. That may make it a bit easier to use/pass in values (plus I think there's a related spec failure).

src/v2/Apps/Auction/Components/AuctionDetails/SaleDetailTimer.tsx
```diff
@@ -42,56 +43,56 @@ export const SaleDetailTimerFragmentContainer = createFragmentContainer(
   }
 )
 
-export const getTimerCopy = (time, hasStarted, lotsAreClosing) => {
-  const { days, hours, minutes, seconds } = time
+// export const getTimerCopy = (time, hasStarted, lotsAreClosing) => {
```
mzikherman, line 46: Can remove the commented stuff.

src/v2/Utils/getSaleOrLotTimerInfo.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  let copy = ""
+  let color = "blue100"
+
+  // Sale has not yet started
+  if (!hasStarted) {
+    if (parsedDays < 1) {
+      copy = "Bidding Starts Today"
+      // Entered extended bidding
+    } else if (extendedBiddingEndAt) {
```
mzikherman, line 37: This should likely be in the `else` section, if we are in the extended bidding period the sale has definitely already started.

src/v2/Utils/getSaleOrLotTimerInfo.ts
```diff
@@ -0,0 +1,91 @@
...
+    } else {
+      copy = `${parsedDays} Day${
+        parsedDays > 1 ? "s" : ""
+      } Until Bidding Starts`
+    }
+    // Sale has started
+  } else {
+    // When the timer is on the sale:
+    if (isSaleInfo) {
+      if (saleHasEnded) {
```
mzikherman, line 49: This line and the next kind of read oddly, if the sale has ended display 'Lots are closing'? I think in the use-case of a cascading sale, the `saleHasEnded` parameter _means_ that the lots have started closing. I wonder if that can be captured better via the grab-bag of parameters being passed in or parameter naming.

mzikherman: I believe the sale and artwork page usage of this , so in general, I think this component is only used when a sale has cascading enabled, so maybe the argument should be renamed to `isLotsClosing` or something like that.

src/v2/Utils/getSaleOrLotTimerInfo.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      // 1-24 hours until close
+      else if (parsedDays < 1 && parsedHours >= 1) {
+        copy = `${parsedHours}h ${parsedMinutes}m`
+      }
+
+      // <60 mins until close
+      else if (parsedDays < 1 && parsedHours < 1) {
+        copy = `${parsedMinutes}m ${parsedSeconds}s`
+        color = "red100"
```
mzikherman, line 85: I think this red state on the lot should just be when it's < 2 minutes (going from `LotTimer`).

src/v2/Utils/getSaleOrLotTimerInfo.ts
annacarey: Yeah `lotsAreClosing` is much better and I've used that elsewhere for this reason. I'll use that instead

src/v2/Utils/getSaleOrLotTimerInfo.ts
annacarey: Good call! This is getting confusing. ALSO it's gonna be so nice when all we have to do is make these exact super small changes to the eigen version for popcorn.

src/v2/Utils/getSaleOrLotTimerInfo.ts
annacarey: (Also added a test case for this)

src/v2/Utils/getSaleOrLotTimerInfo.ts
annacarey: No I had it wrong before. Timer should be red under 1 hour: https://www.figma.com/file/PD7gXzuexzbi4Zz1tZLssx/Popcorn-and-cascading-bidding?node-id=902%3A7045

annacarey: From Figma:

Screen Shot 2022-04-27 at 7 56 23 PM

mzikherman (approved):
🔥

PR #10008, opened 2022-04-29
WIP:chore:update artwork sidebar to respect extended bidding end at

nspinazz89:
The type of this PR is: Chore

This PR solves [BID-365]

### Description

Currently the front end does not respect extended bidding end at when determining whether to show a lot is closed or not. This PR addresses that partially by ensuring we use the extendedBiddingEndAt when possible to determine if the lot is still open. We will also need to do something with websockets or polling for a user who is sitting on a page, but did not just place a bid and would need a refresh to show that the lot is still open. Still need to add specs to this.


mzikherman (dismissed):
Nice, looks like the right fix.

PR #10014, opened 2022-05-02
feat(SWA-348): turn on new landing page

kiryl-zubarau:

This PR solves [SWA-348]

### Description

This PR turns on the new SWA landing page on the `/sell` route. Also, replaces `/consign` route to `/sell`.

[SWA-348]: https://artsyproduct.atlassian.net/browse/SWA-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Consign/consignRoutes.tsx
+++ b/src/v2/Apps/Consign/consignRoutes.tsx
@@ -15,15 +15,6 @@ const NewMarketingLandingApp = loadable(
   }
 )
 
-const MarketingLandingApp = loadable(
-  () =>
-    import(
-      /* webpackChunkName: "consignBundle" */ "./Routes/MarketingLanding/MarketingLandingApp"
-    ),
-  {
-    resolveComponent: component => component.MarketingLandingApp,
-  }
-)
 const OfferDetailApp = loadable(
   () =>
     import(
@@ -121,7 +112,7 @@ const prepareSubmissionFlowStepVariables = data => {
 
 export const consignRoutes: AppRouteConfig[] = [
   {
-    path: "/sell2",
+    path: "/sell",
     getComponent: () => NewMarketingLandingApp,
     onClientSideRender: () => {
       NewMarketingLandingApp.preload()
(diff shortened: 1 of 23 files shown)


src/v2/Apps/Consign/consignRoutes.tsx
```diff
@@ -121,26 +112,41 @@ const prepareSubmissionFlowStepVariables = data => {
...
+          )
+        },
+      },
+      {
+        path: "/",
+        render: ({ match }) => {
+          throw new RedirectException("/sell", 301)
+        },
+      },
+    ],
```
kiryl-zubarau, line 141: @damassi I used 2 subroutes `:splat` and `/` for the redirecting all requests starter with `/consign` to the new `/sell` route. I was thinking about using `/consign(.*)?` path for that but looks like it's not working because versions of `path-to-regexp` are different for the app and Express. In the app, we use `6.1.0`, Express uses `0.1.7`. I tested the route via http://forbeslindesay.github.io/express-route-tester/ and path `/consign(.*)?` match only `/consign` and not match paths like `/consign/foo/bar` for the `0.1.7` version, but works fine for the subsequent versions started from `1.7.0.`.
Looks like it is required to update the Express version to use `(.*)` expression.
Please let me know if you have a better solution for that.

kiryl-zubarau: Also, the problem can't be caught by unit tests because in the app we use an actual version of `path-to-regexp`

iskounen (approved):

src/v2/Apps/Consign/consignRoutes.tsx
kiryl-zubarau: Unfortunately, I was too optimistic, the latest Express version `4.18.1` still uses `path-to-regexp@0.1.7`. The `5.x` version is still in beta and doesn't use path-to-regexp at all. I'll check one more idea but not sure that will help with the problem.

PR #10025, opened 2022-05-03
feat(NX-3105): add fraud prevention banner

tam-kis:
The type of this PR is:  Feature 

This PR solves [NX-3105](https://artsyproduct.atlassian.net/browse/NX-3105)

### Description
Add fraud prevention banner and icon in bottom Artsy Guarantee banner:

Screenshot 2022-05-04 at 12 47 00


package.json
```diff
@@ -91,8 +91,8 @@
     "@artsy/fresnel": "3.5.0",
     "@artsy/gemup": "0.1.0",
     "@artsy/multienv": "^1.2.0",
-    "@artsy/palette": "19.5.1",
-    "@artsy/palette-charts": "18.5.1",
+    "@artsy/palette": "19.6.0-canary.1157.23661.0",
+    "@artsy/palette-charts": "18.6.0-canary.1157.23661.0",
```
lean8086, line 95: I think the canary version is meant only to test work in progress.

damassi:
If we can be sure to update the PR title to match [our convention](https://github.com/artsy/README/issues/327) 🙏

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -163,6 +171,15 @@ const Conversation: React.FC<ConversationProps> = props => {
 
   const [toastBottom, setToastBottom] = useState(0)
 
+  // Show banner
+  const [showBanner, setShowBanner] = useState(true)
+
+  useEffect(() => {
```
araujobarret, line 177: at the top of my head, I don't think this component needs to observe any data change to show the Toast (unless the component is never unmounted when we switch between conversations), I would expect something like:

```jsx
useEffect(() => {
  setShowBanner(true) // if we observe conversationID from the deps of useEffect
  const timer = setTimeout(() => setShowBanner(false), 5000)
  return () => clearTimeout(timer)
}, []) // or we use conversationID in the deps that we define at line 131
```
if you observe `conversationID` here you can start the initial state of `showBanner` as false, then once we have `conversationID` changes will be set to true 

another point, can you please check the timer of the banner? I recall talking to Charles from Eigen and he said 3 seconds, not 5

araujobarret (changes requested):

src/v2/Apps/Conversation/Components/Conversation.tsx
tam-kis: The issue is that the `conversationID` and the url changes every time we change a conversation, so we can see that with `selectedConversationID` . I will look into this option.

This task for Force requires 5 seconds, unless some other decision has been made in the meantime?

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -203,12 +220,21 @@ const Conversation: React.FC<ConversationProps> = props => {
         showDetails={showDetails}
         setShowDetails={setShowDetails}
       />
+      {showBanner && (
+        <GridColumns>
+          <Column start={4} span={6} mt={1}>
+            <Banner variant="brand">
```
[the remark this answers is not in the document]
tam-kis: `Toast` is unable to take an icon, it converts everything into a string, the `Banner` component is able to do that.

src/v2/Apps/Conversation/Components/Conversation.tsx
araujobarret: 5 seconds 👍 , I guess we just have a different time for Eigen I suppose, maybe because the screen is smaller

regarding updating the component I understand, but still, we don't need this extra parameter since we already have `conversationID` defined in the component 😉

src/v2/Apps/Conversation/Components/Conversation.tsx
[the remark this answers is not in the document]
araujobarret: Gotcha, thanks for the clarification 👍

src/v2/Apps/Conversation/Components/Conversation.tsx
tam-kis: I changed it to `conversationID` as suggested.

araujobarret (approved):
All good, please update the PR title as @damassi asked

araujobarret (approved):

PR #10032, opened 2022-05-04
Gkartalis/fx 3930/update navigation links

gkartalis:

This PR solves [FX-3930]

### Description

Updated the links under artworks and artists navigation tabs to be like the screenshots below.

check the [review-app here](https://updated-tab-navigation-links.artsy.net/)

Artworks Tab 
| Before | After | 
|---|---|
|Screenshot 2022-05-04 at 11 42 22|Screenshot 2022-05-04 at 11 40 58|

Artists Tab 
| Before | After | 
|---|---|
|Screenshot 2022-05-04 at 11 42 16|Screenshot 2022-05-04 at 11 40 49|

### DIFF
--- a/src/v2/Components/NavBar/menuData.ts
+++ b/src/v2/Components/NavBar/menuData.ts
@@ -32,34 +32,34 @@ export const ARTISTS_SUBMENU_DATA: MenuLinkData = {
     title: "Artists",
     links: [
       {
-        text: "Trending on Artsy",
-        href: "/collection/highlights-this-month",
+        text: "Trending This Week",
+        href: "/gene/trending-this-week",
       },
       {
-        text: "The Artsy Vanguard Artists",
-        href: "/collection/artsy-vanguard-artists",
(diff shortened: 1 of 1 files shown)


dblandin (dismissed):
LGTM! 👍🏼

src/v2/Components/NavBar/menuData.ts
```diff
@@ -155,16 +155,16 @@ export const ARTWORKS_SUBMENU_DATA: MenuLinkData = {
         href: "/gene/trove",
```
dimatretyak, line 155: **question**: should we change "Trovo" to "Trove: Editor’s Picks"?

FX-3930  Update Navigation Links - Jira 2022-05-04 13-50-58

gkartalis: Good catch @dimatretyak! Yes, I think I missed that

dimatretyak (approved):
Let's ship it 🚀

PR #10037, opened 2022-05-04
feat(auction): Add preview sale states 

damassi:

This PR solves [GRO-963], [GRO-899]

### Description

This updates /auction/id to account for preview states, showing a new status message as well as a rail of current auctions: 

Screen Shot 2022-05-04 at 4 46 53 PM
 
 cc @artsy/grow-devs 

[GRO-963]: https://artsyproduct.atlassian.net/browse/GRO-963?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-899]: https://artsyproduct.atlassian.net/browse/GRO-899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Components/Cells/CellSale.story.tsx
```diff
@@ -0,0 +1,54 @@
...
+          return null
+        }
+
+        if (!props?.viewer?.salesConnection) {
+          return <CellSalePlaceholder />
+        }
+
+        return (
+          <CellSaleFragmentContainer
+            sale={props.viewer?.salesConnection?.edges?.[0]?.node!}
```
damassi, line 44: Sales are published and unpublished all the time, so rather than fix things to an id (brittle), use a connection and grab the first item

jacobherrington (approved):

src/v2/Components/Cells/CellSale.story.tsx
jacobherrington: **suggestion (non-blocking):** A comment might be useful here; it may be difficult or impossible to find this PR comment if that context is handy. 🤷

damassi: It's a storybook story, so not too worried!

PR #10045, opened 2022-05-05
docs(unleash): Add section on accessing flags on server

damassi:
The type of this PR is: **Documentation**

### Description

Noticed that the Unleash docs lacked a section on how to access on the server, so added it!


docs/unleash_a-b_testing.md
```diff
@@ -112,6 +112,26 @@ function MyComponent() {
...
+If needing to access a feature flag on the server before rendering any components, all feature flags are located within local sharify data.
+
+#### Example
+
+```tsx
+const routes = [
+  {
+    path: "/foo",
+    onServerSideRender: ({ res }) => {
+      if (!res.locals.sd.FEATURE_FLAGS.myFlag) {
```
damassi, line 126: @mc-jones - we might want to expand this a bit, but this should communicate the point (i think)

PR #10048, opened 2022-05-06
chore(deps): update dep typescript from 4.3.5 to v4.6.4

gkartalis:

This PR solves [FX-3933]

### Description

- updated typescript from `4.3.5` -> `4.6.4`
- updated relay-compiler-language-typescript from `13.0.3` -> `15.0.0` (due to [this issue](https://github.com/facebook/relay/issues/3649) that I encountered after the TS upgrade)
- regenerated autogenerated files after 👆
- use `getENV` over `sd` in [metaphysics.ts](https://github.com/artsy/force/compare/gkartalis/update-typescript?expand=1#diff-d5b6c2f146a6bbb904bc6a105437e23d4c8ae521997affa0095b53f2c021ff97R2-R11)
- added `eslint-disable-next-line no-restricted-imports` for `superagent` 

cc @artsy/fx-devs 

[FX-3933]: https://artsyproduct.atlassian.net/browse/FX-3933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gkartalis:
Weird that `yarn type-check` is passing locally and failing on the ci 🤔

mdole:
@gkartalis I'm not sure what the solution is exactly but just remembered seeing this thread on the type error you're seeing: https://artsy.slack.com/archives/CP9P4KR35/p1651167945989559

Happy to dig in deeper on Monday if you could use a hand!!

gkartalis:
thanks @mdole ! Sure We can have a look on Monday!

mdole:
@gkartalis just took another look at this, and it's strange because that whole folder should be ignored since Chris added [this](https://github.com/artsy/force/blob/main/tsconfig.json#L41) to the tsconfig.json recently.

I can see in the build step that we at least _seem_ to be copying over the tsconfig file (screenshot below is where we do the copying; the actual mention of tsconfig.json is scrolled way to the right). Maybe try rerunning the job with SSH and confirm that the file exists and contains the values you expect it to?


Screen Shot 2022-05-09 at 3 06 23 PM

gkartalis:
So, together with @mdole we logged in via ssh in circleCI and did the following: 

- confirmed that the tsconfig exists and it includes the exclude `src/lib/passport-local-with-otp` line.
- run `yarn type-check` which succeeded 🤯 

@damassi do you have any idea on what is going on here?

damassi:
@gkartalis - Type-check is 🍏 ! Nothing like converting files from `.js` to `.ts` while still ignoring the folder in `tsconfig` 😄. I suspect there was something going on with the circle cache, and renaming the files busted it.

damassi:
Merging! Thanks again for diving into this tricky lib upgrade @gkartalis and @mdole 🙏

PR #10068, opened 2022-05-12
feat(NX-3138): Adjust checkout redirect behavior for Purchase from inquiry

lean8086:

This PR solves [NX-3138]

### Description

- Implements redirect for purchase for inquiry
- Tidies `if` statements on the Review component

@artsy/negotiate-devs 

[NX-3138]: https://artsyproduct.atlassian.net/browse/NX-3138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Routes/Review/index.tsx
+++ b/src/v2/Apps/Order/Routes/Review/index.tsx
@@ -123,22 +123,26 @@ export class ReviewRoute extends Component<ReviewProps> {
           })
       } else {
         const { order, router, isEigen } = this.props
-        // Buy-mode order redirects to the status page. Eigen must keep the user inside the webview.
-        if (order.mode !== "OFFER" || isEigen) {
+        // Make offer and Purchase order redirects to the status page for Eigen, it must keep the user inside the webview.
+        if (isEigen) {
           return router.push(`/orders/${order.internalID}/status`)
         }
-        // Make offer in inquiry redirects to the conversation page
+        // Make offer and Purchase in inquiry redirects to the conversation page
         if (order.source === "inquiry") {
           return router.push(
             `/user/conversations/${order.conversation?.internalID}`
           )
         }
-        // Make offer from artwork page redirects to the artwork page
         const artworkId = get(
           order,
           o => o.lineItems?.edges?.[0]?.node?.artwork?.slug
         )
-        return router.push(`/artwork/${artworkId}?order-submitted=true`)
+        // Make offer from artwork page redirects to the artwork page and show modal
+        if (order.mode === "OFFER") {
+          return router.push(`/artwork/${artworkId}?order-submitted=true`)
+        }
+        // Purchase from artwork page redirects to the artwork page
+        return router.push(`/artwork/${artworkId}`)
       }
     } catch (error) {
       logger.error(error)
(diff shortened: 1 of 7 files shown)


src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -122,23 +122,32 @@ export class ReviewRoute extends Component<ReviewProps> {
...
-        // Make offer in inquiry redirects to the conversation page
-        if (order.source === "inquiry") {
-          return router.push(
-            `/user/conversations/${order.conversation?.internalID}`
-          )
-        }
-        // Make offer from artwork page redirects to the artwork page
+        const { order, router, isEigen, featureFlags } = this.props
+        const isCBNEnabled =
+          featureFlags?.["conversational-buy-now"]?.flagEnabled
```
lean8086, line 127: I found no examples of unleash's feature flags implemented on class components. I'm reading the `featureFlags` added by the `withSystemContext()`

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -122,23 +122,32 @@ export class ReviewRoute extends Component<ReviewProps> {
...
+        const orderId = order.internalID
+        const conversationId = order.conversation?.internalID
```
lean8086, line 129: There is no `order.conversation` on `CommerceBuyOrder`. To make this work:
1. Add `conversation` to MP stitching ([see here](https://github.com/artsy/metaphysics/blob/main/src/lib/stitching/exchange/v2/stitching.ts#L222-L239))
2. Add `conversation` to the Review query:
```
... on CommerceBuyOrder {
  conversation {
    internalID
  }
}
```

src/v2/Apps/Order/Routes/Review/index.tsx
```diff
@@ -122,23 +122,34 @@ export class ReviewRoute extends Component<ReviewProps> {
...
+          return router.push(`/user/conversations/${conversationId}`)
+        }
+        // Make offer from the artwork page redirects to the artwork page with a confirmation modal
+        if (order.mode === "OFFER") {
+          return router.push(`/artwork/${artworkId}?order-submitted=true`)
+        }
+        // Purchase from the artwork page redirects to the status page
+        if (isCBNEnabled) {
+          return router.push(`/orders/${orderId}/status`)
+        }
```
araujobarret, line 152: isn't missing a default return here like `return router.push(`/artwork/${artworkId}?order-submitted=true`)` ?

lean8086: That was the default before because there were no more use cases after the order was a buy order or from an inquiry.
Now that we need to cover all of them (BN, MO, MOI, BNI) the default return ended up being:
```
return router.push(`/orders/${orderId}/status`)
```
Keep in mind the feature flag will disappear from line 150.

araujobarret: I was more aware if we're covering all cases with the ifs, if we're not these cases we could introduce an issue but if you're sure I guess it's okay to go 🐎

araujobarret (approved):

PR #10070, opened 2022-05-12
feat(artwork-filter): Making loading/zero states sticky

damassi:

### Description

This updates our artwork filter so that the zero state / loading indicator is sticky, preserving the position in the users viewport even if they've scrolled down. (It was confusing to click a filter and to see nothing happen, especially if there are no results and on mobile.)
 
filter

cc @artsy/grow-devs

### DIFF
--- a/src/v2/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
+++ b/src/v2/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
@@ -1,4 +1,4 @@
-import { Spacer, useThemeConfig } from "@artsy/palette"
+import { Box, Spacer, Spinner, useThemeConfig } from "@artsy/palette"
 import * as React from "react"
 import { RelayProp, createFragmentContainer, graphql } from "react-relay"
 import { ArtworkFilterArtworkGrid_filtered_artworks } from "v2/__generated__/ArtworkFilterArtworkGrid_filtered_artworks.graphql"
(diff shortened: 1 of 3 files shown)


src/v2/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
```diff
@@ -127,3 +127,22 @@ export const ArtworkFilterArtworkGridRefetchContainer = createFragmentContainer(
...
+        <Sticky>
+          <Box height={300} width="100%" position="absolute">
+            <Spinner />
+          </Box>
+        </Sticky>
+      )}
+
+      <Box opacity={isLoading ? 0.1 : 1}>{children}</Box>
+    </Box>
+  )
```
dzucconi, line 147: It's always bothered me that this wasn't a skeleton?

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
```diff
@@ -1,31 +1,41 @@
...
+                  </ResetFilterLink>
+                  .
+                </>
+              )}
+            </Message>
+          </Box>
+        )
+      }}
+    </Sticky>
+  </Box>
```
dzucconi, line 38: I'd give this some kind of margin

src/v2/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
damassi: Prob for a separate PR

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
damassi: Good call

src/v2/Components/ArtworkFilter/ArtworkFilterArtworkGrid.tsx
laurabeth: It bothers me that we have a component named skeleton. 👀

laurabeth (approved):
Looks good. :shipit:

PR #10071, opened 2022-05-12
feat: Temporarily suppress logged out hero GRO-1025

jonallured:
We have a contract with a partner about how the carousel on the homepage behaves - it was supposed to always show an ad for a particular auction but we forgot about the signed out user. This PR does the absolutely bare minimum to suppress that item because I believe once that sale ends we'll want to revert. Looks like this:

Screen Shot 2022-05-12 at 1 16 16 PM

https://artsyproduct.atlassian.net/browse/GRO-1025

/cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnits.tsx
+++ b/src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnits.tsx
@@ -16,6 +16,8 @@ interface HomeHeroUnitsProps {
 }
 
 const HomeHeroUnits: React.FC<HomeHeroUnitsProps> = ({ homePage }) => {
+  // TODO: Remove after SSS sale ends
+  const skipHomeHero = true
   const { isLoggedIn } = useSystemContext()
 
   const heroUnits = [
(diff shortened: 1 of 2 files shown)


src/v2/Apps/Home/Components/HomeHeroUnits/HomeHeroUnits.tsx
```diff
@@ -29,6 +30,8 @@ const HomeHeroUnits: React.FC<HomeHeroUnitsProps> = ({ homePage }) => {
         // If logged out, the first hero unit is static and
         // needs to be rendered without the fragment container
         if (!isLoggedIn && i === 0) {
+          if (skipHomeHero) return
```
dzucconi, line 33: Could you add a comment like `TODO: Remove after <whateverdate>`

jonallured: Done!

dzucconi: 🙏

dzucconi (approved):

PR #10074, opened 2022-05-12
fix(article): only preloads necessary images

dzucconi:
I had noticed that this actually made image loading worse. Few things:

* I was preloading _all_ of the images, which is bad enough
* I was preloading every version of every image, which is worse
* Gemini would kind of choke sometimes returning broken images (request queueing + timeouts?)

This fixes it so that we're only ever loading the next and previous images. And only loading the versions that match the actually active media query. Doing preload on responsive picture tags is pretty annoying ATM: https://web.dev/preload-responsive-images/#preload-and-lesspicturegreater


package.json
```diff
@@ -172,6 +172,7 @@
     "jwt-decode": "2.2.0",
     "knox": "0.9.2",
     "lodash": "4.17.21",
+    "map-cursor-to-max": "^1.0.0",
```
dzucconi, line 175: This is already a dep via `use-cursor` just adding it explicitly.

src/v2/Apps/Article/Components/ArticleZoomGallery/ArticleZoomGalleryFigure.tsx
```diff
@@ -31,27 +35,42 @@ const ArticleZoomGalleryFigure: FC<ArticleZoomGalleryFigureProps> = ({
...
+      preload: "(min-width: 767px) and (max-width: 895px)",
+    },
   }
 
   const xs = {
     img: resized(src, { width: 450, height: 450 }),
-    media: "(max-width: 766px)",
+    media: {
+      query: "(max-width: 766px)",
+      preload: "(max-width: 766px)",
```
damassi, line 72: 😓 😓 

It'd be nice to key this off of `THEME.breakpoints` if possible, but all good if there's not a reason to abstract right now.

PR #9948, opened 2022-04-21
feat(press): adds press2 app

dzucconi:
Adds the press2 app. These pages are backed by two Markdown pages managed in Torque:

https://admin.artsy.net/page/in-the-media
https://admin.artsy.net/page/news-and-press-releases

Which I made by converting the JSON blobs from the JSON pages into Markdown.

https://artsy-force-production.s3.amazonaws.com/json/in-the-media.json
https://artsy-force-production.s3.amazonaws.com/json/press-releases.json

https://user-images.githubusercontent.com/112297/164546873-7fa75ba6-78f5-42ee-9194-68c2d0843e37.mov

PR #10076, opened 2022-05-12
chore(press): migrates /press2 => /press; deletes old press app

dzucconi:
Resolves [GRO-1023](https://artsyproduct.atlassian.net/browse/GRO-1023)

This PR migrates the press2 app to production and deletes the old CoffeeScript application.

See: https://github.com/artsy/force/pull/9948 for details on this migration (JSON => Markdown) and where to manage the pages.


dzucconi:
The `/press` redirect is already in the `hardcodedRedirects`: https://github.com/artsy/force/blob/28f80c2e9f8fc557e84370adee678e11b8a6ab6d/src/lib/middleware/hardcodedRedirects.ts#L24

PR #10083, opened 2022-05-13
[Documentation]: ExampleApp improvements

AlicanAkyuz:
The type of this PR is: **Documentation**

This PR solves [WP-67]

### Description

This PR adds example test setup for the ExampleApp using RTL and improves the visibility of the app in our docs. Additionally, it aims to use palette v3 fully within the app. 

[WP-67]: https://artsyproduct.atlassian.net/browse/WP-67?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Example/README.md
+++ b/src/v2/Apps/Example/README.md
@@ -7,7 +7,7 @@ It includes:
 - [x] Routing
 - [x] Relay
 - [x] @principalField
-- [ ] Testing
+- [x] Testing
 - [ ] Analytics
 - [ ] Mention of adding tests to integrity
 
--- a/src/v2/Apps/Example/README.md
+++ b/src/v2/Apps/Example/README.md
@@ -8,7 +8,7 @@ It includes:
 - [x] Relay
 - [x] @principalField
 - [x] Testing
-- [ ] Analytics
+- [x] Analytics
 - [ ] Mention of adding tests to integrity
 
 ## Launch checklist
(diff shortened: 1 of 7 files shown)


src/v2/Apps/Example/README.md
```diff
@@ -7,8 +7,8 @@ It includes:
 - [x] Routing
 - [x] Relay
 - [x] @principalField
-- [ ] Testing
-- [ ] Analytics
+- [x] Testing
+- [x] Analytics
```
AlicanAkyuz, line 11: marked this as done as there are examples of tracking in the app

src/v2/Apps/Example/__tests__/ExampleApp.jest.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { screen } from "@testing-library/react"
+import { findMetaTagBySelector } from "v2/DevTools"
+
+jest.unmock("react-relay")
+
+const { renderWithRelay } = setupTestWrapperTL<ExampleApp_Test_Query>({
+  Component: props => {
+    return (
+      <MockBoot>
+        {/* @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION */}
```
dzucconi, line 15: You'd just use `@ts-ignore` if you wanted to ignore this. This form is from the migration. To actually fix this problem you can just do something like `if (!props.system) return null` or whatever it's complaining about

dzucconi: (Not that it matters a ton in this test code 🤷 )

AlicanAkyuz: thanks for the hint 🙏

damassi:
Thank you!

PR #10091, opened 2022-05-13
feature(NewForYou): implement artwork grid w/ artworksForUser query

laurabeth:

This PR solves [GRO-918] & [GRO-920]

### Description

Add the new for you grid structure & associated query (which will fetch backfill artworks if the user has none). 
Backfill work is in flight, per @jonallured.

[GRO-918]: https://artsyproduct.atlassian.net/browse/GRO-918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-920]: https://artsyproduct.atlassian.net/browse/GRO-920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
+++ b/src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
@@ -3,9 +3,7 @@ import React, { FC, Fragment } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import ArtworkGridItemFragmentContainer from "v2/Components/Artwork/GridItem"
 import { Masonry } from "v2/Components/Masonry"
-import { SystemQueryRenderer } from "v2/System/Relay/SystemQueryRenderer"
 import { extractNodes } from "v2/Utils/extractNodes"
-import { NewForYouArtworksGridQuery } from "v2/__generated__/NewForYouArtworksGridQuery.graphql"
 import { NewForYouArtworksGrid_viewer } from "v2/__generated__/NewForYouArtworksGrid_viewer.graphql"
 
 interface NewForYouArtworksGridProps {
@@ -15,8 +13,7 @@ interface NewForYouArtworksGridProps {
 export const NewForYouArtworksGrid: FC<NewForYouArtworksGridProps> = ({
   viewer,
 }) => {
-  const { artworksForUser } = viewer
-  const artworks = extractNodes(artworksForUser)
+  const artworks = extractNodes(viewer.artworksForUser)
 
   return (
     <Masonry columnCount={[2, 3, 4]}>
@@ -33,37 +30,12 @@ export const NewForYouArtworksGrid: FC<NewForYouArtworksGridProps> = ({
   )
 }
 
-export const NewForYouArtworksGridQueryRenderer: React.FC = () => {
-  return (
-    <SystemQueryRenderer<NewForYouArtworksGridQuery>
-      query={graphql`
-        query NewForYouArtworksGridQuery {
-          viewer {
-            ...NewForYouArtworksGrid_viewer
-          }
-        }
-      `}
-      render={({ props, error }) => {
-        if (error) {
-          console.error(error)
-          return null
-        }
-        if (!props?.viewer) {
-          return null
-        }
-        return <NewForYouArtworksGridFragmentContainer viewer={props.viewer} />
-      }}
-    />
-  )
-}
-
 export const NewForYouArtworksGridFragmentContainer = createFragmentContainer(
   NewForYouArtworksGrid,
   {
     viewer: graphql`
       fragment NewForYouArtworksGrid_viewer on Viewer {
         artworksForUser {
-          totalCount
           edges {
             node {
               internalID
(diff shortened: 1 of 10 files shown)


src/v2/Apps/NewForYou/NewForYouApp.tsx
```diff
@@ -31,6 +22,7 @@ export const NewForYouApp: FC = () => {
         </Column>
       </GridColumns>
       <Spacer mt={4} />
+      <NewForYouArtworksGridQueryRenderer />
```
damassi, line 25: No need to use a query renderer here, can just use the fragment container so that it properly renders on first load from the server. If there were below-the-fold areas, those we would use query renderers with with the `lazyLoad` prop attached.

damassi: Yeah, for something like this we don't need to _only_ render on the client; user is logged in, it's all very deliberate content for one specific user in particular and so performance tuning here is something of a premature optimization. Maybe @dzucconi can comment more on his thinking? 

Can swap out this line with 
```tsx
<NewForYouArtworksGridFragmentContainer viewer={viewer} /> 
```
And achieve the same ends, no other change necessary, only there will no longer be a loading flash and the page will be stable on first render.

dzucconi: Primarily the reason was to actually be able to debug the query because we had no insight into what the error was (MP  is swallowing error and apparently a malformed query can 500 one of the services (There is a bug there that should be fixed)). Moving it to the client-side allowed us to actually see what was happening.

damassi: This MP swallowing errors business that we're seeing all over lately is driving me crazy. I'm going to open a ticket that we can add to as we see them. 

And also, we should look into adding a flag that we can toggle to force client side rendering. Relay will fetch there if a short circuit is thrown into the rehydration data that is mounted on the window, even if there's been a server-side pass. I'll also ticket that.

damassi: Anything that y'all see going forward, please add here: https://artsyproduct.atlassian.net/jira/software/projects/WP/boards/97?issueParent=40613

dzucconi: It's less that MP is swallowing errors and more that errors aren't being handled

damassi: Updated the epic description slightly 👍

src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (!relay.hasMore() || relay.isLoading()) return
+
+    setLoading(true)
+
+    relay.loadMore(25, err => {
+      if (err) console.error(err)
+
+      setTimeout(() => {
+        setLoading(false)
+      }, 2500)
```
dzucconi, line 35: Why the arbitrary delay?

src/v2/Apps/NewForYou/__tests__/NewForYouApp.jest.tsx
```diff
@@ -15,9 +15,7 @@ describe("NewForYouApp", () => {
   it("renders", () => {
     render(
       <MockBoot breakpoint="lg">
-        <SystemContextProvider>
-          <NewForYouApp />
-        </SystemContextProvider>
+        <SystemContextProvider>{/* <NewForYouApp /> */}</SystemContextProvider>
```
damassi, line 18: (Happy to pair on the testing weirdness tomorrow)

src/v2/Apps/NewForYou/newForYouRoutes.tsx
```diff
@@ -1,14 +1,29 @@
...
+  resolveComponent: component => component.NewForYouAppFragmentContainer,
+})
 
 export const newForYouRoutes: AppRouteConfig[] = [
   {
     path: "/new-for-you",
     getComponent: () => NewForYouApp,
     onServerSideRender: ({ req, res }) => {
-      if (!req.user) {
+      if (!res.locals.sd.FEATURE_FLAGS.new_for_you || !req.user) {
```
damassi, line 14: Nice 👍

damassi (approved):
Looks good once the test is in 👍

src/v2/Apps/NewForYou/__tests__/NewForYouApp.jest.tsx
laurabeth: Sounds good!

src/v2/Apps/NewForYou/newForYouRoutes.tsx
laurabeth: Yeah, I think someone suggested this to me at some point. 🤔 😉

src/v2/Apps/NewForYou/__tests__/NewForYouApp.jest.tsx
laurabeth: Fixed.

src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
damassi: Not sure why I added that, but its probably fine to remove here!

jonallured:
Hrm the robots did not listen to that merge on green label so I'll push the button!

PR #10096, opened 2022-05-13
[WIP] Buyer sets bank debit payment method for an order

starsirius:

This PR solves [TX-325](https://artsyproduct.atlassian.net/browse/TX-325)

(Still work-in-progress but would love any feedback!)

### Description

This continues the happy path of the ACH payment and implements the user story:
```
As a buyer checking out an artwork with ACH payment,
After I click "save and continue" on the payment page,
I set the payment on the order and am taken to the review page.
```

Here is the (tentative) experience for a happy path and an error case. More details inline.

### Happy path

ach-set-payment

### Error path

ach-set-payment-error


src/v2/Apps/Order/Components/SetPayment.tsx
```diff
@@ -0,0 +1,46 @@
...
+      } catch (error) {
+        props.onError(error)
+      }
+    }
+
+    setPayment()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  return <div>Setting bank debit payment for this order...</div>
```
starsirius, line 45: TODO: this should be removed.

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -149,6 +151,20 @@ export const PaymentRoute: FC<Props> = props => {
     })
   }
 
+  const setupIntentId = getClientParam("setup_intent")
+  const isSettingPayment =
+    setupIntentId &&
+    getClientParam("setup_intent_client_secret") &&
+    getClientParam("redirect_status") === "succeeded"
```
starsirius, line 158: After redirecting back from Stripe, we get 3 query params. We'll use them as an indication to set payment behind the scene. We may consider making it stricter by checking order state or something.

src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -50,10 +52,14 @@ export const BankDebitForm: FC<Props> = ({ order }) => {
       return
     }
 
+    // Disable the "leave/reload site?" confirmation dialog as we're about to
+    // confirm Stripe payment setup which leaves and redirects back.
+    window.removeEventListener("beforeunload", preventHardReload)
```
starsirius, line 57: To avoid the confirmation alert from popping up during redirect.

Screen Shot 2022-05-13 at 3 55 45 PM

src/v2/Apps/Order/Components/SetPayment.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const { submitMutation: setPaymentMutation } = useSetPayment()
+
+  useEffect(() => {
+    const setPayment = async () => {
+      try {
+        const orderOrError = (
+          await setPaymentMutation({
+            variables: {
+              // TODO: Instead of sending `creditCardId`, we'll send
+              // `setupIntentId` when the mutation is ready.
```
kiryl-zubarau, line 22: The mutation has been updated so `bankAccountId` should now be available

src/v2/Apps/Order/Components/SetPayment.tsx
```diff
@@ -0,0 +1,46 @@
...
+          })
+        ).commerceSetPayment?.orderOrError
+
+        if (orderOrError?.error) {
+          throw orderOrError.error
+        }
+
+        props.onSuccess()
+      } catch (error) {
+        props.onError(error)
```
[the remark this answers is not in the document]
starsirius: yeah, errors can happen in different ways:

- A general exception will be caught and handled here.
- For Exchange order errors, we follow the [`errors as data`](https://graphql-ruby.org/mutations/mutation_errors#errors-as-data) pattern, and need to look into `orderOrError.error` and re-raise to be handled in a similar way.

src/v2/Apps/Order/Components/Mutations/useSetPayment.ts
```diff
@@ -0,0 +1,31 @@
...
+    mutation: graphql`
+      mutation useSetPaymentMutation($input: CommerceSetPaymentInput!) {
+        commerceSetPayment(input: $input) {
+          orderOrError {
+            ... on CommerceOrderWithMutationSuccess {
+              order {
+                id
+                creditCard {
+                  internalID
+                }
```
starsirius, line 16: Not sure if accessing credit card will cause errors but this can be removed now I believe.

starsirius:
> @starsirius the setPayment mutation is now working with bank debit 👍 However, we still can't get to the review step as the orderRoutes_ReviewQuery fails. Should we go ahead and merge this and I can work on making the review step compatible in my next PR?

@rquartararo Yes let's do that! Thank you for taking this through the finish line! 🙏 

Left a question about [accessing credit cards](https://github.com/artsy/force/pull/10096/files#r878935946). Let's merge when ready and iterate!

PR #10102, opened 2022-05-16
feat(TX-297): account balance check 

AlicanAkyuz:

This PR solves [TX-297]

### Description

This PR introduces the check for users' bank account balance in the payment step of the checkout flow. 
- account balance is polled periodically ❌ 
- while account balance check is in progress, a loading interface is displayed ✅ 
- when account does not have enough funds, designed error view is displayed ✅ 
- tests are updated ❌ 

[TX-297]: https://artsyproduct.atlassian.net/browse/TX-297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/BankDebitForm/BankDebitForm.tsx
+++ b/src/v2/Components/BankDebitForm/BankDebitForm.tsx
@@ -2,24 +2,26 @@ import { FC, useState } from "react"
 import { PaymentElement, useStripe, useElements } from "@stripe/react-stripe-js"
 import { Box, Button, Message, Spacer, Text } from "@artsy/palette"
 import { useSystemContext, useTracking } from "v2/System"
-import { useRouter } from "v2/System/Router/useRouter"
 import { LoadingArea } from "../LoadingArea"
 import { preventHardReload } from "v2/Apps/Order/OrderApp"
 
 interface Props {
   order: { mode: string | null; internalID: string }
   returnURL: string
+  isUserHasEnoughFunds: boolean
 }
 
-export const BankDebitForm: FC<Props> = ({ order, returnURL }) => {
+export const BankDebitForm: FC<Props> = ({
+  order,
+  returnURL,
+  isUserHasEnoughFunds,
+}) => {
   const stripe = useStripe()
   const elements = useElements()
   const tracking = useTracking()
-  const { router } = useRouter()
   const { user } = useSystemContext()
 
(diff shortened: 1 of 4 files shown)


src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -45,20 +52,34 @@ export const BankDebitForm: FC<Props> = ({ order }) => {
...
+        },
+      })
+
+      if (setupIntent?.status === "succeeded") {
+        // TODO: poll account balance and set state accordingly
+        setIsInsufficientFunds(true)
 
-    if (error) {
-      console.log("error", error)
+        router.push(`/orders/${order.internalID}/review`)
```
AlicanAkyuz, line 74: @starsirius what do you think about routing like here, instead of providing stripe with a return_url? (since we want to poll the balance before going to the next step).

starsirius: Sorry for the slowness. We asked Stripe about redirect or not, and they highly recommend us to _not_ use the `if_required` parameter because

- some payment methods require redirect (including some banks for ACH, actually) and if we use `if_required` we will need to build and maintain 2 different user experiences.
- it might disrupt compatibility with future payment methods.

The doc mentions:

> Note: Setting if_required requires that you handle successful confirmations for redirect-based and non-redirect based payment methods separately.

src/v2/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -45,20 +52,34 @@ export const BankDebitForm: FC<Props> = ({ order }) => {
...
+      const { setupIntent, error } = await stripe.confirmSetup({
+        elements,
+        redirect: "if_required",
+        confirmParams: {
+          return_url: "https://staging.artsy.net/",
+        },
+      })
+
+      if (setupIntent?.status === "succeeded") {
+        // TODO: poll account balance and set state accordingly
```
starsirius, line 71: Based on the [doc](https://stripe.com/docs/payments/ach-debit/set-up-payment?platform=web#access-data-on-a-financial-connections-bank-account), we might want to poll the account balance _before_ we call `confirmSetup`.

> If you opted to receive balances permissions, we recommend checking a balance at this stage to verify sufficient funds before confirming a payment.

PR #10107, opened 2022-05-16
Updates typography tokens to use 3.1 scale

dzucconi:
Re: https://github.com/artsy/palette/pull/1164

Notably: `md` becomes `sm-display` and `lg` becomes `lg-display`.

This PR will set up the review app. Will post link here once spun up.


dzucconi:
(Ah, I don't need to add the tokens package.)

dzucconi:
https://typography-update.artsy.net/

PR #10109, opened 2022-05-17
[DO NOT MERGE]:(feat): AlicanAkyuz/TX-332/Wire transfer

AlicanAkyuz:

This PR solves [TX-332]

### Description

This PR adds wire transfer as an option to the payment step of the checkout flow. 

#### Video 
https://user-images.githubusercontent.com/42584148/168848774-82ab30a0-a2cf-401d-88b2-fcdb5ba4079f.mov

[TX-332]: https://artsyproduct.atlassian.net/browse/TX-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
+++ b/src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
@@ -1,4 +1,4 @@
-import { FC, RefObject, useState } from "react"
+import React, { FC, RefObject, useState } from "react"
 import {
   Button,
   Clickable,
@@ -8,6 +8,7 @@ import {
   Spacer,
   BorderedRadio,
   RadioGroup,
+  RadioProps,
   Text,
   Collapse,
 } from "@artsy/palette"
@@ -23,7 +24,9 @@ import { Payment_order } from "v2/__generated__/Payment_order.graphql"
 import { CommitMutation } from "../../Utils/commitMutation"
 import { useTracking } from "v2/System"
 import { useRouter } from "v2/System/Router/useRouter"
+import { useFeatureFlag } from "v2/System/useFeatureFlag"
 import { ActionType, OwnerType } from "@artsy/cohesion"
+import { extractNodes } from "v2/Utils/extractNodes"
 
 enum PaymentMethods {
   CreditCard = "credit_card",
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -56,59 +86,137 @@ export const PaymentContent: FC<Props> = props => {
...
+      </Collapse>
+    </Flex>
   )
 }
+
+/*
+returns all available payment methods, by checking relevant feature flags
+TODO: when ACH and wire_transfer FFs is removed, this function can be removed and radios can be moved to PaymentContent
+*/
+const getAvailablePaymentMethods = (
```
rquartararo, line 177: The two-step "Payment Method" and Payment Details" will need to continue to be behind the feature flag until either ACH or wire transfer is released, so I'm sure if this would work in this case as it still shows the radio buttons when neither flag is enabled.

rquartararo:
Overall looking great! 👍  Just left a comment about the feature flagging.

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
rquartararo: Maybe we could have the third flag for "new payment methods" which job it to flag the two-step flow?

rquartararo: Sorry second thought, we might need to check with a designer for this, but if the cases where there is only a single payment method available means there is no two-step process, we could just remove the first step for those cases.

AlicanAkyuz: thank you for flagging this, I see what you mean. 

I think it makes sense to keep the original view (no collapse, just credit card), when neither of the other two is available. please let me know if you disagree.

rquartararo: Yes, I agree!

rquartararo: Yep sounds good to me!

rquartararo:
While testing locally I noticed that the wire transfer flag is set to true, but I don't believe my account has been added to the unleash flag. Maybe I'm missing something?

AlicanAkyuz:
ah yeah it is expected, the flag for wire transfer is set by environment, so it'll be true for everyone in development and false on production.

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -229,6 +204,11 @@ export const PaymentFragmentContainer = createFragmentContainer(
             node {
               artwork {
                 slug
+                partner {
+                  name
+                  # TODO: when it is available, pull below field to check if wire T enabled for partner
+                  # wireTransferEnabled
+                }
```
iskounen, line 211: @starsirius left a [suggestion](https://github.com/artsy/metaphysics/pull/4065#discussion_r876006119) on the PR that @kiryl-zubarau opened to include this flag in Metaphysics. Depending on how that goes, this field could be available on an `order` instead of a `partner`.

PR #10113, opened 2022-05-17
feat(GRO-1028): Onboarding flow implementation

TMcMeans:
The type of this PR is: feat

This PR solves [GRO-1028]

In prepping for the new Onboarding, this code is an investigative spike in setting up a new workflow using the `Inquiry` flow engine as a framework. 

[GRO-1028]: https://artsyproduct.atlassian.net/browse/GRO-1028?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,70 @@
+import { Engine } from "v2/Utils/Engine"
```
damassi, line 1: We should rename this file / variable to something like `DecisionEngine`

TMcMeans: @damassi is the current naming too vague?

damassi: Yep, "engine" could mean a thousand different things

dzucconi: Yeah, good point — I vote for `WorkflowEngine` which I think is the more accurate term of art.

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,70 @@
...
+                "KeepTrackOfInterests",
+                {
+                  whereWouldYouLikeToDiveIn: {
+                    findNewFromArtistsICollect: ["FindNewFromArtistsICollect"],
+                    followGalleriesILove: ["FollowGalleriesILove"],
+                    createArtWishlist: ["CreateArtWishlist"],
+                  },
+                },
+              ],
+            },
```
dzucconi, line 52: Note to self for when we pick it back up on Thursday. I think we might be able to flatten this down into two `views` followed by all of the outcome leafs under a single condition. Which should simplify the flow a little bit.

dzucconi:
https://user-images.githubusercontent.com/112297/169576919-e585f85a-3952-4b38-9de7-5dd405a762cb.mov

src/v2/Apps/Onboarding/OnboardingApp.tsx
```diff
@@ -0,0 +1,108 @@
...
+    case "Done":
+      return <OnboardingDone />
+
+    case VIEW_AUCTION_HIGHLIGHTS:
+    case VIEW_CURATED_ARTWORKS:
+    case VIEW_SEARCH_ARTISTS:
+    case VIEW_SEARCH_ARTWORKS:
+    case VIEW_SEARCH_GALLERIES:
+    case VIEW_TRENDING_ARTISTS:
+    case VIEW_TRENDING_LOTS:
```
dzucconi, line 75: So from here: We should be able to just make components like `OnboardingAuctionHighlights`, `OnboardingSearchGalleries`, etc, and just slot them in.

dzucconi: None of those components need to know anything about the external state other than to call the `next` function to proceed.

TMcMeans: yup yup :)

src/v2/Apps/Onboarding/Views/OnboardingDone.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <Spinner position="static" color="brand" />
+
+      <Spacer mt={2} />
+
+      <Text variant="lg">
+        Great start!
+        <br />
+        We’re adding these to your Artsy home.
+      </Text>
+    </Flex>
```
dzucconi, line 32: From the designs it looks like we'll have different confirmation steps depending on what screen there was; so there probably won't be a single `Done` step anymore.

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
```diff
@@ -0,0 +1,99 @@
...
+    OPTION_KEEP_TRACK_OF_ARTIST_CAREER,
+    OPTION_CURRENTLY_TRENDING_ARTISTS,
+    OPTION_POPULAR_AUCTION_LOTS,
+  ],
+  [OPTION_COLLECTING_ART_THAT_MOVES_ME]: [
+    OPTION_FOLLOW_ARTISTS_I_COLLECT,
+    OPTION_FOLLOW_GALLERIES_I_LOVE,
+    OPTION_CREATE_AN_ART_WISHLIST,
+  ],
+}
```
dzucconi, line 99: Here the question 2 options are just keyed off of the question 1 options. So depending on what you answer there you get a different set of answers here.

dzucconi:
@damassi take another look here. Should be fine to merge.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f2769dcf438f34658304b87cb7f957e3fd550182
</p>

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+          {options.map(option => {
+            return (
+              <Button
+                key={option}
+                size="small"
+                variant="secondaryOutline"
+                onClick={() => {
+                  setAnswers(
+                    prevAnswers => [...prevAnswers, option] as [string, string]
```
damassi, line 60: Might be nice if this `setAnswers` handled the `...prevAnswers` concatination under the hood?

src/v2/Apps/Onboarding/OnboardingApp.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+      <Spacer mt={2} />
+
+      <OnboardingSteps />
+    </OnboardingProvider>
+  )
+}
+
+const OnboardingSteps: FC = () => {
+  const { current, next } = useOnboardingContext()
```
damassi, line 42: non-blocking suggestion: `current` / `next` would be friendlier if they were like `currentStep`, `nextStep`

src/v2/Apps/Onboarding/OnboardingApp.tsx
damassi: Will the switch `case`s above this be refactored into this format? While I lack a bit of context on this, its a bit confusing to see this form
```
case "WhereWouldYouLikeToDiveIn"
```
right alongside this form
```
case VIEW_AUCTION_HIGHLIGHTS:
```
Whats the difference?

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
```diff
@@ -0,0 +1,99 @@
...
+  OPTION_FOLLOW_GALLERIES_I_LOVE,
+  OPTION_GROW_MY_TASTE_IN_ART,
+  OPTION_KEEPING_TRACK_OF_ART_I_AM_INTERESTED_IN,
+  OPTION_KEEP_TRACK_OF_ARTIST_CAREER,
+  OPTION_POPULAR_AUCTION_LOTS,
+} from "../config"
+import { useOnboardingContext } from "../useOnboardingContext"
+
+export const OnboardingQuestionTwo: FC = () => {
+  const { answers, setAnswers, setContext, next } = useOnboardingContext()
```
damassi, line 30: Can we rename `setContext` to something more decision-engine specific? There's simply too much baggage associated with that word (`context`) in a react app, and at a glance its hard to understand what its used for (versus `setAnswers` which.. sets answers).

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const options = QUESTION_2[answers[0]!] as string[]
```
damassi, line 32: 😅

damassi: Things like this should be abstracted a bit if possible. A future dev will have a hard time understanding how this was arrived at.

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
+import { useRef, useState } from "react"
+import { WorkflowEngine } from "v2/Utils/WorkflowEngine"
+import { Context } from "./useOnboardingContext"
+
+interface UseEngine {
+  context: React.RefObject<Context>
+  onDone(): void
+}
+
+export const useEngine = ({ context, onDone }: UseEngine) => {
```
damassi, line 10: `useWorkflowEngine`

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
+import { useRef, useState } from "react"
+import { WorkflowEngine } from "v2/Utils/WorkflowEngine"
+import { Context } from "./useOnboardingContext"
+
+interface UseEngine {
```
damassi, line 5: `UseWorkflowEngine`

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
...
+  context: React.RefObject<Context>
+  onDone(): void
+}
+
+export const useEngine = ({ context, onDone }: UseEngine) => {
+  const engine = useRef(
```
damassi, line 11: `const workflowEngine`

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (engine.current.isEnd()) {
+      onDone()
+      return
+    }
+
+    setCurrent(engine.current.next())
+  }
+
+  return {
+    current,
```
damassi, line 57: Maybe `currentStep`

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
...
+    engine: engine.current,
+    next,
```
damassi, line 59: Maybe `nextStep`

src/v2/Apps/Onboarding/config.ts
```diff
@@ -0,0 +1,92 @@
...
+      onDone()
+      return
+    }
+
+    setCurrent(engine.current.next())
+  }
+
+  return {
+    current,
+    engine: engine.current,
```
damassi, line 58: `workflowEngine`

src/v2/Apps/Onboarding/useOnboardingContext.tsx
```diff
@@ -0,0 +1,89 @@
...
+  setContext: (updatedContext: Partial<Context>) => React.RefObject<Context>
+}>({
+  answers: [],
+  context: createRef<Context>(),
+  current: "",
+  engine: new WorkflowEngine({ workflow: [] }),
+  next: () => {},
+  onDone: () => {},
+  progress: 0,
+  setContext: () => createRef<Context>(),
```
damassi, line 43: Ooh, maybe we literally are setting a new react context? 

Still, i wonder how this can be made less opaque as part of the decision engine and / or onboarding flow. Name-wise its too low-level of a thing to have to be called within individual steps, but if it was named differently it might not feel like so

damassi (approved):
This is looking amazing y'all 👍  There's a couple minor comments throughout that would be good to look at which might improve readability, but nothing totally blocking!

src/v2/Apps/Onboarding/useOnboardingContext.tsx
dzucconi: No, I agree. It's a little weird / should be renamed.

src/v2/Apps/Onboarding/OnboardingApp.tsx
dzucconi: I concur that we should just use the const form for all of them. Makes it a little less error prone.

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
dzucconi: So this is sort of no longer valid now that there will be multi choice. We'll have to just revist this once the flow is final.

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
```diff
@@ -27,9 +27,10 @@ import {
 import { useOnboardingContext } from "../useOnboardingContext"
 
 export const OnboardingQuestionTwo: FC = () => {
-  const { answers, setAnswers, setContext, next } = useOnboardingContext()
+  const { answers, setAnswerTwo, setBasis, next } = useOnboardingContext()
```
damassi, line 30: Don't mean to drag this on, but what does "basis" mean? (Trying to find a way to see this function and understand what it's doing without much outside input.)

src/v2/Apps/Onboarding/useOnboardingContext.tsx
```diff
@@ -10,38 +10,40 @@ import {
...
   current: string
-  engine: WorkflowEngine
+  workflowEngine: WorkflowEngine
   next(): void
   onDone(): void
   progress: number
-  setAnswers: Dispatch<SetStateAction<AnswersState>>
-  setContext: (updatedContext: Partial<Context>) => React.RefObject<Context>
+  setAnswerOne: Dispatch<SetStateAction<Answer>>
+  setAnswerTwo: Dispatch<SetStateAction<Answer>>
```
damassi, line 34: If there were three multiple choice options, would we add another option here?

dzucconi: This whole thing is gonna change so I'm not gonna over think this; but something like that. Or we might model it differently depending.

src/v2/Apps/Onboarding/Views/OnboardingQuestionTwo.tsx
dzucconi: This object is what's needed to make decisions to move to the conditional steps.



from https://www.youtube.com/watch?v=V6DKjEbdYos

PR #10115, opened 2022-05-18
feat: add event tracking for the Get in Touch button

kiryl-zubarau:

This PR solves [TX-335]

### Description

This PR adds event tracking for the Get in Touch button.

### DIFF
--- a/src/v2/Apps/Consign/Routes/MarketingLanding/Components/CtaBanner.tsx
+++ b/src/v2/Apps/Consign/Routes/MarketingLanding/Components/CtaBanner.tsx
@@ -40,6 +40,18 @@ export const CtaBanner = () => {
     })
   }
 
+  const trackGetInTouchClick = () => {
+    trackEvent({
+      // @ts-ignore
+      action: "clickedGetInTouch",
+      // @ts-ignore
+      context_module: "StickyBanner",
+      context_page_owner_type: contextPageOwnerType,
+      label: "Submit an Artwork",
+      user_id: user?.id,
+    })
+  }
+
   return (
     <Sticky>
       {({ stuck }) => {
@@ -59,6 +71,7 @@ export const CtaBanner = () => {
                         as={RouterLink}
                         to="mailto:sell@artsy.net?subject=Inquiry about selling with Artsy"
                         width="100%"
+                        onClick={trackGetInTouchClick}
                         variant="primaryWhite"
                         size="small"
                         mr={2}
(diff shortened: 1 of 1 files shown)


AlicanAkyuz (approved):
lgtm! 

one suggestion is to add a simple test case for this, just in case a future changes breaks this, it'll be caught.

AlicanAkyuz (approved):

src/v2/Apps/Consign/Routes/MarketingLanding/Components/CtaBanner.tsx
```diff
@@ -40,81 +58,81 @@ export const CtaBanner = () => {
...
 
+  const trackGetInTouchClick = () => {
+    trackEvent({
+      // @ts-ignore
+      action: "clickedGetInTouch",
+      // @ts-ignore
+      context_module: "StickyBanner",
+      context_page_owner_type: contextPageOwnerType,
+      label: "Get in Touch",
+      user_id: user?.id,
```
iskounen, line 69: **suggestion (non-blocking):** If the event schema has a field for `user_email` it could be convenient to include that so that we don't have to fetch the email of a user from its ID when we use this data.

src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/CtaBanner.jest.tsx
```diff
@@ -0,0 +1,100 @@
...
+      <CtaBannerContent />
+    </MockBoot>
+  )
+
+const sharedTests = () => {
+  it("tracks get in touch click", () => {
+    fireEvent.click(screen.getByText("Get in Touch"))
+
+    expect(trackEvent).toHaveBeenCalled()
+    expect(trackEvent).toHaveBeenCalledWith({
```
iskounen, line 30: suggestion (non-blocking): It could be good to test for the presence of the `user_id` field if the user is authenticated. (Would that field be `undefined` if they are not authenticated?)

iskounen (approved):

src/v2/Apps/Consign/Routes/MarketingLanding/Components/CtaBanner.tsx
kiryl-zubarau: Added, also warned Audrey about the changes.

src/v2/Apps/Consign/Routes/MarketingLanding/Components/__tests__/CtaBanner.jest.tsx
kiryl-zubarau: Added. 

> Would that field be undefined if they are not authenticated?

Yes, and it's a reason why I don't check it in the test in this case, it's empty.

iskounen (approved):

PR #10129, opened 2022-05-20
feat(FX-3972): send trending sort for artist artwork grids based on Unleash flag

dimatretyak:
The type of this PR is: **Expirement**

This PR solves [FX-3972]

### Description
When a user is viewing an artist artwork grid with the “default” sort option enabled, we want to have the option to either send `-decayed_merch` (our current default sort) or our new `trending_score` (TODO: Clarify this moment) sort parameter to Gravity. We want to be able to change this behavior based on an Unleash feature flag. This will first allow us to add specific Artsy users to the “view trending score-boosted sort” group, so that we can do a qualitative review of the updated grids with stakeholders.

### Acceptance criteria
- [x] Feature flag in Unleash is created
- [x] When the flag is true AND the user is viewing an artist works for sale page, send the new trending score sort param in the fetch for `filteredArtworksConnection`
- [x] When it is `false`, maintain the existing behavior (sending `-decayed_merch`)
- [ ] Correctly track analytics

Then later, we can launch by updating the feature flag to show to all users (or to a portion of users for an A/B test if desired)


dimatretyak:
Necessary changes will be added in another PR

PR #10062, opened 2022-05-11
chore(FX-3927): update copy when no artworks available by an artist

gkartalis:
The type of this PR is: **Enhancement**

This PR solves [FX-3927]

### Description

[Review app](https://empty-artist-artworks.artsy.net/)

We want to display the same messages on mWeb and Desktop.  More specifically:

- No filters applied: [this copy](https://www.figma.com/file/zG9cduJdOaTQAw886QcOwn/Filters?node-id=3914%3A33534) on desktop / mWeb

- Filters applied: the already existing copy on mWeb but on both desktop/mWeb:
 `There aren't any works available that meet the following criteria at this time. Change your filter criteria to view more works. Clear all filters.` 
 where the clear all filters will be a clickable underlined text that clears the applied filters.

##### Desktop

|No filters applied|Filters applied|
|---|---|
|Before||
|Screenshot 2022-05-19 at 11 58 36|Screenshot 2022-05-19 at 11 58 30|
|After||
|Screenshot 2022-05-19 at 11 58 54|Screenshot 2022-05-19 at 11 48 35|

##### mWeb

|No filters applied|Filters applied|
|---|---|
|Before||
|Screenshot 2022-05-19 at 11 54 43|Screenshot 2022-05-19 at 11 54 53|
|After||
|Screenshot 2022-05-19 at 11 48 44|Screenshot 2022-05-19 at 11 48 35|

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FX-3927]: https://artsyproduct.atlassian.net/browse/FX-3927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10130, opened 2022-05-20
chore(FX-3927): Address post-merge comments

gkartalis:

This PR solves [FX-3927]

### Description

Addressed the post merge comments from https://github.com/artsy/force/pull/10062

[FX-3927]: https://artsyproduct.atlassian.net/browse/FX-3927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
@@ -1,28 +1,26 @@
-import { Box, Message, Text } from "@artsy/palette"
+import { Box, BoxProps, Message } from "@artsy/palette"
 import { isEmpty } from "lodash"
 import { useArtworkFilterContext } from "v2/Components/ArtworkFilter/ArtworkFilterContext"
 import { ArtworkGridEmptyState } from "v2/Components/ArtworkGrid/ArtworkGridEmptyState"
 import { Sticky } from "v2/Components/Sticky"
 
-export const ZeroState: React.FC = () => {
+export const ZeroState: React.FC<BoxProps> = props => {
   const { selectedFiltersCounts, resetFilters } = useArtworkFilterContext()
   const hasAppliedFilters = !isEmpty(selectedFiltersCounts)
 
   if (hasAppliedFilters) {
-    return <ArtworkGridEmptyState onClearFilters={resetFilters} />
+    return <ArtworkGridEmptyState onClearFilters={resetFilters} {...props} />
   }
 
   return (
-    <Box width="100%">
+    <Box width="100%" {...props}>
       <Sticky>
         {({ stuck }) => {
           return (
             <Box pt={stuck ? 1 : 0} my={1}>
               <Message title="No works available by the artist at this time">
-                <Text variant="sm" textColor="black60">
-                  Create an Alert to receive notifications when new works are
-                  added
-                </Text>
+                Create an Alert to receive notifications when new works are
+                added
               </Message>
             </Box>
           )
--- a/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
@@ -17,7 +17,7 @@ export const ZeroState: React.FC<BoxProps> = props => {
       <Sticky>
         {({ stuck }) => {
           return (
-            <Box pt={stuck ? 1 : 0} my={1}>
+            <Box pt={stuck ? 1 : 0}>
               <Message title="No works available by the artist at this time">
                 Create an Alert to receive notifications when new works are
                 added
--- a/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
+++ b/src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
@@ -13,14 +13,13 @@ export const ZeroState: React.FC = () => {
   }
 
   return (
-    <Box width="100%" my={1}>
+    <Box width="100%">
       <Sticky>
         {({ stuck }) => {
           return (
-            <Box pt={stuck ? 1 : 0}>
-              <Message>
-                <Text>No works available by the artist at this time</Text>
-                <Text textColor="black60">
+            <Box pt={stuck ? 1 : 0} my={1}>
+              <Message title="No works available by the artist at this time">
+                <Text variant="sm" textColor="black60">
                   Create an Alert to receive notifications when new works are
                   added
                 </Text>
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
```diff
@@ -13,14 +13,13 @@ export const ZeroState: React.FC = () => {
...
       <Sticky>
         {({ stuck }) => {
           return (
-            <Box pt={stuck ? 1 : 0}>
-              <Message>
-                <Text>No works available by the artist at this time</Text>
-                <Text textColor="black60">
+            <Box pt={stuck ? 1 : 0} my={1}>
+              <Message title="No works available by the artist at this time">
+                <Text variant="sm" textColor="black60">
```
dzucconi, line 22: This `Text` node shouldn't be needed at all. Just simple children here.

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
```diff
@@ -8,16 +8,14 @@ interface ArtworkGridEmptyStateProps {
 }
 
 export const ArtworkGridEmptyState: React.FC<ArtworkGridEmptyStateProps> = props => (
-  <Box width="100%" my={1}>
+  <Box width="100%">
     <Sticky>
       {({ stuck }) => {
         return (
-          <Box pt={stuck ? 1 : 0}>
+          <Box pt={stuck ? 1 : 0} my={1}>
```
dzucconi, line 15: This doesn't really address the concern. Golden rule: never use external margins. Find the parent of `ArtworkGridEmptyState` and control spacing there.

src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
```diff
@@ -13,14 +13,13 @@ export const ZeroState: React.FC = () => {
...
-    <Box width="100%" my={1}>
+    <Box width="100%">
       <Sticky>
         {({ stuck }) => {
           return (
-            <Box pt={stuck ? 1 : 0}>
-              <Message>
-                <Text>No works available by the artist at this time</Text>
-                <Text textColor="black60">
+            <Box pt={stuck ? 1 : 0} my={1}>
```
dzucconi, line 20: Shouldn't have an external margin

src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
gkartalis: The reason I went with text here is due to the [figma specs ](https://www.figma.com/file/zG9cduJdOaTQAw886QcOwn/Filters?node-id=3914%3A33534) in order to have `textColor="black60"`

dzucconi: Design System > Figma specs. That Figma is just incorrectly specified.

dzucconi: https://www.figma.com/file/m6gDpKHEWDbYJyrwsVZDBr/Artsy-3.0.0-Design-System?node-id=1020%3A781

dzucconi: In general: if you see yourself doing some one-off modification, bring it up with the design team that they are deviating from the design system.

gkartalis: sounds great! Thanks a lot!

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
gkartalis: **question:** The parent in this case is [ArtistArtworkFilter](https://github.com/artsy/force/blob/07afcef957ee741a6635a71ddc86f4b870811232/src/v2/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx#L71) Do you think that something like this would be the correct approach?

```javascript

      ZeroState={() => (
        <Box my={1}>
          <ZeroState />
        </Box>
      )}
```

damassi: You could spread the props! So `<ZeroState my={1} />`

dzucconi: Ah, yeah that's a bit awkward. Though it's definitely improved. A bit better: `ZeroStateProps extends BoxProps`, then you can just do `<ZeroState my={1} />`. It does make me wonder if _all_ `ZeroStates` should handle margins the same way (I see drill down into where it actually gets rendered and apply padding there.

Just to communicate the problem with external margins: it makes component spacing difficult to debug because you don't know what kind of spacing anything you're looking at in code might have. It also impedes reuse since different contexts may require different spacing. So it's just a good general rule to have. Extending `BoxProps` and controlling the margin from the parent is a fine compromise.

src/v2/Apps/Artist/Routes/WorksForSale/Components/ZeroState.tsx
dzucconi: Also interesting: https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/%F0%9F%9A%AB-Artsy-3.1-Design-System?node-id=6726%3A1796

It looks like in 3.1 this does get updated to look like what you're aiming for. But just using the normal `Message` component as it's intended will ensure that when this gets updated it the changes will be correct everywhere.

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
```diff
@@ -1,29 +1,30 @@
...
+export const ArtworkGridEmptyState: React.FC<ArtworkGridEmptyStateProps> = ({
+  onClearFilters,
+  ...rest
+}) => (
+  <Box width="100%" {...rest}>
     <Sticky>
       {({ stuck }) => {
         return (
-          <Box pt={stuck ? 1 : 0}>
+          <Box pt={stuck ? 1 : 0} my={1}>
```
dzucconi, line 18: Can remove this `my` now

dzucconi (dismissed):
Thanks for bearing with! 🙏 

One little prop removal and should be good to go.

src/v2/Components/ArtworkGrid/ArtworkGridEmptyState.tsx
gkartalis: 🙈

dimatretyak (approved):
@gkartalis thanks for the changes

gkartalis:
@dzucconi My pleasure! Thanks for the comments, really appreciate them, and also thanks for the useful links you've attached!

🤘

dimatretyak:
@dzucconi should we add vertical margin by default [here](https://github.com/artsy/force/blob/978082884d1074a9157fe162d9410705252b7ed8/src/v2/Components/ArtworkGrid/ArtworkGrid.tsx#L245)?

dzucconi:
@dimatretyak hard to say — ideally yes. But the structure of the filters is really something to behold. Use your discretion I suppose. I think this is fine for now.

PR #10134, opened 2022-05-20
chore: Test cleanup

jacobherrington:

### Description

Basically, I noticed that beforeEach is used frequently when beforeAll would work. Then I fixed a couple of other small things. I'm not sure what impact this really has on performance, but it seems worth merging if the impact isn't negligible.


damassi (approved):
Awesome, thank you for this cleanup 👍  I'm def guilty of using `beforeEach` when it could be `beforeAll`. Also thanks for the browserlist update. Feel free to merge on 🍏

PR #10139, opened 2022-05-23
(feat: TX-332): AlicanAkyuz/Wire transfer payment method

AlicanAkyuz:

This PR solves [TX-332]

### Description

This PR adds wire transfer as an option to the payment step of the checkout flow. 

**Implementation Details**
- moved flag-checks to `PaymentContent` component to be able to make it easier to control which method to render. 
- added `getAvailablePaymentMethods` function that adds ACH and WIRE methods to Credit card when they are available. In case both flags are false, current Credit card input is rendered without Collapse 
- placed `onWireTransferContinue` in `Payment/index.tsx` to be consistent with Credit card and debit calls. This function will update backend with `payment_method: 'wire_transfer'` when ready.  

#### Videos
Both flags as disabled:
https://user-images.githubusercontent.com/42584148/169832879-f9e54ca7-9f8c-40d2-b4a1-d30008209688.mov

ACH enabled, wire transfer disabled:
https://user-images.githubusercontent.com/42584148/169833381-3479cb48-4df9-49fb-89cf-80b99aa2da8b.mov

ACH and wire transfer enabled: 
https://user-images.githubusercontent.com/42584148/169833047-65d6a0e0-a354-4296-a552-b6913355b5f6.mov

[TX-332]: https://artsyproduct.atlassian.net/browse/TX-332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi:
This PR looks great @AlicanAkyuz. One suggestion would be, once the feature flag stuff is removed lets just be sure to look carefully at the tests and cover any behavior that might have been missed with more tests (since we're dealing with money and every aspect of $ code should be covered as best as one can. If this is already the case, then feel free to ignore this message!)

AlicanAkyuz:
@damassi thank you for flagging the tests, I definitely agree that we need to improve the tests here. The test for the flag was already there, I just changed the explanation in order to be able to differentiate ACH and wire transfer flag tests. but you are right, they actually don't test anything. I'll pay attention to these tests before the features are on prod 👍

PR #10142, opened 2022-05-23
chore: Convert Rail tests to RTL

jacobherrington:

### Description

These tests were pretty easy to swap to React Testing Library and I think they fit that mental model pretty well.


damassi (dismissed):
Nice! Once 🟢  , hit the 🍏

PR #10150, opened 2022-05-24
refactor: add alert to verify email after update

mc-jones:

This PR solves [PLATFORM-4179]

### Description

This PR adds an alert message after a user updates their email instructing them to verify their email address. It supports the backend changes being made as part of this [ticket](https://artsyproduct.atlassian.net/browse/PLATFORM-4178), and will not be merged until that work is complete. 

#### Considerations

I don't love the two toasts, but I felt it made the action require more clear to the user. If people feel strongly, I'm happy to revert to the alternative approach shown below, with the action required message being part of the success toast message. 

Current Approach:
Screen Shot 2022-05-24 at 4 14 15 PM

Alternative Approach:
Screen Shot 2022-05-24 at 4 14 51 PM

### Follow-up
Update [related integrity test.](https://github.com/artsy/integrity/blob/5dac98724c8500498156d8f491fceea16a5fbd53/cypress/integration/velocity/authentication/userAccount.spec.ts#L36)

[PLATFORM-4179]: https://artsyproduct.atlassian.net/browse/PLATFORM-4179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
+++ b/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
@@ -78,7 +78,7 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
               sendToast({
                 variant: "alert",
                 message:
-                  "Please confirm your new email for this update to take effect.",
+                  "Please confirm your new email for this update to take effect",
               })
             }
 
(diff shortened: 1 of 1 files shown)


artsyjian (approved):
nice!

damassi (dismissed):
I think the double toast definitely gets the point across better and is actually the right use: 
- 1st toast: success
- 2nd toast: notice that a follow up action is needed. 

Maybe @dzucconi has some thoughts? But no blockers from me 👍

dzucconi:
Looks good to me. This is their intended use. If you wanted you could have the alert stay up a bit longer. It accepts a param `ttl` which is specified in milliseconds. The default is `6000`. You can also set it to `Infinity` if you want it to stick around until dismissed.

kajatiger:
I learned that this thing is called "toast" which is kind of funny to me. Thank you for that!
Also I think the approach with them being two individual messages is better and more clear.

joeyAghion (approved):

PR #10161, opened 2022-05-26
chore: scroll to Booths section

dimatretyak:

This PR solves [GRO-]

### DIFF
--- a/src/v2/Apps/Fair/Components/FairBoothRail/FairBoothRail.tsx
+++ b/src/v2/Apps/Fair/Components/FairBoothRail/FairBoothRail.tsx
@@ -18,7 +18,12 @@ import {
   initialBoothFilterState,
   useBoothsFilterContext,
 } from "../BoothFilterContext"
-import { buildUrl } from "v2/Components/ArtworkFilter/Utils/urlBuilder"
+import {
+  paramsToSnakeCase,
+  removeDefaultValues,
+} from "v2/Components/ArtworkFilter/Utils/urlBuilder"
+import qs from "qs"
+import { useRouter } from "v2/System/Router/useRouter"
 
 interface FairBoothRailProps extends BoxProps {
   show: FairBoothRail_show
@@ -36,6 +41,7 @@ export const FairBoothRail: React.FC<FairBoothRailProps> = ({
     contextPageOwnerType,
   } = useAnalyticsContext()
   const { filters } = useBoothsFilterContext()
+  const { match } = useRouter()
   let link: string | null = null
 
   const tappedViewTrackingData: ClickedArtworkGroup = {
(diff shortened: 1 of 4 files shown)


src/v2/Apps/Fair/Components/FairBoothRail/FairBoothRail.tsx
```diff
@@ -43,14 +56,29 @@ export const FairBoothRail: React.FC<FairBoothRailProps> = ({
...
+    const preparedParams = paramsToSnakeCase(params)
+    const queryString = qs.stringify({
+      ...preparedParams,
+      focused_booths: true,
+    })
+    const backHref = `${match.location.pathname}?${queryString}`
+    const encodedBackHref = encodeURIComponent(backHref)
+
+    link = `${show.href}?back_to_fair_href=${encodedBackHref}`
+  }
```
dimatretyak, line 72: Initially, I thought to scroll the page to a specific item from which the user navigated to the Show page. But during testing, undesirable behavior was found that worsened the user's experience with the page.

The problem is that initially, when loading the page, instead of real images for the elements in the Booths section, we display loading placeholders of fixed size. When the actual image is rendered instead of loading placeholder, the final size of the element could change. For this reason, the scroll offset to a specific item in the Booths section, which was made when navigating to the Fair page from "Back to <Fair>" banner, could not correspond to the offset that will be after all the images for the elements are loaded.

src/v2/Apps/Fair/Components/__tests__/FairBoothRail.jest.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+  `,
+})
+
+describe("FairBoothRail", () => {
+  const mockTracking = useTracking as jest.Mock
+  const mockUseRouter = useRouter as jest.Mock
+  const trackEvent = jest.fn()
+
+  beforeEach(() => {
```
jacobherrington, line 50: **suggestion (non-blocking):** Could this be a `beforeAll` call? The tests still pass locally when I make that change. Just wondering!

```suggestion
  beforeAll(() => {
```

dimatretyak: yep, sure

dimatretyak: yep, we can replace `beforeEach` with `beforeAll`

src/v2/Apps/Fair/Routes/FairOverview.tsx
```diff
@@ -14,13 +14,30 @@ import { FairAboutFragmentContainer as FairAbout } from "../Components/FairOverv
...
 }
 
-const FairOverview: React.FC<FairOverviewProps> = ({ fair }) => {
+const FairOverview: FC<FairOverviewProps> = ({ fair }) => {
   const { user } = useSystemContext()
+  const { match } = useRouter()
+  const { focused_booths: focusedBooths } = match.location.query
+  const { scrollTo: scrollToBooths } = useScrollToElement({
+    behavior: "smooth",
+    selectorOrRef: "#jump--BoothsSection",
```
dblandin, line 30: **praise**: Seems like a good compromise to jump to the booths section 👍🏼

dblandin (approved):
LGTM! 💯

PR #10170, opened 2022-05-30
chore: remove `ENABLE_FAIR_PAGE_EXHIBITORS_TAB` feature flag

dimatretyak:

This PR solves [GRO-]

### Description
This PR removes the use of the `ENABLE_FAIR_PAGE_EXHIBITORS_TAB' feature flag, since it was released in September 2021


src/v2/Apps/Fair/__tests__/FairApp.jest.tsx
```diff
@@ -126,32 +120,21 @@ describe("FairApp", () => {
       }),
     })
 
-    fireEvent.click(screen.getByText("Booths"))
+    fireEvent.click(screen.getByText("Exhibitors A-Z"))
```
dblandin, line 123: **question**: Should we remove this click if we're testing click tracking to the artworks tab in this assertion? Having the context_module remain `fairInfo` seems fine to me - unless I'm missing something.

dblandin (dismissed):
One minor question but LGTM overall!

src/v2/Apps/Fair/__tests__/FairApp.jest.tsx
dimatretyak: Done

dblandin (approved):

PR #10180, opened 2022-05-31
chore(article): promotes article2 to production [review app]

dzucconi:
Opening to spin up a review app. Not yet ready for merge.


dzucconi:
For the time being I haven't removed Reaction yet. We're still depending on it in https://www.artsy.net/gender-equality — which we don't have a good redirect replacement for. (The remaining apps all use this 'curation' feature of Writer) (sigh)

cypress/integration/article.spec.ts
```diff
@@ -10,9 +10,10 @@ describe("Article", () => {
         "contain",
         "Agnes Denes’s Manhattan Wheatfield Has Only Grown More Poignant"
       )
+      console.log(cy.title())
```
damassi, line 13: strayy

PR #10181, opened 2022-05-31
chore: remove personnel, margin fix

laurabeth:

This PR solves [GRO-986] [GRO-1016] [GRO-1017]

### Description

A few very small tickets in one PR.

[GRO-986]: https://artsyproduct.atlassian.net/browse/GRO-986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1016]: https://artsyproduct.atlassian.net/browse/GRO-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1017]: https://artsyproduct.atlassian.net/browse/GRO-1017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
@@ -6,6 +6,7 @@ import { ArtworkSidebarAuctionInfoPolling_artwork } from "v2/__generated__/Artwo
 import { ArtworkSidebarAuctionInfoPolling_me } from "v2/__generated__/ArtworkSidebarAuctionInfoPolling_me.graphql"
 import { usePoll } from "v2/Utils/Hooks/usePoll"
 import { useEffect, useRef, useState } from "react"
+import { Spacer } from "@artsy/palette"
 
 type Props = {
   artwork: ArtworkSidebarAuctionInfoPolling_artwork
@@ -54,6 +55,7 @@ export const ArtworkSidebarAuctionPolling: React.FC<Props> = ({
         currentBidChanged={currentBidChanged}
         artwork={artwork}
       />
+      <Spacer my={2} />
       <ArtworkSidebarBidActionFragmentContainer artwork={artwork} me={me} />
     </>
   )
--- a/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
@@ -55,7 +55,7 @@ export const ArtworkSidebarAuctionPolling: React.FC<Props> = ({
         currentBidChanged={currentBidChanged}
         artwork={artwork}
       />
-      <Spacer my={2} />
+      <Spacer mt={2} />
       <ArtworkSidebarBidActionFragmentContainer artwork={artwork} me={me} />
     </>
   )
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
```diff
@@ -54,6 +55,7 @@ export const ArtworkSidebarAuctionPolling: React.FC<Props> = ({
         currentBidChanged={currentBidChanged}
         artwork={artwork}
       />
+      <Spacer my={2} />
```
dzucconi, line 58: We should be using `mt` or `mb` so that we only get valid spacing scale values. (None of this is obvious; long-standing issue)

dzucconi: Also I'd be weary about this: we need to find this exact state. I wouldn't be sure this doesn't compound existing margins which _are already_ correct.

damassi: @dzucconi I don't quite understand 

> We should be using mt or mb so that we only get valid spacing scale values. (None of this is obvious; long-standing issue)

Can you elaborate? I never use `mt` or `mb` and almost always use `my`. 

Also, the two surrounding fragments _should_ always be there.

damassi (commented):
This looks good to me! (Pending @dzucconi's approval)

src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarAuctionInfoPolling.tsx
dzucconi: So like: in the vertical case the margins collapse; which leads to a 'correct' value. But if you inspect you'll see:



If you were to use `mx` then you'd just get completely invalid spacing values.

We really need to update the API for this component to prevent invalid values. But yeah generally I just use `mt` and `ml` to avoid this problem.

laurabeth: guess `my` on a spacer does top and bottom, which actually doubles the expected padding amount? also, I did the footwork on the state. this is the spot, and this component doesn't return in any other circumstance.

dzucconi: Re: the second comment: There's obviously states in which this component looks correct (most of them) and a single state in which it doesn't. Something is wrong. This is perhaps a great example of why you should never ever use external margins in a component.

dzucconi: @laurabeth ah OK. Gotcha; thanks!

dzucconi: Ah yeah I thought this was the outer component 😅

dzucconi (approved):

PR #10198, opened 2022-06-02
feat(webpack): Upgrade to Webpack 5

damassi:

### Description

This upgrades webpack to 5. We're now caching compile output to the filesystem `.cache` folder so after the first cold boot, we should see subsequent initial compiles around ~5-8sec. 

Additionally, added a `COMPILE_LEGACY_CLIENT_IN_DEV` env var, set by default to false. This will skip mounting our legacy webpack config to minimize the amount of files in ram. (It will be uncommon that anyone needs to compile old code.)

There's some follow-up work after this PR: for some reason, saves are triggering multiple HMR rebuilds (same as in webpack 4) and i think this might be contributing to some HMR slowness, but that's an improvement independent of this work. I'm also looking into a long-standing issue of why it takes so long for artsy.net to appear after `yarn start` has completed -- that is, after webpack has already finished compilation. It'd be nice to be able to go to localhost and see the page boot instantly. 

Folks tagged in this PR: Can y'all keep a close eye on overall compilation performance (specifically hot reloading) and let me know what you see? I'm not seeing _that_ much improvement here, but i think we can tune this a bit more.

Let's see how things build on staging. 

cc @artsy/grow-devs


webpack/envs/sharedConfig.js
```diff
@@ -9,12 +9,23 @@ export const productionDevtool = env.webpackDevtool || "source-map"
 export const mode = env.webpackDebug ? "development" : env.nodeEnv
 export const stats = env.webpackStats || "errors-only"
 
+export const experiments = {
+  lazyCompilation: {
+    entries: false,
+    imports: false,
+  },
```
damassi, line 16: Note this flag is currently off. There's some rough edges around lazy compilation and loadable components / bundle-splitting, but will continue looking into a fix here.

damassi:
(Gonna merge this and continue)

PR #10203, opened 2022-06-03
resolves setImmediate is not defined problem

nickskalkin:

This PR solves multiple 500s on staging (presumably after webpack update)

Screenshot 2022-06-03 at 12 10 28

### DIFF
--- a/package.json
+++ b/package.json
@@ -231,6 +231,7 @@
     "styled-system": "5.1.5",
     "stylus": "0.54.5",
     "superagent": "3.8.3",
+    "timers-browserify": "2.0.12",
     "tti-polyfill": "0.2.2",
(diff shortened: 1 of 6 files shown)


dzucconi:
Pretty sure we can just have Webpack polyfill `setImmediate`. No one is going to know to import this down the line. cc @damassi

damassi:
Yup @dzucconi, I had already set that up in the Webpack PR: https://github.com/artsy/force/blob/main/webpack/envs/sharedPlugins.js#L31. 

That's why i'm a bit suspicious about all of this!

dzucconi:
I dont think that plugin includes setImmediate

damassi:
It does inside of `timers`

PR #10207, opened 2022-06-03
fix(standards): replace setImmediate with setTimeout

damassi:

### Description

This replaces our non-standard use of `setImmediate` with `setTimeout`. It turns out that `setImmediate` is a [non-standard browser function](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) and `setTimeout, 0` does the exact same thing. 

Related: 
- https://github.com/artsy/force/pull/10203
- https://github.com/artsy/force/pull/10198


dzucconi (approved):
🙏

dimatretyak (approved):
🔥🔥🔥

jacobherrington (approved):
LGTM! Looks like detect-secrets failed on us 😢

PR #10219, opened 2022-06-07
fix: respects disable scripts flag

dzucconi:
Fixes the error pages attempting to load the layout.

Not entirely sure what this diff does but, this is the source of the issue:
https://github.com/artsy/force/commit/c14a320fa33e9eae665283963f7cea05ae45b053#diff-8527786fa5a1865c6babd7d10de914560a326cbe092000ed5e5e447eb8a4cf03R26

### DIFF
--- a/src/v2/index.ejs
+++ b/src/v2/index.ejs
@@ -22,8 +22,10 @@
   <link rel="preload" href="<%%= fontUrl %>/ll-unica77_medium.woff2" as="font" type="font/woff2" crossorigin />
   <link rel="preload" href="<%%= fontUrl %>/ll-unica77_italic.woff2" as="font" type="font/woff2" crossorigin />
 
-  <!-- Create preload tags for dynamic chunks -->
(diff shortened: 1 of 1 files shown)


damassi (commented):
Interesting - I wonder why this wasn't covered to begin with? 

Also, thoughts on us fixing the layout so that the full page shell is there? I was surprised to see the JS executing properly; that's why I thought it was a layout issue that had changed somehow

dzucconi:
Feel free to attempt but to reiterate the original problem: when you mount the application in an error state, it will execute the conditions to replicate the error state, which isn't desirable. What you want to do is to mount a totally parallel, simplified copy of the application so that you're not sitting inside of the error boundary. The only reason why the application 'works' in this case is because you're seeing a 404 error, so it's just a routing thing. But there [are many other error states to consider.](https://github.com/artsy/force/pull/9153).

dzucconi:
https://github.com/artsy/force/pull/9153#issuecomment-1018720068

> If I try to do that the whole application loads, including the routes, and you ultimately get where you want to go but you see the UI fly around as it mounts the application, executes the route and ultimately replicates the error. What I really want to do is mount a new, simplified application that's just the ErrorPage + AppShell + routes. I have no idea what it'd take to make that happen.

PR #10224, opened 2022-06-07
feat: render review step ui depending on payment method of order

AlicanAkyuz:

This PR solves [TX-380]

### Description

Now that paymentMethod property is available on CommerceOrder, this PR updates the review step of the checkout flow to render the appropriate payment summary depending on order's payment method.


src/v2/Apps/Order/Components/PaymentMethodSummaryItem.tsx
```diff
@@ -24,14 +22,17 @@ export const PaymentMethodSummaryItem = ({
     ...{ textColor: textColor },
   }
   const isACHEnabled = useFeatureFlag("stripe_ACH")
+  const isWireTransferEnabled = useFeatureFlag("wire_transfer")
```
kiryl-zubarau, line 25: I suppose these flags can be removed since we check it on the backend side.

AlicanAkyuz: agreed. 

they are already removed from `PaymentContent` where we render the radio buttons vs. only Credit card. I left them here for extra safety (not to break anything for the current payment method credit card). 

do you know if we are already getting `paymentMethod === 'CREDIT_CARD'` from the API? If so, I'll just change this to a switch on `paymentMethod`.

kiryl-zubarau (approved):

PR #10228, opened 2022-06-07
bug(new for you): paginate nfy grid [DO NOT MERGE]

laurabeth:
The type of this PR is: **Bug**

This PR solves [GRO-1046] [GRO-1035]

### Description

This PR replaces the "show more" functionality with a fully paginated grid so that artworks load in the correct order.

This change set surfaced an issue with the artworksForUser query that will need to be resolved before this PR can be merged:

- When passing a cursor to `artworksForUser` that cursor is not respected
- If the page size is instead increased the query will return a larger result set of unique works, indicating that a new page of works should have also existed
- This was proven to be an issue with the query and not the UI as a result of the following steps:

 <space><space> 1. Run the following query against MP:
```graphql
query {
	artworksForUser(first:2, includeBackfill: true) {
		edges {
			cursor
			node {
				title
			}
		}
		pageInfo {
			hasNextPage
			endCursor
		}
	}
}
```
<space><space> 2. Take note of the title of the 2 returned artworks
<space><space> 3. Copy the cursor of the first object and add it to the list of params for `artworksForUser` with the key `after`, as follows:

```graphql
query {
	artworksForUser(first:2, after: "object-cursor-string", includeBackfill: true) {
		edges {
			cursor
			node {
				title
			}
		}
		pageInfo {
			hasNextPage
			endCursor
		}
	}
}
```
<space><space> 4. Run the query:
<space><space><space><space> - Expected result: The query returns 1 old and one new artwork
<space><space><space><space> - Actual result: the returned set of objects has not changed
<space><space> 5. Re-run these steps against the `artworkConnection` query and compare results

[GRO-1046]: https://artsyproduct.atlassian.net/browse/GRO-1046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1035]: https://artsyproduct.atlassian.net/browse/GRO-1035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi:
@laurabeth - Can we either a) convert this PR to a draft; of b) add `[HOLD]` to the PR title or body just in case someone misses the tiny GH label? It seems obvious but those labels are kinds squinty.

Also, can we elaborate a bit in the PR description on what we found when investigating pagination and the problems with connections? That way that knowledge isn’t lost for when Jon returns and there’s a papertrail.

src/v2/Apps/NewForYou/NewForYouApp.tsx
```diff
@@ -1,24 +1,37 @@
...
       <MetaTags title="New For You" />
       <Text variant="xl" mt={4}>
         New Works For You
       </Text>
-      <Spacer mt={4} />
-      {viewer && <NewForYouArtworksGridFragmentContainer viewer={viewer} />}
+      <Spacer mt={2} />
+      {!isLoggedIn && (
+        <Message variant="warning">
+          <Text>
```
dzucconi, line 27: This `Text` node is not needed.

src/v2/Apps/NewForYou/NewForYouApp.tsx
```diff
@@ -1,24 +1,37 @@
...
+            Already have an account? <Link to="/login">Log in</Link> to see your
```
dzucconi, line 28: Should use `RouterLink` (tbh no clue what `Link` is though)

src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
```diff
@@ -1,117 +1,121 @@
...
-            )
-          })}
-        </Masonry>
+      {viewer.artworksForUser ? (
+        <ArtworksGrid
+          artworks={viewer.artworksForUser}
+          columnCount={[2, 3]}
+          itemMargin={40}
+          loading={loading}
+        />
```
dzucconi, line 61: interestingly `Masonry` can be used fine with this pagination style. It doesn't matter either way, really; though `ArtworksGrid` is much more complex.

damassi: It is more complex, but if we've got an artwork grid it should go through the `<ArtworkGrid>` component so we can properly refactor later.

src/v2/Apps/NewForYou/NewForYouApp.tsx
```diff
@@ -1,24 +1,37 @@
...
+            personalized recommendations.
+          </Text>
+        </Message>
+      )}
+      <Spacer mt={2} />
+      {viewer && <NewForYouArtworksGridRefetchContainer viewer={viewer} />}
```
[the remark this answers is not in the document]
laurabeth: The initial spacer was mt={4}. Did half and half spacing and inserted the message in between. Personally, I think this looks better than 2x mt={4} since the Message area has such a significant amount of padding.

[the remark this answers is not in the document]
dzucconi: That's fine but it counts as external margins if the viewer is null. We don't want this to have any external margin. I just mean to nest it inside of the conditional.

src/v2/Apps/NewForYou/NewForYouApp.tsx
laurabeth: Removed

src/v2/Apps/NewForYou/NewForYouApp.tsx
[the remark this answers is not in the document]
laurabeth: I knew what you meant about the conditional. I think we're crossing wires. Will reach out via slack once everything else here is wrapped up.

jonallured:
Closing in favor of https://github.com/artsy/force/pull/10305.

PR #10236, opened 2022-06-08
fix: removes offset; removes double focus

dzucconi:
This is actually centered. I think you were trying to offset for the line-height of the headline text. Which I don't think we should do.

Also minor fix: don't wrap buttons in links. It creates a double tabstop and an empty button.


jacobherrington (approved):
⌨️

src/v2/Apps/ArtAppraisals/ArtAppraisalsApp.tsx
```diff
@@ -63,9 +63,13 @@ const Header: React.FC = () => {
 
             <Spacer mt={4} />
 
-            <a href="mailto:simon.wills@artsymail.com">
-              <Button>Request an Appraisal</Button>
-            </a>
+            <Button
+              // @ts-ignore
+              as="a"
```
damassi, line 68: Ugh, duh!

PR #10245, opened 2022-06-09
chore: truncate text for last pill

dimatretyak:

### Description
We have to truncate the text for the second pill when hovering. Otherwise, there may be a situation when the contents of these pills exceed the width of the parent container and there is such a visual bug

Image 2022-06-09 13-31-35

The result of how it will look after the text for the second pill is truncated
image

### DIFF
--- a/src/v2/Components/Artwork/HoverDetails.tsx
+++ b/src/v2/Components/Artwork/HoverDetails.tsx
@@ -27,11 +27,22 @@ const HoverDetails: FC<HoverDetailsProps> = ({ artwork }) => {
   return (
     <HoverContainer>
       <Join separator={<Spacer mr={0.5} />}>
-        {pills.map(pill => (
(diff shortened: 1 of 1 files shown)


dzucconi:
Hm: https://palette-storybook.artsy.net/?path=/story/components-pill--long-example-with-truncation

dzucconi:
Does just minWidth={0} work?

dimatretyak:
> Does just minWidth={0} work?

yep, it works

dzucconi:
Cool, should probably just make that default in Palette I guess.

dimatretyak:
I want to clarify a small point with you 😅. What do you think is the best option:
1. Truncate text for all pills when hovering
2. Truncate text only for last pill when hovering

#### Option 1
image

#### Option 2
image

dzucconi:
Oof yeah, sounds like a question for design but it makes more sense that it's just the last one I suppose.

PR #10246, opened 2022-06-09
chore(FX-4007): replace hover to click on rightmost navigation bar elements

dimatretyak:

This PR solves [FX-4007]

Related Palette PR: artsy/palette#1180

Review app: https://replace-hover-on-click.artsy.net/

### Description
Right now you can hover over the activity bell or the profile icon to see menus appear. We’d like to remove the hover effect and instead show/hide the menus on click

### Acceptance Criteria
* Menu should be opened/hided by click instead of hover
* Clicking outside of the menu should cause the menu to hide

[FX-4007]: https://artsyproduct.atlassian.net/browse/FX-4007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/NavBar/NavBarLoggedInActions.tsx
+++ b/src/v2/Components/NavBar/NavBarLoggedInActions.tsx
@@ -1,7 +1,7 @@
 import { useContext } from "react"
 import * as React from "react"
 import { NavBarNotificationsQueryRenderer, NavBarUserMenu } from "./Menus"
-import { AnalyticsSchema, SystemContext, useTracking } from "v2/System"
+import { SystemContext } from "v2/System"
 import {
   BellIcon,
   Dropdown,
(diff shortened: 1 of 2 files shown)


nickskalkin (approved):
👍

src/v2/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -52,25 +46,14 @@ export const NavBarLoggedInActions: React.FC<Partial<
...
-              onClick={() => {
-                trackEvent({
-                  action_type: AnalyticsSchema.ActionType.Click,
-                  subject: AnalyticsSchema.Subject.NotificationBell,
-                  new_notification_count: getNotificationCount(),
-                  destination_path: "/works-for-you",
-                })
-              }}
-            />
-
```
dimatretyak, line 73: FYI I asked Louis if it would be ok if we no longer track this event and am waiting for a response.

PR #10259, opened 2022-06-09
chore(art-appraisals): Update images

damassi:

### Description

This updates the /art-appraisals route with more suitable images (darker header, higher res main). 🙏 for the image export @dzucconi.  

cc @artsy/grow-devs

### DIFF
--- a/src/v2/Apps/ArtAppraisals/ArtAppraisalsApp.tsx
+++ b/src/v2/Apps/ArtAppraisals/ArtAppraisalsApp.tsx
@@ -36,7 +36,7 @@ export const ArtAppraisalsApp: React.FC = () => {
 
 const Header: React.FC = () => {
   return (
-    <FullBleedHeader src="https://files.artsy.net/images/appraisals-header-image.png">
+    <FullBleedHeader src="https://files.artsy.net/images/appraisals-header.jpg">
       <Flex
         position="absolute"
         top={0}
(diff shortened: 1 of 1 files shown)


src/v2/Apps/ArtAppraisals/ArtAppraisalsApp.tsx
```diff
@@ -91,7 +91,7 @@ const Header: React.FC = () => {
 
 const WhyUseArtsy: React.FC = () => {
   const { src, srcSet } = resized(
-    "https://files.artsy.net/images/appraisals-main-image.png",
+    "https://files.artsy.net/images/shinpei-kusanagi-stop-talking-to-yourself-you-will-be-alright-number-3.jpeg",
     {
       width: 2070,
       height: 2160,
```
dzucconi, line 97: Where'd you get these dimensions from? Should be 850x887 from the looks of it.

damassi: I just multiplied by 2 or 3 because you were saying that ideally the images should be real big. Lol, clearly i still do not understand how to handle images on the web 😦

damassi: will update.

dzucconi: :)

`srcSet` handles the 2x-ing. You want to just inspect it at the max breakpoint and use those dimensions.

dzucconi: The thing we want to be really big is the source file which we use to pull resizes/crops from.

PR #10262, opened 2022-06-09
feat(cookie-consent): Sync OneTrust cookie on server

damassi:

### Description

This updates our OneTrust cookie consent flow by syncing the client-side cookie on the server via `secure: true`, which should satisfy Apple's requirements around ITP and get around resetting after 7 days. 

Note the checkmarks in the rows highlighted below; that means `secure: true`, set from the server and overwriting the client cookie set by OneTrust.

Screen Shot 2022-06-09 at 1 23 59 PM

Screen Shot 2022-06-09 at 1 24 04 PM

Opening for discussion.

### DIFF
--- a/src/v2/server.ts
+++ b/src/v2/server.ts
@@ -43,6 +43,27 @@ app.get(
   }
 )
 
+// Experiment with overwriting OneTrust consent cookie with server-side version
+// to get around Safari's 7 day limit for client-side cookies.
+app.get("/set-tracking-preferences", (req, res) => {
+  const { OptanonAlertBoxClosed, OptanonConsent } = req.query
+
+  const cookieConfig = {
+    maxAge: 1000 * 60 * 60 * 24 * 365,
+    httpOnly: false,
+    secure: true,
+  }
+
+  if (OptanonAlertBoxClosed !== "undefined") {
+    res.cookie("OptanonAlertBoxClosed", OptanonAlertBoxClosed, cookieConfig)
+  }
+  if (OptanonConsent) {
+    res.cookie("OptanonConsent", OptanonConsent, cookieConfig)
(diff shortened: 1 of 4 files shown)


src/v2/server.ts
```diff
@@ -43,6 +43,27 @@ app.get(
   }
 )
 
+// Experiment with overwriting OneTrust consent cookie with server-side version
+// to get around Safari's 7 day limit for client-side cookies.
+app.get("/set-tracking-preferences", (req, res) => {
+  const { OptanonAlertBoxClosed, OptanonConsent } = req.query
+
+  const cookieConfig = {
+    maxAge: 1000 * 60 * 60 * 24 * 365,
```
damassi, line 52: Github copilot suggested this /shrug. 1000 years?

kathytafel: I kindof hope that there is some machine running somewhere in 1000 years that overflows this buffer.

artsyjian:
@joeyAghion and I looked at this locally. It all looks good except on page load there are redundant calls to `/set-tracking-preferences`.

onetrust

It turns out that on page load, OneTrust JS calls our [re-consent hook](https://github.com/artsy/force/blob/effc70dbb090cdc510cb9eff167be661d6f260c6/src/lib/analytics/segmentOneTrustIntegration/segmentOneTrustIntegration.ts#L39) despite there's no change to consent yet.

Screen Shot 2022-06-13 at 1 02 16 PM

artsyjian:
Regardless, this is LGTM to me.

joeyAghion:
I also wasn't expecting it to sync tracking preferences on every page load, but only when there are preference _changes_. Is that my misunderstanding?

PR #10263, opened 2022-06-09
feat: unauthenticated users can start Identity Verification

ovasdi:

This PR solves [PLATFORM-4165](https://artsyproduct.atlassian.net/browse/PLATFORM-4165)

#### Start identity verification process without an Artsy account.

This PR decouples identity verification from the _User_ and allows access to the page without a login. It depends on the following [MP PR](https://github.com/artsy/metaphysics/pull/4133) for the flow to work and _should be merged after MP changes are deployed_.

* decouple IdentityVerificationApp from `Me`
* open access to /identity-verification route(s)
* update tests and remove no longer used component
(more in comments 👇)


src/v2/Apps/IdentityVerification/__tests__/IdentityVerificationApp.jest.tsx
```diff
@@ -109,29 +104,6 @@ describe("IdentityVerification route", () => {
       expect(page.text()).toContain("Artsy identity verification")
     })
 
-    it("shows a message if the user does not own the identity verification", async () => {
```
ovasdi, line 112: removing this test as it's no longer relevant.

src/v2/Apps/IdentityVerification/identityVerificationRoutes.tsx
```diff
@@ -51,8 +51,8 @@ export const identityVerificationRoutes: AppRouteConfig[] = [
       query identityVerificationRoutes_IdentityVerificationAppQuery(
         $id: String!
       ) @raw_response_type {
-        me {
-          ...IdentityVerificationApp_me @arguments(id: $id)
+        identityVerification(id: $id) {
```
ovasdi, line 54: Just want to ensure this is the correct way to rewrite this route logic. Compiler seems to be happy with it.

damassi: I don't think the `@raw_response_type` is necessary anymore! Worth trying to remove and see if Relay complains.

ovasdi:
Assigning myself to coordinate rollout with MP changes.

src/v2/Apps/IdentityVerification/identityVerificationRoutes.tsx
ovasdi: @damassi relay didn't complain but I think I do have to re-write [this fixture](https://github.com/artsy/force/blob/ovasdi/PLATFORM-4165/open-access-identity-verification/src/v2/Apps/IdentityVerification/__fixtures__/routes_IdentityVerificationAppQuery.ts#L1) as the `identityVerificationRoutes_IdentityVerificationAppQueryRawResponse` type is no longer available. It is used [here](https://github.com/artsy/force/blob/ovasdi/PLATFORM-4165/open-access-identity-verification/src/v2/Apps/IdentityVerification/__tests__/routes.jest.tsx#L22).

Should I then leverage the other generated type that contains `FragmentRefs` perhaps?
```
export type identityVerificationRoutes_IdentityVerificationAppQueryResponse = {
    readonly identityVerification: {
        readonly " $fragmentRefs": FragmentRefs<"IdentityVerificationApp_identityVerification">;
    } | null;
};
```

I also noticed that a lot of fixtures are not located in the App directories but rather in `src/v2/Apps/__tests__/Fixtures/`, should the fixture be moved there?

damassi: II wouldn't worry about fixture location because we've moved away from the fixture pattern a while back and rely on relay's devtools to generate these things dynamically for us. Not suggesting to refactor the tests to use that, but worth knowing about! 

And no prob about the `raw_response_type`; I think that is related to our old testing patterns.

joeyAghion (approved):

src/v2/Apps/IdentityVerification/identityVerificationRoutes.tsx
ovasdi: Ok. Thanks for bringing it up.

mc-jones:
🥳 Nice job!

PR #10266, opened 2022-06-09
feat(GRO-1061): Render artist insights data on Artist overview page

TMcMeans:
The type of this PR is: feat 

This PR solves: [GRO-1061] 

as part of the Enhanced Artist Metadata epic: [GRO-256]
 
### Description

This PR is to show existing and new Artist insight data on the `/artist` overview page. 

Currently, there are some insights data already living in gravity (i.e. `SOLO_SHOW, GROUP_SHOW, REVIEWED`) which are being used to populate "Career Highlights" but I've also done some work on the backend with an import rake task to auto import new data into gravity, which includes `active_secondary_market`, `critically_acclaimed` and `artsy_vanguard`.

TODOs:
- [x] Clean up dead and commented out code
- [x] Refactor tests
- [x] Responsiveness on mobile
- [ ] Create a review app/feature flag 

Screenshot of how the new artist overview page looks for reference: 
Screen Shot 2022-06-17 at 3 35 18 PM

[GRO-1061]: https://artsyproduct.atlassian.net/browse/GRO-1061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GRO-256]: https://artsyproduct.atlassian.net/browse/GRO-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi (commented):
Yeah, this is way way better than futzing with updating that super old react class code ❇️

dzucconi:
> Create a blue pill/button component that mirrors the badges component

https://palette-storybook.artsy.net/?path=/story/components-pill--variants

```tsx
<Pill variant="badge" disabled>Hello</Pill>
```

src/v2/Components/ArtistInsightsBadges.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export const ArtistInsightsBadges: FC<ArtistInsightsBadgesProps> = ({
+  artist,
+}) => {
+  if (
+    !artist.insights ||
+    !artist.artistHighlights ||
+    !artist.auctionResultsConnection
+  )
+    return null
```
damassi, line 82: This is a super minor nit, but it does make the code less error prone: For single-line conditional returns its totally fine to omit brackets, eg 
```tsx
if (!foo) return null
```
but as soon as things are multiline as above, brackets should be added:
```tsx
if (...) {
  return null
}
```
Another dev without prettier could come along and accidentally add a line break between the `if` and the return and be wondering whats going on; meaning, it's prone to subtle errors.

src/v2/Components/ArtistInsightsPills.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+export const ArtistInsightsPills: FC<ArtistInsightsPillsProps> = ({
+  artist,
+}) => {
+  if (
+    !artist.insights ||
+    !artist.artistHighlights ||
+    !artist.auctionResultsConnection
+  )
+    return null
```
damassi, line 51: ditto

src/v2/Components/ArtistInsightsBadges.tsx
TMcMeans: and its totally ok to have these many conditional checks for data existing right? just making sure my code is sound (after I've updated to multiline conditional checks)

damassi: Oh yeah totally fine! Its all very readable.

PR #10273, opened 2022-06-10
[fix]: AlicanRachelKiril/pull-paymentMethod-upon-setting-payment

AlicanAkyuz:

### Description

This PR fixes the problem that users choosing credit card had to click Continue 2 times in orders to move to the review step. Luckily, this bug did not affect users on productions, since it's caused by new payment method integration, which is behind a feature flag. 

We solved the problem by pulling the paymentMethod on Order when payment is set, so that the redirecting logic could safely check if creditCard info exists in an order when paymentMethod is CreditCard.


kiryl-zubarau (dismissed):

starsirius (dismissed):
Thank you all for the quick fix! 🥇

fladson (dismissed):

PR #10277, opened 2022-06-10
enhancement: return 404 when identity verification does not exist

ovasdi:

This PR solves [PLATFORM-4165](https://artsyproduct.atlassian.net/browse/PLATFORM-4165)

### Description

This is a followup to https://github.com/artsy/force/pull/10263. After removing [this logic](https://github.com/artsy/force/pull/10263/files#diff-5e4ebdf40b75b0a9dac5f65a77a5e245dff68b03561e90c7259fefecc403bd2dL39) and the [wrong-owner component](https://github.com/artsy/force/pull/10263/files#diff-5e8cbe37fe55c565db5dfb15eaa537c04376af52e84c62af9370006930c54bf6L11) I introduced a regression and so now when requesting an identity verification that does not exist the client throws a _500 internal server error_ whereas before we rendered a _wrong-owner_ component. 

To address this regression I thought it would be good to return a _404_ when an identity verification is not found.

Before https://github.com/artsy/force/pull/10263:
Screen Shot 2022-06-10 at 11 19 19 AM

Now:
Screen Shot 2022-06-10 at 2 20 31 PM

### DIFF
--- a/src/v2/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
+++ b/src/v2/Apps/IdentityVerification/IdentityVerificationApp/index.tsx
@@ -18,6 +18,7 @@ import { CompletePassed } from "./CompletePassed"
 import { CompleteWatchlistHit } from "./CompleteWatchlistHit"
 import { RouterLink } from "v2/System/Router/RouterLink"
 import { MetaTags } from "v2/Components/MetaTags"
+import { HttpError } from "found"
 
 const logger = createLogger("IdentityVerificationApp.tsx")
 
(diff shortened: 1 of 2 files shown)


ovasdi:
@damassi i've tried a few different ways to write a test for this state but haven't been able to nail it down. I thought it would be good to cover this state [here](https://github.com/artsy/force/blob/ovasdi/PLATFORM-4165/idv-followup-404/src/v2/Apps/IdentityVerification/__tests__/IdentityVerificationApp.jest.tsx#L48). I tried the following thinking that if response comes back null, it would hit that condition but that doesn't seem to work.

```
      it("returns a 404 when the identity verification is not found", async () => {
        const env = setupTestEnv()
        const page = await env.buildPage({
          mockData: deepMerge(IdentityVerificationAppQueryResponseFixture, {
            identityVerification: null,
          }),
        })

        expect(page.text()).toContain("404")
      })
```

damassi:
A simpler way to test this would be to assert that the page throws an error. Try this:

```tsx
expect(async () => {
  await env.buildPage({
    mockData: deepMerge(IdentityVerificationAppQueryResponseFixture, {
    identityVerification: null,
  }),
}).toThrow()
```

PR #10285, opened 2022-06-13
chore: add ability to pass context module prop for Details and Metadata components

dimatretyak:

### Description
Added the ability to pass the `contextModule` prop, the value of which is used for the "save" button

### DIFF
--- a/src/v2/Components/Artwork/GridItem.tsx
+++ b/src/v2/Components/Artwork/GridItem.tsx
@@ -117,6 +117,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
         artwork={artwork}
         isHovered={isHovered}
         shouldShowHoverSaveButton={!!shouldShowHoverSaveButton}
+        contextModule={contextModule || ContextModule.artworkGrid}
       />
     </div>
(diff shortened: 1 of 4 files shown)


gkartalis (approved):
nice work!

src/v2/Components/Artwork/GridItem.tsx
```diff
@@ -117,6 +117,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
         artwork={artwork}
         isHovered={isHovered}
         shouldShowHoverSaveButton={!!shouldShowHoverSaveButton}
+        contextModule={contextModule || ContextModule.artworkGrid}
```
gkartalis, line 120: **thought(non-blocking):** what do you think of using `??` in favor of `||` here?

```suggestion
        contextModule={contextModule ?? ContextModule.artworkGrid}
```

since `??` will return the right value when `undefined` or `null` which will never happen while `||` will return the right value if left is falsy, but falsy in our case will be only `undefined`.

Does it makes sense? What do you think?

nickskalkin (dismissed):
👍

PR #10304, opened 2022-06-14
refactor(v2): removes modal dialog; majority of sans

dzucconi:
There's a few instances of sans in the legacy app pages (TODO). Otherwise, completely removes `Sans` and the legacy `ModalDialog`


src/v2/Apps/Order/Routes/__tests__/Reject.jest.tsx
```diff
@@ -15,6 +15,22 @@ jest.mock("v2/Utils/getCurrentTimeAsIsoString")
 const NOW = "2018-12-05T13:47:16.446Z"
 require("v2/Utils/getCurrentTimeAsIsoString").__setCurrentTime(NOW)
 
+jest.mock("@artsy/palette", () => {
+  return {
+    ...jest.requireActual("@artsy/palette"),
```
araujobarret, line 20: I hope you don't mind but in the other tests I decided to focus solely in the modal function call like:
```jsx
const mockShowErrorDialog = jest.fn()
jest.mock("v2/Apps/Order/Dialogs", () => ({
  ...jest.requireActual("../../Dialogs"),
  injectDialog: Component => props => (
    <Component {...props} dialog={{ showErrorDialog: mockShowErrorDialog }} />
  ),
}))

// expect mockShowErrorDialog to be called with bla bla
```

So later on I will refactor the missing ones with the same pattern if it's okay on your side

PR #10305, opened 2022-06-14
bugfix(New For You) - Fix grid ordering, expand and cap page size, add login messaging

laurabeth:

This PR solves [GRO-1046] [GRO-1035] [GRO-1079]

### Description

Combo PRs are, in hindsight, a bad idea. But, here we go:

The New for You email grids load horizontally, while grids using the `Masonry` component (as was used in the initial implementation of the corresponding New For You page) load vertically. This was causing some disconnect between the associated recommendations and possibly some confusion for the users, thus we updated the NFY page here to sync the results display. 

Due to the nature of the query used to generate `New for You` recommendations and supplemental backfill, using pagination is not currently an option for this interface. We landed on a quick fix of showing users their top 20 recommendations. This number can be easily changed if required.

The previous implementation of NFY sent unregistered/unauthenticated users to the home page instead of `/new-for-you`. We have changed this behavior to instead allow unauthed users to view the page populated with backfill, while displaying the following messaging: 
`Already have an account? Log in to see your personalized recommendations.`

Screen Shot 2022-06-14 at 6 31 25 PM

[GRO-1046]: https://artsyproduct.atlassian.net/browse/GRO-1046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1035]: https://artsyproduct.atlassian.net/browse/GRO-1035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1079]: https://artsyproduct.atlassian.net/browse/GRO-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
```diff
@@ -1,80 +1,48 @@
...
-        setLoading(false)
-      }, 2500)
-    })
-  }
-
   return (
     <>
       {artworks?.length > 0 ? (
-        <Masonry columnCount={[2, 3, 4]}>
+        <GridColumns gridRowGap={[2, 4]} alignItems="flex-end">
```
dzucconi, line 20: We generally only use `GridColumns` for page layout. What you're looking for is [`ArtworkGrid`](https://github.com/artsy/force/tree/main/src/v2/Components/ArtworkGrid)

src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
```diff
@@ -1,80 +1,48 @@
...
-              </Fragment>
+              </Column>
             )
           })}
-        </Masonry>
+        </GridColumns>
       ) : (
         <Text variant="lg" mt={4} color="black60">
           Nothing yet.
         </Text>
```
dzucconi, line 33: Should maybe be a `Message`?

src/v2/Apps/NewForYou/NewForYouApp.tsx
```diff
@@ -1,23 +1,33 @@
...
       </Text>
-      <Spacer mt={4} />
+      <Spacer mt={2} />
+      {!isLoggedIn && (
+        <Message variant="warning">
+          Already have an account? <RouterLink to="/login">Log in</RouterLink>{" "}
+          to see your personalized recommendations.
+        </Message>
+      )}
+      <Spacer mt={2} />
```
dzucconi, line 30: I still think both of these `Spacer`s should be nested within the conditionals? For instance if you're logged in you get a stack of size 2 Spacers? I don't know what this should be exactly but it seems like it'd be weirdly inconsistent depending on login/out and whether or not there is data.

src/v2/Apps/NewForYou/__tests__/NewForYouApp.jest.tsx
```diff
@@ -30,6 +38,31 @@ describe("NewForYouApp", () => {
...
+
+    expect(
+      screen.queryByText(/(^Already have an account\?)/g)
+    ).not.toBeInTheDocument()
+    expect(screen.queryByText(/(^Log in)/g)).not.toBeInTheDocument()
+    expect(
+      screen.queryByText(/(to see your personalized recommendations\.$)/g)
+    ).not.toBeInTheDocument()
+  })
+
```
dzucconi, line 65: 👍

src/v2/Apps/NewForYou/Components/NewForYouArtworksGrid.tsx
laurabeth: fixed

src/v2/Apps/NewForYou/NewForYouApp.tsx
laurabeth: Fixed

TMcMeans:
lgtm!

TMcMeans:
lgtm!

PR #10307, opened 2022-06-15
feat(FX-4041): Add active filter pills to partner page's artwork grid

dimatretyak:

This PR solves [FX-4041]

Review app: https://selected-filters-for-partner.artsy.net

### Context
We’ve added active filter pills to the artist page’s artworks grid a while ago, which is also where the “Create alert” control lives. We’d like to first expand these active filter pills to other grids before we also support alert creation on those surfaces.

image

### Acceptance Criteria
* As a collector
* When I visit a partner’s artwork grid
* And I enable at least one filter
* Then I should see my active filters represented as pills above the artwork grid
* And "Create Alert" button should **NOT** be displayed

#### Quick links
* [Aicon Gallery](https://selected-filters-for-partner.artsy.net/partner/aicon-gallery/works)
* [Paragon](https://selected-filters-for-partner.artsy.net/partner/paragon/works)
* [Galerie Denise René](https://selected-filters-for-partner.artsy.net/partner/galerie-denise-rene/works)

### Demo

[FX-4041]: https://artsyproduct.atlassian.net/browse/FX-4041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Partner/Routes/Works/index.tsx
+++ b/src/v2/Apps/Partner/Routes/Works/index.tsx
@@ -10,6 +10,7 @@ import {
 import { BaseArtworkFilter } from "v2/Components/ArtworkFilter"
 import { updateUrl } from "v2/Components/ArtworkFilter/Utils/urlBuilder"
 import { Match, RouterState, withRouter } from "found"
+import { ActiveFilterPills } from "v2/Components/SavedSearchAlert/Components/ActiveFilterPills"
 
 interface PartnerArtworkFilterProps {
   partner: Works_partner
@@ -42,7 +43,12 @@ export const Artworks: React.FC<PartnerArtworkFilterProps> = ({
       }
       counts={sidebar?.counts as Counts}
     >
-      <BaseArtworkFilter relay={relay} offset={200} viewer={partner} />
+      <BaseArtworkFilter
+        relay={relay}
+        offset={200}
+        viewer={partner}
+        renderFilterPills={() => <ActiveFilterPills />}
+      />
     </ArtworkFilterContextProvider>
   )
 }
--- a/src/v2/Apps/Partner/Routes/Works/index.tsx
+++ b/src/v2/Apps/Partner/Routes/Works/index.tsx
@@ -47,7 +47,7 @@ export const Artworks: React.FC<PartnerArtworkFilterProps> = ({
         relay={relay}
         offset={200}
         viewer={partner}
-        renderFilterPills={() => <ActiveFilterPills />}
+        FilterPills={<ActiveFilterPills />}
       />
     </ArtworkFilterContextProvider>
   )
--- a/src/v2/Apps/Partner/Routes/__tests__/Works.jest.tsx
+++ b/src/v2/Apps/Partner/Routes/__tests__/Works.jest.tsx
@@ -3,13 +3,17 @@ import { ArtworksRefetchContainer } from "../Works"
 import { graphql } from "react-relay"
 import { Works_Query } from "v2/__generated__/Works_Query.graphql"
 import { useTracking } from "v2/System/Analytics/useTracking"
-import { setupTestWrapper } from "v2/DevTools/setupTestWrapper"
+import {
+  setupTestWrapper,
+  setupTestWrapperTL,
+} from "v2/DevTools/setupTestWrapper"
 import {
   artistAggregation,
   artistNationalityAggregation,
   materialsTermsAggregation,
   mediumAggregation,
 } from "test/fixtures/aggregations"
+import { fireEvent, screen, within } from "@testing-library/react"
 
 jest.unmock("react-relay")
 jest.mock("v2/System/Router/useRouter", () => ({
@@ -32,6 +36,22 @@ const { getWrapper } = setupTestWrapper<Works_Query>({
   ),
   query: graphql`
     query Works_Query($partnerId: String!) @relay_test_operation {
+      partner(id: $partnerId) {
+        ...Works_partner
+      }
+    }
+  `,
+  variables: { partnerId: "pontone-gallery" },
+})
+
+const { renderWithRelay } = setupTestWrapperTL<Works_Query>({
+  Component: ({ partner }) => (
+    <MockBoot user={{ id: "percy-z" }}>
+      <ArtworksRefetchContainer partner={partner!} />
+    </MockBoot>
+  ),
+  query: graphql`
+    query WorksTL_Query($partnerId: String!) @relay_test_operation {
       partner(id: $partnerId) {
         ...Works_partner
       }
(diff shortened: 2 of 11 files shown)


src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -275,10 +275,11 @@ export const BaseArtworkFilter: React.FC<
 
           <Spacer mb={2} />
 
-          {enableCreateAlert && savedSearchEntity && (
+          {savedSearchEntity && (
```
gkartalis, line 278: **suggestion:** what do you think of changing that to 
```suggestion
          {!!savedSearchEntity && (
```
just to make sure that nothing weird will happen. not saying that this won't work in react but I think it is better to explicitly make it to a boolean over a falsy value, what do you think?

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -335,10 +336,11 @@ export const BaseArtworkFilter: React.FC<
             // Safe to remove once artwork masonry uses CSS grid.
             width="100%"
           >
-            {enableCreateAlert && savedSearchEntity && (
+            {savedSearchEntity && (
```
gkartalis, line 339: same here as above

src/v2/Components/SavedSearchAlert/Components/__tests__/SavedSearchAlertArtworkGridFilterPills.jest.tsx
```diff
@@ -78,4 +80,70 @@ describe("SavedSearchAlertArtworkGridFilterPills", () => {
 
     expect(screen.getByText("Banksy")).toBeInTheDocument()
   })
+
+  it("display followed artist pills when 'Artists I Follow' is selected", () => {
```
gkartalis, line 84: **praise:** thanks for the consistent great test cases as always!

gkartalis (approved):

src/v2/Apps/Partner/Routes/__tests__/Works.jest.tsx
```diff
@@ -122,4 +142,63 @@ describe("PartnerArtworks", () => {
       expect(filterWrappers.at(filterIndex).prop("expanded")).toEqual(expanded)
     })
   })
+
+  it("should render selected filters as pills", () => {
```
gkartalis, line 146: **praise:** 👏

src/v2/Components/SavedSearchAlert/Components/__tests__/SavedSearchAlertArtworkGridFilterPills.jest.tsx
dimatretyak: Thanks!

src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
```diff
@@ -385,14 +385,7 @@ const SAVED_SEARCH_ALERT_EDIT_FORM_QUERY = graphql`
...
       artistIDs: $artistIds
-      aggregations: [
-        ARTIST
-        LOCATION_CITY
-        MATERIALS_TERMS
-        MEDIUM
-        PARTNER
-        COLOR
-      ]
+      aggregations: [LOCATION_CITY, MATERIALS_TERMS, MEDIUM, PARTNER, COLOR]
```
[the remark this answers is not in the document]
dimatretyak: We don't need to use data from the `ARTIST` aggregation here, since we get information about artists [here](https://github.com/artsy/force/blob/3742a208af183454173f22e7d19511d80773ce0d/src/v2/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx#L344-L354)

gkartalis (dismissed):

src/v2/Components/ArtworkFilter/index.tsx
```diff
@@ -275,10 +275,11 @@ export const BaseArtworkFilter: React.FC<
 
           <Spacer mb={2} />
 
-          {enableCreateAlert && savedSearchEntity && (
+          {!!savedSearchEntity && (
```
mdole, line 278: @dimatretyak I think we could do a little more "separation of concerns" here. Instead of rendering both filter pills and the "create alert" button when a `savedSearchEntity` is passed, what do you think about separating them so that rendering filter pills is based on a `renderFilterPills` prop and the button on a `renderCreateAlertButton` prop?

src/v2/Apps/Partner/Routes/Works/index.tsx
```diff
@@ -24,6 +26,17 @@ export const Artworks: React.FC<PartnerArtworkFilterProps> = ({
...
+    placeholder: "Partner Page", // TODO: Clarify and update it later
+    defaultArtists: [],
+    owner: {
+      type: OwnerType.partner,
+      slug: partner.slug,
+      id: partner.internalID,
+      name: partner.name ?? "",
+    },
+  }
+
```
mdole, line 39: **minor, non-blocking:** suggest removing this and moving it to a separate PR since it's not necessarily related to rendering the filter pills (see my other comment about making those separate concerns)

mdole (commented):
really nice work @dimatretyak! Left a suggestion about splitting out the logic for rendering pills vs. rendering the create alert button, let me know what you think 🙂

src/v2/Components/ArtworkFilter/index.tsx
dimatretyak: In theory, this is a good idea 🤔. But it seems to me that such changes will only complicate our logic and we will often have to write the duplicate code.

dimatretyak: If it doesn't bother you, can you give an example of the benefits that we will get if we add these changes?

dimatretyak: PR #10391 (simpler solution)
PR #10391 (a more complex solution with duplication of information in some places, but it gives us more flexibility)

src/v2/Apps/Partner/Routes/Works/index.tsx
```diff
@@ -24,6 +26,17 @@ export const Artworks: React.FC<PartnerArtworkFilterProps> = ({
 }) => {
   const { sidebar } = partner
 
+  const savedSearchEntity: SavedSearchEntity = {
+    placeholder: "Partner Page", // TODO: Clarify and update it later
```
nickskalkin, line 30: Don't forget to address this comment :)

nickskalkin (approved):
looks good to me 👍

nickskalkin (approved):

PR #10308, opened 2022-06-15
feat(NX-3140): add purchase submitted message in conversations

tam-kis:

This PR solves [NX-3140](https://artsyproduct.atlassian.net/browse/NX-3140)

### Description

Added new status for purchase submitted in conversations. 
Related to this PR in Exchange: [1202](https://github.com/artsy/exchange/pull/1202)

Screenshot 2022-06-15 at 15 11 56


src/v2/Apps/Conversation/Components/OrderUpdate.tsx
```diff
@@ -55,16 +55,20 @@ export const OrderUpdate: React.FC<OrderUpdateProps> = ({
...
       message = `Offer Accepted`
-    } else if (state === "CANCELED" && stateReason?.includes("_rejected")) {
+    } else if (orderUpdateState === "offer_rejected") {
       color = "red100"
       message = `Offer Declined`
-    } else if (state === "CANCELED" && stateReason?.includes("_lapsed")) {
+    } else if (orderUpdateState === "offer_lapsed") {
       color = "black60"
       message = `Offer Expired`
+    } else if (orderUpdateState === "buy_submitted") {
```
araujobarret, line 68: maybe worth to check what happens if an order is canceled, we still show the `You purchased this artwork` for this case but in reality, the order was canceled or refunded.

araujobarret (approved):
Looks good, I would refactor lots if `if else if` to a more functional way but it's okay for now, I will change it in Eigen to make the component clean, we can replicate here later 👍

src/v2/Apps/Conversation/Components/OrderUpdate.tsx
tam-kis: We will check this with QA and create a new ticket for all the states.

PR #10315, opened 2022-06-15
Fix typos

joeyAghion:
Noticed while debugging mysterious auction registration errors. (The other style changes are automatic, but let me know if you see a problem.)


dzucconi (approved):
😄 thank u

PR #10316, opened 2022-06-15
fix: render status

dzucconi:
The theory here is that some race condition has been exposed and a state set is happening on an unmounted component. `useEffect` runs only on the initial mount (server rehydration), and since the set states happen in the render body directly; here I've just stuck the timeouts in refs then clear them in the opposing conditional branches.

### DIFF
--- a/src/v2/System/Router/RenderStatus.tsx
+++ b/src/v2/System/Router/RenderStatus.tsx
@@ -10,11 +10,14 @@ import { PageLoader } from "./PageLoader"
 import { AppShell } from "v2/Apps/Components/AppShell"
 import { getENV } from "v2/Utils/getENV"
 import { HttpError } from "found"
+import { FC, useRef } from "react"
 
 const logger = createLogger("Artsy/Router/Utils/RenderStatus")
 
 export const RenderPending = () => {
   const { isFetching, setFetching } = useSystemContext()
+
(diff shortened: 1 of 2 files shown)


src/v2/System/Router/RenderStatus.tsx
```diff
@@ -10,79 +10,89 @@ import { PageLoader } from "./PageLoader"
...
    * First, set fetching to ensure that components that are listening for this
    * value have a chance to respond to the fetching state. This is necessary
    * because the `<Renderer>` component below will freeze all updates for the
    * duration of the fetch.
    */
+
   if (!isFetching) {
-    setTimeout(() => setFetching?.(true), 0)
+    timeoutRef.current = setTimeout(() => setFetching?.(true), 0)
+    return undefined
```
dzucconi, line 31: Returning `null` here clears the screen completely.

src/v2/System/Router/RenderStatus.tsx
```diff
@@ -10,79 +10,89 @@ import { PageLoader } from "./PageLoader"
...
+  logger.error(error.data)
 
-  const status = props.error.status || 500
+  const status = error.status || 500
 
   const { isFetching, setFetching } = useSystemContext()
 
+  useEffect(() => {
+    console.log("RenderStatus: Error")
+  }, [])
```
dzucconi, line 88: Woops.

PR #10321, opened 2022-06-16
feat: define initial value for payment method picker

kiryl-zubarau:

This PR solves [TX-400]

### Description

This PR defines the initial value for the payment method picker on the frontend side instead of the backend.

### DIFF
--- a/src/v2/Apps/Order/Utils/__tests__/orderUtils.jest.ts
+++ b/src/v2/Apps/Order/Utils/__tests__/orderUtils.jest.ts
@@ -1,0 +1,90 @@
+import { getInitialPaymentMethodValue } from "./../orderUtils"
+import { isPaymentValid } from "../orderUtils"
+import {
+  CommercePaymentMethodEnum,
+  Payment_order,
+} from "v2/__generated__/Payment_order.graphql"
+
+describe("order utils", () => {
+  describe("is payment valid", () => {
+    it("returns false if payment method unknown", () => {
+      expect(isPaymentValid({ __typename: "%other" })).toEqual(false)
+    })
+
+    it("returns false if payment method details empty", () => {
+      expect(isPaymentValid(null)).toEqual(false)
+    })
+
+    describe("credit card", () => {
+      it("returns false if payment method is wire transfer", () => {
+        expect(isPaymentValid({ __typename: "CreditCard", id: "id" })).toEqual(
+          true
+        )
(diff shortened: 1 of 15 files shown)


AlicanAkyuz (approved):
looks pretty good, I just put a minor comment about the tests.

src/v2/Apps/Order/Utils/__tests__/orderUtils.jest.ts
```diff
@@ -0,0 +1,90 @@
...
+    it("returns false if payment method details empty", () => {
+      expect(isPaymentValid(null)).toEqual(false)
+    })
+
+    describe("credit card", () => {
+      it("returns false if payment method is wire transfer", () => {
+        expect(isPaymentValid({ __typename: "CreditCard", id: "id" })).toEqual(
+          true
+        )
+      })
```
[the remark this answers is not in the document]
kiryl-zubarau: There is a big typo in the name :smile:  Fixed. And yes, isPaymentValid should return true if method is credit card and has id?

src/v2/Apps/Order/redirects.tsx
```diff
@@ -278,5 +264,6 @@ graphql`
     creditCard {
       internalID
     }
+    ...Payment_validation @relay(mask: false)
```
starsirius, line 267: Question: Do we still need `mask: false` in both places if the components use them directly? Or maybe I missed something here.

src/v2/Apps/Order/Utils/orderUtils.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  CommercePaymentMethodEnum,
+  Payment_order,
+} from "v2/__generated__/Payment_order.graphql"
+
+export const isPaymentValid = (
```
starsirius, line 6: #non-blocking. `isPaymentSet` reads better to me because we're not really validating payment.

starsirius (dismissed):
💯  Had a few non-blocking comments and feel free to merge when ready!

src/v2/Apps/Order/redirects.tsx
kiryl-zubarau: If I'm not mistaken then yes, we have to use `mask: false` to get access to the fields described in the fragment from the generated field. In another case, we will have only link to the fragment and we don't have access to the fields.

PR #10327, opened 2022-06-16
Another (Another) Review App for Debugging

dzucconi:
Going to try to manually bisect with a review app; starting at https://github.com/artsy/force/commit/d5682c319975d456ed1977a008bd95f0fcf275f4


dzucconi:
OK: that either wasn't far enough back or the bug is unrelated to recent changes to the artwork page.

PR #10335, opened 2022-06-17
feat: My Collection

olerichter00:

## Description

This PR introduces the initial My Collection page under `/settings/my-collection`. The screen only shows a paginated grid with the artworks of the user's My Collection 

This PR introduces the My Collection page under `/settings/my-collection`. The page shows only a paginated grid with the user's My Collection artworks.

The page is not yet linked in the menu or anywhere else.

This PR solves [CX-2636](https://artsyproduct.atlassian.net/browse/CX-2636)


src/v2/Apps/Settings/Routes/MyCollection/Components/MyCollectionGridItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export const MyCollectionArtworkGridItem: React.FC<MyCollectionArtworkGridItemProps> = ({
+  artwork,
+  lazyLoad = true,
+  contextModule,
+  onClick,
+  ...rest
+}) => {
+  const { user } = useSystemContext()
+  const isTeam = userIsTeam(user)
```
MounirDhahri, line 29: nice hook 😄

olerichter00: I just realised that I can right-click on an image when I'm logged in with my Artsy account :)

src/v2/Apps/Settings/Routes/MyCollection/SettingsMyCollectionRoute.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { SettingsMyCollectionRoute_me } from "v2/__generated__/SettingsMyCollectionRoute_me.graphql"
+import { SettingsMyCollectionArtworksQueryRenderer } from "./Components/SettingsMyCollectionArtworks"
+
+interface SettingsMyCollectionRouteProps {
+  me: SettingsMyCollectionRoute_me
+}
+
+const SettingsMyCollectionRoute: React.FC<SettingsMyCollectionRouteProps> = ({
```
olerichter00, line 10: I mostly copied the code from [SettingsSavesRoute](https://github.com/artsy/force/blob/282d0b6779b2aad7d53f13748dea81695ae4951a/src/v2/Apps/Settings/Routes/Saves/SettingsSavesRoute.tsx#L14).

@dzucconi is there a better/more up-to-date approach I should use instead?

dzucconi:
What's the reasoning for the custom artwork brick?

src/v2/Apps/Settings/Routes/MyCollection/Components/MyCollectionArtworkGridItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+  {
+    artwork: graphql`
+      fragment MyCollectionArtworkGridItem_artwork on Artwork {
+        internalID
+        title
+        image_title: imageTitle
+        image {
+          placeholder
+          url(version: "large")
+          aspect_ratio: aspectRatio
```
dzucconi, line 128: We can ditch this snake casing wherever possible.

src/v2/Apps/Settings/Routes/MyCollection/Components/MyCollectionGridItem.tsx
dzucconi: The fact that we disable this bothers me to no end :)

src/v2/Apps/Settings/Routes/MyCollection/SettingsMyCollectionRoute.tsx
dzucconi: Looks good to me.

dzucconi (approved):
Obviously the brick component doesn't look ready for production, but I'm assuming that there will be follow up. Why do we need a special component here though?

dzucconi:
@olerichter00 cool + I agree with the approach. Though again, why would you not want them clickable? If the detail text stayed the same (it is somewhat configurable BTW), you could just use `pointer-events: none` to disable the interaction.

olerichter00:
> @olerichter00 cool + I agree with the approach. Though again, why would you not want them clickable? If the detail text stayed the same (it is somewhat configurable BTW), you could just use `pointer-events: none` to disable the interaction.

We still don't have a design for the artwork detail page (which will be different from our normal artwork page). I'll give it a try 👍

dzucconi:
> We still don't have a design for the artwork detail page

Oh wow — so there's going to be two different artwork pages?

olerichter00:
@dzucconi I've removed the custom grid item in https://github.com/artsy/force/pull/10335/commits/eaac2871cb6f91097a11f0af099b151e044bbb5e. What do you think?

dzucconi:
@olerichter00 Yeah I think that works well.

PR #10341, opened 2022-06-17
feat: Correctly show fair artwork count to everyone GRO-1065

jonallured:
This ticket was prompted by seeing the total count on a fair artwork list not match the total in the tab. I looked at https://github.com/artsy/force/pull/10168 to see what a fix might be and along the way discovered that part of the problem was that we were suppressing this total count when not signed in. So I have two commit heres that first fix the count for signed in users and then removes that logic to show the count to all users. More chatter here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1655474029822989

We couldn't think of any reason to hide the count for signed out but if something comes up we can revert that last commit and be back to working order. Either way, it looks like this now:

## signed out

Screen Shot 2022-06-17 at 9 12 23 AM

## signed in

Screen Shot 2022-06-17 at 9 12 18 AM

https://artsyproduct.atlassian.net/browse/GRO-1065

/cc @artsy/grow-devs


src/v2/Apps/Fair/Routes/FairArtworks.tsx
```diff
@@ -112,23 +111,19 @@ export const FairArtworksRefetchContainer = createRefetchContainer(
         }
         filtered_artworks: filterArtworksConnection(first: 30, input: $input) {
           id
-          counts @include(if: $shouldFetchCounts) {
```
dzucconi, line 115: 👍

dzucconi (approved):

PR #10343, opened 2022-06-17
fix: safely loads zendesk

dzucconi:
This PR removes `react-zendesk` and better encapsulates its initialization.

### DIFF
--- a/src/v2/Apps/Artwork/Components/ArtworkZendesk.tsx
+++ b/src/v2/Apps/Artwork/Components/ArtworkZendesk.tsx
@@ -1,0 +1,68 @@
+import { FC, useMemo } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { ZendeskWrapper } from "v2/Components/ZendeskWrapper"
+import { isExceededZendeskThreshold } from "v2/Utils/isExceededZendeskThreshold"
+import { ArtworkZendesk_artwork } from "v2/__generated__/ArtworkZendesk_artwork.graphql"
+
+interface ArtworkZendeskProps {
+  artwork: ArtworkZendesk_artwork
+}
+
+const ArtworkZendesk: FC<ArtworkZendeskProps> = ({ artwork }) => {
+  const {
+    isAcquireable,
+    isInquireable,
+    isOfferable,
+    listPrice,
+    isInAuction,
+  } = artwork
+
+  const price = useMemo(() => {
+    if (!listPrice) return null
+
+    switch (listPrice.__typename) {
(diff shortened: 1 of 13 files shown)


src/v2/Utils/isExceededZendeskThreshold.ts
```diff
@@ -1,11 +1,14 @@
...
-      GBP: 7000,
-    }[currencyCode] < amount
-  )
+const CURRENCIES = {
+  USD: 10000,
+  AUD: 13000,
+  EUR: 8000,
+  HKD: 77000,
+  GBP: 7000,
+} as const
```
jonallured, line 7: 🤩

src/v2/Apps/Artwork/Components/ArtworkZendesk.tsx
```diff
@@ -0,0 +1,64 @@
...
+  if (isInquireable && !isAcquireable && !isOfferable) {
+    return null
+  }
+
+  const price =
+    listPrice?.__typename === "Money"
+      ? listPrice
+      : listPrice?.__typename === "PriceRange"
+      ? listPrice.maxPrice
+      : null
```
jonallured, line 30: **nitpick [non-blocking]:** avoid nested ternary

Could this be a map instead of a nested ternary? I'm thinking about something like:

```ts
const map = {
  Money: listPrice,
  PriceRange: listPrice.maxPrice
}
const price = typeToPriceMap[listPrice?.__typename]
```

What do you think?

src/v2/Components/ZendeskWrapper.tsx
```diff
@@ -1,17 +1,37 @@
...
+  mode = "default",
+}) => {
+  const key = MODES[mode]
+
+  useLoadScript({
+    id: "ze-snippet",
+    src: `https://static.zdassets.com/ekr/snippet.js?key=${key}`,
+    removeOnUnmount: true,
+    onReady: () => {
+      window.zEmbed?.show?.()
```
jonallured, line 24: 🤩

jonallured (approved):
Nice one!!

src/v2/Apps/Artwork/Components/ArtworkZendesk.tsx
dzucconi: Yeah that's nicer. Forgot to replace this, I just copied this logic from the old component.

dzucconi: Replaced with a switch statement for type safety.

PR #10351, opened 2022-06-20
chore(FX-4052): Hide "Create Alert" button on artwork page if there are no associated artists

dimatretyak:

This PR solves [FX-4052]

Review app: https://hide-create-alert-cta.artsy.net/

### Description
At the moment, there may be a situation where a user can create a saved search alert for an artwork that does not have associated artists. For this reason, we get undesirable behavior and such alerts are generating a lot of "activity" events (when a work that matches their criteria is published for the first time, meaning all new works) and are definitely not alerting the users to what they thought they were getting when they set them

### Acceptance Criteria
* Button should be hidden, not disabled
* "Create Alert" button should be hidden if there are no associated artists
* Add changes only for the artwork page

#### [Artwork](https://hide-create-alert-cta.artsy.net/artwork/prince-hunting-with-cheetah)
| Before | After |
| before-1 | after-1 |

#### [Bidding Closed Artwork](https://hide-create-alert-cta.artsy.net/artwork/an-okeefe-and-merritt-stove)
| Before | After |
| before-2 | after-2 |

#### Sold artwork ([production link](https://www.artsy.net/artwork/unknown-20th-century-sculptural-wood-lamp))
| Before | After |
| before-3 | after-3 |

[FX-4052]: https://artsyproduct.atlassian.net/browse/FX-4052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -410,7 +414,8 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
       artwork.is_acquireable || artwork.is_offerable
     )
     const shouldRenderButtons = artworkEcommerceAvailable || !!isInquireable
-    const isSecondaryContactGalleryButton = isOfferable || isSold
+    const createAlertAvailable = isSold && (artwork.artists?.length ?? 0) > 0
```
gkartalis, line 417: 🔥👏

src/v2/Apps/Artwork/Components/ArtworkSidebar/index.tsx
```diff
@@ -64,7 +64,8 @@ export const ArtworkSidebar: React.FC<ArtworkSidebarProps> = ({
   const shouldHideDetailsCreateAlertCTA =
     (is_in_auction && hasEnded) ||
     (is_in_auction && lotIsClosed(sale, saleArtwork)) ||
-    is_sold
+    is_sold ||
+    artwork.artists?.length === 0
```
gkartalis, line 68: **suggestion(non-blocking):** what do you think of moving the `artwork.artists?.length === 0` in the beginning of the check (start of [line 65](https://github.com/artsy/force/pull/10351/files#diff-aacf74b81201e9b848f011f880a9891dc317f43d2a1c2d3098ad490d3eda6278R65)) in order to hide it before checking the rest?

gkartalis (approved):
👏 thanks for the test cases and the screenshots 🔥

src/v2/Apps/Artwork/Components/ArtworkSidebar/index.tsx
dimatretyak: Yes, I agree. I have nothing against it 😅

mdole (approved):
nice! thanks for patching this up 👷

PR #10369, opened 2022-06-22
feat(institution-partnerships): adds rotating/crossfading figures to header; removes profiles

dzucconi:
Turns out it was a bit of a pain to add the cross fade ability to the `FullBleedHeader` because of how it does masking for the sticky effect.

Restructures a bit and adds a `FullBleedHeaderFader` component to do this.


src/v2/Components/FullBleedHeader/FullBleedHeaderFader.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    : {
+                        opacity: 0,
+                        pointerEvents: "none",
+                      },
+              })}
+        >
+          {figure.children ?? children}
+        </FullBleedHeaderWithTransition>
+      ))}
+    </Box>
```
dzucconi, line 66: Stacks the current and next header on top of one another. Waits a moment then transitions the opacity of the next one which is stacked on top. Once it's done transitioning we move to the next, moving the top to the bottom, then stacking a new one on top with an opacity of 0. Then repeat.

PR #10370, opened 2022-06-22
feat(NX-3187): add all purchase states in conversations

tam-kis:

This PR solves [NX-3187]

### Description

Add missing states for Purchase in conversation: `buy_approved`,`buy_rejected`, `buy_lapsed`
Refactored states to check if message is for Make Offer or Purchase order.

Screenshot 2022-06-22 at 18 46 28
Screenshot 2022-06-22 at 18 45 05

[NX-3187]: https://artsyproduct.atlassian.net/browse/NX-3187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Conversation/Components/OrderUpdate.tsx
```diff
@@ -55,16 +55,24 @@ export const OrderUpdate: React.FC<OrderUpdateProps> = ({
...
+      } Accepted`
+    } else if (state === "CANCELED" && reasonRejected) {
       color = "red100"
-      message = `Offer Declined`
-    } else if (orderUpdateState === "offer_lapsed") {
-      color = "black60"
-      message = `Offer Expired`
+      message = `${
+        orderUpdateState === "offer_rejected" ? "Offer" : "Purchase"
+      } Declined`
```
lidimayra, line 70: Is `Purchase declined` a state? I thought only offers could be rejected. I would think that the only possibility for a BN order to be canceled/declined would be through expiration (caused by lack of confirmation on the partner side)

tam-kis: We've added it to Exchange but I think we don't have the functionality yet (haven't seen the option in CMS) but I thought we might keep it if we add it in the future, or is it better to remove it until then?

lidimayra: if the scenario is unachievable, I would think it makes more sense not to add it here as it could be misleading...

lidimayra (approved):
LGTM! 🚀

src/v2/Apps/Conversation/Components/OrderUpdate.tsx
```diff
@@ -55,16 +55,21 @@ export const OrderUpdate: React.FC<OrderUpdateProps> = ({
...
+      message = `${
+        orderUpdateState === "offer_approved" ? "Offer" : "Purchase"
+      } Accepted`
     } else if (orderUpdateState === "offer_rejected") {
       color = "red100"
       message = `Offer Declined`
-    } else if (orderUpdateState === "offer_lapsed") {
-      color = "black60"
-      message = `Offer Expired`
+    } else if (state === "CANCELED" && reasonLapsed) {
```
lean8086, line 68: @tam-kis  If I understood correctly, we must add this use case to [order_status_helper](https://github.com/artsy/exchange/blob/main/lib/order_status_helper.rb), right?

PR #10375, opened 2022-06-22
feat: Manually trigger Braze IAM GRO-895

jonallured:
This PR adds a new trick to our app shell: conditionally showing Braze In App Messages! I got some advice from @dzucconi here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1655829978286369

And the approach we sorta settled on was: add a hook to the app shell and adapt the code that Segment suggests with an allow list of routes. Here's that code from the Segment docs:

https://segment.com/docs/connections/destinations/catalog/braze/#other-features

Overall the flow is:

* add a ready handler to window.analytics
* add a subscription to window.appboy with a handler
* when that handler fires, decide if the IAM should be shown

In terms of testing what we have here is I extracted a function that wraps the regex checks and that's what I'm testing. The async nature of this code resisted any further testing.

Note: this does NOT address Eigen - that's been broken off into a separate Jira ticket.

https://artsyproduct.atlassian.net/browse/GRO-895

/cc @artsy/grow-devs @kathytafel

### DIFF
--- a/src/v2/Apps/Components/AppShell.tsx
+++ b/src/v2/Apps/Components/AppShell.tsx
@@ -16,6 +16,7 @@ import { useAuthIntent } from "v2/Utils/Hooks/useAuthIntent"
 import { AppToasts } from "./AppToasts"
 import { useNavBarHeight } from "v2/Components/NavBar/useNavBarHeight"
 import { useProductionEnvironmentWarning } from "v2/Utils/Hooks/useProductionEnvironmentWarning"
+import { useBrazeInAppMessage } from "v2/Utils/Hooks/useBrazeInAppMessage"
 import { useAuthValidation } from "v2/Utils/Hooks/useAuthValidation"
 import { Z } from "./constants"
 import {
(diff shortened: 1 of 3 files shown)


dzucconi:
The one thing I'd suggest here as a sanity check the Braze SDK API is: can you manually force the trigger of a Braze popup on a specific dev-route then verify that this code indeed prevents it from displaying?

jonallured:
> can you manually force the trigger of a Braze popup

I don't believe this is possible locally but it's what I was going to try on staging. 👍 Note too that there's a settings update that has to coincide with this manual triggering:

> Disable your [Automatically Send In-App Messages Destinations setting](https://segment.com/docs/connections/destinations/catalog/braze/#settings). By default, it is enabled when you enable the Braze destination.

From: https://segment.com/docs/connections/destinations/catalog/braze/#other-features

So the testing is something like:

1. merge PR
2. deploy staging
3. turn off setting in Segment for Force Staging
4. trigger IAM for my user
5. go to route that should NOT show that IAM - is it there?
6. go to route that should show that IAM - is it there?
7. deploy to production
8. turn off that setting in Segment for Force Production

/cc @isakelaris @jpoczik

PR #10383, opened 2022-06-23
feat(onboarding): onboarding set up

dzucconi:
This PR does some set up so that we can tackle the remainder in parallel.

This sets up the welcome/questions portion, which will need some polish and animation and then sets up the decisions so that we just need to flesh out the views corresponding to:

- [ ] VIEW_FOLLOW_ARTISTS
- [ ] VIEW_TOP_AUCTION_LOTS
- [ ] VIEW_CURATED_ARTWORKS
- [ ] VIEW_ARTISTS_ON_THE_RISE
- [ ] VIEW_FOLLOW_GALLERIES

https://user-images.githubusercontent.com/112297/175379345-ad09bcb8-c861-434c-b5a0-c518af1d9ed6.mov


src/v2/Apps/Onboarding/useOnboardingContext.tsx
```diff
@@ -2,48 +2,87 @@ import {
...
+
+    case "SET_ANSWER_THREE":
+      return {
+        ...state,
+        questionThree: action.payload,
+      }
+
+    default:
+      return state
+  }
```
dzucconi, line 59: Rather than have a few different states; now there's a centralized state with a reducer.

src/v2/Apps/Onboarding/useOnboardingContext.tsx
```diff
@@ -2,48 +2,87 @@ import {
...
 }
 
+/**
+ * Basis is just State stored in a ref.
+ * It is used to make decisions in the workflow.
+ */
+export type Basis = State
```
dzucconi, line 66: Basis just becomes a simple copy of State; just stuck in a ref so that the workflow engine can access it.

src/v2/Utils/WorkflowEngine.tsx
```diff
@@ -73,4 +75,12 @@ export class WorkflowEngine {
...
     return this.workflow.length - 1
   }
+
+  reset() {
+    this.moves = 0
+    this.index = 0
+    this.workflow = this.__workflow__
+
+    return this.current()
+  }
```
dzucconi, line 85: Added this so that we could reset the workflow which isn't really useful for users but is useful for us when working on/testing the thing.

PR #10395, opened 2022-06-24
feat(onboarding): adds staggered fade transition between views

dzucconi:
The goal with this PR was to implement a method for doing the staggered fade transitions without having to muddy up the components with complex and repetitive styling.

The solution I arrived at with cda564a26bc3013d67e094d6314480eb58b5653f is a very simple hook that allows one to `register` elements in a specific order. Then returns a function to `transition` them in that order. We could make this hook much more generic to support different types of transitions; like transforms — not necessary at this time.

https://user-images.githubusercontent.com/112297/175652649-823d2968-10e9-454f-a7e1-5e29006377e5.mov

### DIFF
--- a/src/v2/Apps/Onboarding/Hooks/useOnboardingFadeTransition.tsx
+++ b/src/v2/Apps/Onboarding/Hooks/useOnboardingFadeTransition.tsx
@@ -1,0 +1,38 @@
+import { useEffect } from "react"
+import { useFadeTransition } from "v2/Utils/Hooks/useFadeTransition"
+import { wait } from "v2/Utils/wait"
+
+interface UseOnboardingFadeTransition {
+  next(): void
+}
+
+export const useOnboardingFadeTransition = ({
+  next,
+}: UseOnboardingFadeTransition) => {
+  const { register, transition, mode, status } = useFadeTransition({
+    initialStatus: "Out",
+  })
+
+  useEffect(() => {
+    const init = async () => {
+      await wait(100)
+      transition("In")
+    }
+
+    init()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  const handleNext = async () => {
+    await transition("Out")
+    await wait(100)
+    next()
+  }
+
+  return {
+    register,
+    handleNext,
+    loading: status === "In" && mode === "Transitioning",
+  }
+}
--- a/src/v2/Utils/Hooks/__tests__/useFadeTransition.jest.tsx
+++ b/src/v2/Utils/Hooks/__tests__/useFadeTransition.jest.tsx
@@ -1,0 +1,63 @@
+import { render } from "@testing-library/react"
+import { renderHook } from "@testing-library/react-hooks"
+import { useFadeTransition } from "../useFadeTransition"
+
+jest.mock("../../wait", () => ({
+  wait: () => Promise.resolve(),
+}))
+
+describe("useFadeTransition", () => {
+  it("sets some defaults", () => {
+    const { result } = renderHook(() => useFadeTransition())
+
+    expect(result.current.mode).toEqual("Resting")
+    expect(result.current.status).toEqual("In")
+  })
+
+  it("registers elements", () => {
+    const { result } = renderHook(() => useFadeTransition())
+
+    const Example = () => {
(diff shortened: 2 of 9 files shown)


src/v2/Utils/Hooks/useFadeTransition.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }, [status])
+
+  const refs = useRef<RefObject<HTMLElement>[]>([])
+
+  /**
+   * @param index The index of the element to register.
+   * Transitions will execute in the order specified by the indexes.
+   * @returns A ref to the element.
+   */
+  const register = useCallback((index: number) => {
```
dzucconi, line 41: `register` very simply creates and returns a ref; then keeps track of that ref in the index you specify when registering it.

src/v2/Utils/Hooks/useFadeTransition.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const ref = createRef<HTMLElement>()
+    refs.current[index] = ref
+    return ref as any // Cast as `any` to get around outdated styled-components typing
+  }, [])
+
+  /**
+   * @returns A promise that resolves when the transition is complete.
+   */
+  const transition = useCallback(
```
dzucconi, line 50: This just reduces over a promise that's a 1/3 as long as the opacity duration, so that the fades overlap a bit. Either fading it in or out, depending on the current state.

src/v2/Apps/Onboarding/Hooks/useOnboardingFadeTransition.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const useOnboardingFadeTransition = ({
+  next,
+}: UseOnboardingFadeTransition) => {
+  const { register, transition, mode, status } = useFadeTransition({
+    initialStatus: "Out",
+  })
+
+  useEffect(() => {
```
dzucconi, line 16: So on mount we transition in.

src/v2/Apps/Onboarding/Hooks/useOnboardingFadeTransition.tsx
```diff
@@ -0,0 +1,38 @@
...
+    init()
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
+  const handleNext = async () => {
+    await transition("Out")
+    await wait(100)
+    next()
+  }
```
dzucconi, line 31: And on `next` we transition out.

src/v2/Utils/Hooks/__tests__/useFadeTransition.jest.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <div ref={result.current.register(1)}>two</div>
+        </div>
+      )
+    }
+
+    render(<Example />)
+
+    expect(
+      result.current.refs.current.map(ref => ref.current!.innerHTML)
+    ).toEqual(["one", "two", "three"])
```
jonallured, line 34: So I think this demonstrates that it doesn't matter which order you call `register` it'll respect the position passed in, right?

src/v2/Utils/Hooks/__tests__/useFadeTransition.jest.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <div ref={result.current.register(0)}>one</div>
+          <div ref={result.current.register(2)}>three</div>
+          <div ref={result.current.register(1)}>two</div>
+        </div>
+      )
+    }
+
+    render(<Example />)
+
+    await result.current.transition("Out")
```
jonallured, line 52: This changes the order, right? I was expecting to see an assertion like this:

```
    expect(
      result.current.refs.current.map(ref => ref.current!.innerHTML)
    ).toEqual(["two", "three", "one"])
```

Or maybe I'm not understanding?

dzucconi: I guess I could have written the spec differently. You'd expect them to be in order because I'm calling them with the correct indexes. I'm executing the calls out of order: `register(0)`, `register(2)`, `register(1)`.

PR #10398, opened 2022-06-27
fix(NX-3190): fix bug on CBN flow for edition sets

lidimayra:

This PR solves [NX-3190]

### Description

This fix is needed because otherwise Conversational Buy Now Flow for artworks containing more than 1 edition set would lead to "Make Offer" flow instead of "Buy Now".

The solution here consists of passing the param between different components so we can control if an order is an offer or purchase. We understand this is not the ideal solution, this is a quick fix so we can release the feature tomorrow. We'll work separately on a proper refactor to avoid the prop drilling ([NX-3208]).

[NX-3190]: https://artsyproduct.atlassian.net/browse/NX-3190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NX-3208]: https://artsyproduct.atlassian.net/browse/NX-3208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Conversation/Components/Conversation.tsx
+++ b/src/v2/Apps/Conversation/Components/Conversation.tsx
@@ -51,7 +51,6 @@ const Loading: React.FC = () => (
 
 const Conversation: React.FC<ConversationProps> = props => {
   const { conversation, relay, showDetails, setShowDetails } = props
-
   const liveArtwork = conversation?.items?.[0]?.liveArtwork
   const artwork = liveArtwork?.__typename === "Artwork" ? liveArtwork : null
   const isCBNEnabled = useFeatureFlag("conversational-buy-now")
@@ -63,6 +62,8 @@ const Conversation: React.FC<ConversationProps> = props => {
   const [showConfirmArtworkModal, setShowConfirmArtworkModal] = useState<
     boolean
   >(false)
+
+  const [createsOfferOrder, setCreatesOfferOrder] = useState<boolean>(true)
 
   const inquiryItemBox = compact(conversation.items).map((i, idx) => {
     const isValidType =
@@ -259,7 +260,10 @@ const Conversation: React.FC<ConversationProps> = props => {
           conversation={conversation}
           refetch={props.refetch}
           environment={relay.environment}
-          openInquiryModal={() => setShowConfirmArtworkModal(true)}
+          openInquiryModal={({ createsOfferOrder }) => {
+            setShowConfirmArtworkModal(true)
+            setCreatesOfferOrder(createsOfferOrder)
+          }}
           openOrderModal={() => setShowOrderModal(true)}
         />
       </NoScrollFlex>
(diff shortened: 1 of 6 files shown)


src/v2/Apps/Conversation/Components/ConversationCTA.tsx
```diff
@@ -39,6 +40,7 @@ export const ConversationCTA: React.FC<ConversationCTAProps> = ({
   conversation,
   openInquiryModal,
   openOrderModal,
+  createsOfferOrder,
```
lean8086, line 43: I see ConversationCTA doesn't need to receive this property, so if we remove it from line 24, then we can remove all the references to it.

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -259,8 +260,12 @@ const Conversation: React.FC<ConversationProps> = props => {
           conversation={conversation}
           refetch={props.refetch}
           environment={relay.environment}
-          openInquiryModal={() => setShowConfirmArtworkModal(true)}
+          openInquiryModal={({ createsOfferOrder }) => {
+            setShowConfirmArtworkModal(true)
+            setCreateOfferOrder(createsOfferOrder)
+          }}
           openOrderModal={() => setShowOrderModal(true)}
+          createsOfferOrder={createsOfferOrder}
```
lean8086, line 268: We don't need to send this prop here

src/v2/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -64,6 +63,8 @@ const Conversation: React.FC<ConversationProps> = props => {
     boolean
   >(false)
 
+  const [createsOfferOrder, setCreateOfferOrder] = useState<boolean>(true)
```
lean8086, line 66: Just a minor typo on the set:
```suggestion
  const [createsOfferOrder, setCreatesOfferOrder] = useState<boolean>(true)
```

src/v2/Apps/Conversation/Components/Reply.tsx
```diff
@@ -58,7 +58,8 @@ interface ReplyProps {
   environment: Environment
   onScroll: () => void
   refetch: RelayRefetchProp["refetch"]
-  openInquiryModal: () => void
+  createsOfferOrder: boolean
```
lean8086, line 61: Same case as in ConversationCTA. I don't think we need `createsOfferOrder` here (`openInquiryModal` is perfect, though)

lean8086 (approved):
LGTM 🚢  Thanks for the swift fix!

lidimayra:
Thanks a lot for the pairing session and for the careful review, @lean8086 💜 

I just pushed the changes removing the unneeded references and the typofix!

PR #10400, opened 2022-06-27
fix(3187): display Offer Declined message for all rejected reasons

tam-kis:

This PR solves [[NX-3187](https://artsyproduct.atlassian.net/browse/NX-3187)]

### Description

Displays `Offer Declined` message for all rejected reasons.

Screenshot 2022-06-27 at 15 56 42


lidimayra (approved):
Great work on the fast identification of the issue and the fix submission!! 👏

PR #10401, opened 2022-06-27
feat(onboarding): animates progress bar

dzucconi:
Closes [GRO-1093](https://artsyproduct.atlassian.net/browse/GRO-1093)

A little tricky: we need to either relocate the progress bar to the parent `OnboardingSteps` so that it transitions smoothly — this is non-ideal, structurally. Or: we can transition to the next state _before_ the step renders, as the old one is transitioning out. This works but is slightly weird since there's a spare render that happens inside the step that picks up the previous progress; since this is a computed value coming from the workflow engine. A bit of a hacky solution then: just debounce the progress value so that we loose that spare render.


dzucconi:
https://user-images.githubusercontent.com/112297/175994717-097e02d8-a7d4-4257-8ee8-fdbffc8dfad0.mov

PR #10406, opened 2022-06-27
chore(RelatedCategories): move to bottom of page

laurabeth:

This PR solves [GRO-1067]

### Description

- Refresh `ArtistGenes` component
- Add `QueryRenderer` to lazy load component data
- Refactor tests to use RTL
- Fix unrelated TS issue to enable push

[GRO-1067]: https://artsyproduct.atlassian.net/browse/GRO-1067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedGeneCategories.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  if (categories.length === 0) return null
+
+  return (
+    <>
+      <Text variant="lg-display">Related Categories</Text>
+
+      <Spacer mt={4} />
+
+      {categories.map(category => {
```
dzucconi, line 28: Since the pills have an `mb={1}` that means that there'll be a `1` external margin. We can wrap `categories` with `<Box mb={-1}>`  which will offset that last row of margins while still allowing for the internal margins when they break lines.

src/v2/Apps/Artist/Routes/Overview/Components/ArtistRelatedGeneCategories.tsx
```diff
@@ -0,0 +1,96 @@
...
+        return (
+          <Pill mr={1} mb={1}>
```
dzucconi, line 30: I'd cast the `Pill` as a link with `as={RouterLink}` since `Pill` is by default a `button`. You'll therefore have two tabstops, one which doesn't do anything. You'll have to `@ts-ignore` the `as` related stuff (which is fine).

laurabeth: fixed

dzucconi (approved):

PR #10413, opened 2022-06-28
fix: Hide Contact Gallery on auction lot

jo-rs:

### Description

After the release of Conversational Buy Now, the new Contact Gallery button was displayed on closed auction lots. This PR makes it so that it doesn't appear on **any** auction lot.

Before:
Screenshot 2022-06-28 at 15 28 10

After:
Screenshot 2022-06-28 at 15 28 58

cc @artsy/negotiate-devs


jo-rs:
Did Peril squash before green? 🤔

chr-tatu:
> Did Peril squash before green? 🤔

@jo-rs: I was in the backend Refinathon group from yesterday and it seems that there is a bug on Peril. Sometimes it squashes before everything is green, if a couple of jobs are green and the other ones are not started anymore. There is [this refinathon idea](https://www.notion.so/artsy/Github-replacement-for-merge-on-green-34bf7c9a06294e0989e34c21b5983ad4) to use GitHub instead of Peril or fix peril.

PR #10417, opened 2022-06-28
chore: deletes unused code; dependencies

dzucconi:
Going through `package.json`, checking if a dependency is used. If not delete it. If it's used in legacy code, see if I can delete that. If it's used in a single place, see if I can refactor it away.

Working my way through some of the low hanging fruit. More to come.


dzucconi:
As soon as we can delete those 3 one-off editorials, _so much code_ is going away.

PR #10418, opened 2022-06-28
chore: deletes unused code; dependencies

dzucconi:
More dependency related code deletion.

More to come.


dzucconi:
redux is gone.

PR #10425, opened 2022-06-29
chore: removes rc-slider dependency

dzucconi:
This PR removes the snowflake `Range` in favor of Palette's `Range` in order to remove the rc-slider dependency. Everything should work the same. The design is slightly different with foregoing the cinched in labels. But I think we reached consensus in the #design-system chat.


src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -94,11 +92,12 @@ export const PriceRangeFilterNew: FC<PriceRangeFilterNewProps> = ({
 
   return (
     <FilterExpandable label={label} expanded={hasSelection || expanded}>
-      <Flex alignItems="flex-end" mt={2}>
+      <Flex>
```
dzucconi, line 95: Top margin removed to be consistent with other `Expandable`s

src/v2/Components/ArtworkFilter/ArtworkFilters/PriceRangeFilterNew.tsx
```diff
@@ -128,39 +128,35 @@ export const PriceRangeFilterNew: FC<PriceRangeFilterNewProps> = ({
...
-      <Box mt={2} mx={`${RANGE_DOT_SIZE / 2}px`}>
-        {bars.length > 0 && !isAllBarsEmpty ? (
-          <Histogram
-            bars={bars}
-            selectedRange={[sliderRange[0], sliderRange[1]]}
-            data-testid="PriceFilterHistogram"
-          />
-        ) : null}
-
-        <Spacer pb={2} />
```
dzucconi, line 140: Never use padding on `Spacer` components since they wouldn't collapse. (Making a less error prone `Spacer` is on my TODOs this week).

src/v2/Components/ArtworkFilter/ArtworkFilters/SizeFilter.tsx
```diff
@@ -291,7 +291,7 @@ export const SizeFilter: React.FC<SizeFilterProps> = ({ expanded }) => {
         textAlign="left"
         onClick={() => setShowCustom(prevShowCustom => !prevShowCustom)}
       >
-        <Text variant="small">{showCustom ? "Hide" : "Show"} custom size</Text>
+        <Text variant="xs">{showCustom ? "Hide" : "Show"} custom size</Text>
```
dzucconi, line 294: Consistent with other text in FIlters

dimatretyak:
@dzucconi sorry for the late response. Thanks for the changes!

PR #10435, opened 2022-07-01
[Refinathon]: AlicanAkyuz/Fix-TS-errors-from-strict-null-check-migration-in-v2Apps

AlicanAkyuz:

This PR solves: https://www.notion.so/artsy/92189f632add4fc3abf177c8ead3279d?v=a888de39a73c4f0ea0af683953932d7f&p=b56a6ca90a2745daabcffd6ef00f1937

### Description
This PR focuses on Artist and Artwork apps to remove comments remaining from strict null check migration.

### DIFF
--- a/src/v2/Apps/Artist/Routes/Consign/Components/ArtistConsignPageViews.tsx
+++ b/src/v2/Apps/Artist/Routes/Consign/Components/ArtistConsignPageViews.tsx
@@ -11,16 +11,12 @@ interface ArtistConsignPageViewsProps {
 
 export const ArtistConsignPageViews: React.FC<ArtistConsignPageViewsProps> = props => {
   const {
-    artist: {
-      name,
-      targetSupply: {
-        // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-        microfunnel: {
-          metadata: { roundedViews, roundedUniqueVisitors },
-        },
-      },
-    },
+    artist: { name, targetSupply },
   } = props
+
+  const roundedViews = targetSupply?.microfunnel?.metadata?.roundedViews || "0"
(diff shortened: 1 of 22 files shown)


src/v2/Apps/Artist/Routes/Consign/Components/ArtistConsignPageViews.tsx
```diff
@@ -11,17 +11,13 @@ interface ArtistConsignPageViewsProps {
...
-        },
-      },
-    },
+    artist: { name, targetSupply },
   } = props
 
+  const roundedViews = targetSupply?.microfunnel?.metadata?.roundedViews || "0"
+  const roundedUniqueVisitors =
+    targetSupply?.microfunnel?.metadata?.roundedUniqueVisitors || "0"
+
```
AlicanAkyuz, line 20: is it ok to set those to `"0"` in case they are `null`? 

I went for this approach because below, we are rendering these values in a string, for example "...works have received more than {roundedViews} views on Artsy this month"

dzucconi: Looks correct to me.

dzucconi: But the fact that a non-nullable number was typed as a nullable string in Metaphysics is super depressing...

AlicanAkyuz: yeah the more I work on this the more I realise that there's work to be done in MP in order to get rid of those `// @ts-expect-error`s in Force

dzucconi (approved):

PR #10443, opened 2022-07-04
[feat][TX-439]: Use SetPaymentByStripeIntent and SetPayment mutations for respective purposes

AlicanAkyuz:

This PR solves [TX-439]

### Description
We divided setPayment mutation in two: `SetPaymentByStripeIntent` and `setPayment`. We need to use these new mutations instead of the current setup where we only use setPayment mutation for all purposes.

This PR uses the two mutations for their regarding purposes instead of the single setPayment mutation. The division is: 

- `setPaymentByStripeIntent` is the entry point for setting up payment methods with the new Stripe intent API
- `setPayment` is now used for setting payment with an existing bank account or wire transfer

#### New account with stripe intent Id
https://user-images.githubusercontent.com/42584148/177154779-9645beca-ca5f-4441-852b-58d2d1ac5139.mov

#### Existing account and wire transfer
https://user-images.githubusercontent.com/42584148/177154927-9b2d09d0-f134-4a31-a7ac-d7d7ef1d2005.mov

[TX-439]: https://artsyproduct.atlassian.net/browse/TX-439?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AlicanAkyuz:
Please do not merge before https://github.com/artsy/metaphysics/pull/4207

PR #10452, opened 2022-07-05
fix(NX-3189): refactor banner so the state doesn't affect Edition set modal

tam-kis:
The type of this PR is: FIX

This PR solves [NX-3189](https://artsyproduct.atlassian.net/browse/NX-3189)

### Description

Refactored Banner as a separate functional component in order to avoid it affecting the Edition Set modal.

https://user-images.githubusercontent.com/17580625/177367870-5c4a1180-4948-4d09-aaa1-58fbf9633f6d.mov


lidimayra (approved):
Way to go!! 👏

PR #10463, opened 2022-07-06
refactor(follow): modernizes artist follow button

dzucconi:
Picking up on https://github.com/artsy/force/pull/10434 — modernizes the artist follow button. Still TODO is the suggestions popover.


dzucconi:
Will follow this PR up with another PR which enables `onFollow` callbacks on all follow buttons.

PR #10466, opened 2022-07-06
chore: removing feature shows limit

pam-:
This PR solves [GRO-1034]

### Description

Previously the FeaturedShows component used to be a grid and so was limited to 6. It’s now a rail so the limit can be removed.

[GRO-1034]: https://artsyproduct.atlassian.net/browse/GRO-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi (approved):
Looks good but we can remove 'take' completely

dzucconi (approved):

PR #10475, opened 2022-07-07
[feat]: AlicanAkyuz/TX-348/Poll bank account balance

AlicanAkyuz:

This PR solves [TX-348]

### Description
This PR adds the feature of checking user's stripe-linked bank account balance before progressing to the review step from the payment step. 

[TX-348]: https://artsyproduct.atlassian.net/browse/TX-348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#### User has insufficient funds
https://user-images.githubusercontent.com/42584148/177786836-e4b00992-67ce-4fb4-bea6-13fcc4ec7976.mov

#### User has enough funds
https://user-images.githubusercontent.com/42584148/177786929-3d08dd5f-1570-418d-b8c5-c2ac636b2cad.mov

### DIFF
--- a/src/v2/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/v2/Apps/Order/Components/BankAccountPicker.tsx
@@ -8,10 +8,9 @@ import {
   Spacer,
   Collapse,
   Button,
-  Message,
-  Text,
 } from "@artsy/palette"
 import { BankDebitDetails } from "./BankDebitDetails"
+import { InsufficientFundsError } from "./InsufficientFundsError"
 import { BankAccountPicker_order } from "v2/__generated__/BankAccountPicker_order.graphql"
 import { useSetPayment } from "v2/Apps/Order/Components/Mutations/useSetPayment"
 import { extractNodes } from "v2/Utils/extractNodes"
@@ -133,25 +132,17 @@ export const BankAccountPicker: FC<Props> = props => {
         </RadioGroup>
       )}
       <Spacer mb={4} />
+
       <Collapse open={bankAccountSelection.type === "new"}>
-        <BankDebitProvider order={order} />
+        <BankDebitProvider
+          order={order}
+          bankAccountHasInsufficientFunds={bankAccountHasInsufficientFunds}
+        />
       </Collapse>
-      {bankAccountHasInsufficientFunds && (
-        <>
-          <Message
-            title="This bank account doesn’t have enough funds."
-            variant="error"
-          >
-            <Text variant="sm">
-              Please choose or link to another bank account or select another
-              payment method.
-            </Text>
-          </Message>
-          <Spacer mt={2} />
-        </>
-      )}
+
       {bankAccountSelection.type === "existing" && (
         <>
+          {bankAccountHasInsufficientFunds && <InsufficientFundsError />}
           <Button
             onClick={handleContinue}
             disabled={
--- a/src/v2/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/v2/Apps/Order/Components/BankAccountPicker.tsx
@@ -22,6 +22,7 @@ interface Props {
   onSetPaymentSuccess: () => void
   onSetPaymentError: (error: Error) => void
   showInsuffiencyFundsError: boolean
+  setShowInsuffiencyFundsError: (arg: boolean) => void
 }
 
 export const BankAccountPicker: FC<Props> = props => {
@@ -31,6 +32,7 @@ export const BankAccountPicker: FC<Props> = props => {
     onSetPaymentSuccess,
     onSetPaymentError,
     showInsuffiencyFundsError,
+    setShowInsuffiencyFundsError,
   } = props
 
   const bankAccountsArray = extractNodes(bankAccounts)
@@ -95,6 +97,7 @@ export const BankAccountPicker: FC<Props> = props => {
       {userHasExistingBankAccounts && (
         <RadioGroup
           onSelect={val => {
+            setShowInsuffiencyFundsError(false)
             if (val === "new") {
               setBankAccountSelection({ type: "new", id: "" })
             } else {
--- a/src/v2/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/v2/Apps/Order/Components/BankAccountPicker.tsx
@@ -8,6 +8,8 @@ import {
   Spacer,
   Collapse,
   Button,
+  Message,
+  Text,
 } from "@artsy/palette"
 import { BankDebitDetails } from "./BankDebitDetails"
 import { BankAccountPicker_order } from "v2/__generated__/BankAccountPicker_order.graphql"
@@ -19,6 +21,7 @@ interface Props {
   me: BankAccountPicker_me
   onSetPaymentSuccess: () => void
   onSetPaymentError: (error: Error) => void
+  showInsuffiencyFundsError: boolean
 }
 
 export const BankAccountPicker: FC<Props> = props => {
@@ -27,6 +30,7 @@ export const BankAccountPicker: FC<Props> = props => {
     order,
     onSetPaymentSuccess,
     onSetPaymentError,
+    showInsuffiencyFundsError,
   } = props
 
   const bankAccountsArray = extractNodes(bankAccounts)
@@ -129,11 +133,25 @@ export const BankAccountPicker: FC<Props> = props => {
       <Collapse open={bankAccountSelection.type === "new"}>
         <BankDebitProvider order={order} />
       </Collapse>
+      {showInsuffiencyFundsError && (
+        <>
+          <Message
+            title="This bank account doesn’t have enough funds."
+            variant="error"
+          >
+            <Text variant="sm">
+              Please choose or link to another bank account or select another
+              payment method.
+            </Text>
+          </Message>
+          <Spacer mt={2} />
+        </>
+      )}
       {bankAccountSelection.type === "existing" && (
         <>
           <Button
             onClick={handleContinue}
-            disabled={!bankAccountSelection.type}
+            disabled={!bankAccountSelection.type || showInsuffiencyFundsError}
             variant="primaryBlack"
             width={["100%", "50%"]}
           >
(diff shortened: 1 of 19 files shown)


src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,140 @@
...
+  setupIntentId,
+  commerceBankAccountBalance,
+  onBalanceCheckComplete,
+  buyerTotalCents,
+  currencyCode,
+}) => {
+  const [shouldStopPolling, setShouldStopPolling] = useState(false)
+  const timeoutID = setTimeout(() => setShouldStopPolling(true), 5000)
+
+  usePoll({
```
starsirius, line 36: 🤯

src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,140 @@
...
+    commerceBankAccountBalance?.balanceCents >= buyerTotalCents &&
+    currencyCode === commerceBankAccountBalance?.currencyCode
+
+  // we receive the balance from Stripe and it's insufficient to buy the artwork
+  const userHasInsufficientBalance =
+    commerceBankAccountBalance?.balanceCents &&
+    commerceBankAccountBalance?.balanceCents < buyerTotalCents &&
+    currencyCode === commerceBankAccountBalance?.currencyCode
+
+  // we didn't receive the balance from Stripe and polling time expired
```
starsirius, line 57: Very clean description/comments.

src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,140 @@
...
+    intervalTime: 1300,
+    key: setupIntentId,
+    clearWhen: shouldStopPolling,
+  })
+
+  // we receive the balance from Stripe and it's enough to buy the artwork
+  const userHasEnoughBalance =
+    commerceBankAccountBalance?.balanceCents &&
+    commerceBankAccountBalance?.balanceCents >= buyerTotalCents &&
+    currencyCode === commerceBankAccountBalance?.currencyCode
```
starsirius, line 49: #non-blocking. I wonder if the variables can be less repetitive (and below) by just using `balanceCents` and `balanceCurrencyCode`.

src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const PollAccountBalancePlaceholder: FC = () => {
+  const SpinnerContainer = styled.div`
+    width: 100%;
+    height: 300px;
+    position: relative;
+  `
+  return (
+    <SpinnerContainer>
+      <Spinner position="relative" color="brand" />
```
starsirius, line 137: 👍  We can get input from design and iterate on the placeholder loading state.

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -225,13 +249,22 @@ export const PaymentRoute: FC<Props> = props => {
...
+            {isSettingPayment &&
+            !isPaymentSetupComplete &&
+            !shouldPollAccountBalance ? (
               <SetPaymentByStripeIntent
                 order={order}
                 setupIntentId={setupIntentId!}
                 onSuccess={onSetPaymentSuccess}
                 onError={onSetPaymentError}
               />
+            ) : shouldPollAccountBalance && setupIntentId ? (
```
starsirius, line 261: Given the account balance polling step heavily depends on data and banks in production and we'll be testing and learning, I suggest we implement a flag (can be directly in code or an Unleash flag) that we can turn on/off the account balance polling step easily in case we need it. So

- when account balance polling is on, the flow would be
```
PaymentContent -> SetPaymentByStripeIntent -> PollAccountBalanceQueryRenderer -> Review
```

- when account balance polling is off, the flow would be
```
PaymentContent -> SetPaymentByStripeIntent -> Review
```

starsirius (commented):
Looking great and awesome! and 💯  for the recording!! @AlicanAkyuz 

I have a suggestion to implement a flag to enable/disable account balance checking, and it seems like we can use a bit more tests in this PR. Very very exciting to see this live!

src/v2/Apps/Order/Routes/Payment/index.tsx
AlicanAkyuz: thank you for this suggestion, makes sense to me 👍

damassi:
👋 Heads up! A major change is incoming in Force, where we flatten out the `src/v2` directory into `src`. This will take place **next Thursday** and be merged in, which could impact PRs with rebase conflicts (since all of our import paths will soon change). If your PR can be merged in please do so by next Weds. (See [this thread](https://artsy.slack.com/archives/C02BC3HEJ/p1657821372559579) for more info.)

src/v2/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -27,6 +31,8 @@ export const BankAccountPicker: FC<Props> = props => {
     order,
     onSetPaymentSuccess,
     onSetPaymentError,
+    showInsuffiencyFundsError,
```
erikdstock, line 34: minor: i would change all of these variables to be about insufficient* funds.

src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const timeoutID = setTimeout(() => setShouldStopPolling(true), 5000)
+
+  usePoll({
+    callback: () => {
+      relay.refetch({ setupIntentId }, null, {}, { force: true })
+    },
+    intervalTime: 1300,
+    key: setupIntentId,
+    clearWhen: shouldStopPolling,
+  })
```
erikdstock, line 43: I almost wonder if the timeout feature could be integrated into the usePoll hook - but not for this pr

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -219,6 +226,7 @@ const getAvailablePaymentMethods = (
   if (availablePaymentMethods.includes("WIRE_TRANSFER")) {
     paymentMethods.push(
       <BorderedRadio
+        key={2}
```
erikdstock, line 229: I know it doesn't matter, but would it make sense to make these keys simply the paymentMethod?

src/v2/Apps/Order/Components/BankAccountPicker.tsx
erikdstock: After reading it a bit more i would name the variable something like `bankAccountHasInsufficientFunds` though i know that is a bit long.

src/v2/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const userHasEnoughBalance =
+    commerceBankAccountBalance?.balanceCents &&
+    commerceBankAccountBalance?.balanceCents >= buyerTotalCents &&
+    orderCurrencyCode === commerceBankAccountBalance?.currencyCode
+
+  // we receive the balance from Stripe and it's insufficient to buy the artwork
+  const userHasInsufficientBalance =
+    commerceBankAccountBalance?.balanceCents &&
+    commerceBankAccountBalance?.balanceCents < buyerTotalCents &&
+    orderCurrencyCode === commerceBankAccountBalance?.currencyCode
```
erikdstock, line 54: It's not clear to me why we are defining these two variables that seem almost identical - or how they differ from the useState `showInsufficiencyFundsError` variable above. It seems these are both either `undefined` or boolean opposites - could L51 be written as 
`const userHasInsufficientFunds = typeof userHasEnoughBalance === 'undefined' ? 'undefined' : !userHasEnoughBalance`? I'm guessing there is an even cleaner way but trying to check my understanding here.

erikdstock (commented):
Looks good, some questions around the complicated polling logic but only a few minor comments.

starsirius (approved):
Very excited to give this a try!! 💯

src/v2/Apps/Order/Components/BankAccountPicker.tsx
AlicanAkyuz: good catch! it's a typo, my intention was `showInsuffientFundsError`. I'll go with your suggestion as it's more descriptive 👍

src/v2/Apps/Order/Routes/Payment/PaymentContent.tsx
AlicanAkyuz: yes it's actually better to use strings as keys!

src/v2/Apps/Order/Components/PollAccountBalance.tsx
AlicanAkyuz: it'd be interesting to try that.

src/v2/Apps/Order/Components/PollAccountBalance.tsx
AlicanAkyuz: thanks both for the suggestions 🙏 

@erikdstock they differ from `showInsufficiencyFundsError` in that they are local variables in this component and are not accessed by any other component. they are used to determine when and with which param to fire `onBalanceCheckComplete`. I would also not rely on types here, I think it'd be even better to cast these variables as booleans specifically. 

@starsirius  the thing is that `commerceBankAccountBalance` can be null too. destructuring `balanceCents` and `currencyCode` from it would cause an error if we do it before checking `commerceBankAccountBalance`'s value (we need to check it since we'd still want to check if polling should stop already when it is null). I'll try to refactor this logic in light of that. let me know if you have any other idea!

AlicanAkyuz:
after simplfying the logic in polling component: 

https://user-images.githubusercontent.com/42584148/179787422-7ac8ab91-afe6-436c-b917-44da01b50fee.mov

starsirius:
Thank you documenting with recordings! 🙇 

One non-blocking thing I noticed is that the error message rendered at the bottom of the page is not very obvious after redirection. I wonder if we can move it to the top of the page. I'll bring it up in the channel.

PR #10476, opened 2022-07-07
fix(followbutton): fixes props spread

dzucconi:
I guess the old button had a click that propagated to the button? I'm actually not sure how this was previously even working. But: relocated the data-test attr onto the actual button; but then also had to get rid of this weird isolated `buttonProps` thing so they merged correctly. Better now + should fix Integrity specs.


src/v2/Components/FollowButton/FollowArtistButton.tsx
```diff
@@ -114,37 +114,36 @@ const FollowArtistButton: React.FC<FollowArtistButtonProps> = ({
...
-              }}
-              buttonProps={rest}
-            />
-          )
-        }}
-      </Popover>
-    </Box>
+        return (
+          <FollowButton
+            data-test="followArtistButton"
```
anandaroop, line 135: 👍🏽

PR #10482, opened 2022-07-07
[wip] New for you page supports url params

jonallured:
This PR pairs with https://github.com/artsy/vortex/pull/498 to show how we could test algorithmic changes on the new-for-you page. What I did was update that page to pass along the values of 3 url params:

* first (defaults to 20)
* includeBackfill (defaults to true)
* version (defaults to null)

With these three url params we can cause the new-for-you page to show us different things during the prototyping phase of algo work. Then we could also use the `version` param to force the page to show a particular version to a user. This could line up nicely with an email test to verify our budget theory but doesn't have to.

Here's a couple examples:

### http://localhost:5000/new-for-you?first=1

Only show the first rec.

Screen Shot 2022-07-07 at 3 46 32 PM

### http://localhost:5000/new-for-you?includeBackfill=false

Don't show me backfill.

Screen Shot 2022-07-07 at 3 46 34 PM

### http://localhost:5000/new-for-you?includeBackfill=false&version=A

I want to see version A recs and no backfill.

Screen Shot 2022-07-07 at 3 46 36 PM

### DIFF
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -9781,6 +9781,12 @@ type MarketingCollection {
 
   # Returns collections by the same artist if more than 4 exist, otherwise returns collections from the same category
   relatedCollections(size: Int = 10): [MarketingCollection!]!
+
+  # Show featured artists section in header. Defaults to true
+  showFeaturedArtists: Boolean!
+
+  # Show artworks rail in header. Defaults to true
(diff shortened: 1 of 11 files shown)


jonallured:
Oops, opened from a fork rather than upstream - will reopen!

PR #10489, opened 2022-07-08
[feat]: AlicanAkyuz/TX-241/Buyer can save bank account during checkout

AlicanAkyuz:

This PR solves [TX-241]

### Description
This PR provides the ability to save/not save a stripe-linked bank account during the checkout flow via a checkbox and explanation tooltip. 

#### Video
https://user-images.githubusercontent.com/42584148/178008582-b53c30d5-874e-4a0d-9acb-a721f0967c73.mov

[TX-241]: https://artsyproduct.atlassian.net/browse/TX-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Apps/Order/Components/SetPaymentByStripeIntent.tsx
+++ b/src/v2/Apps/Order/Components/SetPaymentByStripeIntent.tsx
@@ -7,12 +7,13 @@ import styled from "styled-components"
 interface Props {
   order: Payment_order
   setupIntentId: string
+  saveAccount: string
   onSuccess: () => void
   onError: (object) => void
 }
 
 export const SetPaymentByStripeIntent: FC<Props> = props => {
-  const { setupIntentId } = props
+  const { setupIntentId, saveAccount } = props
   const {
     submitMutation: setPaymentByStripeIntentMutation,
   } = useSetPaymentByStripeIntent()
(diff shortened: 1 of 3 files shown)


src/v2/Apps/Order/Components/SetPaymentByStripeIntent.tsx
```diff
@@ -26,6 +27,7 @@ export const SetPaymentByStripeIntent: FC<Props> = props => {
               input: {
                 id: props.order.internalID,
                 setupIntentId: setupIntentId,
+                oneTimeUse: saveAccount === "true" ? false : true,
```
starsirius, line 30: #minor: Guess this is what prettier formatted. I was thinking it may be more concise to just say
```typescript
oneTimeUse: saveAccount !== "true"
```

src/v2/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -202,6 +202,7 @@ export const PaymentRoute: FC<Props> = props => {
     })
   }
 
+  const saveAccount = getClientParam("saveAccount")
```
starsirius, line 205: #non-blocking: noticed inconsistency of the casing of the query params, i.e. `saveAccount` vs. `setup_intent` below. Not sure what's the recommended casing but would be good to be consistent I think.

starsirius (approved):
Woohoo! 💯

PR #10491, opened 2022-07-08
chore: migrates rss to v2

dzucconi:
Ports `/rss/partner-updates` to TypeScript. Drops `/rss/news` since we no longer publish news.


src/v2/Apps/RSS/rssServerApp.tsx
```diff
@@ -0,0 +1,35 @@
...
+rssServerApp.get("/rss/partner-updates", async (req, res) => {
+  const relayEnvironment = createRelaySSREnvironment({
+    userAgent: req.header("User-Agent"),
+  })
+
+  const { articlesConnection } = await fetchQuery<RssPartnerUpdatesQuery>(
+    relayEnvironment,
+    RSS_PARTNER_UPDATES_QUERY,
+    { channelId: GALLERY_PARTNER_UPDATES_CHANNEL }
+  )
```
dzucconi, line 24: I believe this is the first instance of us just using Relay to fetch server-side without React? So no need for that `metaphysics2` function.

dzucconi: Could also just be a normal `fetch`, I suppose.

src/v2/Apps/RSS/templates/article.ejs
```diff
@@ -0,0 +1,60 @@
+<item>
+  <guid isPermaLink="false"><%= appUrl %><%= article.href %></guid>
+  <title><%= article.thumbnailTitle %></title>
+  <link><%= appUrl %><%= article.href %></link>
+  <author><%= article.byline %></author>
```
dzucconi, line 5: This is technically Invalid FWIW. This is the old behavior but according to spec `author` should be the author's email address. I'm assuming other RSS feeds abuse this field too.

src/v2/Apps/RSS/rssServerApp.tsx
damassi: generated relay types are always good

PR #10499, opened 2022-07-11
chore(v2): ports remaining legacy apps to redirects server app

dzucconi:
_OK_

This PR migrates the last 5 legacy desktop apps. All of them were concerned with the last-resort top-level path redirects and have been consolidated into a single v2 application.

This PR doesn't delete any of the legacy code yet (only the setup that was concerned with the desktop index file); that 🔥 will come in another PR.


src/v2/Apps/Redirects/__tests__/redirectsServerRoutes.jest.ts
```diff
@@ -0,0 +1,227 @@
...
+        },
+      }
+
+      handleFair(req, res, next)
+
+      expect(res.redirect).not.toHaveBeenCalled()
+      expect(next).toHaveBeenCalled()
+    })
+  })
+})
```
dzucconi, line 227: These were the existing tests; just changed to use GraphQL instead of Backbone.

src/v2/Apps/Redirects/redirectsServerRoutes.tsx
```diff
@@ -0,0 +1,219 @@
...
+      "/:id/browse/filter",
+      "/:id/browse/show/:partnerId",
+      "/:id/feed",
+      "/:id/live",
+    ],
+    handleFair
+  )
+
+  // Fair Organziers
+  .get("/:id", handleFairOrganizer)
```
dzucconi, line 217: All this code is basically the same, just updated to use GraphQL.

src/desktop/index.ts
```diff
@@ -1,36 +0,0 @@
-import express from "express"
```
damassi, line 1: ☮️

src/v2/Apps/Redirects/redirectsServerRoutes.tsx
```diff
@@ -0,0 +1,219 @@
...
+  if (res.locals.profile?.owner.__typename !== "Partner") return next()
+
+  const partnerSlug = res.locals.profile.owner.slug
+
+  if (!partnerSlug || !req.path) return next()
+
+  const href: string = `/partner${req.path.replace(req.params.id, partnerSlug)}`
+
+  // /:id/overview => /partner/:id
+  // /:id/about => /partner/:id
```
damassi, line 82: 👍 ty

PR #10502, opened 2022-07-12
chore(v2): ports artist auth modal; deletes apps dir

dzucconi:
Deletes the apps directory. Very close to being able to delete all the models/collections — can use a second set of eyes debugging the `CurrentUser` + passport unwinding.


damassi (approved):
Looks good!

PR #10504, opened 2022-07-12
fix(onboarding): updates crops; adds name

dzucconi:
Updates the crop images on the animated columns; closes [GRO-1087](https://artsyproduct.atlassian.net/browse/GRO-1087)


TMcMeans:
was there any discussion about what would happen if the user puts in gibberish for name or leaves the last name blank? I guess that's not a concern anymore??

dzucconi:
@TMcMeans yeah: they just see whatever they input. We only have one name field so it's just whatever it is in it's entirety; there is no last name (that was kind of the problem).

PR #10508, opened 2022-07-12
chore: deletes legacy code (models, collections, components)

dzucconi:
Ultimately the goal here was to remove _all_ of the backbone related code. To do so required the one remaining place we still used a model, the passport serialization.

I _think_ this should be complete. Will run the full integrity specs against my local before converting off draft.


damassi:
🍻

zephraph:
So proud of you folks 🥲

PR #10509, opened 2022-07-13
Support URL params on New For You page

jonallured:
This PR is well described over on https://github.com/artsy/force/pull/10482 but the tl;dr is that I've updated the `/new-for-you` page so that you can pass `first`, `includeBackfill` and `version` to override defaults.

/cc @artsy/grow-devs @joeyAghion @kathytafel


jonallured:
Ok this is green so I'm going to merge!

PR #10513, opened 2022-07-13
enhancement(2fa): includes password in enable second factor mutation

ovasdi:

This PR solves [PLATFORM-4319](https://artsyproduct.atlassian.net/browse/PLATFORM-4319)

### Description

`EnableSecondFactor` mutation includes a user _password_ in the input vars.

---

This change is related to an accepted [security bounty submission](https://artsyproduct.atlassian.net/browse/PLATFORM-4316) that surfaced: _Password is validated but not required on OTP changes_.

When setting up 2fa (App or Sms) we require a user to confirm their password. The password is then sent with the initial _Create[[App](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx#L136)|[Sms](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/SmsSecondFactor/index.tsx#L137)]SecondFactor_ mutation but it is not sent with the following mutation that enables second factor, _[EnableSecondFactor](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx#L96)_.

The password is _currently optional_ in both, [MP](https://github.com/artsy/metaphysics/blob/f4bbd051032bf47caa233b7e6fb56f6cb3e73214/_schemaV2.graphql#L7471) and [Gravity](https://github.com/artsy/gravity/blob/000b3af443c8b770a0066875497efaf19dfdc2d5/app/graphql/mutations/enable_second_factor.rb#L7). We want to make this a requirement and thus the first step is to pass the password from the client.

### DIFF
--- a/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
+++ b/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
@@ -41,6 +41,7 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
     false
   )
   const [stagedSecondFactor, setStagedSecondFactor] = useState(null)
+  const [confirmedPassword, setConfirmedPassword] = useState("")
   const [isDisabling] = useState(false) // ???
   const [isCreating, setCreating] = useState(false)
 
@@ -139,6 +140,7 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
       })
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
       setStagedSecondFactor(response.createAppSecondFactor.secondFactorOrErrors)
+      setConfirmedPassword(password)
       setShowConfirmPassword(false)
       setShowSetupModal(true)
     } catch (error) {
@@ -246,6 +248,7 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
       <AppSecondFactorModal
         show={showSetupModal}
         secondFactor={stagedSecondFactor}
+        confirmedPassword={confirmedPassword}
         onComplete={onComplete}
         onClose={() => setShowSetupModal(false)}
       />
--- a/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
+++ b/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/Modal.tsx
@@ -42,10 +42,11 @@ interface AppSecondFactorModalProps {
   onComplete: () => void
   // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
   secondFactor: CreateAppSecondFactorMutationResponse["createAppSecondFactor"]["secondFactorOrErrors"]
+  confirmedPassword: string
 }
 
 export const AppSecondFactorModal: React.FC<AppSecondFactorModalProps> = props => {
-  const { secondFactor, onComplete } = props
+  const { secondFactor, confirmedPassword, onComplete } = props
   const { relayEnvironment } = useSystemContext()
 
   const [showForm, setShowForm] = useState(true)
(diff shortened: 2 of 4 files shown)


src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
```diff
@@ -41,6 +41,7 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
     false
   )
   const [stagedSecondFactor, setStagedSecondFactor] = useState(null)
+  const [confirmedPassword, setConfirmedPassword] = useState("")
```
ovasdi, line 44: Do I need to include `setConfirmedPassword("")` in the `resetState`? It appears that we do this for all _state flags_ in [AppSecondFactor](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx#L49) however I am not clear on whether this is necessary seeing as in [SmsSecondFactor](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/SmsSecondFactor/index.tsx#L50) not _all_ the flags are 'reset'.

ovasdi: Wondering, what is the reason for the reset?

joeyAghion (approved):
I don't see problems here, assuming it works and the confirmed password is carried from the prompt to the eventual mutation the same way it is for other sequences of operations.

I can't answer your question though, so might be worth getting input from others.

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
mc-jones, line 44: Nit, non-blocking. Do we want to change this a bit since technically this is a password _to be_ confirmed? So maybe something like `passwordToBeConfirmed`, or `passwordConfirmation` to more accurately describe that state of the string?

mc-jones: Based on the [note](https://github.com/artsy/force/blob/main/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/SmsSecondFactor/index.tsx#L49) from @dzucconi, looks like we shouldn't need to worry about this?

Maybe we were seeing some stale state and were testing resetting them manually for some reason? Looks like this only is being done after a `mutation` [error](https://github.com/artsy/force/blob/0fa8a5e244e1eb0856957ff63d433dd96cb4e48c/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx#L116). Not sure if you saw anything while testing a mutation failure?

mc-jones (approved):

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
ovasdi: > You may need to add it. The whole mess was getting into unrecoverable states on error, so I added that `resetState` to kind of take a sledgehammer to it.

Thanks for elaborating. For consistency, I'll add it in.

ovasdi: Renamed in [f4f9062](https://github.com/artsy/force/pull/10513/commits/f4f90621b865c57cdf1ab376d6f9a7462e5d1db6)

PR #10520, opened 2022-07-14
feat(onboarding): adds query param trigger

dzucconi:
This PR allows the onboarding modal to be triggered from the homepage with the existence of a query param: `?onboarding=true`


src/v2/Components/Onboarding/Hooks/useOnboarding.tsx
```diff
@@ -0,0 +1,29 @@
...
+      )
+    },
+  })
+
+  return {
+    isVisible,
+    showOnboarding: showDialog,
+    hideOnboarding: hideDialog,
+    onboardingComponent: dialogComponent,
+  }
```
damassi, line 28: Did not know about the `useDialog` hook 👍

PR #10525, opened 2022-07-14
chore(passport): Delete unused passport analytics code

damassi:

### Description

This PR deletes some passport analytics code that has never executed. Context around this is:
- In [this PR](https://github.com/artsy/force/pull/10516) I noticed that we had analytics stuff that wasn't executing
- [This line](https://github.com/artsy/force/pull/10516/files#diff-e5737d1f4fe31daada480392ee41e22228a6725b24417cf486d725ad8d8e97c1L7) was the culprit
- Looking in the [options file](https://github.com/artsy/force/blob/main/src/lib/passport/lib/options.js), there's no `SEGMENT_WRITE_KEY`
- And so in the aforementioned PR I fixed the issue by pointing at sharify, which in turn started to cause problems in our smoke tests due to missing ENV var

But really, lets just delete this code. It looks like some legacy stuff from old artsy/passport lib that hasn't worked in a while. No reason to keep it around, and we track login stuff elsewhere.


ovasdi (approved):
LGTM. May be worth mentioning [this slack](https://artsy.slack.com/archives/CA8SANW3W/p1657819276076089) thread for additional context.

dzucconi:
This code does get executed though?

damassi:
Thanks for the flag -- looks like this options export was getting mutated on init. See [convo here](https://artsy.slack.com/archives/C01ADJNCS5D/p1657904394813909) for more info. Reverting.

PR #10529, opened 2022-07-14
chore: include PROCESSING_APPROVAL state in queries for active orders in conversation view.

erikdstock:
The type of this PR is: bugfix
Related: [TX-467] - the same change in eigen will complete it.

### Description
This PR adds PROCCESSING_APPROVAL to every graphql order query in our conversation app that uses the `states:` argument to filter orders by state if that query is also used to find an `activeOrder`. The hope is that this will cause orders that are processing approval to still appear with their associated conversations and not disappear completely.

[TX-467]: https://artsyproduct.atlassian.net/browse/TX-467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


starsirius (approved):
🙏  Thank you for the quick investigation and fix!

damassi:
Can we update the PR title before merge 🙏

PR #10536, opened 2022-07-15
chore: migrates v2 text variants to v3

dzucconi:
Most of this is minor remapping of text variants.

I did wind up just rebuilding the search feedback form, which was pretty far gone.


dzucconi:
https://github.com/artsy/palette/pull/1194 is releasable once this is merged

PR #10541, opened 2022-07-15
feat: create backup second factor mutation includes user password

ovasdi:

This PR solves [PLATFORM-4322](https://artsyproduct.atlassian.net/browse/PLATFORM-4322)

### Description

* _setting up_ or _regenerating_ second factor codes prompts to confirm password
* `useCreateSettingsBackupSecondFactors` mutation includes a password in the input vars

---

For the same reasons as with https://github.com/artsy/force/pull/10513 this PR includes a password in the mutation when creating backup second factor codes. Additionally, this adds a _confirm password_ step when setting up and regenerating backup codes.

The password is currently optional in [MP](https://github.com/artsy/metaphysics/blob/f4bbd051032bf47caa233b7e6fb56f6cb3e73214/_schemaV2.graphql#L6693) and [Gravity](https://github.com/artsy/gravity/blob/000b3af443c8b770a0066875497efaf19dfdc2d5/app/graphql/mutations/create_backup_second_factors.rb#L4) and will be switched to required following this change.


damassi (approved):
👌

src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/SettingsEditSettingsTwoFactorBackupCodes.tsx
```diff
@@ -97,21 +104,32 @@ export const SettingsEditSettingsTwoFactorBackupCodes: FC<SettingsEditSettingsTw
...
             </Button>
           )}
         </Flex>
       </Flex>
 
+      <ConfirmPasswordModal
+        show={showConfirmPassword}
+        onConfirm={handleGenerate}
+        onCancel={() => setShowConfirmPassword(false)}
+        title="Regenerate Backup Codes"
```
ovasdi, line 129: I chose to call this "Regenerate" seeing as whenever the user (initially) sets up 2fa, backup codes are automatically generated for them. In fact, I am curious if there is ever a state when a user will see "Set up" backup codes button?

mc-jones: Nope. We conditionally [hide](https://github.com/artsy/force/blob/49e09ce13c0b6f8fd62794ebf65e85b45340c094/src/v2/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/SettingsEditSettingsTwoFactor.tsx#L33) the backup second-factor component unless the user has any form of 2fa, so they are only ever able to regenerate codes after setting up some other form of 2fa. This is by design, as we don't necessarily want users to rely on backup codes as their main form of 2fa.

PR #10549, opened 2022-07-18
[WIP](feat)show: show events on show subpage

TMcMeans:
The type of this PR is: _feat_

This PR solves [GRO-1024]

### Description

The intention of this PR is to show events for shows on their respective info subpages for the user. Currently this PR is a WIP for a couple reasons: 
- the show event data is query-able but strange in presentation and there will need to be some cleanup of the `startAt` & `endAt` times as well as whether or not to use the `eventType` or `title` to display. 

Screen Shot 2022-07-18 at 1 24 03 PM

- also we don't have enough show events to QA for so I'm hunting around for those at the moment 

- here is where show events should go (??) on the show subpage assuming there isn't any pushback on that

Screen Shot 2022-07-18 at 2 01 25 PM

[GRO-1024]: https://artsyproduct.atlassian.net/browse/GRO-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Show/Components/ShowEvents.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const ShowEventsFragmentContainer = createFragmentContainer(ShowEvents, {
+  show: graphql`
+    fragment ShowEvents_show on Show {
+      events {
+        title
+        eventType
+        description
+        startAt
+        endAt
```
damassi, line 48: @TMcMeans - note that these fields should be configurable and return any format that you'd like. See [here](https://github.com/artsy/force/blob/15693e3b173c8daa510d72014eb0bf378a6eacd7/src/v2/Apps/Home/Components/HomeCurrentFairs.tsx#L179-L180) for an example. 

Also, check out this list of date formats: https://docs.oracle.com/cd/E41183_01/DR/Date_Format_Types.html#:~:text=MM%2FDD%2FYY,(17%2F02%2F2009)

TMcMeans: oh awesome, that solves the problem of having to do any data cleanup for sure thanks!

src/v2/Apps/Show/ShowApp.tsx
```diff
@@ -27,6 +27,7 @@ import {
 } from "v2/Components/ArtworkFilter/ArtworkFilterContext"
 import { RouterLink } from "v2/System/Router/RouterLink"
 import { BackToFairBannerFragmentContainer as BackToFairBanner } from "./Components/BackToFairBanner"
+import { ShowEventsFragmentContainer as ShowEvents } from "./Components/ShowEvents"
```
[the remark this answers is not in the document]
damassi: That would be really awesome 👍 I think the parent app was setup before we formalized this rule and then we never went back and cleaned up

src/v2/Apps/Show/Components/ShowEvents.tsx
damassi: There's a few fields in metaphysics like that. Another example is [`description`](https://github.com/artsy/force/blob/main/src/v2/Apps/Artwork/Components/ArtworkDetails/ArtworkDetailsAboutTheWorkFromArtsy.tsx#L70). And here's it defined in [Metaphysics](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/gene.ts#L125). Lil helpers that will transform inputs into new fields that one can configure based on the need.

PR #10552, opened 2022-07-18
chore(i18n): Only enable debug mode when ENABLE_I18N_DEBUG=true

damassi:

### Description

Noticed that i18n debug logging was enabled in prod; this fixes that by enabling it only when 
```sh
ENABLE_I18N_DEBUG=true
```


jonallured:
tyyyyy 🤗

PR #10557, opened 2022-07-19
feat(TX-285): bank debit setup error handling

rquartararo:

This PR solves [TX-285]

### Description

This PR adds error handling to the `CreateBankDebitSetupForOrder` mutation, as well as adds an error message for the user if creating the setup fails and we are unable to render the Stripe payment element. 

Screenshot 2022-07-19 at 10 31 25 AM

[TX-285]: https://artsyproduct.atlassian.net/browse/TX-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/v2/Components/BankDebitForm/__tests__/BankDebitProvider.jest.tsx
+++ b/src/v2/Components/BankDebitForm/__tests__/BankDebitProvider.jest.tsx
@@ -1,0 +1,40 @@
+import { mount } from "enzyme"
+import { BankDebitProvider } from "../BankDebitProvider"
+import React from "react"
+import { BankDebitForm } from "../BankDebitForm"
+
+const setHookState = state =>
+  jest.fn().mockImplementation(() => [state, () => {}])
+
+const reactMock = require("react")
+
+describe("BankDebitProvider", () => {
+  const getWrapper = (props: any = {}) =>
+    mount(
+      <BankDebitProvider
+        order={{
+          internalID: "1234",
+          mode: "BUY",
+          bankAccountId: "bank-id-1",
+          " $refType": "BankAccountPicker_order",
+        }}
+      />
+    )
+
+  it("renders bank debit form", () => {
+    reactMock.useState = setHookState({
+      clientSecret: "client-secret",
+    })
+    const wrapper = getWrapper()
+    expect(wrapper.find(BankDebitForm).length).toBe(1)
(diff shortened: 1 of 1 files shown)


src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -63,10 +95,12 @@ export const BankDebitProvider: FC<Props> = ({ order }) => {
 
   return (
     <div>
-      {clientSecret && (
+      {clientSecret ? (
```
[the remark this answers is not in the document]
AlicanAkyuz: looks fine to me 👍  

assuming that we don't need the users to know the details of the error – whether it's caused by our system or stripe etc.

AlicanAkyuz (approved):

src/v2/Components/BankDebitForm/BankDebitProvider.tsx
```diff
@@ -5,31 +5,63 @@ import { getENV } from "v2/Utils/getENV"
...
+    <>
+      <Message
+        title="Bank transfer is not available at the moment"
+        variant="error"
+      >
+        <Text variant="sm">
+          Refresh the page or select another payment method.
+        </Text>
+      </Message>
+      <Spacer mt={2} />
```
damassi, line 30: @rquartararo - A good habit for working with layouts is to always handle outer margin and padding outside of the component. Since `Message` has all of our common spacer values mixed in ([via `Flex`](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Message/Message.tsx#L42) you can use the `MessageProps` type: 

```tsx
const BankSetupErrorMessage: React.FC<MessageProps> = props => {
  return (
    <Message {...props}>...</Message>
  )
}
```
and then later
```tsx
<BankSetupErrorMessage mt={2} />
```

damassi: But an even simpler way to do this (while preserving the literate coding style) would be to assign `Message` it to a variable:

```tsx
const BankSetupErrorMessage = Message
```
or skip the intermediate variable altogether and just use `<Message>` in the main body of the code.

PR #10561, opened 2022-07-19
fix: bank debit error message render

rquartararo:

### Description

While testing [TX-285](https://github.com/artsy/force/pull/10557) on staging with network throttling, I noticed that the error message appears before we get the client secret and then goes away after, which isn't ideal. I've refactored the error message to use its own hook, instead of relying on the client secret hook. 

### DIFF
--- a/src/v2/Components/BankDebitForm/__tests__/BankDebitProvider.jest.tsx
+++ b/src/v2/Components/BankDebitForm/__tests__/BankDebitProvider.jest.tsx
@@ -22,6 +22,9 @@ describe("BankDebitProvider", () => {
     )
 
   it("renders bank debit form", () => {
+    reactMock.useState = setHookState({
+      clientSecret: "client-secret",
+    })
     const wrapper = getWrapper()
     expect(wrapper.find(BankDebitForm).length).toBe(1)
   })


AlicanAkyuz (approved):
it also slipped my eye that could happen! 

your solution looks good to me. one suggestion is to update the condition to render the error component, so that there's no way for users to see both `BankDebitForm` and `BankSetupErrorMessage` at the same time (very unlikely to happen, but why not avoid it). 

`{!clientSecret && bankDebitSetupError && <BankSetupErrorMessage />}`

iskounen (approved):
LGTM. Thank you for testing this out with the network throttled!

AlicanAkyuz:
@rquartararo to solve the issue you mentioned, I think we'll need to mock the async operation (`submitMutation()` call in useEffect) in the tests. For this,

bring in the wrapper function:
```
import { CreateBankDebitSetupForOrder } from "v2/Components/BankDebitForm/Mutations/CreateBankDebitSetupForOrder"
```

tell jest you will mock it: 
```
jest.mock("v2/Components/BankDebitForm/Mutations/CreateBankDebitSetupForOrder")
```

mock it (within `describe`):
```
const mockCreateBankDebitSetupForOrder = CreateBankDebitSetupForOrder as jest.Mock
```

and finally mock its return value (which is actually what we need): 
```
mockCreateBankDebitSetupForOrder.mockImplementation(() => ({
      submitMutation: jest.fn(),
}))
```

let me know if this makes sense! happy to pair as well.

damassi:
> When adding this, it's causing an async issue with the tests as we need to wait for the mutation to finish before mocking the client secret hook. I'm struggling to figure out how to wait for the useEffect to finish in Jest

#dev-help is the place to go for questions like this before a merge, especially if it was noted by another teammate in review 🙏  Lots of Jest experts in the bunch.

damassi:
Np 👍 internal context should always be the guide in things like this.

PR #10568, opened 2022-07-20
feat(CX-2657): My Collection coming soon banner

lordkiz:

This PR solves [CX-2657]

### Description
- Adds a banner about MyCollection coming soon on web.

Screenshot 2022-07-20 at 11 33 15

[CX-2657]: https://artsyproduct.atlassian.net/browse/CX-2657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Settings/Routes/MyCollection/Components/SettingsMyCollectionArtworks.tsx
```diff
@@ -28,6 +34,10 @@ const SettingsMyCollectionArtworks: FC<SettingsMyCollectionArtworksProps> = ({
   relay,
 }) => {
   const [loading, setLoading] = useState(false)
+  const [hasDismissedMessage, setHasDismissedMessage] = useState(
+    window.localStorage.getItem("HAS_SEEN_MY_COLLECTION_MESSAGE_BANNER") ===
```
damassi, line 38: Ah, is this working? Since this relies on `window` I would expect this to break during the SSR pass. Is this component only mounted on the client via a `QueryRenderer`?

lordkiz: @damassi it works on dev. I am not sure if this component is only mounted via a QueryRenderer. I should move this logic into a hook that only runs client, like useEffect 🤔

dariakoko (dismissed):
LGTM! Thanks Kizito

How come localStorage has not been used in a like manner in force before? 😅 Was something else used, or there was just no need for that?

src/v2/Apps/Settings/Routes/MyCollection/Components/SettingsMyCollectionArtworks.tsx
damassi: Interesting -- i would investigate this and be sure. If things are executing within a query renderer it should be fine (since query renderers only execute on the client) but not that this code will break if things ever get moved outside of a query renderer into our formal server-side-rendering flow. 

If it works though, no reason to move to a hook! We should just be sure to understand why it's working.

src/v2/Apps/Settings/Routes/MyCollection/Components/SettingsMyCollectionArtworks.tsx
```diff
@@ -62,8 +76,31 @@ const SettingsMyCollectionArtworks: FC<SettingsMyCollectionArtworksProps> = ({
...
+          <Flex flexDirection="row" justifyContent="space-between">
+            <Box />
+            <Text mx={1}>
+              Access all the My Collection features on the Artsy app. Coming
+              soon also on web.
+            </Text>
+            <Clickable onClick={dismissMyCollectionMessage}>
+              <CloseIcon />
+            </Clickable>
+          </Flex>
```
dzucconi, line 101: You should avoid messing with the internals of this component so much and just add support for an `onDismiss` callback within the Palette component.

src/v2/Apps/Settings/SettingsApp.tsx
```diff
@@ -29,7 +31,7 @@ const SettingsApp: React.FC<SettingsAppProps> = ({ me, children }) => {
         Hi {me.name}
       </Text>
 
-      <RouteTabs my={4}>
+      <RouteTabs my={SETTINGS_ROUTE_TABS_MARGIN}>
```
dzucconi, line 34: 🤨

src/v2/Apps/Settings/Routes/MyCollection/Components/SettingsMyCollectionArtworks.tsx
lordkiz: @dzucconi We don't have a dismissible Message Component. And none of the Banner variants match the design.

src/v2/Apps/Settings/Routes/MyCollection/Components/SettingsMyCollectionArtworks.tsx
lordkiz: Following recent changes and refactors, I encountered the expected window not available error. Now using it in useEffect instead.

damassi: Sounds good! Lets keep an eye on UX flashing in an unstable way (server renders, then this banner suddely appears on client-side mount), and if so we can return to this and perhaps use a cookie instead. Nothing blocking here, though 👍

lordkiz:
> How come localStorage has not been used in a like manner in force before? 😅 Was something else used, or there was just no need for that?

@dariakoko This is like my 2nd or 3rd PR to Force. I am not familiar with all the parts yet. But we use localStorage for some logic related to featureFlags/experiments. I suppose we have not had a need for it beyond that.

dariakoko (approved):
LGTM!
Let's merge?

lordkiz:
@dariakoko Yes

PR #10571, opened 2022-07-20
feat(FX-4039): expand auto-selection of user's preferred measurement system for size filter on all grids

dimatretyak:
The type of this PR is: **Feature**

This PR solves [FX-4039]

### Description

[FX-4039]: https://artsyproduct.atlassian.net/browse/FX-4039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10572, opened 2022-07-20
feat(FX-4039): keep user preferences in context

dimatretyak:

This PR solves [FX-4039]

### Description
A couple of months ago, [changes](https://github.com/artsy/force/pull/9471) were added that allow to automatically apply the user's preferred metric for the size filter. For example, if the user specified `inches` as the preferred metric, then `in` toggle will be selected automatically selected in the size filter. If `centimeters`, then `cm` toggle (see attached videos [here](https://github.com/artsy/force/pull/9471) to get more context)

There is such a [Jira ticket](https://artsyproduct.atlassian.net/browse/FX-4039) on the FX board, in which the above changes should be expanded to all other grids (at the moment these changes are only available on artist artworks grid). 

[Here](https://github.com/artsy/force/pull/10571) you can find some progress for FX-4039 task. But if you take a closer look at the changes, then note that we need to add too many extra changes in order to automatically apply the user's preferred metric (for example, as in [this commit](https://github.com/artsy/force/pull/10571/commits/b263987afc7ca007fc30f2a5e22964b46b33d845), where support for artist series grid was added)

Therefore, I thought it would be a good idea to store the user's preferred settings in some single store (for example, in system context) and be able to access these settings at any time (for example, using a custom react hook).

I will be glad to listen to any suggestions/advices about this approach

[FX-4039]: https://artsyproduct.atlassian.net/browse/FX-4039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/middleware/userPreferencesMiddleware.ts
```diff
@@ -0,0 +1,35 @@
...
+export const userPreferencesMiddleware = async (
+  req: ArtsyRequest,
+  res: ArtsyResponse,
+  next: NextFunction
+) => {
+  if (!!req.user) {
+    let metric
+
+    try {
+      const endpoint = `${API_URL}/api/v1/me`
```
damassi, line 17: Can we use GraphQL for this, rather than relying on our gravity endpoint? Then we'll get proper generated types. Here's [a good example](https://github.com/artsy/force/blob/main/src/v2/Apps/RSS/rssServerApp.tsx#L19-L27) of how that can be done.

damassi: (Just a non-blocking suggestion)

dimatretyak: Thank you for the example, I will take a look at it 🙏

src/v2/System/Router/buildClientAppContext.ts
```diff
@@ -27,6 +27,7 @@ export const buildClientAppContext = (
       mediator,
       user: sd.CURRENT_USER,
       featureFlags: sd.FEATURE_FLAGS,
+      userPreferences: sd.USER_PREFERENCES,
```
dimatretyak, line 30: **question**: maybe it would be better if we store it in the `user` field? Something like this

```javascript
{
  analytics: {
    contextPageOwnerSlug: pageSlug,
    contextPageOwnerType: pageType,
  },
  mediator,
  user: {
    ...sd.CURRENT_USER,
    preferences: sd.USER_PREFERENCES,
  },
  featureFlags: sd.FEATURE_FLAGS,
  ...context,
}
```

damassi: This is a good question. One thing about the `user` object is that it maps 1:1 with the model. If we were to extend it it would almost be better if we were to make the change in Gravity so that no additional field or object merging is needed (if I'm understanding correctly)

dimatretyak: > If we were to extend it it would almost be better if we were to make the change in Gravity so that no additional field or object merging is needed

we can query additional fields via `userKeys` [here](https://github.com/artsy/force/blob/b1603f4b74c73dea950dcf4869a5294c6e8c7f0d/src/middleware.ts#L198) (see example below). But the main problem is that the `length_unit_preference` will contain the value that was relevant at the time of executing [this request](https://github.com/artsy/force/blob/b1603f4b74c73dea950dcf4869a5294c6e8c7f0d/src/lib/passport/lib/app/token_login.js#L43) and as far as I understood, the values specified in `userKeys` will not be automatically updated until the next user authorization (or are there moments that I missed?)

```javascript
app.use(
  artsyPassport({
    ...
    userKeys: [
      "collector_level",
      "default_profile_id",
      "email",
      "has_partner_access",
      "id",
      "lab_features",
      "name",
      "paddle_number",
      "phone",
      "roles",
      "type",
      "length_unit_preference", // 👈 query new field
    ],
  })
)
```

src/lib/middleware/userPreferencesMiddleware.ts
dimatretyak: Done

src/v2/Utils/fetchUserPreferences.ts
```diff
@@ -0,0 +1,18 @@
+import { Environment, fetchQuery, graphql } from "relay-runtime"
+import { fetchUserPreferencesQuery } from "v2/__generated__/fetchUserPreferencesQuery.graphql"
+
+export const fetchUserPreferences = (relayEnvironment: Environment) => {
+  return fetchQuery<fetchUserPreferencesQuery>(
```
damassi, line 5: Nice 👍

src/Utils/fetchUserPreferences.ts
```diff
@@ -0,0 +1,18 @@
...
+  )
+}
+
+const USER_PREFERENCES_QUERY = graphql`
+  query fetchUserPreferencesQuery {
+    me {
+      lengthUnitPreference
+    }
+  }
+`
```
damassi, line 18: This file might be more appropriately placed in the `/System` folder, close to where its used (and would be modified later)

dimatretyak: do you mean to place this file under the path `src/System/fetchUserPreferences.ts`?

dimatretyak: done

PR #10573, opened 2022-07-20
refactor(caching): Remove full-page caching and cleanup sharify

damassi:

### Description

After doing some analysis we've found that the full-page caching doesn't do much in terms of performance gains and so we're going to remove it along with its associated complexity. 

Additionally, cleaned up our sharify implementation to how it was before, removing unnecessary friction and further complexity. 

#### To delete once merged:
- [ ] PAGE_CACHE_ENABLED
- [ ] PAGE_CACHE_EXPIRY_SECONDS
- [ ] PAGE_CACHE_TYPES
- [ ] PAGE_CACHE_VERSION


src/v2/System/Client/syncSharify.ts
```diff
@@ -1,34 +0,0 @@
...
-      (navigator.userAgent.match(/Android/i) &&
-        navigator.userAgent.match(/Mobile/i))
-  )
-
-  // Load sensitive data from cookies.
-  sd["ARTSY_XAPP_TOKEN"] = Cookies.get("ARTSY_XAPP_TOKEN")
-  sd["IP_ADDRESS"] = Cookies.get("IP_ADDRESS")
-  sd["REQUEST_ID"] = Cookies.get("REQUEST_ID")
-  sd["SESSION_ID"] = Cookies.get("SESSION_ID")
-}
```
damassi, line 34: All this stuff is here: https://github.com/artsy/force/blob/b123a61861519addd6b54bf03ba66ffb9bd0a2bd/src/lib/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts#L11

Which is another thing on the list, to clean up this abstraction too.

mzikherman (approved):
Nice! I thought ripping it out would be even simpler, I hadn't recalled the weird Sharify implications about serving the cached page w/ those 'non cacheable' pieces.

PR #10580, opened 2022-07-21
feat [CX-2658]: Add My Collection tab to Settings tabs

MrSltun:

This PR solves [CX-2658]

### Description
This PR adds My Collection tab inside SettingsApp when `my-collection-web` feature flag is enabled

image

### Follow Up
When we release My Collection on Web, we should remove `TABS_WITH_MYC` and add My Collection to `TABS`

[CX-2658]: https://artsyproduct.atlassian.net/browse/CX-2658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/v2/Apps/Settings/SettingsApp.tsx
```diff
@@ -20,7 +21,23 @@ const TABS = [
...
+  { name: "Order History", url: "/settings/purchases" },
+  { name: "Bids", url: "/settings/auctions" },
+  { name: "Payments", url: "/settings/payments" },
+  { name: "Shipping", url: "/settings/shipping" },
+]
+
 const SettingsApp: React.FC<SettingsAppProps> = ({ me, children }) => {
+  const isMyCollectionEnabled = useFeatureFlag("my-collection-web")
+
+  const tabs = isMyCollectionEnabled ? TABS_WITH_MYC : TABS
```
dariakoko, line 40: I think instead of using two different arrays you can push a new element to the specific position of the array
Should be something like this
```TABS.splice(2, 0, { name: "My Collection", url: "/settings/my-collection" })```

Comment (dariakoko) [reply src/v2/Apps/Settings/SettingsApp.tsx:40]: But I don't mind your approach as well :)

Comment (MrSltun) [reply src/v2/Apps/Settings/SettingsApp.tsx:40]: It was my first thought, but when I tried it with a useEffect there was a weird behavior with it, sometimes it doesn't render, and sometimes it renders after a delay (check video)
but maybe it's my device, maybe it could be solved with SSR

https://user-images.githubusercontent.com/20655703/180221650-882e4eac-c893-4a9a-abaa-1effc082b3dc.mp4

Comment (olerichter00) [reply src/v2/Apps/Settings/SettingsApp.tsx:40]: You could give [compact](https://lodash.com/docs/4.17.15#compact) a try:

```
const TABS = compact([
  ...
  isMyCollectionEnabled && { name: "My Collection", url: "/settings/my-collection" },
  ...
])
```

olerichter00:
Closing in favour of https://github.com/artsy/force/pull/10592

PR #10587, opened 2022-07-22
feat: keep pills extraction logic as lib

dimatretyak:

### ToDo
- [x] Create [new repository](https://github.com/artsy/extract-filter-pills)
- [x] Don't use `lodash`
- [ ] Setup CI/CD
- [ ] Add more tests
- [ ] Add `prettier`
- [ ] Add `husky`
- [ ] Setup `jest`
- [ ] Fix TS warnings/errors

### DIFF
--- a/src/Components/SavedSearchAlert/Components/ActiveFilterPillsAndCreateAlert.tsx
+++ b/src/Components/SavedSearchAlert/Components/ActiveFilterPillsAndCreateAlert.tsx
@@ -10,7 +10,6 @@ import { DEFAULT_METRIC } from "Components/ArtworkFilter/Utils/metrics"
 import { AuthModalOptions } from "Utils/openAuthModal"
 import { ContextModule, Intent } from "@artsy/cohesion"
 import { SavedSearchCreateAlertButton } from "./SavedSearchCreateAlertButton"
-import { prepareForPillsExtract } from "../Utils/extractPills"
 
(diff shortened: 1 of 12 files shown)


dzucconi:
I would strongly recommend against this course of action and wonder if logic could be extracted into Metaphysics?

dimatretyak:
> could be extracted into Metaphysics

I have tried this and the main problem with this approach is that the user will have a bad UX. For example:
- the user has selected some filters
- we send query to Metaphysics to get human-readable labels for pills
- we will get a "jumping" and "delayed" state when the query is successfully completed and pills will be displayed to the user (see an example of how it might look below). As a solution to this problem, we can display the loading state (for example, instead of a real pill, display a "loading" pill), but this also does not solve the problem completely, since we do not the final size of the pill with real content

https://user-images.githubusercontent.com/3513494/180789210-fe51b053-6250-4ed7-9335-14d29c6149a3.mp4

dzucconi:
@dimatretyak could you not prefetch all of the labels ahead of time? TBH I'd rather have the loading state than this logic extracted to its own library. We went down this path before with Backbone and it was, in short, a _nightmare_ from a maintenance perspective.

dimatretyak:
> We went down this path before with Backbone and it was, in short, a nightmare from a maintenance perspective

Thanks for sharing this 🙏
I will see what can be done and how to solve the problems I have encountered if we decide to continue with Metaphysics approach

PR #10598, opened 2022-07-25
experiment(compiler): Add experimental SWC compiler

damassi:
> 👋 Looking for volunteers to test this 

The type of this PR is: **Experiment**

### Description

Similar to Next.js, this PR updates force to use [the Rust-based SWC compiler](https://swc.rs/) for compilation performance. Still need to run the numbers a bit more and compare against a few things as the results so far are milder than I'd expect. However, that said, everything is just about working! Force compiles and things run as one would expect. 

Given the bleeding-edge nature of this work, the time savings need to be substantial enough en-masse to justify moving away from Babel. That Next.js is using it as its main compiler is promising, though. 

#### Update (8/5/2022): 

Turning this flag on by default, which should apply to both tests and local development. To opt-out, set this flag in your `.env`:  
```sh
EXPERIMENTAL_SWC_COMPILER_DISABLED=true
```

A handful of tests (~12) were disabled in [this commit](https://github.com/artsy/force/pull/10598/commits/27ed054f9f7f1316abb5c5b50a02b05e0393afd1) and flagged with a comment: 
```
// FIXME: SWC_COMPILER_MIGRATION
```
Many of these tests were anomalous; we should go through and resolve but not hold up the performance wins in this PR in the meantime. 

Also note that SWC compilation is *not* enabled in prod; this is only a dev-time DX / CI test runner performance improvement. 

---

> ⚠️ To use this feature be sure to prepend the **`EXPERIMENTAL_SWC_COMPILER=true`** env flag to any commands

### Changes: 
- Add [`@swc/jest`](https://swc.rs/docs/usage/jest) for compiling Jest tests with SWC
- Add [`@swc/plugin-relay`](https://github.com/swc-project/plugins/tree/main/packages/relay) for compiling `relay` graphql tags
- Add [`@swc/plugin-styled-components`](https://github.com/swc-project/plugins/tree/main/packages/styled-components) for extracting styled-components style tags for SSR rendering

Have also added a new CI job (`test-swc`) so that we can compare the old against the new) 
 
Screen Shot 2022-07-25 at 6 25 07 PM

### Issues: 
- There are currently about 11 failing tests

### Some Test Commands: 
- `EXPERIMENTAL_SWC_COMPILER=true yarn start`, then make changes here and there and resave
- `EXPERIMENTAL_SWC_COMPILER=true yarn build:client:dev`
- `EXPERIMENTAL_SWC_COMPILER=true yarn jest Apps/Auction`
- `EXPERIMENTAL_SWC_COMPILER=false yarn start`
- `EXPERIMENTAL_SWC_COMPILER=false yarn build:client:dev`
- `EXPERIMENTAL_SWC_COMPILER=false yarn jest Apps/Auction`

### Initial Results:
- `yarn jest Apps/Auction`: 35s babel, 18s SWC. With cache 13s
- `yarn build:client:dev` : 75s babel, 38s SWC. With cache 12s


araujobarret (dismissed):
The speed improvement is incredible and my Mac fan thanks you also! Plus no hacks added 👏

AlicanAkyuz (dismissed):
really looking forward to having this experiment on main 🍾

damassi:
Awesome y'all. I need to fix a few tests but once that's done we can get this into `main` 👍

damassi:
> Plus no hacks added 

@araujobarret - this surprised me too 😄. Not many hacks, but it was quite hard to find how all the pieces fit together since there are no docs. I need to open a PR to the SWC repo giving an example.

damassi:
My results:
- `yarn jest Apps/Auction: 35s babel, 18s SWC. With cache 13s`
- `yarn build:client:dev : 75s babel, 38s SWC. With cache 12s`

PR #10599, opened 2022-07-25
fix: password required in CreateBackupSecondFactorsInput

ovasdi:

This PR solves: [PLATFORM-4340](https://artsyproduct.atlassian.net/browse/PLATFORM-4340)

### Description

This updates `useCreateSettingsBackupSecondFactorsMutation.ts` to no longer default to an empty object as a password is now required by the schema. [MP pr here](https://github.com/artsy/metaphysics/pull/4247).

This would resolve the _type-error_ observed in https://github.com/artsy/force/pull/10590. Failed job run [here](https://app.circleci.com/pipelines/github/artsy/force/34427/workflows/fa9e10a7-1851-493c-aec0-733c6e7040ae/jobs/265378).

### DIFF
--- a/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/useCreateSettingsBackupSecondFactorsMutation.ts
+++ b/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/useCreateSettingsBackupSecondFactorsMutation.ts
@@ -10,7 +10,7 @@ export const useCreateSettingsBackupSecondFactors = () => {
   const { relayEnvironment } = useSystemContext()
 
   const submitCreateSettingsBackupSecondFactors = (
-    input: CreateBackupSecondFactorsInput = {}
+    input: CreateBackupSecondFactorsInput
   ): Promise<useCreateSettingsBackupSecondFactorsMutationResponse> => {
     return new Promise((resolve, reject) => {
       commitMutation<useCreateSettingsBackupSecondFactorsMutation>(


joeyAghion (approved):

src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/useCreateSettingsBackupSecondFactorsMutation.ts
```diff
@@ -10,7 +10,7 @@ export const useCreateSettingsBackupSecondFactors = () => {
   const { relayEnvironment } = useSystemContext()
 
   const submitCreateSettingsBackupSecondFactors = (
-    input: CreateBackupSecondFactorsInput = {}
+    input: CreateBackupSecondFactorsInput
```
damassi, line 13: if this input is required, any reason why we don't have `CreateBackupSecondFactorsInput!` (note the `!`)

damassi: (Non-blocking, just thinking out loud that that's the more correct type-safe way of doing this0

ovasdi: @damassi I tried that by the editor keeps auto-correcting it to `!CreateBackupSecondFactorsInput`. Here is what I see:

Screen Shot 2022-07-25 at 2 39 35 PM

mzikherman: @damassi This is vanilla TS, I think you're thinking about GraphQL types + nullability? This should already be sufficient as-is since its not an optional argument.

damassi: Ah, apologies, to elaborate, this should be related to the internals of the generated type as it relates to the query itself. The `!` would apply to the graphql query, not the typescript definition.

damassi: yup exactly @mzikherman - referring to graphql

ovasdi: @damassi do you mean this? https://github.com/artsy/force/blob/9fb0f70c7abda2db06a79d80d4a77dea68c53334/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/useCreateSettingsBackupSecondFactorsMutation.ts#L30

damassi: yep 👍

PR #10608, opened 2022-07-26
fix(TX-499): order app spacing

rquartararo:
The type of this PR is: **FIX/REFACTOR**

This PR solves [TX-499]

### Description
This PR fixes the spacing issues in the order app routes. 
This PR also adds `OrderRouteContainer` so we can more easily update all route layouts in the future. 

### Screenshots 
Screenshot 2022-07-26 at 10 04 09 AM

Screenshot 2022-07-26 at 10 03 47 AM

[TX-499]: https://artsyproduct.atlassian.net/browse/TX-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -1,17 +1,11 @@
 import { Button, Flex, Spacer } from "@artsy/palette"
 import { Accept_order } from "__generated__/Accept_order.graphql"
-import { TwoColumnLayout } from "Apps/Order/Components/TwoColumnLayout"
 import { track } from "react-tracking"
 import { RouteConfig, Router } from "found"
 import { Component } from "react"
 import { Media } from "Utils/Responsive"
-import {
-  OrderStepper,
-  counterofferFlowSteps,
-} from "../../Components/OrderStepper"
-
+import { counterofferFlowSteps } from "../../Components/OrderStepper"
 import { RelayProp, createFragmentContainer, graphql } from "react-relay"
-
 import { AcceptOfferMutation } from "__generated__/AcceptOfferMutation.graphql"
 import { ConditionsOfSaleDisclaimer } from "Apps/Order/Components/ConditionsOfSaleDisclaimer"
 import { ShippingSummaryItemFragmentContainer as ShippingSummaryItem } from "Apps/Order/Components/ShippingSummaryItem"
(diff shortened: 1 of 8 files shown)


fladson (dismissed):

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -367,6 +367,7 @@ export const ReviewRoute: FC<ReviewProps> = props => {
         currentStep="Review"
         steps={order.mode === "OFFER" ? offerFlowSteps : buyNowFlowSteps}
       />
+      <Spacer mb={4} />
```
AlicanAkyuz, line 370: **question**: do we know why payment route is wrapped by a Box (above in 365) while other routes are not?

AlicanAkyuz (approved):
thank you for making the UI better 🙏  the changes look good to me. 

I have a suggestion that can also be considered in a future iteration: it looks like we have `<OrderStepper /> `and `<Spacer />` in the route indexes very consistently. we could simply our job by abstracting this layout into a simple wrapper. something like this roughly: 

```
const myHOC = ({ currentStep, steps, children }) => {
  return (
    <>
      <OrderStepper currentStep={currentStep} steps={steps} />
      <Spacer mb={4} />
      {children}
    </>
  )
}
```

it might be even possible to put `<TwoColumnLayout />` into the wrapper and pass a content explicitly.

src/Apps/Order/Routes/Review/index.tsx
AlicanAkyuz: ah okay, thank you for figuring it out!

damassi:
(Since `myHOC` is technically still a react component, we should make sure to use `CapsCase` there -- something like `StepperContainer`)

AlicanAkyuz:
@rquartararo yes, this is looking good!

damassi:
@AlicanAkyuz - Just for clarity sake, one thing that came to me when I woke up today was the nature of higher-order components just in general. What they are is an example of [partial application](https://www.freecodecamp.org/news/how-to-use-partial-application-to-improve-your-javascript-code-5af9ad877833/), where one invokes a function to return another function (vs just passing through `children`, in react):

```js
const toggleVisibility = (options) => {
  return (props) => {
      if (options.show) {
        return <div>...</div>
      } else {
        return ....
      }
    }
  }
}

const MyComponent = toggleVisibility({ show: true })

<MyComponent /> 
```

Wanted to drop this note just to explain a bit how they work under the hood.

AlicanAkyuz:
@damassi I guess HOC was the wrong term to use, I meant a simple wrapper for consistent spacing across the checkout steps. thank you for the detailed note 🙏

AlicanAkyuz (approved):

PR #10612, opened 2022-07-26
fix: Add back partner supplied bios GRO-1170

jonallured:
This PR adds back the partner supplied bios - here's a few takes:

### with a partner supplied bio

Screen Shot 2022-07-26 at 4 29 52 PM

### with an editorial bio

Screen Shot 2022-07-26 at 4 30 06 PM

### with no bio at all

Screen Shot 2022-07-26 at 4 30 01 PM

https://artsyproduct.atlassian.net/browse/GRO-1170

/cc @artsy/grow-devs


TMcMeans:
lgtm but I'm a little confused about where the conditional for rendering partner bios vs generic/no bios is, is that supposed to be `showCredit`?? Also, did we eventually make sense of where the CVs are supposed to go??

jonallured:
Good call @TMcMeans I def went too fast and could have added back the tests too - I have done that and this is ready for another review! 🙌 

Your question about conditional rendering is a good one and caused me to make some other changes too. First let's remember that editorial bios run up top next to the artist's name whereas partner bios run under the Notable Works rail. So the conditional logic is split:

* in `ArtistHeader` we check for a partner bio and suppress the editorial bio (here)[https://github.com/artsy/force/blob/jonallured/fix-add-back-partner-supplied-bios/src/Apps/Artist/Components/ArtistHeader/ArtistHeader.tsx#L26]
* then here in this PR i've added a bit to `ArtistCareerHighlights` that checks for the partner bio and runs it if should

That second bit could have used a better name so I switched things around and made a boolean called `showPartnerBio` that checks the same things as before but makes it more clear I think. What do you think?

---

In terms of CV, that's outside the scope of this PR. We did decide that we could add the CV link elsewhere on the page but I'm going to follow up with that in a separate PR.

TMcMeans:
I like the `showPartnerBio` boolean being parsed out into a variable it makes things much more readable.

TMcMeans (approved):

PR #10614, opened 2022-07-26
feat(onboarding): Add new onComplete handler to context, invoking the updateUserProfile mutation

damassi:

This PR solves [GRO-1168]

### Description

Part 1 of a two PR body of work. This PR updates the onboarding flow's context with a new `onComplete` function that can be called when we're certain onboarding has completed. When invoked, it calls the `updateUserProfile` with a `completedOnboarding: true` value. 

Also renames `onDone` to `onClose`.  

Follow-up PR will wire up to the final thank you screen, to be invoked. 

cc @artsy/grow-devs 

[GRO-1168]: https://artsyproduct.atlassian.net/browse/GRO-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Onboarding/OnboardingApp.tsx
+++ b/src/Apps/Onboarding/OnboardingApp.tsx
@@ -13,7 +13,7 @@ export const OnboardingApp: FC = () => {
   }
 
   return (
-    <OnboardingProvider onDone={handleDone}>
+    <OnboardingProvider onClose={handleDone}>
       <OnboardingDebug />
 
       <Spacer mt={2} />
--- a/src/Components/Onboarding/Hooks/useOnboardingContext.tsx
+++ b/src/Components/Onboarding/Hooks/useOnboardingContext.tsx
@@ -75,11 +75,8 @@ const OnboardingContext = createContext<{
   current: string
   dispatch: React.Dispatch<Action>
   next(): void
-  // Called when the user has finished the onboarding process
   onComplete(): void
-  // Called on close and other cancel actions
-  // TODO: Should we rename this to onClose?
-  onDone(): void
+  onClose(): void
   progress: number
   state: State
   workflowEngine: WorkflowEngine
@@ -88,19 +85,19 @@ const OnboardingContext = createContext<{
   dispatch: () => {},
   next: () => {},
   onComplete: () => {},
-  onDone: () => {},
+  onClose: () => {},
   progress: 0,
   state: DEFAULT_STATE,
   workflowEngine: new WorkflowEngine({ workflow: [] }),
 })
 
 interface OnboardingProviderProps {
-  onDone(): void
+  onClose(): void
 }
 
 export const OnboardingProvider: FC<OnboardingProviderProps> = ({
   children,
-  onDone,
+  onClose,
 }) => {
   const basis = useRef<State>(DEFAULT_STATE)
   const { relayEnvironment } = useSystemContext()
@@ -120,7 +117,7 @@ export const OnboardingProvider: FC<OnboardingProviderProps> = ({
 
   const { workflowEngine, current, next, reset: __reset__ } = useConfig({
     basis,
-    onDone,
+    onClose,
   })
 
   const [state, dispatch] = useReducer(reducer(__reset__), DEFAULT_STATE)
@@ -139,7 +136,7 @@ export const OnboardingProvider: FC<OnboardingProviderProps> = ({
         dispatch,
         next,
         onComplete: handleComplete,
-        onDone,
+        onClose,
         progress,
         state,
         workflowEngine,
--- a/src/Components/Onboarding/Hooks/useOnboardingContext.tsx
+++ b/src/Components/Onboarding/Hooks/useOnboardingContext.tsx
@@ -6,6 +6,8 @@ import {
   useReducer,
   useEffect,
 } from "react"
+import { useSystemContext } from "System"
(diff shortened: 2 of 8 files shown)


src/Components/Onboarding/Hooks/useOnboardingContext.tsx
```diff
@@ -73,6 +75,10 @@ const OnboardingContext = createContext<{
   current: string
   dispatch: React.Dispatch<Action>
   next(): void
+  // Called when the user has finished the onboarding process
+  onComplete(): void
+  // Called on close and other cancel actions
+  // TODO: Should we rename this to onClose?
```
laurabeth, line 81: Is this called whenever someone closes onboarding, or do we check to see if they've actually filled out all answers?

If the former, I would rename.

laurabeth (approved):

src/Components/Onboarding/Hooks/useOnboardingContext.tsx
damassi: That's the thing. I would probably defer to @dzucconi when he is back on the thinking here. So far `onDone` is seemingly being used as close, because we need a mechanism to close the modal in a few different places, from a centralized context. I added `onComplete` in order to differentiate (ie, fire `onComplete` inside of a `useEffect` when @TMcMeans 's new thank you screen mounts)

laurabeth: Well, it seems like "onComplete" is appropriate here. Complete means that the process of onboarding is finished, closed or I guess "done" in this case, means the user left without completing. 

In this case, then, I actually don't like the use of "onDone" and would change that to "onClose" instead.

damassi: Digging into this a bit further, and noticed this line: https://github.com/artsy/force/blob/main/src/Components/Onboarding/config.ts#L44-L48

So this seems right: 
- When the workflow is "done", call `onClose` (close the modal)
- Workflow can be "done" any number of ways, including by a user exiting early -- so close the modal
- This is different than `onComplete`, which says: user has successfully completed the onboarding process.

So in sum, will rename 😄

laurabeth: We did it! 💃🏽

src/Apps/Onboarding/OnboardingApp.tsx
```diff
@@ -13,7 +13,7 @@ export const OnboardingApp: FC = () => {
   }
 
   return (
-    <OnboardingProvider onDone={handleDone}>
+    <OnboardingProvider onClose={handleDone}>
```
TMcMeans, line 16: small nitpicky thing- wouldnt it make sense to call this handleClose now for naming convention sake??

src/Components/Onboarding/Hooks/useOnboardingContext.tsx
```diff
@@ -73,33 +75,49 @@ const OnboardingContext = createContext<{
...
+  const { relayEnvironment } = useSystemContext()
+  const { submitUpdateMyUserProfile } = useUpdateMyUserProfile({
+    relayEnvironment,
+  })
+
+  const handleComplete = async () => {
+    try {
+      await submitUpdateMyUserProfile({
+        completedOnboarding: true,
+      })
```
damassi, line 112: this is the only relevant addition

TMcMeans: oh ok well then disregard any other comments then for sure

TMcMeans: am i correct in seeing that this function handles the mutation to update the user profile? not the collector profile??

TMcMeans (approved):

src/Components/Onboarding/Hooks/useOnboardingContext.tsx
damassi: This mutation will eventually call [this line](https://github.com/artsy/gravity/blob/fec1f717d96a144c46e88aa66731bde8403c01a8/app/api/v1/me_endpoint.rb#L69); in this case we're just re-wiring-back-up what existed before.

src/Apps/Onboarding/OnboardingApp.tsx
damassi: Ah, its actually backwards (this is a good react idiom to memorize):
- Callback handlers defined as props are always named `onSomething` (which matches native events like `onClick`, `onMouseDown`, etc. This means something is being passed in from the _outside_
- Handlers defined within the body of a component are always `handleSomething`. This means that a function is being invoked from the _inside_.
- This makes it clear when reading code that anything with `on` is passed in externally, while anything with `handle` is managed locally. 

Here's a small example:

```tsx
const MyComponent = ({ onClose }) => {
  const handleClose = () => {
    if (onClose) {
      onClose()
    }
    
    // do something else 
  }
  
  return (
    return (
      <div onClick={handleClose}>Click me</div>
    )
  )
}
```

damassi: Lemme know if this makes sense! can explain more

TMcMeans: oh ok that totally makes sense!

PR #10616, opened 2022-07-27
fix: PX-5140 - artist url redirect

Mitch-Henson:

This PR solves [PX-5140]

### Description

It seems that the redirection was broken with the latest changes. The reason for this (and why the tests did not pick this up) is that the tests were passing in the valid `req.route.path` which doesn't actually happen in the real world. When passed an array like is done [here](https://github.com/artsy/force/blob/c9b7c872408e743614419fd0703874b4cee56357/src/Apps/Redirects/redirectsServerRoutes.tsx#L162-L176), the specific path which triggers the function from being run is lost and instead the entire array of routes are passed through to the function, e.g.

```
/:id,/:id/overview,/:id/about,/:id/shows,/:id/works,/:id/collection,/:id/shop,/:id/artists,/:id/artist/:artistId,/:id/articles,/:id/contact
```

This breaks all logic that relies on this specific path, including the artist URL redirect which is found [here](https://github.com/artsy/force/blob/c9b7c872408e743614419fd0703874b4cee56357/src/Apps/Redirects/redirectsServerRoutes.tsx#L93-L99).

I have broken out the logic so that each function only needs to deal with the routes it cares about to avoid this issue. Additionally, as part of these changes, I have removed a route redirection for fair info which was not working and also potentially has not been working for a while. Some discussion about this can be found [here](https://artsy.slack.com/archives/CP9P4KR35/p1658915137403989)

[PX-5140]: https://artsyproduct.atlassian.net/browse/PX-5140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Redirects/__tests__/redirectsServerRoutes.jest.ts
+++ b/src/Apps/Redirects/__tests__/redirectsServerRoutes.jest.ts
@@ -1,11 +1,165 @@
 import { Request, NextFunction } from "express"
 import {
-  handlePartner,
+  handlePartnerOverview,
+  handlePartnerWorks,
+  handlePartnerArtist,
+  handlePartnerGenericRedirect,
+  handleFairArtworks,
   handleFair,
   ResWithProfile,
 } from "../redirectsServerRoutes"
 
-describe("handlePartner", () => {
+describe("handlePartnerOverview", () => {
+  let req: Request
+  let res: ResWithProfile
+  let next: jest.Mock<NextFunction>
+
+  beforeEach(() => {
+    jest.resetModules()
+
+    res = ({
+      locals: {
+        profile: {
+          owner: {
+            __typename: "Partner",
+            slug: "white-cube",
+          },
+        },
+      },
+      redirect: jest.fn(),
+    } as unknown) as ResWithProfile
+
+    next = jest.fn()
+  })
+
+  it.each([
+    [
+      "/:id/overview,/:id/about",
+      "/partner/white-cube",
+      "/white-cube/overview",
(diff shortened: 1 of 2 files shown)


leamotta (approved):
Thanks for digging this problem out! Looks good to me, I'm adding @damassi for an extra pair of eyes on the review!

Mitch-Henson:
@leamotta From what I can see on slack @damassi is actually out until the 26th of September so maybe we shouldn't wait for him. Also from his [comment on slack](https://artsy.slack.com/archives/CP9P4KR35/p1658920138102019?thread_ts=1658915137.403989&cid=CP9P4KR35), he may not be the best person to talk to, just the person who moved things over.

Mitch-Henson:
These changes triggered a lot of discussion which has led to the creation of a [followup ticket](https://artsyproduct.atlassian.net/browse/FX-4115). This ticket should mean that these changes can be merged to deal with the issue in production and allow the questions raised in the slack thread to be addressed by the appropriate team

damassi:
@Mitch-Henson - def tag me on force PRs if you have any questions! I keep an on on all the PRs that come through here and can always answer any questions that may come up. 

(Also, where does it say i'm out until Sept 26th in slack?)

damassi:
(PR looks great 👌)

Mitch-Henson:
@damassi I just confused two different people. I thought that @\dzucconi had your github tag. Sorry for the confusion!

PR #10618, opened 2022-07-27
feat: update test attributes

kiryl-zubarau:

This PR solves [TX-281]

### Description

This PR updates test attributes for the save and continue buttons

[TX-281]: https://artsyproduct.atlassian.net/browse/TX-281?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -145,6 +145,7 @@ export const BankAccountPicker: FC<Props> = props => {
         <>
           {bankAccountHasInsufficientFunds && <InsufficientFundsError />}
           <Button
+            data-test="bank-transfer-save-existing"
             onClick={handleContinue}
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -145,6 +145,7 @@ export const BankAccountPicker: FC<Props> = props => {
         <>
           {bankAccountHasInsufficientFunds && <InsufficientFundsError />}
           <Button
+            data-test="bank-transfer-save-existing"
```
damassi, line 148: This is bit nitpicky, but in general: but why not `bankTransferSaveExisting`? We use camelCase everywhere else, and this string will still be used in code (tests)

PR #10623, opened 2022-07-27
feat(onboarding): Call onComplete on complete

damassi:

### Description

This dispatches the `onboarding_complete` event to the user model in gravity, called when a user hits the thank you page. 

cc @artsy/grow-devs

### DIFF
--- a/src/Components/Onboarding/Views/OnboardingThankYou.tsx
+++ b/src/Components/Onboarding/Views/OnboardingThankYou.tsx
@@ -1,9 +1,16 @@
 import { Box, Flex, Text } from "@artsy/palette"
-import { FC } from "react"
+import { FC, useEffect } from "react"
 import { OnboardingFigure } from "../Components/OnboardingFigure"
 import { OnboardingSplitLayout } from "../Components/OnboardingSplitLayout"
+import { useOnboardingContext } from "../Hooks/useOnboardingContext"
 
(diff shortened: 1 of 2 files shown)


src/Components/Onboarding/Views/OnboardingThankYou.tsx
```diff
@@ -1,9 +1,16 @@
 import { Box, Flex, Text } from "@artsy/palette"
-import { FC } from "react"
+import { FC, useEffect } from "react"
 import { OnboardingFigure } from "../Components/OnboardingFigure"
 import { OnboardingSplitLayout } from "../Components/OnboardingSplitLayout"
+import { useOnboardingContext } from "../Hooks/useOnboardingContext"
 
 export const OnboardingThankYou: FC = () => {
+  const { onComplete } = useOnboardingContext()
```
damassi, line 8: _Might_ be a case of over-engineering here, but wanted to keep the complete action decoupled from the thank you page itself since there might be multiple paths to completion, independent of this page (eg, landing on a gene page where one clicks off to an artwork. The gene page is technically _also_ a completion moment -- we can't intercept clicks to the artworks to then redirect to thank you which then redirects to artwork page. And anyways, the thank you comes from the loading interstitial before the artworks page is displayed)

TMcMeans (approved):

PR #10637, opened 2022-07-29
feat: Update staff info on advisory page 1164

jonallured:
~You know me, I just really like a good `quick win` PR on a Friday before vacay.~ 😎 🌴 🍹 

I realized that this PR would be a good chance to try out the i18n work that @lidimayra and friends have put so much time into and I think it worked pretty well! I first made the changes to the page that the ticket called for and then I took a step back and extracted certain strings to the en-EU file. Here's how I thought about the page:

<details>

<summary>meet the specialists: i18n pass</summary>

meet-the-specialists

</details>

In that screenshot I have labels various bits of text with the i18n keys I used to populate the page. I skipped the sections in black with the following reasoning:

* those questions at the top seem pretty static and ripe for extraction but they contain HTML - how should i handle that?
* the information about the people on the page has been extracted already into objects - that doesn't seem worth translating, right?

I'm curious to hear feedback about what I extracted and the names used for the keys - fire away!

https://artsyproduct.atlassian.net/browse/1164

/cc @artsy/grow-devs

### DIFF
--- a/src/System/i18n/locales/en-US/translation.json
+++ b/src/System/i18n/locales/en-US/translation.json
@@ -1,24 +1,30 @@
 {
-  "navbar":{
-    "show":"Show",
-    "shows":"Shows",
-    "buy":"Buy",
-    "artists":"Artists",
-    "artworks":"Artworks",
-    "galleries":"Galleries",
-    "auctions":"Auctions",
-    "fairs":"Fairs",
-    "museums":"Museums",
-    "viewingRooms":"Viewing Rooms",
-    "nfts":"NFTs",
-    "signup":"Sign Up",
-    "login":"Log In",
-    "sell":"Sell",
-    "priceDatabase":"Price Database",
-    "editorial":"Editorial",
-    "downloadApp":"Download Apps",
+  "navbar": {
+    "show": "Show",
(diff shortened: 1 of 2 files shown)


src/System/i18n/locales/en-US/translation.json
```diff
@@ -1,24 +1,30 @@
...
-    "museums":"Museums",
-    "viewingRooms":"Viewing Rooms",
-    "nfts":"NFTs",
-    "signup":"Sign Up",
-    "login":"Log In",
-    "sell":"Sell",
-    "priceDatabase":"Price Database",
-    "editorial":"Editorial",
-    "downloadApp":"Download Apps",
+  "specialists": {
```
damassi, line 2: pls alphabetize all keys before this all gets out of hand 🙏 

Also, this makes me wonder if we can use something other than `.json` here so that we can add comments; it'd be good to be able to add something like:

```
// /meet-the-specialists
"specialists: { ...
```

so people know where the copy ends up living. But that's for another day

damassi (approved):

laurabeth (approved):
FAN 👏🏽 tas 👏🏽 tic 👏🏽

PR #10645, opened 2022-07-29
chore: Extract string for cv link text

jonallured:
Returning to the topic of artist CV links and https://github.com/artsy/force/pull/10631. After that PR was merged I realized that this would have been a good time to extract a string into the locale file so I'm opening a PR to do this.

I extracted the string and used the hook/helper to update the two call sites in the implementation code. Then I went back to the testing code and replaced the const with the actual string itself.

What do folks think about this - is this the right way to do this?

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Artist/Components/ArtistHeader/__tests__/ArtistHeader.jest.tsx
+++ b/src/Apps/Artist/Components/ArtistHeader/__tests__/ArtistHeader.jest.tsx
@@ -3,7 +3,7 @@ import { ArtistHeaderFragmentContainer } from "../ArtistHeader"
 import { useTracking } from "react-tracking"
 import { setupTestWrapperTL } from "DevTools/setupTestWrapper"
 import { screen } from "@testing-library/react"
-import { CV_LINK_TEXT } from "../ArtistHeader"
+import "System/i18n/i18n"
 
 jest.unmock("react-relay")
 jest.mock("react-tracking")
@@ -56,7 +56,7 @@ describe("ArtistHeader", () => {
     expect(screen.getByText("USA, Jan 1 1980")).toBeInTheDocument()
     expect(screen.getByText("111 Followers")).toBeInTheDocument()
     expect(screen.getByText("biographyBlurbText")).toBeInTheDocument()
-    const cvLink = screen.getByText(CV_LINK_TEXT)
+    const cvLink = screen.getByText("See all past shows and fair booths")
     expect(cvLink).toHaveAttribute(
       "href",
       expect.stringContaining("/artist/artistSlug/cv")
@@ -71,7 +71,9 @@ describe("ArtistHeader", () => {
     })
 
     expect(screen.queryByText("biographyBlurbText")).not.toBeInTheDocument()
-    expect(screen.queryByText(CV_LINK_TEXT)).not.toBeInTheDocument()
+    expect(
+      screen.queryByText("See all past shows and fair booths")
+    ).not.toBeInTheDocument()
   })
 
   it("hides follows if count is zero", () => {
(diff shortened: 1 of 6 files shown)


damassi (dismissed):

src/Apps/Artist/Components/ArtistHeader/__tests__/ArtistHeader.jest.tsx
```diff
@@ -56,7 +55,7 @@ describe("ArtistHeader", () => {
     expect(screen.getByText("USA, Jan 1 1980")).toBeInTheDocument()
     expect(screen.getByText("111 Followers")).toBeInTheDocument()
     expect(screen.getByText("biographyBlurbText")).toBeInTheDocument()
-    const cvLink = screen.getByText(CV_LINK_TEXT)
+    const cvLink = screen.getByText("See all past shows and fair booths")
```
damassi, line 58: ty

src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistCareerHighlights.jest.tsx
```diff
@@ -80,6 +79,8 @@ describe("ArtistCareerHighlights", () => {
       expect.stringContaining("partner/number-one-best-gallery")
     )
     expect(screen.getByText("this artist rocks")).toBeInTheDocument()
-    expect(screen.queryByText(CV_LINK_TEXT)).toBeInTheDocument()
+    expect(
+      screen.queryByText("See all past shows and fair booths")
+    ).toBeInTheDocument()
```
lidimayra, line 84: I guess we have to add an `import "System/i18n/i18n"` for this to work here
(I'm actually wondering if maybe it would be better to import this on `setupTestWrapper`? not sure 🤔 )

lidimayra (approved):
Thank you so much for putting effort into it, @jonallured!! 🙌 💜 :heartsy: 😁 

I left a small comment, but it LGTM, so I'm approving it!! \o/

src/Apps/Artist/Routes/Overview/Components/__tests__/ArtistCareerHighlights.jest.tsx
damassi: We could import it lower down, in `src/tests.ts`!

jonallured: Good call, I just updated this individual test with that import and it went green. Then I took it out and moved it to the test setup file as @damassi mentioned and that worked too - thanks all!

jonallured: oh hrm, I now see that adding this to the test.ts file breaks a few tests:

```
src/Apps/Search/Components/__tests__/SearchMeta.jest.tsx
src/Apps/Artist/Components/__tests__/ArtistMeta.jest.tsx
src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/__tests__/ContactInformation.jest.tsx
src/lib/__tests__/openAuthModal.jest.ts
```

What do yall think? Maybe I can separate the test setup improvement from this larger ticket?

lidimayra: Yeah, I agree with the idea of handling the test setup improvement in a separate PR

jonallured: @damassi @lidimayra ok cool then I'll reverse out that test setup change and get this landed. Can come back through with the improvement later. 👍

lidimayra (approved):

PR #10646, opened 2022-07-30
feat: filter some params from logs

erikdstock:

This PR adds a small filtering function to our middleware to remove sensitive query parameters from artsy logs. I paired with @artsyjian on the equivalent change [here](https://github.com/artsy/substance/pull/281) in our ingress controller and there was some question whether we could accomplish a similar goal using our logspout configuration here. However, this approach seems to work, and stops logs going to STDOUT locally from displaying sensitive params either. Because the match used to filter is a regular expression (roughly `${term}=(.+)&`) can match on a term up to the equals sign we could simply exclude all keys ending with `secret` which seems like it would be safe. As it is we are targetting a specific key.

### DIFF
--- a/src/lib/middleware/morgan.ts
+++ b/src/lib/middleware/morgan.ts
@@ -26,12 +26,22 @@ function skipAssets(req: ArtsyRequest, res: ArtsyResponse): boolean {
   )
 }
 
+const BANNED_PARAMS = ["setup_intent_client_secret"]
+export function filterParams(url: string): string {
+  const cleaned = BANNED_PARAMS.reduce((prevUrl, key) => {
+    const re = new RegExp(`${key}=([^&]+)`)
+    return prevUrl.replace(re, (match, p1) => match.replace(p1, "[FILTERED]"))
+  }, url)
+
+  return cleaned
+}
+
 export function logFormat(
   tokens: any,
   req: ArtsyRequest,
   res: ArtsyResponse
 ): string {
-  const url = tokens.url(req, res)
+  const url = filterParams(tokens.url(req, res))
   const status = tokens.status(req, res)
 
   return (
--- a/src/lib/middleware/__tests__/morgan.jest.ts
+++ b/src/lib/middleware/__tests__/morgan.jest.ts
@@ -1,0 +1,53 @@
+import { StringDecoder } from "string_decoder"
+import { logFormat } from "../morgan"
+
+const mockUrl = jest.fn()
+let tokens = {
+  status: jest.fn(),
+  method: jest.fn(),
+  "remote-addr": jest.fn(),
+  "response-time": jest.fn(),
+  "user-agent": jest.fn(),
+  url: mockUrl,
+}
+
+let req = {} as any
+let res = {} as any
+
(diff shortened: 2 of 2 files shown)


src/lib/middleware/morgan.ts
```diff
@@ -26,12 +26,22 @@ function skipAssets(req: ArtsyRequest, res: ArtsyResponse): boolean {
   )
 }
 
+const BANNED_PARAMS = ["setup_intent_client_secret"]
```
damassi, line 29: #superminor - a line-break between the variable and function declaration would make this nicer to read

src/lib/middleware/morgan.ts
```diff
@@ -26,12 +26,22 @@ function skipAssets(req: ArtsyRequest, res: ArtsyResponse): boolean {
...
+export function filterParams(url: string): string {
```
damassi, line 30: Does this need to be exported?

src/lib/middleware/morgan.ts
```diff
@@ -26,12 +26,22 @@ function skipAssets(req: ArtsyRequest, res: ArtsyResponse): boolean {
...
+  const cleaned = BANNED_PARAMS.reduce((prevUrl, key) => {
+    const re = new RegExp(`${key}=([^&]+)`)
+    return prevUrl.replace(re, (match, p1) => match.replace(p1, "[FILTERED]"))
```
damassi, line 33: Can we rename `p1` to something meaningful to someone who has no knowledge of security lingo?

src/lib/middleware/morgan.ts
```diff
@@ -26,12 +26,22 @@ function skipAssets(req: ArtsyRequest, res: ArtsyResponse): boolean {
   )
 }
 
+const BANNED_PARAMS = ["setup_intent_client_secret"]
+export function filterParams(url: string): string {
+  const cleaned = BANNED_PARAMS.reduce((prevUrl, key) => {
+    const re = new RegExp(`${key}=([^&]+)`)
```
damassi, line 32: Also #superminor, reading this reducer code would be a bit easier if `key` -> `bannedParam` and similar to the comment below we don't abbreviate variables:

```tsx
const paramToMask = new RegExp(`${bannedParam}=([^&]+)`)
const maskedUrl = prevUrl.replace(paramToMask, (match, p1) => match.replace(p1, "[FILTERED]"))
return maskedUrl 
```

```tsx

```

src/lib/middleware/morgan.ts
damassi: Also `filterParams` isn't quite accurate; better something like `maskParams`

damassi (dismissed):
A couple of minor changes, but otherwise really like the nice and simple solution here 👍

src/lib/middleware/morgan.ts
erikdstock: I was borrowing from [rails' `filter_parameters`](https://api.rubyonrails.org/classes/ActionDispatch/Http/FilterParameters.html) but yeah, i think i agree

PR #10648, opened 2022-08-01
fix(CX-2680): fix my collection web phase 1 issues

lordkiz:

This PR solves [CX-2680]

### Description
- Fixes all outstanding issues as outlined in the attached Jira ticket

[CX-2680]: https://artsyproduct.atlassian.net/browse/CX-2680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
+++ b/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
@@ -13,7 +13,7 @@ import { resized } from "Utils/resized"
 import { Media } from "Utils/Responsive"
 
 const image = resized(
-  "https://files.artsy.net/images/my-collection-get-app-img.jpg",
+  "https://files.artsy.net/images/my-coll-get-app-img.jpg",
   {
     width: 910,
     height: 652,
@@ -85,7 +85,7 @@ const DesktopLayout: React.FC = () => {
             height="100%"
             srcSet={image.srcSet}
             lazyLoad
-            alt="My Collection app hero image"
+            alt="Screenshot of My Collection sample screen on the App against a painting background"
           />
         </ResponsiveBox>
       </Column>
--- a/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
+++ b/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
@@ -56,37 +56,25 @@ const AppStoreBadges = () => (
   </Flex>
 )
 
-const Title = () => {
-  const title = "Get More from My Collection on the App"
-  return (
-    <>
-      <Media greaterThanOrEqual="lg">
-        <Text variant="xxl" mt={1}>
-          {title}
-        </Text>
-      </Media>
-      <Media lessThan="lg">
-        <Text variant="lg" mt={1}>
-          {title}
-        </Text>
-      </Media>
-    </>
-  )
-}
+const DescriptionColumn = () => (
+  <>
+    <Text variant={["lg", "lg", "xxl"]} mt={1}>
+      Get More from My Collection on the App
+    </Text>
+
+    <Text variant="sm" mt={1} mb={4}>
+      Discover all the features of My Collection on the Artsy app. Coming soon:
+      The full range of features will also be available on the website.
+    </Text>
+    <AppStoreBadges />
+  </>
+)
 
 const DesktopLayout: React.FC = () => {
   return (
     <GridColumns gridRowGap={4} alignItems="center">
       <Column span={6}>
-        <Title />
-
-        <Text variant="sm" mt={1} mb={4}>
-          Discover all the features of My Collection on the Artsy app. Coming
-          soon: The full range of features will also be available on the
-          website.
-        </Text>
-
-        <AppStoreBadges />
+        <DescriptionColumn />
       </Column>
 
       <Column span={6}>
@@ -121,14 +109,7 @@ const MobileLayout: React.FC = () => {
         </ResponsiveBox>
       </Column>
       <Column span={6}>
-        <Title />
-
-        <Text variant="sm" mt={1} mb={4}>
-          Discover all the features of My Collection on the Artsy app. Coming
-          soon: The full range of features will also be available on the
-          website.
-        </Text>
-        <AppStoreBadges />
+        <DescriptionColumn />
       </Column>
     </GridColumns>
   )
--- a/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
+++ b/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
@@ -9,16 +9,59 @@ import {
   Text,
 } from "@artsy/palette"
 import { resized } from "Utils/resized"
+import { Media } from "Utils/Responsive"
 
-export const MyCollectionAppDownload: React.FC = () => {
-  const image = resized(
-    "https://files.artsy.net/images/my-collection-get-app-img.jpg",
-    {
-      width: 910,
-      height: 652,
-    }
-  )
+const image = resized(
+  "https://files.artsy.net/images/my-collection-get-app-img.jpg",
+  {
+    width: 910,
+    height: 652,
+  }
+)
 
+const AppStoreBadges = () => (
+  <Flex>
+    <Box
+      width={120}
+      height={40}
+      as="a"
+      // @ts-ignore
+      href="https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      <Image
+        src="https://files.artsy.net/images/download-ios-app.svg"
+        width="100%"
+        height="100%"
+        alt="ios app store button"
+        loading="lazy"
+      />
+    </Box>
+
+    <Spacer ml={2} />
+
+    <Box
+      width={120}
+      height={40}
+      as="a"
+      // @ts-ignore
+      href="https://play.google.com/store/apps/details?id=net.artsy.app"
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      <Image
+        src="https://files.artsy.net/images/download-android-app.svg"
+        width="100%"
+        height="100%"
+        alt="android play store button"
+        loading="lazy"
+      />
+    </Box>
+  </Flex>
+)
+
+const DesktopLayout: React.FC = () => {
   return (
     <GridColumns gridRowGap={4} alignItems="center">
       <Column span={6}>
@@ -32,45 +75,7 @@ export const MyCollectionAppDownload: React.FC = () => {
           website.
         </Text>
 
-        <Flex>
-          <Box
-            width={120}
-            height={40}
-            as="a"
-            // @ts-ignore
-            href="https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            <Image
-              src="https://files.artsy.net/images/download-ios-app.svg"
-              width="100%"
-              height="100%"
-              alt="ios app store button"
-              loading="lazy"
-            />
-          </Box>
-
-          <Spacer ml={2} />
-
-          <Box
-            width={120}
-            height={40}
-            as="a"
-            // @ts-ignore
-            href="https://play.google.com/store/apps/details?id=net.artsy.app"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            <Image
-              src="https://files.artsy.net/images/download-android-app.svg"
-              width="100%"
-              height="100%"
-              alt="android play store button"
-              loading="lazy"
-            />
-          </Box>
-        </Flex>
+        <AppStoreBadges />
       </Column>
 
       <Column span={6}>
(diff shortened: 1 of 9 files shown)


src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
```diff
@@ -9,16 +9,59 @@ import {
...
+)
+
+const AppStoreBadges = () => (
+  <Flex>
+    <Box
+      width={120}
+      height={40}
+      as="a"
+      // @ts-ignore
+      href="https://apps.apple.com/us/app/artsy-buy-sell-original-art/id703796080"
```
olerichter00, line 29: 👀

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
```diff
@@ -81,22 +83,30 @@ const MyCollectionRoute: FC<MyCollectionRouteProps> = ({ me, relay }) => {
...
+                      Coming soon also on web.
+                    </Text>
+                  </Link>
+
+                  {/**TODO: Replace Close with Clickable
+                   * if you can get clickable to not center its children
+                   * */}
+                  <Close onClick={dismissMyCollectionMessage}>
+                    <CloseIcon />
+                  </Close>
```
olerichter00, line 106: What do you think about using the [Clickable](https://github.com/artsy/force/blob/b96e157ba9abfbe321c9b3f9ca198b7401f232a2/src/Components/Modal/Modal.tsx#L71) here?

olerichter00 (commented):
Great work!

I've left a few comments in the code and maybe we can clarify if the images should align with the text or be centered. Right now it looks like this and it's not quite clear looking at the Figma design.

Screenshot 2022-08-01 at 15 37 34

lordkiz:
> Great work!
> 
> I've left a few comments in the code and maybe we can clarify if the images should align with the text or be centered. Right now it looks like this and it's not quite clear looking at the Figma design.
> 
> Screenshot 2022-08-01 at 15 37 34

@olerichter00 Looks like we need a Layout for Md devices like Pads, atleast for the title. On it.

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
lordkiz: It was originally with Clickable it is almost impossible for the children not to center with this flex layout. https://github.com/artsy/force/pull/10648/files#diff-e8f31aad1a0c51e2e8ff1301402f54800d53bd23779054b4b696ad828351db43R101

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
olerichter00: Yeah, but I don't think it's a good practice because we have a component for links like this. Especially if we have to add a `// ts-ignore` to it.

src/Components/Artwork/GridItem.tsx
```diff
@@ -43,6 +48,19 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
     ? transform(imageURL, { width, height })
     : { src: "", srcSet: "" }
 
+  const [linkTo, setLinkTo] = useState(artwork.href)
+
+  useEffect(() => {
+    if (!!disableRouterLinking) {
+      setLinkTo(window.location.href)
```
damassi, line 55: I'm wondering why this has to happen in a `useEffect`? If the user has JS disabled then this will break navigation during SSR rendering. 

Also, to get the current location we should use our router directly (which is SSR compatable):

```ts
const { match } = useRouter()
...
const url = match.location.href
```

src/Components/Artwork/GridItem.tsx
```diff
@@ -43,6 +48,19 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
...
+    } else {
+      if (linkTo !== artwork.href) {
+        setLinkTo(artwork.href)
+      }
+    }
```
damassi, line 60: To make this lightly less brittle in the future we should guard against `if` conditions in `else` statements (which might not match if the condition changes, leading to a clickable area) and just fall back to an always available URL:

```tsx
} else {
  url = artwork.href
}
```

src/Components/Artwork/GridItem.tsx
```diff
@@ -125,6 +147,10 @@ const Link = styled(RouterLink)`
   left: 0;
 `
 
+const DisabledLink = styled(Link)`
```
damassi, line 150: `Link`'s are deprecated - we should just use `RouterLink` here, even if it's disabled

src/Components/Artwork/Metadata.tsx
```diff
@@ -27,13 +30,30 @@ export const Metadata: React.FC<MetadataProps> = ({
...
+  React.useEffect(() => {
+    if (!!disableRouterLinking) {
+      setLinkTo(window.location.href)
+    } else {
+      if (linkTo !== artwork.href) {
+        setLinkTo(artwork.href)
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [disableRouterLinking])
```
damassi, line 50: Similar to comment above, this doesn't need to be set in a useEffect. Can use `useRouter` hook (from `src/System/Router`) and get access to SSR-friendly `location.href`

src/Components/Artwork/Metadata.tsx
```diff
@@ -47,13 +67,18 @@ export const Metadata: React.FC<MetadataProps> = ({
...
+        showHoverDetails={showHoverDetails}
         showSaveButton={showSaveButton}
         contextModule={contextModule}
       />
-    </RouterLink>
+    </LinkContainer>
   )
 }
 
+const DisabledLink = styled(RouterLink)`
```
damassi, line 78: 👍

src/Components/Artwork/Metadata.tsx
```diff
@@ -15,6 +16,8 @@ export interface MetadataProps
   hideArtistName?: boolean
   hideSaleInfo?: boolean
   isHovered?: boolean
+  showHoverDetails?: boolean
+  disableRouterLinking?: boolean
```
damassi, line 20: Alphabetize 🙏

src/Components/Artwork/GridItem.tsx
```diff
@@ -19,6 +20,8 @@ interface ArtworkGridItemProps extends React.HTMLAttributes<HTMLDivElement> {
   onClick?: () => void
   hideSaleInfo?: boolean
   showSaveButton?: boolean
+  showHoverDetails?: boolean
+  disableRouterLinking?: boolean
```
damassi, line 24: alpha

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
lordkiz: Thanks. But which existing component is suitable for this? I am not aware of any besides good old `<div>`

src/Components/Artwork/GridItem.tsx
lordkiz: Thank youuuuu! 🙏

src/Components/Artwork/GridItem.tsx
lordkiz: Thanks. We don't need this with `match.location.href`

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
olerichter00: What about `RouterLink`?

src/Components/Artwork/GridItem.tsx
lordkiz: `Link` here is actually a styled `RouterLink`

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
lordkiz: I didn't know we could use RouterLink for external urls.

olerichter00: @damassi `RouterLink` for external urls, right?

damassi: Yup! it will fall back to a regular link if it can't find the url passed to it in our router link list. See https://github.com/artsy/force/blob/main/src/System/Router/RouterLink.tsx#L31-L45

damassi (dismissed):

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
```diff
@@ -8,71 +8,106 @@ import {
...
+    </RouterLink>
+  </Flex>
+)
+
+const Title = () => {
+  const title = "Get More from My Collection on the App"
+  return (
+    <>
+      <Media greaterThanOrEqual="lg">
+        <Text variant="xxl" mt={1}>
```
dzucconi, line 64: `variant` is a responsive prop: `variant={['xxl', 'lg]}` this usage of `Media` is unnecessary. It's extraction into its own component is also unnecessary and results in an isolated external margin.

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
```diff
@@ -8,71 +8,106 @@ import {
...
-        </Flex>
+      <Column span={6}>
+        <ResponsiveBox aspectHeight={652} aspectWidth={910} maxWidth="100%">
+          <Image
+            src={image.src}
+            width="100%"
+            height="100%"
+            srcSet={image.srcSet}
+            lazyLoad
+            alt="My Collection app hero image"
```
dzucconi, line 100: https://palette.artsy.net/guides/a11y/#alternative-text

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
```diff
@@ -81,22 +83,30 @@ const MyCollectionRoute: FC<MyCollectionRouteProps> = ({ me, relay }) => {
...
+                  <Box />
+                  <Link to="/settings/my-collection#download-app-banner">
+                    <Text mx={1}>
+                      Access all the My Collection features on the Artsy app.
+                      Coming soon also on web.
+                    </Text>
+                  </Link>
+
+                  {/**TODO: Replace Close with Clickable
+                   * if you can get clickable to not center its children
```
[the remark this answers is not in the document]
dzucconi: Never use `onClick` on `div`s

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
```diff
@@ -177,3 +189,11 @@ export const MyCollectionRouteRefetchContainer = createRefetchContainer(
   },
   MY_COLLECTION_ROUTE_QUERY
 )
+
+const Link = styled(RouterLink)`
```
dzucconi, line 193: `RouterLink` accepts `textDecoration`

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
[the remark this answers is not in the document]
lordkiz: Per design, the CloseIcon should be at the top right corner. With Clickable, it is always centered. Any ideas how to prevent this?

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionAppDownload.tsx
lordkiz: Updated

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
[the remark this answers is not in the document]
olerichter00: You could set the position to `absolute` to position the button in the top right corner like this:

```
                <Flex flexDirection="row" justifyContent="center">
                  <RouterLink
                    to="/settings/my-collection#download-app-banner"
                    textDecoration="none"
                  >
                    <Text mx={1}>
                      Access all the My Collection features on the Artsy app.
                      Coming soon also on web.
                    </Text>
                  </RouterLink>

                  <Clickable
                    onClick={dismissMyCollectionMessage}
                    position="absolute"
                    top={0}
                    right={0}
                    p={1}
                  >
                    <CloseIcon />
                  </Clickable>
                </Flex>
```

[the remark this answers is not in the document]
olerichter00: The [Modal close icon](https://github.com/artsy/force/blob/641734ee0df20c307ae9fa6cc5f765737b005199/src/Components/Modal/Modal.tsx#L71) is implemented in a similar way.

olerichter00 (changes requested):
Let's address https://github.com/artsy/force/pull/10648#discussion_r934768904, https://github.com/artsy/force/pull/10648#discussion_r934769653, and https://github.com/artsy/force/pull/10648#discussion_r934770194 and then merge this PR. 
Please let me know if I can help with any of these.

olerichter00 (approved):

MounirDhahri:
@lordkiz can you address the other comments separately

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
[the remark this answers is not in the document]
olerichter00: Here is a draft PR with the approach: https://github.com/artsy/force/pull/10662

PR #10653, opened 2022-08-02
fix(onboarding): Make slightly more responsive

damassi:

### Description

This PR adjusts the layout slightly to be slightly more responsive. 

cc @artsy/grow-devs

### DIFF
--- a/src/Components/Onboarding/Components/OnboardingSplitLayout.tsx
+++ b/src/Components/Onboarding/Components/OnboardingSplitLayout.tsx
@@ -19,7 +19,7 @@ export const OnboardingSplitLayout: FC<OnboardingSplitLayoutProps> = ({
       <Box
         display={["none", "block"]}
         bg="black100"
-        flexBasis="50%"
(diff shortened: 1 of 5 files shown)


damassi:
Going to merge this and will follow up with any comments if they appear!

src/Components/Onboarding/Components/OnboardingSplitLayout.tsx
```diff
@@ -19,7 +19,7 @@ export const OnboardingSplitLayout: FC<OnboardingSplitLayoutProps> = ({
       <Box
         display={["none", "block"]}
         bg="black100"
-        flexBasis="50%"
+        flexBasis={["30%", "50%"]}
```
dzucconi, line 22: This doesn't even display at this breakpoint; why would it ever be set to 30%?

damassi: Whatever it did, did something that worked.. not sure

PR #10664, opened 2022-08-02
feat: disable submit button on review step until stripe is loaded (take II)

starsirius:

This PR solves [TX-421]

### Description

This is a take II of https://github.com/artsy/force/pull/10376. It's almost identical and I'll comment inline the difference.

Since the previous attempt caused an issue, I assign this to myself and plan to merge and monitor in the morning ET.

[TX-421]: https://artsyproduct.atlassian.net/browse/TX-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Review/index.tsx
+++ b/src/Apps/Order/Routes/Review/index.tsx
@@ -358,7 +358,20 @@ export const ReviewRoute: FC<ReviewProps> = props => {
     props.router.push(`/orders/${props.order.internalID}/shipping`)
   }
 
-  const { order, isCommittingMutation, isEigen } = props
+  const { order, isCommittingMutation, isEigen, stripe } = props
+  const submittable = !!stripe
+
+  const SubmitButton: FC = () => (
+    <Button
+      variant="primaryBlack"
+      width="100%"
+      loading={isCommittingMutation}
+      disabled={!submittable}
+      onClick={() => onSubmit()}
+    >
+      Submit
+    </Button>
+  )
 
   return (
     <Box data-test="orderReview">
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -358,7 +358,20 @@ export const ReviewRoute: FC<ReviewProps> = props => {
...
+  const { order, isCommittingMutation, isEigen, stripe } = props
+  const submittable = !!stripe
+
+  const SubmitButton: FC = () => (
+    <Button
+      variant="primaryBlack"
+      width="100%"
+      loading={isCommittingMutation}
+      disabled={!submittable}
+      onClick={() => onSubmit()}
```
starsirius, line 370: This line is the only difference compared to https://github.com/artsy/force/pull/10376. It was incorrectly implemented as

```jsx
onClick={onSubmit}
```

fladson (approved):
LGTM, I was just thinking if `isCommittingMutation` covers a timeout issue. 
Do we display something if Stripe is not reachable in a considerable amount of time?

starsirius:
Good point! I can do some research on timeout and share back.

starsirius:
Going to merge and babysit for a bit and see if it fixes the flakiness.

src/Apps/Order/Routes/Review/index.tsx
damassi: This is non-blocking and minor in this case, but it's always good to define components outside of other components at the top level. Have gotten into a lot of subtle trouble here in the past.

starsirius:
OK the failure hasn't happened for the past 20 builds so I'm going to call it a success! 🎉

PR #10668, opened 2022-08-03
chore(jest): Update to Jest@28

damassi:

### Description

Upgrades Jest to 28.


package.json
```diff
@@ -209,6 +209,7 @@
     "@babel/preset-typescript": "7.13.0",
     "@babel/register": "7.13.8",
     "@graphql-inspector/core": "1.14.0",
+    "@graphql-tools/jest-transform": "^1.2.0",
```
araujobarret, line 212: this plugin looks better than the other, let's see for how long they'll maintain it 😂

araujobarret (commented):
Looks that went smooth 🍯

package.json
damassi: yah that's a good question. I had to upgrade because of a jest breaking change that the other lib hasn't switched over to, but so far so good.

PR #10669, opened 2022-08-03
ci(test): Speed up tests on CI via parallelism 

damassi:
Related https://github.com/artsy/force/pull/10668 

The type of this PR is: **CI**

#### Jest 

Jest 28 comes with [a new `--shard` argument](https://jestjs.io/docs/next/cli#--shard) which allows one to split the tests across processes all run in parallel. 

Test speeds went from 6+ minutes per run [down to ~2m](https://app.circleci.com/pipelines/github/artsy/force/34836/workflows/ba6af729-8d73-4696-ae37-b00a0445ff4b/jobs/267865/parallel-runs/0?filterBy=ALL), so a 2/3rds reduction in time. If we wanted to decrease it further we could increase the `parallelism` value in our circle config.

The only question here is: Will this in any way constrain other jobs being executed on CircleCI outside of Force?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,10 +137,9 @@ jobs:
           aws_access_key_id: $AWS_ACCESS_KEY_ID
           aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
     working_directory: /app
-    environment:
-      - JEST_JUNIT_OUTPUT: reports/junit/js-test-results.xml
+    parallelism: 4
     steps:
-      - run: yarn test
+      - run: yarn test --shard=$(expr $CIRCLE_NODE_INDEX + 1)/$CIRCLE_NODE_TOTAL
 
   type-check:
     docker:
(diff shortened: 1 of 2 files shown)


joeyAghion (approved):
This seems fine to experiment with. It _will_ limit other builds' CI throughput for the next month or so until we're force-upgraded to Circle's "performance" plan, after which concurrency will not be a bottleneck (but this will use more "credits" so cost more).

Is the 6 minute -> 2 minute decrease _worth it_ though? If a developer can run the whole test suite in this kind of time locally, I'd expect them to rely _less_ on frequent pushes to working branches just for test results. Are developers using Circle for this kind of test feedback still?

damassi:
> Are developers using Circle for this kind of test feedback still 

All the time! I know myself and numerous other devs who only rely on CI and never ever run things locally, unless it's a single focused test. It's common.

PR #10670, opened 2022-08-03
fix(TX-500): remove underline from US bank account only label 

rquartararo:

This PR solves [TX-500]

### Description

This PR is to remove the hover and focus underline style on the radio button "US banks only" label for ACH. This is a one-off fix until a solution for this is developed in Palette.  

My solution was to use the radio description which doesn't have an underline and change its position.

### Screenshots 
Screenshot 2022-08-03 at 9 52 49 AM
Screenshot 2022-08-03 at 9 54 49 AM

[TX-500]: https://artsyproduct.atlassian.net/browse/TX-500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/PaymentContent.tsx
+++ b/src/Apps/Order/Routes/Payment/PaymentContent.tsx
@@ -227,7 +227,7 @@ const getAvailablePaymentMethods = (
   let paymentMethod: CommercePaymentMethodEnum = "CREDIT_CARD"
   const paymentMethods = [
     <BorderedRadio
-      className="payment-method"
+      data-test-id="credit-card"
       key="CREDIT_CARD"
       value={paymentMethod}
       label={
@@ -243,7 +243,7 @@ const getAvailablePaymentMethods = (
   if (availablePaymentMethods.includes("WIRE_TRANSFER")) {
     paymentMethods.push(
       <BorderedRadio
-        className="payment-method"
+        data-test-id="wire-transfer"
         key="WIRE_TRANSFER"
         value={(paymentMethod = "WIRE_TRANSFER")}
         label={
@@ -260,7 +260,7 @@ const getAvailablePaymentMethods = (
   if (availablePaymentMethods.includes("US_BANK_ACCOUNT")) {
     paymentMethods.unshift(
       <RadioWithLabel
-        className="payment-method"
+        data-test-id="us-bank-account"
         key="US_BANK_ACCOUNT"
         value={(paymentMethod = "US_BANK_ACCOUNT")}
         label={
--- a/src/Apps/Order/Routes/__tests__/Payment.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Payment.jest.tsx
@@ -319,7 +319,7 @@ describe("Payment", () => {
     })
 
     it("tracks when the user selects the credit card payment method", async () => {
-      page.selectPaymentMethod(1)
+      page.selectPaymentMethod("CreditCard")
 
       expect(trackEvent).toHaveBeenLastCalledWith({
         action: "clickedPaymentMethod",
@@ -334,8 +334,8 @@ describe("Payment", () => {
     })
 
     it("tracks when the user selects the bank payment method", async () => {
-      page.selectPaymentMethod(1)
-      page.selectPaymentMethod(0)
+      page.selectPaymentMethod("CreditCard")
+      page.selectPaymentMethod("USBankAccount")
 
       // We toggle back and forth to test the selecttion starting from bank account
       expect(
@@ -344,7 +344,7 @@ describe("Payment", () => {
     })
 
     it("renders credit card element when credit card is chosen as payment method", async () => {
-      page.selectPaymentMethod(1)
+      page.selectPaymentMethod("CreditCard")
 
       const creditCardCollapse = page
         .find(CreditCardPickerFragmentContainer)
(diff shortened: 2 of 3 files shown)


src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -218,6 +229,7 @@ const getAvailablePaymentMethods = (
   let paymentMethod: CommercePaymentMethodEnum = "CREDIT_CARD"
   const paymentMethods = [
     <BorderedRadio
+      id="payment-method"
```
iskounen, line 232: **suggestion:** Depending on the available payment methods, this component can render up to 3 DOM elements with the same `id` attribute, which are expected to be unique. Please consider using the `class` attribute instead, or renaming the IDs to be more specific (ex. `payment-method-credit-card`).

Since these IDs are only being used for tests, my preference would be to use a more specific `data-test-id` attribute for each

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -248,19 +261,21 @@ const getAvailablePaymentMethods = (
   // when available, unshift ACH since it's the first option we want to offer
   if (availablePaymentMethods.includes("US_BANK_ACCOUNT")) {
     paymentMethods.unshift(
-      <BorderedRadio
+      <RadioWithLabel
```
iskounen, line 264: **question:** Will the credit card and wire transfer radio buttons still have an underline when they are hovered or selected?

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -206,6 +206,17 @@ const USBankOnlyLabel = styled(Text)`
   color: ${themeGet("colors.orange150")};
   padding: 1px 5px;
   border-radius: 2px;
+  align-self: center;
+`
+
+const RadioWithLabel = styled(BorderedRadio)`
+  div {
+    flex-direction: row;
+    .jDlIOQ {
```
iskounen, line 215: **question:** I am unsure about using this class name as a CSS selector because it looks like a generated name, and I'm not sure if that will change over time. Is this a reliable class name to use for this purpose?

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -248,19 +261,21 @@ const getAvailablePaymentMethods = (
...
-            <USBankOnlyLabel ml={0.5} variant="xs">
-              US bank account only
-            </USBankOnlyLabel>
           </>
         }
-      />
+      >
+        <USBankOnlyLabel ml={0.5} variant="xs">
+          US bank account only
+        </USBankOnlyLabel>
```
iskounen, line 277: **question:** Does moving this component out of the `label` attribute fix the issue, or does the new set of CSS rules in the styled component fix the issue? (Or is it both?)

src/Apps/Order/Routes/__tests__/Payment.jest.tsx
```diff
@@ -354,12 +355,14 @@ describe("Payment", () => {
     })
 
     it("renders bank element when bank transfer is chosen as payment method", async () => {
-      page.selectPaymentMethod(3)
+      page.selectPaymentMethod(0)
```
iskounen, line 358: **suggestion (non-blocking):** This is out of scope for this ticket, but it would be nice to ask for the specific payment method that you need without having to remember its index. If you were to add a `data-test-id` for each payment method, we can refactor this method to select the specific payment method that we want to test.

src/Apps/Order/Routes/Payment/PaymentContent.tsx
iskounen: That make sense. Thank you for the explanation!

src/Apps/Order/Routes/Payment/PaymentContent.tsx
iskounen: I asked this question before you explained how the fix works, and this makes sense now too.

src/Apps/Order/Routes/__tests__/Utils/OrderAppTestPage.tsx
```diff
@@ -108,8 +108,18 @@ export class OrderAppTestPage extends RootTestPage {
...
+      case "CreditCard":
+        this.find("label[data-test-id='credit-card']").simulate("click")
+        break
+      case "WireTransfer":
+        this.find("label[data-test-id='wire-transfer']").simulate("click")
+        break
+      case "USBankAccount":
+        this.find("label[data-test-id='us-bank-account']").simulate("click")
+        break
+    }
```
iskounen, line 122: **suggestion (non-blocking):** This method is a big improvement over the previous one because now the test code is readable and the parameter isn't cryptic.

I think we can make a small improvement on the parameter (which needs to be a specific string) by breaking this up into 3 methods: `selectCreditCard()`, `selectWireTransfer()`, `selectUSBankAccount()`. That's more methods, but each one will be small and simple, so it's easy to read and easy to use.

iskounen (approved):
Thank you for making this happen! This is what I had in mind, and it's cool to see how much easier it is to read the test code.

I have one suggestion on the `selectPaymentMethod` method, but I think it's a matter of preference, so it's non-blocking.

PR #10684, opened 2022-08-04
chore(about2): Updates UI

damassi:

### Description

Finishes up some UI stuff around /about2, including adding nav buttons to the links. 

cc @artsy/grow-devs

### DIFF
--- a/src/Apps/About2/AboutApp2.tsx
+++ b/src/Apps/About2/AboutApp2.tsx
@@ -2,7 +2,9 @@ import {
   Box,
   BoxProps,
   Button,
+  Column,
   Flex,
+  GridColumns,
   Image,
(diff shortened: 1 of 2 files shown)


src/Apps/About2/AboutApp2.tsx
```diff
@@ -16,7 +18,7 @@ import {
   FullBleedHeader,
   FullBleedHeaderOverlay,
 } from "Components/FullBleedHeader"
-import { Masonry } from "Components/Masonry"
+import { scrollIntoView } from "Utils/scrollHelpers"
```
damassi, line 21: Had to use this rather than hook because we're calling the function from outside a react component, and no way to pass in.

PR #10687, opened 2022-08-05
fix(onboarding): Link position

damassi:

### Description

Fixes small issue with font weight and position to match close


src/Components/Onboarding/Components/OnboardingProgress.tsx
```diff
@@ -24,10 +24,11 @@ export const OnboardingProgress: FC<OnboardingProgressProps> = ({
     <Box position="relative">
       <BackLink
         to="/"
-        top={-20}
-        left={-5}
+        top={-19}
+        left={-17}
```
damassi, line 28: 😬

PR #10675, opened 2022-08-03
fix: Prevent My Collection artwork click scrolling and remove cursor: not-allowed style

olerichter00:
The type of this PR is: **Bugfix**

This PR solves [CX-2693]

### Description

- Prevents My Collection artwork click scrolling by not passing any URL.
- Removes `cursor: not-allowed;` artwork link.

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CX-2693]: https://artsyproduct.atlassian.net/browse/CX-2693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10688, opened 2022-08-05
fix(RouterLink): Prevent Disabled RouterLink from routing to top of page

lordkiz:

This PR solves []

### Description
- When a RouterLink's `to` prop is passed a `null`, onClick automatically jumps to top of page. This PR fixes that.

### DIFF
--- a/src/Components/Artwork/Metadata.tsx
+++ b/src/Components/Artwork/Metadata.tsx
@@ -6,7 +6,6 @@ import { BoxProps } from "@artsy/palette"
 import { RouterLink } from "System/Router/RouterLink"
 import { AuthContextModule } from "@artsy/cohesion"
 import styled from "styled-components"
-import { useRouter } from "System/Router/useRouter"
 
 export interface MetadataProps
   extends BoxProps,
@@ -37,12 +36,11 @@ export const Metadata: React.FC<MetadataProps> = ({
   showSaveButton,
   ...rest
 }) => {
-  const { match } = useRouter()
   const LinkContainer = disableRouterLinking ? DisabledLink : RouterLink
 
   return (
     <LinkContainer
-      to={disableRouterLinking ? match.location.pathname : artwork.href}
+      to={disableRouterLinking ? null : artwork.href}
       display="block"
       textDecoration="none"
       textAlign="left"
@@ -65,7 +63,7 @@ export const Metadata: React.FC<MetadataProps> = ({
 }
 
 const DisabledLink = styled(RouterLink)`
-  cursor: not-allowed;
+  cursor: default;
 `
 
 export default createFragmentContainer(Metadata, {
--- a/src/System/Router/RouterLink.tsx
+++ b/src/System/Router/RouterLink.tsx
@@ -42,7 +42,9 @@ export const RouterLink: React.FC<RouterLinkProps> = React.forwardRef(
       return <RouterAwareLink to={to ?? ""} {...deprecated} {...rest} />
     }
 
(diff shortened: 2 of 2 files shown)


olerichter00 (approved):
Thanks for the fix 🤔 

But wasn't a fix like this already merged in https://github.com/artsy/force/pull/10675?

olerichter00:
Ah, this PR fixes the issue for the metadata and https://github.com/artsy/force/pull/10675 only fixed the issue for the grid item itself.

lordkiz:
> But wasn't a fix like this already merged in #10675?

- When we pass null to `to`, since null is not one of our routes the RouterLink uses RouteUnawareLink which is a styled `a` tag. But it passes an empty string to this `a` tag which just takes the user to the top of the page. I just changed it to undefined.
- Also changed the cursor for Metadata

damassi:
@lordkiz, @olerichter00 - can you hover over a link and take a screenshot of the browser status bar and paste it in here? 

Does it say `http://whateverlink/undefined`? If so, we should revert this and fix the problem at its origin, which is in the `MetaData` component. If the link is undefined, we shouldn't wrap the component with a link. Having urls with `undefined` in them isn't great for SEO and other things. 

(Also, whenever changes are made to to code in the `System` folder, be sure to tag @dzucconi or me or @araujobarret as a reviewer on the PR please 🙏)

src/System/Router/RouterLink.tsx
```diff
@@ -42,7 +42,9 @@ export const RouterLink: React.FC<RouterLinkProps> = React.forwardRef(
       return <RouterAwareLink to={to ?? ""} {...deprecated} {...rest} />
     }
 
-    return <RouterUnawareLink href={to ?? ""} {...deprecated} {...rest} />
+    return (
+      <RouterUnawareLink href={to ?? undefined} {...deprecated} {...rest} />
```
damassi, line 46: ☝️

lordkiz:
> @lordkiz, @olerichter00 - can you hover over a link and take a screenshot of the browser status bar and paste it in here?
> Does it say `http://whateverlink/undefined?`

@damassi No, it doesn't.

https://user-images.githubusercontent.com/18648835/183135837-28f91c59-a325-46da-82c7-2b7306a19ef9.mov


> (Also, whenever changes are made to to code in the System folder, be sure to tag @dzucconi or @damassi or @araujobarret as a reviewer on the PR please 🙏)

Noted. Sure. Thanks 🙏

dzucconi:
Why would we ever want a link to null?

damassi:
Yah @lordkiz we should fix that. We can conditionally render the `RouterLink` component vs using `undefined` and keep the behavior as it was.

src/Components/Artwork/Metadata.tsx
```diff
@@ -65,7 +63,7 @@ export const Metadata: React.FC<MetadataProps> = ({
 }
 
 const DisabledLink = styled(RouterLink)`
-  cursor: not-allowed;
+  cursor: default;
 `
```
dzucconi, line 67: This method of disabling the link is _not good_. We shouldn't have any empty anchor tag hanging out like this — it's still keyboard focusable.

dzucconi: You should likely completely swap out the tag here.

dzucconi: This is a good example of "fixing" the symptom rather than the root cause.

lordkiz:
@damassi @dzucconi This PR has been reverted here: https://github.com/artsy/force/pull/10714 🙏

PR #10699, opened 2022-08-05
chore(navbar): Remove focus state on logo in navbar

damassi:

This PR solves [GRO-1194]

### Description

As a number of people have flagged this as (seemingly) a bug, this removes the blue focus state on the logo in the navbar. If we can figure out a way to make the behavior feel a little more solid then lets give it a shot (as I assume this was done for a specific reason). 

[GRO-1194]: https://artsyproduct.atlassian.net/browse/GRO-1194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/NavBar/NavBarPrimaryLogo.tsx
+++ b/src/Components/NavBar/NavBarPrimaryLogo.tsx
@@ -2,7 +2,6 @@ import * as React from "react"
 import { ArtsyMarkIcon } from "@artsy/palette"
 import styled from "styled-components"
 import { RouterLink, RouterLinkProps } from "System/Router/RouterLink"
-import { themeGet } from "@styled-system/theme-get"
 
 export const NavBarPrimaryLogo: React.FC<Omit<
   RouterLinkProps,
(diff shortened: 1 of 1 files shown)


dzucconi:
this needs a focus visible state

dzucconi:
just change to focus visible rather than making it inaccessible. there should be some examples to follow if u search

damassi:
Will do

damassi:
@dzucconi looking around this is what i see:

Screen Shot 2022-08-05 at 11 13 43 AM

Mind giving me an example of what i should do? Not sure

src/Components/NavBar/NavBarPrimaryLogo.tsx
```diff
@@ -24,12 +23,4 @@ const HitArea = styled(RouterLink)`
   > svg {
     box-sizing: content-box;
   }
-
-  &:focus {
```
dzucconi, line 28: Sorry: just switch this to `&.focus-visible {`

PR #10703, opened 2022-08-06
feat(TX-494): AlicanAkyuz/Checkout flow step transition and some refactoring

AlicanAkyuz:
The type of this PR is: **Feature / Enhancement**

This PR solves [TX-494]

### Description
This PR improves the transition from /payment to /review in the checkout flow for all payment methods, where several async operations are happening. For credit card, wire transfer and saved bank accounts, we only have a simple call to MP; for new bank accounts, we have several such as setting Stripe intent, redirect from stripe and polling account balance. 

While working on this task, I cleaned up/simplify `PaymentRoute` so that it's easier to work on that route in the future. The main clean ups include: 
- creating a single Save and Continue button to be used in all cases 
- moving the handling of Stripe's redirect to a dedicated hook 
- reducing prop drilling 

[TX-494]: https://artsyproduct.atlassian.net/browse/TX-494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

#### Video: setting payment with wire, existing and new credit card, and existing and new bank account

https://user-images.githubusercontent.com/42584148/183362588-6d558699-b29e-477b-b70e-81cd7e74d2dd.mov


src/Apps/Order/Routes/__tests__/Payment.jest.tsx
```diff
@@ -303,46 +310,6 @@ describe("Payment", () => {
       expect(page.text()).toContain("Bank transfer")
     })
 
-    it("tracks the initially-selected payment method on load like any other selection", async () => {
```
AlicanAkyuz, line 306: moved these tests down, inside 'tracking' section

src/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -13,28 +11,29 @@ import {
 } from "@artsy/palette"
 import { useSystemContext } from "System"
 import { useTracking } from "react-tracking"
-import { LoadingArea } from "../LoadingArea"
```
AlicanAkyuz, line 16: deprecated

src/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -13,28 +11,29 @@ import {
...
 import { InsufficientFundsError } from "Apps/Order/Components/InsufficientFundsError"
 import { preventHardReload } from "Apps/Order/OrderApp"
+import { SaveAndContinueButton } from "Apps/Order/Components/SaveAndContinueButton"
+import { getENV } from "Utils/getENV"
 
 interface Props {
   order: { mode: string | null; internalID: string }
-  returnURL: string
```
AlicanAkyuz, line 22: we don't need to pass returnURL as a prop, it can be defined and used in this component

src/Apps/Order/Components/SaveAndContinueButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export const SaveAndContinueButton: FC<{
+  onClick?: () => void
+  media?: { [key: string]: string }
+  loading?: boolean
+  disabled?: boolean
+}> = ({ onClick, media, loading = false, disabled = false }) => (
+  <>
+    {media ? (
+      <Media {...media}>
```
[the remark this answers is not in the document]
AlicanAkyuz: thanks for letting me know, didn't know. I'll check if it can be quickly removed in this PR.

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -218,9 +199,9 @@ const RadioWithLabel = styled(BorderedRadio)`
 `
 
 /*
-returns all available payment methods, by checking relevant feature flags
-TODO: when ACH and wire_transfer FFs is removed, this function can be removed and radios can be moved to PaymentContent
-*/
+ * returns all available payment methods in the form of BorderRadio
```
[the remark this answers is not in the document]
AlicanAkyuz: I'd assume yes, since payment methods are determined by the API now, we'll always need to check it per order (unless all methods are available for all orders at some point, which I don't see happening).

src/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -73,73 +77,61 @@ export const BankDebitForm: FC<Props> = ({
...
-                <InfoCircleIcon />
-              </Clickable>
-            </Tooltip>
-          </Flex>
+            <Clickable ml={0.5} style={{ lineHeight: 0 }}>
+              <InfoCircleIcon />
+            </Clickable>
+          </Tooltip>
         </Flex>
-        <Spacer mt={4} />
```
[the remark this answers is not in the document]
AlicanAkyuz: thank you for catching, I'm checking 🤔

araujobarret (commented):
The 2 binaries files added from .swc should be ignored, right?

src/Components/BankDebitForm/BankDebitForm.tsx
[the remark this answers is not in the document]
AlicanAkyuz: just pushed a commit with more consistent spacing, let me know if it still doesn't look right to you!

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -121,32 +129,29 @@ export const PaymentRoute: FC<Props> = props => {
         return
       }
 
-      const orderOrError = (
+      const error = (
         await setOrderPayment({
```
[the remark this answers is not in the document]
AlicanAkyuz: one thing we could do is to pull the order from the response and check against it, something like 

```
 const order = (
        await setOrderPayment({
          input: {
            paymentMethod: "CREDIT_CARD",
            paymentMethodId: result?.creditCardId,
            id: props.order.internalID!,
          },
        })
 ).commerceSetPayment?.orderOrError?.order
 
 if (!order.id) throw error
```

would that make it less confusing? I'm also open to other ideas if you have any.

src/Apps/Order/Components/SaveAndContinueButton.tsx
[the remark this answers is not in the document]
AlicanAkyuz: currently I don't see anything obvious to replace it with and the use of Media seems very common across the app. I'll leave it for now and tackle it in another PR.

src/Apps/Order/Routes/Payment/index.tsx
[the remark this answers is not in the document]
AlicanAkyuz: yup, this is how it was. I changed it back to that pattern, now knowing that it's clearer.

damassi:
@AlicanAkyuz - This bot is gonna yell until the PR title is formatted right:

```
feat(the thing): the title
```

(I think the brackets vs the parens are throwing it off)

src/Apps/Order/Components/SaveAndContinueButton.tsx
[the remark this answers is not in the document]
damassi: Yeah, getting rid of Media is going to be really tricky 😓. Until its ready to go feel free to still keep using it tho!

src/Apps/Order/Hooks/useStripePaymentBySetupIntentId.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+        if (error) throw error
+        setIsPaymentSetupSuccessful(true)
+      } catch (error) {
+        setIsPaymentSetupSuccessful(false)
+      }
+    }
+
+    // pull necessary params from Stripe redirect URL
+    const saveAccount = getClientParam("save_account")
```
damassi, line 44: For router consistency (and SSR support) we should always grab this off of the router itself, versus the window:
```tsx
const { match } = useRouter()

match.location... 
````

File: src/Apps/Order/Components/SaveAndContinueButton.tsx [hunk 9]
```diff
@@ -0,0 +1,38 @@
+import { FC } from "react"
+import { Button, Spacer } from "@artsy/palette"
+import { Media } from "Utils/Responsive"
+
+export const SaveAndContinueButton: FC<{
+  onClick?: () => void
+  media?: { [key: string]: string }
+  loading?: boolean
+  disabled?: boolean
+}> = ({ onClick, media, loading = false, disabled = false }) => (
```
damassi, line 10: As code is always returned to (and never truly complete!) we can be sure that in the future some dev will open this file and want to either expand it with a new feature, or console.log a pro, or any number of other things. This implicit return will make that much harder, requiring the dev to rewrite the format to have a (normal) explicit return. 

The convention is: if its a single liner, implicit returns are totally cool; think `const z = (x) => x * y`. 

Anything more than one line means there's some sufficient complexity therein and it will be a pain for someone else to have to unwind it later. 

There's a note about this in here: https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-implicit-returns-on-react-components

src/Apps/Order/Components/ProcessingPayment.tsx
```diff
@@ -0,0 +1,23 @@
+import { FC } from "react"
+import styled from "styled-components"
+import { Flex, Spinner, Text } from "@artsy/palette"
+
+export const ProcessingPayment: FC = () => (
```
damassi, line 5: Same here, no implicit returns pls 🙏 It looks and feels tidy, but it always ends up a maintainence burden over time

src/Apps/Order/Mutations/SetPaymentByStripeIntent.ts
```diff
@@ -0,0 +1,52 @@
+import { commitMutation, Environment, graphql } from "relay-runtime"
+import {
+  SetPaymentByStripeIntentMutation,
+  CommerceSetPaymentByStripeIntentInput,
+  SetPaymentByStripeIntentMutationResponse,
+} from "__generated__/SetPaymentByStripeIntentMutation.graphql"
+
+export const SetPaymentByStripeIntent = (
```
damassi, line 8: This mutation looks good, but can we use our `useMutation` hook and follow this pattern? (And try to spread the word :) https://github.com/artsy/force/blob/main/src/Apps/Settings/Routes/DeleteAccount/useDeleteAccount.ts

damassi (commented):
Just a couple minor comments 👍 

Given how complex the order app is and my total lack of context around additions / changes circa 2022 I might be way off here but I wanted to post as food for thought.

One thing that I keep thinking when I see PRs come across around this part of the app is that the quantity of state seems to be increasing fairly dramatically lately, with a lot of component level `setState` calls being added and state setter functions being passed around as props / threaded through the app.

Is it time to start thinking about another approach to handling state updates? I'm thinking specifically about `useReducer` and a react context. Something **centralized**. 

The general flow is that one always starts a new app with `useState`, and after a certain point in time when features start getting added, one migrates to something more orderly. This is sometimes hard to do (just in terms of a refactor), but short of this one runs the risk of digging a pretty deep hole for themselves in terms of managing complexity and making changes through time, due to 
- Individual components managing state (other devs can't figure out where something is being set), leading to harder to track down bugs, and other problems with scale 
- Props passing. Every time one passes a prop from component A to component B one binds component A to component B. As an app scales out, this coupling makes it difficult to make changes, move stuff around, etc, because there's always the expectation that the component needs a dependency (prop) to work. 
- The more this happens, the more imperative it becomes that a context is used, so that everything is centrally managed _there_ and not _locally_ within the component. This solves the bonding that occurs with props passing and makes it so that one can just use a hook (which is easily mocked in tests) to grab the props / state / setters and getters when needed.

In an ideal world, for anything that is sufficiently (and statefully) complex, the only props that are passed to a component are layout props such as margin and padding. Everything else is communicated through hooks, context and updated via `useReducer`. Then, when a new feature is added (meaning, more state), the dev can go directly to the context file and have a central place to make changes. No more playing thread the needle. 

Again, this is me just musing here, though I would suggest thinking about what organizing the state in the order app might look like. A bit of up front effort here could yield a very significant complexity reduction in the long run.

src/Apps/Order/Components/SaveAndContinueButton.tsx
AlicanAkyuz: makes sense, thanks for the heads up 👍

AlicanAkyuz:
@damassi I absolutely agree with your suggestion on having a central context and a redux-like state management circle in state-heavy apps. I would particularly like having centralised actions that are the only _pathway_ towards updating the context, simply because it takes me too much time to track and find what is updated where in complex apps like this.

I think the Order App is a good place to try that approach out, at least in an experimental PR and/or a demo app. my worry is that it'll break something in such a critical app (since the test setup would also need to be updated with context stuff, don't think they'll be reliable enough). 

I'll create a backlog ticket on TX board to try that out on a FF. (I would prefer not to do this experiment in this PR).

AlicanAkyuz:
#### Video after the changes
The major change (after talking to our design team) is that the 'Payment Processing...' copy is not displayed when user transitions for Shipping step to Payment step. 

https://user-images.githubusercontent.com/42584148/183655398-5627d910-8258-4a33-9e11-6f2beda1be3f.mov

AlicanAkyuz:
@rquartararo I have to bring it back for now (design team prefers to leave that part as it is).

damassi:
> my worry is that it'll break something in such a critical app (since the test setup would also need to be updated with context stuff, don't think they'll be reliable enough).

Yeah this is totally my concern too. The ship might have sailed here. 

That said, an incremental approach would be to start with a single route, get that centralized, then another, then another (but only if necessary -- the only reason to do this would be if there's a lot of state and props-passing; if not, no reason to refactor). Once a few individual routes moved into route-specific contexts, the final refactor pass would be to combine everything into one thing.

src/Apps/Order/Components/SaveAndContinueButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export const SaveAndContinueButton: FC<{
+  onClick?: () => void
+  media?: { [key: string]: string }
+  loading?: boolean
+  disabled?: boolean
+}> = ({ onClick, media, loading = false, disabled = false }) => {
+  return (
+    <>
+      <Spacer mb={4} />
```
damassi, line 13: Spacing stuff defined external to the component pls or passed in as a prop 🙏

damassi (approved):
🌟

PR #10705, opened 2022-08-08
feat(TX-493): AlicanAkyuz/Balance check for saved bank accounts

AlicanAkyuz:

This PR solves [TX-493]

### Description
This PR allows checking a previously saved bank account's balance before moving to the review step in the checkout flow. The feature is already in place for newly-added bank accounts; this PR utilises the same flow for saved bank accounts. 

#### Video with insufficient and sufficient balances

https://user-images.githubusercontent.com/42584148/183430187-34d05ca4-4aa8-407f-8954-0f4354abe513.mov

[TX-493]: https://artsyproduct.atlassian.net/browse/TX-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AlicanAkyuz:
currently I set the base as my other PR: https://github.com/artsy/force/pull/10703 

I'll change the base to main here once it's merged there.

src/Apps/Order/Components/InsufficientFundsError.tsx
```diff
@@ -4,6 +4,7 @@ import { Spacer, Message, Text } from "@artsy/palette"
 export const InsufficientFundsError: FC = () => {
   return (
     <>
+      <Spacer mt={2} />
```
damassi, line 7: As a general rule, layout spacing should always be defined external to the component; eg, 

```tsx
export const InsufficientFundsError: FC<BoxProps> = (props) => {
  return (
    <Box {...props}>...</Box>
  )
}
```

Then one can do

```tsx
<InsufficientFundsError mt={2} />
```

Or even simpler, where changes don't need to be made to this component: 

```tsx
<Spacer mt={2} /> 

<InsufficientFundsError /> 
```

PR #10733, opened 2022-08-10
feat: AlicanAkyuz/Fix connecting another account after insufficient funds error

AlicanAkyuz:

### Description
The PR fixes the issue that a user should be able to click on Save and Continue button after receiving an insufficient funds error on their initially-connected account. 

The solution here is to set the relevant error state to false when a new bank account connection is completed. 

#### Video 
https://user-images.githubusercontent.com/42584148/183929157-ff669b79-7408-4aa1-bb12-c4a81651ea8a.mov


src/Components/BankDebitForm/BankDebitForm.tsx
```diff
@@ -35,9 +37,11 @@ export const BankDebitForm: FC<Props> = ({
   const [isSaveAccountChecked, setIsSaveAccountChecked] = useState(true)
   const tracking = useTracking()
 
-  const trackPaymentElementEvent = event => {
+  const onPaymentElementChange = event => {
     const trackedEvents: any[] = []
```
AlicanAkyuz, line 41: PS: I don't think we need to create an array and push events to it, and then loop over it to track. If anyone has no objection, I'll remove it here: https://github.com/artsy/force/pull/10703

AlicanAkyuz:
merged this fix here: https://github.com/artsy/force/commit/978a5c60c3f4dfcbc5a914e0517c37c49fab6961

PR #10735, opened 2022-08-10
feat(onboarding): wire-up user onboarding flow analytics tracking

TMcMeans:

This PR solves [GRO-1153]

### Description

This is a follow up PR to wire up the analytics tracking function calls into the Onboarding app at specific stages for the user in the onboarding flow. I'll also be including tests in this PR. 

[GRO-1153]: https://artsyproduct.atlassian.net/browse/GRO-1153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Onboarding/Hooks/useOnboardingTracking.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  const userStartedOnboarding = useCallback(() => {
+    const payload: StartedOnboarding = {
+      action: ActionType.startedOnboarding,
+    }
+
+    trackEvent(payload)
+  }, [trackEvent])
+
+  const trackQuestionOne = useCallback(
```
damassi, line 24: Worth noting that `useCallback` on web is a performance optimization. If we're not seeing performance bottlenecks, then `useCallback` falls generally under the umbrella of premature optimization. 

(This is slightly different on mobile where every bit of performance matters -- noticed you adding this @laurabeth. But on web, its a need-to basis, otherwise it just adds unnecessary complexity.)

damassi: But its indeed harmless to add here either way.

TMcMeans: I think using the useCallback() here solves the problem that I'm seeing of multiple analytics calls firing at once, anywhere between 8-10 of the same call firing and its a bit strange?

TMcMeans: not a performance bottleneck, but strange behavior overall

laurabeth: I read that the rule of thumb here is that internal functions don't need useCallback, but handlers that are passed through to child components (excepting state handlers), or those that will be leveraged in dependency arrays, should be wrapped to avoid unnecessary rerenders. This situation seems to warrant it. 

I will try to find a citation for this so I'm not just blowing smoke. 😤

damassi: That reasoning seems correct @laurabeth - so the issue is the over-rendering. Good call with the `useCallback` then (and this is indeed a performance thing -- we don't want multiple renders trigging multiple function calls!)

laurabeth: I think the actual fix here (correct me if I'm wrong @TMcMeans) was also linked to the placement/trigger of the tracking call within the component. But, with hooks like this it always seems best to just optimize from the outset. That way if someone comes in and either uses the hook w/o check contents we minimize risk, or if they do look at the code and add to it via copy/pasta we're setting the best possible precedent. 🪙 🪙 <- 2 cents

damassi (approved):
Looks good!

PR #10736, opened 2022-08-10
fix(): Alican/Provide data-test id to button element directly instead of wrapper

AlicanAkyuz:

### Description
Currently our Integrity tests are [failing](https://app.circleci.com/pipelines/github/artsy/integrity/13620/workflows/884ea3e4-d687-4d16-845d-68bfd3a0890b/jobs/67584) because it looks for specific test Ids on button elements. This PR attempts to fix this issue by moving test IDs from the wrapper component `SaveAndCotinuebutton` to `Button` element directly.


araujobarret (commented):
thanks for fixing it 👍

PR #10750, opened 2022-08-12
feat(TX-493): Alican/Saved bank accounts balance check

AlicanAkyuz:

This PR solves [TX-493]

### Description

This PR brings account balance check for saved bank accounts (in addition to newly-link bank accounts). 

[TX-493]: https://artsyproduct.atlassian.net/browse/TX-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AlicanAkyuz:
I will squeeze in [tracking](https://artsyproduct.atlassian.net/browse/TX-568) into this PR, except for that one please feel free to review!

AlicanAkyuz:
@rquartararo thank you! 

I'll look into the 2nd issue now (incorrect account is selected after balance check error). 

For the 1st issue, I'm assuming that you did not get the balance at all, therefore you were able to move to the review step. (one way to test this is to check your call to commerceBalance in the network tab and see if we get balance at all or it's just null). let me know if you test it and still problematic.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -16,8 +17,11 @@ interface Props {
   me: BankAccountPicker_me
   bankAccountHasInsufficientFunds: boolean
   setBankAccountHasInsufficientFunds: (arg: boolean) => void
-  onBankAccountContinue: () => void
   setIsSavingPayment: (arg: boolean) => void
+  setBalanceCheckComplete: (arg: boolean) => void
+  setSelectedBankAccountId: (arg: string) => void
+  bankAccountSelection: BankAccountSelection
+  setBankAccountSelection: (arg: BankAccountSelection) => void
```
damassi, line 24: @AlicanAkyuz - this is non-blocking but worth noting, re these passed call-back functions:

The long-term React idiom for callback functions passed into a component is `onSomethingFoo` and local `useState` functions are `setSomethingFoo`

```
<MyComponent onBankBalanceCheckComplete={handleBankBalanceCheckComplete} />
```

A future dev reading this code will see these passed props and think that the state is locally set in this component, only to realize that what's actually being invoked is a callback function passed in as a prop. 

I'll comment further below to help explain the thinking...

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -26,24 +30,21 @@ export const BankAccountPicker: FC<Props> = props => {
...
-
-  interface BankAccountSelection {
-    type: BankAccountSelectionType
-    id?: string
-  }
 
   const { submitMutation: setPaymentMutation } = useSetPayment()
 
   const handleContinue = async () => {
+    setBalanceCheckComplete(false)
```
damassi, line 45: It's nice to see the consistency of the below sitting side by side

```
setBalanceCheckComplete(false)
setIsSavingPayment(true)
```

But there's a big difference between these two functions! One calls out _externally_ to the component and one is sets state _internally_. For an application _that lacks centralized state management_ to scale well over time crucial to maintain the distinctions between these two things at a glance. 

Why? Because one can be absolutely sure that the local `setIsSavingPayment` is being managed adequately here, where 
`setBalanceCheckComplete` -- being an external callback that's passed in -- is a wildcard. Was the right thing passed in at runtime to the prop? Is it defined? etc. Being a prop and external to the component, one can't control all of the details (even with TypeScript). So having idiomatic flags like `onBalanceCheckComplete` tells the future you that one needs to look at this piece of the code carefully, and in the future when for certain there will be a bug, a dev will know that this is external code executing, and we need to track it down.

damassi: (This comment applies to all other locations in the PR where `onActionFoo` callbacks are seemingly being mixed in as local state)

src/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -62,16 +68,22 @@ const PollAccountBalance: FC<PollAccountBalanceProps> = ({
       balanceCents >= buyerTotalCents && currencyCode === orderCurrencyCode
     )
 
-    onBalanceCheckComplete(!enoughBalance)
+    onBalanceCheckComplete(
```
damassi, line 71: Here, reading the code, i'm like "ok, this is external, what is this 👀 really doing 👀. It's importance is immediately elevated at a glance.

src/Apps/Order/Mutations/SetPaymentByStripeIntent.ts
```diff
@@ -0,0 +1,52 @@
+import { commitMutation, Environment, graphql } from "relay-runtime"
+import {
+  SetPaymentByStripeIntentMutation,
+  CommerceSetPaymentByStripeIntentInput,
+  SetPaymentByStripeIntentMutationResponse,
+} from "__generated__/SetPaymentByStripeIntentMutation.graphql"
+
+export const SetPaymentByStripeIntent = (
```
damassi, line 8: I've flagged this before, but want to flag again: We really need to stop writing custom mutations every time we mutate! Please use the `useMutation` hook as [seen here](https://github.com/artsy/force/blob/main/src/Apps/Settings/Routes/DeleteAccount/useDeleteAccount.ts)

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -184,22 +227,50 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
   }
 
-  // pushes to the review step, when payment is set with an existing bank account
-  const onBankAccountContinue = () => {
-    props.router.push(`/orders/${props.order.internalID}/review`)
-  }
-
   // fired when balance check is done: either sets error state or moves to /review
-  const onBalanceCheckComplete = (displayInsufficientFundsError: boolean) => {
+  const onBalanceCheckComplete = (
```
damassi, line 231: I know this code was here before, but its important to call out for clarity sake. Idiomatically this form is also incorrect per long-term patterns in the react community: 

Local functions that _handle_ actions are always prefixed with `handle`, eg: `handleClick`:

```tsx
const Foo = () => {
  const handleClick = () => {
    ... 
  }
  
  return <Box onClick={handleClick} />
}
```

Or even

```tsx
const Foo = () => {
  const handleClick = () => {
    ... 
  }
  
  return <Bar onClick={handleClick} />
}

const Bar = ({ onClick }) => {
  const [isLoading, setIsLoading] = useState(false) 
  
  const handleClick = () => {
      setIsLoading(true)
      onClick?.() 
  }
  
  return <Box onClick={handleClick}
}
```

The rule is: 
- callback functions passed _down_ to components are defined as `onActionFoo` (think: `onClick`, `onMouseDown`, etc -- old patterns well defined in js community)
- local functions that _handle_ callback related actions are prefixed with `handleActionFoo`

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -184,22 +227,50 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
+
     if (displayInsufficientFundsError) {
       setBankAccountHasInsufficientFunds(true)
       return
     }
+
+    onPaymentStepComplete()
+  }
+
+  const onPaymentStepComplete = () => {
```
damassi, line 259: `onPaymentStepComplete` -> `handlePaymentStepComplete`

Not trying to be pedantic with these naming things! Since React is a functional programming paradigm one of the most crucial things one can do is learn to describe the direction of data flow within the system (this goes for all functional langs). 

- Is the data flow going down? `onFoo={handleFoo}` 
- Is the data being invoked from above? `onFoo()` 
- Is data being invoked locally? `handleFoo()` 
- Is data being set locally? `setFoo()` 
- Is data being set externally from above? `setFoo()` 
- Are we passing a setter down so that it's being invoked below the parent? `onSetFoo={setFoo}` which is then in the child component calling back up to the parent? `onSetFoo(true)`

See how just by naming things consistently one can "see" at a glance exactly which way the data is flowing through a system? These rules only grow in importance the less centralized the state management is.

damassi (commented):
Very nice PR @AlicanAkyuz 👍 I left a handful of detailed (non-blocking) comments around idiomatic React stuff; take em or leave em! Just good to know kinda things.

src/Apps/Order/Mutations/SetPaymentByStripeIntent.ts
AlicanAkyuz: this is my bad to forget to delete it! I actually have it [with useMutation](https://github.com/artsy/force/blob/main/src/Apps/Order/Mutations/useSetPaymentByStripeIntentMutation.ts) and we are [using that](https://github.com/artsy/force/blob/main/src/Apps/Order/Hooks/useStripePaymentBySetupIntentId.tsx#L11) one. 

This custom mutation is not used at all, so I'll just get rid of it

AlicanAkyuz:
woo @damassi, thanks for those comments about conventions, really appreciate it and it totally makes sense to have that separation with `handle`, `on` and `set` prefixes.

damassi:
Really appreciate you going through to update those call-sites @AlicanAkyuz 🙏 Nice work on this 👍

PR #10751, opened 2022-08-12
fix(CX-2744): Fix design issues in My Collection

MrSltun:

This PR solves [CX-2744]

### Description
This PR updates font sizes in the Settings app, and adds a "NoImageIcon" when there are no images for an artwork inside my collection, plus fixes a URL to point to `/my-collection/artwork/id` instead of `/artwork/id`

image

[CX-2744]: https://artsyproduct.atlassian.net/browse/CX-2744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/Metadata.tsx
```diff
@@ -92,6 +103,7 @@ export default createFragmentContainer(Metadata, {
   artwork: graphql`
     fragment Metadata_artwork on Artwork {
       ...Details_artwork
+      internalID
```
MrSltun, line 106: this change caused all the query updates 😅

olerichter00: We could also pass `to` as a prop instead of checking inside of `Metadata` if the component is rendered in the My Collection artwork screen. Like this, we would keep the component more general.

src/Components/Artwork/GridItem.tsx
```diff
@@ -96,12 +96,27 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
...
-              maxWidth="100%"
-            />
+            <>
+              <ResponsiveBox
+                aspectWidth={4}
+                aspectHeight={3}
+                position="relative"
+                maxWidth="100%"
+              />
+              {!!isMyCollectionArtwork && (
```
olerichter00, line 106: I wonder if this could be the default behavior of this component.  Removing the `isMyCollectionArtwork`check shouldn't affect other artwork grids because all artworks outside of My Collection should have images.

olerichter00 (commented):
Great work! I've added a few suggestions we can discuss.

src/Components/Artwork/Metadata.tsx
```diff
@@ -20,6 +20,7 @@ export interface MetadataProps
   isHovered?: boolean
   showHoverDetails?: boolean
   showSaveButton?: boolean
+  isMyCollectionArtwork?: boolean
```
MounirDhahri, line 23: can you add some tests to this file

MounirDhahri (approved):

PR #10756, opened 2022-08-15
feat: further normalize to improve gallery location filter matches 

dblandin:
Jira: [FX-4161](https://artsyproduct.atlassian.net/browse/FX-4161)

The /galleries page currently has a location filter to narrow down displayed galleries.

The filter is backed by lists of cities generated by Gravity and exposed via Metaphysics. The filtering currently takes place client-side using a simple string lowercase / include mechanism.

This is not inclusive of cities whose names contain accented characters. When searching for Bogotá or Périgueux, if you replace accented characters when unaccented variation, the expected results do not show up. Short of switching to a completely different back-end implementation, I've opted for simpler client-side patch here by further normalizing the characters.

The use of `String.normalize` seems widely supported in browsers today: https://caniuse.com/mdn-javascript_builtins_string_normalize

## Screen Recordings

**Before**

**After**


damassi:
Nice! I wasn't aware of the `normalize` method.

PR #10759, opened 2022-08-15
feat(CX-2707): My Collection artwork details - Image viewer

MrSltun:

This PR solves [CX-2707]

### Description
This PR adds the image viewer to My Collection artworks details 

### Screenshots

|  | One Image | Multiple Images | No Images |
| -- | -- | -- | -- |
| Mobile | image | image | image |
| Tablet | image | image | image |
| Desktop | image | image | image |

[CX-2707]: https://artsyproduct.atlassian.net/browse/CX-2707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi:
@MrSltun - note that the Danger warning is coming from the formatting of the PR title - its expecting something that look like this:

`type(context): The description`. (It doesn't like the brackets.0

src/Apps/MyCollection/Routes/MyCollectionArtwork/MyCollectionArtwork.tsx
```diff
@@ -19,18 +20,18 @@ const MyCollectionArtwork: React.FC<MyCollectionArtworkProps> = ({
...
-        <MyCollectionArtworkMetaFragmentContainer artwork={artwork} />
-        <GridColumns>
-          <Column span={8}>
-            <Box width={[335, 780]} height={[320, 870]} bg="black5" />
-          </Column>
+      <MyCollectionArtworkMetaFragmentContainer artwork={artwork} />
+
+      <GridColumns gridRowGap={[4, null]} pt={[2, 6]}>
+        <Column span={8}>
+          <MyCollectionArtworkImageBrowserFragmentContainer artwork={artwork} />
```
olerichter00, line 27: Can we just use the same component we use on the artwork page? This would be great because we wouldn't have to maintain two different components.


```suggestion
          <ArtworkImageBrowserFragmentContainer artwork={artwork} />
```

MounirDhahri (dismissed):
LGTM

src/Apps/MyCollection/Routes/MyCollectionArtwork/MyCollectionArtwork.tsx
MrSltun: I like your idea of reusing the current implementation of the image browser, but we need to update a few things in the ArtworkImageBrowser to match the design and looks for My Collection Artworks

olerichter00: If the image browser design is different on the "Artwork" page and the "My Collection Artwork" page, I would first question if it is necessary to have these two designs. IMHO we should keep the designs between the two pages as consistent as possible.

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkImageBrowser/__tests__/MyCollectionArtworkImageBrowser.jest.tsx
```diff
@@ -0,0 +1,75 @@
...
+            width: 800,
+            height: 600,
+            src: "image.jpg",
+            srcSet: "image.jpg 1x",
+          }),
+        },
+        false
+      )
+
+      expect(screen.getByTestId("artwork-lightbox-box")).toHaveStyle(
```
damassi: Some are of the opinion that using `getByTestId` should be avoided, but I personally very much appreciate the unambiguous simplicity of it and try to encourage its use (over other more general methods)

src/Apps/MyCollection/Routes/MyCollectionArtwork/MyCollectionArtwork.tsx
damassi: @MrSltun - Agree with @olerichter00 here and we should def return to this in a follow-up PR! This is the sort of thing that Engineering and Design need to work together on, rather than just blindly building out designs that the design team hands over, duplicating stuff, using extra time, and creating tech debt. 

Engineers have a lot more inside knowledge of the app than the design team. Not only because we're always poking around different areas of the app, but also because there's been so much turn over on design, and new folks don't know whats what or what patterns are available. Often times it's up to us to inform them.  

As an engineer its often hard to think like a designer, but I would encourage you to use your power here and if you see something that could be consolidated into one, or if something else seems off, don't hesitate to push back and encourage design to re-review and try to incorporate your feedback. Product designers are like engineers and they too like consistency; it's mostly just a knowledge issue.

PR #10778, opened 2022-08-17
fix(onboarding): add data-test id for integration testing

TMcMeans:
The type of this PR is: fix

### Description

This PR is part of the work that fixes the integration tests failing in integrity. Adding a `data-test-id` helps select the skip button in cypress tests. 

### DIFF
--- a/src/Components/Onboarding/Views/OnboardingWelcome.tsx
+++ b/src/Components/Onboarding/Views/OnboardingWelcome.tsx
@@ -61,6 +61,7 @@ export const OnboardingWelcome = () => {
               as={RouterLink}
               to="/"
               onClick={onClose}
+              data-test="onboarding-skip-button"
             >
               Skip
             </Button>


laurabeth (approved):
🔥

src/Components/Onboarding/Views/OnboardingWelcome.tsx
```diff
@@ -61,6 +61,7 @@ export const OnboardingWelcome = () => {
               as={RouterLink}
               to="/"
               onClick={onClose}
+              data-test="onboarding-skip-button"
```
damassi, line 64: (totally no blocking trivial comment): worth noting that in a JS world there's no need to use hyphens in variable values and they're pretty foreign syntax-wise (and mostly a legacy html idiom). `data-test="onboardingSkipButton"` works just as well, matches rest of the code

TMcMeans: hey, fixing the integration tests now so ill update on a follow-up pr addressing new integration tests for onboarding flow.

damassi: no need to follow up! just a side comment for the future

PR #10790, opened 2022-08-18
chore: run detect-secrets in CI without recovery logic

ovasdi:

### Description

Back in the fall of 2021, we [expanded the run command](https://github.com/artsy/force/pull/8648) to recover from '_random_' baseline mutations. This was necessary (at the time) as it was causing developer friction and failed builds in CI.

Seeing as the tool has had a couple of releases since then and some projects ([eigen](https://github.com/artsy/eigen/blob/main/scripts/secrets-check-all)) do not experience the same issue, I wanted to remove the conditional recovery logic and verify if the extra logic is still necessary.

I picked Force to try this out on because:

1. Force was one of the projects that had this issue in the past.
2. Force has a lot of daily builds so if the recovery logic is still necessary, the problem should surface (hopefully) relatively quickly.

Following this change, if we notice that detect-secrets step stars failing randomly we will revert this PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,27 +213,7 @@ jobs:
     working_directory: /usr/src/app
     steps:
       - checkout
-      - run: |
-          cp .secrets.baseline /tmp/.secrets.baseline
-          set +e && set +o pipefail
-          detect-secrets-hook --baseline .secrets.baseline $(git ls-files)
-          code=$?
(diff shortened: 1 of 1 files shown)


damassi (approved):
Lets give it a shot 👍

damassi:
(About to update circle config file so gonna merge!)

PR #10792, opened 2022-08-18
fix(onboarding): handle /signup case slightly better

damassi:

### Description

This updates how we handle onboarding redirects from the static `artsy.net/signup` page, instead relying on internal auth form logic rather than setting a fallback on `AUTHENTICATION_REDIRECT_TO` (which broke a few integrity tests). 

Something that we very much need to follow up on once things have settled is refactoring the onboarding auth flow logic to have a formalized onboarding launch mechanism, versus us piggy-backing on `redirectTo` which is brittle and opaque. It works well enough right now, but the case in this PR illustrates why it won't scale: as soon as we have a signup example that _doesn't_ use `redirectTo` things break; a future dev will certainly not know why. (Filed a ticket [here](https://artsyproduct.atlassian.net/browse/GRO-1239).)

cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Authentication/Utils/__tests__/helpers.jest.ts
+++ b/src/Apps/Authentication/Utils/__tests__/helpers.jest.ts
@@ -4,6 +4,7 @@ import {
   handleSubmit,
   maybeUpdateRedirectTo,
   setCookies,
+  updateURLWithOnboardingParam,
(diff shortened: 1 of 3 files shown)


laurabeth (approved):
Nice! Thanks so much for following up on this.

PR #10822, opened 2022-08-23
feat(CX-2745): Add Upload Photos CTA on MyC artwork without images

MrSltun:

This PR solves [CX-2745]

### Description
This PR adds the Upload Photos CTA when an artwork in My Collection is without any photos
This change is hidden behind `my-collection-web-phase-3` feature flag and the URL sends the user the edit artwork page

### Screenshot
| Desktop | Tablet | Mobile |
| -- | -- | -- |
| image | image | image|

### Follow up
Update the URL to the upload photos link cc @olerichter00 

[CX-2745]: https://artsyproduct.atlassian.net/browse/CX-2745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olerichter00 (commented):
Good work so far!

I wonder if there is still a case where we would show the "No Image" image 🤔 

If not, I'd move the logic whether to show the images or the "Upload image" CTA outside of the image artwork browser into the component `MyCollectionArtwork`. The reason here is separation of concerns. The image browser component should not include "My Collection"-related logic if possible and only show images.

MrSltun:
> Good work so far!
> 
> I wonder if there is still a case where we would show the "No Image" image within the image browser 🤔
> 
> If not, I'd move the logic whether to show the images or the "Upload image" CTA outside of the image artwork browser into the component `MyCollectionArtwork`. The reason here is separation of concerns. The image browser component should not include "My Collection"-related logic if possible and only show images.
I agree with that idea, let's maybe create an image browser for my collection that handles the no image/upload photos CTA logic when it doesn't have photos, and uses the artwork image browser when there are images

olerichter00:
Not sure if we really need another component to handle both cases. Couldn't we just render the image browser or the cta conditionally in `MyCollectionArtwork` like this?

```
{artwork.images.length ? (
            <ArtworkImageBrowserFragmentContainer
              artwork={artwork}
              isMyCollectionArtwork
            />
          ) : (
            <ArtworkNoImageComponent />
          )}
```

MrSltun:
> Not sure if we really need another component to handle both cases. Couldn't we just render the image browser or the cta conditionally in `MyCollectionArtwork` like this?
> 

Currently, what I did is change the icon with the button, so if I need to change it, I would need to copy the gray background (Image placeholder) into the new component to keep the code simple (that's my approach)

olerichter00:
That would work. I mean the approach that's implemented right now works fine, but intuitively I'd move all the code that is related to this "Upload Photo" CTA (or the broken image icon, which we can remove) to the My Collection app because it's highly coupled to the My Collection artwork page and not really useful anywhere else.

MounirDhahri (commented):
LGTM, @olerichter00's suggestion here also makes sense to me. If the component was only used inside my collection, the ImageBrowser would be the right place for it, the issue here is that the same component is used everywhere in Artsy and I am afraid if we don't pay attention to decoupling it, we could end up in few months having a component that's tightly coupled to many others and then it suddenly does lots of stuff and we better keep it as simple as possible and doing only one job

olerichter00:
@MrSltun that's exactly my point and that is why I would **not** render the `NoImageComponent` inside of the `ImageBrowser` component and instead render it in `MyCollectionArtworkImage` like this:

```
{artwork.images.length ? (
  <ArtworkImageBrowserFragmentContainer
    artwork={artwork}
    isMyCollectionArtwork
  /> ) : (
  <ArtworkNoImageComponent /> 
  )
}
```

After this change, the `ImageBrowser` component would be in the same state it has been before we started developing My Collection for web.

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkImageBrowser/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              aspectWidth={1}
+              aspectHeight={1}
+            >
+              <MyCollectionArtworkNoImageComponent
+                artworkID={artwork.internalID}
+              />
+            </ResponsiveBox>
+          </Flex>
+        </Media>
+      </>
```
olerichter00, line 53: This should work as well: 

```suggestion
      <Flex maxWidth={["100%", 600]} mx="auto">
        <ResponsiveBox
          data-testid="artwork-browser-no-image-box"
          bg="black10"
          mx={[0, 2, 4]}
          maxWidth="100%"
          aspectWidth={1}
          aspectHeight={1}
        >
          <MyCollectionArtworkNoImageComponent artworkID={artwork.internalID} />
        </ResponsiveBox>
      </Flex>
```

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkImageBrowser/MyCollectionArtworkImageBrowser.tsx
```diff
@@ -0,0 +1,43 @@
+import { AddCircleIcon, Button, Flex, NoImageIcon } from "@artsy/palette"
+import { RouterLink } from "System/Router/RouterLink"
+import { useFeatureFlag } from "System/useFeatureFlag"
+
+interface MyCollectionArtworkNoImageComponentProps {
+  artworkID?: string
+}
+
+export const MyCollectionArtworkNoImageComponent: React.FC<MyCollectionArtworkNoImageComponentProps> = ({
```
olerichter00, line 9: Let's rename this file to `MyCollectionArtworkNoImageComponent.tsx`.

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkImageBrowser/index.tsx
olerichter00, line 53: This should work as well:

```suggestion
      <Flex maxWidth={["100%", 600]} mx="auto">
        <ResponsiveBox
          data-testid="artwork-browser-no-image-box"
          bg="black10"
          mx={[0, 2, 4]}
          maxWidth="100%"
          aspectWidth={1}
          aspectHeight={1}
        >
          <MyCollectionArtworkNoImageComponent artworkID={artwork.internalID} />
        </ResponsiveBox>
      </Flex>
```

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkImageBrowser/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { createFragmentContainer } from "react-relay"
+import { graphql } from "relay-runtime"
+import { Media } from "Utils/Responsive"
+import { MyCollectionArtworkImageBrowser_artwork } from "__generated__/MyCollectionArtworkImageBrowser_artwork.graphql"
+import { MyCollectionArtworkNoImageComponent } from "./MyCollectionArtworkImageBrowser"
+
+interface MyCollectionArtworkImageBrowserProps {
+  artwork: MyCollectionArtworkImageBrowser_artwork
+}
+const MyCollectionArtworkImageBrowser: React.FC<MyCollectionArtworkImageBrowserProps> = ({
```
olerichter00, line 12: And maybe we can rename this file to `MyCollectionArtworkImageBrowser`. I'm not quite sure if we still use `index.tsx` files in **Force** for components. In **Eigen** we try to avoid them ([see RFC](https://github.com/artsy/eigen/pull/6986)).

olerichter00 (commented):
Looking good! I've added a few suggestions regarding file names and a suggestion to decomplexify the code.

dariakoko (dismissed):
LGTM! Thank you Sultan

olerichter00 (dismissed):

PR #10830, opened 2022-08-24
chore(grow): Address misc tickets

damassi:

### Description

This PR solves
- SEO OpenGraph images on home and about https://artsyproduct.atlassian.net/browse/GRO-1207
- Update NavBar with new NFT link: https://artsyproduct.atlassian.net/browse/GRO-1186 
- Persist Editorial in NavBar: https://artsyproduct.atlassian.net/browse/GRO-1233 
- Only show /auctions WFY tab if following artist from auction: https://artsyproduct.atlassian.net/browse/GRO-1192

cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Auctions/Components/CuritorialRailsTabBar.tsx
+++ b/src/Apps/Auctions/Components/CuritorialRailsTabBar.tsx
@@ -15,12 +15,11 @@ interface CuritorialRailsTabBarProps {
 export const CuritorialRailsTabBar: React.FC<CuritorialRailsTabBarProps> = ({
   viewer,
 }) => {
-  const isFollowingArtists = !!viewer.me?.followsAndSaves?.artistsConnection
-    ?.totalCount!
+  const showWorksForYouTab = !!viewer.followedArtistsInAuction?.counts?.total
 
   return (
     <Tabs mb={4}>
-      {isFollowingArtists && (
+      {showWorksForYouTab && (
         <Tab name="Works For You">
           <Join separator={<Spacer mt={2} />}>
             <MyBidsFragmentContainer me={viewer.me!} />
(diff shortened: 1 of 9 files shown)


src/System/Router/interceptLinks.ts
```diff
@@ -12,9 +12,10 @@ interface InterceptLinksProps {
  */
 export function interceptLinks({ router, routes }: InterceptLinksProps) {
   catchLinks(window, href => {
+    const [pathname, _search] = href.split("?")
     // FIXME: PR upstream; `matchRoutes` is missing from type definition
     // @ts-ignore
-    const foundUrl = router.matcher.matchRoutes(routes, href)
+    const foundUrl = router.matcher.matchRoutes(routes, pathname)
```
damassi, line 18: Finally got around to fixing this long-standing issue. 

Before, what would happen is that our `<RouterLink to='/some-path?foo=bar'>` link would perform hard jumps between pages because the additional query param wouldn't match against our routes, and return false (we don't abstractly define query params in our routes definitions). And so this fixes that by splitting on the query param and matching _only_ against the pathname. Quick and easy

damassi, line 18: Finally got around to fixing this long-standing issue. 

Before, what would happen is that our `<RouterLink to='/some-path?foo=bar'>` link would perform hard jumps between pages because the additional query param wouldn't match against our routes, and return false (we don't abstractly define query params in our routes definitions). And so this fixes that by splitting on the query param and matching _only_ against the pathname. Quick and easy

src/Apps/Auctions/Components/CuritorialRailsTabBar.tsx
```diff
@@ -47,14 +48,22 @@ export const CuritorialRailsTabBarFragmentContainer = createFragmentContainer(
...
-        me {
-          ...MyBids_me
-          followsAndSaves {
-            artistsConnection {
-              totalCount
+        followedArtistsInAuction: saleArtworksConnection(
+          includeArtworksByFollowedArtists: true
+          isAuction: true
+          liveSale: true
+          first: 1
```
laurabeth, line 55: non-blocking: `first: 0` fetches metadata without having to return the objects in the list. Maybe it would be good practice to leverage this pattern here to return only the count? 

```graph
saleArtworksConnection(includeArtworksByFollowedArtists: true, first: 0, isAuction: true, liveSale: true) {
		counts {
			total
		}
	}
```

laurabeth (dismissed):
One note, but overall looks good. ✨

src/Apps/Auctions/Components/CuritorialRailsTabBar.tsx
damassi: Oh that's a great idea and I always forget about this. Will do 👍

src/System/Router/interceptLinks.ts
laurabeth: Awesome! This is a big win. 🎉

PR #10831, opened 2022-08-25
fix(TX-619)- Do not send bankAccountId as null

AlicanAkyuz:

This PR solves [TX-619]

### Description

This PR fixes the problem that when account balance check polling receives null initially and then the actual balance value, it still allows users to proceed to the review step, even though the received value is less than the order value. 

The fix is to send bankAccountId as a string explicitly instead of defaulting to null, even when we are using setupIntentId for a newly added bank account. 

Note that this problem is surfaced only since yesterday, when we added bankAccountId to the query for already saved bank account. 

#### Video
https://user-images.githubusercontent.com/42584148/186608637-45dc30d6-9809-4fca-ada7-06e60dc1423c.mov

[TX-619]: https://artsyproduct.atlassian.net/browse/TX-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/PollAccountBalance.tsx
+++ b/src/Apps/Order/Components/PollAccountBalance.tsx
@@ -125,7 +125,9 @@ export const PollAccountBalanceQueryRenderer: FC<PollAccountBalanceQueryRenderer
     <SystemQueryRenderer<PollAccountBalanceQuery>
       environment={relayEnvironment}
       placeholder={<SavingPaymentSpinner />}
-      variables={bankAccountId ? { bankAccountId } : { setupIntentId }}
+      variables={
+        bankAccountId ? { bankAccountId } : { setupIntentId, bankAccountId: "" }
+      }
       query={BALANCE_QUERY}
       render={({ props }) => {
         if (!props?.commerceBankAccountBalance) {
@@ -134,7 +136,7 @@ export const PollAccountBalanceQueryRenderer: FC<PollAccountBalanceQueryRenderer
 
         return (
           <PollAccountBalanceRefetchContainer
-            commerceBankAccountBalance={props?.commerceBankAccountBalance}
+            commerceBankAccountBalance={props.commerceBankAccountBalance}
             setupIntentId={setupIntentId}
             bankAccountId={bankAccountId}
             {...rest}


src/Apps/Order/Components/PollAccountBalance.tsx
```diff
@@ -125,7 +125,9 @@ export const PollAccountBalanceQueryRenderer: FC<PollAccountBalanceQueryRenderer
     <SystemQueryRenderer<PollAccountBalanceQuery>
       environment={relayEnvironment}
       placeholder={<SavingPaymentSpinner />}
-      variables={bankAccountId ? { bankAccountId } : { setupIntentId }}
+      variables={
+        bankAccountId ? { bankAccountId } : { setupIntentId, bankAccountId: "" }
```
starsirius, line 129: Interesting! Can you help me understand why that makes a difference? Also, if there a good way to test this?

AlicanAkyuz: thank you for merging 🙏  

I'm not entirely sure on how it solves the problem. here is my findings if you are interested: 

Screenshot 2022-08-26 at 14 34 17
Screenshot 2022-08-26 at 14 34 26

these ⬆️  screenshots are from the previous version (`variables={bankAccountId ? { bankAccountId } : { setupIntentId }})`. I realised that when we pass `setupIntentId` as a param, `bankAccountId` would be null in the 1st poll and and empty string in the 2nd and the following polls. it shouldn't be! this messed up the refetch container that we use in the polling. 

the reason for this is that in Payment/index.tsx, we are rendering `PaymentContent` with blocked display: 

```
{displayLoading && <SavingPaymentSpinner />}
              {/* keep PaymentContent mounted but displayed none while 
                  displayLoading is true; needed to handle Stripe redirect */}
              <Flex
                flexDirection="column"
                style={displayLoading ? { display: "none" } : {}}
              >
                <PaymentContent
                  commitMutation={props.commitMutation}
                  paymentMethod={selectedPaymentMethod}
                  me={props.me}
                  ...
```

we needed to do that in order to display loading spinner and keep `PaymentElement` mounted while still communicating with Stripe. but this also caused the problem about `bankAccountId` initially being null and then empty string. when we get the redirect back from stripe, `PaymentContent` is re-rendered but not display. when it's re-rendered, it re-renders BankAccountPickerFragmentContainer (the children of Collapse component are rendered but not displayed even if its open property is false). _I think_ this was the root of the problem. with the change, we are explicitly passing `bankAccountId` as empty string when we pass `setupIntentId`.

AlicanAkyuz: for manual testing, I have followed your suggestion in ticket (changing the definition of balance freshness in Exchange for testing different cases). thank you so much for that clever idea! 

for automated tests, I haven't seen any examples of testing refetch container with polling logic in Force, so I'm a bit unsure about how to go about it. if it's desirable though, I would have some bandwidth to work on creating a test suite for this.

starsirius: Thanks for the pointer! Then I was curious about why the other scenario (polling by bankAccountId) does not affected by this. I Googled a bit and found [this](https://github.com/facebook/relay/issues/2244#issuecomment-355054944) that seems relevant. Let me explain my theory and see if it makes sense to you! (disclaimer: I'm not familiar with Relay so I could be totally wrong.)

- First, the empty data for `setupIntentId` and `bankAccountId` is different. One is `null`, the other is `""`. The variables used in `relay.refetch` are different
  ```javascript
  { setupIntentId: '123', bankAccountId: '' }   // when polling by setupIntentId
  { setupIntentId: null, bankAccountId: '456' } // when polling by bankAccountId
  ```
- Before this fix, the variables that the parent QueryRenderer only contains non-blank one. So
  ```javascript
  { setupIntentId: '123' } // when polling by setupIntentId
  { bankAccountId: '456' } // when polling by bankAccountId
  ```
- The `null` vs. `""` may have different implication when it comes to subscriptions to the refetch query in the refetch container:
  * When the parent container has variables `{ setupIntentId: '123' }` and the refetch query uses variables `{ setupIntentId: '123', bankAccountId: '' }` , the parent container is _not_ subscribed to the changes from the refetch query. My interpretation: it's saying that "refetch query with variables `{ setupIntentId: '123', bankAccountId: '' }` would not affect components with variables `{ setupIntentId: '123' }`. So it's not passed with refetched data.
  * When the parent container has variables `{ bankAccountId: '456' }` and the refetch query uses variables `{ setupIntentId: null, bankAccountId: '456' }` , the parent container _is_ subscribed to the changes from the refetch query. My interpretation: it's saying that "refetch query with variables `{ setupIntentId: null, bankAccountId: '456' }` _would_ affect components with variables `{ bankAccountId: '456' }`. So it's passed with refetched data.

I wonder if always making keys in the variables of the [parent QueryRenderer](https://github.com/artsy/force/blob/0575667b106af3e438b9b363fb30a18e00135bf6/src/Apps/Order/Components/PollAccountBalance.tsx#L128-L130) and the [refetch query](https://github.com/artsy/force/blob/0575667b106af3e438b9b363fb30a18e00135bf6/src/Apps/Order/Components/PollAccountBalance.tsx#L39) match would be recommended to address this problem. So something like
```jsx
<SystemQueryRenderer<PollAccountBalanceQuery>
  placeholder={<SavingPaymentSpinner />}
  variables={{ setupIntentId, bankAccountId }}
  // ...
>

// in refetch container
relay.refetch({ setupIntentId, bankAccountId }, //...)
```

Relay is hard! I could be totally misunderstand things, and thank you for reading!

PR #10732, opened 2022-08-10
feat: My Collection artwork upload flow

olerichter00:
The type of this PR is: **Feature**

This PR solves [CX-2697](https://artsyproduct.atlassian.net/browse/CX-2697)

#### To Do

- [ ] Add tests

### Description

This PR implements the My Collection artwork upload flow (without image upload).

| Web | mWeb |
| --- | --- |
|Screenshot 2022-08-16 at 11 07 30| Screenshot 2022-08-16 at 11 07 40|

[Figma](https://www.figma.com/file/AF7eVKe0eErR5gB6DDRiB2/My-Collection-Web?node-id=607%3A21642)

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10834, opened 2022-08-25
feat: delete artwork flow in edit artwork

dariakoko:

This PR solves [CX-2757]

### Description

Added delete artwork flow in edit artwork

| Web | Mobile |
| --- | --- |
| image | image |

+ corrected the layout of the modal -- removed extra margin
image

[CX-2757]: https://artsyproduct.atlassian.net/browse/CX-2757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
@@ -127,7 +127,6 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
                 ? `/my-collection/artwork/${artwork.internalID}`
                 : "/my-collection"
             }
-            mt={4}
             width="100%"
             data-testid="leave-button"
           >
@@ -144,7 +143,7 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
         </>
       }
     >
-      <Text mt={4}>
+      <Text>
         {isEditing
           ? "Changes you have made so far will not be saved."
           : "Your artwork will not be added to My Collection."}
@@ -179,7 +178,6 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
         <>
           <Button
             onClick={handleDelete}
-            mt={4}
             width="100%"
             data-testid="submit-delete-button"
           >
@@ -196,7 +194,7 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
         </>
       }
     >
-      <Text mt={4}>This artwork will be removed from My Collection.</Text>
+      <Text>This artwork will be removed from My Collection.</Text>
     </ModalDialog>
   )
 
--- a/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
@@ -31,7 +31,7 @@ import {
   MyCollectionArtworkDetailsValidationSchema,
   validateArtwork,
 } from "./Utils/artworkValidation"
-import { useCreateOrUpdateArtwork } from "./Utils/useCreateOrUpdateArtwork"
+import { useCreateOrUpdateOrDeleteArtwork } from "./Utils/useCreateOrUpdateOrDeleteArtwork"
 
 const logger = createLogger("MyCollectionArtworkForm.tsx")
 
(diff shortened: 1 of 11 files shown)


src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -106,6 +112,92 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
...
+          : "Your artwork will not be added to My Collection."}
+      </Text>
+    </ModalDialog>
+  )
+
+  const handleDelete = async () => {
+    try {
+      await deleteArtworkRequest({
+        artworkId: artwork?.internalID!,
+      }).then(() => router.push({ pathname: "/settings/my-collection" }))
```
olerichter00, line 158: **(suggestion):** If you're already using `await` here, you could get rid of the callback function like this:

```suggestion
      await deleteArtworkRequest({
        artworkId: artwork?.internalID!,
      })
      
      router.push({ pathname: "/settings/my-collection" })
```

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -106,6 +112,92 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
...
+        variant: "error",
+        message: "An error occurred",
+        description: "Please contact support@artsymail.com",
+      })
+
+      return
+    }
+  }
+
+  const deletionModal = (
```
olerichter00, line 172: What do you think about moving both Modals into their own files? This file is getting pretty long....

src/Apps/MyCollection/Routes/EditArtwork/Utils/useCreateOrUpdateOrDeleteArtwork.ts
```diff
@@ -1,15 +1,18 @@
...
 import { useCreateArtwork } from "../Mutations/useCreateArtwork"
 import { useUpdateArtwork } from "../Mutations/useUpdateArtwork"
+import { useDeleteArtwork } from "../Mutations/useDeleteArtwork"
 
 export type ArtworkInput =
   | MyCollectionCreateArtworkInput
   | MyCollectionUpdateArtworkInput
 
-export const useCreateOrUpdateArtwork = () => {
+export const useCreateOrUpdateOrDeleteArtwork = () => {
```
olerichter00, line 12: I wonder if it makes sense to do this change here. We could use `const { submitMutation: deleteArtwork } = useDeleteArtwork()` directly in [MyCollectionArtworkForm.tsx](https://github.com/artsy/force/pull/10834/files#diff-7445b7d2b9d4df4ded453cf7511b68c1234ad7d6168cabd72b81d75309fa2cce)`. I've just added this helper to upudate-or-create an artwork with one function call.

olerichter00 (changes requested):
This looks great! I've added some suggestions and comments.

src/Apps/MyCollection/Routes/EditArtwork/Utils/useCreateOrUpdateOrDeleteArtwork.ts
dariakoko: I don't like the idea of having requests in separate files. I think it makes sense to have everything in one place
What if I keep upudate-or-create function and add another one only for deletion to the same file?

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
dariakoko: Agree :)

src/Apps/MyCollection/Routes/EditArtwork/Utils/useCreateOrUpdateOrDeleteArtwork.ts
olerichter00: It seems to be the preferred pattern in Force to use the `useMutation` pattern and than use `commitMutation` directly in the component that uses the file ([see comment from @damassi ](https://github.com/artsy/force/pull/10732#discussion_r942810460)).

dariakoko: Aaaah, alright, I didn't know
Changing to the preferred way

olerichter00: Thanks 🙇

olerichter00 (approved):

olerichter00 (approved):

PR #10836, opened 2022-08-25
fix(TX-618) - analytics unknown schema errors

AlicanAkyuz:

This PR solves [TX-618]

### Description

This PR fixes the error seen in console when 2 events are tracked: 'unknown analytics schema is used'. The reason for the problem was that the correct `ActionType` property was missing in the event bodies. 

[TX-618]: https://artsyproduct.atlassian.net/browse/TX-618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -227,8 +227,6 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
     }
   }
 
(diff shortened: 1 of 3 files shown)


starsirius (approved):
Thank you for the quick fix!!

PR #10838, opened 2022-08-25
feat(FX-4169): allow Artsy employees to set up frequency of sending notifications

dimatretyak:

This PR solves [FX-4169]

Review app: https://notifications-frequency.artsy.net/

### Description
The goal of this ticket is to allow Artsy employees to configure instant Saved Alerts push notifications. Simply put an Admin only section (or better for any Artsy employee?) to expose a new "Enable instant notifications (beta)" when "Mobile Alerts" is checked. That would be enough to let us more easily test it out.

### Acceptance Criteria
* As an Artsy employee
* When setting up a saved alert and selecting “Mobile Alerts”
* I see a radio button group and can choose between daily and instant push notifications
* And my choice is saved in Gravity

#### Create flow
https://user-images.githubusercontent.com/3513494/186661712-90cf3116-466b-4aa3-8bde-387b07d14da4.mp4

#### Edit flow
https://user-images.githubusercontent.com/3513494/186661799-0c9d60d9-8bd5-4c52-9087-26b8eb578db4.mp4

[FX-4169]: https://artsyproduct.atlassian.net/browse/FX-4169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gkartalis (dismissed):
Looks great! Only one thought 🤔 should we show this feature to either **admins** or only **engineering**? 

Just thinking of the cases where people who don't have context that this is an artsy debugging specific feature might be confused. 

(don't consider this blocking, feel free to ignore and merge)

dimatretyak:
> should we show this feature to either admins or only engineering?

good question. maybe it would be better if we hide these changes under the unleash flag and display them only for specific developers. @nickskalkin what do you think about it?

nickskalkin (approved):
looking great! 👍

nickskalkin:
@gkartalis @dimatretyak good question! I think it's ok to enable this feature for all artsy employees:

1) there is an explanation what this feature does and that it is experimental
2) We can send a slack message to the product channel inviting people to play around with this feature

dimatretyak:
> @gkartalis @dimatretyak good question! I think it's ok to enable this feature for all artsy employees:
> 
> 1. there is an explanation what this feature does and that it is experimental
> 2. We can send a slack message to the product channel inviting people to play around with this feature

Make sense. thanks for the clarification!

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
```diff
@@ -0,0 +1,50 @@
+import { BorderedRadio, Box, RadioGroup, Text } from "@artsy/palette"
+import { themeGet } from "@styled-system/theme-get"
+import styled from "styled-components"
+import { useSystemContext } from "System"
+import { isArtsyEmail } from "Utils/isArtsyEmail"
```
dblandin, line 5: **suggestion**: I know this Email-based utility is already in use but I think we should prefer the `userIsTeam` util to determine whether the currently authenticated user works at Artsy. Seems less brittle than Email-based detection. What do you think?

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Box>
+        <Text variant="sm-display" mb={1}>
+          Frequency
+        </Text>
+        <RadioGroup defaultValue={defaultFrequence} onSelect={onSelect}>
+          <BorderedRadio
+            value="instant"
+            label={
+              <Text>
+                Instant <Warning>(experimental feature)</Warning>
```
dblandin, line 31: **suggestion**: Thoughts in including admin only in this message?

Instant (beta / admin only)

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+              </Text>
+            }
+          >
+            <Text color="black60">
+              Frequency of sending notifications ~ 1 time per hour
```
dblandin, line 36: **suggestion**: Maybe "Max 1 notification per hour“?

dblandin (commented):

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
dimatretyak: Yes, I agree with you. This approach would be more correct to use, since any user can potentially sign up with an email that includes `@artsymail.com` and use the account without confirmation

https://user-images.githubusercontent.com/3513494/186893848-418df42c-3c1d-4e77-ad6b-3af95f2e72f4.mp4

dimatretyak: but I think we should also use `userIsAdmin`

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
dimatretyak: updated. you can check changes in review app

src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx
dblandin: Ahh, I think "internal only" would work with the team role.

dblandin: To clarify, if we change the copy to be "internal only", I think that would align with the `team` role check. All provisioned Artsy employee accounts have the `team`, user role.

dimatretyak: did I understand you correctly:
1. here change copy from `Instant (beta / admin only)` to `Instant (beta / internal only)`
2. here we leave [the check](https://github.com/artsy/force/blob/71e3079e0620c7f9a13a5ff90498cc073ab6b28e/src/Components/SavedSearchAlert/Components/FrequencyRadioButtons.tsx#L18) only for `userIsTeam`

is that right?

dblandin: Yes, perfect!

dimatretyak: Done
image

dblandin (approved):

PR #10846, opened 2022-08-25
feat: run some staging pods in kubernetes spot tier.

artsyjian:
The type of this PR is: Feat

This PR solves PLATFORM-4514

### Description

Part of spike on using AWS Spot instances.
[Had created spot tier](https://github.com/artsy/substance/pull/283) in staging Kubernetes cluster.

### DIFF
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -93,6 +93,17 @@ spec:
                 operator: In
                 values:
                 - foreground
+                - foreground-spot
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: tier
+        whenUnsatisfiable: ScheduleAnyway
+        labelSelector:
+          matchLabels:
+            app: force
+            layer: application
+            component: web
+
 ---
 apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
--- a/hokusai/staging.yml
+++ b/hokusai/staging.yml
@@ -103,7 +103,6 @@ spec:
             app: force
             layer: application
             component: web
-
 ---
 apiVersion: autoscaling/v2beta2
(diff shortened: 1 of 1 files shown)


hokusai/staging.yml
```diff
@@ -93,6 +93,17 @@ spec:
                 operator: In
                 values:
                 - foreground
+                - foreground-spot
```
artsyjian, line 96: allow pods to run on nodes [labeled with foreground-spot](https://github.com/artsy/substance/blob/4809569bb58fd86f20eab2ce513ccd253dde4435/clusters/staging/federation/kubernetes-staging-leo.artsy.systems/spec.yml#L183).

hokusai/staging.yml
```diff
@@ -93,6 +93,17 @@ spec:
...
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: tier
```
artsyjian, line 99: spread pods by `tier` label.

hokusai/staging.yml
```diff
@@ -93,6 +93,17 @@ spec:
                 operator: In
                 values:
                 - foreground
+                - foreground-spot
+      topologySpreadConstraints:
+      - maxSkew: 1
```
artsyjian, line 98: when spreading pods between two groups, the max difference (in pod count) between the groups is 1.

hokusai/staging.yml
```diff
@@ -93,6 +93,17 @@ spec:
...
+        topologyKey: tier
+        whenUnsatisfiable: ScheduleAnyway
```
artsyjian, line 100: if spreading is not achievable, for example, if there are no instance available in `foreground-spot`, just schedule everything in `foreground`.

PR #10847, opened 2022-08-25
fix(404): `/oembed` route now 404s 

damassi:

### Description

Updates our `routes.ts` file with a 404 redirect for `/oembed`

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -36,6 +36,7 @@ import { settingsRoutes } from "Apps/Settings/settingsRoutes"
 import { showRoutes } from "Apps/Show/showRoutes"
 import { showsRoutes } from "Apps/Shows/showsRoutes"
 import { viewingRoomRoutes } from "Apps/ViewingRoom/viewingRoomRoutes"
+import { HttpError } from "found"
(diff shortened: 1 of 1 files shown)


joeyAghion:
This does seem like an odd way to accomplish a 404 error. Clients would get a non-error response (`301`) to the route, follow it, and then get a `404` at the new location. I also don't see other redirects to error statuses in this list, so is there just not a pattern in Force yet for this?

I see a few `throw new HttpError(404)` calls in other places. Is a simple express route that does that possible (or is that complicated by react app overhead)?

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 07860bc44fd8d7d26fb34ba0a9ca87eca787f8a8
</p>

joeyAghion (approved):

PR #10851, opened 2022-08-26
fix: My Collection artwork form header 

olerichter00:
Resolves [CX-2843](https://artsyproduct.atlassian.net/browse/CX-2843) & [2844](https://artsyproduct.atlassian.net/browse/CX-2844)

## Description

This PR changes the copy of and removes the `Stepper` component from the artwork form header. This solves the following problems:

- Text was overlapping the sticky header
- Chevron Icon was visible at the end of the stepper.


MounirDhahri (dismissed):

MrSltun (approved):
🌟

MounirDhahri (approved):

PR #10856, opened 2022-08-26
refactor(FX-4185): ArtworkApp to functional component and add ff for new artworksidebar

gkartalis:

This PR solves [FX-4185]

### Description

paired on this with @mdole 

- [x] refactoring ArtworkApp component to a functional one
- [x] added `fx-force-artwork-sidebar` unleash feature flag
- [x] use ☝️ ff to toggle between old / new artworksidebar 

[FX-4185]: https://artsyproduct.atlassian.net/browse/FX-4185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -150,83 +113,117 @@ export class ArtworkApp extends React.Component<Props> {
...
+   */
+  useEffect(() => {
+    if (shouldRenderSubmittedOrderModal()) {
+      // TODO: Look into using router push
+      // this.props.router.replace(this.props.match.location.pathname)
+      window.history.pushState({}, null, props.match.location.pathname)
+    }
+    track()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
dzucconi, line 144: What happens if you navigate from one artwork to another artwork directly?

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -150,83 +113,117 @@ export class ArtworkApp extends React.Component<Props> {
...
+          artwork.artists.map(artist => {
+            if (!artist) return null
+
+            return (
+              <ArtistInfoQueryRenderer key={artist.id} slug={artist.slug} />
+            )
+          })}
+      </Join>
+    </>
+  )
```
dzucconi, line 170: I would extract this into its own component. Generally, you want to avoid declaring a component within the body of another.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -150,83 +113,117 @@ export class ArtworkApp extends React.Component<Props> {
...
+  const track = useCallback(() => {
+    trackPageview()
+    trackProductView()
+    trackLotView()
+  }, [trackPageview, trackProductView, trackLotView])
 
-          <PricingContextQueryRenderer slug={artwork.slug} />
+  const shouldRenderSubmittedOrderModal = useCallback(() => {
+    return !!props.match.location.query["order-submitted"]
+  }, [props.match.location.query])
```
dimatretyak, line 126: **question**: can we do something like this?

```javascript
const shouldRenderSubmittedOrderModal = !!props.match.location.query["order-submitted"]

...

useEffect(() => {
  if (shouldRenderSubmittedOrderModal) {
    // TODO: Look into using router push
    // this.props.router.replace(this.props.match.location.pathname)
    window.history.pushState({}, null, props.match.location.pathname)
  }
  track()
}, [])

...

{shouldRenderSubmittedOrderModal && (
  <SubmittedOrderModalFragmentContainer slug={artwork.slug} me={me} />
)}
...

```

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -150,83 +113,117 @@ export class ArtworkApp extends React.Component<Props> {
...
-              )
-            })}
-        </Join>
-      </>
-    )
+  useEffect(() => {
+    if (shouldTrackPageView) {
+      track()
+    }
+  }, [shouldTrackPageView, track])
```
dimatretyak, line 150: **suggestion**: i think we have to skip first render. maybe we can use something [like this](https://github.com/artsy/palette/blob/main/packages/palette/src/utils/useUpdateEffect.ts)?

dimatretyak: **suggestion**: remove deps for `useEffect` and don't use `useCallback` for `trackProductView`, `trackLotView`, `trackPageview` (by doing this, we will get the same effect as before with `componentDidUpdate`)

src/Apps/Artwork/Components/ArtworkSidebar2/index.tsx
```diff
@@ -0,0 +1,3 @@
+import { Flex } from "@artsy/palette"
```
damassi, line 1: No need to add an index file... just import from `ArtworkSidebar2` directly

damassi (dismissed):
We should triple check that the pageview tracking is working as it should. The artwork page is weird in that it sends along additional data (circumventing the automatic flow), and getting this to work correctly a few years ago could best be described as a "hack"

src/Apps/Artwork/ArtworkApp.tsx
dimatretyak: [SOME UPDATE] I think we can use your implementation, but just keep `shouldTrackPageView` in deps
```javascript
useEffect(() => {
  if (shouldTrackPageView) {
    track()
  }
}, [shouldTrackPageView])
```

https://user-images.githubusercontent.com/3513494/187208627-d808f40b-7215-423c-9b0f-5062f67ecb99.mp4

src/Apps/Artwork/ArtworkApp.tsx
damassi: Just looking for things like double page views, and that things are tracked. We previously had an issue where navigating from artwork to artwork didn't yield multiple page views, that kind of thing.

gkartalis: Perfect thanks! We did a side by side comparison and everything works smoothly

PR #10863, opened 2022-08-26
fix(misc-tickets): Debounce pageview tracking and hypenate partner badges 

damassi:

This PR solves
- https://artsyproduct.atlassian.net/browse/GRO-1242 (onboarding events firing more than once)
- https://artsyproduct.atlassian.net/browse/GRO-1210 (hyphenate gallery ownership badges) 

### Description

Fixes the first ticket by adding a debounce to our analytics middleware. This is good protection (just in general) but more specific to the ticket, we were getting three page view calls because we 
- create account and then
- redirect back to page where user was (page view)
- and trigger onboarding by updating url bar with query param (page view)
- then replace url by stripping `onboarding=true` query param (page view)

which is three routing events (push, push, replace), which all tap into our analytics middleware. So this adds some protection against that by ensuring we can never fire these rapid fire page view events (which should never even be possible from a UX perspective)

The second ticket just hyphenates the display of our categories. Quick fix vs needing to update a few dbs and old rails app vibrations haml UI. Doesn't seem like that big of a deal but happy to go through the hoops there. 

Screen Shot 2022-08-26 at 2 16 02 PM

Screen Shot 2022-08-26 at 2 15 20 PM

cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Partner/PartnerApp.tsx
+++ b/src/Apps/Partner/PartnerApp.tsx
@@ -54,7 +54,10 @@ export const PartnerApp: React.FC<PartnerAppProps> = ({
 
         <FullBleed mb={[2, 4]}>
           {firstEligibleBadgeName ? (
-            <Marquee speed="static" marqueeText={firstEligibleBadgeName} />
+            <Marquee
+              speed="static"
+              marqueeText={firstEligibleBadgeName.replace(" ", "-")} // hypenate gallery badges
+            />
           ) : (
             <Separator />
           )}
--- a/src/System/Analytics/trackingMiddleware.ts
+++ b/src/System/Analytics/trackingMiddleware.ts
@@ -1,7 +1,8 @@
 import { ActionTypes } from "farce"
-import { data as sd } from "sharify"
 import { get } from "Utils/get"
 import { match } from "path-to-regexp"
+import { debounce } from "lodash"
+import { getENV } from "Utils/getENV"
 
 /**
  * PageView tracking middleware for use in our router apps. Middleware conforms
@@ -16,6 +17,8 @@ interface TrackingMiddlewareOptions {
 }
 
 export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
+  const trackPageview = debounce(window.analytics?.page!, 30)
+
   return store => next => action => {
     const { excludePaths = [] } = options
     const { type, payload } = action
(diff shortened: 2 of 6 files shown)


src/Apps/Partner/PartnerApp.tsx
```diff
@@ -54,7 +54,10 @@ export const PartnerApp: React.FC<PartnerAppProps> = ({
 
         <FullBleed mb={[2, 4]}>
           {firstEligibleBadgeName ? (
-            <Marquee speed="static" marqueeText={firstEligibleBadgeName} />
+            <Marquee
+              speed="static"
+              marqueeText={firstEligibleBadgeName.replace(" ", "-")} // hypenate gallery badges
+            />
```
jonallured, line 60: Could also do this:

```
const galleryBadges = ["Black-Owned", "Women-Owned"]
```

But whatevs I don't feel strongly one way or the other very easy to change this stuff. Also I agree that it's not worth messing around in an admin interface. 👍

jonallured (approved):
I'm slightly concerned that this change (debounce events) will solve this situation but then elsewhere we're actually relying on this odd behavior and it'll take some amount of time for us to notice it. 😬 

But yeah we should do this. I guess I'll just tag @artsy/data here for visibility!

damassi:
> I'm slightly concerned that this change (debounce events) will solve this situation but then elsewhere we're actually relying on this odd behavior

I had this thought too, but in reality a page view is like a user landing on a new page. If a user can't click from page a to page b and back to page a in 5 milliseconds then any tracking of such is invalid. 

This also fixes other areas i've noticed we update the url really quickly, such as on the artwork grid pages where we set the users `in/metric` preferences as well as on auction/id registration. All of those quick updates are invalid page view tracking events.

src/System/Analytics/trackingMiddleware.ts
```diff
@@ -16,6 +17,8 @@ interface TrackingMiddlewareOptions {
 }
 
 export function trackingMiddleware(options: TrackingMiddlewareOptions = {}) {
+  const trackPageview = debounce(window.analytics?.page!, 30)
```
dzucconi, line 20: So — what about this broke tracking @damassi

damassi: No longer tracked page on transition between SPA pages

PR #10871, opened 2022-08-29
feat(FX-4176): Implement All / Alerts  tabs on the Activity Panel standalone page.

dimatretyak:

This PR solves [FX-4176]

Review app: https://tabs-for-activity-panel.artsy.net/notifications

### Acceptance Criteria
* Implement All / Alerts  tabs on the Activity Panel standalone page.
* Use `notificationTypes` field exposed in [this ticket](https://artsyproduct.atlassian.net/browse/FX-4153).

[FX-4176]: https://artsyproduct.atlassian.net/browse/FX-4176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Notifications/NotificationsApp.tsx
+++ b/src/Apps/Notifications/NotificationsApp.tsx
@@ -1,4 +1,4 @@
-import { Box } from "@artsy/palette"
+import { Spacer } from "@artsy/palette"
 import { MetaTags } from "Components/MetaTags"
 import { Notifications } from "Components/Notifications/Notifications"
 
@@ -7,12 +7,9 @@ export const NotificationsApp: React.FC = () => {
     <>
       <MetaTags title="Notifications | Artsy" pathname="/notifications" />
 
-      <Box
-        // Unconventional value here to keep visual rhythm
-        mt="30px"
-      >
-        <Notifications />
-      </Box>
+      <Spacer mt={4} />
+
+      <Notifications />
     </>
   )
 }
--- a/src/Apps/Notifications/NotificationsApp.tsx
+++ b/src/Apps/Notifications/NotificationsApp.tsx
@@ -1,12 +1,18 @@
+import { Box } from "@artsy/palette"
 import { MetaTags } from "Components/MetaTags"
-import { NotificationsListQueryRenderer } from "Components/Notifications/NotificationsList"
+import { Notifications } from "Components/Notifications/Notifications"
 
 export const NotificationsApp: React.FC = () => {
   return (
     <>
       <MetaTags title="Notifications | Artsy" pathname="/notifications" />
 
-      <NotificationsListQueryRenderer />
+      <Box
+        // Unconventional value here to keep visual rhythm
+        mt="30px"
+      >
+        <Notifications />
+      </Box>
     </>
   )
 }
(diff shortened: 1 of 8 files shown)


src/Components/Notifications/NotificationsList.tsx
```diff
@@ -102,19 +111,29 @@ export const NotificationsListFragmentContainer = createPaginationContainer(
   }
 )
 
-export const NotificationsListQueryRenderer = () => {
+interface NotificationsListQueryRendererProps {
+  type: NotificationType
```
dimatretyak, line 115: **question**: do you think it might be better to use an **automatically generated enum**?

```javascript
interface NotificationsListQueryRendererProps {
  type?: NotificationTypesEnum
}

...

<NotificationsListQueryRenderer type="ARTWORK_PUBLISHED" /> // All tab
<NotificationsListQueryRenderer type="ARTWORK_ALERT" /> // Alerts tab
<NotificationsListQueryRenderer type="ARTWORK_PUBLISHED" /> // Other supported types
```

src/Apps/Notifications/NotificationsApp.tsx
```diff
@@ -1,12 +1,18 @@
...
+import { Notifications } from "Components/Notifications/Notifications"
 
 export const NotificationsApp: React.FC = () => {
   return (
     <>
       <MetaTags title="Notifications | Artsy" pathname="/notifications" />
 
-      <NotificationsListQueryRenderer />
+      <Box
+        // Unconventional value here to keep visual rhythm
```
dzucconi, line 11: Generally we use `4` — what about this requires a 3?

dimatretyak: the offset of 30px is specified in the design 😅

image

dimatretyak: but I agree with you that it is better to make 40px so that there is a correspondence to other places

image

dimatretyak: but I agree with you that it is better to make 40px so that there is a correspondence to other places

damassi: @dimatretyak - The design is incorrect: we have no "30" space in our design system. We need to take it back to design.

dblandin: I just left a comment for Ryan in the Figma spec [here](https://www.figma.com/file/SlzmAJV6zqJDmAO8GoSjiC?node-id=1336:82114#257544863).

I see that the spacing was changed to `4` (=40px). This UI isn't live yet (it's behind a feature flag). We'll make sure that the implementation is aligned with design (and that the design is aligned with the design system), well before launching anything.

damassi: Cool 👍 its more just a matter of developers knowing that we can go back to design and mention the inconsistency versus just creating whatever is send to us. Sometimes it's just an accident on the part of the designer, or a new designer not familiar with the constraints of the design sys (and so on).

PR #10873, opened 2022-08-29
feat: Add MyC Save button loading state when processing images

olerichter00:
## Description

Add MyC Save button loading state when processing images.


MrSltun (approved):
LGTM! 🚀

PR #10875, opened 2022-08-29
fix: Replacing my-collection routes in redirects to avoid 404s

olerichter00:
## Description

Replacing `/my-collection` with `/settings/my-collection` in redirects to avoid 404s. This is probably related to the redirect.


MounirDhahri (approved):
makes sense

PR #10888, opened 2022-08-30
feat[FX-4145]: add new artists section in artwork sidebar

mdole:

This PR solves [FX-4145]

### Description

With 1 artist
Screen Shot 2022-08-30 at 18 41 07
---
With > 4 artists, not expanded
Screen Shot 2022-08-30 at 18 37 35
---
With > 4 artists, expanded
Screen Shot 2022-08-30 at 18 37 32
---
With cultural maker
Screen Shot 2022-08-30 at 18 38 02

[FX-4145]: https://artsyproduct.atlassian.net/browse/FX-4145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damassi:
Nice working through the palette changes 👌 It is indeed better.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -198,7 +198,7 @@ export const ArtworkApp: React.FC<Props> = props => {
 
         <Column span={4} pt={[0, 2]}>
           {isNewArtworkSidebarEnabled ? (
-            <ArtworkSidebar2 />
+            <ArtworkSidebar2FragmentContainer artwork={artwork} />
```
gkartalis, line 201: 🔥

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2Artists.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export const ArtworkSidebar2Artists: React.FC<ArtistsProps> = ({
+  artwork: { artists, culturalMaker },
+}) => {
+  if (!artists) return null
+
+  const showMoreText = `${artists.length - ARTISTS_TO_DISPLAY} more`
+
+  return (
+    <div>
```
gkartalis, line 33: **question:** is there a reason that you picked a `<div>` here over a `<>` fragment or a `<Box>` ?

gkartalis (approved):

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2Artists.tsx
mdole: Not really - it's a personal preference thing. Basically, why use `<>` when it's actually just a `<div>`? AFAIK there's not any advantage to it, it's just saving 3 characters (6 if you count the closing tag) and using components where we could be writing HTML sometimes gets us into trouble 🤷 

But I'm definitely not consistent about it either!

PR #10889, opened 2022-08-30
fix: specify payment method for each Payment Element

starsirius:

### Description

We started to see Integrity failure roughly after https://github.com/artsy/force/pull/10859. There seems to be 2 reasons:

- Stripe seemed to change their iframe implementation again, and our [iframe selector in Integrity](https://github.com/artsy/integrity/blob/0cdd676c94025bb0dd40d988031d0da71ec68e0a/cypress/utils/checkAndAddBankAccount.ts#L12) will find 2 matches.

Screen Shot 2022-08-30 at 1 50 02 PM

- After https://github.com/artsy/force/pull/10859, we'll render 2 sections with `[data-test="bankTransferSection"]` (one for ACH one for SEPA). The selection then becomes ambiguous.

This PR updates the `data-test` attribute to be payment method specific so our selector in Integrity can be more reliable.

### DIFF
--- a/src/Apps/Order/Routes/Payment/PaymentContent.tsx
+++ b/src/Apps/Order/Routes/Payment/PaymentContent.tsx
@@ -154,7 +154,7 @@ export const PaymentContent: FC<Props> = props => {
         <BankAccountPickerFragmentContainer
           me={me}
           order={order}
-          paymentMethod={paymentMethod}
+          paymentMethod="US_BANK_ACCOUNT"
           bankAccountHasInsufficientFunds={bankAccountHasInsufficientFunds}
           onSetBankAccountHasInsufficientFunds={
             onSetBankAccountHasInsufficientFunds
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -175,7 +175,7 @@ export const PaymentContent: FC<Props> = props => {
         <Spacer mb={2} />
         <BankDebitProvider
           order={order}
-          paymentMethod={paymentMethod}
+          paymentMethod="SEPA_DEBIT"
```
starsirius, line 178: It seems more reasonable to pass in the specific payment method; otherwise, we'll render multiple `Collapse`s with the same payment method. Let me know if I misunderstood things here!

AlicanAkyuz (approved):

PR #10892, opened 2022-08-31
fix: conversation list large screen scroll

jo-rs:
The type of this PR is: Fix

This PR solves [NX-3274]

### Description

Implements `IntersectionObserver` on the conversation list and replaces hard-coded variables with constants. This fixes an issue where users wouldn't get more conversations loaded on larger screens, (hopefully) preventing it from resurfacing.

Before:
before

After:
after

_I'm creating a ticket to also implement `IntersectionObserver` on the actual conversation as a follow-up._

[NX-3274]: https://artsyproduct.atlassian.net/browse/NX-3274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Conversation/__tests__/ConversationApp.jest.tsx
```diff
@@ -54,6 +54,16 @@ describe("Conversation app", () => {
...
+  beforeEach(() => {
+    // IntersectionObserver isn't available in test environment
+    const mockIntersectionObserver = jest.fn()
+    mockIntersectionObserver.mockReturnValue({
+      observe: () => null,
+      unobserve: () => null,
+      disconnect: () => null,
+    })
+    window.IntersectionObserver = mockIntersectionObserver
+  })
```
jo-rs, line 66: Mocking `IntersectionObserver` since it's not available on test runs.

src/Apps/Conversation/Components/ConversationList.tsx
```diff
@@ -27,12 +31,12 @@ const SpinnerContainer = styled.div`
 `
 
 const ConstrainedHeightContainer = styled(Flex)`
-  height: calc(100vh - 103px);
+  height: calc(100vh - ${DESKTOP_NAV_BAR_HEIGHT}px);
```
jo-rs, line 34: There was a pixel added here which was causing the initial issue.

lidimayra: No waaaay!!!! 🤯

lidimayra (approved):
Not familiar with `IntersectionObserver`, but we have a bug and this PR fixes it, therefore LGTM! 🚀

src/Apps/Conversation/Components/ConversationList.tsx
```diff
@@ -86,11 +88,14 @@ const ConversationList: React.FC<ConversationsProps> = props => {
...
-            <Spinner />
-          </SpinnerContainer>
-        )}
+        <Flex>
+          {!!fetchingMore && (
+            <SpinnerContainer>
+              <Spinner />
+            </SpinnerContainer>
+          )}
+          <Box ref={ref as any} height={1} />
```
araujobarret, line 97: That's what I imagined, well done here!

PR #10896, opened 2022-08-31
feat: sending a message to eigen on successful order

pvinis:

This PR is part of [MOPLAT-311]

### Description
We need to get a callback so that we know the order was successful. That way we can do some work after that.

https://user-images.githubusercontent.com/100233/187646529-d92b7988-ab9a-4847-b668-ad7f6c3650ac.mov

[MOPLAT-311]: https://artsyproduct.atlassian.net/browse/MOPLAT-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Review/index.tsx
+++ b/src/Apps/Order/Routes/Review/index.tsx
@@ -129,24 +129,28 @@ export const ReviewRoute: FC<ReviewProps> = props => {
           o => o.lineItems?.edges?.[0]?.node?.artwork?.slug
         )
 
-        if (
-          order.mode === "OFFER" &&
-          isEigen &&
-          order.source === "artwork_page"
-        ) {
-          window?.ReactNativeWebView?.postMessage(
-            JSON.stringify({
-              key: "goToInboxOnMakeOfferSubmission",
-              orderCode: order.code,
-              message: `The seller will respond to your offer by ${order.stateExpiresAt}. Keep in mind making an offer doesn’t guarantee you the work.`,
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -129,24 +129,28 @@ export const ReviewRoute: FC<ReviewProps> = props => {
...
+            )
+            // We cannot expect Eigen to respond all the time to messages sent from the webview
+            // a default fallback page is safer for old/broken Eigen versions
+            setTimeout(() => {
+              router.push(`/orders/${order.internalID}/status`)
+            }, 500)
+            return
+          }
+
+          if (order.mode === "BUY") {
```
araujobarret, line 149: I'm from mobile phone here(ooo for the next 2 weeks) but I would recommend a fallback for older Eigen versions that have no interaction with this new dispatched message. Maybe moving line 143 block down?

pvinis: what will happen if its an older version eigen? 🤔 the message will be posted, but no interaction will happen, right?

if so, that is totally fine for us.

araujobarret: Yep, if that's the expected it's fine!

damassi:
👍

PR #10900, opened 2022-08-31
feat(FX-4192): add artwork title to sidebar2

gkartalis:

This PR solves [FX-4192]

### Description

Adds artwork title to new sidebar

(hidden behind a ff)
Screenshot 2022-08-31 at 14 37 26

[FX-4192]: https://artsyproduct.atlassian.net/browse/FX-4192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gkartalis:
note that we decided [here](https://artsy.slack.com/archives/C9SATFLUU/p1661948315026059) to have the artwork title as italic and the year as not italic

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
```diff
@@ -23,6 +21,16 @@ export const ArtworkSidebar2: React.FC<ArtworkSidebarProps> = props => {
   return (
     <Flex flexDirection="column">
       <ArtworkSidebar2ArtistsFragmentContainer artwork={artwork} />
+      <Flex flexDirection="row">
+        <Text color="black60" variant="lg-display" fontStyle="italic">
+          {artwork.title}
+        </Text>
+        <Text color="black60" variant="lg-display">
+          , {artwork.date}
+        </Text>
```
mdole, line 30: @gkartalis I think we'll likely want to put these pieces of text in the same `<Text>` or us `as="span"` or find some other way of avoiding awkward formatting for long titles.

We might actually just want to steal what's in the current component verbatim since it seems to handle a few edge cases already: https://github.com/artsy/force/blob/470b6a3189f9592f4e5298687088e095f432b999/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarTitleInfo.tsx#L20-L23

Example: http://localhost:5000/artwork/euan-roberts-green-balenciaga-pink-tiles

Screen Shot 2022-08-31 at 15 22 30

mdole (changes requested):

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
gkartalis: good catch

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
```diff
@@ -23,6 +21,14 @@ export const ArtworkSidebar2: React.FC<ArtworkSidebarProps> = props => {
   return (
     <Flex flexDirection="column">
       <ArtworkSidebar2ArtistsFragmentContainer artwork={artwork} />
+      <Text color="black60" variant="lg-display">
+        <i>{artwork.title?.trim()}</i>
+        {artwork.date &&
+          artwork.date.replace(/\s+/g, "").length > 0 &&
+          ", " + artwork.date}
```
dimatretyak, line 28: **suggestion**: what do you think about this?

```javascript
const getArtworkDate = () => {
  const formattedDate = artwork.date?.replace(/\s+/g, "") ?? ""

  if (formattedDate.length > 0) {
    return `, ${artwork.date}`
  }

  return
}
```

mdole (approved):
Ty for the updates!

PR #10901, opened 2022-08-31
feat(TX-638): Introduce React Context to Payment Route

AlicanAkyuz:
The type of this PR is: **Experiment**

This PR solves [TX-638]

### Description

This PR introduces React context to the payment route of the Order App. Official docs are here: https://reactjs.org/docs/context.html

The context has a reducer that is available to the children of the provider for accessing and updating the context state. 

This is just a small example of how this would work and no actual data is currently used. good news is that we can incrementally adopt this approach if we agreed to adopt it.

Some todos: 
- come up with a simple test suite that works with context and useReducer 

[TX-638]: https://artsyproduct.atlassian.net/browse/TX-638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Payment/context.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useReducer, createContext, Dispatch } from "react"
+import { paymentReducer } from "./reducer"
+import { PaymentAction, PaymentState } from "./types"
+
+const initialPaymentState = {
+  selectedBankAccountId: "",
+}
+
```
AlicanAkyuz, line 8: for now I just added a single property to the state and it is **not** used in the actual app. 

if we want to adopt this approach, this would be the place where `useState` definitions would be translated into.

src/Apps/Order/Routes/Payment/context.tsx
```diff
@@ -0,0 +1,32 @@
...
+  dispatch: () => null,
+})
+
+const PaymentProvider: React.FC = ({ children }) => {
+  const [state, dispatch] = useReducer(paymentReducer, initialPaymentState)
+
+  // for developers' convenience only
+  if (process.env.NODE_ENV === "development") {
+    console.log({ paymentState: state })
+  }
```
AlicanAkyuz, line 23: I thought it'd be good to log the state to the console, which can help us the way redux-devtools does. the same can also be done for fired actions.

src/Apps/Order/Routes/Payment/reducer.ts
```diff
@@ -0,0 +1,13 @@
+import { PaymentState, PaymentAction, PaymentActions } from "./types"
+
+export const paymentReducer = (state: PaymentState, action: PaymentAction) => {
+  switch (action.type) {
+    case PaymentActions.SET_SELECTED_BANK_ACCOUNT_ID:
+      return {
+        ...state,
+        selectedBankAccountId: action.payload,
+      }
```
AlicanAkyuz, line 9: I kept the action right in the reducer, which might turn into a long long switch statement if we moved all state updaters here.

another idea is to have actions in a separate file. just focused functions. any idea?

src/Apps/Order/orderRoutes.tsx
```diff
@@ -136,7 +137,11 @@ export const orderRoutes: AppRouteConfig[] = [
         }
       }
 
-      return <Component {...props} />
+      return (
+        <PaymentProvider>
+          <Component {...props} />
+        </PaymentProvider>
+      )
```
AlicanAkyuz, line 144: @damassi I'd like to get your feedback especially on this part. it kinda feels off to provide the wrapper in the file where routes are defined. 

do you have any feedback?

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -71,6 +73,17 @@ export interface BankAccountSelection {
...
+  console.log({ state })
+
+  useEffect(() => {
+    dispatch({
+      type: PaymentActions.SET_SELECTED_BANK_ACCOUNT_ID,
+      payload: "XXX",
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
+
```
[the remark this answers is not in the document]
AlicanAkyuz: ps: it doesn't have to be in an effect!

src/Apps/Order/Routes/Payment/reducer.ts
AlicanAkyuz: overall I think it'd be good to keep the reducer logic-free.

src/Apps/Order/Routes/Payment/context.tsx
damassi: We should make the message a bit more verbose via our logger, just so we know it's a dev thing. Something like:
```
import createLogger from 'Utils/logger'

const logger = createLogger("Order/Routes/Payment/PaymentContext.ts") 
...
logger.log(...)

File: src/Apps/Order/Routes/Payment/context.tsx [hunk 6]
```diff
@@ -0,0 +1,32 @@
+import React, { useReducer, createContext, Dispatch } from "react"
```
damassi, line 1: context.ts -> `OrderPaymentContext.ts'

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -71,6 +73,17 @@ export interface BankAccountSelection {
 }
 
 export const PaymentRoute: FC<PaymentRouteProps> = props => {
+  const { state, dispatch } = useContext(PaymentContext)
```
damassi, line 76:
```tsx
const { state, dispatch } = useOrderPaymentContext() 
```

src/Apps/Order/Routes/Payment/index.tsx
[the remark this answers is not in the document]
damassi: Not strictly necessary, but its common to wrap the dispatches behind helpful context functions passed into the context; eg, 

```tsx
const contextValues = {
  setSelectedBankAccountID: (payload) => {
    dispatch({
      type: PaymentActions.SET_SELECTED_BANK_ACCOUNT_ID,
      payload: "XXX",
    })
  }
}
```
and then later, from the callsite:

```tsx
const { setSelectedBankAccountID } = useOrderPaymentContext()
```

src/Apps/Order/Routes/Payment/reducer.ts
```diff
@@ -0,0 +1,13 @@
+import { PaymentState, PaymentAction, PaymentActions } from "./types"
```
damassi, line 1: `orderPaymentReducer.ts`

src/Apps/Order/Routes/Payment/reducer.ts
damassi: Generally speaking, logic lives in the reducer. That way one doesn't have to go all around in order to update logic relating to state; the dev can go to one file and make changes necessary, keeping a strict separation of concerns. If things grow in complexity then further organization might be necessary, but for the most part this is all that's needed. 

Curious what your thoughts are @AlicanAkyuz?

src/Apps/Order/Routes/Payment/types.ts
```diff
@@ -0,0 +1,26 @@
...
+type ActionMap<M extends { [index: string]: any }> = {
+  [Key in keyof M]: M[Key] extends undefined
+    ? {
+        type: Key
+      }
+    : {
+        type: Key
+        payload: M[Key]
+      }
+}
```
damassi, line 26: totally fine to keep a types file! I have found it easier in experience to keep types related to reducer / context all inside the context file tho just for ease of access. But no strong opinions here.

src/Apps/Order/orderRoutes.tsx
```diff
@@ -5,6 +5,7 @@ import { ErrorPage } from "Components/ErrorPage"
 import { Redirect, RedirectException } from "found"
 import { graphql } from "react-relay"
 import { AppRouteConfig } from "System/Router/Route"
+import { PaymentProvider } from "./Routes/Payment/context"
```
damassi, line 8:
```tsx
import { OrderPaymentContextProvider } from './Routes/Payment/OrderPaymentContext'
```

It's more verbose, but works better at scale to keep naming as descriptive as possible and 1:1 with the file name (where applicable)

src/Apps/Order/orderRoutes.tsx
damassi: This feels right! The idea here is to incrementally adopt and expand context state over time. So you have a context for the payment portion of the app -- it should only wrap that part of the route, since it's the root. Then later, if state expands, you wrap another route root with another context, then another. And finally, as the final refactor, you can combine all of the state together into one `OrderContext` and merge each of the files together into one single context and then access slices of the state via a single context hook. But this isn't strictly necessary, and incrementally adopting by opting in to certain routes the way you're doing here is def the way to go. Each route is effectively its own sub-app.

damassi:
Psyched @AlicanAkyuz! So far so good 👍

src/Apps/Order/Routes/Payment/types.ts
AlicanAkyuz: I think it makes sense, will move!

src/Apps/Order/orderRoutes.tsx
AlicanAkyuz: thanks, @damassi!

src/Apps/Order/Routes/Payment/index.tsx
AlicanAkyuz: yup! I'm secretly hoping that testing would also be easier with a hook.

src/Apps/Order/Routes/Payment/OrderPaymentContext.tsx
```diff
@@ -0,0 +1,85 @@
+import React, { useReducer, useContext, createContext, Dispatch } from "react"
+import { CommercePaymentMethodEnum } from "__generated__/Payment_order.graphql"
+import { orderPaymentReducer } from "./orderPaymentReducer"
+import createLogger from "Utils/logger"
+
+const logger = createLogger("[dev: OrderPaymentContext] state:")
+
+/* ACTIONS ------------------------------------------------------------------ */
```
damassi, line 8: This is super minor and feel free to ignore, but we don't usually see this kind of comment form throughout the codebase. Something as simple as 

```
/**
 * Actions 
 */ 
```

Is sufficient

src/Apps/Order/Routes/Payment/OrderPaymentContext.tsx
```diff
@@ -0,0 +1,100 @@
...
+}>({
+  state: initialOrderPaymentState,
+  dispatch: () => null,
+})
+
+/**
+ * Provider
+ */
+
+const OrderPaymentProvider: React.FC = ({ children }) => {
```
iskounen, line 67: **nitpick:** Would `OrderPaymentContextProvider` be more accurate name for this component?

src/Apps/Order/Routes/Payment/orderPaymentReducer.ts
```diff
@@ -0,0 +1,28 @@
...
+  action: OrderPaymentAction
+) => {
+  switch (action.type) {
+    case OrderPaymentActions.SET_SELECTED_BANK_ACCOUNT_ID: {
+      return {
+        ...state,
+        selectedBankAccountId: action.payload,
+      }
+    }
+    case OrderPaymentActions.SET_SELECTED_PAYMENT_METHOD: {
```
iskounen, line 18: **suggestion:** Would it make sense to set `selectBankAccountId` to `null` if the selected payment method is credit card or wire transfer?

src/Apps/Order/Routes/Payment/index.tsx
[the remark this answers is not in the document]
damassi: @AlicanAkyuz - thoughts on this ^ comment?

erikdstock:
nit: i would organize the new context and reducer-related files in a subdirectory here
image

src/Apps/Order/Routes/Payment/OrderPaymentContext.tsx
AlicanAkyuz: def makes sense, I missed the consistency there. thank you!

src/Apps/Order/Routes/Payment/orderPaymentReducer.ts
AlicanAkyuz: that's a good suggestion. for now we just wanted to see how the whole flow would work, we haven't put much effort into the reducer logic yet. will do this for sure!

src/Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext.tsx
```diff
@@ -0,0 +1,100 @@
...
+  )
+}
+
+/**
+ * Hook
+ */
+
+const useOrderPaymentContext = () => {
+  const { state, dispatch } = useContext(OrderPaymentContext)
+  return { state, dispatch }
```
kiryl-zubarau, line 91: As I mentioned at the knowledge share we can return an action creator instead of dispatch.
Screenshot from 2022-09-09 14-49-32
And then use:
Screenshot from 2022-09-09 14-51-36

This example uses the hook as a place where we can create an function which will create an action and pass the action to dispatch. Then the function can be used at the needed place.

src/Apps/Order/Routes/Payment/index.tsx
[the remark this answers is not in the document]
AlicanAkyuz: sounds like a good option! another way we can do this is by wrapping dispatch calls by helper functions in the hook like @kiryl-zubarau suggested below. I'll push a commit in a bit with that, curious to get your opinion on it.

src/Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext.tsx
damassi: Yup, @AlicanAkyuz - this is basically the same thing I was saying in my comment up above! Then everything is very readable by just using autocomplete / intellisense, without us having to wire up dispatches and actions and so on.

src/Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext.tsx
```diff
@@ -0,0 +1,113 @@
...
+      payload,
+    })
+
+  const setSelectedPaymentMethod = payload =>
+    dispatch({
+      type: OrderPaymentActions.SET_SELECTED_PAYMENT_METHOD,
+      payload,
+    })
+
+  return { ...state, setSelectedBankAccountId, setSelectedPaymentMethod }
```
damassi, line 104: 👌 yup, exactly

src/Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export {
+  useOrderPaymentContext,
+  OrderPaymentContextProvider,
+  OrderPaymentState,
+  OrderPaymentAction,
+  OrderPaymentActions,
+}
```
damassi, line 113: Rather than exporting everything at the bottom lets just use our conventional 

```tsx
export const theThing = 
```
where exports are colocated with the actual definition

damassi (commented):
This PR is looking good @AlicanAkyuz and co, all the bases and patterns are covered 👍

src/Apps/Order/Routes/Payment/index.tsx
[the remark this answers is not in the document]
damassi: Cool, so i'll consider this resolved 👍

src/Components/BankDebitForm/__tests__/BankDebitForm.jest.tsx
```diff
@@ -4,6 +4,7 @@ import { useSystemContext } from "System"
 import { useTracking } from "react-tracking"
 import { PaymentTestQueryRawResponse } from "__generated__/PaymentTestQuery.graphql"
 import { BankDebitForm } from "../BankDebitForm"
+import { useOrderPaymentContext } from "../../../Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext"
```
damassi, line 7: `"Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext"` should just work here and below

damassi (commented):
@AlicanAkyuz - One thing that would be good to add here before merge is an pattern for how to test. (And then follow-up with tests where necessary)

AlicanAkyuz:
@damassi how do you mean? I did add tests for behaviours by mocking the hook only. do you mean to test the calls to the reducer and updates to the context?

damassi:
Oh I see 👍 

Happy that this was merged, thanks for the work on this 💯

PR #10913, opened 2022-09-01
fix: Add my-collection route to history when saving an artwork

olerichter00:

### Description

Currently, when saving an artwork after editing it, the user will be navigated to the artwork detail page. At this point it's not easy to navigate back to the My Collection overview because pressing the back button would navigate the user again back to the edit page. 

This PR fixes this by replacing the route with the `/my-collection` route after saving the artwork. 


src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2Artists.tsx
```diff
@@ -33,7 +33,7 @@ export const ArtworkSidebar2Artists: React.FC<ArtistsProps> = ({
     <div>
       <ShowMore
         initial={ARTISTS_TO_DISPLAY}
-        variant={"lg-display"}
+        variant="lg-display"
```
olerichter00, line 36: Just a small fix.

dariakoko (approved):

PR #10914, opened 2022-09-01
feat: add i18n support in tests

gkartalis:

### Description

As we were pairing with @mdole we realized that we haven't setup i18n for tests, we came across it while developing a feature and we were trying to `queryByText` but we were seeing only the key rendered with `screen.debug()`.

This pr solves this, now i18n translations are going to be available to us when testing.

### DIFF
--- a/src/DevTools/setupTestWrapper.tsx
+++ b/src/DevTools/setupTestWrapper.tsx
@@ -13,12 +13,12 @@ import { MockResolvers } from "relay-test-utils/lib/RelayMockPayloadGenerator"
 import i18n from "System/i18n/i18n"
 
 // overide mount to provide access to i18n inside the setupTestWrapper function
-const mount = ui =>
-  enzMount(<I18nextProvider i18n={i18n}>{ui}</I18nextProvider>)
+const mount = children =>
+  enzMount(<I18nextProvider i18n={i18n}>{children}</I18nextProvider>)
 
 // overide render to provide access to i18n inside the setupTestWrapperTL function
-const render = ui =>
-  rtlRender(<I18nextProvider i18n={i18n}>{ui}</I18nextProvider>)
+const render = children =>
+  rtlRender(<I18nextProvider i18n={i18n}>{children}</I18nextProvider>)
 
 type SetupTestWrapper<T extends OperationType> = {
   Component: React.ComponentType<T["response"]>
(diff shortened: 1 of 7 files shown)


mdole (approved):
😍 🤘

lidimayra (approved):
Yay, thank you, @gkartalis and @mdole !!! ❤️ 

Also pinging @jonallured and @damassi  here for awareness as we were discussing the ideal approach to do this on [this thread](https://github.com/artsy/force/pull/10645#discussion_r934434815)

gkartalis:
I will further look into it tomorrow since it is #ff

src/tests.ts
```diff
@@ -4,6 +4,8 @@ import Adapter from "@wojtekmaj/enzyme-adapter-react-17"
 import "regenerator-runtime/runtime"
 import { format } from "util"
 import "@testing-library/jest-dom"
+// this import makes sure that the translations are accessible in the test environment
+import "System/i18n/i18n"
```
damassi, line 8: 🙏

damassi (approved):

src/tests.ts
araujobarret: Thanks for the solution! Can you check if we had previous tests mocking it that don't need it anymore?

gkartalis:
Went through and:

- [x] removed the references of the import in other components
- [x] tried out a different approach in order to prevent the tests from breaking

src/DevTools/setupTestWrapper.tsx
```diff
@@ -9,6 +10,15 @@ import {
   RelayMockEnvironment,
 } from "relay-test-utils"
 import { MockResolvers } from "relay-test-utils/lib/RelayMockPayloadGenerator"
+import i18n from "System/i18n/i18n"
+
+// overide mount to provide access to i18n inside the setupTestWrapper function
+const mount = ui =>
```
araujobarret, line 16: nit: call it children instead of ui

araujobarret (dismissed):
Great job!

lidimayra (approved):
Many thanks for putting efforts into improving our i18n setup!! 💟 🚀

dimatretyak (approved):
🏎🏎🏎

mdole (approved):

PR #10880, opened 2022-08-30
fix(CX-2855): Update sell through rate in MyC Artist Market

MrSltun:
The type of this PR is: **Fix**

This PR solves [CX-2855]

### Description
This PR updates the Sell Through Rate value in the artist market in My Collection artwork details page

|before|after
|--|--|
|image | image |

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CX-2855]: https://artsyproduct.atlassian.net/browse/CX-2855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #10915, opened 2022-09-01
fix: Adjust MyC insights sell-through rate formatting

olerichter00:
Addresses https://github.com/artsy/force/pull/10880#pullrequestreview-1093488341

## Description

Adjust MyC insights sell-through rate formatting.

| Before | After |
| --- | --- |
| Screenshot 2022-09-01 at 16 28 06|Screenshot 2022-09-01 at 16 33 02 |

### DIFF
--- a/src/Apps/Artwork/Utils/insightUtils.ts
+++ b/src/Apps/Artwork/Utils/insightUtils.ts
@@ -1,0 +1,9 @@
+export const formatSellThroughRate = (sellThroughRate: number | null) => {
+  if (sellThroughRate === null) {
+    return ""
+  }
+
+  const sellThroughRatePercentage = (sellThroughRate as number) * 100
+  // show up to 2 decimal places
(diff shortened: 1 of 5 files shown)


src/Apps/Artwork/Utils/insightUtils.ts
```diff
@@ -0,0 +1,9 @@
+export const formatSellThroughRate = (sellThroughRate: number | null) => {
```
MounirDhahri, line 1: **note:** I added a ticket to add this to metaphysics in the future and update it in Eigen.

MounirDhahri (approved):

src/Apps/Artwork/Utils/insightUtils.ts
olerichter00: Reading my mind! Let's merge these fixes right now and add the field to MP in the next step.

PR #10919, opened 2022-09-01
feat: don't show confirmation dialog

dariakoko:

This PR solves [CX-2870]

### Description

Don't show confirmation dialog when the user has not made any changes | edit/create artwork flow

[CX-2870]: https://artsyproduct.atlassian.net/browse/CX-2870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
@@ -15,6 +15,7 @@ import { BackLink } from "Components/Links/BackLink"
 import { MetaTags } from "Components/MetaTags"
 import { Sticky, StickyProvider } from "Components/Sticky"
 import { Form, Formik } from "formik"
+import { isEqual } from "lodash"
 import { useState } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { RouterLink } from "System/Router/RouterLink"
@@ -189,6 +190,15 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
     }
   }
 
+  const leaveForm = () => {
+    router.replace({ pathname: "/settings/my-collection" })
+    router.push({
+      pathname: isEditing
+        ? `/my-collection/artwork/${artwork.internalID}`
+        : "/settings/my-collection",
+    })
+  }
+
   return (
     <>
       <MetaTags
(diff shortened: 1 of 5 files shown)


src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -253,7 +256,16 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
...
                                 as={RouterLink}
-                                onClick={() => setShouldShowBackModal(true)}
+                                onClick={() => {
+                                  const formIsDirty = !isEqual(
+                                    initialValues,
+                                    values
+                                  )
+
+                                  return formIsDirty
+                                    ? setShouldShowBackModal(true)
```
MounirDhahri, line 266: **suggestion:** I know this was done in a different PR, but I feel like we improve the naming here. Maybe something like `showLeaveWithoutSavingModal` or `showQuitWithoutSavingModal`

src/Apps/MyCollection/Routes/EditArtwork/__tests__/MyCollectionArtworkForm.jest.tsx
```diff
@@ -421,6 +450,19 @@ describe("Create artwork", () => {
         pathname: "/settings/my-collection",
       })
     })
+
+    it("navigates to the previous screen when the form has not been changed", async () => {
```
MounirDhahri, line 454: **praise:** 👏👏👏

src/Apps/MyCollection/Routes/EditArtwork/__tests__/MyCollectionArtworkForm.jest.tsx
```diff
@@ -408,9 +430,16 @@ describe("Create artwork", () => {
   })
 
   describe("Back Link behavior", () => {
-    it("opens modal on press", async () => {
+    it("opens modal on press when the form has been changed", async () => {
```
MounirDhahri, line 433: **praise:** 👏

MounirDhahri (dismissed):
LGTM, I left a small non-blocking comment though that would be great if we address it

olerichter00 (commented):
Have you tried using Formik's render prop [dirty](https://formik.org/docs/api/formik#dirty-boolean) instead? If we could use this prop, we wouldn't have to add additional logic with an equality check.

```
dirty: boolean

Returns true if values are not deeply equal from initial values, false otherwise. dirty is a readonly computed property and should not be mutated directly.
```

MrSltun (dismissed):
Great work Daria! 🌟🌟🌟🌟🌟

olerichter00 (approved):

PR #10928, opened 2022-09-06
feat(tx-420): alican/payment method change from review step

AlicanAkyuz:

This PR solves [TX-420]

### Description

This PR allows that when a buyer changes payment method in the review step and goes back to the payment step, they see their unsaved bank account pre-selected. 

https://user-images.githubusercontent.com/42584148/188600027-20554121-3e8e-4d9e-ab60-9cc063e54fcc.mov

[TX-420]: https://artsyproduct.atlassian.net/browse/TX-420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -52,11 +52,7 @@ export const BankAccountPicker: FC<Props> = props => {
 
   const bankAccountsArray: BankAccountRecord[] = extractNodes(bankAccounts)
 
-  if (
-    order.paymentMethodDetails &&
-    order.paymentMethodDetails.internalID &&
-    order.paymentMethodDetails.last4
-  ) {
+  if (order?.paymentMethodDetails?.internalID) {
     // if account on order is not saved on user's profile
     const isOrderBankSaved = bankAccountsArray.find(
       bank => bank.last4 === order.paymentMethodDetails?.last4
@@ -101,7 +97,7 @@ export const BankAccountPicker: FC<Props> = props => {
 
   return (
     <>
-      {bankAccountsArray && bankAccountsArray.length > 0 && (
+      {bankAccountsArray?.length > 0 && (
         <RadioGroup
           data-test="bankAccounts"
           onSelect={val => {
@@ -128,7 +124,7 @@ export const BankAccountPicker: FC<Props> = props => {
 
               return (
                 <BorderedRadio value={internalID} key={internalID}>
-                  <BankDebitDetails last4={last4!} />
+                  <BankDebitDetails last4={last4} />
                 </BorderedRadio>
               )
             })
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -1,4 +1,4 @@
-import { FC, useState, useEffect } from "react"
+import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 
 import { BankDebitProvider } from "Components/BankDebitForm/BankDebitProvider"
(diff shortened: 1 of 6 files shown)


src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -45,7 +50,23 @@ export const BankAccountPicker: FC<Props> = props => {
...
+    // if account on order is not saved on user's profile
+    const isOrderBankSaved = bankAccountsArray.find(
+      bank => bank.last4 === order.paymentMethodDetails?.last4
+    )
+
+    if (!isOrderBankSaved) {
+      // populate banks array with the account on order
+      bankAccountsArray.unshift(order.paymentMethodDetails as BankAccountRecord)
+    }
+  }
```
erikdstock, line 69: Responding to https://github.com/artsy/force/pull/10928#discussion_r963843402 which is not appearing in the diff, my understanding of what is going on here:

- for the first `if` to enter this code,  `order.paymentMethodDetails` must have an `internalID` and a `last4` - is this because `last4` is unique to bank accounts, so we avoid checking on other payment methods? If so I would make that assumption more explicit, maybe by checking payment method directly.
- `isOrderBankSaved` on L60 makes sense, but I think testing `internalID` makes more sense if it is available on both objects.

Question on L67: Is there any chance that mutating this object could cause strange behavior since `bankAccountsArray` is sourced from relay? I'm thinking the answer is no, since even if the code ran multiple times we would be working with a new array (from `extractNodes`) each time. Probably nothing.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -176,6 +197,12 @@ export const BankAccountPickerFragmentContainer = createFragmentContainer(
         internalID
         mode
         bankAccountId
+        paymentMethodDetails {
+          ... on BankAccount {
+            internalID
+            last4
+          }
+        }
```
erikdstock, line 205: Ah, i now see we are only querying bank account payment methods.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -80,7 +101,7 @@ export const BankAccountPicker: FC<Props> = props => {
 
   return (
     <>
-      {bankAccountsArray.length > 0 && (
+      {bankAccountsArray && bankAccountsArray.length > 0 && (
```
erikdstock, line 104: is it possible for `me.bankAccounts` (the source of this variable) to be null/undefined? The types above seem to suggest it is always present.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -107,7 +128,7 @@ export const BankAccountPicker: FC<Props> = props => {
 
               return (
                 <BorderedRadio value={internalID} key={internalID}>
-                  <BankDebitDetails last4={last4} />
+                  <BankDebitDetails last4={last4!} />
```
erikdstock, line 131: Why is this non-null assertion necessary given the `BankAccountRecord` interface?

erikdstock (commented):
A few non-blocking/drive-by questions, makes sense from here but interested in your thoughts!

damassi:
@AlicanAkyuz - that peril bot is gonna keep complaining on every PR until the title is formatted exactly as it expects :) 

```
chore(context): Description 
```

src/Apps/Order/Components/BankAccountPicker.tsx
AlicanAkyuz: @erikdstock, Rachel's comment has a point. I initially made the fields in BankAccountRecord interface nullable (`internalID: string | null...`), therefore I was checking the presence of the fields in `paymentMethodDetails`, but ut's unnecessary now. will remove.

AlicanAkyuz: what's going on here is that: 

L53: we extract the saved bank accounts on user's profile into `bankAccountsArray`

L55: we check if order has a bankAccount set on it as the paymentMethod (it can only be bank account at this point as you recognised in L201) 

L61: we are trying to find if the bankAccount on order is a saved bank account on user's profile 

L65: if the bank account on order is not saved in users's profile, we are placing it in the beginning of `bankAccountsArray`, so that when user goes to the Review step, clicks on 'Change payment method' and comes back to the payment page, we are able to display the unsaved bank account that they initially set on Order.

src/Apps/Order/Components/BankAccountPicker.tsx
AlicanAkyuz: `bankAccounts` on `me` can be an array or null. we then extract this `bankAccounts` coming from the API into `bankAccountsArray` which can also include the unsaved bank account. I'll shorten this to` bankAccountsArray?.length` 👍

src/Apps/Order/Components/BankAccountPicker.tsx
AlicanAkyuz: it's actually not, good catch! will remove.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -45,7 +50,23 @@ export const BankAccountPicker: FC<Props> = props => {
     clientSecret,
   } = props
 
-  const bankAccountsArray = extractNodes(bankAccounts)
+  const bankAccountsArray: BankAccountRecord[] = extractNodes(bankAccounts)
+
+  if (
+    order.paymentMethodDetails &&
+    order.paymentMethodDetails.internalID &&
+    order.paymentMethodDetails.last4
```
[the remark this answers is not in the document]
AlicanAkyuz: good catch. I initially made the fields in BankAccountRecord interface nullable but it's unnecessary now. removed 👍

PR #10933, opened 2022-09-06
fix: fix admin/clear-cache route

mdole:

### Description

While [investigating](https://artsy.slack.com/archives/CCS8TMQ0K/p1662478073254789) a shortcut that had been updated and is seemingly very cached, discovered that the `/admin/clear-cache` route & functionality was broken. 

This fixed it when running locally, though I will freely admit that I'm not sure if it's the "correct" way to fetch a user in Force nowadays. LMK if I goofed!


damassi:
Looks good!

PR #10952, opened 2022-09-08
fix: Block text injection on settings page GRO-1172

jonallured:
This PR adds a layer of text injection protection that we were lacking. It updates two sides of a handshake so that a bad actor can't populate our error toast via the URL. Here's an example of the intended usage:

Screen Shot 2022-09-08 at 3 19 54 PM

Examples of abuses:

```
http://localhost:4000/settings/edit-settings?error=Please%20contact%20support%20at%20bad-actor@example.com
http://localhost:4000/settings/edit-settings?error=already-linked&provider=Please%20contact%20bad-actor@example.com
http://localhost:4000/settings/edit-settings?error=already-linked&provider=Google%20Please%20contact%20bad-actor@example.com
```

I have verified each of these does nothing with this new blocking approach. 👍 

Shout out to @joeyAghion for helping me see that this was an easy way to solve the problem! 🤗 

https://artsyproduct.atlassian.net/browse/GRO-1172

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsLinkedAccounts.tsx
+++ b/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsLinkedAccounts.tsx
@@ -24,6 +24,8 @@ interface SettingsEditSettingsLinkedAccountsProps {
   me: SettingsEditSettingsLinkedAccounts_me
 }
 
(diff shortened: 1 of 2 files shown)


damassi:
Well that was quick and easy!

damassi (approved):

PR #10970, opened 2022-09-12
feat[FX-4095]: use curated collection pages instead of genes

mdole:

This PR solves [FX-4095]

### Description

This PR accomplishes a few things:
- Update navigation links to point to collections instead of gene pages (e.g. `/collection/trove-editors-picks` instead of `/gene/trove`)
- Update "curated genes" to redirect to curated collections (e.g. navigating to `/gene/trove` will redirect you to `/collection/trove-editors-picks`)
- Replace genes in onboarding with collections (Paired with Eigen version [here](https://github.com/artsy/eigen/pull/7396))
- Update homepage Trove rail to pull works from a collection and direct clicks to the collection instead of to the gene

[FX-4095]: https://artsyproduct.atlassian.net/browse/FX-4095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

For the full plan for launching these changes, see [Notion](https://www.notion.so/artsy/Curated-collections-launch-plan-8181a3ef06884b66a6b71664e65e5598#ecc09d24f7744e1ba990e8a938bc1400)

### DIFF
--- a/src/Apps/Gene/Server/__tests__ /redirectGeneToCollection.jest.ts
+++ b/src/Apps/Gene/Server/__tests__ /redirectGeneToCollection.jest.ts
@@ -1,0 +1,35 @@
+import { redirectGeneToCollection } from "../redirectGeneToCollection"
+
+describe("redirectGeneToCollection", () => {
+  it("does not redirect for a non-migrated gene", () => {
+    const req = {
+      params: {
+        slug: "contemporary",
+      },
+    }
+    const res = {
+      redirect: jest.fn(),
+      locals: { sd: {} },
+    }
+    redirectGeneToCollection({ req, res })
+    expect(res.redirect).not.toHaveBeenCalled()
+  })
+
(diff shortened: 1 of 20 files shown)


gkartalis (approved):
Looking good! Left one question about the `destination_page_owner_id`.

Another non blocking thought what do you think of changing [line 93](https://github.com/artsy/force/pull/10970/files#diff-4d0c3ec0e9992a9a4c0570eb8f3a13bfd996c104d9e706e8a9ab0885ebf7427aR93) to be instead of string something like `"trove-slug" | "other-slugs"` ?

The reason for this is that we are hardcoding the strings and a small typo might cause the collection to not be rendered if we change anything there 🤔 


Happy to hear your thoughts on this

PR #10980, opened 2022-09-14
feat:(FX-4219): edition selection artwork screen

gkartalis:

This PR solves [FX-4219]

### Description

> ⚠️ Feature is behind a feature flag

Adds Edition selection in artwork screen, functionality and tests will be added on next ticket ([FX-4149](https://artsyproduct.atlassian.net/browse/FX-4149))

#### Screenshots

|Desktop| mWeb|
|---|---|
|Screenshot 2022-09-14 at 17 22 48|Screenshot 2022-09-14 at 17 23 26|

[FX-4219]: https://artsyproduct.atlassian.net/browse/FX-4219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
@@ -14,6 +14,7 @@ import { useTimer } from "Utils/Hooks/useTimer"
 import { useState } from "react"
 import { lotIsClosed } from "Apps/Artwork/Utils/lotIsClosed"
 import { useAuctionWebsocket } from "Components/useAuctionWebsocket"
+import { ArtworkSidebar2EditionSetFragmentContainer } from "./ArtworkSidebar2EditionSets"
 
 export interface ArtworkSidebarProps {
   artwork: ArtworkSidebar2_artwork
@@ -74,6 +75,9 @@ export const ArtworkSidebar2: React.FC<ArtworkSidebarProps> = props => {
 
       <Spacer mt={2} />
       <ArtworkSidebar2DetailsFragmentContainer artwork={artwork} />
+      <Separator />
+
+      <ArtworkSidebar2EditionSetFragmentContainer artwork={artwork} />
 
       {!isSold && artworkEcommerceAvailable && (
         <>
@@ -123,6 +127,7 @@ export const ArtworkSidebar2FragmentContainer = createFragmentContainer(
         ...ArtworkSidebar2ArtworkTitle_artwork
         ...ArtworkSidebar2Artists_artwork
         ...ArtworkSidebar2Details_artwork
+        ...ArtworkSidebar2EditionSets_artwork
         ...ArtworkSidebar2ShippingInformation_artwork
         ...ArtworkSidebar2PartnerInfo_artwork
         ...ArtworkSidebar2CreateArtworkAlert_artwork
--- a/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
@@ -74,7 +74,6 @@ export const ArtworkSidebar2: React.FC<ArtworkSidebarProps> = props => {
 
       <Spacer mt={2} />
       <ArtworkSidebar2DetailsFragmentContainer artwork={artwork} />
-      <Spacer mt={2} />
 
       {!isSold && artworkEcommerceAvailable && (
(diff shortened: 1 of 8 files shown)


src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
```diff
@@ -74,7 +75,9 @@ export const ArtworkSidebar2: React.FC<ArtworkSidebarProps> = props => {
 
       <Spacer mt={2} />
       <ArtworkSidebar2DetailsFragmentContainer artwork={artwork} />
-      <Spacer mt={2} />
+      <Separator />
```
dimatretyak, line 78: **question**: won't we have a problem with the fact that `Spacer` will be rendered twice ([here](https://github.com/artsy/force/blob/a7e04afef41e8cee76dc9fd09e640bcaa364b643/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx#L78) and [here](https://github.com/artsy/force/blob/a7e04afef41e8cee76dc9fd09e640bcaa364b643/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx#L95)) if there is nothing to render in edition sets and `artworkEcommerceAvailable` is `false`?

dimatretyak (approved):

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2SizeInfo.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  return (
+    <React.Fragment data-testid="size-info">
+      <Media greaterThanOrEqual="md">
+        <Box color="black60">
+          <Text variant="sm-display">
+            {hasInDimensions && dimensions.in}{" "}
+            {hasInDimensions && hasCmDimensions && "|"}{" "}
+            {hasCmDimensions && dimensions.cm}
```
dimatretyak, line 37: **suggestion**: move it to helper method

```javascript
const renderDimensions = () => {
  if (hasInDimensions && hasCmDimensions) {
    return `${dimensions.in} | ${dimensions.cm}`
  }

  return dimensions?.cm ?? dimensions?.in
}
```

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2EditionSets.tsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  return (
+    <RadioGroup>
+      {
+        editionSets?.map((editionSet, index) => (
+          <React.Fragment key={editionSet?.id}>
+            <Box py={2}>{renderEditionSet(editionSet)}</Box>
+            {index !== editionSets.length - 1 && <Separator />}
```
dimatretyak, line 85: **question**: maybe we can use [Join](https://palette.artsy.net/elements/join/) component here?

dimatretyak:
will this behavior be changed in the future?

for-george

gkartalis:
> will this behavior be changed in the future and it won't be rendered like this (when edition set is 0 or 1)?

Yes, it will be changed in the next pr, sorry had to specify that in the description

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2.tsx
gkartalis: I will open a follow up pr after all the tickets are addressed fixing all these 🤞

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2EditionSets.tsx
gkartalis: Unfortunately we can't use join here since the separator from Join has to be an element, and in our case when the condition in line 85 is false we don't render one and we get this type error:

`Type 'Element | undefined' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.`
or
`Type 'Element | null' is not assignable to type 'ReactElement<any, string | JSXElementConstructor<any>>'.`

dimatretyak: got it. thanks for checking!

dimatretyak (approved):

PR #10981, opened 2022-09-14
feat: User sees show events on the info page GRO-1024

jonallured:
This PR adds back something that was removed when we re-built the more info page for shows: `PartnerShowEvent` records. Here's the Figma I was following:

https://www.figma.com/file/k465puwy6YQOcEESEkppni/Smaller-Tasks-2022?node-id=24%3A952

What I did here was update the GQL to include the `events` array and then I broke out an `EventList` component. I didn't start with that component extracted but then when it came time to test it was handy to have a way to filter down to that part of the page.

<details>

<summary>Heres some screenshots</summary>

Screen Shot 2022-09-14 at 10 53 36 AM
Screen Shot 2022-09-14 at 10 53 10 AM

</details>

https://artsyproduct.atlassian.net/browse/GRO-1024

/cc @artsy/grow-devs @starsirius

### DIFF
--- a/src/Apps/Show/Routes/ShowInfo.tsx
+++ b/src/Apps/Show/Routes/ShowInfo.tsx
@@ -24,6 +24,8 @@ export const ShowInfo: React.FC<ShowInfoProps> = ({
   show,
   show: { about, pressRelease, partner, hasLocation },
 }) => {
+  const events = show.events || []
+
   return (
     <>
       <Text as="h1" variant="xl" my={4}>
@@ -62,6 +64,8 @@ export const ShowInfo: React.FC<ShowInfoProps> = ({
                 </HTML>
               </Box>
             )}
+
+            {events.length > 0 && <EventList events={events} />}
           </Join>
         </Column>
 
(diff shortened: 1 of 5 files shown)


src/Apps/Show/Routes/ShowInfo.tsx
```diff
@@ -95,6 +99,36 @@ export const ShowInfo: React.FC<ShowInfoProps> = ({
   )
 }
 
+const EventList = ({ events }) => {
```
damassi, line 102: Can we add a type here?

TMcMeans:
would it make sense to configure the dateTimeRange to show the local time instead of UTC?

TMcMeans:
Also, just surfacing that the CMS doesn't have guards in place for the description, title and event type so partners might put random stuff in that will be shown on the page.

jonallured:
Yeah that's a good call out @TMcMeans about the timezone. This field in MP is hardcoded to UTC:

https://github.com/artsy/metaphysics/blob/main/src/schema/v2/show_event.ts#L28

So I guess I'm not sure what to do here - any thoughts @damassi? What do we typically do with regard to timezones in Force? 🤔

src/Apps/Show/Routes/ShowInfo.tsx
jonallured: Done!

TMcMeans:
isnt there a function in force called `DateTimeFormat()` that does just that?? @damassi correct me here if I'm wrong

TMcMeans:
im looking at some of the auctions sale specs and it seems like there are formattedStart & endTimes that can be fetched and rendered client-side but I dunno if the show events have any of those properties.

jonallured:
Well I found this: https://github.com/artsy/force/blob/main/src/Apps/Auction/Components/AuctionDetails/AuctionDetailsStartTime.tsx#L12-L15

And I read that file but I don't see how this is done client-side and where we pass the user's timezone?

TMcMeans:
yep, that's the file I was looking in but those formatted times are fetched from the query, and then rendered client-side.

jonallured:
Screen Shot 2022-09-21 at 3 39 28 PM

Ok with https://github.com/artsy/metaphysics/pull/4396 I was able to update the logic in MP so that a user's timezone is taken into account - see screenshot above! I'm going to merge this and get it out to production. 😅

damassi (approved):
Looks great, and no weird client-side timezone fetch mess either (like on /auction/id) 👍

PR #10983, opened 2022-09-14
feat: Increase the max number of works per artist for new for you GRO-1297

jonallured:
This PR updates both the rail on the homepage and the actual /new-for-you page so that they will show more works per artist. I considered extracting a constant for this but then changed my mind. 🤷 

https://artsyproduct.atlassian.net/browse/GRO-1297

/cc @artsy/grow-devs @shaq31415926 @isakelaris

### DIFF
--- a/src/Apps/Home/Components/HomeNewWorksForYouRail.tsx
+++ b/src/Apps/Home/Components/HomeNewWorksForYouRail.tsx
@@ -116,7 +116,11 @@ export const HomeNewWorksForYouRailQueryRenderer: React.FC = () => {
       environment={relayEnvironment}
       query={graphql`
         query HomeNewWorksForYouRailQuery {
-          artworksForUser(includeBackfill: true, first: 20) {
+          artworksForUser(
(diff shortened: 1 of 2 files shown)


src/Apps/Home/Components/HomeNewWorksForYouRail.tsx
```diff
@@ -116,7 +116,11 @@ export const HomeNewWorksForYouRailQueryRenderer: React.FC = () => {
       environment={relayEnvironment}
       query={graphql`
         query HomeNewWorksForYouRailQuery {
-          artworksForUser(includeBackfill: true, first: 20) {
+          artworksForUser(
+            includeBackfill: true
+            first: 20
+            maxWorksPerArtist: 3
```
pam-, line 122: +1 to the point you made in the PR desc about setting a constant! in case we gotta change this again in the future

pam- (approved):
smol comment! otherwise :shipit:

PR #10985, opened 2022-09-14
feat: User sees complete rules for password and better validations GRO-1277

jonallured:
This PR started with a simply copy change so that users see all the rules for a valid password. Then I also update the Yup schema for the password field to throw errors when those rules are violated and provide better error messages to the user. This happens client-side so the user doesn't have to wait for the server to tell them something was wrong.

I then wrote unit tests that use this Yup schema to demonstrates the various ways a password can be invalid and finally a case for a password that follows the rules.

<details>

<summary>Here's some screenshots</summary>

Screen Shot 2022-09-14 at 3 33 00 PM
Screen Shot 2022-09-14 at 3 33 30 PM
Screen Shot 2022-09-14 at 3 33 37 PM
Screen Shot 2022-09-14 at 3 33 40 PM
Screen Shot 2022-09-14 at 3 33 47 PM

</details>

https://artsyproduct.atlassian.net/browse/GRO-1277

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Authentication/Routes/__tests__/ResetPasswordRoute.jest.tsx
+++ b/src/Apps/Authentication/Routes/__tests__/ResetPasswordRoute.jest.tsx
@@ -49,11 +49,11 @@ describe("ResetPasswordRoute", () => {
 
   it("resets the password and redirects", async () => {
     fireEvent.change(screen.getByPlaceholderText("New password"), {
-      target: { name: "password", value: "secretsecret" },
+      target: { name: "password", value: "Secretsecret1" },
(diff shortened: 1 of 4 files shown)


jonallured:
Note too that I wondered if these changes could blow up Cypress so I poked around and I _think_ it's fine? I see that we set an env var for the password that's typed into forms there. I checked the value and it's a random string of alpha/numeric that should satisfy the client side checks. So yeah I think it'll be safe?

damassi (approved):
A+

PR #10986, opened 2022-09-15
fix(FX-4233): encode search term for tabs

dimatretyak:

This PR solves [FX-4233]

### Steps to Reproduce
* In the global search bar, enter a search term that contains an ampersand e.g. art sales & research 
* Disregard the dropdown and press Enter for full-page search results
* Note that the Artworks tab should be highlighted by default but isn’t
* Select any other tab
* Note that the # of results has changed, and that there is an un-escaped ampersand in the URL bar

### Current Outcome
* Tab highlighting is inconsistent
* Total hit counts is inconsistent between tabs

### Expected Outcome
* Artworks tab is always highlighted initially
* Hit counts is stable between tabs


anandaroop (approved):
Thank you!

PR #10987, opened 2022-09-15
Fix: show an error message when user types an artist that isn't not on artsy desktop

dariakoko:

This PR solves [CX-2904]

### Description

[Figma](https://www.figma.com/file/AF7eVKe0eErR5gB6DDRiB2/My-Collection-Web?node-id=2063%3A61703)
When the query returns an empty array - set an error. The error message is defined in the MyCollectionArtworkDetailsValidationSchema.
The Banner content is centred and not left aligned (as per design) because this is a standard Palette component.

Note: this PR fixed the same bug on SWA flow as well

image
image


MounirDhahri:
@dariakoko iydm, can you also share a video for the issue here - to see it in action

dariakoko:
Sure @MounirDhahri 
Added to the description

MounirDhahri:
Niiice awesome

MounirDhahri (approved):
LGTM - thanks for adding the video

dariakoko:
I'm thinking, the user will see the list only if he enters two letters, should we add another error message in this case? @MounirDhahri

MounirDhahri:
I think it's fine as it is now since it's been working well for the SWA flow - maybe you can confirm with product as well but I think it's okay.

MrSltun (approved):
Great work Daria! 🌟 🌟

PR #10998, opened 2022-09-16
test: remove old relay test tools part IV

araujobarret:

### Description

Sibling of #10344
Removes old relay testing tool from Shipping route tests under Order app. 
We still have 4 more files using it that are going to be removed in the next PRs, we hope that unlocks Relay 13 upgrade

Note: this was a tricky one to refactor, this kind of integration test when needed to be refactored becomes intense. I would say that raises the flag of maybe moving it to smoke tests since it's a very important collector happy flow.


src/Apps/Order/Utils/shippingUtils.ts
```diff
@@ -154,7 +154,7 @@ export const getDefaultShippingQuoteId = (order: Shipping_order) => {
 
   const cleanQuotes = compact(shippingQuotes?.edges?.map(quote => quote?.node))
 
-  return cleanQuotes ? cleanQuotes[0].id : undefined
```
araujobarret, line 157: idk what kind of magic our old relay tool was doing, but happens cleanQuotes or is an empty array or with elements, never null/undefined, so this error popped once I removed the old testing tools. So far no issues/bugs were introduced

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -560,9 +560,11 @@ export const ShippingRoute: FC<ShippingProps> = props => {
     shippingQuotes.length > 0 &&
     !shippingQuoteId
 
+  // TODO: consider to move this block to a useEffect
   if (useDefaultArtsyShippingQuote) {
     const defaultShippingQuoteId = getDefaultShippingQuoteId(order)
-    setShippingQuoteId(defaultShippingQuoteId)
+    shippingQuoteId !== defaultShippingQuoteId &&
```
araujobarret, line 566: this is just a control to avoid loops, I've tried moving to a useEffect this block but broke other tests, just flagging for future interactions with this file

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -192,346 +231,435 @@ describe("Shipping", () => {
...
       await page.selectPickupOption()
       fillInPhoneNumber(page.root, { isPickup: true, value: "2813308004" })
-      expect(mutations.mockFetch).not.toHaveBeenCalled()
+      expect(mockCommitMutation).not.toHaveBeenCalled()
+
       await page.clickSubmit()
-      expect(mutations.mockFetch).toHaveBeenCalledTimes(1)
-      expect(mutations.lastFetchVariables.input.fulfillmentType).toBe("PICKUP")
+      expect(mockCommitMutation).toHaveBeenCalledTimes(1)
+      expect(mockCommitMutation).toHaveBeenCalledWith(
```
araujobarret, line 412: since we don't have the custom variables to watch we only check if the mocked mutation is being called with the right values

damassi (approved):

PR #10999, opened 2022-09-16
chore (CX-2731): Add tracking to My Collection

MrSltun:

This PR solves [CX-2731]

### Description
This PR adds tracking to My Collection
This is a follow up from artsy/cohesion#354

[CX-2731]: https://artsyproduct.atlassian.net/browse/CX-2731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -116,6 +124,11 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
         artworkLocation: values.artworkLocation,
       })
 
+      // Adding tracking after a successfully adding an artwork
+      if (!isEditing) {
+        trackEvent(tracks.saveCollectedArtwork())
+      }
```
olerichter00, line 130: **suggestion:** I could imagine that the most important info here is that we don't track if the user saves an artwork after editing. What do you think about emphasizing this in the comment? For example like this:

```suggestion
      //  Track saving only when the user adds a new artwork.
      if (!isEditing) {
        trackEvent(tracks.saveCollectedArtwork())
      }
```

olerichter00 (approved):
Looks good 💯

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -433,3 +449,23 @@ export const MyCollectionArtworkFormFragmentContainer = createFragmentContainer(
     `,
   }
 )
+
+const tracks = {
```
damassi, line 453: One pattern worth mentioning is that we've been moving tracking into a single hook lately, similar to 
- https://github.com/artsy/force/blob/main/src/Apps/Article/useArticleTracking.ts 
- https://github.com/artsy/force/blob/main/src/Apps/Auction/Hooks/useAuctionTracking.ts

Then its easy to mock the whole file to test, and there's one centrally located place to change, add, adjust tracking through time (versus having it littered throughout app. It's been working really nicely.

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -49,6 +50,10 @@ export interface MyCollectionArtworkFormProps {
 export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = ({
   artwork,
 }) => {
+  const {
+    deleteCollectedArtwork,
+    saveCollectedArtwork,
```
MounirDhahri, line 55: **suggestion:** do we maybe have some patterns around naming here, this might be confusing to reviewers or someone reading the code later who didn't see where the method is coming from. Maybe something like:
```suggestion
    deleteCollectedArtwork: trackDeleteCollectedArtwork,
    saveCollectedArtwork: trackSaveDeletedArtwork,
    saveCollectedArtwork,
```

src/Apps/MyCollection/Routes/EditArtwork/__tests__/MyCollectionArtworkForm.jest.tsx
```diff
@@ -400,6 +411,15 @@ describe("Edit artwork", () => {
         })
       )
 
+      expect(trackingSpy).toHaveBeenCalledWith({
```
MounirDhahri, line 414: **suggestion:** although it's okay here to compare object, comparing snapshots as well would be also fine here

MounirDhahri (commented):
Niice, thanks for wrapping this work! I only left a minor comment about naming

olerichter00 (approved):
Looks great 🌟

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -49,6 +50,10 @@ export interface MyCollectionArtworkFormProps {
 export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = ({
   artwork,
 }) => {
+  const {
+    deleteCollectedArtwork: trackDeleteCollectedArtwork,
+    saveCollectedArtwork: trackSaveCollectedArtwork,
+  } = useMyCollectionTracking()
```
damassi, line 56: 🙏 tyty

MounirDhahri (approved):

PR #11011, opened 2022-09-19
test: remove old relay test tools part V

araujobarret:

### Description

Removes old relay testing tool from reminiscent components.


src/Apps/IdentityVerification/__tests__/IdentityVerificationApp.jest.tsx
```diff
@@ -1,32 +1,37 @@
...
-        <IdentityVerificationAppFragmentContainer {...props} />
-      </ToastsProvider>
-    ),
+describe("IdentityVerification route", () => {
+  let sendProps
+  let trackEvent
+
+  const { getWrapper } = setupTestWrapper({
+    Component: (props: any) => {
+      const propsFiltered = sendProps ? props : {}
```
araujobarret, line 25: this is a bit hackish but as far as I tested our current relay setup is not allowing to mock resolver types as `null/undefined` 🤷

src/Apps/Order/Components/__tests__/SavedAddresses.jest.tsx
```diff
@@ -204,9 +170,7 @@ describe("Saved Addresses", () => {
       expect(modal.props().show).toBe(false)
       button.simulate("click")
 
-      setTimeout(() => {
-        expect(modal).toHaveLength(1)
-      }, 0)
+      expect(modal).toHaveLength(1)
```
araujobarret, line 173: I really don't know how this kind of misusage assertion inside a timeout passed through review, we should be aware that doesn't work and make the test assertion a false positive 🚓

src/DevTools/setupTestWrapper.tsx
```diff
@@ -138,11 +138,11 @@ export const setupTestWrapper = <T extends OperationType>({
 }: SetupTestWrapper<T>) => {
   const getWrapper = (
     mockResolvers: MockResolvers = {},
-    componentProps: {} = {}
+    componentProps: {} = {},
+    mockedEnv?: ReturnType<typeof createMockEnvironment>
   ) => {
-    const env = createMockEnvironment()
+    const env = mockedEnv ?? createMockEnvironment()
```
araujobarret, line 144: the only change in our relay tools I've had to make, we'll revisit that in the future for sure as soon as we introduce hooks and have mixed Fragment/Hooks/etc in the same file, the test probably will need to control the env outside for handling data/dealing with mutations, this is open to suggestion but we should enforce the same pattern to use in all our React-Relay fronts, Force, Eigen, etc.

damassi (approved):
🚀

PR #11024, opened 2022-09-21
feat: Add CWA header Get in Touch CTA functionality

olerichter00:
Addresses [CX-2937]

## Description

Add CWA header Get in Touch CTA functionality.

## Follow-Up: 

Enable `get-in-touch-flow-web` [feature flag in Unleash](https://unleash.artsy.net/projects/default/features/get-in-touch-flow-web)

[CX-2937]: https://artsyproduct.atlassian.net/browse/CX-2937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Consign/Routes/MarketingLanding/Components/Header.tsx
```diff
@@ -25,6 +23,17 @@ export const Header: React.FC = () => {
...
+  const trackGetInTouchClick = () => {
+    trackEvent({
+      action: "clickedGetInTouch",
+      context_module: "Header",
+      context_page_owner_type: contextPageOwnerType,
+      label: "Get in Touch",
+      user_id: user?.id,
+      user_email: user?.email,
+    })
+  }
```
MounirDhahri, line 35: **note:** I think we can keep this as it is now since the screen also contains other hard typed action names, I would keep an eye on this in the future in case we extend it - if so we will need to migrate those actions to cohesion

MounirDhahri (approved):

src/Apps/Consign/Routes/MarketingLanding/Components/Header.tsx
olerichter00: Good point! We could also use the [usePageNameTracking pattern](https://github.com/artsy/force/blob/main/src/Apps/Article/useArticleTracking.ts#L13). I can create a ticket to clean up tracking for the whole MarketingLanding app.

MounirDhahri: great, thanks Ole

PR #11038, opened 2022-09-23
feat(FX-4149): commercial and auction sections

gkartalis:

This PR solves [FX-4149]

### Description

Changes are behind a feature flag

review app [here](https://artwork-sidebar-redesign.artsy.net/)

Commercial section of the new artwork section
includes: 

- all commercial and auction buttons
- tracking
- mutations for the commercial actions

was QAed (4 Oct) and you can find the QA board [here](https://www.notion.so/artsy/QA-Artwork-Page-Web-mWeb-02c8adf99f5947c3ac7866c6a2bb4ade)

[FX-4149]: https://artsyproduct.atlassian.net/browse/FX-4149?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dblandin (approved):
Implementation looks good to me and we had a QA session covering all of the different artwork states yesterday so let's get this out on staging! 💯

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2AuctionTimer.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <>
+      {sale &&
+        saleArtwork &&
+        !lotIsClosed(sale, saleArtwork) &&
+        (sale?.cascadingEndTimeIntervalMinutes ? (
+          <>
+            <Spacer mt={2} />
+            {/* @ts-ignore RELAY UPGRADE 13 */}
```
damassi, line 26: @gkartalis - all of these ignores can be removed now!

gkartalis: woops good catch! will do a follow up removing everything that I missed

src/Apps/Artwork/Components/ArtworkSidebar2/__tests__/ArtworkSidebar2CommercialButtons.jest.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+jest.unmock("react-relay")
+
+describe("ArtworkSidebar2CommercialButtons", () => {
+  let user
+  beforeAll(() => {
+    mediator.on("open:auth", () => {})
+  })
+
+  let mockEnvironment = createMockEnvironment()
```
damassi, line 20: Curious - was this necessary to pass into `<MockBoot>`?

gkartalis: Talked this through with @araujobarret some days ago.

What we wanted to do here was to get access to the mutations that were called after the initial render. The problem was that we were getting the classic error no pending operations in the list. 

After adding the `mockEnvironment` to the MockBoot we were able to locate them with one caveat:
- the initial query was resolved by one environment (the one on setupTestWrapper) but all the rest were resolved by the other one (the one created in the test). 🤯

Also encountered another issue and have it on my list to check it out tomorrow: wasn't able to mock the router in this test.

What I want to do eventually is to migrate all the existing tests from [here](https://github.com/artsy/force/blob/main/src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercial.jest.tsx) to the new spec.

I have it on my list for tomorrow, so any advice is highly appreciated

src/DevTools/MockBoot.tsx
```diff
@@ -18,7 +27,7 @@ export const MockBoot: React.SFC<{
       headTags={headTags}
       context={mockContext as ClientContext}
       user={user}
-      relayEnvironment={null as any}
+      relayEnvironment={relayEnvironment ?? (null as any)}
```
gkartalis, line 30: @damassi and this is why I also added this here, because without this I wasn't able to have access to the mutation that happens when we press the commercial buttons

src/Apps/Artwork/Components/ArtworkSidebar2/__tests__/ArtworkSidebar2CommercialButtons.jest.tsx
damassi: Here's a Q: Is `MockBoot` necessary? Generally we can get away with testing our components in strict isolation, and then mocking out other component deps that might interfere. Is it because of `<Media>` usage? 

Otherwise, this is a really good-to-know. I haven't done too much mutation testing so I can't comment much further but I would continue pushing on as you have! Figuring out all of these quirks and streamlining our testing flows is 👍

damassi: cc @dzucconi, you might be interested here as well (and might have some ideas)

dzucconi: > What we wanted to do here was to get access to the mutations that were called after the initial render.

Not sure why that'd be necessary in a UI spec? I also don't see it in the specs below.

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2ArtsyGuarantee.tsx
```diff
@@ -1,14 +1,32 @@
...
+`
+const SecureLockIconBlue = styled(SecureLockIcon)`
+  .securelock-opening {
+    fill: ${themeGet("colors.brand")};
+  }
+`
+const VerifiedIconBlue = styled(VerifiedIcon)`
+  .verified-checkmark {
+    fill: ${themeGet("colors.brand")};
+  }
```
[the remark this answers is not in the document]
damassi: That was all removed in https://github.com/artsy/force/pull/11115

PR #11040, opened 2022-09-23
fix: issue related to image removal

MounirDhahri:

This PR solves issue related to images removal

### Description


MounirDhahri:
when removing images, we don't delete images from the local store, this update makes sure that this happens

PR #11052, opened 2022-09-27
fix(onboarding): Cap max height

damassi:

### Description

This fixes the issue where the modal was always expanding to fill 90% of the height on desktop: 

Screen Shot 2022-09-26 at 6 15 23 PM

cc @artsy/grow-devs


laurabeth (approved):
Looks great! :shipit:

PR #11058, opened 2022-09-27
fix(GRO-1242): prevents triggering analytics on route change

dzucconi:
Closes: [GRO-1242](https://artsyproduct.atlassian.net/browse/GRO-1242)

There's definitely a few places in the code that use `router.replace` not expecting it to trigger an analytics pageview. (And a few places where they want `push` or a proper link.)

I don't think this is the _correct_ fix. The correct way would be to switch the analytics tracking middleware to track on the `PUSH` event only. But as far as I can tell, it is never triggered (instead `UPDATE_LOCATION` is).

### DIFF
--- a/src/Utils/Hooks/useOnboardingModal.ts
+++ b/src/Utils/Hooks/useOnboardingModal.ts
@@ -3,10 +3,11 @@ import { useEffect } from "react"
 import { useOnboarding } from "Components/Onboarding"
 import { useSystemContext } from "System"
 import { useRouter } from "System/Router/useRouter"
+import { stringify } from "qs"
 
 export const useOnboardingModal = () => {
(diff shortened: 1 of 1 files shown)


dzucconi:
Hm, closing for a second. This miraculously interferes with modal close.

dzucconi:
lol: so the reason it does is because manipulating the URL like this doesn't alter internal router state. And the modal is triggered via the router state...

This honestly might be a not fixing situation.

damassi:
One possible approach is to add _yet another_ weird gotcha in the analyticsMiddleware (which needs to be cleaned up) to account for this one condition, and pass that condition along as history state from the onboarding hook

dzucconi:
Cursed

PR #11059, opened 2022-09-27
fix(GRO-1242): prevents triggering analytics on route change

dzucconi:
Closes: [GRO-1242](https://artsyproduct.atlassian.net/browse/GRO-1242)

Initially: https://github.com/artsy/force/pull/11058

Read that description; which is still valid. The problem with that approach is that manipulating the history state directly does not update anything in the routing store. The data in the routing store is used in triggering the modal in the first place.

Here I've just added an additional argument to exclude referrers from tracking 😵‍💫 — sorry.

### DIFF
--- a/src/Utils/Hooks/useOnboardingModal.ts
+++ b/src/Utils/Hooks/useOnboardingModal.ts
@@ -1,5 +1,5 @@
 import { omit } from "lodash"
-import { useEffect } from "react"
+import { useEffect, useRef } from "react"
 import { useOnboarding } from "Components/Onboarding"
 import { useSystemContext } from "System"
 import { useRouter } from "System/Router/useRouter"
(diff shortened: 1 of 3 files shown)


src/Utils/Hooks/useOnboardingModal.ts
```diff
@@ -19,14 +20,17 @@ export const useOnboardingModal = () => {
...
+      return
 
     showOnboarding()
 
     router.replace({
       ...match.location,
       query: omit(match.location.query, "onboarding"),
     })
+
+    initialized.current = true
```
dzucconi, line 33: This effect ran twice. Which again: shouldn't be a problem. In the case of the exclusion, it counts the referrer as `/undefined` on the second go round. Which is... interesting.

damassi (approved):
This is a reasonable solution!

PR #11062, opened 2022-09-27
fix(footer): Fix mobile footer caption

damassi:

### Description

This addresses the responsive footer, which wasn't tackled in the previous PR.


src/Components/Footer/FooterDownloadAppBanner.tsx
```diff
@@ -16,58 +16,57 @@ const MOBILE_COVER_IMAGE = resized(
...
+            height={320}
+            width="100%"
+            lazyLoad
+            alt=""
+            style={{ objectFit: "cover", objectPosition: "center top" }}
+          />
+        </Media>
+      </Column>
+      <Column span={12} order={[1, 2]}>
+        <Text variant="xs" fontStyle="italic" textAlign="right" mt={-10} pb={1}>
```
damassi, line 65: This negative margin was the only way i could figure out how to remove the row gap.

damassi: Went this way because the `order` prop was messing up the layout on mobile

dzucconi, line 65: (You can just use `-1`)

dzucconi: Just use a responsive rowGap?

src/Components/Footer/FooterDownloadAppBanner.tsx
```diff
@@ -16,58 +16,57 @@ const MOBILE_COVER_IMAGE = resized(
...
-          alignItems="center"
-          justifyContent="center"
-          order={[2, 1]}
-          px={2}
-          py={[6, 2]}
-        >
-          <Text variant="xl" textAlign="center" mb={1}>
-            Get the Artsy app
-          </Text>
+    <GridColumns borderBottom="1px solid" borderColor="black10">
```
dzucconi, line 19: `gridRowGap={[0, 2]}` or whatever

src/Components/Footer/FooterDownloadAppBanner.tsx
dzucconi: The `pb={1}` is also weird and I would avoid it. If you want external space, find the parent.

src/Components/Footer/FooterDownloadAppBanner.tsx
damassi: ah yeah much better

src/Components/Footer/FooterDownloadAppBanner.tsx
```diff
@@ -16,58 +16,62 @@ const MOBILE_COVER_IMAGE = resized(
...
-          py={[6, 2]}
-        >
-          <Text variant="xl" textAlign="center" mb={1}>
-            Get the Artsy app
-          </Text>
+    <GridColumns
+      gridRowGap={1}
+      borderBottom="1px solid"
+      borderColor="black10"
+      pb={1}
```
dzucconi, line 23: I mean the actual parent context.

dzucconi: (Consider how annoying it is to import a component with some kind of external spacing.)

damassi: Oh no I understand! in this case there's a border around the grid columns to define the footer; the border still represents the edge, but now that there's the caption we need some internal padding so that there's space between the image, caption and border edge. It's not an external margin for layout. 
 
 Screen Shot 2022-09-27 at 1 00 52 PM

dzucconi: OH, sorry I missed the border! Proceed!

src/Components/Footer/FooterDownloadAppBanner.tsx
damassi: ah, in the past negatives wouldn't apply to space scale, didn't know this was fixed

PR #11075, opened 2022-09-28
feat: preconnect webfonts and preload font css

dzucconi:
A minor thing but in my [non-scientific] testing this removed webfonts from being a factor in holding up the LCP.

The domain preconnect is incase we're on a page that has another font (mainly medium-italic, though it's rare). And preload the CSS because that's an obvious critical one we need.


damassi (approved):

jonallured (approved):
LGTM!! 🎉

PR #10747, opened 2022-08-12
feat(TX-517): Alican/Order shipping route to FC

AlicanAkyuz:
The type of this PR is: **Refactor**

This PR solves [TX-517]

### Description

This PR refactors /shipping route in Order App to a functional component. 

[TX-517]: https://artsyproduct.atlassian.net/browse/TX-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11076, opened 2022-09-28
fix (PX-5256): Remove duplicated call to setShipping mutation on address changes

jpotts244:
The type of this PR is: Fix

This PR solves [PX-5256](PX-5256)

### Description
Seeing the following error on checkout plus the "something went wrong" modal if:
- Order is Artsy Shipping
- on any address changes (addition of new address or selecting between multiple saved addresses)

Screen Shot 2022-09-28 at 3 48 56 PM

The error in console reads as:
```
instrument.js:109 Order/Routes/Shipping/index.tsx | Error: Mutliple simulataneous mutations is not currently supported
```

Looks like this started with the new use of `useEffect` [that calls setShipping mutation](https://github.com/artsy/force/blob/main/src/Apps/Order/Routes/Shipping/index.tsx#L165-L170)

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -492,10 +492,6 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       setSelectedAddressID(value)
       setShippingQuotes(null)
       setShippingQuoteId(undefined)
-
-      if (checkIfArtsyShipping()) {
-        selectShipping()
-      }
     }
   }
 


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -492,10 +492,6 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       setSelectedAddressID(value)
       setShippingQuotes(null)
       setShippingQuoteId(undefined)
-
-      if (checkIfArtsyShipping()) {
-        selectShipping()
-      }
     }
```
jpotts244, line 495: Did some testing locally looking like all data that is persisted on the order is getting saved correctly and sent to ARTA (our shipping logistics provider) I think the [useEffect](https://github.com/artsy/force/blob/main/src/Apps/Order/Routes/Shipping/index.tsx#L165-L170) above is capturing these cases.  

Do you forsee any issues with this? Looks to be working but I am a novice react developer 😸

jpotts244: Feedback most welcome!

Mitch-Henson: You are right in that this seems to have been introduced from the previous [refactor](https://github.com/artsy/force/pull/10747). The [previous code](https://github.com/artsy/force/blob/9b2b173bb1662217366a5ef847662c0cbec493de/src/Apps/Order/Routes/Shipping/index.tsx#L166-L174) only ran once on mount, which meant that this section you have changed was needed to not double run. 

From what I can see, there is two ways to solve this. The way you have right here, or you can change the useEffect to only run once on mount e.g. use `[]` and then leave this code as is. I'm not aware of all the potential cases, but your solution is definitely feasible from what I can see. That's just my 2c though and happy to hear from the actual experts!

jpotts244: I think the useEffect does have the `[]` already, with the selected address id within: https://github.com/artsy/force/pull/11076/files#diff-e0afcd1142a7fdd8a803a66f3f2490eb8db1d25fe9d66bd889fefd4852937ec6L165-L170

Trying to play around with it locally and I am breaking more things. Will keep digging for a bit

kiryl-zubarau: I think @Mitch-Henson means that you can remove dependencies of the `useEffect` hook and leave just an empty array which means that the hook will run only on the component mount.

jpotts244: Hm yes, running locally with the empty array I am running into another bug where it is submitting the first saved address and not the saved address that I've selected.

kiryl-zubarau: Have you moved back this part in `selectSavedAddress` method?

 ```
if (checkIfArtsyShipping()) {
  selectShipping()
}
```

kiryl-zubarau (approved):

Mitch-Henson (approved):
🚢

PR #11082, opened 2022-09-30
feat(artwork): Update video carousel to support order

damassi:
#### Depends on: 
- https://github.com/artsy/metaphysics/pull/4419 (needs to be merged first!) 

This PR solves [GRO-1328]

### Description

This updates the artwork page with support for videos being pushed to the front or the back: 

artworkpage-vid

cc @artsy/grow-devs 

[GRO-1328]: https://artsyproduct.atlassian.net/browse/GRO-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowserLarge.tsx
```diff
@@ -93,7 +93,7 @@ const ArtworkImageBrowserLarge: React.FC<ArtworkImageBrowserLargeProps> = ({
             my={2}
             maxHeight={maxHeight}
             artwork={artwork}
-            activeIndex={index}
+            activeIndex={artwork.setVideoAsCover ? index - 1 : index}
```
laurabeth, line 96: non-blocking[nitpick]: I would alias artwork as figures for the sake of consistency.

damassi: Not sure I follow, can you elaborate? The only thing that has changed in this file is add another field to our relay query, which I read in here

laurabeth: Oh! Nevermind... ignore this. By the time we get here videos are already filtered out of the array and sent elsewhere, so these are just images. 🧠✨

damassi: Yep yep, all the previous work in MP worked out well here. (Even changes there were super minimal.)

PR #11083, opened 2022-09-30
fix: spinner when user comes back to the payment step

AlicanAkyuz:

### Description

This PR fixes the bug reported in TX channel recently. More context in the [slack thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1664477051141039). 

The problem is caused by the fact that the flag to render a spinner in the payment context remains true when a user leaves the payment route with Credit Card as the payment method. This PR makes sure that the flag is set to false when leaving the route.


kiryl-zubarau (approved):
LGTM!

PR #11091, opened 2022-09-30
chore(tooling): add no-relative-path-imports

laurabeth:
Addresses https://artsyproduct.atlassian.net/browse/WP-116

### Description

Add eslint plugin & rule to disallow the use of relative paths, with an exception for same folder.


damassi:
👌 👌

PR #11094, opened 2022-10-01
chore(relay): Upgrade to Relay 13

damassi:

All the hard work credit in this pr goes to @araujobarret -- opening it from over here to get a read on CI and see what it has to say

#### Review App:
- https://github.com/artsy/force/pull/11096


src/Apps/Artwork/Components/ArtworkSidebar2/__tests__/ArtworkSidebar2Links.jest.tsx
```diff
@@ -26,6 +26,7 @@ describe("ArtworkSidebar2Links", () => {
     ArtworkSidebar2Links_Test_Query
   >({
     Component: ({ artwork }) => (
+      // @ts-ignore RELAY_UPGRADE 13
```
damassi, line 29: @araujobarret - these type errors are really mysterious. Any thought as to what the problem is? Looking at the error it's like `fragment ref Foo$data cannot be assigned to Fragment ref Foo$data`, but like... they're identical. I can't figure out and I squinted so much at it.  All of the ones below are similar (as I assume you also found)

damassi:
Merging this and will monitor

PR #11105, opened 2022-10-03
chore(relay): Add official relay devtools extension to VSCode config

damassi:

### Description

This adds the official [VSCode Relay](https://marketplace.visualstudio.com/items?itemName=meta.relay) extension to our VSCode `extensions.json` file, prompting for an install for devs who open force. 

It's pretty sweet and adds intellisense / click-to-definition into graphql code throughout (pr also deletes our old, hacky apollo solution for the same):

intellisense


araujobarret (commented):
Finally, Apollo is gone 🙌 

shouldn't we add the relay plugin to our official pack of plugins too?
https://github.com/artsy/vscode-artsy/blob/master/package.json

damassi:
Unfortunately I'm not sure how to access that! Looks like our login doesn't work, but we should follow up and try it one more time to be sure

PR #11118, opened 2022-10-05
chore(PX-5227): Update artworks where unlisted to have non-index metatags

jpotts244:
The type of this PR is: Chore

This PR solves [PX-5227](https://artsyproduct.atlassian.net/browse/PX-5227)

### Description
Updating artworks that will be marked as unlisted to have non-index meta tags
In order to block unlisted artwork pages from being indexed, we could add non-index metatags, which should prevent crawlers from including them.

Running locally can confirm seeing the noindex metatag
Screen Shot 2022-10-05 at 10 00 51 AM


src/Apps/Artwork/Components/ArtworkMeta.tsx
```diff
@@ -68,11 +68,24 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
...
+    const { artwork } = this.props
+
+    if (artwork?.visibilityLevel == "UNLISTED") {
+      return (
+        <>
+          <Meta name="robots" content="noindex, follow" />
+        </>
+      )
+    }
+  }
```
jpotts244, line 88: Following the pattern we were using for Artist meta
https://github.com/artsy/force/blob/main/src/Apps/Artist/Components/ArtistMeta.tsx#L187-L197

nspinazz89 (approved):
Not sure that a spec is possible for this sort of thing, so LGTM, just one small question out of curiosity!

Mitch-Henson (approved):

src/Apps/Artwork/Components/ArtworkMeta.tsx
```diff
@@ -68,11 +68,24 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
...
     )
   }
+
+  maybeRenderNoIndex() {
+    const { artwork } = this.props
+
+    if (artwork?.visibilityLevel == "UNLISTED") {
+      return (
+        <>
+          <Meta name="robots" content="noindex, follow" />
```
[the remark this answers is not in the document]
jpotts244: Well I found the docs for "nofollow" on google:

> Do not follow the links on this page. If you don't specify this directive, Google may use the links on the page to discover those linked pages. Learn more about [nofollow](https://developers.google.com/search/docs/crawling-indexing/qualify-outbound-links).

Maybe adding the `nofollow` tag is something we actually want here? what do you think @nspinazz89 @ansor4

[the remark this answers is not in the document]
ansor4: Feels like follow is fine here, like we don't really care if Google indexes the pages that are linked in this one: artist, partner, etc. I guess a `nofollow` could be safer in terms of eliminating possible edge cases, I just don't see those cases yet.

ansor4 (approved):
Any way we can add a test for this?

src/Apps/Artwork/Components/__tests__/ArtworkMeta.jest.tsx
```diff
@@ -0,0 +1,95 @@
...
+      mount(
+        <MockBoot>
+          <ArtworkMeta artwork={artwork} />
+        </MockBoot>
+      )
+
+      const tags = getTags()
+
+      expect(tags.meta.find(tag => tag.name === "robots")).toEqual(undefined)
+    })
```
jpotts244, line 93: would love to clean up some of the duplication here and get a better listed/published case test but for now, just checking the presence of the Robot facing meta tag

src/Apps/Artwork/Components/ArtworkMeta.tsx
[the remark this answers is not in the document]
nspinazz89: yeah I agree with @ansor4 that we are fine with follow from this page. That is a good piece of information that I'm sure will be useful at some point though so good job @jpotts244!

src/Apps/Artwork/Components/__tests__/ArtworkMeta.jest.tsx
nspinazz89: I'm not a Jest expert but this seems solid enough to me for a first pass.

nspinazz89 (approved):
This seems good enough to me @jpotts244! Great job adding that spec!

src/Apps/Artwork/Components/ArtworkMeta.tsx
```diff
@@ -68,11 +68,24 @@ export class ArtworkMeta extends Component<ArtworkMetaProps> {
         <SeoDataForArtwork artwork={artwork} />
 
         {this.renderImageMetaTags()}
+        {this.maybeRenderNoIndex()}
```
ansor4, line 71: non-blocking: The `maybe` prefix seems a bit odd and an antipattern. Maybe we can simplify this altogether with something like this?
```
render() {
    const addNoIndex = artwork?.visibilityLevel == "UNLISTED"
    ...
    return (
        ...
        {addNoIndex && <Meta name="robots" content="noindex, follow" />}
    )
}
```

ansor4 (approved):
minor non blocking comment, i'll let you decide if you want to make the change!

src/Apps/Artwork/Components/ArtworkMeta.tsx
jpotts244: Sounds great, will add!

ansor4 (approved):
thanks for the changes! looks good to merge from me!

Mitch-Henson (approved):

PR #11126, opened 2022-10-05
[Review App] Scale images by area in artwork shelves

dzucconi:
This converts the artwork shelf to target an area (default 40,000) when coming up with a width to resize to. This allows for a more visually balanced shelf component. Previously we were just constraining everything by a specific width (default 200px), which would lead to large portrait oriented images and small landscape ones (the artwork brick columns still suffer from this problem).

----

Additionally I've gone around and replaced all of the bespoke skeletons with a single placeholder type which should prevent layout shifting for these components.

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/SoldRecentlyOnArtsy.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/SoldRecentlyOnArtsy.tsx
@@ -1,17 +1,11 @@
 import { ContextModule, OwnerType } from "@artsy/cohesion"
-import {
-  Box,
-  Flex,
-  Skeleton,
-  SkeletonBox,
-  SkeletonText,
-  Spacer,
-  Text,
-} from "@artsy/palette"
+import { Flex, Skeleton, SkeletonText, Spacer, Text } from "@artsy/palette"
 import { shuffle } from "lodash"
-import { Fragment } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
-import { ShelfArtworkFragmentContainer } from "Components/Artwork/ShelfArtwork"
+import {
+  ShelfArtworkFragmentContainer,
+  ShelfArtworkPlaceholder,
+} from "Components/Artwork/ShelfArtwork"
 import { Rail } from "Components/Rail"
 import { useTracking } from "react-tracking"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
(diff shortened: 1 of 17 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/SoldRecentlyOnArtsy.tsx
```diff
@@ -58,29 +52,32 @@ export const SoldRecentlyOnArtsy: React.FC<SoldRecentlyOnArtsyProps> = ({
...
-                  onClick={trackArtworkItemClick(artwork, i)}
-                  // FIXME:
-                  contextModule={ContextModule.artworkRecentlySoldGrid as any}
-                />
+              <ShelfArtworkFragmentContainer
+                key={artwork.internalID}
+                artwork={artwork}
+                hideSaleInfo
+                lazyLoad
+                area={60000}
```
damassi, line 60: Hunch is that this is gonna be kinda hard for devs to remember / understand, at least initially. Can you break it down a bit for me just so I understand better? (once i understand completely we might want to come up with a better prop name.. maybe)

dzucconi: I wouldn't have exposed it otherwise.

damassi: Still, might be good to make the name more clear... `maxHorizontalArea` or something. "Area" is so vague! (No strong feelings about it tho other than "whoa, what does area do", which I don't know)

dzucconi: I agree it's weird but it's not a maxHorizontalArea. It's 'how much area do u want this image to occupy?'

damassi: Oh I see. Feels weird no matter how you describe it. I've got no naming ideas other than to suggest there's a good comment on the interface describing its function; `area={60000}` is foreign and indiscernible in terms of function, and whats that super large number? Anywaysss

dzucconi: Yeah like 'totalNumberOfPixelsInImage' isn't a ton better.

damassi: right

PR #11127, opened 2022-10-05
feat(art-quiz): create results page structure

laurabeth:
The type of this PR is: **feature**

This PR solves [GRO-1258]

### Description
This PR creates a basic structure for the user's art quiz results.

Screen Shot 2022-10-21 at 12 04 33 PM

🚧Still to-do🚧
 - [ ] Remove context and lift logic into ArtQuizApp
 - [ ] Finalize ☝🏽 logic
 - [ ] Add animations
 - [ ] Add tooltips
 - [ ] Add mutation to dislike button
 - [ ] Add mutation to back button
 - [ ] Replace fixture data w/ query
 - [ ] Write integration tests
 - [ ] Finalize folder structure
 - [ ] Add tab content

[GRO-1258]: https://artsyproduct.atlassian.net/browse/GRO-1258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/ArtQuiz/ArtQuizResults.tsx
+++ b/src/Apps/ArtQuiz/ArtQuizResults.tsx
@@ -1,18 +1,14 @@
 import { Box, Button, Spacer, Tab, Tabs, Text } from "@artsy/palette"
 import { FC } from "react"
 import { useTranslation } from "react-i18next"
-import { getENV } from "Utils/getENV"
 
 export const ArtQuizResults: FC = () => {
   const { t } = useTranslation()
-  const name: string | undefined = getENV("CURRENT_USER")?.name
-  const customGreeting: string = `${name}'s ${t("artQuizPage.results.title")}`
-  const defaultGreeting: string = `Your ${t("artQuizPage.results.title")}`
 
   return (
     <Box>
       <Text mt={6} variant="xl">
-        {name ? customGreeting : defaultGreeting}
+        {t("artQuizPage.results.title")}
       </Text>
       <Text color="black60">{t("artQuizPage.results.subtitle")}</Text>
       <Spacer mt={2} />
@@ -20,7 +16,7 @@ export const ArtQuizResults: FC = () => {
         {t("artQuizPage.results.emailButton")}
       </Button>
       <Spacer mt={6} />
-      <Tabs>
+      <Tabs justifyContent="fill">
         <Tab name={t("artQuizPage.results.tabs.worksYouLiked")} />
         <Tab name={t("artQuizPage.results.tabs.recommendedCollections")} />
         <Tab name={t("artQuizPage.results.tabs.recommendedArtists")} />
--- a/src/Apps/ArtQuiz/ArtQuizResults.tsx
+++ b/src/Apps/ArtQuiz/ArtQuizResults.tsx
@@ -16,7 +16,7 @@ export const ArtQuizResults: FC = () => {
         {t("artQuizPage.results.emailButton")}
       </Button>
       <Spacer mt={6} />
-      <Tabs justifyContent="fill">
+      <Tabs fill>
         <Tab name={t("artQuizPage.results.tabs.worksYouLiked")} />
         <Tab name={t("artQuizPage.results.tabs.recommendedCollections")} />
         <Tab name={t("artQuizPage.results.tabs.recommendedArtists")} />
--- a/src/Apps/ArtQuiz/ArtQuizResults.tsx
+++ b/src/Apps/ArtQuiz/ArtQuizResults.tsx
@@ -1,0 +1,30 @@
+import { Box, Button, Spacer, Tab, Tabs, Text } from "@artsy/palette"
+import { FC } from "react"
+import { useTranslation } from "react-i18next"
+import { getENV } from "Utils/getENV"
+
+export const ArtQuizResults: FC = () => {
+  const { t } = useTranslation()
+  const name: string | undefined = getENV("CURRENT_USER")?.name
+  const customGreeting: string = `${name}'s ${t("artQuizPage.results.title")}`
+  const defaultGreeting: string = `Your ${t("artQuizPage.results.title")}`
+
+  return (
+    <Box>
+      <Text mt={6} variant="xl">
+        {name ? customGreeting : defaultGreeting}
+      </Text>
+      <Text color="black60">{t("artQuizPage.results.subtitle")}</Text>
+      <Spacer mt={2} />
+      <Button variant="secondaryBlack">
+        {t("artQuizPage.results.emailButton")}
+      </Button>
+      <Spacer mt={6} />
+      <Tabs>
+        <Tab name={t("artQuizPage.results.tabs.worksYouLiked")} />
+        <Tab name={t("artQuizPage.results.tabs.recommendedCollections")} />
+        <Tab name={t("artQuizPage.results.tabs.recommendedArtists")} />
+      </Tabs>
+    </Box>
+  )
+}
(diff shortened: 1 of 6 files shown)


src/Apps/ArtQuiz/ArtQuizResults.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <Text mt={6} variant="xl">
+        {t("artQuizPage.results.title")}
+      </Text>
+      <Text color="black60">{t("artQuizPage.results.subtitle")}</Text>
+      <Spacer mt={2} />
+      <Button variant="secondaryBlack">
+        {t("artQuizPage.results.emailButton")}
+      </Button>
+      <Spacer mt={6} />
+      <Tabs justifyContent="fill">
```
dzucconi, line 19: `<Tabs fill>`

laurabeth: fixed 👍🏽

src/Apps/ArtQuiz/ArtQuizWelcome.tsx
```diff
@@ -34,28 +36,27 @@ export const ArtQuizWelcome: FC<ArtQuizWelcomeProps> = ({ onStartQuiz }) => {
...
+                <Text variant={["md", "lg-display"]}>
                   {t("artQuizPage.welcomeScreen.subtitle2")}
                 </Text>
               </Box>
 
-              <Spacer my={1} />
-              <Box width="100%">
+              <Spacer my={6} />
+
+              <Box width="100%" my={6}>
```
dzucconi, line 59: Does this margin actually do anything with the `Spacer` above that it's collapsing into?

dzucconi: (If they *are* stacking, that'd be an "illegal" spacing value)

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
```diff
@@ -0,0 +1,53 @@
...
+      tabs: {
+        worksYouLiked: "Works You Liked",
+        recommendedCollections: "Recommended Collections",
+        recommendedArtists: "Recommended Artists",
+      },
+    },
+  },
+}
+
+jest.mock("react-i18next", () => {
```
dzucconi, line 20: Ugh, this is gonna be annoying.

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
```diff
@@ -0,0 +1,53 @@
+import { ArtQuizResults } from "Apps/ArtQuiz/ArtQuizResults"
+import { render, screen } from "@testing-library/react"
+
+const translationFileMock = {
```
dzucconi, line 4: We *really* shouldn't have to do this — there's no way to load this automatically? This is wild.

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
```diff
@@ -0,0 +1,53 @@
...
+    ["title", "Explore Your Quiz Results"],
+    [
+      "subtitle",
+      "Explore these collections and artists recommended for you based on your saved works. Follow them to see their latest works on your Artsy home.",
+    ],
+    ["email button", "Email My Results"],
+    ["tab", "Works You Liked"],
+    ["tab", "Recommended Collections"],
+    ["tab", "Recommended Artists"],
+  ])("displays the expected %s text", (_key, text) => {
```
dzucconi, line 49: This seems... excessive for a simple render spec.

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
gkartalis: oooooh wait, no need to do that anymore! Maybe my fault for not communicating the change better but the setupTestWrapper files use these two, one for [enzyme](https://github.com/artsy/force/blob/main/src/DevTools/setupTestWrapper.tsx#L16) one for [rtl](https://github.com/artsy/force/blob/main/src/DevTools/setupTestWrapper.tsx#L20)

src/Apps/ArtQuiz/ArtQuizWelcome.tsx
laurabeth: They are stacking, yes.

dzucconi: Oh wait 12 is a valid value. So this is fine — but it'd be easier to understand if it was just a single spacer rather than two stacked.

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
laurabeth: Why?

dzucconi: It's just like: Did the thing render correctly, y or n? A single spec will tell you that and is easier to read and maintain vs a dynamically constructed set of specs for every significant piece of text on the page.

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
laurabeth: It was my intention to complete this PR and then lift out into some sort of a global mock for i18n. 

Do we really want to wrap every test for a translated component in the full test wrapper? It seems like we should be able to test simple components in relative isolation.

dzucconi: You could just use the re-exported `render` as @gkartalis points out.

laurabeth: Agreed. Just looked at the file and saw that those are separate exports. 

@gkartalis did you try to create a global default? Did that implementation fall short somehow?

gkartalis: I went with the simplest solution at that point that I found, which was to re-export the testing functions wrapped, but happy to go with any other solution if you have something better in mind 🙂

src/Apps/ArtQuiz/__tests__/ArtQuizResults.jest.tsx
dzucconi: `blocking` The `NavBar` used to have a similar spec. It would break constantly and is a pain to maintain. It makes it difficult to search for the relevant spec when you see it in your test output. There's no benefit here, just a simple: did a single string render is fine. And in fact nothing of consequence is even being tested because the strings it's looking for are mocked.

src/Apps/ArtQuiz/__tests__/ArtQuizResultsLoader.jest.tsx
```diff
@@ -1,22 +1,21 @@
 import React from "react"
-import { render, screen } from "@testing-library/react"
+import { screen } from "@testing-library/react"
 import {
   ArtQuizResultsLoader,
   waitTime,
 } from "Apps/ArtQuiz/Components/ArtQuizResultsLoader"
 import { act } from "react-dom/test-utils"
+import { render } from "DevTools/setupTestWrapper"
```
dzucconi, line 8: I think you're right that we need to globally mock this. This is painful to have to know and is hidden. This is fine for now, but I'll make a ticket to actually fix this.

dzucconi: Resolved here: https://github.com/artsy/force/pull/11238 — will have to rebase.

dzucconi (approved):
👍 👍

PR #11135, opened 2022-10-07
feat: Initialize SWA form images when submitting an artwork from My Collection

olerichter00:

This PR solves [CX-2952](https://artsyproduct.atlassian.net/browse/CX-2952)

### Description

This PR implements the initialization of images in the Sell With Artsy form when submitting an artwork from My Collection.
All images from the My Collection artwork will be automatically reuploaded and processed and added as assets to the submission.

Screenshot 2022-10-10 at 11 03 52

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
@@ -65,7 +65,7 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
 
   useEffect(() => {
     const imagesToUpload = values.photos.filter(
-      c => !(c.geminiToken || c.url) && !c.loading
+      c => !(c.geminiToken || c.url) && !c.loading && !c.errorMessage
     )
 
     if (imagesToUpload.length) {
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
@@ -12,7 +12,7 @@ import { BackLink } from "Components/Links/BackLink"
 import { PhotoThumbnail } from "Components/PhotoUpload/Components/PhotoThumbnail"
 import { normalizePhoto, Photo } from "Components/PhotoUpload/Utils/fileUtils"
 import { Form, Formik } from "formik"
-import { useRef, useState } from "react"
+import { useEffect, useRef, useState } from "react"
 import {
   createFragmentContainer,
   Environment,
@@ -161,6 +161,16 @@ export const UploadPhotos: React.FC<UploadPhotosProps> = ({
         initialErrors={initialErrors}
       >
         {({ values, setFieldValue, isValid, isSubmitting }) => {
+          useEffect(() => {
+            console.log({
+              photos: values.photos,
+            })
+            console.log(
+              "photos",
+              values.photos.map(p => p.removed)
+            )
+          }, [values.photos])
+
           const handlePhotoDelete = (photo: Photo) => {
             photo.removed = true
             photo.abortUploading?.()
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
@@ -1,8 +1,16 @@
 import { Box, Button, Text } from "@artsy/palette"
 import { SubmissionStepper } from "Apps/Consign/Components/SubmissionStepper"
+import {
+  useAddAssetToConsignmentSubmission,
+  useRemoveAssetFromConsignmentSubmission,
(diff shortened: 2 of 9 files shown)


src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
```diff
@@ -65,7 +65,7 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
 
   useEffect(() => {
     const imagesToUpload = values.photos.filter(
-      c => !(c.geminiToken || c.url) && !c.loading
+      c => !(c.geminiToken || c.url) && !c.loading && !c.errorMessage
```
olerichter00, line 68: Without filtering out artworks with an error message, this effect will try to re-add images that could not be loaded, which can lead to an infinite loop.

Alternatively, we could add a retry counter and only retry to upload an image 3 times. But I've decided against this solution because for me it seems to be too complex. The user can also just reupload the image manually.

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
```diff
@@ -1,13 +1,20 @@
-import { AssetCredentials } from "../Mutations/getGeminiCredentialsForEnvironment"
+import { AssetCredentials } from "Components/PhotoUpload/Mutations/getGeminiCredentialsForEnvironment"
 import { Photo } from "./fileUtils"
 
+const externalPhotoToFile = async (photo: Photo) => {
+  const response = await fetch(photo.externalUrl!)
+  const blob = await response.blob()
+
+  return new File([blob], photo.name)
+}
```
olerichter00, line 9: This function fetches the artwork image and returns a file that can then be used to upload the image to S3.

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
```diff
@@ -68,8 +71,19 @@ export const getUploadPhotosFormInitialValues = (
...
           loading: false,
-        })) || [],
+        })) || []
+  } else if (myCollectionArtwork) {
+    photos =
+      myCollectionArtwork?.images
+        ?.map(image => ({
+          name: "Automatically added",
+          externalUrl: image?.url!,
+          type: "image/jpg",
```
MounirDhahri, line 81: **thought:** do we only allow jpg images?

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
MounirDhahri: **nitpick:** thanks for the comment, maybe we can also add it to the code with the method name 👀

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
```diff
@@ -23,16 +30,27 @@ export const uploadFileToS3 = (
...
+    } catch (error) {
+      reject(new Error("Initializing artwork photo failed."))
+      return
+    }
+
     const data = {
-      acl: acl,
-      "Content-Type": photo.file.type,
+      acl,
+      "Content-Type": photo.file.type || "image/jpg",
```
MounirDhahri, line 46: **thought:** I am wondering if it would be possible to always have the file type here to avoid the `|| image/jpg` here

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
```diff
@@ -23,16 +30,27 @@ export const uploadFileToS3 = (
...
+    const isExternalPhoto = photo.externalUrl
+
+    let file
+
+    try {
+      file = isExternalPhoto ? await externalPhotoToFile(photo) : photo.file
+    } catch (error) {
+      reject(new Error("Initializing artwork photo failed."))
+      return
+    }
```
MounirDhahri, line 42: **Q:** I am wondering if you considered moving this logic to where the upload file method is being called to always send files here in order to avoid initializing the file here

MounirDhahri (commented):
Thanks for the nice PR @olerichter00 . I left some small comments to ask for some details, please let me know if something wasn't clear

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
olerichter00: Good idea! I've added a comment and renamed the function to `fetchExternalFile`.

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
olerichter00: The image from the artwork comes as a jpg. If the user uploads a file manually, we support more file types and take the type directly from the uploaded file.

src/Components/PhotoUpload/Utils/uploadFileToS3.tsx
olerichter00: You're right, the file type should always be set. I've removed the `|| "image/jpg"`.

MounirDhahri (approved):
Nice, thanks for addressing the review comments quickly. LGTM 🙏

PR #11136, opened 2022-10-07
chore: remove leftovers from relay upgrade

araujobarret:

### Description

Removes Relay ignore leftovers

### DIFF
--- a/src/Apps/Order/orderRoutes.tsx
+++ b/src/Apps/Order/orderRoutes.tsx
@@ -127,7 +127,6 @@ export const orderRoutes: AppRouteConfig[] = [
             { order }
           )
 
-          // FIXME RELAY_UPGRADE
           if (redirect === null) {
             // Work around to ensure that all of the "stays on the..." tests
             // remain valid for Relay 13 upgrade, since with the new approach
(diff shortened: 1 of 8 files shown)


dzucconi (approved):

src/Apps/Partner/Components/PartnerArtists/PartnerArtistList/PartnerArtists.tsx
```diff
@@ -41,7 +41,6 @@ export const PartnerArtists: React.FC<PartnerArtistsProps> = ({
       <PartnerArtistListFragmentContainer
         partnerSlug={slug}
         scrollTo={scrollTo!}
-        // FIXME: RELAY UPGRADE
         artists={artists as any}
```
dzucconi, line 44: Should remove the `any`

src/Apps/Order/orderRoutes.tsx
```diff
@@ -127,7 +127,6 @@ export const orderRoutes: AppRouteConfig[] = [
             { order }
           )
 
-          // FIXME RELAY_UPGRADE
```
dzucconi, line 130: I guess this still looks relevant: there's some code here that would need to be walked back.

damassi: Wasn't able to solve this one, but this maintains that the logic we were testing as valid within the test, albeit annoyingly

src/Apps/Partner/Components/PartnerArtists/PartnerArtistList/PartnerArtists.tsx
dzucconi: It should be fine if you use the extractNodes util on the parent connection

PR #11137, opened 2022-10-07
fix: Remove callback to avoid removed photos to appear again in the SWA form

olerichter00:
## Description

Removes `callback` to avoid removed photos to appear again in the SWA form when adding a new photo.

### Steps to reproduce the bug / QA steps

1. Submit an artwork
2. Add artwork details
3. Add some photos
4. Remove some photos
5. Add a new photo
6. **Removed photos shouldn't appear again**

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
@@ -9,7 +9,7 @@ import {
 } from "Components/PhotoUpload/Utils/fileUtils"
 import { useFormikContext } from "formik"
 import * as React from "react"
-import { useCallback, useEffect, useState } from "react"
+import { useEffect, useState } from "react"
 import { FileRejection } from "react-dropzone"
 import { useSystemContext } from "System"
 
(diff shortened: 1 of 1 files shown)


src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
```diff
@@ -74,15 +74,15 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
     }
   }, [values.photos])
 
-  const onDrop = useCallback((acceptedFiles: File[]) => {
+  const onDrop = (acceptedFiles: File[]) => {
```
olerichter00, line 77: I think using callback here leads to `values.photos` still pointing to an older object (after the last photo has been added). Because of this all new files are added but also all old files that have been removed are added again.

MounirDhahri: good thinking of removing this. i think no need for optimising here

MounirDhahri (approved):

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
olerichter00: Exactly. Even adjusting the `useCallback` could use to similar bugs in the future.

PR #11141, opened 2022-10-10
fix(FX-4319): commercial buttons section overflow

dimatretyak:

This PR solves [FX-4319]

Review app: [commercial-buttons](http://commercial-buttons.artsy.net/)
Quick artwork example: [Link](https://commercial-buttons.artsy.net/artwork/damon-zucconi-editioned-cg-price-range-arta-us-2)

### Problems
* Commercial buttons starts overflowing the sidebar when screen width < 1075
* iPhone SE (first generation) does the same when contact gallery is secondary

### Solution
We discussed with Kseniya and decided that the buttons will be placed under each other by default (column view). When the screen width is `1192px` or more (`xl` breakpoint according to the list of [supported breakpoints](https://palette.artsy.net/guides/responsive/#all-breakpoints)), then place the buttons in a row

#### Problem №1 - when screen width < 1075
| Before | After |
| before-1 | after-1 | 

#### Problem №2 - when the device has a small screen (e.g. iPhone SE first generation)
| Before | After |
| before-2 | after-2 | 

[FX-4319]: https://artsyproduct.atlassian.net/browse/FX-4319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
@@ -6,6 +6,7 @@ import {
 import {
   Button,
   Flex,
+  Join,
   Separator,
   Spacer,
   Text,
@@ -27,6 +28,7 @@ import { ContextModule, Intent } from "@artsy/cohesion"
 import { openAuthModal } from "Utils/openAuthModal"
 import currency from "currency.js"
 import { useTranslation } from "react-i18next"
+import { Media } from "Utils/Responsive"
 
 interface SaleMessageProps {
   saleMessage: string | null
@@ -288,10 +290,7 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
     }
 
     return (
-      <>
-        <Spacer mt={2} />
-        <ArtworkSidebarCreateAlertButtonFragmentContainer artwork={artwork} />
-      </>
+      <ArtworkSidebarCreateAlertButtonFragmentContainer artwork={artwork} />
     )
   }
 
@@ -326,21 +325,32 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
         </>
       )}
 
-      <Flex>
-        {artwork.isSold && renderCreateAlertButton()}
-        {artwork.isAcquireable && (
-          <Button
-            width="100%"
-            size="large"
-            loading={isCommitingCreateOrderMutation}
-            onClick={handleCreateOrder}
-          >
-            {t("artworkPage.sidebar.commercialButtons.buyNow")}
-          </Button>
-        )}
-        {artwork.isOfferable && (
-          <>
-            <Spacer ml={artwork.isAcquireable ? 1 : 0} />
+      <Flex flexDirection={["column", "column", "column", "column", "row"]}>
+        <Join
+          separator={
+            <>
+              <Media at="xl">
+                <Spacer ml={1} />
+              </Media>
+
+              <Media lessThan="xl">
+                <Spacer mt={1} />
+              </Media>
+            </>
+          }
+        >
+          {artwork.isSold && renderCreateAlertButton()}
+          {artwork.isAcquireable && (
+            <Button
+              width="100%"
+              size="large"
+              loading={isCommitingCreateOrderMutation}
+              onClick={handleCreateOrder}
+            >
+              {t("artworkPage.sidebar.commercialButtons.buyNow")}
+            </Button>
+          )}
+          {artwork.isOfferable && (
             <Button
               variant={
                 artwork.isAcquireable ? "secondaryBlack" : "primaryBlack"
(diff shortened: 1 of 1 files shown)


src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
```diff
@@ -287,10 +289,50 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
       return null
     }
 
+    return (
+      <ArtworkSidebarCreateAlertButtonFragmentContainer artwork={artwork} />
+    )
+  }
+
+  const renderButtons = () => {
```
dzucconi, line 297: We should really avoid this pattern of having a 'render' function inside of another render function. Ideally should be extracted to its own component. If you really don't want to do that then wrap in `useMemo`. (It's caused subtle bugs in production in the past.)

dzucconi:
(OH, I see it's a feature flag. What about just simply changing "Contact Gallery" to "Contact" at the smaller size with a simple `<span display={["none", "inline"]}>Gallery</span>`

dzucconi:
Also couldn't you just achieve this proposed fix with a responsive prop for `flexDirection`? (I'm just convinced we over rely `Media`)

dimatretyak:
> Also couldn't you just achieve this proposed fix with a responsive prop for `flexDirection`? (I'm just convinced we over rely `Media`)

we can do this, but in this case it will look something [like this in the code](https://github.com/artsy/force/blob/5fcfb8285177c3ad23e43ce88428df63a1331d34/src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx#L328) (it seemed to me that it was better to use `Media` and `renderButtons` pattern)

dzucconi:
Responsive props are going to be much more light weight and are preferable. I only really reach for `Media` if it's like two totally different designs.

dzucconi:
The use of `Media` for simply a copy change is kind of interesting, I guess. Though I don't see what benefit that has over the `span`

dimatretyak:
> The use of `Media` for simply a copy change is kind of interesting, I guess. Though I don't see what benefit that has over the `span`

I mean, here we use `react-i18next` and it won't work so easy to use `span` with `display={["none", "inline"]}`

dimatretyak:
> What about just simply changing "Contact Gallery" to "Contact" at the smaller size

We discussed this moment with Kseniya and decided to leave the current implementation

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
```diff
@@ -326,21 +325,32 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
...
+          separator={
+            <>
+              <Media at="xl">
+                <Spacer ml={1} />
+              </Media>
+
+              <Media lessThan="xl">
+                <Spacer mt={1} />
+              </Media>
+            </>
```
dzucconi, line 339: These Spacers aren't any different? (And if they were: responsive props)

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
```diff
@@ -326,21 +325,32 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
...
+        <Join
+          separator={
+            <>
+              <Media at="xl">
+                <Spacer ml={1} />
+              </Media>
+
+              <Media lessThan="xl">
+                <Spacer mt={1} />
+              </Media>
```
dzucconi, line 338: `<Spacer ml={[0, 0, 0, 1]} mt={[1, 0, 0, 0]} />` or something.

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
```diff
@@ -326,21 +325,32 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
...
-            loading={isCommitingCreateOrderMutation}
-            onClick={handleCreateOrder}
-          >
-            {t("artworkPage.sidebar.commercialButtons.buyNow")}
-          </Button>
-        )}
-        {artwork.isOfferable && (
-          <>
-            <Spacer ml={artwork.isAcquireable ? 1 : 0} />
+      <Flex flexDirection={["column", "column", "column", "column", "row"]}>
```
dzucconi, line 328: This is probably correct but why wouldn't `["column", "row"]` work here?

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
```diff
@@ -326,21 +324,20 @@ const ArtworkSidebar2CommerialButtons: React.FC<ArtworkSidebar2CommercialButtons
...
+        <Join separator={<Spacer ml={[0, 0, 0, 0, 1]} mt={[1, 1, 1, 1, 0]} />}>
```
dzucconi, line 328: And actually just simply `<Spacer mt={1} ml={1} />` would work too! 😛

dimatretyak: thank you!

src/Apps/Artwork/Components/ArtworkSidebar2/ArtworkSidebar2CommercialButtons.tsx
dzucconi: 👍 👍

dzucconi (approved):

PR #11147, opened 2022-10-11
feat(FX-4363): Activity panel: Remove artworks counter

nickskalkin:

This PR solves [FX-4363]

### Description

Removes counters from individual notifications since backend returns not complete data which makes it impossible to display a correct number. This is a temporary solution until the issue is resolved on BE side.

Screenshot 2022-10-11 at 11 53 57

[FX-4363]: https://artsyproduct.atlassian.net/browse/FX-4363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Notifications/NotificationItem.tsx
+++ b/src/Components/Notifications/NotificationItem.tsx
@@ -15,7 +15,6 @@ const UNREAD_INDICATOR_SIZE = 8
 
 const NotificationItem: React.FC<NotificationItemProps> = ({ item }) => {
   const artworks = extractNodes(item.artworksConnection)
-  const remainingArtworksCount = (item.artworksConnection?.totalCount ?? 0) - 4
 
   const getNotificationType = () => {
     if (item.notificationType === "ARTWORK_ALERT") {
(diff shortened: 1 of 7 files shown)


src/Components/Notifications/NotificationItem.tsx
```diff
@@ -15,7 +15,6 @@ const UNREAD_INDICATOR_SIZE = 8
 
 const NotificationItem: React.FC<NotificationItemProps> = ({ item }) => {
   const artworks = extractNodes(item.artworksConnection)
-  const remainingArtworksCount = (item.artworksConnection?.totalCount ?? 0) - 4
```
dimatretyak, line 18: **suggestion**: I think you also need to remove [totalCount](https://github.com/artsy/force/blob/4740ec7917ae209321fd275ba892a890e215ecdc/src/Components/Notifications/NotificationItem.tsx#L97) from the query, since this data is not used

dimatretyak (dismissed):

src/Components/Notifications/NotificationItem.tsx
nickskalkin: thank you @dimatretyak, just did it!

PR #11150, opened 2022-10-11
feat(FX-4329): Auctions Registration Countdown in Banner 

dimatretyak:

This PR solves [FX-4329]

Review app: [countdown-for-top-artwork-banner](https://countdown-for-top-artwork-banner.artsy.net/)

Countdown example: [artwork/joan-miro-la-serpentine-1](https://countdown-for-top-artwork-banner.artsy.net/artwork/joan-miro-la-serpentine-1)

### Acceptance Criteria
* Only show on desktop web
* Not conditional on user registration status

### Demo
demo

[FX-4329]: https://artsyproduct.atlassian.net/browse/FX-4329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/TopContextBar.tsx
```diff
@@ -51,32 +53,36 @@ export const TopContextBar: React.FC<TopContextBarProps> = ({
...
+
+          <Text variant="xs" lineHeight={1} lineClamp={2}>
+            {children}
+          </Text>
+        </Flex>
 
-        <Text variant="xs" lineHeight={1} lineClamp={2}>
-          {children}
-        </Text>
+        {rightContent}
```
dzucconi, line 85: Fine with this but just wondering — is the `rightContent` even necessary — could you just nest this layout inside the `children` as is?

dimatretyak: Yes, that's what I [originally did](https://github.com/artsy/force/pull/11150/commits/c34b2a1e72e2cedcfe34b27ee85decd49e75c0dc). but it seems to me that the current implementation will be better 😅

dzucconi: It'd only be better if this was a pattern that we intend on using elsewhere. (Which is maybe the case? A question for the design team)

gkartalis (approved):

PR #11151, opened 2022-10-11
chore(FX-4321): change Contact For Price checks

gkartalis:

This PR solves [FX-4321]

### Description

Changed in gravity the string `Contact For Price` to be `Contact for price`, and as a result, we should also change the checks that we had in place.

this is the related gravity pr.
https://github.com/artsy/gravity/pull/15726

[FX-4321]: https://artsyproduct.atlassian.net/browse/FX-4321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
@@ -91,7 +91,9 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
   renderSaleMessage(saleMessage: string) {
     return (
       <Text variant="lg-display" data-test="SaleMessage">
-        {saleMessage === "Contact For Price" ? "Price on request" : saleMessage}
+        {saleMessage.toLowerCase() === "contact for price"
+          ? "Price on request"
+          : saleMessage}
       </Text>
     )
   }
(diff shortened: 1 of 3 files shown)


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
```diff
@@ -91,7 +91,9 @@ export class ArtworkSidebarCommercialContainer extends React.Component<
   renderSaleMessage(saleMessage: string) {
     return (
       <Text variant="lg-display" data-test="SaleMessage">
-        {saleMessage === "Contact For Price" ? "Price on request" : saleMessage}
+        {saleMessage.toLowerCase() === "contact for price"
+          ? "Price on request"
+          : saleMessage}
```
dimatretyak, line 96: **suggestion**: maybe put this check into helper method? something like this:

```javascript
const checkIsContactForPrice = (message: string) => message.toLowerCase() === "contact for price"
```

dimatretyak (approved):
LGTM 👍

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercial.tsx
gkartalis: We're going to delete these files either way after the new Artwork screen is released so I was thinking of leaving that as it is.
This change is mainly for the transitional phase until we release it

PR #11152, opened 2022-10-11
fix: Adjust error message when initial artwork image loading fails in SWA flow

olerichter00:
Addresses [CX-2952]

## Description

Adjust error message when initial artwork image loading (from My Collection) fails in SWA flow.
Screenshot 2022-10-11 at 15 21 44

[Slack Thread with more context](https://artsy.slack.com/archives/C01B2P6LJUU/p1665492951106379)

[CX-2952]: https://artsyproduct.atlassian.net/browse/CX-2952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
@@ -52,7 +52,9 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
       photo.loading = false
 
       if (!geminiToken) {
-        photo.errorMessage = `Photo could not be added: ${photo.name}`
+        photo.errorMessage = photo.externalUrl
+          ? "Artwork image could not be automatically added. Please add a photo."
+          : `Photo could not be added: ${photo.name}`
         setFieldValue("photos", values.photos)
         return
       }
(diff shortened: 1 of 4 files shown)


src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
```diff
@@ -52,7 +52,9 @@ export const UploadPhotosForm: React.FC<UploadPhotosFormProps> = ({
       photo.loading = false
 
       if (!geminiToken) {
-        photo.errorMessage = `Photo could not be added: ${photo.name}`
+        photo.errorMessage = photo.externalUrl
+          ? "Artwork image could not be automatically added. Please add a photo."
+          : `Photo could not be added: ${photo.name}`
```
MounirDhahri, line 57: **thought:** not sure if the name here is going to be descriptive enough since it's probably going to be something random

MounirDhahri (dismissed):
LGTM, let's fix this the tests and get this merged

dariakoko (dismissed):

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/Components/UploadPhotosForm.tsx
olerichter00: Good point! Let's tackle this separately to see if we can make the copy here more clear as well.

PR #11157, opened 2022-10-12
feat(CX-2890): Show high demand info icon on MyCollectionArtwork

lordkiz:

This PR solves [CX-2890]

### Description
- Displays the `High Demand` info and Icon on a MyCollectionArtwork of a P1 artist with high demand rank

Screenshot 2022-10-11 at 15 53 18

[CX-2890]: https://artsyproduct.atlassian.net/browse/CX-2890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/Details.tsx
```diff
@@ -132,6 +135,17 @@ const SaleInfoLine: React.FC<DetailsProps> = props => {
   )
 }
 
+const HighDemandInfo = () => {
+  return (
+    <Flex flexDirection="row">
+      <HighDemandIcon />
+      <Text variant="sm-display" color="blue100" ml={0.5}>
```
MrSltun, line 142: **Suggestion:** the variant according to the design is `xs` 🌚 
```suggestion
      <Text variant="xs" color="blue100" ml={0.5}>
```

MrSltun (commented):
Looks good! 🌟  Thank you Kizito! 🙌 
I just left some comments from a design perspective :)

dariakoko:
@lordkiz Could you, please, pull and merge main into this branch? It shows 136 changes files currently 😮

src/Components/Artwork/Details.tsx
lordkiz: Thank youuuu! 🙏

lordkiz:
> @lordkiz Could you, please, pull and merge main into this branch? It shows 136 changes files currently 😮

@dariakoko those are mostly __generated__ files. Unlike in Eigen, we still commit those in Force. Or has this changed? 🤔

dariakoko (dismissed):
Ah, alright, all good then

Thank you Kizito!

MrSltun (dismissed):
Thank you Kizito! 🌟

src/Components/Artwork/Details.tsx
```diff
@@ -28,6 +30,7 @@ interface DetailsProps {
   hideArtistName?: boolean
   hidePartnerName?: boolean
   isHovered?: boolean
+  isMyCollectionArtwork?: boolean
```
olerichter00, line 33: **suggestion (non-blocking):** What do you think about keeping this prop independent from the concept of My Collection? We could rename it to `showHighDemandIcon` and open the possibility to use the prop outside of My Collection in the future without confusion.

src/Components/Artwork/__tests__/Details.jest.tsx
```diff
@@ -24,6 +24,7 @@ describe("Details", () => {
   const getWrapper = async (
     response: Details_Test_Query$rawResponse["artwork"],
     restProps?: {
+      isMyCollectionArtwork?: boolean
```
olerichter00, line 27: Same comment as above about renaming the prop.

olerichter00 (approved):
Looks good 🌟 

I've just added a comment about the name of the prop.

src/Components/Artwork/Details.tsx
lordkiz: Hmmmm. This is a great point. I will update the PR. Thank you 👍

olerichter00 (approved):

src/Components/Artwork/Metadata.tsx
```diff
@@ -54,6 +54,7 @@ export const Metadata: React.FC<MetadataProps> = ({
         hidePartnerName={hidePartnerName}
         hideArtistName={hideArtistName}
         isHovered={isHovered}
+        showHighDemandIcon={isMyCollectionArtwork}
```
MrSltun, line 57: **suggestion(non-blocking):** maybe we should add a separate prop for the `showHighDemandIcon` instead of using `isMyCollectionArtwork`?

src/Components/Artwork/Details.tsx
```diff
@@ -321,6 +348,14 @@ export const DetailsFragmentContainer = createFragmentContainer(Details, {
...
       sale_message: saleMessage
       cultural_maker: culturalMaker
+      artist {
+        targetSupply {
+          isP1
+        }
+      }
+      marketPriceInsights {
+        demandRank
+      }
```
MrSltun, line 358: **question:** I'm concerned that this might cause the artwork list across Force will be a bit slower since with every request to MP (and gravity) there will be another one to Vortex to get the market price insights, is there a way to have a conditional way to query only if `showHighDemandInfo` is true? cc @olerichter00

src/Components/Artwork/Metadata.tsx
lordkiz: @MrSltun I think that if we do that we will have to pass the prop all the way form Artwork component down here. But we already have `isMyCollectionArtwork` for which we want to show High Demand icon. That will be like doing the same thing twice, right? 🤔

src/Components/Artwork/Details.tsx
lordkiz: Hmmmm. Good point 👍 . Curious what the others think too.

src/Components/Artwork/Metadata.tsx
MrSltun: I understand your point, but my idea was about having this feature for non-MyCollection artworks, but that's not related to the current ticket :)

lordkiz: Oh I see 👀👍🏾.  However, `Metadata.tsx` already uses `isMyCollectionArtwork` prop to determine the behaviour of the routerLink.

PR #11174, opened 2022-10-13
fix(home): tightens up homepage placeholders

dzucconi:
Minor thing; going around trying to improve some of the placeholders to reduce layout shift.

In a separate PR I'm going to extract some work from https://github.com/artsy/force/pull/11126 which included a bunch of placeholder improvements (— not sure when we'll get to a point where we can start breaking the image caches).

### DIFF
--- a/src/Apps/Auctions/Components/MyBids/MyBids.tsx
+++ b/src/Apps/Auctions/Components/MyBids/MyBids.tsx
@@ -5,7 +5,6 @@ import { MyBidsBidHeaderFragmentContainer } from "./MyBidsBidHeader"
 import { MyBidsBidItemFragmentContainer } from "./MyBidsBidItem"
 import {
   Box,
-  Flex,
   Button,
   Join,
   Separator,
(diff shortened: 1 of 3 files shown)


src/Apps/Auctions/Components/MyBids/MyBids.tsx
```diff
@@ -133,77 +129,6 @@ export const MyBidsFragmentContainer = createFragmentContainer(MyBids, {
   `,
 })
 
-const MyBidsPlaceholder: React.FC = () => {
```
dzucconi, line 136: As discussed here: https://artsy.slack.com/archives/C01ADJNCS5D/p1665433604940789

src/Apps/Auctions/Components/MyBids/MyBids.tsx
```diff
@@ -214,6 +139,7 @@ export const MyBidsQueryRenderer: React.FC = () => {
   return (
     <SystemQueryRenderer<MyBidsQuery>
       lazyLoad
+      debugPlaceholder
```
dzucconi, line 142: WOOPS

PR #11193, opened 2022-10-18
fix(TX-828): Postal code validation in shipping step address form

AlicanAkyuz:

This PR solves [TX-828]

### Description
Currently the address validation util used for the address form in the shipping step of the checkout flow does not capture wrong postal code submissions by the user. As a result of that, we go ahead and make a request to our API to set the (invalid) address on the order, which results in an error as illustrated below: 

Screenshot 2022-10-18 at 14 37 01

This PR adds a validator function for postal code for addresses in the US and CA, so that: 
- user is informed about the wrong information they submitted 
- we avoid trying to set an invalid address on the order

[TX-828]: https://artsyproduct.atlassian.net/browse/TX-828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
+++ b/src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
@@ -12,6 +12,9 @@ describe("formValidators/validatePostalCode", () => {
   it("returns no error when postal code is valid for US", () => {
     expect(validatePostalCode("15601", "US")).toBe(null)
   })
+  it("returns no error when a 9-digit postal code is valid for US", () => {
+    expect(validatePostalCode("88310-7241", "US")).toBe(null)
+  })
   it("returns no error when postal code is valid for CA", () => {
     expect(validatePostalCode("M3J3N3", "CA")).toBe(null)
   })
--- a/src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
+++ b/src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
@@ -7,7 +7,7 @@ import {
 
 describe("formValidators/validatePostalCode", () => {
   it("returns error when postal code is invalid for US", () => {
-    expect(validatePostalCode("XX", "US")).toBe("This field is required")
+    expect(validatePostalCode("XX", "US")).toBe("Postal code is invalid")
   })
   it("returns no error when postal code is valid for US", () => {
     expect(validatePostalCode("15601", "US")).toBe(null)
@@ -19,7 +19,7 @@ describe("formValidators/validatePostalCode", () => {
     expect(validatePostalCode("M3J3N3", "CA")).toBe(null)
   })
   it("returns error when postal code is invalid for CA", () => {
-    expect(validatePostalCode("YY", "CA")).toBe("This field is required")
+    expect(validatePostalCode("YY", "CA")).toBe("Postal code is invalid")
   })
 })
 
@@ -206,7 +206,7 @@ describe("formValidators/validateAddress", () => {
       const result = validateAddress(address)
 
       expect(result.hasErrors).toEqual(true)
-      expect(result.errors.postalCode).toEqual("This field is required")
+      expect(result.errors.postalCode).toEqual("Postal code is invalid")
     })
 
     it("returns no error for a valid postalCode for US", () => {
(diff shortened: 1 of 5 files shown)


erikdstock:
This seems like a reasonable change, but just to make sure we don't step on each others' toes or duplicate work i believe that NX is also working on a more robust address validation project right now! Will link in slack.

lidimayra:
Oh this is super cool!! CCing @pvinis here as I think he'll be happy to see this. He suggested this exact change yesterday!

And thanks for the heads up, @erikdstock ! The address validation project was deprioritized on NX side, so I believe it's all good!! 🚀

src/Apps/Order/Utils/formValidators.tsx
```diff
@@ -8,6 +8,15 @@ export const validatePresence = (value: string): string | null => {
   return null
 }
 
+export const validatePostalCode = (postalCode: string, countryCode: string) => {
+  let postalCodeRegex: RegExp =
+    countryCode === "US"
+      ? /^([0-9]{5})(?:[-\s]*([0-9]{4}))?$/
+      : /^([A-Z][0-9][A-Z])\s*([0-9][A-Z][0-9])$/
+
+  return postalCodeRegex.test(postalCode) ? null : "This field is required"
```
iskounen, line 17: **suggestion:** If a postal code is provided, but does not match the regex, this error message won't make sense to the user. I would suggest keeping `validatePresence(postalCode)` in the `postalCode` validation check below.

src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
```diff
@@ -1,5 +1,24 @@
...
+  validatePresence,
+  validatePostalCode,
+} from "Apps/Order/Utils/formValidators"
+
+describe("formValidators/validatePostalCode", () => {
+  it("returns error when postal code is invalid for US", () => {
+    expect(validatePostalCode("XX", "US")).toBe("This field is required")
+  })
+  it("returns no error when postal code is valid for US", () => {
+    expect(validatePostalCode("15601", "US")).toBe(null)
```
iskounen, line 13: **suggestion (non-blocking):** Your US ZIP Code regex supports 9-digit ZIP Codes as well, so it would be nice to add a test to ensure that we don't lose that ability if the regex is changed in the future.

iskounen:
Hi @desaijay26 👋 Thank you for your interest in contributing to Force. It sounds like you are having trouble with the current OSS setup instructions that we have documented. It is possible that our documentation needs to be updated to reflect recent changes to Force’s infrastructure.

To surface this problem to the team, please open a new issue in this repository with the steps that you followed and the errors that you encountered. We prefer to keep our pull request discussions focused on the change at hand, so opening an issue is the best way to get support 🙂

src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
```diff
@@ -1,5 +1,24 @@
...
+  })
+  it("returns error when postal code is invalid for CA", () => {
```
starsirius, line 15: The description of this test and the one below should be swapped?

src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
```diff
@@ -1,5 +1,24 @@
...
-import { validateAddress, validatePresence } from "../formValidators"
+import {
+  validateAddress,
+  validatePresence,
+  validatePostalCode,
+} from "Apps/Order/Utils/formValidators"
+
+describe("formValidators/validatePostalCode", () => {
+  it("returns error when postal code is invalid for US", () => {
+    expect(validatePostalCode("XX", "US")).toBe("This field is required")
```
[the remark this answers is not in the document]
AlicanAkyuz: I'll check with Viviana and Erin about this. thank you.

src/Apps/Order/Utils/formValidators.tsx
AlicanAkyuz: makes sense. that way we can return/display different messages to the user for empty and wrong postal code. thank you!

src/Apps/Order/Utils/__tests__/formValidators.jest.tsx
AlicanAkyuz: thank you for this suggestion.

PR #11197, opened 2022-10-18
fix: show edition set size on single edition works

mdole:

### Description

Due to some faulty logic in the artwork sidebar, we were hiding dimensions on single-edition works (i.e. editioned works where only one edition is available on Artsy).

### Before:

Screen Shot 2022-10-18 at 21 28 30

### After: 
Screen Shot 2022-10-18 at 21 28 48


anandaroop (approved):
Nice work 😎

PR #11211, opened 2022-10-19
fix: Remove bogus canonical tag from auction results GRO-1357

jonallured:
I can't be sure but I believe the intention here was to set the page title but rather than using the react-head approach we were using the MetaTags approach. This didn't work well because it caused us to render _all_ the meta tags again (incorrectly) so this PR just gets us back on track.

https://artsyproduct.atlassian.net/browse/GRO-1357

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -1,4 +1,5 @@
 import * as DeprecatedAnalyticsSchema from "@artsy/cohesion/dist/DeprecatedSchema"
+import { Title } from "react-head"
 import { ContextModule, Intent } from "@artsy/cohesion"
 import {
   Box,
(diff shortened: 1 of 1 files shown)


dzucconi:
Wouldn't the correct fix to remove the extra canonical tag — where ever that is? We use `MetaTags` because that component does everything correctly and consistently.

jonallured:
Yeah there might be more to this, I'm not totally sure. What I found when comparing the auctions results page with the overview and works for sale pages was that all of them get these:

https://github.com/artsy/force/blob/main/src/Apps/Artist/Components/ArtistMeta.tsx#L164
https://github.com/artsy/force/blob/main/src/Apps/Artist/Components/ArtistMetaCanonicalLink.tsx

But only the auctions results page was ALSO using `MetaTags`. So my assumption is that there's something special about the artist pages where we wanna do the meta tags like this. 🤷  But maybe not - maybe this was done before we had `MetaTags`? IDK, but I believe this PR does fix the reported bug - what do you think I should do?

dzucconi:
Interesting; lemme take a closer look!

dzucconi:
OK yeah I see now: the ArtistMeta specifies everything for the artist app and you're right, we can just change the title.

(I'm pretty sure though that a lot of those tags are either wrong or pointless.)

dzucconi (approved):

dzucconi:
I do think it'd be smart to revisit this using the `MetaTags` implementation and remove some of those pointless old open graph tags.

PR #11171, opened 2022-10-13
feat(FX-4322): Auctions Watch Lot - In Product Information Overlay

dimatretyak:
The type of this PR is: **Feat**

This PR solves [FX-4322]

Review app: [watch-log-pop-up](https://watch-log-pop-up.artsy.net/)

Live auctions: [with specified registration deadline](https://watch-log-pop-up.artsy.net/auction/dima-live-auction-test-1), [without specified registration deadline](https://watch-log-pop-up.artsy.net/auction/dima-live-auction-test-2)

### Description
* We will only display the “register to bid” button (“registration pending”, “verify identity”, “approved to bid” and etc buttons **will not** be displayed) in the pop-up
* We show the pop-up until the user register (pressed “register to bid” button)
* Do not show popover if registration deadline is specified and has already passed

#### With specified registration deadline
https://watch-log-pop-up.artsy.net/artwork/634ed29627729d00066067aa

https://user-images.githubusercontent.com/3513494/196486036-cc814e9f-1da1-4761-9a2b-d44f46f6858b.mp4

#### Without specified registration deadline
https://watch-log-pop-up.artsy.net/artwork/634ed371bd5df100066e34b2

https://user-images.githubusercontent.com/3513494/196490382-013f416e-2517-45c8-8b46-5f3c5db2d591.mp4

[FX-4322]: https://artsyproduct.atlassian.net/browse/FX-4322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11227, opened 2022-10-21
feat(FX-4388): display register to bid popover after auth flow

dimatretyak:

Related Force PR: artsy/force#11171

This PR solves [FX-4388]

Review app: [show-register-to-bid-popup-after-auth](https://show-register-to-bid-popup-after-auth.artsy.net/)

Artwork example: https://show-register-to-bid-popup-after-auth.artsy.net/artwork/henri-van-noordenburg-untitled-1

### Demo
https://user-images.githubusercontent.com/3513494/197199913-824e6591-d82e-413f-8255-d5a92cc2d18b.mp4

[FX-4388]: https://artsyproduct.atlassian.net/browse/FX-4388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
@@ -7,7 +7,7 @@ import {
   Popover,
   Text,
 } from "@artsy/palette"
-import * as React from "react"
+import { useEffect, useCallback, useState } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useSaveArtwork } from "Components/Artwork/SaveButton/useSaveArtwork"
 import { ArtworkActionsSaveButton_artwork$data } from "__generated__/ArtworkActionsSaveButton_artwork.graphql"
@@ -15,6 +15,7 @@ import { UtilButton } from "./UtilButton"
 import { ArtworkAuctionRegistrationPanelFragmentContainer } from "Apps/Artwork/Components/ArtworkImageBrowser/ArtworkAuctionRegistrationPanel"
 import { useSystemContext } from "System"
 import { useTranslation } from "react-i18next"
+import { mediator } from "Server/mediator"
 
 interface ArtworkActionsSaveButtonProps {
   artwork: ArtworkActionsSaveButton_artwork$data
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
```diff
@@ -50,12 +52,32 @@ const ArtworkActionsSaveButton: React.FC<ArtworkActionsSaveButtonProps> = ({
...
+      setPopoverVisible(true)
+    }
+  }, [ignoreAuctionRegistrationPopover])
+
+  const onPopoverClosed = useCallback(() => {
+    setPopoverVisible(false)
+  }, [])
+
+  useEffect(() => {
+    mediator.on("artwork:save", openAuctionRegistrationPopover)
```
dzucconi, line 66: Interesting that this is still useful — as we've been looking to remove this event bus for a while. Is there an easy way around not using it?

dimatretyak: Unfortunately, I didn't find another easy way. I tried this solution, but since the component is unmounted and `shouldDisplayPopover` is reset, it won't work 🤔

```javascript
const shouldDisplayPopover = useRef(false)

...

useEffect(() => {
  if (isLoggedIn && shouldDisplayPopover.current) {
    shouldDisplayPopover.current = false
    setPopoverVisible(true)
  }
}, [isLoggedIn])

...

<Popover {...other props} >
  {({ anchorRef }) => {
    return (
      <UtilButton
        {...other props}
        onClick={() => {
          handleSave()

          if (!ignoreAuctionRegistrationPopover) {
            if (isLoggedIn) {
              setPopoverVisible(true)
              return
            }

            shouldDisplayPopover.current = true
          }
        }}
      />
    )
  }}
</Popover>
```

dzucconi: Yeah, could do something with a query param perhaps. But I guess I'm not opposed to us having an event bus — it's a little weird though in terms of patterns that exist. The more React way would be to set something in the system context or some other parent context.

dimatretyak: If you don't mind, then I'll leave the current solution and on one of the feature Friday I'll see what can be done with an event bus

dzucconi: Sounds good to me! Thanks!

dblandin (approved):

PR #11238, opened 2022-10-24
fix(i18n): globally mocks translation function

dzucconi:
This fixes the mock so that you don't have to use a monkey patched render etc.

### DIFF
--- a/src/Components/NavBar/__tests__/NavBar.jest.tsx
+++ b/src/Components/NavBar/__tests__/NavBar.jest.tsx
@@ -2,11 +2,9 @@ import { BellIcon, EnvelopeIcon, SoloIcon } from "@artsy/palette"
 import { SystemContextProvider } from "System"
 import { useTracking } from "react-tracking"
 import { mount } from "enzyme"
-import { NavBar } from "../NavBar"
-import { NavBarMobileMenuInboxNotificationCount } from "../NavBarMobileMenu/NavBarMobileMenuInboxNotificationCount"
+import { NavBar } from "Components/NavBar/NavBar"
+import { NavBarMobileMenuInboxNotificationCount } from "Components/NavBar/NavBarMobileMenu/NavBarMobileMenuInboxNotificationCount"
 import { mediator } from "Server/mediator"
-import { I18nextProvider } from "react-i18next"
-import i18n from "System/i18n/i18n"
 
 jest.mock("Components/Search/SearchBar", () => {
   return {
@@ -37,9 +35,7 @@ describe("NavBar", () => {
   const getWrapper = ({ user = null, isEigen = false } = {}) => {
     return mount(
       <SystemContextProvider user={user} isEigen={isEigen}>
-        <I18nextProvider i18n={i18n}>
-          <NavBar />
-        </I18nextProvider>
+        <NavBar />
       </SystemContextProvider>
     )
   }
--- a/src/DevTools/mockReactI18n.tsx
+++ b/src/DevTools/mockReactI18n.tsx
@@ -1,0 +1,15 @@
+import i18n from "i18next"
+import config from "System/i18n/config"
+
+i18n.init(config)
+
+export const useTranslation = () => {
+  return {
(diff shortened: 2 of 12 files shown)


src/DevTools/mockReactI18n.tsx
```diff
@@ -0,0 +1,14 @@
...
+export const useTranslation = () => {
+  return {
+    t: (path: string | string[]) => {
+      const key = Array.isArray(path) ? path.join(".") : path
+      const value = get(translations, key, key)
+
+      return value
+    },
+  }
+}
```
dzucconi, line 14: cc @damassi

dzucconi: OK looks like I'll just have to inject the i18n `t` function here directly, preloaded with the en file. Because we use some of the other 'features' (like pluralization).

gkartalis (approved):
👌 thanks for fixing that, way much cleaner! Looks good when the pluralization test starts passing 👀

src/System/i18n/config.ts
```diff
@@ -0,0 +1,24 @@
+import { getENV } from "Utils/getENV"
+
+// TODO: Serve static assets on the server-side
```
dzucconi, line 3: @gkartalis @lidimayra what does this mean in this context?

src/System/i18n/config.ts
```diff
@@ -0,0 +1,24 @@
...
+  debug: getENV("ENABLE_I18N_DEBUG") === "true",
+  resources: {
+    ["en-US"]: {
+      translation: translations,
+    },
+  },
+  detection: {
+    order: ["querystring"],
+    lookupQuerystring: "locale",
+  },
```
dzucconi, line 16: We won't be using this option? Is there anything else in this config can we remove?

src/Components/NavBar/__tests__/NavBar.jest.tsx
```diff
@@ -2,11 +2,9 @@ import { BellIcon, EnvelopeIcon, SoloIcon } from "@artsy/palette"
 import { SystemContextProvider } from "System"
 import { useTracking } from "react-tracking"
 import { mount } from "enzyme"
-import { NavBar } from "../NavBar"
-import { NavBarMobileMenuInboxNotificationCount } from "../NavBarMobileMenu/NavBarMobileMenuInboxNotificationCount"
+import { NavBar } from "Components/NavBar/NavBar"
+import { NavBarMobileMenuInboxNotificationCount } from "Components/NavBar/NavBarMobileMenu/NavBarMobileMenuInboxNotificationCount"
```
dzucconi, line 6: Unrelated but I suspect we can just run an autofix for all of these.

src/System/i18n/config.ts
gkartalis: not 100% sure what this is referring to since I wasn't involved in the planning/implementation, but I remember some conversations for the actual translations being moved to a 3rd party like [transifex](https://www.transifex.com/) or [locize](https://locize.com/) , maybe it was something like this 🤔

lidimayra: This is actually related to what was commented here: https://github.com/artsy/force/pull/9797#discussion_r840560932

We paired on it a bit, but we never managed to actually make it work, so we decided to move forward without it, as we didn't want it to block the implementation because of this.

dzucconi: What's the benefit of doing that vs simply requiring?

lidimayra: As I understand, this would be an optimization for the future. The server-side rendering would prevent us from loading unnecessary files if we have several JSON files for different locales.

@araujobarret might have the proper context on this.

damassi: > this would be an optimization for the future

We should probably cross that bridge when we get there and keep it simple for now

dzucconi: Presumably we'd only load this JSON in memory on the server boot — It's not going to affect performance in any meaningful way.

PR #11240, opened 2022-10-24
fix(theme): uses a single theme provider

dzucconi:
Noticed that components weren't adhering to the design system in onboarding. Turns out it was sitting inside of the v2 theme context. This PR removes that possibility from happening in the future.


jonallured (approved):
🏆

PR #11242, opened 2022-10-25
fix: Type guard in logger

AlicanAkyuz:

### Description
Adding a type guard for `shouldLogErrorToSentry` property in error logger.


dzucconi (approved):
🙏

PR #11249, opened 2022-10-27
feat:PX-5250 unlisted artwork banner

nspinazz89:
The type of this PR is: Feat

This PR solves: [PX-5250]

### Description

This PR adds a non-scrollable full width banner to the Artwork page when it's been `unlisted`. This lets the collector know what that this artwork is not findable to the general public on Artsy. The banner contains a link that pops open a modal that further explains the value proposition.

image
image

[PX-5250]: https://artsyproduct.atlassian.net/browse/PX-5250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -76,7 +76,8 @@ const BelowTheFoldArtworkDetails: React.FC<BelowTheFoldArtworkDetailsProps> = ({
 export const ArtworkApp: React.FC<Props> = props => {
   const isNewArtworkSidebarEnabled = useFeatureFlag("fx-force-artwork-sidebar")
   const { artwork, me, referrer, tracking, shouldTrackPageView } = props
-  const unlistedArtwork = artwork?.visibilityLevel == "UNLISTED"
+  const showUnlistedArtworkBanner =
+    artwork?.visibilityLevel == "UNLISTED" && artwork?.partner
 
   const trackPageview = useCallback(() => {
     const { listPrice, availability, is_offerable, is_acquireable } = artwork
@@ -181,7 +182,7 @@ export const ArtworkApp: React.FC<Props> = props => {
       {artwork.sale && (
         <CascadingEndTimesBannerFragmentContainer sale={artwork.sale} />
       )}
-      {unlistedArtwork && artwork.partner && (
+      {showUnlistedArtworkBanner && (
         <UnlistedArtworkBannerFragmentContainer partner={artwork.partner} />
       )}
 
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -29,6 +29,7 @@ import { UseRecordArtworkView } from "./useRecordArtworkView"
 import { Router, Match } from "found"
 import { WebsocketContextProvider } from "System/WebsocketContext"
 import { CascadingEndTimesBannerFragmentContainer } from "Components/CascadingEndTimesBanner"
+import { UnlistedArtworkBannerFragmentContainer } from "Components/UnlistedArtworkBanner"
 import { useCallback, useEffect } from "react"
 import { useFeatureFlag } from "System/useFeatureFlag"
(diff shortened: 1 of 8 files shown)


src/Components/UnlistedArtworkBanner.tsx
```diff
@@ -0,0 +1,47 @@
...
+            onClick={() => {
+              setOpen(true)
+            }}
+          >
+            private listing
+          </Clickable>{" "}
+          {partner?.name ? `from ${partner.name}.` : "."}
+        </Text>
+      </FullBleedBanner>
+      <UnlistedBannerModal show={open} onClose={() => setOpen(false)} />
```
nspinazz89, line 33: Was following the pattern of extracting the modal out into its own component here.

src/Apps/Artwork/Components/UnlistedBannerModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <ModalDialog
+      onClose={onClose}
+      title="Private Listings"
+      footer={
+        <Button onClick={onClose} width="100%">
+          OK
+        </Button>
+      }
+    >
+      <Text variant="sm" data-testid="unlisted-modal-text">
```
nspinazz89, line 25: It was super tricky to get the text without these test ids. But I saw that we were using these data-testids to facilitate that in a couple other places.

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -75,6 +76,8 @@ const BelowTheFoldArtworkDetails: React.FC<BelowTheFoldArtworkDetailsProps> = ({
 export const ArtworkApp: React.FC<Props> = props => {
   const isNewArtworkSidebarEnabled = useFeatureFlag("fx-force-artwork-sidebar")
   const { artwork, me, referrer, tracking, shouldTrackPageView } = props
+  const showUnlistedArtworkBanner =
+    artwork?.visibilityLevel == "UNLISTED" && artwork?.partner
```
leamotta, line 80: wondering if we can remove the partner validation based on the [slack discussion](https://artsy.slack.com/archives/C9ZJYFDNF/p1666805404390339), what do you think?

src/Apps/Artwork/Components/UnlistedBannerModal.tsx
leamotta: It was not rendering or what was the problem? I wonder what that property is used for 🤔

src/Apps/Artwork/ArtworkApp.tsx
nspinazz89: so we get type errors when you try to remove the partner check that I wasn't able to resolve even with doing partner?  in the props of the element. It still works, but typescript doesn't like it. And it would fail in CI

leamotta (dismissed):
Not an FE expert but this looks good to me!

src/Apps/Artwork/Components/UnlistedBannerModal.tsx
jpotts244: +++ yes definitely pro to the testID, I was running into the issue on my side too. 
It was because of the nested child elements which contained different parts of the text, needed a better matcher to find the full sentence and not just the text of the deepest element

src/Apps/Artwork/ArtworkApp.tsx
jpotts244: [side note]: Wonder if there are stricter type checks in eigen, I actually was getting yelled at by typescript for doing the string check, 
Needed it to be of [type enum ](https://github.com/artsy/eigen/pull/7540/files#diff-9f418fccaa09fb8ea23d492fa300d6e3be69881dc6fbe84fc3c979e114e36533R36-R40) and then [check the object](https://github.com/artsy/eigen/pull/7540/files#diff-9f418fccaa09fb8ea23d492fa300d6e3be69881dc6fbe84fc3c979e114e36533R123)

jpotts244 (approved):
🏅 Seems reasonable to me!

PR #11259, opened 2022-10-27
feat: Secretly support Braze content cards GRO-1385

jonallured:
This PR adds support on the homepage for a query param that will show us Braze content cards. This is useful as we build towards completely migrating to Braze for this content. An intermediary step like this let's us tinker in the Braze dashboard with the cards before users actually see them.

A follow up PR should include:

* proper placeholder
* log impressions to Braze
* log clicks to Braze
* suppress segment tracking
* (optionally) set a standard height to avoid ragged bottoms

And so I can work through that while other folks work on getting the stuff done in Braze.

https://artsyproduct.atlassian.net/browse/GRO-1385

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Home/Components/HomeContentCards.tsx
+++ b/src/Apps/Home/Components/HomeContentCards.tsx
@@ -1,0 +1,3 @@
+export const HomeContentCards = () => {
+  return <h1>Content Cards Go Here</h1>
+}
--- a/src/Apps/Home/Components/HomeContentCards.tsx
+++ b/src/Apps/Home/Components/HomeContentCards.tsx
@@ -1,3 +1,64 @@
-export const HomeContentCards = () => {
-  return <h1>Content Cards Go Here</h1>
+import React, { useEffect, useState } from "react"
+import { HeroCarousel } from "Components/HeroCarousel/HeroCarousel"
+import {
+  HomeHeroUnit,
+  StaticHeroUnit,
+} from "Apps/Home/Components/HomeHeroUnits/HomeHeroUnit"
+
+interface ContentCardProps {
+  card: any
+  index: any
+}
+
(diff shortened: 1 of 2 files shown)


src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { useEffect, useState } from "react"
+import { HeroCarousel } from "Components/HeroCarousel/HeroCarousel"
+import {
+  HomeHeroUnit,
+  StaticHeroUnit,
+} from "Apps/Home/Components/HomeHeroUnits/HomeHeroUnit"
+
+interface ContentCardProps {
+  card: any
+  index: any
```
dzucconi, line 10: Can we type these?

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const heroUnit = {
+    backgroundImageURL: card.imageUrl,
+    creditLine: extras.credit,
+    heading: extras.label,
+    href: card.url,
+    linkText: card.linkText,
+    subtitle: card.description,
+    title: card.title,
+  } as StaticHeroUnit
```
dzucconi, line 24: `const heroUnit: StaticHeroUnit = {` would be preferable

src/Apps/Home/Components/HomeContentCards.tsx
jonallured: omg I totally forgot about this thank you - will fix!

jonallured: For those Braze types do I just do something like `yarn add -D @braze/web-sdk`? And then I can import them? And that won't make it into the prod bundle because of it being a dev dep?

PR #11265, opened 2022-10-28
fix: correctly scrolls considering sticky offset

dzucconi:
Closes [GRO-1384](https://artsyproduct.atlassian.net/browse/GRO-1384)

[Diff ignoring whitespace](https://github.com/artsy/force/pull/11265/files?diff=unified&w=1)

This is an interesting one. Both in the problem and the solution(s).

## TLDR
If you want to programmatically scroll to an element, set a position to scroll to using `useJump/Jump`:

```tsx
import { useJump, Jump } from "Utils/Hooks/useJump"

// ...

const { jumpTo } = useJump()

// ...

<Button onClick={() => jumpTo("some-unique-identifier")}></Button>

// ...

<Jump id="some-unique-identifier" />

````

------

Firstly, _the problem_: In a number of places we use `Sticky` pieces of UI and _also_ need to programmatically scroll elements into view. Normally this isn't an issue, in our case the problem is two fold: our persistent nav can change height, and with elements sticking, we need to know the height of the nav and the combined height of whatever is also stuck to the top of the screen.

This is one of those "I'm not sure if this ever worked" kind of things, though I really did think it was working, documentation on these hooks said it worked, and no one told me it didn't. 😅  Unfortunately, while although the `useScrollTo` hook correctly calculated the height of the nav, it would ignore the `Sticky` elements.

------

Initially my first design for solving this was simple: "OK we just need to know what's stuck when you click on a scroll to trigger" — so I stored the `stuck` status in the `Sticky` context and summed those and... additional elements might stick to the nav bar as you're scrolling, so what you _really_ need to do is know what is stuck to the viewport once you're _already_ scrolled to that element.

Additionally what makes this a bit more annoying than most is that you need to coordinate behavior between remote calls. Like ideally the hook could return bound refs that you could then just sprinkle around and then scroll to, but scrolling typically happens between components that know nothing about one another and implementing a context registry seems like overkill.

So ultimately this PR introduces a new hook and component `useJump` and `Jump`. `Jump` accepts an `id` and `useJump` returns a function `jumpTo` which accepts an `id`.

To explain how this works I need to explain how `Sticky` stacking works. Each `Sticky` registers itself with an entry in a context provider and that entry keeps track of the node's `height`. The entries are made according to where they are in the page. So elements at the bottom of the page will be at the end of the array. When an element sticks, it finds its index in that array and then sums up the top offset to find where it should stick.

Internally `Jump` registers an entry in this same `Sticky` context with a `height` of `0`. When we click to scroll to it, we can find its index in that array and similarly sum up the top offset.

------

And there's still 3 other methods of scrolling that exist in the app that I need to unwind:

- [ ] `scrollIntoView`
- [ ] `ScrollIntoView`
- [ ] `window.scrollTo`

### CODE REVIEW

File: src/Apps/Artist/Components/ArtistInsights/ArtistInsightPills.tsx [hunk 1]
```diff
@@ -1,42 +1,40 @@
-import { Flex, Pill, Spacer } from "@artsy/palette"
+import { Flex, Pill } from "@artsy/palette"
 import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { ArtistInsightPills_artist$data } from "__generated__/ArtistInsightPills_artist.graphql"
-import { useScrollTo } from "Utils/Hooks/useScrollTo"
```
damassi, line 5: should we use an eslint import restriction?

dzucconi: I deleted the entire file.

src/Utils/Hooks/useJump.tsx
```diff
@@ -0,0 +1,73 @@
...
+  offset?: number
+}
+
+const NAMESPACE = "JUMP"
+
+/**
+ * Returns a `jumpTo` function that scrolls with an automatically calculated offset
+ * based on the nav and the current state of the sticky component.
+ */
+export const useJump = ({ behavior = "smooth", offset = 0 }: UseJump = {}) => {
```
damassi, line 18: Thoughts on this living in Palette?

dzucconi: It's dependent on the `Sticky` implementation.

damassi (dismissed):
A+

src/Components/Sticky/Sticky.tsx
```diff
@@ -28,35 +28,41 @@ import { useSticky } from "./StickyProvider"
  */
 export const Sticky: React.FC<
   Pick<ReactStickyProps, "bottomBoundary"> & {
+    // TODO: Remove this prop!
     withoutHeaderOffset?: boolean
```
dzucconi, line 32: Plan is to get rid of this by just making header a `Sticky`.

src/Components/Sticky/Sticky.tsx
```diff
@@ -28,35 +28,41 @@ import { useSticky } from "./StickyProvider"
...
   const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
+
   const [stuck, setStuck] = useState(false)
+
   const containerRef = useRef<HTMLDivElement | null>(null)
+
   const headerOffset = withoutHeaderOffset ? 0 : isMobile ? mobile : desktop
 
   useEffect(() => {
+    console.log("jump registering", containerRef.current)
```
dzucconi, line 48: Woops

src/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -142,36 +139,36 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
...
-                    scrollTo()
-                  }}
-                  noTitle
-                />
-                {(!customValue || customValue < minPrice) && (
-                  <MinPriceWarning
-                    isPriceRange={!!artwork?.isPriceRange}
-                    minPrice={asCurrency(minPrice) as string}
-                    orderID={order.internalID}
+              <Jump key="price-option-custom" id="price-option-custom">
```
dimatretyak, line 151: **question**: maybe we should pass `key` prop for `BorderedRadio`?

src/Apps/Shows/Routes/ShowsCity.tsx
```diff
@@ -60,14 +60,10 @@ export const ShowsCity: React.FC<ShowsCityProps> = ({
...
-  const currentShowsRef = useRef<HTMLDivElement | null>(null)
-  const { scrollTo } = useScrollToElement({
-    selectorOrRef: currentShowsRef,
-    offset: 20,
-  })
+  const { jumpTo } = useJump({ offset: 20 })
 
   const handleClick = (cursor: string, page: number) => {
-    scrollTo()
+    jumpTo("CurrentShows")
```
dimatretyak, line 66: **question**: maybe it would be better if we keep `id` as constant?

src/Components/Pagination/index.tsx
```diff
@@ -17,6 +17,7 @@ export interface PaginationProps {
     | Pagination_pageCursors$data
     | CommercePagination_pageCursors$data
     | undefined
+  // FIXME: Replace with `useJump` implementation
```
dimatretyak, line 20: 🙏

dimatretyak (dismissed):
Greta change, thank you!

src/Utils/Hooks/useJump.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
+
+  const { mobile, desktop } = useNavBarHeight()
+
+  const { stickies } = useSticky()
+
+  const jumpTo = (
```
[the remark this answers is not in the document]
dzucconi: Not totally necessary, I think, but a fine thing to do just-in-case.

src/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx
```diff
@@ -197,21 +199,15 @@ export const BuyerGuaranteeIndex: FC = () => {
...
+        <Column span={6} pt={[0, 6]} order={2}>
           <Text variant="lg-display">Authenticity Guarantee</Text>
 
           <Text variant="sm" mt={2}>
             We are dedicated to being the world’s most trustworthy marketplace
             to buy and sell art. In the rare case that a work purchased through
             Artsy’s secure checkout is found to be inauthentic, report the issue
             within 180 days from delivery arrival and we’ll help facilitate a
-            refund for the sale.
+            refund for the sale.wh
```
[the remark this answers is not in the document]
dzucconi: 😅

src/Apps/Order/Components/PriceOptions.tsx
dzucconi: Yeah, [RadioGroup](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/RadioGroup/RadioGroup.tsx) is a very weird old component that needs to be revisited — I'll modernize it at some point 🤷

src/Apps/Shows/Routes/ShowsCity.tsx
dzucconi: Updated in this file; though I don't really feel strongly either way. I suppose it could help to export these to keep them in sync across multiple components. Though now that it's an `id` on a purpose built component, you know now what it does and it's less likely to be removed or changed because of that.

PR #11275, opened 2022-10-31
fix(partner): updates artist list and scrolling

dzucconi:
Working towards [GRO-1383](https://artsyproduct.atlassian.net/browse/GRO-1383)

[Diff without whitespace changes](https://github.com/artsy/force/pull/11275/files?diff=unified&w=1)

This begins to update some of the UI patterns in the Partner app. Starting with, primarily the artists list. Here instead of manually columnizing and then inserting into a carousel, we just simply use CSS columns. This looks the same or better in some instances. Only weirdness here is a little workaround to implement column headers. Additionally the scrolling has been simplified by using the new `Jump/useJump` component/hook. We can get rid of the double sub-nav render and all the manual offsetting.


src/Apps/Partner/Components/PartnerArtists/PartnerArtistList/PartnerArtistList.tsx
```diff
@@ -1,95 +1,150 @@
...
-  artists: PartnerArtistList_artists$data
-  distinguishRepresentedArtists: boolean
-  partnerSlug: string
-  scrollTo: ScrollIntoViewProps
-  displayFullPartnerPage: boolean
-}
+// Values here used to implement column headers in CSS
+const PADDING = 0.5
+const OFFSET =
+  parseInt(THEME_V3.textVariants.sm.lineHeight, 10) + PADDING * 10 * 2
```
dzucconi, line 12: Honestly we should just design around this. I don't like this hack. It's made even weirder because of Safari. Basically; we calculate an offset value here, then use that to negatively transform the column header (because Safari can't handle negative top margins that break out of a container (!?)), then apply a negative margin to the first element in the column.

src/Utils/ScrollIntoView.tsx
```diff
@@ -1,26 +0,0 @@
...
-        width={["100%"]}
-        onClick={() => {
-          scrollIntoView(this.props)
-        }}
-      >
-        {this.props.children}
-      </Box>
-    )
-  }
-}
```
dzucconi, line 26: 👋

damassi (approved):

PR #11278, opened 2022-11-01
fix: Hide price estimate section when artwork has a submission

olerichter00:
## Description

We need to hide the price estimate section on the My Collection artwork page when the artwork has a submission.


dariakoko (approved):
⭐x5

PR #11280, opened 2022-11-01
fix(partner): more modernization

dzucconi:
Closes: [GRO-1383](https://artsyproduct.atlassian.net/browse/GRO-1383)

OK, `PartnerApp` should be in good shape now. On to the `FairApp` next.


araujobarret (commented):
Good stuff 🚀

PR #11283, opened 2022-11-01
fix: scroll clean up

dzucconi:
Working towards [GRO-1401](https://artsyproduct.atlassian.net/browse/GRO-1401)

More programmatic scroll clean up — replacing most of the `scrollIntoView` usage. Remaining is components that use the `Pagination#scrollTo`  prop which will be in the next PR.


anandaroop (approved):
🙏🏽

PR #11288, opened 2022-11-02
chore(volley): updates image metrics

dzucconi:
Closes [GRO-1406](https://artsyproduct.atlassian.net/browse/GRO-1406)

* I can't use the `pageType` tagging that the page load metrics use. Those rely on something server-side and aren't updated as the client-side route changes. Instead I went with just pulling off the first path dir, tagged with `root-path`. Now that we no longer have top-level vanity routes, this should be sufficent: `artwork`, `article`, `collect`... etc.
* I added a tag for device pixel ratio. Not sure if this is necessary but would be nice to have to better determine where large images are coming from.
* I didn't add a cap to the transfer size because I just don't know where that top is. It'd be good to get some insight here.

### DIFF
--- a/src/Utils/Hooks/__tests__/useImagePerformanceObserver.jest.ts
+++ b/src/Utils/Hooks/__tests__/useImagePerformanceObserver.jest.ts
@@ -45,25 +45,34 @@ describe("useImagePerformanceObserver", () => {
 
     result.current.current = [
       {
-        initiatorType: "img",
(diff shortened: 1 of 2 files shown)


joeyAghion (approved):
👍 from me.

Noting that the `transfer-size` tag's meaning and values will change with this release. Might be a little funky to slice in datadog, but presumably we can filter by time window to _only_ see the updated bucket values.

PR #11292, opened 2022-11-02
fix(artwork): correctly scales video container

dzucconi:
Closes [GRO-1302](https://artsyproduct.atlassian.net/browse/GRO-1302)

The problem is two fold:
* We need to determine if the height of the video is taller than the height of the image, and include that in our value for `maxHeight`
* We need to scale the video within the `MAX_DIMENSION` (800px) in case it's larger than that (which is what images do).


src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
```diff
@@ -35,10 +32,11 @@ const ArtworkVideoPlayer: FC<ArtworkVideoPlayerProps> = ({
       {...rest}
     >
       <ResponsiveBox
-        mx={[0, "70px"]}
```
dzucconi, line 38: I realize this is for the UI — I'm wondering if it's needed. Going to come back to this in the morning.

I think we could just show/hide the arrows like we do in the zoom view.

src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
```diff
@@ -35,10 +32,11 @@ const ArtworkVideoPlayer: FC<ArtworkVideoPlayerProps> = ({
...
-        maxWidth="100%"
-        aspectWidth={activeVideo.width}
-        aspectHeight={activeVideo.height}
+        maxWidth={MAX_DIMENSION}
+        maxHeight={MAX_DIMENSION}
```
dzucconi, line 36: Here we just use the `ResponsiveBox` to scale the video container since we can't do it via GraphQL or with the `resize` function. This will fit it inside of an 800x800px box and then responsively scale it down if it's smaller.

src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
```diff
@@ -61,10 +59,11 @@ const ArtworkVideoPlayerFragmentContainer = createFragmentContainer(
       fragment ArtworkVideoPlayer_artwork on Artwork {
         figures {
           ... on Video {
-            type: __typename
+            __typename
             url
-            height
-            width
+            # Fields need to be aliased to prevent conflicting types
```
dzucconi, line 64: Annoying!

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -28,23 +29,51 @@ export const ArtworkImageBrowser: React.FC<ArtworkImageBrowserProps> = ({
...
-  const maxHeight = Math.max(
-    ...images.map(image => {
-      const scaled = scale({
-        width: image.width!,
-        height: image.height!,
-        maxWidth: MAX_DIMENSION,
-        maxHeight: MAX_DIMENSION,
+  // Here we pre-emptively scale the figures to figure out which is going
+  // to have the largest height. We use this to set a fixed height on the container
+  // so that the UI doesn't shift around.
```
dzucconi, line 34: This was the main problem.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -28,23 +29,51 @@ export const ArtworkImageBrowser: React.FC<ArtworkImageBrowserProps> = ({
...
+              height: figure.height,
+              maxWidth: MAX_DIMENSION,
+              maxHeight: MAX_DIMENSION,
+            }).height
+          }
+
+          case "Video": {
+            return scale({
+              width: figure.videoWidth,
+              height: figure.videoHeight,
```
dzucconi, line 53: 😝

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkImageBrowser.tsx
```diff
@@ -28,23 +29,51 @@ export const ArtworkImageBrowser: React.FC<ArtworkImageBrowserProps> = ({
...
-    })
-  )
+  const defaultIndex = useMemo(() => {
+    return (
+      figures.findIndex(figure => {
+        if (!("isDefault" in figure)) return false
+        return !!figure.isDefault
+      }) ?? 0
+    )
+  }, [figures])
```
dzucconi, line 74: And actually; we should rename this since `isDefault` doesn't even make sense here. It should be like `viewInRoomIndex` or something.

damassi (approved):
Thank you for the fix!

src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
dzucconi: Re-added it; I'm going to circle back to get rid of this though.

PR #11294, opened 2022-11-03
fix(artwork): prevents re-render

dzucconi:
Closes: [GRO-1411](https://artsyproduct.atlassian.net/browse/GRO-1411)

So turns out this problem was a little weird than I had initially thought. I ticketed this thinking updating the index was causing re-renders — what was actually happening was a bit more subtle. This change: https://github.com/artsy/force/commit/f3dc33a53cafd1b4172f2f4067ab1111135e82f3#diff-9f79588cb81cf2099b264bdfb20a943d6b5eb82b78d6ccfbbfad94b0fbe0e528 mixed up the variables `index` and `i`. 

What we _want_ to be doing is render the swiper with all of the images in them. As you swipe through them it updates the index that you're viewing which in turn updates the progress indicator dots on the bottom. What this change did was render a swiper with N copies of the active image. So as you swiped through it would update that index which would then update the swiper to re-render the entire thing with N copies of the new image.

Thank you to @MrSltun for the bug report in #practice-web.


dzucconi:
Yeah, I'll make that change @jonallured — though I think what happened here was just a copy paste from the `Large` version of this component where it does in fact work like this (swapping out the active image entirely).

PR #11299, opened 2022-11-03
feat(video): Add analytics

damassi:

This PR solves [GRO-1362]

### Description

This adds analytics to our video player code. 

Going to merge this to get things to staging, but will follow up with any comments tomorrow before launch! 

cc @artsy/grow-devs 

[GRO-1362]: https://artsyproduct.atlassian.net/browse/GRO-1362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
```diff
@@ -12,12 +15,59 @@ interface ArtworkVideoPlayerProps extends BoxProps {
...
+
+  useEffect(() => {
+    window.addEventListener("blur", trackClickedPlayVideo)
+
+    trackViewedVideo()
+
+    return () => {
+      window.removeEventListener("blur", trackClickedPlayVideo)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
damassi, line 68: Only want this on mount, and never again

src/Apps/Artwork/Components/ArtworkVideoPlayer/__tests__/ArtworkVideoPlayer.jest.tsx
```diff
@@ -1,33 +1,132 @@
...
+    mockUseTracking.mockReturnValue({ trackEvent: jest.fn() })
+  })
+
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
 
   it("renders with video figure", () => {
-    const { container } = render(
-      <ArtworkVideoPlayer {...defaultProps([image, video])} />
```
damassi, line 22: @laurabeth - if a component is a relay component we should always be using our relay test tools which auto generate fixture data for us

src/Apps/Artwork/Components/ArtworkVideoPlayer/__tests__/ArtworkVideoPlayer.jest.tsx
```diff
@@ -1,33 +1,132 @@
...
+      await waitFor(() => {
+        expect(spy.mock.calls[1]).toEqual([
+          {
+            action: "clickedPlayVideo",
+            context_owner_id: "foo",
+            context_owner_slug: "bar",
+            context_owner_type: "artwork",
+          },
+        ])
+      })
```
damassi, line 129: It took me so freakin long to figure this one out 😓

src/Apps/Artwork/Components/ArtworkVideoPlayer/ArtworkVideoPlayer.tsx
```diff
@@ -12,12 +15,59 @@ interface ArtworkVideoPlayerProps extends BoxProps {
...
+  }, [])
```
damassi, line 69: On mount, and only on mount, no matter what

dzucconi: You need to return a teardown for that event handler or it will double bind on other pages

dzucconi: Theres also no real reason to not use exhaustive deps. You're seeing this effect get called more than once?

damassi: Well it was more: By passing those two functions in as deps I then had to wrap each in a `useCallback`, and then define the dependencies for each of those. All of that is totally irrrelevant and unnecessarily complicated if all i want to do is run an effect exactly one time, on mount

damassi: (Can fix if you want tho, just that the linter then flagged those other functions. Either way)

PR #11315, opened 2022-11-07
chore(storybook): Upgrade to latest enable webpack 5

damassi:

### Description

We've been getting some weird storybook build failures since adding `@artsy/icons`. However, upgrading to latest and enabling webpack 5 fixes it (and improves storybook compilation perf quite dramatically)


src/tests.ts
```diff
@@ -13,7 +13,6 @@ track.mockImplementation(() => x => x as any)
 useTracking.mockImplementation(() => ({ trackEvent: jest.fn() }))
 
 jest.mock("lodash/debounce", () => jest.fn(e => e))
-jest.mock("react-sizeme", () => jest.fn(() => d => d))
```
damassi, line 16: Nowhere in the app, not even in `yarn.lock`, yet never failed before now which is 🤔

PR #11332, opened 2022-11-09
feat(artwork): hide next/prev nav when idle

dzucconi:
OK — there's no ticket here; so we can take or leave this. Just wanting to make it so that videos fill the space more optimally, as if they were sized like the images.

Previously we had `70px` of horizontal padding on either side to accommodate the next/previous arrows. This PR does a few things:

* Removes that padding so that the arrows overlay the video directly, just like they currently do on images.
* Changes the blend mode to difference so that they are always visible.
* Extracts the idle detection hook that I used in the deep zoom viewer to hide the nav after a few seconds of inactivity and to redisplay it when you move the cursor.
* Reduced the click area of the arrows so as to not interfere with the video UI.
* Added index support to the progress dashes at the bottom so those go to the figure when clicked.

https://user-images.githubusercontent.com/112297/200911655-b82c95d7-15aa-447e-8094-2806627ed876.mov

### DIFF
--- a/src/Utils/Hooks/useDetectActivity.ts
+++ b/src/Utils/Hooks/useDetectActivity.ts
@@ -1,0 +1,49 @@
+import { throttle } from "lodash"
+import { useEffect, useMemo, useRef, useState } from "react"
+
+interface UseDetectActivity {
+  waitTime: number
+}
+
+/**
+ * Detect when user is idle within a component
+ */
+export const useDetectActivity = (
+  { waitTime }: UseDetectActivity = { waitTime: 2500 }
+) => {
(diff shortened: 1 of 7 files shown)


src/Utils/Hooks/useDetectActivity.ts
```diff
@@ -0,0 +1,49 @@
...
+import { useEffect, useMemo, useRef, useState } from "react"
+
+interface UseDetectActivity {
+  waitTime: number
+}
+
+/**
+ * Detect when user is idle within a component
+ */
+export const useDetectActivity = (
```
damassi, line 11: 👍

damassi (approved):

PR #11336, opened 2022-11-10
feat(TX-875): Implement new 'us accounts only' design

AlicanAkyuz:

This PR solves [TX-875]

### Description

This PR implements the new design for how we display "Us bank accounts only" copy for ACH payment method. 

Screenshot 2022-11-10 at 09 55 09

[TX-875]: https://artsyproduct.atlassian.net/browse/TX-875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/PaymentContent.tsx
+++ b/src/Apps/Order/Routes/Payment/PaymentContent.tsx
@@ -15,9 +15,6 @@ import {
   Tooltip,
   Flex,
 } from "@artsy/palette"
-import styled from "styled-components"
-import { themeGet } from "@styled-system/theme-get"
-
 import { Payment_me$data } from "__generated__/Payment_me.graphql"
 import {
   Payment_order$data,
@@ -112,7 +109,6 @@ export const PaymentContent: FC<Props> = props => {
 
       {/* Credit card */}
       <Collapse open={selectedPaymentMethod === "CREDIT_CARD"}>
-        {/* @ts-ignore RELAY_UPGRADE 13  */}
         <CreditCardPickerFragmentContainer
           commitMutation={commitMutation}
           me={me}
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -245,10 +222,10 @@ const getAvailablePaymentMethods = (
           </>
         }
       >
-        <USBankOnlyLabel ml={0.5} variant="xs">
+        <Text ml="24px" variant="xs" color="black60">
```
mdole, line 225: @AlicanAkyuz how would you feel about including this in the `label` (I think we could basically wrap this + the other text in a `Flex`) to avoid having to use a custom `px` margin value that's not in our design system?

rquartararo: @mdole An issue with this is the label has a hover underline that we don't want for description.

rquartararo: Ideally, we would create the option to add an icon to radio labels, but this hasn't been prioritized.

AlicanAkyuz: @mdole yeah that's a good suggestion! we actually thought about this as Rachel pointed out, but it got de-prioritised (it took a lot of back-and-forth discussions to come up with that label in the first place...).

AlicanAkyuz: I'm going to merge this as-is, let's create a ticket on our backlog to do the palette work (if there isn't already one).

mdole: gotcha gotcha. sounds good, thanks!

rquartararo: @AlicanAkyuz this change needs to be added to the design system before it can be added to Palette, so maybe we can create a ticket for design?

rquartararo: @AlicanAkyuz i can do it :)

PR #11339, opened 2022-11-10
fix(images): favors larger on thumbnail components

dzucconi:
Switches to larger for all thumbnail components.


joeyAghion (approved):

dzucconi:
Merging.

PR #11347, opened 2022-11-11
feat(TX-875): Introduce PhoneNumber input to address form

AlicanAkyuz:

This PR solves [TX-875]

### Description

This PR integrate PhoneNumberInput into setting's address form, to be re-used in the checkout flow for adding and editing an address. 

[TX-875]: https://artsyproduct.atlassian.net/browse/TX-875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Settings/Routes/Shipping/Components/SettingsShippingAddressForm.tsx
```diff
@@ -134,10 +141,39 @@ export const SettingsShippingAddressForm: FC<SettingsShippingAddressFormProps> =
...
         handleBlur,
         setFieldValue,
+        setFieldError,
         isValid,
         isSubmitting,
         submitForm,
       }) => {
+        const handlePhoneNumberChange = async (region, number) => {
+          if (region && number && relayEnvironment) {
+            const phoneInformation = await getPhoneNumberInformation(
```
dzucconi, line 151: Can the component just do this for us?

There's no real reason that consumers of the component should have to perform the query.

The phone number input could also handle the debouncing (if it doesn't already).

AlicanAkyuz: yes debouncing and moving the call to PhoneNumberInput are my next steps. 

the only thing that bothers me is how to access `isValid` property in this component (that we get through `getPhoneNumberInformation`).

dzucconi: Just pass it along in as an argument in some custom `onChange` callback?

AlicanAkyuz: I feel like we need `isValid` here to tell Formik it's ok.

AlicanAkyuz: > Just pass it along in as an argument in some custom onChange callback?

will do! thanks for the idea.

dzucconi:
Working thru a related integrity failure: there should be some kind of network indicator IMO (though this will be difficult to do from a design perspective) as well as consideration for what to do if this times out and how we indicate that state to the user. Design problems. Do we let potentially invalid data thru? Etc.

dzucconi:
We also need to debounce the API calls here. Since this is called on every change they are going to lead to poor performance while typing and erroneous validations since it's going to validate half-typed states.

PR #11353, opened 2022-11-14
fix(CX-3151): fix duplication of MyCollection after submission

lordkiz:

This PR solves [CX-3151]

### Description

- Fixes the duplication of MyCollection Artwork after submission. 
- Fixes submission_id not being updated on MyCollection Artwork submission because the source is forced to `web_inbound` instead of `my_collection`

[CX-3151]: https://artsyproduct.atlassian.net/browse/CX-3151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
+++ b/src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
@@ -1,0 +1,35 @@
+import { themeProps } from "@artsy/palette"
+import { useFeatureFlag } from "System/useFeatureFlag"
+import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
+
+function typedArray<T extends string>(...elems: T[]): T[] {
+  return elems
+}
+
+export const submissionFlowSteps = typedArray(
+  "Artwork Details",
+  "Upload Photos",
+  "Contact Information"
+)
+
+export const submissionFlowStepsMobile = typedArray(
+  "Artwork",
+  "Photos",
+  "Contact"
+)
+
+export const useSubmissionFlowSteps = () => {
+  const enableFlowReorder = useFeatureFlag(
+    "reorder-swa-artwork-submission-flow"
+  )
+  const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
+  if (enableFlowReorder && isMobile) {
+    return typedArray("Contact", "Artwork", "Photos")
+  } else if (enableFlowReorder) {
+    return typedArray("Contact Information", "Artwork Details", "Upload Photos")
+  }
+  if (isMobile) {
+    return submissionFlowStepsMobile
+  }
+  return submissionFlowSteps
+}
--- a/src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
+++ b/src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
@@ -2,31 +2,32 @@ import { themeProps } from "@artsy/palette"
 import { useFeatureFlag } from "System/useFeatureFlag"
 import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
 
-function typedArray<T extends string>(...elems: T[]): T[] {
-  return elems
-}
+type MobileSteps = "Artwork" | "Photos" | "Contact"
 
-export const submissionFlowSteps = typedArray(
+type DesktopSteps = "Artwork Details" | "Upload Photos" | "Contact Information"
+
+export const submissionFlowSteps: DesktopSteps[] = [
   "Artwork Details",
   "Upload Photos",
-  "Contact Information"
-)
+  "Contact Information",
+]
 
-export const submissionFlowStepsMobile = typedArray(
+export const submissionFlowStepsMobile: MobileSteps[] = [
   "Artwork",
   "Photos",
-  "Contact"
-)
+  "Contact",
+]
 
-export const useSubmissionFlowSteps = () => {
+export const useSubmissionFlowSteps = (): DesktopSteps[] | MobileSteps[] => {
   const enableFlowReorder = useFeatureFlag(
     "reorder-swa-artwork-submission-flow"
   )
   const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
   if (enableFlowReorder && isMobile) {
-    return typedArray("Contact", "Artwork", "Photos")
-  } else if (enableFlowReorder) {
-    return typedArray("Contact Information", "Artwork Details", "Upload Photos")
+    return ["Contact", "Artwork", "Photos"]
+  }
+  if (enableFlowReorder) {
+    return ["Contact Information", "Artwork Details", "Upload Photos"]
   }
   if (isMobile) {
     return submissionFlowStepsMobile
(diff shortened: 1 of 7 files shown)


src/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/ArtworkDetails.tsx
```diff
@@ -128,11 +129,11 @@ export const ArtworkDetails: React.FC<ArtworkDetailsProps> = ({
         utmSource: utmParams?.utmSource,
         utmTerm: utmParams?.utmTerm,
         sessionID: !isLoggedIn ? getENV("SESSION_ID") : undefined,
+        myCollectionArtworkID: artworkId && isFirstStep ? artworkId : undefined,
```
MounirDhahri, line 132: **nitpick:** it would be great to add a comment here on why this is important. I would even go to add add a test to avoid this regression in the future

src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
```diff
@@ -92,6 +93,9 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
             userPhone: phone.international,
             state: isLastStep ? "SUBMITTED" : "DRAFT",
             sessionID: !isLoggedIn ? getENV("SESSION_ID") : undefined,
+            myCollectionArtworkID:
+              artworkId && isFirstStep ? artworkId : undefined,
+            source: isFirstStep && artworkId ? "MY_COLLECTION" : undefined,
```
MounirDhahri, line 98: **nitpick:** likewise here 🙏

MounirDhahri (dismissed):
Thanks for fixing this Kizito. I left some minor comments that shouldn't be blocking, I would appreciate though if you also link some video as well for similar work in the future.

src/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/ArtworkDetails.tsx
lordkiz: Thanks. True, a test would make sense.

MounirDhahri:
Let's merge this once the tests are fixed. I will try to rerun them again

MounirDhahri (dismissed):
Great job Kizito. Thanks 🙌

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
```diff
@@ -0,0 +1,35 @@
+import { themeProps } from "@artsy/palette"
+import { useFeatureFlag } from "System/useFeatureFlag"
+import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
+
+function typedArray<T extends string>(...elems: T[]): T[] {
+  return elems
+}
```
dzucconi, line 7: What's this all about? It looks like you just want `["Foo", "Bar"] as const`?

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
```diff
@@ -0,0 +1,35 @@
...
+  const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
+  if (enableFlowReorder && isMobile) {
+    return typedArray("Contact", "Artwork", "Photos")
+  } else if (enableFlowReorder) {
+    return typedArray("Contact Information", "Artwork Details", "Upload Photos")
+  }
+  if (isMobile) {
+    return submissionFlowStepsMobile
+  }
+  return submissionFlowSteps
```
dzucconi, line 34: I would flatten out all these conditionals.

dariakoko (approved):

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
lordkiz: Oh, I thought this is as flat as it can be, since I had avoided nesting.

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
lordkiz: Yes, it is the same 👍 . Do you think it is cleaner to just do `["Foo", "Bar"] as const`? With that approach I have to add `as const` to every array I return in  useSubmissionFlowSteps in order to correctly type the `currentStep` prop in `SubmissionStepper`

dzucconi: Yep that's definitely preferred than adding code like this. That's a feature of TypeScript — we should use it.

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
dzucconi: Oh, yes you're right of course. The else if threw me. I have a personal preference to never use else or else if in favor of short-circuiting returns. I would prefer that of course. Though that is preference.

I also want to make sure you realize that the `__internal__useMatchMedia(themeProps.mediaQueries.xs)` is not SSR friendly.

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
dzucconi: Though I definitely feel like I'm missing something here as to why that would be necessary. Either way: I would strongly endeavor to remove something like this.

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
dzucconi: I mean in this case actually the `else if` doesn't actually do anything. Just make it an if

src/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/ArtworkDetails.tsx
```diff
@@ -74,6 +73,7 @@ export const ArtworkDetails: React.FC<ArtworkDetailsProps> = ({
   const initialErrors = validate(initialValue, artworkDetailsValidationSchema)
 
   const artworkId = myCollectionArtwork?.internalID
+  console.log("artworkId", artworkId, "isFirstStep", isFirstStep)
```
dzucconi, line 76: 🔍

lordkiz: 🤦

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
lordkiz: Oh. Yeah true. Will update. 🙌 .
We only use `__internal__useMatchMedia` to determine what labels to put on the Submission steps

src/Apps/Consign/Hooks/useSubmissionFlowSteps.ts
lordkiz: It is not exactly necessary. It just makes typing the `currentStep` prop easier and reads better in VSCode because hovering it tells exactly which String Literal is expected as opposed to showing the type name.
But I will remove it.

PR #11355, opened 2022-11-14
fix(flashbanner): fixes client/server mismatch; clean up

dzucconi:
Closes: [GRO-1410](https://artsyproduct.atlassian.net/browse/GRO-1410)

This must have been causing a client/server mismatch render error since it's been implemented. Oddly though only in Safari did it manifest as the flashing placeholder issue.

### DIFF
--- a/src/Components/FlashBanner/index.tsx
+++ b/src/Components/FlashBanner/index.tsx
@@ -6,11 +6,17 @@ import { graphql } from "react-relay"
 import { useSystemContext } from "System"
 import { EmailConfirmationCTA } from "Components/FlashBanner/EmailConfirmationCTA"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
-import { isServer } from "Server/isServer"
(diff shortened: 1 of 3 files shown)


src/Components/FlashBanner/index.tsx
```diff
@@ -86,8 +86,6 @@ const TrackedFlashBanner = track({
 export const FlashBannerQueryRenderer: React.FC = () => {
   const { relayEnvironment, user } = useSystemContext()
 
-  if (isServer) return null
```
dzucconi, line 89: The problem.

damassi:
It's like winning a prize when these client/server mismatches get resolved. We need to fix ~all of them~ before we can upgrade React.

dzucconi:
cc @dimatretyak related to `enableActivityPanel`

dzucconi:
Acceptance is acting up. Shouldn't affect anything there. Merging.

dimatretyak:
I faced this error on localhost in safari on the `main` branch

image

PR #11344, opened 2022-11-11
fix: Adjust Demand Icon spacing 

olerichter00:
Addresses [CX-3077]

## Description

Adjust Demand Icon spacing.

[CX-3077]: https://artsyproduct.atlassian.net/browse/CX-3077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11358, opened 2022-11-15
fix: Correct demand Icon spacing

olerichter00:
Addresses an issue in [CX-3077](https://artsyproduct.atlassian.net/browse/CX-3077)

## Description

Correct demand Icon spacing ([more context](https://github.com/artsy/force/pull/11344#discussion_r1021939091)).

### DIFF
--- a/src/Components/Artwork/Details.tsx
+++ b/src/Components/Artwork/Details.tsx
@@ -139,8 +139,8 @@ const HighDemandInfo = () => {
   return (
     <Flex flexDirection="row" alignItems="center">
       <HighDemandIcon width={16} height={16} />
-      <Text variant="xs" color="blue100" ml="3px">
-        High Demand
+      <Text variant="xs" color="blue100">
+        &nbsp;High Demand
       </Text>
     </Flex>
   )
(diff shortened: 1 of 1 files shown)


src/Components/Artwork/Details.tsx
```diff
@@ -139,7 +139,7 @@ const HighDemandInfo = () => {
   return (
     <Flex flexDirection="row" alignItems="center">
       <HighDemandIcon width={16} height={16} />
-      <Text variant="xs" color="blue100" ml={0.3}>
+      <Text variant="xs" color="blue100" ml="3px">
```
dzucconi, line 142: I guess but it sounds like you just want a normal `" "` or something

olerichter00: Hmm, placing a `{" "}` behind`<HighDemandIcon width={16} height={16} />` or inside of `Text`  doesn't seem to work 🤔

dzucconi: Inside of `Text` `&nbsp;` would work

dzucconi:
You'll need to rebase with `main` for acceptance to pass.

src/Components/Artwork/Details.tsx
olerichter00: It works 👍

PR #11360, opened 2022-11-15
fix(flashbanner): removes use of window (!)

dzucconi:
Thanks to @dimatretyak for catching this.

Previously there was a `isServer` check that was preventing this code from executing. But _that_ was causing server/client mismatch rendering errors. So when that was fixed, this broke.

😵‍💫

### DIFF
--- a/src/Components/FlashBanner/index.tsx
+++ b/src/Components/FlashBanner/index.tsx
@@ -1,7 +1,6 @@
 import * as DeprecatedAnalyticsSchema from "@artsy/cohesion/dist/DeprecatedSchema"
 import { useMemo } from "react"
 import * as React from "react"
-import qs from "qs"
 import { graphql } from "react-relay"
 import { useSystemContext } from "System"
 import { EmailConfirmationCTA } from "Components/FlashBanner/EmailConfirmationCTA"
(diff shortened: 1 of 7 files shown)


dimatretyak (approved):
LGTM 👍

dzucconi:
OK with ea8b1876ecb290c9ac460a812a31f1d161b237d8 I restructured the component to make more sense and rewrote all the specs. I'll want to revisit this and consider migrating it to just use toasts, which will be simpler and more consistent. There's buttons in the email related banners which looks crazy and there's no feedback on loading states etc. I suspect it'd be a good usecase to flesh out toast actions.

src/Components/FlashBanner/index.tsx
```diff
@@ -1,120 +1,111 @@
...
+      data-test="flashMessage"
+      data-testid="flashMessage"
+    >
       {content}
     </FullBleedBanner>
   )
 }
 
-const TrackedFlashBanner = track({
-  context_module: DeprecatedAnalyticsSchema.ContextModule.FlashBanner,
```
dzucconi, line 83: This didn't do anything

src/Components/FlashBanner/index.tsx
```diff
@@ -1,120 +1,111 @@
...
       render={({ props, error }) => {
         if (error) {
           console.error(error)
-          return null
+          return <FlashBannerFragmentContainer />
         }
 
         if (!props?.me) {
-          return null
+          return <FlashBannerFragmentContainer />
```
dzucconi, line 102: Just because you're logged in doesn't mean we don't want to display a message.

src/Components/FlashBanner/index.tsx
```diff
@@ -1,120 +1,111 @@
...
+    `,
+  }
+)
 
-export const FlashBannerQueryRenderer: React.FC = () => {
-  const { relayEnvironment, user } = useSystemContext()
+export const FlashBannerQueryRenderer: FC = () => {
+  const { user } = useSystemContext()
 
   return user ? (
```
dzucconi, line 86: This ternary is an odd optimization but I'm allowing it.

PR #11364, opened 2022-11-15
chore: Add placeholder content cards GRO-1324

jonallured:
This PR uses our Skeleton components to draw a placeholder for the content cards so that we avoid layout shifting. For the text I used an example of content from a card that's running today. I'll play with this on production to see if it's jarring and we can adjust that content as needed.

<details>
<summary>Here's some screenshots</summary>
Screen Shot 2022-11-15 at 1 49 01 PM
Screen Shot 2022-11-15 at 1 49 08 PM

Screen Shot 2022-11-15 at 1 47 01 PM
Screen Shot 2022-11-15 at 1 47 18 PM
Screen Shot 2022-11-15 at 1 47 23 PM

 
</details>

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Home/Components/HomeContentCards.tsx
+++ b/src/Apps/Home/Components/HomeContentCards.tsx
@@ -5,6 +5,70 @@ import {
   StaticHeroUnit,
 } from "Apps/Home/Components/HomeHeroUnits/HomeHeroUnit"
 import { CaptionedImage as BrazeContentCard } from "@braze/web-sdk"
+import {
+  Box,
+  ResponsiveBox,
+  GridColumns,
+  Column,
+  SkeletonBox,
+  SkeletonText,
+  Spacer,
+} from "@artsy/palette"
+import { Media } from "Utils/Responsive"
+
+const HomeContentCardPlaceholder = () => {
+  return (
+    <GridColumns bg="black5" width="100%">
+      <Column span={6} bg="black30">
+        <>
+          <Media at="xs">
+            <ResponsiveBox aspectWidth={3} aspectHeight={2} maxWidth="100%" />
+          </Media>
+          <Media greaterThan="xs">
+            <Box height={[300, 400, 500]} />
+          </Media>
+        </>
+      </Column>
+      <Column span={6}>
+        <GridColumns height="100%">
+          <Column
+            start={[2, 3]}
+            span={[10, 8]}
+            display="flex"
+            justifyContent="center"
+            flexDirection="column"
+            py={4}
+          >
+            <Media greaterThan="xs">
+              <SkeletonText variant="xs">Artsy Auction</SkeletonText>
+              <Spacer mt={2} />
+            </Media>
+            <SkeletonText variant={["lg-display", "xl", "xl"]}>
+              Post-War and Contemporary
+            </SkeletonText>
+            <Spacer mt={[1, 2]} />
+            <SkeletonText variant={["xs", "sm-display", "lg-display"]}>
+              Bid on works by Alice Neel, Ugo Rondinone, Robert Nava, and
+              more—and benefit the Immediate Abortion Access Fund to support
+              reproductive justice for all.
+            </SkeletonText>
+            <Media greaterThan="xs">
+              <Spacer
+                // Unconventional value here to keep visual rhythm
+                mt="30px"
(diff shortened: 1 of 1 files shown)


src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -5,6 +5,70 @@ import {
...
+  Column,
+  SkeletonBox,
+  SkeletonText,
+  Spacer,
+} from "@artsy/palette"
+import { Media } from "Utils/Responsive"
+
+const FakeCard = () => {
+  return (
+    <GridColumns bg="black5" width="100%">
```
dzucconi, line 21: Why is the whole thing this grey?

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -5,6 +5,70 @@ import {
...
+      <Column span={6} bg="black30">
+        <>
+          <Media at="xs">
+            <ResponsiveBox aspectWidth={3} aspectHeight={2} maxWidth="100%" />
+          </Media>
+          <Media greaterThan="xs">
+            <Box height={[300, 400, 500]} position="relative" />
```
dzucconi, line 28: `position="relative"` not needed

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -5,6 +5,70 @@ import {
...
+  ResponsiveBox,
+  GridColumns,
+  Column,
+  SkeletonBox,
+  SkeletonText,
+  Spacer,
+} from "@artsy/palette"
+import { Media } from "Utils/Responsive"
+
+const FakeCard = () => {
```
dzucconi, line 19: Should name this like `HomeContentCardPlaceholder`

src/Apps/Home/Components/HomeContentCards.tsx
jonallured: That's just copy/paste from the herounit component! What do you think it should be? 🤔

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -5,6 +5,70 @@ import {
...
+              <SkeletonBox mt={2} width={200} height={50} />
+            </Media>
+            <Media at="xs">
+              <Spacer mt={1} />
+              <SkeletonBox mt={2} width={200} height={50} />
+            </Media>
+          </Column>
+        </GridColumns>
+      </Column>
+    </GridColumns>
```
dzucconi, line 69: 👍

src/Apps/Home/Components/HomeContentCards.tsx
jonallured: Also copy/paste but you're right I see no visual difference when I remove it. 👍

src/Apps/Home/Components/HomeContentCards.tsx
dzucconi: Oh, you're right. Somehow missed that 👍

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -51,15 +115,16 @@ export const HomeContentCards: React.FC = () => {
...
-      {cards.map((card, index) => (
-        <ContentCard card={card} key={card.id} index={index} />
-      ))}
-    </HeroCarousel>
-  )
+  const realCards = cards.map((card, index) => (
+    <ContentCard card={card} key={card.id} index={index} />
+  ))
+
+  const heroCards = cards.length < 1 ? placeholderCards : realCards
```
dzucconi, line 127: What we typically do is write a placeholder for the entire component and then wrap it in `Skeleton` which performs that pulsing animation and disables interaction.

jonallured: Ok update, we chatted about this here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1668542811835079?thread_ts=1668525827.325359&cid=C01ADJNCS5D

And ended up bailing on the idea of the Skeleton wrapper because of the wrinkle that this is in the context of a full bleed component and it's not behaving as we'd like. Can come back to this if we choose to!

PR #11372, opened 2022-11-16
Revert "feat: removed the use of `isServer`"

dimatretyak:
Reverts artsy/force#11363


damassi:
@dimatretyak - would you mind updating the PR description with the reason why this needed to get reverted? Then in the future we won't make the same mistake twice!

damassi:
(ty)

PR #11375, opened 2022-11-16
refactor: nav bar indicators

dimatretyak:

Review app: [nav-bar-indicators](https://nav-bar-indicators.artsy.net/)

### Description
It seems to me it would be more correct to consider relay response as the only source of truth (ignoring cookies), since the new activity panel is now available in Eigen also.

If we continue to use a cookie-based approach, we might encounter such a problem:
* User has unread notifications
* User opens Force, it was saved in cookie that the user has unread notifications
* User marked all notifications as read in Eigen a little later
* User opens Force again
* The network request in which we query the unread notifications count is still in progress and has not completed. For this reason, the unread notifications count will be used [from cookie](https://github.com/artsy/force/blob/main/src/Components/NavBar/helpers.ts#L34)
* Since cookies will show that the user has unread notifications, the unread notifications indicator will be displayed (although a little earlier the user marked all notifications as read in Eigen)
* The network request will be completed. the unread notifications count will be `0` and therefore the unread indicator will be hidden


dzucconi:
Love to see this — thank you @dimatretyak.

I would add a follow up ticket that we should likely be periodically polling for changes since the this query will only execute on the initial page load and not any subsequent navigation.

dzucconi:
If I had to guess (I checked the original PR and couldn't find any mention of it) — the original cookie based implementation was to cache it between page loads so that you didn't see it flash off then on again?

dzucconi (dismissed):

dimatretyak:
> we should likely be periodically polling for changes since the this query will only execute on the initial page load and not any subsequent navigation

Agree with it. I will create a Jira ticket

PR #11376, opened 2022-11-16
chore: errors on exhaustive-deps

dzucconi:
Simply ignoring the warnings leads to very difficult to refactor components and hooks. There's very little effort involved with actually addressing the errors. Ignoring them tends to paper over actual bugs that will when the component is extended or altered. If we're to ignore errors then you should be forced to include a comment explaining why you are doing so (there's valid reasons but they tend to be rare in practice.)


damassi (approved):
👌

PR #11383, opened 2022-11-16
Fix: make Zendesk z-index lower than OneTrust's cookie banner

pam-:

This PR solves [GRO-1377]

### Description

This PR does the following things:
* Moves the onboarding styling z-index 🔥 hotfix 🔥 to the `OneTrustConsentStyles` component and sets it to 999 to keep it below the onboarding modal's (9999)
* Sets the Zendesk widget's z-index to 99
  * This was tested locally by mimicking the cookie banner, i.e. making a div with a z-index > 99 and confirming that the widget stays below.

### Next steps:
* Merge upon 👍 ✅ 
* QA on staging 
* Deploy if all looks good

### DIFF
--- a/src/Apps/Components/AppShell.tsx
+++ b/src/Apps/Components/AppShell.tsx
@@ -17,7 +17,6 @@ import { useNavBarHeight } from "Components/NavBar/useNavBarHeight"
 import { useProductionEnvironmentWarning } from "Utils/Hooks/useProductionEnvironmentWarning"
 import { useAuthValidation } from "Utils/Hooks/useAuthValidation"
 import { Z } from "./constants"
-import { createGlobalStyle } from "styled-components"
 import { useDidMount } from "Utils/Hooks/useDidMount"
 import { useOnboardingModal } from "Utils/Hooks/useOnboardingModal"
 import { useImagePerformanceObserver } from "Utils/Hooks/useImagePerformanceObserver"
@@ -111,17 +110,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
 
       {onboardingComponent}
 
-      {isMounted && <OneTrustModalOverlayHotfixStyles />}
+      {isMounted}
     </Flex>
   )
 }
-
-/**
- * This is a workaround for the cookie consent banner overlay appearing on top
- * of our modals. This positions it below so that modal buttons are clickable.
- */
-export const OneTrustModalOverlayHotfixStyles = createGlobalStyle`
-  #onetrust-banner-sdk {
-    z-index: 1 !important;
-  }
-`
--- a/src/Components/OneTrustConsentStyles.ts
+++ b/src/Components/OneTrustConsentStyles.ts
@@ -412,6 +412,7 @@ export const OneTrustConsentStyles = createGlobalStyle`
       ${toStyle({
         boxShadow: DROP_SHADOW,
         borderTop: `1px solid ${THEME.colors.black15}`,
+        zIndex: "1 !important",
       })}
 
       .ot-sdk-container {
--- a/src/Components/OneTrustConsentStyles.ts
+++ b/src/Components/OneTrustConsentStyles.ts
@@ -412,6 +412,7 @@ export const OneTrustConsentStyles = createGlobalStyle`
       ${toStyle({
         boxShadow: DROP_SHADOW,
         borderTop: `1px solid ${THEME.colors.black15}`,
+        // keeping the banner below the onboarding modal
         zIndex: "999 !important",
       })}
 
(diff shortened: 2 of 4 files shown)


src/Apps/Components/AppShell.tsx
```diff
@@ -111,17 +110,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
 
       {onboardingComponent}
 
-      {isMounted && <OneTrustModalOverlayHotfixStyles />}
```
dzucconi, line 114: Whoa, I did not know this was here

src/Apps/Components/AppShell.tsx
```diff
@@ -111,17 +110,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
...
+      {isMounted}
```
dzucconi, line 113: You can get rid of this `isMounted` and the related hook call.

src/Components/OneTrustConsentStyles.ts
```diff
@@ -412,6 +412,8 @@ export const OneTrustConsentStyles = createGlobalStyle`
       ${toStyle({
         boxShadow: DROP_SHADOW,
         borderTop: `1px solid ${THEME.colors.black15}`,
+        // keeping the banner below the onboarding modal
+        zIndex: "999 !important",
```
dzucconi, line 416: So: this `toStyle` function automatically `!important`s everything in it. Also: we _want_ to keep this below the modal? It seems to me that this should be above everything.

dzucconi: Also, you probably know this but you're next steps didn't list it. Changing this file doesn't actually change the OneTrust stuff. The remaining steps need to be followed:

```
* - in Boot.tsx:
 * `import { OneTrustConsentStyles } from "Components/OneTrustConsentStyles"`
 * and include the `<OneTrustConsentStyles />` component in the render method.
 *
 * When you're ready, you can just return only the `<OneTrustConsentStyles />`
 * component from Boot.tsx, then copy out the compiled styles. Log into
 * onetrust.com and find the template. Click on "Styling", then paste into the
 * "Custom CSS" field. Save the template, then go to "Integration" > "Scripts".
 * Click the environment and publish.
```

src/Components/ZendeskWrapper.tsx
```diff
@@ -21,6 +21,12 @@ export const ZendeskWrapper: FC<ZendeskWrapperProps> = ({
     src: `https://static.zdassets.com/ekr/snippet.js?key=${key}`,
     removeOnUnmount: true,
     onReady: () => {
+      window.zESettings = {
+        webWidget: {
+          // keeping the widget below the cookie banner
+          zIndex: 99,
```
dzucconi, line 27: I'd set this to `1`

src/Components/OneTrustConsentStyles.ts
dzucconi: I suspect a much lower value is sufficient. We avoid a complex z-index hierarchy, generally speaking. You'll notice that we only use the values `0`, `1` and very rarely: `2`. Things tend to have a natural stacking order and the Modal is a legacy oddity.

dzucconi: Though now that I remember it, I did extract a Z index object for some of the weird ones. Check `Apps/Components/constants` and extract these into those.

src/Components/ZendeskWrapper.tsx
dzucconi: I _do_ wish we were able to trigger this with our own UI. Did you find out if that was possible?

src/Components/OneTrustConsentStyles.ts
pam-: roger! i was following the modal's 9999 -- will mod

pam-: (will add the ontrust stuff to the PR desc too)

src/Components/ZendeskWrapper.tsx
pam-: Yep! Afaik We can change the basic look of the button (image, text, color) 
I think that should be a separate ticket, though!

src/Components/OneTrustConsentStyles.ts
pam-: looking at z-indices from `constants`: the footer is 1, so the consent banner should be 1, and the zendesk widget 0. will push in a bit

src/Components/ZendeskWrapper.tsx
dzucconi: Oh, I'm looking for more than that. Like I want us to have our own component trigger it; rather than pass it styles. But sure.

dzucconi (approved):

PR #11384, opened 2022-11-16
feat(GRO-1437): rebuilds auth dialog

dzucconi:
I'm just going to open a draft here to update progress.

This is what's left on my TODO:

- [x] Support passing title
- [x] Support redirectTo
- [x] Support [images](https://drive.google.com/drive/u/2/search?q=2x_Evergreen)
- [x] Support intent
- [x] Handle redirects
- [x] Wire up social auth
- [x] showRecaptchaDisclaimer
- [x] Set post login/signup cookies (afterSignUpAction)
- [x] Ensure state reset on close?
- [x] Pop up a toast that you're already authenticated if attempting to open while logged in
- [ ] Support usage within SSR auth routes
- [ ] Investigate `autoFocus` behavior within Modals
- [ ] Double check for any existing data-test attrs
- [ ] Add specs
- [ ] Support contextModule
- [ ] Wire up analytics tracking
- [ ] Extract copy to en file
- [ ] Ensure terms needs to be checked when using social authentication
- [ ] Pass Integrity checks
- [ ] Clean up error messages
- [ ] Validate that afterAuthActions still execute correctly when they are serialized in the query params during Gravity auth

-------

https://user-images.githubusercontent.com/112297/202306804-2693b7a0-acf1-4c3b-bef2-19ce4ae4650d.mov

### DIFF
--- a/src/Apps/Debug/DebugAuth.tsx
+++ b/src/Apps/Debug/DebugAuth.tsx
@@ -1,0 +1,140 @@
+import { AuthIntent, Intent } from "@artsy/cohesion"
+import {
+  Button,
+  Checkbox,
+  Input,
+  Join,
+  Select,
+  Spacer,
+  Text,
+} from "@artsy/palette"
+import {
+  AuthDialogMode,
+  AUTH_DIALOG_MODES,
+} from "Components/AuthDialog/AuthDialogContext"
+import { useAuthDialog } from "Components/AuthDialog"
+import { FC, useState } from "react"
+import { Title } from "react-head"
+
+export const DebugAuth: FC = () => {
+  const { showAuthDialog } = useAuthDialog()
+
+  const [state, setState] = useState<
+    Required<Parameters<typeof showAuthDialog>[0]>
+  >({
+    mode: "Login",
+    options: {},
+  })
+
+  return (
+    <>
+      <Title>AuthDialog</Title>
+
+      <Spacer y={4} />
+
+      <Join separator={<Spacer y={4} />}>
+        <Button
+          variant="primaryBlue"
+          onClick={() => {
+            showAuthDialog(state)
+          }}
+        >
+          Open AuthDialog with options
+        </Button>
+
+        <Text as="pre" variant="xs" bg="black5" py={0.5} px={1}>
+          {JSON.stringify(state, null, 2)}
+        </Text>
+
+        <Select
+          title="View"
+          selected={state.mode}
+          options={AUTH_DIALOG_MODES.map(mode => ({
+            value: mode,
+            text: mode,
+          }))}
+          onSelect={(mode: AuthDialogMode) => {
+            setState(prevState => ({ ...prevState, mode }))
+          }}
+        />
+
+        <Input
+          title="Title"
+          value={state.options.title}
+          placeholder="Copy to display as title"
+          onChange={({ target: { value: title } }) => {
+            setState(prevState => ({
+              ...prevState,
+              options: { ...prevState.options, title },
+            }))
+          }}
+        />
+
+        <Input
+          title="Redirect to"
+          value={state.options.redirectTo}
+          placeholder="Pathname to redirect to after authentication"
+          onChange={({ target: { value: redirectTo } }) => {
+            setState(prevState => ({
+              ...prevState,
+              options: { ...prevState.options, redirectTo },
+            }))
+          }}
+        />
+
+        <Select
+          title="Intent"
+          value={state.options.intent}
+          options={AUTH_INTENTS.map(intent => ({
+            value: intent,
+            text: intent,
+          }))}
+          onSelect={(intent: AuthIntent) => {
+            setState(prevState => ({
+              ...prevState,
+              options: { ...prevState.options, intent },
+            }))
+          }}
+        />
+
+        <Checkbox
+          selected={state.options.image}
+          onSelect={image =>
+            setState(prevState => ({
+              ...prevState,
+              options: { ...prevState.options, image },
+            }))
+          }
+        >
+          Display image panel?
+        </Checkbox>
+      </Join>
+    </>
+  )
+}
+
+const AUTH_INTENTS: AuthIntent[] = [
+  Intent.bid,
+  Intent.buyNow,
+  Intent.consign,
+  Intent.createAlert,
+  Intent.followArtist,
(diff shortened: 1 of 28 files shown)


src/Components/AuthDialog/AuthDialogContext.tsx
```diff
@@ -0,0 +1,73 @@
...
+    dispatch({ type: "SHOW" })
+  }
+
+  return (
+    <AuthDialogContext.Provider
+      value={{ state, dispatch, showAuthDialog, hideAuthDialog }}
+    >
+      {children}
+
+      {state.isVisible && <AuthDialog onClose={hideAuthDialog} />}
```
dzucconi, line 66: So I'm designing the usage API to be such that, there's a single instance of this modal that we show/hide. You import and use `const { showAuthDialog } = useAuthDialog()` and then can use `showAuthDialog({ ... })` with options for how it's displayed.

src/Components/AuthDialog/Views/AuthDialogLogin.tsx
```diff
@@ -0,0 +1,198 @@
...
+  // If auth code then require it
+  authenticationCode: Yup.string()
+    .when("mode", {
+      is: "OnDemand",
+      then: Yup.string().required("Authentication code required."),
+    })
+    .when("mode", {
+      is: "TwoFactor",
+      then: Yup.string().required("Authentication code required."),
+    }),
```
dzucconi, line 197: We previously weren't validating this stuff.

dzucconi: I'm also wondering if I should validate the format. @joeyAghion can we expect this to always be a numeric sequence?

src/Utils/recaptcha.ts
```diff
@@ -1,30 +1,40 @@
...
 const logger = createLogger("recaptcha.ts")
 
-export const recaptcha = (action: RecaptchaAction, cb?: any) => {
-  if (sd.RECAPTCHA_KEY) {
+// TODO: Should return a Promise instead of accepting a callback
+export const recaptcha = (
+  action: RecaptchaAction,
+  callback?: RecaptchaCallback
+) => {
+  if (getENV("RECAPTCHA_KEY")) {
```
damassi, line 11: 👍 (though we might want to update citadel with this so folks always have it)

damassi (commented):
✨

src/Utils/recaptcha.ts
dzucconi: I didn't add this, just changed it to use `getENV`. Should be there already.

src/Components/AuthDialog/Views/AuthDialogLogin.tsx
joeyAghion: Nope! One-time codes are 6-digit numerics, but backup codes are longer and alphanumeric. You could do some minimal validation of minimum length though.

dzucconi:
Documenting the existing modal options which are supported in the current auth flow:

🟢 if we will continue to support
🔴 if we will drop support
🟡 if I'm not sure yet

-------

* `afterSignUpAction`: 🟢 This controls the after auth intent. A better name for this would be `afterAuthAction` since it has nothing to do with sign ups specifically.
* `contextModule`: 🟢 Gets passed into the analytics tracking
* `copy`: 🟢 refers to the modal title, really
* `destination`: 🔴  This sets a cookie with a location to redirect to. Nothing appears to check for it. The only thing that was using it was the modals opened on the article app. They are just calling for the redirect to just be the page you're already on, which is going to happen by default, which is why this appears to work.
* `image`: 🟢 this shows up as the image which occupies a left pane on the modal dialog
* `intent`: 🔴 this is a shortcut for constructing a bunch of options. It gets passed into the views even though it doesn't do anything there. It's only referenced in [openAuthModal](https://github.com/artsy/force/blob/aa3bebac4c496a1d746cfb1ec86adce0f3d909c1/src/Utils/openAuthModal.ts). IMO this is just confusing. The better way to do this would be to write a wrapper function that sits above `openAuthModal` rather than this being an option.
* `mode`: Essentially refers to which view you want to initially open. `Login | SignUp | ForgotPassword`
* `oauthLogin`: 🔴 This appears to only be used on the server side routes so it's not really an option that needs to be passed. It flags the server-side route to redirect to something if you're logged in. I'm not 100% sure why we need to check for this because if you're logged in and there's a redirect-to we'd follow it anyway. cc @damassi 
* `redirectTo`: 🟢 Which page to redirect to after authentication is successful
* `signupReferer`: 🟡 This is relevant to the server-side routes. It gets set using a sharify var named `AUTHENTICATION_REFERER` which is populated with `req.header("Referer") || req.hostname` — ultimately this gets passed into the sign up form as a field. We want to support this still but I don't think this needs to be threaded through as a modal option explicitly and can just be pulled out using `getENV`.
* `triggerSeconds`: 🔴 This used to be used to delay opening the the modal while you scrolled. No longer needed.

joeyAghion:
`oauthLogin` was added when we switched from api.artsy.net to www.artsy.net serving the form for logging into other oauth clients. It was necessary because Force's previous behavior was to redirect away from the log-in form when landing on it as an already-authenticated user. With this parameter, re-authenticating and redirecting back to the next oauth step is supported.

dzucconi:
I'm going to merge once green. This just enables a testbed under the `/debug/auth` path. All the functionality should be there but we're missing analytics and the SSR route integrations.

src/Apps/Debug/DebugAuth.tsx
```diff
@@ -0,0 +1,140 @@
...
+  Intent.registerToBid,
+  Intent.requestConditionReport,
+  Intent.saveArtwork,
+  Intent.signup,
+  Intent.viewAuctionResults,
+  Intent.viewArtist,
+  Intent.viewEditorial,
+  Intent.viewFair,
+  Intent.viewViewingRoom,
+]
```
damassi, line 140: This is cool -- so its an auth form builder of some kind?

dzucconi: Yeah basically: https://staging.artsy.net/debug/auth — I just wanted a way for us to test this in all it's configurations.

damassi: (Might be worth keeping around)

PR #11385, opened 2022-11-17
feat: Debounce phone number validation call

AlicanAkyuz:

### Description

This PR introduces debouncing to the phone number validation.


dzucconi (dismissed):
🙏

dzucconi:
So actually: this seems a bit weird to me. Wouldn't want to debounce `validatePhoneNumber`? Can you just confirm that Relay isn't making extraneous requests?

PR #11389, opened 2022-11-17
chore(CX-3066): update My Collection insights tracking events

MrSltun:

This PR solves [CX-3066]

### Description
This is a follow up on artsy/cohesion#365

This PR updates the addCollectedArtwork tracking event to include My Collection insights as the context owner type, and also updates the auction results item toggle tracking event

[CX-3066]: https://artsyproduct.atlassian.net/browse/CX-3066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -42,14 +43,16 @@ export const ArtistAuctionResultItem: FC<Props> = props => {
...
-      context_page: DeprecatedAnalyticsSchema.PageName.ArtistAuctionResults,
-      action_type:
-        DeprecatedAnalyticsSchema.ActionType.AuctionResultItemClicked,
-      current: {
-        expanded: expand,
-      },
-    })
+    if (pathname.startsWith("/my-collection/artwork")) {
+      trackClickedAuctionResultItem(
+        expand,
```
MounirDhahri, line 48: **nitpick:** this naming feels a bit off to me, would you mind updating it to maybe `collapsed` or `closed` since `expand=!expanded` or maybe even better, use it just as it is

MounirDhahri (approved):
LGTM, thanks Sultan! only a small comment and please feel free to merge afterwards

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
MrSltun: The idea here is to track the event of clicking since the element gets rendered collapsed, we set the initial state `expanded` to be false, and when the user interacts with the element, we change the state and track that event.

The naming `expanded` seems clear due to the action that happens when interacting with the element, but if we change it to `collapsed` it would insinuate that the element is open by default (the initial value needs to be false) unless we refactor the code to change the name

dariakoko (approved):

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
MounirDhahri: thanks for the clarification. I understand the reasoning behind it better now, but I still though feel though that we can do better with the naming, maybe something like `toBeExpanded` or so. This is not a blocking comment though, it's more of a nitpick

MounirDhahri (approved):

PR #11397, opened 2022-11-17
chore: return singular or plural form for "Lot/Lots" label

dimatretyak:

This PR solves [FX-4445]

#### One lot
https://lot-or-lots-label.artsy.net/auction/dima-test-auction-1

demo-1

#### Multiple lots
https://lot-or-lots-label.artsy.net/auction/dima-test-auction-2

demo-2

[FX-4445]: https://artsyproduct.atlassian.net/browse/FX-4445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Auction/Components/AuctionDetails/AuctionDetails.tsx
```diff
@@ -33,6 +33,16 @@ const AuctionDetails: React.FC<AuctionDetailsProps> = ({ sale, me }) => {
...
     !!sale.cascadingEndTimeIntervalMinutes && !sale.isClosed
 
+  const getEndTimeIntervalMessage = () => {
+    const { cascadingEndTimeIntervalMinutes } = sale
+
+    if (sale.artworksConnection?.totalCount === 1) {
+      return `Lot close at ${cascadingEndTimeIntervalMinutes}-minute intervals`
+    }
+
+    return `Lots close at ${cascadingEndTimeIntervalMinutes}-minute intervals`
```
anandaroop, line 43: **thought (non-blocking):** Learned from [the recent i18n presentation](https://artsy.slack.com/archives/C02BC3HEJ/p1668697223204379?thread_ts=1668523213.247079&cid=C02BC3HEJ) that we have access to [pluralization helpers via react-i18next](https://react.i18next.com/misc/using-with-icu-format#plural).

src/Apps/Auction/Components/AuctionDetails/__tests__/AuctionDetails.jest.tsx
```diff
@@ -85,4 +85,36 @@ describe("AuctionDetails", () => {
...
+        Sale: () => ({
+          isClosed: false,
+          cascadingEndTimeIntervalMinutes: 5,
+          artworksConnection: {
+            totalCount: 1,
+          },
+        }),
+      })
+
+      const message = "Lot close at 5-minute intervals"
```
anandaroop, line 101: **thought (blocking):** This doesn't seem right to me. If there is a single lot, there are no intervals between lots, so this message doesn't really make sense. Maybe it makes more sense to suppress this message altogether in the single-lot case?

Looking at the screenshot in the description, I'd say this problem applies to all three messages — maybe we can clarify what the correct copy should be, or if there should be any message at all in the single-lot case? 

"X days until lot closes" (the one in blue) might be the only one that really works?

anandaroop (changes requested):
Thanks for tackling this and the upstream MP work 👏🏽 

I think there's still work to do to clarify the expected text and behavior, see below...

dimatretyak:
Closed based on [this comment](https://artsy.slack.com/archives/C9SATFLUU/p1669044416854419?thread_ts=1669043681.173219&cid=C9SATFLUU)

PR #11398, opened 2022-11-17
feat: Send content card events to braze GRO-1324

jonallured:
This PR updates the Braze content card code to use their SDK to send events. Because we're loading these things ourselves, we are responsible for sending clicks and impressions so that's what this PR does.

Note that I started by duplicating the `HomeHeroUnit` and then spent about a dozen commits pruning it down to just what I needed in this case. Ultimately we'll end up removing that component once we fully ship Braze content cards but not yet. We still use a query param to load the Braze ones so this acts as a feature flag for now. Hopefully we can remove all that stuff next week!

While this project was meant to focus only on replicating how the existing cards behave we had an opportunity to chat about the images and improve things there - more here:

https://artsy.slack.com/archives/C01ADJNCS5D/p1668700831526749?thread_ts=1668624990.972969&cid=C01ADJNCS5D

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Home/Components/HomeContentCard.tsx
+++ b/src/Apps/Home/Components/HomeContentCard.tsx
@@ -17,7 +17,6 @@ import { HomeHeroUnitCredit } from "./HomeHeroUnits/HomeHeroUnitCredit"
 
 interface ContentCard {
   creditLine?: string
-  heading?: string | JSX.Element
 }
 
 interface HomeContentCardProps {
@@ -33,7 +32,6 @@ export const HomeContentCard: React.FC<HomeContentCardProps> = ({
 
   const contentCard: ContentCard = {
     creditLine: extras.credit,
-    heading: extras.label,
   }
 
   const image = cropped(card.imageUrl!, {
@@ -129,10 +127,10 @@ export const HomeContentCard: React.FC<HomeContentCardProps> = ({
               style={{ display: "block", textDecoration: "none" }}
             >
               <Media greaterThan="xs">
-                {contentCard.heading && (
+                {extras.label && (
                   <>
                     <Text variant="xs" color="black100">
-                      {contentCard.heading}
+                      {extras.label}
                     </Text>
 
                     <Spacer mt={2} />
--- a/src/Apps/Home/Components/HomeContentCard.tsx
+++ b/src/Apps/Home/Components/HomeContentCard.tsx
@@ -44,7 +44,6 @@ export const HomeHeroUnit: React.FC<HomeHeroUnitProps> = ({
   heroUnit,
   index,
 }) => {
-  const layout = "a"
   const { trackEvent } = useTracking()
 
   const handleTrackEvent = useCallback(() => {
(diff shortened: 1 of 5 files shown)


src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -121,10 +97,18 @@ export const HomeContentCards: React.FC = () => {
...
 
   const realCards = cards.map((card, index) => (
-    <ContentCard card={card} key={card.id} index={index} />
+    <HomeContentCard card={card} key={card.id} index={index} />
   ))
 
   const heroCards = cards.length < 1 ? placeholderCards : realCards
 
-  return <HeroCarousel>{heroCards}</HeroCarousel>
+  const handleChange = newIndex => {
```
dzucconi, line 105: What about this index is "new"?

src/Components/HeroCarousel/HeroCarousel.tsx
```diff
@@ -7,20 +7,24 @@ export interface HeroCarouselProps {
   children: ReactNode
   /** Only utilizes full-bleed at desktop sizes */
   fullBleed?: boolean
+  onChange?: (newIndex) => void
```
dzucconi, line 10: Same q: re: "new"

dzucconi (approved):
Looks good. I would use a consistent name for the indexes though.

src/Apps/Home/Components/HomeContentCard.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const image = cropped(card.imageUrl!, {
+    width: 1270,
+    height: 500,
+  })
+
+  const cardLink = card.url ?? ""
+
+  const handleClick = () => {
+    const appboy = (window as any).appboy
```
damassi, line 37: Can we update the Window type to include appboy?

damassi: And should we alias the variable name to `const braze` so that folks don't get confused in the future?

src/Apps/Home/Components/HomeContentCards.tsx
```diff
@@ -121,10 +97,18 @@ export const HomeContentCards: React.FC = () => {
...
+    const appboy = (window as any).appboy
```
damassi, line 106: ditto here

src/Apps/Home/Components/HomeContentCard.tsx
jonallured: What if I promise to return to this once we've finished upgrading our segment integration?? We're on an older version of the web sdk and the next major release from them switches to `window.braze` so then I could do all that stuff and it'll finally make sense. 😝 

I'm hoping to get that migration done any week now if I can just get 3 vendors to play nice (OneTrust, Segment, Braze). 😅

src/Apps/Home/Components/HomeContentCards.tsx
jonallured: Yeah idk, I guess I was thinking "new" in terms of it changing from the previous value but that's silly - will update!

src/Apps/Home/Components/HomeContentCard.tsx
damassi: Oh yeah np!

jonallured:
Ok gonna merge and kick the tires. 👍

PR #11412, opened 2022-11-18
fix(useauthintent): unifies auth intent approach

dzucconi:
Closes: [GRO-1448](https://artsyproduct.atlassian.net/browse/GRO-1448)

I'll just convert this off draft on Monday but this should be fine. I need to follow a submission end to end on staging as well to test that end of it.

Part of the effort in rebuilding our authentication modal is just decoding what the old one was doing. So this is the first of a few PRs I'll make to simplify things and remove options and props, as I discover more.

-------

I've removed the usage of `mediator` from the post-authentication action intent code. IMO we should just keep a React way of doing things here. Instead of triggering and subscribing to an event bus, just lift the state up. So that's what this does: creates an AuthIntent context and sets a single value in it. When you've observed the change in value you just clear it using the provided function.

I've also moved the associate submission mutation from it's own hook into this one, which is what this should have been doing. We've got a single approach to handling this stuff: if we need to perform some action on behalf of the user after authentication than let's use that one approach instead of creating more solutions to the same problem. I do realize all of the authentication code is nearly impossible to follow, so that's why it's getting cleaned up and rebuilt.


src/Apps/Article/ArticleApp.tsx
```diff
@@ -34,13 +34,11 @@ const ArticleApp: FC<ArticleAppProps> = ({ article }) => {
       // TODO: Onboarding is triggered based on contents of redirectTo
       // prop. Move this to `afterSignupAction.action`
       redirectTo: location.pathname,
-      afterSignUpAction: { action: "editorialSignup" },
```
dzucconi, line 37: These literally didn't do anything.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
```diff
@@ -54,23 +56,27 @@ const ArtworkActionsSaveButton: React.FC<ArtworkActionsSaveButtonProps> = ({
...
     setPopoverVisible(false)
   }, [])
 
+  const { value, clearValue } = useAuthIntent()
+
   useEffect(() => {
-    mediator.on("artwork:save", openAuctionRegistrationPopover)
+    // If we are coming in after authentication triggered by a save
+    // *maybe* open the auction registration popover
+    if (!value || value.action !== "save") return
```
dzucconi, line 74: So you can see here: we wait for the value to change to something we're looking for...

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
```diff
@@ -54,23 +56,27 @@ const ArtworkActionsSaveButton: React.FC<ArtworkActionsSaveButtonProps> = ({
...
+
+    maybeOpenAuctionRegistrationPopover()
 
-    return () => {
-      mediator.off("artwork:save")
-    }
-  }, [openAuctionRegistrationPopover])
+    clearValue()
```
dzucconi, line 78: Then clear it when we're done.

src/Apps/Authentication/Utils/useAuthForm.ts
```diff
@@ -21,36 +21,32 @@ export function useAuthForm({
...
-        "afterSignUpAction",
-        JSON.stringify({ action, kind, objectId })
-      )
-    }
-  }
+  // FIXME: Convection should link using the `afterSignUpAction` param,
+  // not `submissionId` so that we don't have to do this.
+  //
+  // If there's a `submissionId` constructs an `afterSignUpAction` from it. Otherwise
+  // falls back to the `afterSignUpAction` param.
```
dzucconi, line 42: I'll add a ticket for this but we should use the `afterSignUpAction` (poorly named but, one thing at a time), query param. Which will get handled by our `useAuthIntent` hook.


Would look something like:
`?afterSignUpAction%5Baction%5D=follow&afterSignUpAction%5Bkind%5D=artist&afterSignUpAction%5BobjectId%5D=andy-warhol`

src/Components/Authentication/Types.ts
```diff
@@ -61,11 +56,6 @@ export interface ModalOptions {
    * Hook to be called after the modal has closed
    */
   afterClose?: () => void
-  /**
-   * MOBILE ONLY
-   * Used to construct afterSignupAction from query params
-   */
-  action?: AfterSignUpAction["action"]
```
dzucconi, line 68: This was previously used by martsy to communicate with this 'new' component. I do not know why it wasn't nested in the first place. It no longer does anything.

dzucconi: Ditto for the rest of these `MOBILE ONLY` props

src/System/Router/Boot.tsx
```diff
@@ -87,9 +88,11 @@ export const Boot = track(undefined, {
...
                     <ToastsProvider>
                       <StickyProvider>
-                        <FocusVisible />
-                        <SiftContainer />
-                        {children}
+                        <AuthIntentProvider>
+                          <FocusVisible />
+                          <SiftContainer />
+                          {children}
+                        </AuthIntentProvider>
```
dzucconi, line 95: I do kinda wanna write a little compose providers function to clean this up.

src/Utils/Hooks/useAuthIntent/__tests__/useAuthIntent.jest.ts
```diff
@@ -244,27 +280,96 @@ describe("runAuthIntent", () => {
...
+          action: "createAlert",
+          kind: "artist",
+          objectId: "banksy",
+        })
+
+        expect(Cookies.expire).toBeCalledWith("afterSignUpAction")
+      })
+    })
+
+    // There is no failure branch for this action
```
dzucconi, line 311: TBH, I think we could have allowed the UI to be opened and then had the create button pop open the auth modal with the appropriate intent. I wonder if we should send out toasts for success though.

src/System/Router/Boot.tsx
damassi: If you can pull it off i'll send ya $5. We've all tried...

src/Utils/Hooks/useAuthIntent/mutations/AuthIntentAssociateSubmissionMutation.ts
```diff
@@ -0,0 +1,33 @@
+import { commitMutation, Environment, graphql } from "relay-runtime"
+import { AuthIntentAssociateSubmissionMutation } from "__generated__/AuthIntentAssociateSubmissionMutation.graphql"
+import { AuthIntentMutation } from "./types"
+
+export const associateSubmissionMutation: AuthIntentMutation = (
```
damassi, line 5: this might just be a refactor, but why not use our `useMutation` hook?

dzucconi: Yeah I left a TODO about it; just didn't feel like going to crazy here. The specs, etc.

damassi: yah np at all

dzucconi:
And just deleting some options that aren't even used in cf307afc647203ac0b0e20f4bcad724e49e78054

src/Apps/Authentication/Utils/helpers.ts
```diff
@@ -176,23 +177,17 @@ export const maybeUpdateRedirectTo = (
...
     Cookies.set("destination", destination, {
       expires: 60 * 60 * 24,
     })
   }
-
-  if (submissionId) {
-    Cookies.set("submissionId", submissionId, {
-      expires: 60 * 60 * 24,
-    })
-  }
```
AlicanAkyuz, line 195: the reason we have `submissionId` stuff here is to connect an anonymous user who signs up to artsy to their submission. the scenario is: 

- user makes a submission anonymously
- user gets an email encouraging to sign up to Artsy to see their submission on MyCollection (the email includes a link with `submissionId` in the params)
- we set `submissionId` cookie 
- after they sign up, we check if there's a `submissionId` cookie 
- reach out to Convection to add the user to the submission

AlicanAkyuz: I'm not entirely sure if this is still the case since CX took over this flow for a while ago. maybe @MounirDhahri is aware of any changes.

src/Utils/Hooks/useAuthIntent/index.tsx
```diff
@@ -65,35 +75,66 @@ export const runAuthIntent = async (
...
           }
+
           break
+
         case "save":
-          triggerEvent(mediator, "artwork:save", value)
           return saveArtworkMutation(relayEnvironment, value.objectId)
+
+        case "associateSubmission":
+          return associateSubmissionMutation(relayEnvironment, value.objectId)
```
AlicanAkyuz, line 85: ok never mind my comment above, I just saw this 👍

dimatretyak (dismissed):
Looks great. Thank you for these changes 👌

src/Apps/Authentication/Utils/helpers.ts
MounirDhahri: yes, that's still the case

src/Utils/Hooks/useAuthIntent/index.tsx
```diff
@@ -32,30 +31,41 @@ const isValid = (value: any): value is AfterAuthAction => {
...
-  user: User,
+export const runAuthIntent = async ({
+  user,
+  relayEnvironment,
+  onSuccess,
+}: {
+  user: User
   relayEnvironment: RelayModernEnvironment
-) => {
+  onSuccess: (value: AfterAuthAction) => void
```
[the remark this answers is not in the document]
dzucconi: This only get's used in a single place where it *is* required `useRunAuthIntent` which gets run on app mount. So this isn't something for other devs to consume really.

src/Utils/Hooks/useAuthIntent/index.tsx
dzucconi: I was unable to test this end to end because the form is unable to validate sometimes. Going to assume these mutations work.

src/Apps/Authentication/Utils/useAuthForm.ts
```diff
@@ -12,45 +17,49 @@ interface UseAuthFormProps {
...
 
-    if (isClient) {
-      Cookies.set(
-        "afterSignUpAction",
-        JSON.stringify({ action, kind, objectId })
-      )
-    }
-  }
+  // FIXME: We have to make the cookie now because apparently passing it into
+  // options like this doesn't even work!
```
dzucconi, line 57: I will revisit all of this after the rebuild is done and I need to integrate the new form into the static routes, where this is an issue.

PR #11418, opened 2022-11-21
[WIP] feat: show attribution class name when media xs, display about info li…

dariakoko:
…ke in eigen

This PR solves [CX-2953]

### Description

- display attribution class name on mWeb
- display artwork about info on mWeb like in Eigen

| Before | After |
| --- | --- |
| image | image |

[CX-2953]: https://artsyproduct.atlassian.net/browse/CX-2953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi:
I suspect most of this can be handled without Fresnel and would encourage you to see if it's really necessary (maybe!).

MounirDhahri:
@dzucconi could you maybe share more details on what you mean?

src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkSidebar/MyCollectionArtworkSidebarMetadata.tsx
```diff
@@ -70,17 +77,37 @@ const MetadataField = ({ label, value }) => {
...
+        </Box>
+      </Media>
+      <Media at="xs">
+        <Flex flexDirection="row" justifyContent="space-between" mb={1}>
+          <Text variant="sm" color="black100" pr={1}>
+            {label}
+          </Text>
+
+          <Text
+            style={{ flex: 1, maxWidth: "70%" }}
```
dzucconi, line 101: Both of these are normal props that we support, use them.

dzucconi:
@MounirDhahri it's really difficult to review without checking out the branch to understand where its usage is actually necessary. I just know from experience: I very rarely need it, and so when I see it interspersed into simple looking components I treat it as a code smell.

There's a penalty we do incur with it that we should be aware of: we're rendering twice the amount of UI on the server. It also makes code very difficult to read when it's interwoven in a lot of places.

Generally, I'll use it in a wrapper component to switch on two totally different approaches to a UI. [For example](https://github.com/artsy/force/blob/main/src/Components/HeroCarousel/HeroCarousel.tsx).

And, while though it's totally fine to use — I don't want to scare anyone off here — I do want us to be aware of what it is we're doing and what the alternatives are. Very frequently using responsive `display`, `flexDirection`, `order`, etc, is more than adequate, and easier to read and maintain.

MounirDhahri:
@dzucconi thanks a lot for the feedback, I wasn't aware of the drawbacks of using Fresnel, I will share this with the rest of the CX team to raise more awareness about it. cc @artsy/collector-experience

dzucconi:
Don't bend over backwards to avoid it; it'll probably be clear to the developer when its usage is necessary. It's just typically not clear to the reviewer, which makes it tricky! Just generally be aware that we've [got responsive props](https://styled-system.com/responsive-styles/) and that they are lighter weight, easier to use, and clearer.

MounirDhahri:
@dariakoko let me know if there is something not clear here and would gladly support with it

dariakoko:
Thanks for your comments! I'll review my code tomorrow. 🙏

damassi:
@dariakoko - Going to close this as it appears to be stale. Feel free to reopen if you still need it!

PR #11422, opened 2022-11-22
refactor(TX-876): consolidate address forms in checkout flow

rquartararo:

This PR solves [TX-876]

### Description

This PR aims to consolidate the address forms in Force by replacing the add and edit `AddressModal` in the checkout flow with the `SettingsShippingAddressForm` (with phone number validation). Replacing the main checkout address form will happen in a following PR. 

**Checklist**

- [x] Edit Address
- [x] Edit Default Address
- [x] Create New Address
- [ ] Delete Address
- [ ] Fix tests


src/Apps/Order/Utils/shippingUtils.ts
```diff
@@ -128,15 +127,22 @@ export const convertShippingAddressForExchange = (
 }
 
 export const convertShippingAddressToMutationInput = (
-  address: SavedAddressType
-): UserAddressAttributes => {
-  return omit(
-    {
-      ...address,
-      name: address?.name || "",
+  address: SavedAddressType | null
```
AlicanAkyuz, line 130: we could consider importing `SettingsShippingAddress_address$data `and adding it to potential types here `address: SavedAddressType | SettingsShippingAddress_address$data | null `

rquartararo: thank you!

src/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -146,18 +141,12 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
...
-    refetchAddresses(() => {
-      if (address?.createUserAddress) {
-        onAddressCreate && onAddressCreate(address.createUserAddress)
-      } else if (address?.updateUserAddress) {
-        onAddressEdit && onAddressEdit(address.updateUserAddress)
-      }
-    })
-
-    onShowToast && onShowToast(true, "Saved")
+  const onSuccess = (id: string) => {
```
rquartararo, line 144: @AlicanAkyuz I re-added the onAddressCreate/Edit functions from `index/shipping` as I realized that they include some additional logic for ARTA.

rquartararo: I also changed them to accept the address ID instead of the entire address as we only pass the id to `onSucess()` in the form component

AlicanAkyuz: makes sense! however, I think we need the full address passed in case of editing (or else we should look into refetching addresses in `index/shipping` after an address is edited here).

rquartararo: Yep! I needed to update this to pass the address for your previous bug fix

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -251,21 +249,12 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-      const shipToAddress = isCreateNewAddress()
-        ? address
-        : convertShippingAddressForExchange(
-            editedAddress
-              ? editedAddress
-              : addressList.find(
-                  address => address.internalID == selectedAddressID
-                )!
-          )
+      let phoneNumberFormatted: string = `+${
```
AlicanAkyuz, line 252: can be a `const` now

src/Components/Address/ShippingAddressForm.tsx
```diff
@@ -72,18 +76,26 @@ interface SettingsShippingAddressFormProps {
...
   onClose,
   address,
+  onSuccess,
+  onDelete,
 }) => {
   const { submitMutation: submitAddAddress } = useAddAddress()
   const { submitMutation: submitEditAddress } = useEditAddress()
   const { submitMutation: submitSetDefaultAddress } = useSetDefaultAddress()
   const { sendToast } = useToasts()
+  const [showDeleteDialog, setShowDeleteDialog] = useState<boolean>(false)
```
AlicanAkyuz, line 98: we don't need to specify type `<boolean>` here, since we have an initial value and it's inferred

src/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -40,15 +47,16 @@ interface SavedAddressesProps {
...
-    address: UpdateUserAddressMutation$data["updateUserAddress"]
-  ) => void
+  onAddressDelete?: (id: string) => void
+  onAddressEdit?: (address: {
+    internalID: string
+    attributes: ShippingAddress
+  }) => void
+  onAddressCreate?: (address: {
+    internalID: string
+    attributes: ShippingAddress
```
AlicanAkyuz, line 57: since the param types are same and used in the parent as well, maybe it's a good idea to have a type for this.

rquartararo: good idea!

PR #11425, opened 2022-11-22
fix(CX-3130): Remove "share followed artists..." from Inquiry Flow

MrSltun:

This PR solves [CX-3130]

### Description
This PR removes the "Share followed artists, categories, and galleries" from the inquiry flow

### Screenshots

| mWeb | Desktop |
| -- | -- |
| image | image |

[CX-3130]: https://artsyproduct.atlassian.net/browse/CX-3130

### DIFF
--- a/src/Components/Inquiry/__tests__/Views/InquiryBasicInfo.jest.tsx
+++ b/src/Components/Inquiry/__tests__/Views/InquiryBasicInfo.jest.tsx
@@ -89,8 +89,6 @@ describe("InquiryBasicInfo", () => {
     fill(wrapper, "profession", "Carpenter")
     fill(wrapper, "phone", "555-555-5555")
 
-    wrapper.find('[role="checkbox"]').first().simulate("click")
-
     wrapper.find("form").simulate("submit")
 
     await flushPromiseQueue()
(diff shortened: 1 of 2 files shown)


src/Components/Inquiry/__tests__/Views/InquiryBasicInfo.jest.tsx
```diff
@@ -89,16 +89,14 @@ describe("InquiryBasicInfo", () => {
...
-
     wrapper.find("form").simulate("submit")
 
     await flushPromiseQueue()
 
     expect(mockSubmitUpdateMyUserProfile).toBeCalledWith({
       profession: "Carpenter",
       phone: "555-555-5555",
-      shareFollows: false,
+      shareFollows: true,
```
MounirDhahri, line 99: **Q:** do you know why this is still being sent here?

MounirDhahri (approved):
LGTM, I only have one minor comment that is not probably blocking since it's what we will do anyway later

src/Components/Inquiry/__tests__/Views/InquiryBasicInfo.jest.tsx
MrSltun: It is defined in the update profile mutation query in the inquiry flow, and when the user is created it defaults to `nil` [in Gravity](https://github.com/artsy/gravity/blob/main/app/models/domain/user.rb#L118), this way we make sure to set it to true when the user inquires about an artwork.

Another approach would be that we set the `share_follows` to `true` on Gravity when the user sends a message and remove it from Force.

But I'm open to other approaches as well :)

PR #11430, opened 2022-11-22
feat: remove cta banner on mweb, add buttons on mweb

dariakoko:

This PR solves [CX-2954]

### Description
For mWeb:
- remove the banner that appears on top of the screen when scrolling
- add two buttons to the bottom of the page instead
- add tracking events for the new buttons on the bottom of the page (aligned with Audrey)

telegram-cloud-photo-size-2-5469829057379157367-y

[CX-2954]: https://artsyproduct.atlassian.net/browse/CX-2954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Consign/Routes/MarketingLanding/Components/CtaBanner.tsx
```diff
@@ -73,36 +73,9 @@ export const CtaBannerContent = () => {
...
-              size="small"
-              width="100%"
-            >
-              Submit an Artwork
-            </Button>
-          </Flex>
-        </Media>
-
-        <Media greaterThanOrEqual="sm">
+    <Media greaterThan="xs">
```
dariakoko, line 76: How do you feel about using Media here? If not good, let's pair to fix it?

I think it should wrap the CtaBanner component on line 27. I tried doing this, and it breaks the header - when I scroll down and then up again, the content of the header (text and two buttons) disappears. Looks like it is hidden somewhere under the image.
When I use `<FullBleed />` instead of `<FullBleedHeader />` component in the Header.tsx - I don't experience this issue

MounirDhahri: it's okay to use Media here

dariakoko: Perfect! Thanks

PR #11431, opened 2022-11-22
WIP:bug:ensure image in partner artist rail is scoped to partner

nspinazz89:
The type of this PR is: Fix

This PR solves [:px-5309]
### Description

This PR reverts back to the old way we had the artist rail because the recent change was no longer scoping to specific Partner. 

### DIFF
--- a/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
+++ b/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
@@ -1,15 +1,13 @@
 import * as React from "react"
-import { Shelf } from "@artsy/palette"
+import { Box, Shelf } from "@artsy/palette"
+import { compact } from "lodash"
+import { PartnerArtistsCarouselItemFragmentContainer } from "./PartnerArtistsCarouselItem"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useSystemContext } from "System"
 import { PartnerArtistsCarouselRendererQuery } from "__generated__/PartnerArtistsCarouselRendererQuery.graphql"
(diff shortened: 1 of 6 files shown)


dzucconi:
We don't want to do this

dzucconi (changes requested):
We don't want to use this UI. The only issue is the href? If so that's fine we can support a custom href in the Cell

dzucconi:
In fact you barely need to do anything here: `Cell` supports the `to` attr — so just use it.

nspinazz89:
> We don't want to use this UI. The only issue is the href? If so that's fine we can support a custom href in the Cell

The issue isn't just the href, it's that the large square image shown for an artist in a rail isn't scoped to a partner artist. As an example, the gallery Robin Rile Fine, currently if you were to visit, https://staging.artsy.net/partner/robin-rile-fine-art and scroll down to Banksy you would see the the "Kissing Coppers" image which is not a work that this gallery has. 
image
vs my local branch after reverting back to the way we had previously had it.
image

I think the problem is with the graphql query itself, but was hesitant to change the query directly in the CellArtist component as I had assumed it was a more general component that was being used elsewhere....but having double checked that I don't see it being imported anywhere else so I will go and change that component directly now.

dzucconi:
OH — alright I see. In which case we'll want to make a new Cell component. Say: `CellPartnerArtist` following the patterns laid out in `CellArtist`

nspinazz89:
> OH — alright I see. In which case we'll want to make a new Cell component. Say: `CellPartnerArtist` following the patterns laid out in `CellArtist`

Sounds good! Also I had found the ticket for the original design specified that small circular image, but that is removed by CellArtist. Was that a conscious/more recent design change? If so I will follow that convention.

nspinazz89:
Closing this PR in favor of https://github.com/artsy/force/pull/11462 which takes a better approach via new CellPartnerArtist Component.

PR #11437, opened 2022-11-23
refactor(auth): removes destination

dzucconi:
I guess at some point this must have done something, but no longer. Previously whatever was passed as `destination` was set as a cookie named `"destination"`. This cookie is never actually checked anywhere.

### DIFF
--- a/src/Apps/Articles/ArticlesApp.tsx
+++ b/src/Apps/Articles/ArticlesApp.tsx
@@ -8,24 +8,18 @@ import { getENV } from "Utils/getENV"
 import { ArticlesApp_viewer$data } from "__generated__/ArticlesApp_viewer.graphql"
 import { useScrollToOpenAuthModal } from "Utils/Hooks/useScrollToOpenAuthModal"
 import { ContextModule, Intent } from "@artsy/cohesion"
-import { useRouter } from "System/Router/useRouter"
 
 interface ArticlesAppProps {
   viewer: ArticlesApp_viewer$data
 }
 
 const ArticlesApp: FC<ArticlesAppProps> = ({ viewer }) => {
-  const {
-    match: { location },
-  } = useRouter()
-
   useScrollToOpenAuthModal({
     key: "editorial-signup-dismissed",
     modalOptions: {
       intent: Intent.viewEditorial,
       contextModule: ContextModule.popUpModal,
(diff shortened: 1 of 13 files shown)


src/Apps/Articles/ArticlesApp.tsx
```diff
@@ -8,24 +8,18 @@ import { getENV } from "Utils/getENV"
...
-    match: { location },
-  } = useRouter()
-
   useScrollToOpenAuthModal({
     key: "editorial-signup-dismissed",
     modalOptions: {
       intent: Intent.viewEditorial,
       contextModule: ContextModule.popUpModal,
       copy: "Sign up for the latest in art market news",
-      destination: location.pathname,
```
dzucconi, line 28: This option was only being used in modals that were opened on article pages. `redirectTo` is backfilled to `location.href` in the function that ultimately checks this. So you don't need to specify it anyway.

src/Apps/Authentication/Utils/helpers.ts
```diff
@@ -177,17 +176,11 @@ export const maybeUpdateRedirectTo = (
...
 
   if (afterSignUpAction) {
     Cookies.set(AFTER_AUTH_ACTION_KEY, JSON.stringify(afterSignUpAction))
   }
-
-  if (destination) {
-    Cookies.set("destination", destination, {
-      expires: 60 * 60 * 24,
-    })
-  }
```
dzucconi, line 190: This was the only real thing this option did. Which is never actually checked anywhere.

src/Apps/Authentication/Utils/helpers.ts
```diff
@@ -159,35 +158,39 @@ export const updateURLWithOnboardingParam = (url: string) => {
...
+  intent?: AuthIntent
+}) => {
+  // Only trigger onboarding for sign ups without a commercial intent
+  return !!(
+    type === ModalType.signup &&
+    intent &&
+    !COMMERCIAL_AUTH_INTENTS.includes(intent)
+  )
+}
+
```
dzucconi, line 175: cc @damassi — extracted out the logic for triggering onboarding into its own function, which is then also used in the analytics track call. This technically means our analytics data for onboarding is inaccurate. Don't know how much that matters, ultimately.

PR #11456, opened 2022-11-30
feat: User sees fallback cards after timeout GRO-1451

jonallured:
This PR adds a new trick to the content cards on the homepage: a 1 second timeout that renders fallback content. I also included a way to override this fallback timing in order to force the rendering of the fallback content for ease of QAing.

<details>
<summary>Some Screenshots</summary>

Screen Shot 2022-11-29 at 9 44 35 PM

Screen Shot 2022-11-29 at 9 44 42 PM

</details>

I took a while to land this solution. After some pairing with @dzucconi where we were banging our heads against stale state, I started over and took it slower. I read some things and found that generally speaking `useRef` and `setTimeout` are a good way to ensure that your state is up-to-date when the callback fires.

But along the way I also found that I wanted to extract things and work at a single level of abstraction so there's some moves here to that end. I wanted a boss component so that's `HomeContentCards`. It's where all the state is managed and where we pick between the types of cards to render. Those card types are:

* `BrazeCards` - the cards that you see when we get a successful response from the Braze API
* `FallbackCards` - the cards we run when something goes wrong
* `PlaceholderCards` - the skeleton cards that hold the layout while we're waiting to hear from Braze

That boss component has three `useEffect` hooks:

* one that depends on nothing and sets `appboy` when segment is fully loaded
* one that depends on `appboy` and sets up listeners and the timeout
* one that keeps our card count ref in sync

Once I realized that I could separate these hooks I went from a puzzle to nice declarative code. And from there it's essentially an if/elseif/else but everybody knows @dzucconi hates those so I did it with a ternary and early return. Looks cute. 💃 

TODO:

* [x] clean up those anys
* [x] take a run at getting some tests around this shapeshiftyness

https://artsyproduct.atlassian.net/browse/GRO-1451

/cc @artsy/grow-devs


src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
```diff
@@ -0,0 +1,71 @@
...
+    appboy.requestContentCardsRefresh()
+
+    return () => {
+      appboy.removeSubscription(subscriptionId)
+      clearTimeout(timeout)
+    }
+  }, [appboy, timeoutAmount])
+
+  useEffect(() => {
+    cardsLengthRef.current = cards.length
```
dzucconi, line 62: Interesting!

src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { useRouter } from "System/Router/useRouter"
+import { paramsToCamelCase } from "Components/ArtworkFilter/Utils/urlBuilder"
+
+const sortCards = (lhs, rhs) => {
+  const lhsPosition = (lhs.extras || {}).position || lhs.id
+  const rhsPosition = (rhs.extras || {}).position || rhs.id
+  return lhsPosition > rhsPosition ? 1 : -1
+}
+
+const DEFAULT_TIMEOUT_AMOUNT = 1000
```
dzucconi, line 15: `non-blocking`: I *suspect* this default will wind up being a bit too aggressive. I'd probably opt for something closer to like 5 seconds.

dzucconi (approved):

src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export const HomeContentCards: React.FC = () => {
+  const { match } = useRouter()
+  const { brazeTimeoutAmount } = paramsToCamelCase(match?.location.query) as {
+    brazeTimeoutAmount?: number
```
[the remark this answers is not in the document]
jonallured: IDK, maybe it's over-engineering but I wanted it for when I show this to folks to QA.

src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
jonallured: I don't have strong feelings here so I'm happy to move to 5 seconds. 👍

jonallured:
Ok with [b242fa8](https://github.com/artsy/force/pull/11456/commits/b242fa8d1abbf8402df689caf119813c559d95b3) I have added some specs. I paired with @dzucconi and @laurabeth on them and we were able to get most of it working but for whatever reason that last test just won't behave. I was authorized by @dzucconi to mark it pending because "who cares". 😝

jonallured:
Ok going to merge this now that my list of todos is complete and the PR is green. 👍

src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
jonallured: Turns out this was a Bad Idea ™️ - much better to just update the ref prior to setting the cards...avoids a race condition!

PR #11462, opened 2022-11-30
fix:ensure artist carousel image is scoped to partner artist

nspinazz89:
The type of this PR is: Fix

This PR solves [PX-5309]
### Description

This PR addresses an issue where the artwork image on an artist rail is associated with an Artsy wide Artist instead of the PartnerArtist as it should be. 

Current incorrect image: 
image

image

After update Correct Image:
image

[PX-5309]: https://artsyproduct.atlassian.net/browse/PX-5309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
+++ b/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
@@ -6,8 +6,8 @@ import { PartnerArtistsCarouselRendererQuery } from "__generated__/PartnerArtist
 import { PartnerArtistsCarousel_partner$data } from "__generated__/PartnerArtistsCarousel_partner.graphql"
 import { PartnerArtistsCarouselPlaceholder } from "./PartnerArtistsCarouselPlaceholder"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
-import { CellArtistFragmentContainer } from "Components/Cells/CellArtist"
-import { extractNodes } from "Utils/extractNodes"
+import { CellPartnerArtistFragmentContainer } from "Components/Cells/CellPartnerArtist"
+import { compact } from "lodash"
 import { FollowArtistButtonQueryRenderer } from "Components/FollowButton/FollowArtistButton"
 import { ContextModule } from "@artsy/cohesion"
 
@@ -27,25 +27,27 @@ export const PartnerArtistsCarousel: React.FC<PartnerArtistsCarouselProps> = ({
   ) {
     return null
   }
-
-  const artists = extractNodes(partner.artistsConnection)
+  const { slug } = partner
+  const artists = compact(partner.artistsConnection.edges)
 
   return (
     <Shelf alignItems="flex-start">
-      {artists.map(artist => (
-        <CellArtistFragmentContainer
-          key={artist.internalID}
-          artist={artist}
-          to={`/partner/${partner.slug}/artists/${artist.slug}`}
-          FollowButton={
-            <FollowArtistButtonQueryRenderer
-              id={artist.internalID}
-              contextModule={ContextModule.recommendedArtistsRail}
-              size="small"
-            />
-          }
-        />
-      ))}
+      {artists
+        .filter(artist => artist !== null)
+        .map(artist => (
+          <CellPartnerArtistFragmentContainer
+            key={artist.node?.internalID}
+            artistPartnerEdge={artist}
+            to={`/partner/${slug}/artists/${artist.node?.slug}`}
+            FollowButton={
+              <FollowArtistButtonQueryRenderer
+                id={artist.node?.internalID!}
+                contextModule={ContextModule.recommendedArtistsRail}
+                size="small"
+              />
+            }
+          />
+        ))}
     </Shelf>
   )
 }
@@ -63,10 +65,11 @@ export const PartnerArtistsCarouselFragmentContainer = createFragmentContainer(
         ) {
           edges {
             node {
-              ...CellArtist_artist
+              id
               internalID
               slug
             }
+            ...CellPartnerArtist_partnerArtist
           }
         }
       }
--- a/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
+++ b/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
@@ -65,7 +65,6 @@ export const PartnerArtistsCarouselFragmentContainer = createFragmentContainer(
         ) {
           edges {
             node {
-              id
               internalID
               slug
             }
--- a/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
+++ b/src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
@@ -8,8 +8,6 @@ import { PartnerArtistsCarouselPlaceholder } from "./PartnerArtistsCarouselPlace
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
 import { CellPartnerArtistFragmentContainer } from "Components/Cells/CellPartnerArtist"
 import { compact } from "lodash"
-import { FollowArtistButtonQueryRenderer } from "Components/FollowButton/FollowArtistButton"
-import { ContextModule } from "@artsy/cohesion"
 
 const PAGE_SIZE = 20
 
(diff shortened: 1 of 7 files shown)


src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
```diff
@@ -27,25 +27,27 @@ export const PartnerArtistsCarousel: React.FC<PartnerArtistsCarouselProps> = ({
...
-            <FollowArtistButtonQueryRenderer
-              id={artist.internalID}
-              contextModule={ContextModule.recommendedArtistsRail}
-              size="small"
-            />
-          }
-        />
-      ))}
+      {artists
+        .filter(artist => artist !== null)
```
nspinazz89, line 36: Probably not necessary, but wanted to have extra protection here.

damassi: `compact` should filter this out up above

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
```diff
@@ -63,10 +65,11 @@ export const PartnerArtistsCarouselFragmentContainer = createFragmentContainer(
         ) {
           edges {
             node {
-              ...CellArtist_artist
+              id
```
damassi, line 68: Is this being used? Note that `id` = relay globally unique ID and has no relationship to our API data. (id = internalID in terms of data)

nspinazz89: good point I think you're right and its not necessary.

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
nspinazz89: I have tried calling that on artists before but it was throwing an error because it's not actually an array?
image

damassi: Ah, should do something like 
```tsx
const artists = extractNodes(partner.artistsConnection)
```

I wouldn't worry about compacting or filtering here though. The pattern we typically use is in a comment below

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
```diff
@@ -27,25 +27,27 @@ export const PartnerArtistsCarousel: React.FC<PartnerArtistsCarouselProps> = ({
...
-              id={artist.internalID}
-              contextModule={ContextModule.recommendedArtistsRail}
-              size="small"
-            />
-          }
-        />
-      ))}
+      {artists
+        .filter(artist => artist !== null)
+        .map(artist => (
```
damassi, line 37: We should always [avoid implict returns](https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-implicit-returns-on-react-components) in react components

damassi: And the reason is.. code always gets returned to / expanded, even if it seems final. Like for example right now 😄 

```tsx
if (!artist) {
  return null
}
```

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
```diff
@@ -27,25 +27,27 @@ export const PartnerArtistsCarousel: React.FC<PartnerArtistsCarouselProps> = ({
...
+          <CellPartnerArtistFragmentContainer
+            key={artist.node?.internalID}
+            artistPartnerEdge={artist}
+            to={`/partner/${slug}/artists/${artist.node?.slug}`}
```
dzucconi, line 41: This should be the default.

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
```diff
@@ -27,25 +27,27 @@ export const PartnerArtistsCarousel: React.FC<PartnerArtistsCarouselProps> = ({
...
+            FollowButton={
+              <FollowArtistButtonQueryRenderer
+                id={artist.node?.internalID!}
+                contextModule={ContextModule.recommendedArtistsRail}
```
dzucconi, line 45: I don't think this needs the `recommendedArtistsRail` context since that's not what this is. We don't necessarily need a contextModule; in which case you can just leave this as the default by not passing in a `FollowButton`

src/Components/Cells/CellPartnerArtist.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const CellPartnerArtist: FC<CellPartnerArtistProps> = ({
+  artistPartnerEdge,
+  mode = "RAIL",
+  displayCounts,
+  FollowButton,
+  ...rest
+}) => {
+  const width = mode === "GRID" ? "100%" : DEFAULT_CELL_WIDTH
+  const artwork = extractNodes(artistPartnerEdge.artworksConnection)
```
dzucconi, line 34:
```suggestion
  const [artwork] = extractNodes(artistPartnerEdge.artworksConnection)
```

src/Components/Cells/CellPartnerArtist.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const image = artwork[0].image?.cropped
```
dzucconi, line 35:
```suggestion
  const image = artwork.image?.cropped
```

src/Components/Cells/CellPartnerArtist.tsx
```diff
@@ -0,0 +1,140 @@
...
+      )}
+    </Box>
+  )
+}
+
+export const CellPartnerArtistFragmentContainer = createFragmentContainer(
+  CellPartnerArtist,
+  {
+    artistPartnerEdge: graphql`
+      fragment CellPartnerArtist_partnerArtist on ArtistPartnerEdge {
```
dzucconi, line 116: You have access to the `partner` on this edge so you can pull out the href to construct the nested route.

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
nspinazz89: so that is what I tried originally but then that doesn't work because the CellPartnerArtistsFragmentContainer needs that fragmentRef 
image
vs now artists just being fully deconstructed (I think that is the right term here, but not sure) 
image

dzucconi: Yep, we're not extracting nodes here since this expects an edge. It's super annoying.

dzucconi: If we fixed our connection types to better reflect nullability this wouldn't be an issue.

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
dzucconi: Yah

src/Components/Cells/CellPartnerArtist.tsx
nspinazz89: ah cool good point!

src/Components/Cells/CellPartnerArtist.tsx
nspinazz89: ah nice that's cleaner!

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
nspinazz89: ah cool didn't know that! Will keep that in mind in future!

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
damassi: Ah 👍

src/Apps/Partner/Components/PartnerArtists/PartnerArtistsCarousel/PartnerArtistsCarousel.tsx
nspinazz89: ah cool done as well!

jpotts244 (commented):
Looks good to me as a novice, just cc-ing @dzucconi & @damassi for one final review

dzucconi (approved):
Looks good!

PR #11465, opened 2022-12-01
fix: broken link from artwork page to auction house

anandaroop:

This PR solves [FX-4484]

### Description

Fixes a bug affecting all auction artwork pages.

The artwork page renders an auction partner entity header that attempts to link out to a non-existent auction profile page. Such profile pages for auction partners are not reachable on artsy.net (I'm not sure if they ever were.)

Screen Shot 2022-11-30 at 11 19 01 PM

The fix here is to simply display that^ entity header without enclosing it in an `<a>`, behavior which is already supported in the `EntityHeaderPartner` component.

(There is also likely a fix to be made in MP, which maybe should not [construct these invalid links](https://github.com/artsy/metaphysics/blob/498f9193668e640ead83f3acf012a6732e1d4e72/src/schema/v2/partner/partner.ts#L568) in the case of auction partners) 

[FX-4484]: https://artsyproduct.atlassian.net/browse/FX-4484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nspinazz89 (dismissed):
Looks good to me 🚀

anandaroop:
Note that in https://github.com/artsy/force/pull/11465/commits/528ce3222907367c91a4d7613257b22df25da566 I updated the approach here based on some conversations in Slack [here](https://artsy.slack.com/archives/C9SATFLUU/p1669911920794469) and [here](https://artsy.slack.com/archives/C9SATFLUU/p1669921875727559?thread_ts=1668633169.052509&cid=C9SATFLUU)

TLDR: after going way down the rabbit hole I've decided to simply mirror [the check that determines whether a partner page can even be rendered in the first place](https://github.com/artsy/force/blob/77d3ed29eb605ad3c5e5d05982dc03e82ed39db8/src/Apps/Partner/PartnerApp.tsx#L29-L31). That brings this entity link into sync with reality (without needing to reference whether it's for an auction, private partner page, etc.)

PR #11472, opened 2022-12-01
feat: Launch Braze Content Cards GRO-1324

jonallured:
This PR will launch the Braze content cards on web! 🎉 

It also fixes (at least) two things:

1. A race condition where Braze cards aren't rendered
2. A bug where client-side routing breaks the call to Braze for cards

So yeah, let's merge this tomorrow morning and behold the glory.

ALSO

Over on https://github.com/artsy/force/pull/11456 I thought I was a real hotshot and knew it was fine to pend that spec but it turns out it was revealing that race condition bug. 😬 This fixes the bug and also makes the test pass IMAGINE THAT.

Finally I wanted to mention that in a follow up to this PR I will be deleting the `HomeHeroUnit` code. I could have done that here but for my sanity I want it to be easy to rollback in case something goes wrong.

/cc @artsy/grow-devs @isakelaris @jpoczik

### DIFF
--- a/src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
+++ b/src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
@@ -29,17 +29,28 @@ export const HomeContentCards: React.FC = () => {
   const cardsLengthRef = useRef(cards.length)
 
   useEffect(() => {
-    window.analytics?.ready(() => {
+    if (appboy) return
+
+    if (window.appboy) {
       setAppboy(window.appboy)
-    })
-  })
+    } else if (window.analytics) {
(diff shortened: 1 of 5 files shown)


src/Apps/Home/Components/HomeContentCards/HomeContentCards.tsx
```diff
@@ -29,17 +29,28 @@ export const HomeContentCards: React.FC = () => {
   const cardsLengthRef = useRef(cards.length)
 
   useEffect(() => {
-    window.analytics?.ready(() => {
+    if (appboy) return
+
+    if (window.appboy) {
```
dzucconi, line 34: Ah!

PR #11491, opened 2022-12-05
fix(cellpartnerartist): guard for artists without artworks

dzucconi:
Minor fix for artists that don't have artworks.


nspinazz89 (approved):
LGTM 🚀

PR #11505, opened 2022-12-06
feat(CX-3178): update account page

dariakoko:

This PR solves [CX-3178]

### Description

"Full Name" field was removed and replaced with "Price Range" field + moved components to the Account tab

| Web | mWeb |
| --- | --- |
| image |  image |

[CX-3178]: https://artsyproduct.atlassian.net/browse/CX-3178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
+++ b/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
@@ -9,10 +9,11 @@ import {
   useToasts,
 } from "@artsy/palette"
 import { PRICE_BUCKETS } from "Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileAboutYou"
-import { email, password } from "Components/Authentication/Validators"
+import { email, name, password } from "Components/Authentication/Validators"
 import { Form, Formik } from "formik"
 import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { useFeatureFlag } from "System/useFeatureFlag"
 import * as Yup from "yup"
 import { SettingsEditSettingsInformation_me$data } from "__generated__/SettingsEditSettingsInformation_me.graphql"
 import { useUpdateSettingsInformation } from "./useUpdateSettingsInformation"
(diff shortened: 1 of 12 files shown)


src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
```diff
@@ -84,14 +98,9 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
...
             const updatedMe = updateMyUserProfile?.me
 
-            resetForm({
-              values: {
-                name: updatedMe?.name ?? name,
-                email: updatedMe?.email ?? email,
-                phone: updatedMe?.phone ?? phone,
-                password: "",
-              },
-            })
```
dariakoko, line 94: I am not sure why we need this, If you feel like we can remove it, let's do so. I think there might be a reason to keep it tho

I cannot keep using the `resetForm`, so using `setFieldValue` instead
Can't use `resetForm` because in the response we're not getting the `priceRange`, getting only `priceRangeMin` and `priceRangeMax`. And I can't use `resetForm` with a value missing. In order not to add extra logic I'm not resetting this field

src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
```diff
@@ -152,6 +150,24 @@ export const SettingsEditSettingsInformation: React.FC<SettingsEditSettingsInfor
...
 
+              <Select
+                title="Price Range"
+                options={PRICE_BUCKETS}
+                selected={values.priceRange}
+                onSelect={value => {
+                  setFieldValue("priceRange", value)
+
+                  // We don't actually accept a priceRange,
+                  // so have to split it into min/max
```
olerichter00, line 161: 👍

olerichter00 (approved):
Looks good 💯

src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsInformation.tsx
MounirDhahri: probably it's being used to make sure that the formik `dirty` boolean is set to false afterwards - contrarily to using `setField`. If the `dirty` is not being used, this shouldn't be an issue

MounirDhahri (dismissed):
Niice. 

Are the changes above hidden behind a feature flag @dariakoko

olerichter00:
@dariakoko I'll merge 👍

PR #11507, opened 2022-12-07
fix: save user's coordinates

kiryl-zubarau:
The type of this PR is: Fix

This PR solves [TX-926]

### Description

This PR fix user's timezone by saving `coordinates` to the user's location.


src/Components/LocationAutocompleteInput.tsx
```diff
@@ -103,6 +103,7 @@ export const LocationAutocompleteInput: FC<LocationAutocompleteInputProps> = ({
 
   const handleSuggestionsFetchRequested = useMemo(
     () => debounce(updateSuggestions, DEBOUNCE_DELAY),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
rquartararo, line 106: It was brought up recently not to disable this error unless absolutely necessary. A solution could be to wrap `updateSuggestions` in a useCallback hook.

kiryl-zubarau: Updated

rquartararo (dismissed):
Looks great 👍

PR #11512, opened 2022-12-07
docs: fix typo

patrinoua:
update wording

### DIFF
--- a/docs/best_practices.md
+++ b/docs/best_practices.md
@@ -46,9 +46,9 @@ Our design system [Palette](https://github.com/artsy/palette) is used for most U
 
 ### Commonly used components
 
-- [Box](https://palette.artsy.net/elements/box/) - is a div which we can pass extra props to, like margins for example. The content is displayed vertically.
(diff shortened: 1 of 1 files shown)


dzucconi:
This isn't accurate either. `Flex` isn't a component for displaying content horizontally. It's a [`Box`](https://styled-system.com/guides/build-a-box/) with `display: flex` enabled. Nothing more. Neither `Box` nor `Flex` have anything inherently to do with the direction that content is displayed in.

dzucconi:
I would just remove this section and link to the [Palette docs](https://palette.artsy.net/)

patrinoua:
@dzucconi @damassi I used your wording [link](https://artsy.slack.com/archives/C2TQ4PT8R/p1670258912263329?thread_ts=1670258695.085059&cid=C2TQ4PT8R).
However, please suggest a better way do describe when to use Flex and when to use Box with a few words, (rather than describe what they are made of).

dzucconi:
Yeah, I find that wording to be inaccurate/misleading.

I don't think this doc should explain when to use a basic CSS property.

> `Flex` is a `Box` component with `display: flex` set as default.

Is sufficiently clear. You should know what that means if you're contributing to Force.

PR #11516, opened 2022-12-08
feat: Update Edit Profile page

MrSltun:

This PR solves [CX-3174]

### Description

Added Edit Profile page component 
Route is settings/edit-settings (to discuss later with the team and Audrey, it probably should be renamed)

| Web | mWeb |
| --- | --- |
| image | image image |

[CX-3174]: https://artsyproduct.atlassian.net/browse/CX-3174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Auction/Components/Form/Utils/validationSchemas.ts
```diff
@@ -50,3 +50,9 @@ export const biddingValidationSchemas = {
     selectedBid,
   }),
 }
+
+export const editProfileVerificationSchema = Yup.object().shape({
```
olerichter00, line 54: **suggestion:** What do you think about moving the schema from `src/Apps/Auction/....` to `src/Apps/ColelctorProfile/...`?

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
```diff
@@ -0,0 +1,221 @@
...
+          <Input
+            title="Full name"
+            placeholder="Full name"
+            name="full-name"
+            required
+            maxLength={256}
+            value={values.name}
+            onChange={({ target: { value } }) => {
+              setFieldValue("name", value)
+            }}
```
olerichter00, line 98: **suggestion:** You could use Formik' [handleChange](https://formik.org/docs/api/formik#handlechange-e-reactchangeeventany--void) helper here ([example](https://github.com/artsy/force/blob/main/src/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/Components/ArtworkDetailsForm.tsx#L235)).

```suggestion
            onChange={handleChange}
```

olerichter00 (approved):
Looks good 🌟 

I've left a few minor comments.

src/Apps/Auction/Components/Form/Utils/validationSchemas.ts
dariakoko: Ouch why it is in the Auction
Will move to CollectorProfile 👌

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
dariakoko: Hmmm can't make this change quickly
@MrSltun did you implement it like that for a reason?

olerichter00: Hmm, shouldn't replacing `setFieldValue` with `onChange` just work for most of these inputs?

dariakoko: In this case I can't edit the initial value

MrSltun: I was testing `handleChange` but it wasn't updating the values for some reason, so I replaced it with the current implementation

MrSltun: maybe `handleChange` didn't work since the field name is different than the value name? (like `name="full-name"` while the value is `name`)

dariakoko: Oh wow, looks like you're right

dariakoko: Works with everything, but breaks the location field
I'd keep it as it is, location requires custom updates, with normalize function

olerichter00: Let's change it to `handleChange` wherever it is possible and use a custom handler for the location field ([similar to the SWA form](https://github.com/artsy/force/blob/main/src/Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/Components/ArtworkDetailsForm.tsx#L192)).

dariakoko: I had an [issue with the `<LocationAutocompleteInput />` component again](https://artsy.slack.com/archives/CP9P4KR35/p1662033427441369), didn't realise that yesterday
Now all is working with the suggested changes  🎉

dariakoko:
I'll merge it @MrSltun

PR #11529, opened 2022-12-09
feat: Use curated trending artists for home page GRO-1381

jonallured:
Another sibling PR to https://github.com/artsy/metaphysics/pull/4605 and https://github.com/artsy/eigen/pull/7839 this one just updates Force and is independent of the merging of the other ones.

https://artsyproduct.atlassian.net/browse/GRO-1381

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/Home/Components/HomeTrendingArtistsRail.tsx
+++ b/src/Apps/Home/Components/HomeTrendingArtistsRail.tsx
@@ -28,7 +28,7 @@ const HomeTrendingArtistsRail: React.FC<HomeTrendingArtistsRailProps> = ({
 }) => {
   const { trackEvent } = useTracking()
 
-  const artists = extractNodes(viewer.artistsConnection)
+  const artists = extractNodes(viewer.curatedTrendingArtists)
 
   if (artists.length === 0) {
     return null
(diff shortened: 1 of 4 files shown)


src/Apps/Home/Components/HomeTrendingArtistsRail.tsx
```diff
@@ -96,7 +96,7 @@ export const HomeTrendingArtistsRailFragmentContainer = createFragmentContainer(
   {
     viewer: graphql`
       fragment HomeTrendingArtistsRail_viewer on Viewer {
-        artistsConnection(sort: TRENDING_DESC, first: 99) {
+        curatedTrendingArtists(first: 99) {
```
anandaroop, line 99: `99` 🧐 wonder what that was/is about!

anandaroop (approved):
Works for me, but I'm wondering about that 99, haha.

src/Apps/Home/Components/HomeTrendingArtistsRail.tsx
jonallured: Actually yeah now that I'm pausing long enough to actually read that argument I'm not sure what it's doing haha. Any thoughts @dzucconi @damassi ?

dzucconi: I have no idea why this is set to fetch a crazy large number (why 99 and not 100 too?) but can we actually change this to something sensible like 20? It's been on my mental list for a while; looks ridiculous on the homepage.

jonallured: Yeah I'm happy to change it - holler if anybody knows of a good reason not to!!!

dzucconi: I was assuming the original assumption was that there was some set number of them and wanted to fetch them all.

PR #11536, opened 2022-12-12
fix: Edit artwork form back button

olerichter00:
## Description

Fix the back button on the edit artwork page to behave correctly.


src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
```diff
@@ -226,7 +226,11 @@ export const MyCollectionArtworkForm: React.FC<MyCollectionArtworkFormProps> = (
   }
 
   const goBack = () => {
-    router.go(-1)
```
olerichter00, line 229: The user has to press the button twice after editing the form 🤔

dariakoko (approved):

src/Apps/MyCollection/Routes/EditArtwork/MyCollectionArtworkForm.tsx
MrSltun: but in this case, users who would add artworks from the Insights tab would be sent to my collection

PR #11543, opened 2022-12-12
fix: Guard against ready but no appboy GRO-1465

jonallured:
While doing some troubleshooting with @dzucconi we found that it was possible to have Segment fire the ready event for the the Braze SDK to not exist on window. This PR captures that case in a test and guards against it.

https://artsyproduct.atlassian.net/browse/GRO-1465

/cc @artsy/grow-devs


dzucconi (approved):
🙏

PR #11550, opened 2022-12-13
refactor(featureflags): returns better types

dzucconi:
@mzikherman correctly pointed out that I wasn't checking for `enabled`. This wasn't a bug in the code because of the guarded access but it was technically incorrect.

I changed the return type of `getFeatureVariant` to better reflect the actual return. Unleash's provided types are not very good. I've left alone the hooks since I don't want to go around and fix all the call sites and they incidentally work fine as is. But something to be aware of.

### DIFF
--- a/src/Utils/resizer.ts
+++ b/src/Utils/resizer.ts
@@ -30,9 +30,12 @@ const DEFAULT_IMAGE_SERVICE: ImageService = "gemini"
 
 export const getImageService = (): ImageService => {
   const variant = getFeatureVariant("image-service")
-  const requestedImageService = variant?.payload?.value as ImageService
 
-  return requestedImageService || DEFAULT_IMAGE_SERVICE
+  if (!!variant && "payload" in variant && variant.payload && variant.enabled) {
+    return (variant.payload.value || DEFAULT_IMAGE_SERVICE) as ImageService
+  }
+
+  return DEFAULT_IMAGE_SERVICE
 }
 
 export const crop = (


src/System/useFeatureFlag.tsx
```diff
@@ -18,6 +18,11 @@ interface VariantTrackingProperties {
   payload?: string
 }
 
+interface DisabledVariant {
+  enabled: false
+  name: "disabled"
+}
```
dzucconi, line 24: Honestly: weird that they do this.

src/Utils/resizer.ts
```diff
@@ -30,9 +30,12 @@ const DEFAULT_IMAGE_SERVICE: ImageService = "gemini"
 
 export const getImageService = (): ImageService => {
   const variant = getFeatureVariant("image-service")
-  const requestedImageService = variant?.payload?.value as ImageService
 
-  return requestedImageService || DEFAULT_IMAGE_SERVICE
+  if (!!variant && "payload" in variant && variant.payload && variant.enabled) {
```
dzucconi, line 34: Checks that: we have the variant, the variant isn't a disabled variant, the variant has a payload and for good measure, that it's enabled. Even though if it has a payload by default it should be enabled. 🤷 Trust no one.

dzucconi:
We should also address how and when we send those `console.error`s — which also feel like they should be `warn`. The specs are noisy now.

PR #11552, opened 2022-12-13
feat: add MyC artworks to the my collection tab on collector profile

dariakoko:

This PR solves [CX-3246]

### Description

I reused MyCollectionRouteRefetchContainer, changed only the name of the container, component, query and other things like that 
**_I didn't hide the changes behind a FF_**, I think if it's off, the user won't be able to reach this page anyway. I would test the experience with FF off during the QA session when all work is done.

| Web | mWeb |
| --- | --- |
| image | image |

[CX-3246]: https://artsyproduct.atlassian.net/browse/CX-3246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun (approved):
🌟

MounirDhahri (approved):
LGTM!
I would have prefered though if you have imported my collection from the folder where it exists right now. this way, in case we make changes there, they will be in both places. also, we won't loose any git history

dariakoko:
@MounirDhahri agree! will do it now
Yesterday had some issues with importing the module, but I get that the issue was now

src/Apps/Settings/Routes/MyCollection/MyCollectionRoute.tsx
```diff
@@ -29,7 +29,7 @@ import {
 import { MyCollectionRoute_me$data } from "__generated__/MyCollectionRoute_me.graphql"
 import { EmptyMyCollectionPage } from "./Components/EmptyMyCollectionPage"
 
-interface MyCollectionRouteProps {
+export interface MyCollectionRouteProps {
```
dariakoko, line 32: We need this export to be able to reuse the component in another file

src/Apps/CollectorProfile/Routes/MyCollection/CollectorProfileMyCollectionRoute.tsx
```diff
@@ -1,14 +1,3 @@
...
-
-  if (!isCollectorProfileEnabled) {
-    return null
-  }
-
-  return <Text>Ja Hallo!</Text>
-}
-
-export const CollectorProfileMyCollectionRouteFragmentContainer = CollectorProfileMyCollectionRoute
+export const CollectorProfileMyCollectionRouteRefetchContainer = MyCollectionRouteRefetchContainer
```
olerichter00, line 3: **question:** Can we just remove this file and import `MyCollectionRouteRefetchContainer` directly wherever it is needed?

dariakoko: I'd keep this file just to make it clear that this is a different thing
When cleaning up the FF I'd remove this file and just move the MyCollectionRouteRefetchContainer here + renamed some variables, queries, etc

olerichter00: Sounds good 👍  To be honest, I cannot tell what `CollectorProfileMyCollectionRoute` is supposed to be 🤔.

dariakoko: MyCollection, but in the collector profile 😅

dariakoko: collector-profile/my-collection

olerichter00: Ooooh, I see :)

PR #11553, opened 2022-12-13
feat(artquiz): wires up quiz page flow

dzucconi:
This picks up from @laurabeth's PR and refactors the context and wires up the entire flow from start to finish.

## To do

- [ ] Wire up GraphQL layer
- [ ] UI animations and transitions
- [ ] Result tab contents
- [ ] Email dialog

I'll follow this up with the wireframe of the email dialog and then move on to animations and polish.

## Notes

* I removed all the GraphQL stuff: we don't need to make assumptions about fetching and it wasn't mergeable. The context layer gets dramatically simpler. We should avoid making assumptions about what we *think* we'll need to implement in favor of getting the mechanics immediately working.
* We can remove a large amount of layout complexity by avoiding manually calculating full screen dimensions. We can just naturally fill the screen. I had to add a `height="100%"` to an element in the `AppShell` for this to work.
* I'm aware that there's a single place where will run into manually measuring: the image in the main section will have to scale proportionally according to **both** a fluid width and height. We actually have an abstraction around the work required there that was built for articles: [ArticleZoomGalleryResponsiveBox](https://github.com/artsy/force/blob/main/src/Apps/Article/Components/ArticleZoomGallery/ArticleZoomGalleryResponsiveBox.tsx#L16). We'll extract this into a general use component, but only once we have real data and real images.

-----

https://user-images.githubusercontent.com/112297/207452743-878da737-dbac-4772-b6e2-cf5037edeff9.mov


src/DevTools/flushPromiseQueue.ts
```diff
@@ -2,5 +2,5 @@
  * test utility to wait for all promises to be resolved
  */
 export function flushPromiseQueue() {
-  return new Promise(r => setTimeout(r, 0))
+  return new Promise(jest.requireActual("timers").setImmediate)
```
dzucconi, line 5: Looks like this is causing a spec failure — Needed to be done to support usage in specs with fake timers.

src/Apps/ArtQuiz/Components/ArtQuizMain.tsx
```diff
@@ -0,0 +1,83 @@
...
+  Text,
+} from "@artsy/palette"
+import { useArtQuizContext } from "Apps/ArtQuiz/ArtQuizContext"
+import { FC } from "react"
+import { RouterLink } from "System/Router/RouterLink"
+
+interface ArtQuizMainProps {}
+
+export const ArtQuizMain: FC<ArtQuizMainProps> = () => {
+  const { total, index, onNext, onPrevious } = useArtQuizContext()
```
dzucconi, line 18: It occurs to me now that we don't even need a context? All the state is isolated in this lone component. Will just remove. If we need to lift state, we can always do that.

dzucconi:
Also a bit more on the need for a context. Right now we have `start` => `middle` => `end` which are all represented in routes. If this was any more complex, routing-wise, I'd want us to have a context for entire sake of managing the current route, so that we could decouple knowledge about what's before and after any given view. As it stands this is fine since it's only a single view that has to know about it.

PR #11563, opened 2022-12-14
feat(artquiz): implements animation

dzucconi:
I thought I'd open this early as draft and update it as I approach the problem, since we don't do animation that often.

I'm approaching this in 3 steps:
- [x] Animate the buttons
- [x] Animate the cards on desktop
- [x] Animate the cards on mobile via swipe

I'll be using the Web Animations API (edit: used simple transitions for the swipes for now), which is quite nice and well supported with our user base. I'll leave comments on these individual steps.

### DIFF
--- a/src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
+++ b/src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
@@ -1,13 +1,14 @@
-import { Clickable, ClickableProps } from "@artsy/palette"
+import {
+  Clickable,
+  ClickableProps,
+  CloseIcon,
+  HeartFillIcon,
+  HeartIcon,
+} from "@artsy/palette"
 import React, { FC, useRef } from "react"
 import { useMode } from "Utils/Hooks/useMode"
 
 type Mode = "Pending" | "Animating" | "Done"
-
-interface ArtQuizButtonProps extends ClickableProps {
-  children: JSX.Element | ((props: { mode: Mode }) => JSX.Element)
-  onClick: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
-}
 
 const ANIMATION_DURATION = 250
 
@@ -23,7 +24,13 @@ const KEYFRAME_ANIMATION_OPTIONS: KeyframeAnimationOptions = {
   iterations: 1,
 }
 
+interface ArtQuizButtonProps extends ClickableProps {
+  variant: "Like" | "Dislike"
+  onClick: (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+}
+
 export const ArtQuizButton: FC<ArtQuizButtonProps> = ({
+  variant,
   children,
   onClick,
   ...rest
@@ -65,9 +72,18 @@ export const ArtQuizButton: FC<ArtQuizButtonProps> = ({
       py={4}
       px={6}
       onClick={handleClick}
+      aria-label={variant}
       {...rest}
     >
-      {typeof children === "function" ? children({ mode }) : children}
+      {variant === "Dislike" && <CloseIcon width={40} height={40} title="" />}
+
+      {variant === "Like" && mode === "Animating" && (
+        <HeartFillIcon width={40} height={40} title="" />
+      )}
+
+      {variant === "Like" && mode !== "Animating" && (
+        <HeartIcon width={40} height={40} title="" />
+      )}
     </Clickable>
   )
 }
--- a/src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
+++ b/src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
@@ -1,7 +1,6 @@
 import { Clickable, ClickableProps } from "@artsy/palette"
 import React, { FC, useRef } from "react"
 import { useMode } from "Utils/Hooks/useMode"
-import { wait } from "Utils/wait"
 
 type Mode = "Pending" | "Animating" | "Done"
 
@@ -27,6 +26,7 @@ const KEYFRAME_ANIMATION_OPTIONS: KeyframeAnimationOptions = {
 export const ArtQuizButton: FC<ArtQuizButtonProps> = ({
   children,
   onClick,
+  ...rest
 }) => {
   const [mode, setMode] = useMode<Mode>("Pending")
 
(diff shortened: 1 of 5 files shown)


dzucconi:
First up is the buttons via ebd974d662e24f002da4da7e56e0028edc57d819 - which is pretty straightforward. Here's what that looks like:

https://user-images.githubusercontent.com/112297/207681747-50b8c032-42ab-4ea1-8a66-469f72070ee2.mov

-----

One thing you'll quickly realize is that you can't accomplish this using a traditional `css` animation property because you need to cancel any in progress animation to prevent it piling up if you click faster than the runtime of the animation. You also can't really achieve this via transition because you want to ease over the entire cycle of the animation.

src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+    // Click executes immediately
+    onClick(event)
+
+    // Animation plays on
+    animationRef.current = nodeRef.current.animate(
+      KEYFRAMES,
+      KEYFRAME_ANIMATION_OPTIONS
+    )
+
+    await animationRef.current.finished
```
dzucconi, line 57: Setting a normal timer for the duration of the animation wouldn't be accurate.

src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    setMode("Done")
+
+    await wait(ANIMATION_DURATION)
```
dzucconi, line 61: This will get adjusted as we tweak the card animations. It feels awkward right now.

src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    setMode("Pending")
+  }
+
+  return (
+    <Clickable ref={nodeRef as any} py={4} px={6} onClick={handleClick}>
+      {typeof children === "function" ? children({ mode }) : children}
```
dzucconi, line 68: Here I'm just passing along the state of the component to the children so that I can use that to fill the heart. I could have also just hardcoded the two variants of this button, but I thought this was cleaner.

src/Apps/ArtQuiz/Components/ArtQuizButton.tsx
damassi: might be a bit early here (not much use elsewhere) but could see this being easily generalized into a simple `useAnimation` hook that accepts some keyframes

dzucconi:
Next with 6d560b3f23c38ddef2b23ba792615650e28174fe I've added the card animations. I implemented this in the simplest way possible to wrap my head around it but this might change a bit more depending on how the mobile implementation goes and if any optimizations are needed once we have real images.

Here's what it looks like:

https://user-images.githubusercontent.com/112297/207728553-4d1436bd-8160-4da0-80d1-c16534f2512a.mov

------

* I wound up storing the like/dislike state using `useReducer` but depending on how we go about this with Relay, it might not be needed at all. So I'm expecting that to get deleted.
* To do the animations I'm just stacking every 'card' and applying one of 4 modes to them: `Pending | Active | Liked | Disliked`, those modes map to a set of transforms and we transition the transforms. Part of the reason this is done this way is because you could potentially move very quickly through the stack. We could prune a render window around some number of the elements as an optimization.
* I think I'll have to handle the animations differently depending on how the mobile swipe feels. I might want to convert them to use the Web Animation API so that you could "scrub" thru them depending on where you are in mid-swipe. This might be in addition or it might replace the current implementation. The naive implementation, which I'll do first, is just to check which direction you swiped and run the appropriate action and corresponding transition. We'll see what feels good.
* I'm noticing a very weird bug where the 'fill' state of the heart will sometimes create a moment where clicks are rejected. I'm not sure what's causing that yet but I think it's something to do with the render interfering with the click event. Like it gets removed between mouse down and up since there's a render that slips in between. Will investigate further tomorrow.

src/Apps/ArtQuiz/Components/ArtQuizCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+    index: 0,
+  })
+
+  return {
+    activeCard: state.cards[state.index] as T & { mode: Mode },
+    activeIndex: state.index,
+    cards: state.cards as (T & { mode: Mode })[],
+    dispatch,
+  }
+}
```
dzucconi, line 131: As I mentioned, I'm expecting this reducer state to just get deleted. I think we'll only need to keep track of the active index and will just use the like/dislike state on the `QuizArtworkEdge`s.

src/Apps/ArtQuiz/Components/ArtQuizMain.tsx
```diff
@@ -80,27 +90,47 @@ export const ArtQuizMain: FC<ArtQuizMainProps> = () => {
...
+          justifyContent="center"
+          alignItems="center"
+          flex={1}
+          flexDirection="column"
+        >
+          {cards.map((card, i) => {
+            const mode = i === activeIndex ? Mode.Active : card.mode
+
+            return (
+              <ArtQuizCard key={card.id} mode={mode} position="absolute">
```
dzucconi, line 104: I'll want to replace these with sample images (including retina sized srcSets) to see if the performance changes.

dzucconi:
Alright: added a pretty naive swipe implementation but it's maybe simple enough to work. Let's see how it feels on a real device:

https://user-images.githubusercontent.com/112297/207929621-861f9153-39dd-4f49-8687-831e2b61c7a3.mov

-----

One thing I've overlooked for the time being is triggering the button animations on swipe; I'll follow up with that.

src/Apps/ArtQuiz/Hooks/useSwipe.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    document.addEventListener("touchstart", handleTouchStart)
+    document.addEventListener("touchend", handleTouchEnd)
+
+    return () => {
+      document.removeEventListener("touchstart", handleTouchStart)
+      document.removeEventListener("touchend", handleTouchEnd)
+    }
+  }, [distance, onLeft, onRight])
+}
```
damassi, line 46: Should we move this to `Utils/Hooks`? Or again too early

dzucconi: I think too early. We may not wind up using it. We'll see. I'll move it out if it proves useful.

PR #11569, opened 2022-12-15
Revert "fix:prevent overlapping text on firefox"

leamotta:
Reverts artsy/force#11498

The original fix breaks the column arrangement in Chrome:
Ricardo Fernandes (1)

While not covering all cases in Firefox:
Screenshot 2022-12-15 at 14 45 31

Reverting the fix so we can return the proper display in Chrome while we figure out a cross-browser solution


nspinazz89 (approved):
LGTM 🚀

PR #11598, opened 2022-12-19
fix: differentiates city by displaying regions

dzucconi:
Closes: [GRO-720](https://artsyproduct.atlassian.net/browse/GRO-720)


anandaroop (commented):
🙏🏽

PR #11599, opened 2022-12-19
feat(art-quiz): connect quiz artworks screen to MP

laurabeth:

This PR solves [GRO-1435] [GRO-1257]

### Description

- Add a `startingIndex` param to `useArtQuizCards`. This allows us to check a users interacted quiz artworks and restart the quiz where they left off.
- Add `useDislikeArtworks` to allow us to capture negative signals from the quiz
- Lift `ArtQuizMain` component contents into `ArtQuizArtworks` to avoid prop drilling
- Add save and dislike mutations to click handlers
- Add top level mutations into `artQuizRoutes` 

[GRO-1435]: https://artsyproduct.atlassian.net/browse/GRO-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRO-1257]: https://artsyproduct.atlassian.net/browse/GRO-1257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/ArtQuiz/Components/ArtQuizCard.tsx
```diff
@@ -113,13 +113,20 @@ const reducer = <T,>(state: State<T>, action: Action) => {
 }
 
 interface UseArtQuizCards<T> {
+  //** An array of card objects */
   cards: T[]
+  findIndexCallback?: (card: T) => boolean
```
laurabeth, line 118: Realize I should switch this name to be more descriptive in the context of this component. Will also add useful pragma.

src/Apps/ArtQuiz/Components/ArtQuizMain.tsx
```diff
@@ -13,21 +15,46 @@ import {
...
 
   const { cards, activeIndex, dispatch } = useArtQuizCards({
-    cards: EXAMPLE_CARDS,
+    cards: quiz.quizArtworkConnection!.edges!.map(edge => ({
+      ...edge!.node,
+      interactedAt: edge!.interactedAt,
+    })),
+    findIndexCallback: card => {
+      if ("interactedAt" in card) {
+        console.log("#################### interactedAt", card.interactedAt)
```
laurabeth, line 39: Will remove. 🤡

src/Apps/ArtQuiz/Hooks/DislikeArtworkMutation.ts
```diff
@@ -0,0 +1,35 @@
+import { Environment, commitMutation, graphql } from "react-relay"
+import {
+  DislikeArtworkMutation,
+  DislikeArtworkInput,
+  DislikeArtworkMutation$data,
+} from "__generated__/DislikeArtworkMutation.graphql"
+
+export const DislikeArtwork = (
```
damassi, line 8: Can use our `useMutation` hook ([here's an example](https://github.com/artsy/force/blob/main/src/Apps/Settings/Routes/DeleteAccount/useDeleteAccount.ts) of how to use)

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -1,13 +1,49 @@
...
-      <MetaTags title="Art Taste Quiz | Artsy" />
+      <MetaTags />
 
-      <ArtQuizMain />
+      <ArtQuizMain quiz={quiz} />
     </>
   )
 }
+
+export const ArtQuizArtworksFragmentContainer = createFragmentContainer(
```
laurabeth, line 21: This isn't correct. Will move query contents down a level.

src/Apps/ArtQuiz/Hooks/DislikeArtworkMutation.ts
laurabeth: Oh nice. Will take a look.

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -16,12 +19,12 @@ const ArtQuizResults = loadable(() => import("./Routes/ArtQuizResults"), {
 export const artQuizRoutes: AppRouteConfig[] = [
   {
     path: "/art-quiz",
+    displayFullPage: true,
```
dzucconi, line 22: I actually don't think this particular flag _does_ anything.

dzucconi (commented):
This is looking good. I'm pleasantly surprised at how painless this integration looks!

We talked about this offline but moving the query root from `User` to `Me` will get rid of the need for the `prepareVariables` work.

src/Apps/ArtQuiz/Components/ArtQuizCard.tsx
dzucconi: What about just accepting an `initialIndex` prop and calculating this outside in the component that calls the hook?

laurabeth: _Too easy @dzucconi_. 😝

dzucconi: :) but yeah there's precedent for that as a prop. [Our older carousel and swiper for instance.](https://github.com/search?q=repo%3Aartsy%2Fpalette%20initialIndex&type=code)

laurabeth: No, I know... I making fun of myself. I fixed it.

src/Apps/ArtQuiz/Hooks/useDislikeArtwork.ts
```diff
@@ -0,0 +1,20 @@
...
+export const useDislikeArtwork = () => {
+  return useMutation<useDeleteArtworkMutation>({
+    mutation: graphql`
+      mutation useDislikeArtworkMutation($input: DislikeArtworkInput!)
+        @raw_response_type {
+        dislikeArtwork(input: $input) {
+          artwork {
+            id
+            slug
+            is_disliked: isDisliked
```
dzucconi, line 14: We don't need to alias this. Assuming the old one was aliased for some unrelated compatibility issue that's no longer a problem.

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -1,13 +1,213 @@
...
+import { useRouter } from "System/Router/useRouter"
+
+interface ArtQuizArtworksProps {
+  quiz: ArtQuizArtworks_quiz$data
+}
+
+export const ArtQuizArtworks: FC<ArtQuizArtworksProps> = ({ quiz }) => {
+  const { router } = useRouter()
+  const resume = quiz.quizArtworkConnection!.edges!.findIndex(edge => {
+    if (edge && "interactedAt" in edge) {
```
dzucconi, line 33: Wondering why we need the type guard here?

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -1,13 +1,213 @@
...
+      return edge.interactedAt === null
+    }
+    return false
```
dzucconi, line 36: You should be able to write this as a single condition rather than an if statement

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -1,13 +1,213 @@
...
+    startingIndex: resume || 0,
+  })
+  const currentArtwork = quiz.quizArtworkConnection!.edges![activeIndex]!.node!
+
+  const mutationProps = {
+    artwork: currentArtwork,
+    contextModule: "artQuiz" as AuthContextModule,
+  }
+  const { handleSave } = useSaveArtwork({
+    ...mutationProps,
```
dzucconi, line 53: Why not just inline `mutationProps` — doesn't look like it's used elsewhere?

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -21,12 +22,11 @@ const ArtQuizResults = loadable(() => import("./Routes/ArtQuizResults"), {
 export const artQuizRoutes: AppRouteConfig[] = [
   {
     path: "/art-quiz",
+    hideFooter: true,
```
dzucconi, line 25: `non-blocking`: I don't believe this does anything. Ideally it would apply to all child routes but I'm pretty sure it's just ignored.

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -38,18 +38,68 @@ export const artQuizRoutes: AppRouteConfig[] = [
...
+          const { me } = props as any
+
+          if (!me?.quiz) {
+            return
+          }
+          const { completedAt } = me.quiz
+
+          if (completedAt) {
+            throw new RedirectException("/art-quiz/results")
+          }
```
dzucconi, line 65: 👍

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -38,18 +38,68 @@ export const artQuizRoutes: AppRouteConfig[] = [
...
+          if (!(Component && props)) {
+            return
+          }
+          const { me } = props as any
+
+          if (!me?.quiz) {
+            return
+          }
+
+          return <Component {...props} quiz={me.quiz} />
```
dzucconi, line 96: `blocking`: We should just follow how we do this elsewhere and the route component should have a fragment on `Me` rather than us pulling it off in this render function.

dzucconi (commented):
Looks good. Good to merge after addressing that one inconsistency on the route fragment.

src/Apps/ArtQuiz/artQuizRoutes.tsx
laurabeth: It definitely does do something.

dzucconi: What does it do then if we have it also on the child routes?

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
laurabeth: fixed

src/Apps/ArtQuiz/Hooks/useDislikeArtwork.ts
laurabeth: removed

src/Apps/ArtQuiz/artQuizRoutes.tsx
laurabeth: rabble rabble

PR #11605, opened 2022-12-19
feat(GRO-1481): fully responsive, 100% width/height scaling box

dzucconi:
Closes: [GRO-1481](https://artsyproduct.atlassian.net/browse/GRO-1481)

---- 

<del>**NOTE**: @laurabeth Take a look at 29881d00f398840ab1c46da463fa0bf1b754abba I'll just roll off that commit; just wanted to show an example of the integration, minimal as that is. I don't want to create any conflicts with your open PR.</del>

----

It's surprising that this winds up being an edge case, because you'd think this would show up more often, but it's really only in the case of some kind of full-screen/lightbox -like experience. The only place we do this currently is on article zooms. Artwork zooms just use our deep zoom pyramids.

So what this does under the hood is use our `useSizeAndPosition` hook to track the geometry of the container. In this case it's a 100% width and 100% height box, placed somewhere on the page. Then use those values to proportionally scale a child box. So given the aspect width and height; it'll just fit it inside. What's nice about this solution is that it doesn't pollute the parent component with any concerns about resize events or measuring. We just pass it our image width and height, and stick an image in it, as if it were a normal `ResponsiveBox`.

It really feels to me like there's a big gap in the browser implementation how `aspect-ratio` is used inside of a flex context. Like I want `object-fit: contain` for non-[replaced elements](https://developer.mozilla.org/en-US/docs/Web/CSS/Replaced_element). And maybe there is a way to do this with just plain CSS. If so I'd love to know how!


src/Components/FullyResponsiveBox.tsx
```diff
@@ -3,17 +3,18 @@ import { FC, useEffect, useState } from "react"
 import { useSizeAndPosition } from "Utils/Hooks/useSizeAndPosition"
 import { scale } from "proportional-scale"
 
-interface ArticleZoomGalleryResponsiveBoxProps extends BoxProps {
+export interface FullyResponsiveBoxProps extends BoxProps {
```
damassi, line 6: Lol @dzucconi having "ResponsiveBox" and "FullyResponsiveBox" is clearly quite confusing

dzucconi: How about `FullscreenBox`

dzucconi: Sure. This is a totally different use-case than a normal `ResponsiveBox` (and a pretty rare one). Good call out. `FullscreenBox` makes sense to me. Any objections or better ideas?

dzucconi: `ResponsiveBox` should probably be `AspectRatioBox`...

`FullscreenAspectRatioBox`?

damassi: FullscreenBox 👍 that makes perfect sense

laurabeth (dismissed):
Looks good once naming convention is sorted.

src/Components/FullyResponsiveBox.tsx
dzucconi: Images, not necessarily: like it could be a video or iframe. But the function of it is: scale up in both dimensions, maintaining aspect ratio.

dzucconi:
Went with `FullscreenAspectRatioBox`. Merging on green.

dzucconi:
[Scratch that](https://artsy.slack.com/archives/C01ADJNCS5D/p1671561907606629), `FullscreenBox` it is!

PR #11607, opened 2022-12-20
fix(NX-3342): artwork detail wasn't closable on medium-size screens

lean8086:

This PR solves [NX-3342]

### Description

- There was a mismatch between:
    - when we showed the sidebar positioned as absolute, and
    - when we showed the sidebar header to make it closable.
- The reason seems to be that using different values on the attributes (i.e.: `display={["none", "none", "flex", "flex"]`) wasn't applying the changes at the same widths as `breakpoints`. We use breakpoints to change the `position` of the sidebar ([see here](https://github.com/artsy/force/blob/main/src/Apps/Conversation/Components/DetailsSidebar.tsx#L34-L68))
- I made sure either DetailsHeader or DetailIcon is always visible (both allow you to close the sidebar)
- For that, I rewrote those attributes' media queries with useMatchMedia (still client-side queries), to match:
    - when the sidebar positions `absolute` covering the Details button,
    - then display the sidebar header with the X button
    - and vice versa

**Before**
image

**After**
| | xs / sm | md / lg | xl |
| --- | --- | --- | --- |
| Closed | image | image | image |
| Open | image | image | image |

cc @artsy/negotiate-devs 

[NX-3342]: https://artsyproduct.atlassian.net/browse/NX-3342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Conversation/Components/ConversationHeader.tsx
+++ b/src/Apps/Conversation/Components/ConversationHeader.tsx
@@ -91,7 +91,22 @@ const LargeConversationHeader: FC<ConversationHeaderProps> = props => {
             <>{props.children}</>
           )}
         </Box>
-        <DetailIcon showDetails={showDetails} setShowDetails={setShowDetails} />
+        <Media lessThan="xl">
+          <DetailIcon
+            showDetails={showDetails}
+            setShowDetails={setShowDetails}
+          />
+        </Media>
+        {/* When opened, DetailsSidebar is positioned on top of the messages (on XL
+        layouts only). In those cases, DetailsHeader is redundant with DetailIcon */}
+        <Media at="xl">
+          {!showDetails && (
+            <DetailIcon
+              showDetails={showDetails}
+              setShowDetails={setShowDetails}
+            />
+          )}
+        </Media>
       </Flex>
       <Separator />
     </Flex>
--- a/src/Apps/Conversation/Components/DetailsHeader.tsx
+++ b/src/Apps/Conversation/Components/DetailsHeader.tsx
@@ -8,9 +8,11 @@ import {
   Box,
   InfoCircleIcon,
   Clickable,
+  themeProps,
 } from "@artsy/palette"
 
 import { LARGE_SCREEN_HEADER_HEIGHT } from "./ConversationHeader"
+import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
 
 export interface DetailsProps {
   showDetails: boolean
@@ -21,15 +23,23 @@ interface DetailsHeaderProps extends DetailsProps {}
 
 export const DetailsHeader: FC<DetailsHeaderProps> = props => {
   const { showDetails, setShowDetails } = props
+  const isMobile =
+    __internal__useMatchMedia(themeProps.mediaQueries.xs) ||
+    __internal__useMatchMedia(themeProps.mediaQueries.sm)
+
+  if (isMobile) {
+    return null
+  }
+
   return (
     <StackableBorderBox
       p={0}
       flexDirection="column"
       width={showDetails ? "376px" : "0"}
       maxWidth={showDetails ? "auto" : "0"}
-      display={["none", "none", "flex", "flex", "flex"]}
       {...props}
       borderTop="none !important"
+      borderBottom="none !important"
     >
       <Flex
         flexDirection="row"
@@ -54,8 +64,16 @@ export const DetailsHeader: FC<DetailsHeaderProps> = props => {
 
 export const DetailIcon: React.FC<DetailsProps> = props => {
   const { showDetails, setShowDetails } = props
+  const isXL = __internal__useMatchMedia(themeProps.mediaQueries.xl)
+
+  // When opened, DetailsSidebar is positioned on top of the messages (on XL
+  // layouts only). In those cases, DetailsHeader is redundant with DetailIcon
+  if (showDetails && isXL) {
+    return null
+  }
+
   return (
-    <Box display={["initial", "initial", showDetails ? "none" : "inline"]}>
+    <Box>
       <Flex flexDirection="row" alignItems="center" pr={1}>
         <InfoCircleIcon />
         <Clickable cursor="pointer" textDecoration="underline">
--- a/src/Apps/Conversation/Components/DetailsHeader.tsx
+++ b/src/Apps/Conversation/Components/DetailsHeader.tsx
@@ -39,7 +39,6 @@ export const DetailsHeader: FC<DetailsHeaderProps> = props => {
       maxWidth={showDetails ? "auto" : "0"}
       {...props}
       borderTop="none !important"
-      borderBottom="none !important"
     >
       <Flex
         flexDirection="row"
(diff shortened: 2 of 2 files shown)


src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -21,15 +23,23 @@ interface DetailsHeaderProps extends DetailsProps {}
 
 export const DetailsHeader: FC<DetailsHeaderProps> = props => {
   const { showDetails, setShowDetails } = props
+  const isMobile =
+    __internal__useMatchMedia(themeProps.mediaQueries.xs) ||
+    __internal__useMatchMedia(themeProps.mediaQueries.sm)
+
+  if (isMobile) {
```
araujobarret, line 30: that's kinda okayish but I think of another way to use the responsive props, for example:
```jsx
return (
   <StackableBorderBox display={["none", "inherit"]}
   ...
```
is it possible for this case?

araujobarret: I know you wrote in the PR 
```
The reason seems to be that using different values on the attributes (i.e.: display={["none", "none", "flex", "flex"] wasn't applying the changes at the same widths as breakpoints.
```
But wanted to double check if you've tried here

araujobarret: nvm gh display messed up when I was taking a look here, I see it was removed 🤦

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -21,15 +23,23 @@ interface DetailsHeaderProps extends DetailsProps {}
...
   return (
     <StackableBorderBox
       p={0}
       flexDirection="column"
       width={showDetails ? "376px" : "0"}
       maxWidth={showDetails ? "auto" : "0"}
-      display={["none", "none", "flex", "flex", "flex"]}
       {...props}
       borderTop="none !important"
+      borderBottom="none !important"
```
araujobarret, line 42: I wonder if we have to do this if the `StackableBorderBox` props are spread/used correctly on Palette

tam-kis: Have you tried `borderBottom: "0px"` ?

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -21,13 +23,20 @@ interface DetailsHeaderProps extends DetailsProps {}
 
 export const DetailsHeader: FC<DetailsHeaderProps> = props => {
   const { showDetails, setShowDetails } = props
+  const isMobile =
+    __internal__useMatchMedia(themeProps.mediaQueries.xs) ||
+    __internal__useMatchMedia(themeProps.mediaQueries.sm)
```
damassi, line 28: Both of these use `window` internally which might blow up SSR rendering; can you give that a test? 

Fine with using that here tho `__internal__` was meant to mean: don't use, its only for other internal implementations! But i can see it being useful occasionally (if its not breaking SSR rendering)

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -54,8 +63,16 @@ export const DetailsHeader: FC<DetailsHeaderProps> = props => {
 
 export const DetailIcon: React.FC<DetailsProps> = props => {
   const { showDetails, setShowDetails } = props
+  const isXL = __internal__useMatchMedia(themeProps.mediaQueries.xl)
```
damassi, line 66: Dito here - does this break SSR rendering?

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -54,8 +63,16 @@ export const DetailsHeader: FC<DetailsHeaderProps> = props => {
...
+
+  // When opened, DetailsSidebar is positioned on top of the messages (on XL
+  // layouts only). In those cases, DetailsHeader is redundant with DetailIcon
+  if (showDetails && isXL) {
```
damassi, line 70: I wish we had easy access to breakpoint props on the JS code side, but alas 😢 

One thing we can do to get around this (which sucks a bit DX wise) is:

```tsx
<Media at="xs">{renderXS()}</Media>
<Media at="greaterThanXS">{renderGreaterThanXS()}</Media>
```

src/Apps/Conversation/Components/DetailsHeader.tsx
lean8086: I compared SSR on both (my changes locally and the current version on staging) and the behavior is the same. Though, I'm not sure if that's an accurate test because the sidebar only opens client-side 🤔

src/Apps/Conversation/Components/DetailsHeader.tsx
lean8086: Using Media sounds good. I'll give it a try now!

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -22,40 +23,41 @@ interface DetailsHeaderProps extends DetailsProps {}
...
-      {...props}
-      borderTop="none !important"
-    >
-      <Flex
-        flexDirection="row"
-        height={LARGE_SCREEN_HEADER_HEIGHT}
-        justifyContent="space-between"
-        alignItems="flex-end"
-        pb={1}
+    <Media greaterThan="sm">
```
lean8086, line 26: One of the 2 changes in this file is wrapping the component in this Media.

src/Apps/Conversation/Components/DetailsHeader.tsx
```diff
@@ -22,40 +23,41 @@ interface DetailsHeaderProps extends DetailsProps {}
 export const DetailsHeader: FC<DetailsHeaderProps> = props => {
   const { showDetails, setShowDetails } = props
   return (
-    <StackableBorderBox
-      p={0}
-      flexDirection="column"
-      width={showDetails ? "376px" : "0"}
-      maxWidth={showDetails ? "auto" : "0"}
-      display={["none", "none", "flex", "flex", "flex"]}
```
lean8086, line 30: Removing this prop was one of the 2 changes in this file.

src/Apps/Conversation/Components/DetailsHeader.tsx
damassi: If the sidebar is only open on the client, then no prob! Just had to call it out 🙂  

In any case, the updated implementation with Media makes sense 👍

damassi (approved):

src/Apps/Conversation/Components/ConversationHeader.tsx
```diff
@@ -91,7 +91,22 @@ const LargeConversationHeader: FC<ConversationHeaderProps> = props => {
...
+        {/* When opened, DetailsSidebar is positioned on top of the messages (on XL
+        layouts only). In those cases, DetailsHeader is redundant with DetailIcon */}
+        <Media at="xl">
+          {!showDetails && (
+            <DetailIcon
+              showDetails={showDetails}
+              setShowDetails={setShowDetails}
+            />
+          )}
+        </Media>
```
lean8086, line 109: I wish there was another way of rendering the DetailIcon always except on `showDetails && screenXL`

PR #11618, opened 2022-12-20
fix(art-quiz): fixes fragment access; uses fullscreen box

dzucconi:
This PR does two things.

* The `/art-quiz/artworks` was erroring due to the fragment name property not being updated
* Added the `FullscreenBox` usage and updated the images to work with it.

### DIFF
--- a/src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
+++ b/src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
@@ -3,7 +3,6 @@ import { ArtQuizArtworks_me$data } from "__generated__/ArtQuizArtworks_me.graphq
 import { AuthContextModule } from "@artsy/cohesion"
 import {
   ArrowLeftIcon,
-  Box,
   Clickable,
   Image,
   Flex,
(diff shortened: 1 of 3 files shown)


src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -22,26 +21,29 @@ import { useSaveArtwork } from "Components/Artwork/SaveButton/useSaveArtwork"
...
 
 interface ArtQuizArtworksProps {
   me: ArtQuizArtworks_me$data
 }
 
 export const ArtQuizArtworks: FC<ArtQuizArtworksProps> = ({ me }) => {
   const { router } = useRouter()
-  const resume = me.quiz!.quizArtworkConnection!.edges!.findIndex(edge => {
+
+  const edges = me.quiz.quizArtworkConnection!.edges!
```
dzucconi, line 33: I *reeeeeeaaaally* want to fix these connection types to be non-nullable. I guess we could also have a helper type to cast them that way as a temporary hack...

src/Apps/ArtQuiz/Routes/ArtQuizArtworks.tsx
```diff
@@ -180,7 +193,7 @@ export const ArtQuizArtworks: FC<ArtQuizArtworksProps> = ({ me }) => {
 export const ArtQuizArtworksFragmentContainer = createFragmentContainer(
   ArtQuizArtworks,
   {
-    quiz: graphql`
+    me: graphql`
```
dzucconi, line 196: 💥

PR #11624, opened 2022-12-21
fix(art-quiz): temporarily fixes height

dzucconi:
I wanted to just get this back to a working state while I take a look at the `AppShell` today. I'm thinking a smarter direction would be to just support the notion of "layouts" and have a few available, instead of a concoction of boolean flags like `hideNav`, `hideFooter` etc. Would be easier to understand and maintain. But in the meantime...


laurabeth (approved):
🤦🏽‍♀️ Again...? Anyway, thanks for the quick fix.

PR #11628, opened 2022-12-22
fix: Hide collected artists when not existing in upload flow

olerichter00:
## Description

Hide the collected artists section when the user hasn't collected any artists yet in the artwork upload flow (according to the Figma design).


MounirDhahri (approved):
good catch

PR #11630, opened 2022-12-22
feat: ignore header offset when `isEigen` is true

dimatretyak:

Review app: [ignore-header-offset](https://ignore-header-offset.artsy.net/)

### Description
Ignore header offset in sticky when `isEigen` is true, since the nav bar menu [will not be rendered](https://github.com/artsy/force/blob/aa217617287577dff6e9c2b2710df63fdcd195f0/src/Components/NavBar/NavBar.tsx#L58-L60)

#### Before
https://user-images.githubusercontent.com/3513494/209139712-b7888df4-5338-4897-a807-46c0422d4e69.mp4

#### After
https://user-images.githubusercontent.com/3513494/209139795-d34d6a14-da3e-44e3-afaf-a7f91514f1be.mp4

### DIFF
--- a/src/Components/Sticky/Sticky.tsx
+++ b/src/Components/Sticky/Sticky.tsx
@@ -4,7 +4,6 @@ import ReactSticky, { Props as ReactStickyProps } from "react-stickynode"
 import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
 import { useNavBarHeight } from "Components/NavBar/useNavBarHeight"
 import { useSticky } from "./StickyProvider"
-import { useSystemContext } from "System"
 
 /**
  * Wrap a component to have it stick below the main nav.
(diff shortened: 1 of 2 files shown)


src/Components/Sticky/Sticky.tsx
```diff
@@ -32,17 +33,23 @@ export const Sticky: React.FC<
...
   const { desktop, mobile } = useNavBarHeight()
+  const { isEigen } = useSystemContext()
 
   const isMobile = __internal__useMatchMedia(themeProps.mediaQueries.xs)
 
   const [stuck, setStuck] = useState(false)
 
   const containerRef = useRef<HTMLDivElement | null>(null)
 
+  const withoutHeaderOffset = _withoutHeaderOffset ?? isEigen
```
dimatretyak, line 52: or is it better to place this logic in `useNavBarHeight'? Something like this

```javascript
const useNavBarHeight = () => {
  const { isLoggedIn, isEigen } = useSystemContext()

  let mobile = isLoggedIn ? MOBILE_LOGGED_IN_NAV_HEIGHT : MOBILE_NAV_HEIGHT
  let desktop = DESKTOP_NAV_BAR_HEIGHT

  if (isEigen) {
    mobile = 0
    desktop = 0
  }

  return { height: [mobile, desktop], mobile, desktop }
}
```

dzucconi: Yeah, this winds up getting used in a few places so it'd make sense to move it to `useNavBarHeight`. I'm surprised it took this long for this to surface!

dimatretyak: Done

dzucconi (approved):

PR #11638, opened 2022-12-22
feat: app layouts

dzucconi:
I don't particularly feel like merging a site wide change like this before the holiday break but please leave some feedback.

The problem I noticed was that the layout in `AppShell` is difficult to work with. Sometimes hiding footer/navigation is confusing. There's a `displayFullPage` flag that is pointless. And there's a need for alternate layouts that were achieved in hacky ways (see the Order app, for instance).

So this just removes those flags and introduces a few different layouts which can be opted into at the route level. They are simple to understand. I think the naming could use some work but 🤷 .

## Layouts

* `Default`: current layout with fixed nav and footer. Contained by app container.
* `NavOnly`: Same as default but without the footer.
* `ContainerOnly`: Same as default but without the nav and without the footer.
* `Blank`: Completely empty layout
* `LogoOnly`: App container + our logotype. Used in the error pages. Should probably be used by the order app and various pages on the my collection app.

### DIFF
--- a/src/System/Router/Route.tsx
+++ b/src/System/Router/Route.tsx
@@ -1,35 +1,38 @@
 import { RouteSpinner } from "System/Relay/renderWithLoadProgress"
 import { RouteConfig, HttpError, Match } from "found"
 import BaseRoute from "found/Route"
-import * as React from "react";
-import { CacheConfig, GraphQLTaggedNode } from "relay-runtime";
-import { ArtsyRequest, ArtsyResponse } from "Server/middleware/artsyExpress";
-import { NextFunction } from "express";
+import * as React from "react"
+import { CacheConfig, GraphQLTaggedNode } from "relay-runtime"
+import { ArtsyRequest, ArtsyResponse } from "Server/middleware/artsyExpress"
+import { NextFunction } from "express"
+import { LayoutVariant } from "Apps/Components/Layouts"
+import { RemoveIndex } from "Utils/typeSupport"
 
 interface RouteConfigProps extends RouteConfig {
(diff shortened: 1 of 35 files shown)


damassi (approved):
💯 I don't see anything controversial in here that should prevent a merge. Its just way way better and also fundamentally simpler / straightforward. 

I'd add a small entry in the best_practices doc about it and then 👍 on merging if you feel comfortable (the risk seems low).

damassi:
Also maybe update the PR description with a list of all the new layouts so we can get a quick glance

dzucconi:
@damassi updated description with a list of the layouts.

src/System/Router/Route.tsx
```diff
@@ -1,35 +1,38 @@
...
   children?: AppRouteConfig[]
-  displayFullPage?: boolean
   fetchIndicator?: FetchIndicator
-  hideNav?: boolean
-  hideFooter?: boolean
+  /** FIXME: Remove. Avoid poluting global route config with application specific concerns */
   hideNavigationTabs?: boolean
   ignoreScrollBehavior?: boolean
   ignoreScrollBehaviorBetweenChildren?: boolean
+  layout?: LayoutVariant
```
damassi, line 19: Should this be required?

dzucconi: No, just defaults to the default layout.

damassi: ehh scratch that, fallback is better

dblandin (approved):
Looks fantastic 💯

PR #11642, opened 2022-12-23
fix(NX-3408): Set the iframe to use the total width

lidimayra:

This PR solves [NX-3408]

### Description
After updating the deprecated `Modal` component to `ModalDialog` (#11301), the iframe being rendered inside the modal had its width affected.

Changes here ensure the iframe will use 100% of the available width when the modal is triggered (this seemed to be the usual behaviour when we were using the `Modal` component).

Is this the best way to go here? As I understand we don't use this component today in any other spot besides the conversation.

## Before
before

## After
after

cc @artsy/negotiate-devs 

[NX-3408]: https://artsyproduct.atlassian.net/browse/NX-3408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Conversation/Components/OrderModal.tsx
+++ b/src/Apps/Conversation/Components/OrderModal.tsx
@@ -4,6 +4,7 @@ import styled from "styled-components"
 
 export const StyledIframe = styled("iframe")`
   height: 60vh;
+  width: 100%;
   border: none;
 `
 


jo-rs (approved):
Thank you for fixing, Lidi! 🏅 
Is everything looking ok on mobile web?

lidimayra:
Thank you!! 🙌 💜 

And good call, I checked now on mobile, and it looks alright (the columns get split into rows):
Screenshot 2022-12-23 at 14 39 21

Screenshot 2022-12-23 at 14 39 31

lean8086 (approved):

PR #11657, opened 2022-12-29
fix: Add space to High Demand icon on artwork grid

olerichter00:
## Description

Add space to High Demand icon on artwork grid.

| Before | After |
| --- | --- |
|Screenshot 2022-12-29 at 15 12 04 |Screenshot 2022-12-29 at 15 11 21 |

### DIFF
--- a/src/Components/Artwork/Details.tsx
+++ b/src/Components/Artwork/Details.tsx
@@ -140,7 +140,7 @@ const HighDemandInfo = () => {
     <Flex flexDirection="row" alignItems="center">
       <HighDemandIcon width={16} height={16} />
       <Text variant="xs" color="blue100" ml={0.3}>
-        High Demand
+        &nbsp;High Demand
       </Text>
     </Flex>
   )


src/Components/Artwork/Details.tsx
```diff
@@ -140,7 +140,7 @@ const HighDemandInfo = () => {
     <Flex flexDirection="row" alignItems="center">
       <HighDemandIcon width={16} height={16} />
       <Text variant="xs" color="blue100" ml={0.3}>
-        High Demand
+        &nbsp;High Demand
```
MrSltun, line 143: **Suggestion:** should we use `ml` instead? 👀

MrSltun (approved):
Looks great! 🌟 I just left a small suggestion, but feel free to merge without it :)

src/Components/Artwork/Details.tsx
olerichter00: I've had this discussion before and someone suggested using the `&nbsp;` instead because it's not really meant to be a spacing between components.

PR #11674, opened 2023-01-04
chore: show existing artwork images from local store

MounirDhahri:

### Description
This PR resolves [CX-3296]

https://user-images.githubusercontent.com/11945712/210832556-fd7a38a7-bb65-46e3-9680-7b35888afff0.mov

Images take a while to be first in queue in Gemini and it's why they take a while to be fully ready

[CX-3296]: https://artsyproduct.atlassian.net/browse/CX-3296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
@@ -19,8 +19,9 @@ export interface MyCollectionArtworkFormImagesProps {
   saveImagesToLocalStorage: (artworkId: string) => Promise<string | undefined>
 }
 export const MyCollectionArtworkFormImages = forwardRef<
-  MyCollectionArtworkFormImagesProps
->((_, formImagesRef) => {
+  MyCollectionArtworkFormImagesProps,
+  { isEditing?: boolean }
+>(({ isEditing = false }, formImagesRef) => {
   const [errors, setErrors] = useState<Array<FileRejection>>([])
   const [localImages, setlocalImages] = useState<
     Array<LocalImage & { photoID: string }>
@@ -48,6 +49,7 @@ export const MyCollectionArtworkFormImages = forwardRef<
         saveImagesToLocalStorage,
       }
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     [localImages, saveImagesToLocalStorage]
   )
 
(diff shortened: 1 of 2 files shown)


src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
```diff
@@ -125,6 +128,20 @@ export const MyCollectionArtworkFormImages = forwardRef<
     const imageAlreadyAdded = localImages.find(
       localImage => localImage.photoID === photoID
     )
+
+    // Handle images added through the select artwork step
+    if (!isEditing && currentSrc.includes("cloudfront.net")) {
```
MounirDhahri, line 133: I am using `isEditing` here to make this only happens when the user is adding an artwork for the first time to avoid duplicates

src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
```diff
@@ -19,8 +19,9 @@ export interface MyCollectionArtworkFormImagesProps {
   saveImagesToLocalStorage: (artworkId: string) => Promise<string | undefined>
 }
 export const MyCollectionArtworkFormImages = forwardRef<
-  MyCollectionArtworkFormImagesProps
->((_, formImagesRef) => {
+  MyCollectionArtworkFormImagesProps,
+  { isEditing?: boolean }
+>(({ isEditing = false }, formImagesRef) => {
```
olerichter00, line 24: **suggestion:** Is this prop really needed? When editing an artwork, there shouldn't be any images from already existing artworks, right? I'd suggest to remove this prop if it's not needed to reduce complexity.

src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
olerichter00, line 133: 👍

olerichter00 (approved):
Looks good. I've added a suggestion. Feel free to merge 🚀

src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormImages.tsx
MounirDhahri: my goal here was to separate between the following:
- images that are added from an existing artwork
- images that are added from the same artwork while editing it (they source also contains `cloudFront`)

Both images urls contain `cloudFront`

I wasn't able to separate between them without explicitly sending the flag

olerichter00: Oh, I understand 👍

PR #11681, opened 2023-01-04
feat: Separate Saves page in Collector Profile

MrSltun:

This PR solves [CX-3175]

### Description
This PR separates the Saves page from the Saves and Follows into it's own tab

### Screenshots

| | Desktop | mWeb |
|--|--|--|
| page | image | image |
| pagination | image | image |


src/Apps/CollectorProfile/Routes/Saves/CollectorProfileSavesRoute.tsx
```diff
@@ -0,0 +1,160 @@
...
+            }
+          }
+        }
+      }
+    `,
+  },
+  COLLECTOR_PROFILE_SAVES_QUERY
+)
+
+const COLLECTOR_PROFILE_SAVES_PLACEHOLDER = (
```
MounirDhahri, line 130: **thought:** is there a specific reason you used this new pattern instead of the way we usually use?
```suggestion
const CollectorProfileSavesPlaceholder = () => (
```

src/Apps/CollectorProfile/CollectorProfileApp.tsx
```diff
@@ -8,7 +8,8 @@ import { CollectorProfileApp_me$data } from "__generated__/CollectorProfileApp_m
 const TABS = [
   { name: "My Collection", url: "/collector-profile/my-collection" },
   { name: "Insights", url: "/collector-profile/insights" },
-  { name: "Saves & Follows", url: "/collector-profile/saves" },
+  { name: "Saves", url: "/collector-profile/saves" },
+  { name: "Follows", url: "/collector-profile/follows" },
```
MounirDhahri, line 12: **issue:** since we don't have the follows page yet, let's hide this change behind a separate feature flag and keep the saves & follows tab until then

src/Apps/CollectorProfile/Routes/Saves/CollectorProfileSavesRoute.tsx
MrSltun: I saw the pattern already in a few places in Force, that's why I used it, but I have no issue changing it :)

MounirDhahri: if that's the case then it's fine 🙌

MounirDhahri: no need to change it

dariakoko (approved):
Looks great, thanks Sultan!

PR #11682, opened 2023-01-05
fix(TX-988): correct phone number is committed for pickup orders

AlicanAkyuz:

This PR solves [TX-988]

### Description
This PR fixes the problem reported by Applause in the new app release. the problem is that when users select pick-up shipping option, type in their phone number, and they have a saved address, we are storing the phone number from the saved address instead of the typed-in number. 

#### Video 
https://user-images.githubusercontent.com/42584148/210779467-d95a41cf-e954-4ed9-8974-df3c51cb449f.mov

[TX-988]: https://artsyproduct.atlassian.net/browse/TX-988?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -262,10 +262,11 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                 )!
           )
 
-      const shipToPhoneNumber = isCreateNewAddress()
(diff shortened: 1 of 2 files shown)


kiryl-zubarau (approved):
LGTM!

PR #11689, opened 2023-01-05
feat(authdialog): loose ends

dzucconi:
This PR ties up a few loose ends. Next PR will include feature flagging code for enabling it in a single location.


src/Components/AuthDialog/Hooks/useAfterAuthenticationRedirect.ts
```diff
@@ -37,6 +36,17 @@ export const useAfterAuthenticationRedirect = () => {
       return
     }
 
+    // Validate that it's a safe URL to redirect to.
```
dzucconi, line 39: Replaces https://github.com/artsy/force/blob/main/src/Apps/Authentication/Server/redirectPostAuth.ts

damassi (approved):

PR #11690, opened 2023-01-05
feat(authentication): removes explicit terms checkbox

dzucconi:
This removes the terms checkbox as it's no longer explicitly needed. Will open an Integrity pull request before merging.

| GDPR | Non-GDPR |
| ---- | ---- |
|  |


dzucconi:
Merging.

PR #11698, opened 2023-01-09
fix: Hide artist selection skip button behind feature flag in MyC upload flow

olerichter00:
## Description

We need to hide the artist selection **skip** button behind when the feature flag `"cx-my-collection-personal-artists-for-web"` is disabled. This is needed in case we want to enable the new uploading flow without allowing users to create artists (just in case 😄 )


MounirDhahri (dismissed):
Niice! thanks

MrSltun (approved):
🌟

PR #11699, opened 2023-01-09
fix: Reduce number of columns in MyC Select Artist step

olerichter00:
## Description

Reduce number of columns in MyC Select Artist step.

### DIFF
--- a/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtistStep.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtistStep.tsx
@@ -124,7 +124,7 @@ export const MyCollectionArtworkFormArtistStep: React.FC<MyCollectionArtworkForm
           <Spacer y={1} />
           <GridColumns width="100%">
             {collectedArtists.map(artist => (
-              <Column span={[6, 4]} key={artist.internalID} mt={1}>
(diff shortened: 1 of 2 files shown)


src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtistStep.tsx
```diff
@@ -141,6 +141,8 @@ export const MyCollectionArtworkFormArtistStep: React.FC<MyCollectionArtworkForm
           </GridColumns>
         </>
       )}
+
+      <Spacer y={4} />
```
olerichter00, line 145: Adding some space here at the bottom of the page.

src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormHeader.tsx
```diff
@@ -51,6 +51,7 @@ export const MyCollectionArtworkFormHeader: React.FC<MyCollectionArtworkFormHead
                       flexDirection="row"
                       justifyContent="space-between"
                       py={2}
+                      minHeight={[70, 90]}
```
olerichter00, line 54: To make sure the height of the header is the same for each step (even without the skip button).

MounirDhahri (approved):

PR #11706, opened 2023-01-09
fix(CX-3275): Fix Phone Input Populated with wrong values in Consignment Inquiry Form

lordkiz:

This PR solves [CX-3275]

### Description
- When populating the form with the user's already saved phone number, we rely on the international and national formatted numbers gotten from Metaphysics. However these formats - national and international - when used to populate the form, we can end up with unintended final phone number values sent to the BE and we cannot handle this case by case (or country by country)
- This PR uses the raw phone number as is in gravity and replaces the country code in order to get the phone number we actually want to save as the phone value. 
More Context: https://artsy.slack.com/archives/C01B2P6LJUU/p1671712313793239

#### Before
Screenshot 2023-01-09 at 17 10 50

#### After
Screenshot 2023-01-09 at 17 13 04

[CX-3275]: https://artsyproduct.atlassian.net/browse/CX-3275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/ConsignmentInquiry/ConsignmentInquiry.tsx
+++ b/src/Apps/Consign/Routes/ConsignmentInquiry/ConsignmentInquiry.tsx
@@ -27,17 +27,22 @@ const logger = createLogger("ConsignmentInquiry/ConsignmentInquiry.tsx")
 
 const getContactInformationFormInitialValues = (
   me: ConsignmentInquiry_me$data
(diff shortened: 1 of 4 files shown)


olerichter00 (approved):
Thanks for the fix 🙏

PR #11718, opened 2023-01-10
chore(authdialog): enables auth dialog wrapper for intentful buttons

dzucconi:
Continuing the auth dialog rollout (currently at 25% in just the top-right auth links), this PR expands the surface to auth dialogs that open when you click actions that have after authentication intents (save, follow, etc).

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateAlertButton.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateAlertButton.tsx
@@ -10,7 +10,6 @@ import {
   SavedSearchEntityCriteria,
   SearchCriteriaAttributes,
 } from "Components/SavedSearchAlert/types"
-import { AuthModalOptions } from "Utils/openAuthModal"
 import { ArtworkSidebarCreateAlertButton_artwork$data } from "__generated__/ArtworkSidebarCreateAlertButton_artwork.graphql"
 
 interface ArtworkSidebarCreateAlertButtonProps {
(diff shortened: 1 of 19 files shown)


src/Components/FollowButton/FollowGeneButton.tsx
```diff
@@ -28,7 +29,7 @@ const FollowGeneButton: React.FC<FollowGeneButtonProps> = ({
   onFollow,
   ...rest
 }) => {
-  const { isLoggedIn, mediator } = useSystemContext()
```
damassi, line 31: ❇️

src/Components/SavedSearchAlert/Components/ActiveFilterPillsAndCreateAlert.tsx
```diff
@@ -60,7 +45,20 @@ export const ActiveFilterPillsAndCreateAlert: React.FC<ActiveFilterPillsAndCreat
         criteria={criteria}
         metric={metric}
         aggregations={aggregations}
-        getAuthModalOptions={getAuthModalOptions}
```
damassi, line 63: 👍

damassi (approved):

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCreateAlertButton.tsx
```diff
@@ -70,30 +69,25 @@ const ArtworkSidebarCreateAlertButton: FC<ArtworkSidebarCreateAlertButtonProps>
...
-        action: "createAlert",
-        kind: "artworks",
-        objectId: artwork.internalID,
-      },
-      contextModule: ContextModule.artworkSidebar,
-      intent: Intent.createAlert,
-      redirectTo: location.href,
-      copy: "Sign up to create your alert",
-    }
-  }
```
dzucconi, line 89: I think this stuff was wrapped in a function because of the call on `window.location` — that's a default anyway so removing that minor bit of indirection. I'm not really convinced this is a good way to pass these options anyway but it's fine for now. Can revisit when we retire the legacy dialog.

dzucconi: I think the _only_ thing you'd have to actually pass is the `contextModule` (which is what we do in a lot of these components). The rest can be inferred.

PR #11721, opened 2023-01-11
feat(art-quiz): limit results related works

laurabeth:

This PR solves [GRO-1482]

### Description

After a user takes the art quiz we want to show them works related to the ones they saved. Acceptance criteria for this task were explained as follows:

- If user likes only 1 artwork, show ALL related
- If user likes less than 3 artworks, show [8] per liked
- If user likes more than 3 artworks, show [4] per liked

This set of criteria for the related works results, however, does not apply to the quiz results email, which will have a smaller subset of related works. That being said, it didn't feel too out of line to simply pass a size argument to the pre-existing query.

Maybe worth revisiting a `results` query that encapsulates all results data at some point in the future.

[GRO-1482]: https://artsyproduct.atlassian.net/browse/GRO-1482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/ArtQuiz/Components/ArtQuizResultsRecommendedArtworks.tsx
+++ b/src/Apps/ArtQuiz/Components/ArtQuizResultsRecommendedArtworks.tsx
@@ -3,11 +3,10 @@ import { createFragmentContainer, graphql } from "react-relay"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
 import { ArtQuizResultsRecommendedArtworks_me$data } from "__generated__/ArtQuizResultsRecommendedArtworks_me.graphql"
 import { ArtQuizResultsRecommendedArtworksQuery } from "__generated__/ArtQuizResultsRecommendedArtworksQuery.graphql"
-import { extractNodes } from "Utils/extractNodes"
 import { Masonry } from "Components/Masonry"
 import ArtworkGridItemFragmentContainer from "Components/Artwork/GridItem"
 import { Spacer } from "@artsy/palette"
-import { shuffle, uniqBy } from "lodash"
+import { compact, shuffle, uniqBy } from "lodash"
 import { ArtworkGridPlaceholder } from "Components/ArtworkGrid"
 
 interface ArtQuizResultsRecommendedArtworksProps {
@@ -17,14 +16,16 @@ interface ArtQuizResultsRecommendedArtworksProps {
 const ArtQuizResultsRecommendedArtworks: FC<ArtQuizResultsRecommendedArtworksProps> = ({
   me,
 }) => {
-  const artworks = shuffle(
-    uniqBy(
-      me.quiz.savedArtworks.flatMap(artwork => {
-        if (!artwork.layer) return []
+  const artworks = compact(
+    shuffle(
+      uniqBy(
+        me.quiz.savedArtworks.flatMap(artwork => {
+          if (!artwork.layer) return []
 
-        return extractNodes(artwork.layer.artworksConnection)
-      }),
-      "internalID"
+          return artwork.related || []
+        }),
+        "internalID"
+      )
     )
   )
 
@@ -47,18 +48,22 @@ export const ArtQuizResultsRecommendedArtworksFragmentContainer = createFragment
   ArtQuizResultsRecommendedArtworks,
   {
     me: graphql`
-      fragment ArtQuizResultsRecommendedArtworks_me on Me {
+      fragment ArtQuizResultsRecommendedArtworks_me on Me
+        @argumentDefinitions(limit: { type: "Int" }) {
         quiz {
           savedArtworks {
             layer(id: "main") {
               artworksConnection {
                 edges {
                   node {
-                    ...GridItem_artwork
                     internalID
                   }
                 }
               }
+            }
+            related(size: $limit) {
+              ...GridItem_artwork
+              internalID
             }
           }
         }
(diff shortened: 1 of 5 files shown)


src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
```diff
@@ -22,6 +28,16 @@ export const ArtQuizResultsTabs: FC = () => {
     })
   }
 
+  const calculatePerArtworkRecommendationsLimit = useCallback(() => {
+    if (savedQuizArtworksCount <= 1) return 100
+    if (savedQuizArtworksCount <= 3) return 8
+    return 4
+  }, [savedQuizArtworksCount])
```
dzucconi, line 35: This doesn't need to be in a callback and can just be inlined in the memo

src/Apps/ArtQuiz/Routes/ArtQuizResults.tsx
```diff
@@ -11,7 +11,10 @@ interface ArtQuizResultsProps {
 }
 
 const ArtQuizResults: FC<ArtQuizResultsProps> = ({ me }) => {
-  const hasSavedArtworks = me.quiz.savedArtworks.length > 0
+  const savedQuizArtworksCount = useMemo(() => me.quiz.savedArtworks.length, [
```
dzucconi, line 14: You don't need to memoize accessing a length

src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
```diff
@@ -22,6 +28,16 @@ export const ArtQuizResultsTabs: FC = () => {
     })
   }
 
+  const calculatePerArtworkRecommendationsLimit = useCallback(() => {
+    if (savedQuizArtworksCount <= 1) return 100
+    if (savedQuizArtworksCount <= 3) return 8
+    return 4
```
dzucconi, line 34: 100, 8, or 4?

src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
laurabeth: ohhh... good call.

src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
laurabeth: 100 is no upper limit.

dzucconi: Ah OK. Might be helpful to inline this explanation in a comment:

```
If user likes only 1 artwork, show ALL related
1 liked = all related
If user likes less than 3 artworks, show [8] per liked
2 liked = 16 artworks
3 liked = 24 artworks
If user likes more than 3 artworks, show [4] per liked
4 liked = 16 artworks
5 liked = 20 artworks
….
16 liked = 64 artworks
```

src/Apps/ArtQuiz/Components/ArtQuizResultsRecommendedArtworks.tsx
```diff
@@ -47,15 +47,16 @@ export const ArtQuizResultsRecommendedArtworksFragmentContainer = createFragment
...
   {
     me: graphql`
-      fragment ArtQuizResultsRecommendedArtworks_me on Me {
+      fragment ArtQuizResultsRecommendedArtworks_me on Me
+        @argumentDefinitions(limit: { type: "Int" }) {
         quiz {
           savedArtworks {
             layer(id: "main") {
-              artworksConnection {
+              artworksConnection(first: $limit) {
```
dzucconi, line 55: Isn't it the case that the limit should be applied to the entire set of recommendations and not `per` each individual artwork?

laurabeth:
@dzucconi does this make sense though? I know you talked about wanting to do this in MP. But I think there might have been a miscommunication about the necessity for 2 calls on this tab?

Idk, thought I'd toy around with it, and it turned out to be sort of uncomplicated.

src/Apps/ArtQuiz/Components/ArtQuizResultsRecommendedArtworks.tsx
laurabeth: no, it's per artwork.

dzucconi: Ah, yep you're right. This works for me

dzucconi (dismissed):
Yeah, I was assuming the limit would have to be applied to the entire set of works; but this is simpler than I thought. Looks good to me.

src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
jonallured: Could also extract this as a function and then have the tests explain the requirements!

src/Apps/ArtQuiz/Routes/ArtQuizResults.tsx
laurabeth: i'm always so paranoid about passing un-memoized values into child components. 😦 fixed.

dzucconi: :) The length of an array isn't computed on the fly, so it's really no different than accessing any other property on an object.

src/Apps/ArtQuiz/Components/ArtQuizResultsTabs.tsx
laurabeth: I question the decision to extract or export a function exclusively for the sake of testing. Have heard arguments for and against, and still not sure where I land.

To your point though, @jonallured, I should test this component. But that sounds like after lunch Laura's problem.

🥪

damassi: The reason: extracted functions are extremely easy to test. Its a matter of dependency injection; rather than having to mock out the world, you simply mock out the function variable inputs. Done and done.

damassi: (Would also appreciate a comment here to avoid having to maintain magic numbers lacking context)

laurabeth: Have a ticket open in the art quiz epic to flesh out the related test suite. Will address this there.

PR #11727, opened 2023-01-11
chore: Deprecate Masonry in favor of ArtworkGrid

olerichter00:
## Description

Deprecate Masonry in favor of ArtworkGrid.


dzucconi (approved):

olerichter00:
Closing this PR because `Masonry` can be useful when not used for artworks (e.g. as a placeholder). Instead of deprecating it, maybe we should [fix the layout problems](https://github.com/artsy/force/pull/11733)

PR #11728, opened 2023-01-11
chore: Trove rebrand GRO-1496

jonallured:
My process for this rebrand was to search the codebase for the string `trove` and then update matches as I encountered them. At this point here's what that looks like:

```
jon@mister-sinister:~/code/force(jonallured/chore-trove-rebrand)% fd trove | wc -l
       0

jon@mister-sinister:~/code/force(jonallured/chore-trove-rebrand)% ag trove -l --stats
src/Apps/Home/Components/HomeEmergingPicksArtworksRail.tsx
src/Apps/Home/Components/HomeContentCards/FallbackCards.tsx
src/Apps/Home/__tests__/HomeEmergingPicksArtworksRail.jest.tsx
src/Apps/Collect/Utils/collectionToArtistSeriesSlugMap.tsx
src/Apps/Gene/Server/__tests__ /redirectGeneToCollection.jest.ts
src/Apps/Gene/Utils/geneToCollectionMap.ts
12 matches
6 files contained matches
```

So no filenames match that string and the remaining instances in the codebase are all required to stay there. I do have one remaining todo here - I need to update the Fallback cards for the homepage but the creative isn't ready yet. I'll follow up on that once I hear from @isakelaris that she's done with her part in the Braze dashboard.

https://artsyproduct.atlassian.net/browse/GRO-1496

/cc @artsy/grow-devs

### DIFF
--- a/src/Apps/ArtQuiz/Components/ArtQuizTrendingCollections.tsx
+++ b/src/Apps/ArtQuiz/Components/ArtQuizTrendingCollections.tsx
@@ -47,7 +47,7 @@ export const ArtQuizTrendingCollectionsFragmentContainer = createFragmentContain
             "artists-on-the-rise"
             "finds-under-1000-dollars"
             "top-auction-lots"
-            "trove-editors-picks"
+            "curators-picks-emerging"
             "contemporary-now"
           ]
         ) {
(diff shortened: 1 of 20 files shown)


jonallured:
Ok let's do this!

PR #11732, opened 2023-01-12
chore: Change "Market News" title on home page to "Artsy Editorial"

dimatretyak:

This PR solves [FX-4542]

### Demo
| Before | After |
| before | after | 

[FX-4542]: https://artsyproduct.atlassian.net/browse/FX-4542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Home/Components/HomeFeaturedMarketNews.tsx
+++ b/src/Apps/Home/Components/HomeFeaturedMarketNews.tsx
@@ -125,7 +125,7 @@ const HomeFeaturedMarketNewsContainer: React.FC = ({ children }) => {
   return (
     <>
       <Flex justifyContent="space-between" alignItems="center">
-        <Text variant="xl">Market News</Text>
+        <Text variant="xl">Artsy Editorial</Text>
 
         <Text
           variant="sm-display"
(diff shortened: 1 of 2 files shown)


src/Apps/Home/Components/HomeFeaturedMarketNews.tsx
```diff
@@ -125,7 +125,7 @@ const HomeFeaturedMarketNewsContainer: React.FC = ({ children }) => {
   return (
     <>
       <Flex justifyContent="space-between" alignItems="center">
-        <Text variant="xl">Market News</Text>
+        <Text variant="xl">Artsy Editorial</Text>
```
gkartalis, line 128: **question(non blocking):** I am not 100% sure what is the preferred way to add strings. When should we add it on the i18n file and when not, will open a thread on #practice-web to check!

gkartalis (approved):
⭐️

src/Apps/Home/Components/HomeFeaturedMarketNews.tsx
gkartalis: left a question [here](https://artsy.slack.com/archives/C2TQ4PT8R/p1673527351037269) to know what to do for the next copy changes that we will encounter

PR #11733, opened 2023-01-12
feat: New Masonry component

olerichter00:
## Description

Draft of a new `Masonry` component that replaces [Masonry](https://github.com/artsy/force/blob/main/src/Components/Masonry.tsx) and [ArtworkGrid](https://github.com/artsy/force/blob/main/src/Components/ArtworkGrid/ArtworkGrid.tsx#L9) .

I've implemented two different interfaces. The first one uses `children` and the second one is similar to ReactNative's [FlatList](https://reactnative.dev/docs/flatlist). 

**Interface 1:**

```jsx
<Masonry2
  aspectRatios={artworks.map(artwork => artwork.image?.aspectRatio)}
  onLoadMore={handleLoadMore}
>
  {artworks.map(artwork => {
    return (
      <ArtworkGridItemFragmentContainer artwork={artwork} />
    )
  })}
</Grid>
```

**Interface 2:**

```jsx
<Masonry3
  data={artworks}
  renderItem={artwork => {
    return (
     <ArtworkGridItemFragmentContainer artwork={artwork} />
    )
  }}
  onLoadMore={handleLoadMore}
/>
```

Screenshot 2023-01-12 at 19 15 21

### DIFF
--- a/src/Components/Grid.tsx
+++ b/src/Components/Grid.tsx
@@ -1,0 +1,148 @@
+import { Box, Flex } from "@artsy/palette"
+import { range } from "lodash"
+import { ReactNode, useEffect, useRef } from "react"
+import { Media, valuesWithBreakpointProps } from "Utils/Responsive"
+
+export interface GridProps {
+  aspectRatios: (number | undefined | null)[]
+  children: ReactNode[]
+  columnCount?: number[]
+  onLoadMore?: () => any
+  sectionMargin?: number
+  style?: React.CSSProperties | undefined
+}
+
+export const Grid: React.FC<GridProps> = ({
+  columnCount = [2, 3, 4],
+  style,
+  onLoadMore,
+  ...props
+}) => {
+  const containerRef = useRef<HTMLElement>(null)
+
+  useLoadMore(onLoadMore, containerRef)
+
+  const breakpointProps = valuesWithBreakpointProps(columnCount)
+
+  const breakpoints = breakpointProps.map(([columnCount, mediaProps], i) => {
+    return (
+      <Box ref={containerRef as any} style={style}>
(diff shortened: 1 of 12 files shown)


damassi:
And to be clear, we're keeping the flowing masonry style? Because otherwise, it's [as simple as this](https://github.com/artsy/force/blob/main/src/Components/ArtworkGrid/ArtworkGrid.tsx#L215-L230), which appears on the Auction page (so we should account for this mode)

Also, it'd be great to avoid using our `Media` component. Is there a way to implement this without it? Ideally we stick to styled-system's responsive props or some other model. (`Media` does not behave well with React 18 and we're going to have to address this soon.)

damassi:
FWIW, restarted the discussion around the grid [over here](https://artsy.slack.com/archives/C9XJKPY9W/p1673540414048299), so lets keep an eye on developments in this direction.

olerichter00:
@damassi I have messed up the naming (updated the description and title). The new component would replace [Masonry](https://github.com/artsy/force/blob/main/src/Components/Masonry.tsx), which is used in some places and [causes some problems](https://artsy.slack.com/archives/C01B2P6LJUU/p1672413032860489).

It uses the logic of [ArtworkGrid](https://github.com/artsy/force/blob/main/src/Components/ArtworkGrid/ArtworkGrid.tsx) but decouples it from everything else (as suggested by @dzucconi).

damassi:
That sounds good! With the rename it makes more sense. We're going to be running some A/B tests around this shortly so please tag me on any PRs that replace existing grids and note that this might still need to incorporate the logic from the ArtworkGrid for said test. 👍

damassi:
@olerichter00 - Going to close this as it appears to be stale. Feel free to reopen if you still need it!

PR #11741, opened 2023-01-13
feat(TX-980): Private sale checkout flow

AlicanAkyuz:

This PR solves [TX-980]

### Description

This PR adjusts the checkout flow steps for private sale orders. If an order's source is private_sale, user sees only payment and reviews steps, and if they attempt to reach /shipping, they are redirected to payment. 

#### Video 

https://user-images.githubusercontent.com/42584148/212288346-444db502-49a5-48e7-85ed-9296f102fd09.mov

[TX-980]: https://artsyproduct.atlassian.net/browse/TX-980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fladson (approved):
Nice! I thought this would require way more changes.

kiryl-zubarau (approved):
LGTM!

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -352,7 +353,13 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
       <OrderRouteContainer
         currentStep="Payment"
-        steps={order.mode === "OFFER" ? offerFlowSteps : buyNowFlowSteps}
+        steps={
+          order.mode === "OFFER"
+            ? offerFlowSteps
+            : order.source === "private_sale"
+            ? privateFlowSteps
+            : buyNowFlowSteps
+        }
```
damassi, line 362: nonblocking suggestion: nested ternaries like this tend to always be super hard to reason about later on by devs who didn't write the code (and are bug prone / tend to be unwound during debugging, similar to implicit returns in react components. Best to just avoid it in favor of more boring `if` statements or a switch.

```tsx
let steps
if (offerMode === "OFFER") {
  steps = offerFlowSteps
} else {
  if (order.soruce === 'private_sale') {
    steps = privateFlowSteps
  } else {
    steps = buyNowSteps
  }
}
```
or the switch form, which is a bit clever but also quite clear:
```tsx
const steps = (() => {
  switch (true) {
    case offer.mode === "OFFER": 
      return offerFlowSteps
    case order.source === 'private_sale':
      return privateFlowSteps
    default:
      return buyNowFlowSteps
  }
})()
```

starsirius (approved):
Nice one!

PR #11748, opened 2023-01-16
feat(TX-991): payment step for private sale orders

AlicanAkyuz:
The type of this PR is: **Feat**

This PR solves [TX-991]

### Description

This PR updates the payment step for private sale orders. 
- Email at the footer is updated 
- Explanation texts updated for wire_transfer 
- Private conditions of sale link UI added 
- Extra conditions of sale is added 

Notes
1- We need to update the link for Private conditions of sale 
2- When Exchange offers, we need to pull the extra conditions of sale  from there 

#### Video 
https://user-images.githubusercontent.com/42584148/212628364-9d0f53f8-4eff-498e-b42b-d4f0f26f8a23.mov

[TX-991]: https://artsyproduct.atlassian.net/browse/TX-991?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11749, opened 2023-01-16
feat(TX-982): Private sales review step changes

AlicanAkyuz:

This PR solves [TX-982]

### Description

This PR updates the review step for private sale orders. 
1- button texts is updated to 'Complete Purchase' 
2- 'Change' option for shipment is hidden 
3- PS conditions of sales and extra condition of sales are added 
4- footer email is updated 

#### Video 
https://user-images.githubusercontent.com/42584148/212665602-cefbcd48-031b-4237-a711-988b1d7bd792.mov

[TX-982]: https://artsyproduct.atlassian.net/browse/TX-982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AlicanAkyuz:
note: I have rebased this from https://github.com/artsy/force/pull/11748 because some changes in this other PR applies here as well.

src/Components/StepSummaryItem.tsx
```diff
@@ -25,11 +26,15 @@ const LockIconPositioner = styled(Flex)`
...
   title,
   onChange,
+  orderSource,
   children,
   locked,
   ...others
 }) => {
   const showHeading = title || (onChange && !locked)
+  const allowChangeShipping =
+    !locked && onChange && orderSource !== "private_sale"
```
starsirius, line 36: Looks like the `locked` flag works well for our use case. Instead of expanding the logic in the `StepSummaryItem`, I wonder if we can pass in `locked` for private sale orders. It will display the lock icon, and I think it makes sense.

AlicanAkyuz: ah cool, thanks for pointing this out, I didn't know in which cases `locked` would be true. I'll update the component.

src/Apps/Order/Components/PaymentMethodSummaryItem.tsx
```diff
@@ -47,7 +52,7 @@ export const PaymentMethodSummaryItemFragmentContainer = createFragmentContainer
   {
     order: graphql`
       fragment PaymentMethodSummaryItem_order on CommerceOrder {
-        paymentMethod
```
AlicanAkyuz, line 50: removed this because we are using `paymentMethodDetails` to render relevant data.

src/Components/StepSummaryItem.tsx
```diff
@@ -22,14 +22,15 @@ const LockIconPositioner = styled(Flex)`
   right: 0;
 `
 
-export const StepSummaryItem: React.SFC<StepSummaryItemProps> = ({
+export const StepSummaryItem: React.FC<StepSummaryItemProps> = ({
```
AlicanAkyuz, line 25: small chore, unrelated.

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -402,6 +418,7 @@ export const ReviewRoute: FC<ReviewProps> = props => {
               <ShippingSummaryItem
                 order={order}
                 onChange={onChangeShippingAddress}
+                locked={order.source === "private_sale"}
```
starsirius, line 421: 👍

starsirius (approved):
LGTM

PR #11761, opened 2023-01-17
fix(authdialog): corrects how email sign up is passed

dzucconi:
I had made the assumption that the `signUp` function was just passing along the data it was given. As it happens it was selecting specific values, but the object I passed it still satisfied the type. So while the test was running we weren't including `agreed_to_receive_emails` in the registration payload.

### DIFF
--- a/src/Components/AuthDialog/Views/AuthDialogSignUp.tsx
+++ b/src/Components/AuthDialog/Views/AuthDialogSignUp.tsx
@@ -49,15 +49,23 @@ export const AuthDialogSignUp: FC<AuthDialogSignUpProps> = ({
       validateOnBlur={false}
       initialValues={{
         ...INITIAL_VALUES,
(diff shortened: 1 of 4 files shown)


src/Components/AuthDialog/Views/AuthDialogSignUp.tsx
```diff
@@ -49,15 +49,23 @@ export const AuthDialogSignUp: FC<AuthDialogSignUpProps> = ({
...
         setFieldValue("mode", "Loading")
 
         try {
-          const { user } = await signUp(values)
+          const { user } = await signUp({
+            email,
+            name,
+            password,
+            agreedToReceiveEmails,
+          })
```
dzucconi, line 68: Going forward, will avoid passing an object and prefer explicitly destructuring and constructing a payload.

PR #11766, opened 2023-01-18
feat(TX-948): replace artwork data to versioned data in checkout

rquartararo:

This PR solves [TX-948]

### Description

This PR replaces the artwork data with versioned artwork data for the `ArtworkSummaryItem` and `ItemReview` components in the checkout flow. 

NOTE: We have also hidden artwork details only for **private sales** orders where artwork data was not able to be replaced with versioned data. We will eventually unhide the fields once we can retrieve the data. 

### Private sales w/ hidden info

Screenshot 2023-01-18 at 12 02 17 PM


src/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -65,24 +68,27 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
...
+        {!privateSale && (
+          <Text variant="sm" overflowEllipsis color="black60">
+            {name}
+          </Text>
+        )}
+        {!privateSale && (
+          <Text variant="sm" color="black60">
+            {shippingOrigin}
+          </Text>
+        )}
```
AlicanAkyuz, line 83: suggestion: it could reduce the amount of checks if we could render `name` and `shippingOrigin` within one `!privateSale `check.

AlicanAkyuz (approved):
put a couple small suggestions. everything LGTM.

src/Apps/Order/Components/ArtworkSummaryItem.tsx
rquartararo: good idea!

AlicanAkyuz (approved):

PR #11767, opened 2023-01-18
feat(FX-4535): mark specific notification as read when clicking on an entry in the activity panel

dimatretyak:

This PR solves [FX-4535]

Review app: [mark-notification-as-read](https://mark-notification-as-read.artsy.net/)

### Demo
https://user-images.githubusercontent.com/3513494/213193037-3bdd51be-1f56-4058-a9d6-64c7c6d3cfe9.mp4

[FX-4535]: https://artsyproduct.atlassian.net/browse/FX-4535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Notifications/NotificationItem.tsx
+++ b/src/Components/Notifications/NotificationItem.tsx
@@ -7,6 +7,11 @@ import styled from "styled-components"
 import { themeGet } from "@styled-system/theme-get"
 import { ActionType } from "@artsy/cohesion"
 import { useTracking } from "react-tracking"
+import { useSystemContext } from "System"
+import createLogger from "Utils/logger"
+import { markNotificationAsRead } from "Components/Notifications/Mutations/markNotificationAsRead"
+
+const logger = createLogger("NotificationItem")
 
 interface NotificationItemProps {
   item: NotificationItem_item$data
@@ -16,6 +21,7 @@ const UNREAD_INDICATOR_SIZE = 8
 
 const NotificationItem: React.FC<NotificationItemProps> = ({ item }) => {
   const { trackEvent } = useTracking()
+  const { relayEnvironment } = useSystemContext()
   const artworks = extractNodes(item.artworksConnection)
   const remainingArtworksCount = item.objectsCount - 4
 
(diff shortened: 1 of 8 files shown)


dblandin (approved):
LGTM!

Non-blocking but I think it would be wise to create an upkeep ticket to setup some more robust tests that allow to us exercise some of the tap/click behavior in the activity panel UI.

Hopefully [these docs](https://github.com/artsy/force/blob/main/src/DevTools/README.md#guide-to-createtestenv) are still up-to-date. Would be nice to protect ourselves against regressions by simulating a tap on one of these notification items and ensuring the we're marking as read _and_ navigating to the correct place.

dimatretyak:
> Non-blocking but I think it would be wise to create an upkeep ticket to setup some more robust tests that allow to us exercise some of the tap/click behavior in the activity panel UI

Yes, it's a good idea. I originally planned to add tests, but due to some limitations of how we mock relay env for tests, this could not be done.

gkartalis (approved):

nickskalkin (approved):

src/Components/Notifications/NotificationItem.tsx
```diff
@@ -31,16 +37,39 @@ const NotificationItem: React.FC<NotificationItemProps> = ({ item }) => {
...
+        item.internalID
+      )
+      const responseOrError = response.markNotificationAsRead?.responseOrError
+      const errorMessage = responseOrError?.mutationError?.message
+
+      if (errorMessage) {
+        throw new Error(errorMessage)
+      }
+    } catch (error) {
+      logger.error(error)
```
nickskalkin, line 58: **question / non-blocking**: What is our policy for handling exceptions / errors of frontend side? Should we send an error to Sentry here?

dimatretyak: when we call `logger.error', the error will be [logged in sentry](https://github.com/artsy/force/blob/80d11482e601df71506dac918ea8702d816eb0b7/src/Utils/logger.ts#L23-L25)

PR #11775, opened 2023-01-18
refactor: moves over deprecated countdown timer component

dzucconi:
So this moves over the Palette component so that we can drop Luxon from its dependencies and remove Palette from handling this kind of thing.

I'll move this off draft once the Palette PR is ready to go.

Also note that I'm still leaving this as deprecated. It's *another* one for the Order app to move over. It seems we've landed on using the `useTimer` hook for this sort of thing, which is _much_ simpler. So I've recommended that as a replacement.


src/Components/__tests__/CountdownTimer.jest.tsx
```diff
@@ -104,7 +104,7 @@ describe("CountdownTimer", () => {
 
     const text = timer.text()
     expect(text).toMatchInlineSnapshot(
-      `"Time remaining01d 00h 00m 00s leftRespond by Dec 4, 8:50am ESTExpired offers end the negotiation process permanently."`
+      `"01d 00h 30m 00s leftRespond by Dec 4, 8:50am ESTExpired offers end the negotiation process permanently."`
```
dzucconi, line 107: This spec was just wrong. As you can see it wasn't asserting on that half hour difference. It works now so 🤷

yarn.lock
```diff
@@ -15046,6 +15051,11 @@ luxon@^1.15:
   resolved "https://registry.yarnpkg.com/luxon/-/luxon-1.28.1.tgz#528cdf3624a54506d710290a2341aa8e6e6c61b0"
   integrity sha512-gYHAa180mKrNIUJCbwpmD0aTu9kV0dREDrwNnuyFAsO1Wt0EVYSZelPnJlbj9HplzXX/YWXHFTL45kvZ53M0pw==
 
+luxon@^3.2.1:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/luxon/-/luxon-3.2.1.tgz#14f1af209188ad61212578ea7e3d518d18cee45f"
+  integrity sha512-QrwPArQCNLAKGO/C+ZIilgIuDnEnKx5QYODdDtbFaxzsbZcc/a7WFq7MhsVYgRlwawLtvOUESTlfJ+hc/USqPg==
```
dzucconi, line 15057: This was a transitive dependency via Palette.

dzucconi:
I also don't think it'd be particularly difficult from removing this from our dependencies in Force. We really only use it in tests so it could just be a devDep.

PR #11788, opened 2023-01-19
feat(art-quiz): add auth redirects to artQuizRoutes

laurabeth:

This PR solves [GRO-1507]

### Description

This PR adds a redirect for logged out users who hit any of the /art-quiz routes. Now, instead of seeing a blank screen or a 500, a logged out user will be redirected to /login and redirected back to the art quiz once they've signed in.


src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -1,8 +1,26 @@
 import loadable from "@loadable/component"
-import { RedirectException } from "found"
+import { Match, RedirectException } from "found"
 import { graphql } from "react-relay"
+import { ArtsyResponse } from "Server/middleware/artsyExpress"
 import { AppRouteConfig } from "System/Router/Route"
 
+const LOGIN_COPY = "Log in to take the Art Quiz."
+const REDIRECT_URL = `/login?redirectTo=/art-quiz&afterSignUpAction=/art-quiz&copy=${LOGIN_COPY}`
```
dzucconi, line 8: `afterSignUpAction=/art-quiz` isn't a valid afterSignUpAction, you can remove that param.

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -1,8 +1,26 @@
...
+
+const artQuizServerSideRedirect = ({ res }: { res: ArtsyResponse }) => {
+  if (!res.locals.sd.FEATURE_FLAGS["art-quiz"].flagEnabled) {
+    res.redirect("/")
+  }
+  if (!res.locals.sd.CURRENT_USER) {
+    res.redirect(REDIRECT_URL)
+  }
+}
+const artQuizClientSideRedirect = ({ match }: { match: Match }) => {
```
dzucconi, line 18: Why would we need a client-side redirect?

laurabeth: The assumption was that this would be relevant if the user logged out mid quiz or something, but I didn't think too deeply about it. Probably incorrect?

dzucconi: Yeah, safe to remove this. Logging out does a hard-refresh anyway.

laurabeth: done

TMcMeans:
**question [non-blocking]:** 

Wondering if it's better to redirect to login or send them to the artsy `/` homepage? No opinion either way, but do you think we should stick to convention here?

dzucconi (approved):

src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -1,8 +1,21 @@
 import loadable from "@loadable/component"
 import { RedirectException } from "found"
 import { graphql } from "react-relay"
+import { ArtsyResponse } from "Server/middleware/artsyExpress"
 import { AppRouteConfig } from "System/Router/Route"
 
+const LOGIN_COPY = "Log in to take the Art Quiz."
+const REDIRECT_URL = `/login?redirectTo=/art-quiz&copy=${LOGIN_COPY}`
```
TMcMeans, line 8: **nitpick [non-blocking]:** 

I see now that we're directing the user to login to take the art quiz via url. If they've already taken it and go straight to the results, I think this copy would be a little strange to read. Maybe we can make it more generalized?

TMcMeans (approved):
Got a couple small non-blocking nitpicky comments :) but otherwise LGTM 👍🏾 🥇

dzucconi:
You would send them to login with a `redirectTo` to the quiz page. If you're hitting this route then by default you want to login.

src/Apps/ArtQuiz/artQuizRoutes.tsx
TMcMeans: I mean since the login is a pop-up modal this isn't a big deal, disregard!

dzucconi: We can target that separately by redirecting on only the results route.

dzucconi: (It's not a pop up actually)

TMcMeans:
> You would send them to login with a `redirectTo` to the quiz page. If you're hitting this route then by default you want to login.

does this hold true for users who log out while viewing their results page though? just making sure all bases are covered

dzucconi:
> for users who log out while viewing their results page though

I'd imagine this is an extremely rare event, but sure, why not? Presumably you would want to pick up where you left off. No one is going to log out, then immediately log back in in order to navigate to the homepage.

TMcMeans:
I'm gonna hit merge on this. We can prolly check out how the auth redirect feels for navigating from the results page during our next round of QA??

PR #11790, opened 2023-01-20
fix(TX-1009): review page will not error out if no image

rquartararo:

This PR solves [TX-1009]

### Description

This PR makes it so when the artwork version image "resized" is null on the review page, the UI will not error out. 

Screenshot 2023-01-20 at 9 52 44 AM

[TX-1009]: https://artsyproduct.atlassian.net/browse/TX-1009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/ItemReview.tsx
+++ b/src/Apps/Order/Components/ItemReview.tsx
@@ -34,11 +34,9 @@ export const ItemReview: React.FC<ItemReviewProps> = ({
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
       medium,
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-      image: {
-        resized: { url },
-      },
+      image,
       // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/ItemReview.tsx
```diff
@@ -75,7 +73,14 @@ export const ItemReview: React.FC<ItemReviewProps> = ({
           {shortDescription}
         </Text>
       </Flex>
-      <Image maxHeight={375} width={185} src={url} alt={title} />
+      {image && (
```
AlicanAkyuz, line 76: does it make sense to check `image.resized` property before we render `Image`? 🤔

rquartararo: I was thinking if image.resized was null then image would be as well? I can add it in just in to be safe though.

AlicanAkyuz: you might be right, I'm not sure if it's possible to have image without resized. adding the check for extra safety sounds good 👍

starsirius (approved):
LGTM! Thank you for the quick fix! 🙏

PR #11801, opened 2023-01-23
fix: Simplify My Collection local image optimization

olerichter00:

This PR solves [CX-3336]

### Description

Local images are used while Gemini processes newly uploaded images.

This PR fixes some issues we had with local images in My Collection. Instead of saving images by artwork ID, they will be saved by image ID. This will simplify the use of local images and should reduce the complexity.

Sorry for the huger PR! Most of the files are generated by Relay 🙃 

[CX-3336]: https://artsyproduct.atlassian.net/browse/CX-3336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/GridItem.tsx
```diff
@@ -61,6 +63,14 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
     rest.onMouseLeave?.(event)
   }
 
+  const localImagePlaceholder =
```
MounirDhahri, line 66: **praise:** nice touch

MounirDhahri: **Q** I don't get what you are trying to do here 👀 would you mind maybe explaining it to me? we can later add a comment to the code here maybe

src/Components/CountdownTimer.tsx
```diff
@@ -26,7 +26,6 @@ export const CountdownTimer: React.FC<CountdownTimerProps> = ({
   countdownStart,
 }) => {
   const currentTime = useCurrentTime({ syncWithServer: true })
-  console.log({ currentTime })
```
MounirDhahri, line 29: thank you

src/Utils/localImageHelpers.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export const getLocalImage = async (
+  key: string
+): Promise<LocalImage | null> => {
+  const storeKey = `${IMAGE_KEY_PREFIX}_${key}`
+  const imageJSONString = await localforage.getItem(storeKey)
+
+  if (!imageJSONString) return null
+
+  return JSON.parse(imageJSONString as string)
```
MounirDhahri, line 37: **suggestion:** should we maybe here check if the image expired or not and if yes delete it 👀?

MounirDhahri (dismissed):
LGTM, thanks for making this much needed change. I left very few non-blocking comments, please let me know if something wasn't clear

PR #11804, opened 2023-01-23
feat: Hide AP blue dot indicator if there are no new notifications

nickskalkin:

This PR solves [FX-4538]

### Description

As a user
Given I have unseen notifications
When I visit the activity panel
The top-level indicator goes away
And I continue to see unread indicators to individual notifications.

[FX-4538]: https://artsyproduct.atlassian.net/browse/FX-4538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Notifications/util.ts
```diff
@@ -14,3 +17,28 @@ const isArtworksBasedNotification = (
...
+  if (!lastNotificationDatetime) {
+    return false
+  }
+
+  if (!window.localStorage.getItem(LAST_SEEN_NOTIFICATION_PUBLISHED_AT_KEY)) {
+    return true
+  }
+
+  return (
+    Date.parse(lastNotificationDatetime) >
```
dimatretyak, line 39: **suggestion**: use `DateTime.fromISO` (this approach is usually used in code) instead of `Date.parse`

dimatretyak: **suggestion**: some refactor suggestions

```javascript
export const hasNewNotifications = lastNotificationDatetime => {
  const prevLastNotificationDatetime = window.localStorage.getItem(LAST_SEEN_NOTIFICATION_PUBLISHED_AT_KEY)

  if (!lastNotificationDatetime) {
    return false
  }

  if (!prevLastNotificationDatetime) {
    return true
  }

  const prevDate = DateTime.fromISO(prevLastNotificationDatetime)
  const newDate = DateTime.fromISO(lastNotificationDatetime)

  return newDate > prevDate
}
```

src/Components/NavBar/NavBarMobileMenu/NavBarMobileMenuNotifications.tsx
```diff
@@ -51,7 +60,7 @@ export const NavBarMobileMenuNotifications: React.FC<NavBarMobileMenuNotificatio
             }}
           >
             Activity
-            {hasNotifications && <Indicator />}
+            {hasNewNotifications(notification?.publishedAt) && <Indicator />}
```
dimatretyak, line 63: we should also use `hasNotifications` here

src/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -123,6 +133,14 @@ export const NavBarLoggedInActionsQueryRenderer: React.FC<{}> = () => {
       environment={relayEnvironment}
       query={graphql`
         query NavBarLoggedInActionsQuery {
+          notificationsConnection(first: 1) {
```
dimatretyak, line 136: **question**: maybe it's worth query more than 1 artworks here? so how do we then do some data filtering?

src/Components/Notifications/NotificationsList.tsx
```diff
@@ -40,10 +40,17 @@ const NotificationsList: React.FC<NotificationsListProps> = ({
...
+
   const nodes = extractNodes(viewer.notifications).filter(node =>
     shouldDisplayNotification(node)
   )
 
+  useEffect(() => {
+    if (nodes?.length > 0) {
+      setLastSeenNotificationDate(nodes[0].publishedAtAbsolute)
+    }
+  }, [nodes])
```
dimatretyak, line 52: `useEffect` will be called every time the component is updated (since the reference type to `nodes` will change every time). what do you think about this code?

```javascript
const nodes = extractNodes(viewer.notifications).filter(node =>
  shouldDisplayNotification(node)
)
const recentNotification = nodes[0]
const recentNotificationPublishedAt = recentNotification?.publishedAtAbsolute

useEffect(() => {
  if (recentNotificationPublishedAt) {
    setLastSeenNotificationDate(recentNotificationPublishedAt)
  }
}, [recentNotificationPublishedAt])
```

src/Components/Notifications/util.ts
```diff
@@ -14,3 +17,28 @@ const isArtworksBasedNotification = (
...
+
+  window.localStorage.setItem(LAST_SEEN_NOTIFICATION_PUBLISHED_AT_KEY, date)
+}
+
+export const hasNewNotifications = lastNotificationDatetime => {
+  if (!lastNotificationDatetime) {
+    return false
+  }
+
+  if (!window.localStorage.getItem(LAST_SEEN_NOTIFICATION_PUBLISHED_AT_KEY)) {
```
dimatretyak, line 34: @dzucconi @damassi what do you think about this approach? will there be any problems with server rendering here? Should we check `isServer`/`typeof window !== "undefined"` here?

src/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -60,7 +70,7 @@ export const NavBarLoggedInActions: React.FC<Partial<
           >
             <BellIcon title="Notifications" fill="currentColor" />
 
-            {hasNotifications && (
+            {hasNewNotifications(notification?.publishedAt) && (
```
dimatretyak, line 73: we should use/check `hasNotifications` here

src/Components/Notifications/NotificationsList.tsx
dimatretyak: IMO we should call `setLastSeenNotificationDate` only when `type === "all"` (discussed this point on the zoom call)

src/Components/Notifications/util.ts
```diff
@@ -14,3 +17,28 @@ const isArtworksBasedNotification = (
 ) => {
   return ["ARTWORK_ALERT", "ARTWORK_PUBLISHED"].includes(notificationType)
 }
+
+export const setLastSeenNotificationDate = date => {
```
dzucconi, line 21: Why extract this? Just inline this in the `useEffect` so it isn't misused. If it is extracted however you should annotate the type.

src/Components/Notifications/util.ts
```diff
@@ -14,3 +17,28 @@ const isArtworksBasedNotification = (
...
+  if (!date) {
+    return
+  }
+
+  window.localStorage.setItem(LAST_SEEN_NOTIFICATION_PUBLISHED_AT_KEY, date)
+}
+
+export const hasNewNotifications = lastNotificationDatetime => {
```
dzucconi, line 29: This whole function could just be a compound boolean statement rather than a series of conditionals that return booleans.

dzucconi: There should be no `any` types in this codebase.

src/Components/Notifications/__tests__/util.jest.ts
```diff
@@ -1,4 +1,8 @@
-import { shouldDisplayNotification } from "Components/Notifications/util"
+import {
+  shouldDisplayNotification,
+  hasNewNotifications,
```
dzucconi, line 3: Really, this can't be used like this. I'd consider just extracting all of this into a hook, and set a state flag in a useEffect.

src/Components/NavBar/NavBarLoggedInActions.tsx
nickskalkin: increased to 3, thanks!

src/Components/Notifications/NotificationsList.tsx
nickskalkin: done, thanks!

src/Components/Notifications/UnseenNotificationsIndicator.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  if (!prevLastNotificationDatetime) {
+    return true
+  }
+
+  const prevDate = DateTime.fromISO(prevLastNotificationDatetime)
+  const newDate = DateTime.fromISO(lastNotificationDatetime)
+
+  return newDate > prevDate
+}
```
dzucconi, line 80: Why not just have this hook execute the query for the data it needs? (We can support Relay hooks)

nickskalkin:
closing this PR without merging, we've decided to take a slightly different approach involving BE changes

PR #11813, opened 2023-01-23
chore(feature-flags): switch to warn and only warn during development

dzucconi:
The noise in the specs can be misleading when looking at the CI logs. It seems to me that: these aren't errors but are warnings, and are only really useful to developers in development mode. This is open to debate, of course.


damassi:
haha @dzucconi - was looking at this on Friday and ran out of time! Ty!!!

PR #11818, opened 2023-01-24
fix(auction-results): Update Auction Results awaiting results logic

MrSltun:
The type of this PR is: **TYPE**

This PR solves [PROJECT-XX]

### Description
This PR updates the logic of the Awaiting Results condition for the auction results 


MounirDhahri (approved):
LGTM - I would have recommended adding a test to avoid the same regression in the future but after reflecting about this, I think the right thing to do here is to move this to metaphysics. not necessarily now but let's create a ticket to do it in Jira

MounirDhahri (approved):

MounirDhahri:
🚢🚢🚢

PR #11846, opened 2023-01-30
feat(FX-4526): Create component for Saves and Wishlist rail

dimatretyak:

This PR solves [FX-4526]

Review app: [saves-and-wishlist](https://saves-and-wishlist.artsy.net/collector-profile/saves2)

### Figma designs
* [Desktop view](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB/Save-%2B-Wishlist?node-id=308%3A216953&t=rOlJYAozFRqXRwb9-0)
* [mWeb view](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB/Save-%2B-Wishlist?node-id=466%3A144453&t=rOlJYAozFRqXRwb9-0)

### Demo
https://user-images.githubusercontent.com/3513494/215774846-21576c87-1000-4b7a-847a-cc87c98e2d44.mp4

[FX-4526]: https://artsyproduct.atlassian.net/browse/FX-4526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -1,24 +1,53 @@
 import { Shelf } from "@artsy/palette"
-import { SavesItem } from "Apps/CollectorProfile/Routes/Saves2/Components/SavesItem"
+import { SavesItemFragmentContainer } from "Apps/CollectorProfile/Routes/Saves2/Components/SavesItem"
+import { orderBy } from "lodash"
 import { FC, useState } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { extractNodes } from "Utils/extractNodes"
+import { CollectorProfileSaves2Route_me$data } from "__generated__/CollectorProfileSaves2Route_me.graphql"
 
-const items = Array.from(Array(10))
+interface CollectorProfileSaves2RouteProps {
+  me: CollectorProfileSaves2Route_me$data
+}
 
-export const CollectorProfileSaves2Route: FC = () => {
+const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
+  me,
+}) => {
   const [selectedIndex, setSelectedIndex] = useState(0)
+  const collections = extractNodes(me.collectionsConnection)
+  // Placing the default collection at the top of the list
+  const sortedCollections = orderBy(collections, ["default"], ["desc"])
 
   return (
     <Shelf showProgress={false}>
-      {items.map((_, index) => (
-        <SavesItem
-          key={`saves-item-${index}`}
-          title="Saves"
-          artworksCount={36}
+      {sortedCollections.map((collection, index) => (
+        <SavesItemFragmentContainer
+          key={collection.internalID}
+          item={collection}
           isSelected={selectedIndex === index}
-          imagesLayout={index === 0 ? "grid" : "stacked"}
+          imagesLayout={collection.default ? "grid" : "stacked"}
           onClick={() => setSelectedIndex(index)}
         />
       ))}
     </Shelf>
   )
 }
+
+export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContainer(
+  CollectorProfileSaves2Route,
+  {
+    me: graphql`
+      fragment CollectorProfileSaves2Route_me on Me {
+        collectionsConnection(first: 10) {
+          edges {
+            node {
+              internalID
+              default
+              ...SavesItem_item
+            }
+          }
+        }
+      }
+    `,
+  }
+)
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -1,10 +1,12 @@
 import { Shelf } from "@artsy/palette"
 import { SavesItem } from "Apps/CollectorProfile/Routes/Saves2/Components/SavesItem"
-import { FC } from "react"
+import { FC, useState } from "react"
 
 const items = Array.from(Array(10))
 
 export const CollectorProfileSaves2Route: FC = () => {
+  const [selectedIndex, setSelectedIndex] = useState(0)
+
   return (
     <Shelf showProgress={false}>
       {items.map((_, index) => (
@@ -12,8 +14,9 @@ export const CollectorProfileSaves2Route: FC = () => {
           key={`saves-item-${index}`}
           title="Saves"
           artworksCount={36}
-          isSelected={index === 0}
+          isSelected={selectedIndex === index}
           imagesLayout={index === 0 ? "grid" : "stacked"}
+          onClick={() => setSelectedIndex(index)}
         />
       ))}
     </Shelf>
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -1,6 +1,21 @@
-import { Text } from "@artsy/palette"
+import { Shelf } from "@artsy/palette"
+import { SavesItem } from "Apps/CollectorProfile/Routes/Saves2/Components/SavesItem"
 import { FC } from "react"
 
+const items = Array.from(Array(10))
+
 export const CollectorProfileSaves2Route: FC = () => {
-  return <Text>/collector-profile/saves2</Text>
+  return (
+    <Shelf showProgress={false}>
+      {items.map((_, index) => (
+        <SavesItem
+          key={`saves-item-${index}`}
+          title="Saves"
+          artworksCount={36}
+          isSelected={index === 0}
+          imagesLayout={index === 0 ? "grid" : "stacked"}
+        />
+      ))}
+    </Shelf>
+  )
 }
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -38,7 +38,7 @@ export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContai
   {
     me: graphql`
       fragment CollectorProfileSaves2Route_me on Me {
-        collectionsConnection(first: 10) {
+        collectionsConnection(first: 20) {
           edges {
             node {
               internalID
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex, Image, Spacer } from "@artsy/palette"
 import { prepareImageURLs } from "Apps/CollectorProfile/Routes/Saves2/Utils/prepareImageURLs"
 import { FC } from "react"
+import { cropped } from "Utils/resized"
 
 interface FourUpImageLayoutProps {
   imageURLs: (string | null)[]
@@ -53,5 +54,18 @@ const RowImage: FC<RowImageProps> = ({ url }) => {
     )
   }
 
-  return <Image width={SIZE} height={SIZE} preventRightClick src={url} />
+  const image = cropped(url, {
+    width: LARGE_IMAGE_SIZE,
+    height: LARGE_IMAGE_SIZE,
+  })
+
+  return (
+    <Image
+      width={SIZE}
+      height={SIZE}
+      preventRightClick
+      src={image.src}
+      srcSet={image.srcSet}
+    />
+  )
 }
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
@@ -1,4 +1,5 @@
 import { Box, Flex, Image, Spacer } from "@artsy/palette"
+import { prepareImageURLs } from "Apps/CollectorProfile/Routes/Saves2/Utils/prepareImageURLs"
 import { FC } from "react"
 
 interface FourUpImageLayoutProps {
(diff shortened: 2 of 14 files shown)


src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
```diff
@@ -0,0 +1,73 @@
...
+        borderColor="black10"
+        aria-label="Image placeholder"
+      />
+    )
+  }
+
+  const image = cropped(url, {
+    width: LARGE_IMAGE_SIZE,
+    height: LARGE_IMAGE_SIZE,
+  })
```
dimatretyak, line 61: **question**: or is it better to query cropped image from backend (see implementation [here](https://github.com/artsy/force/commit/5aa0da21f98805578aab198414c82993780a88eb))?

src/Apps/CollectorProfile/Routes/Saves2/Utils/prepareImageURLs.ts
```diff
@@ -0,0 +1,7 @@
+export const prepareImageURLs = (imageURLs: (string | null)[]) => {
+  // Ensure we have an array of exactly 4 images
+  return Array.from(Array(4)).reduce(
+    (acc, _, i) => [...acc, imageURLs[i] ?? null],
+    [] as string[]
+  )
+}
```
dimatretyak, line 7: **question**: should we add additional logic so that non-empty image URLs always go at the beginning of the array? Something like this `from ["url-1", null, "url-2", null] to ["url-1", "url-2", null, null]`

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -1,6 +1,53 @@
...
+  me: CollectorProfileSaves2Route_me$data
 }
+
+const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
+  me,
+}) => {
+  const [selectedIndex, setSelectedIndex] = useState(0)
+  const collections = extractNodes(me.collectionsConnection)
+  // Placing the default collection at the top of the list
+  const sortedCollections = orderBy(collections, ["default"], ["desc"])
```
dimatretyak, line 19: **question**: maybe we should place this logic on the backend side?

src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
dzucconi: It doesn't make a difference — whatever is easier. In this case, using the `cropped` function is easier.

src/Apps/CollectorProfile/Routes/Saves2/Components/SavesItem.tsx
```diff
@@ -0,0 +1,74 @@
+import { Box, Flex, Text } from "@artsy/palette"
+import { FourUpImageLayout } from "Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout"
+import { StackedImageLayout } from "Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout"
+import { FC } from "react"
+import { createFragmentContainer, graphql } from "react-relay"
+import { extractNodes } from "Utils/extractNodes"
+import { SavesItem_item$data } from "__generated__/SavesItem_item.graphql"
+
+const ITEM_DROP_SHADOW = "0px 2px 12px rgba(0, 0, 0, 0.13)"
```
dzucconi, line 9: Use one of our shadows or get design to alter it: https://github.com/artsy/palette/blob/main/packages/palette/src/helpers/shadow.ts

src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
dimatretyak: > It doesn't make a difference

oh, good to know. Thanks!

src/Apps/CollectorProfile/Routes/Saves2/Components/SavesItem.tsx
dimatretyak: ok, I'll clarify this point with the designer

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
```diff
@@ -0,0 +1,75 @@
...
+  imageURLs: (string | null)[]
+}
+
+interface StackImageProps {
+  url: string | null
+  index: number
+}
+
+const LARGE_IMAGE_SIZE = 190
+const SMALL_IMAGE_SIZE = 112
```
dzucconi, line 16: I don't like how specific these numbers are. Did design think about if these things scale at all? or how they relate to the grid?

src/Apps/CollectorProfile/Routes/Saves2/Utils/prepareImageURLs.ts
dzucconi: What's the situation in which you'd be expecting to not have images? (I know it's possible but just should be rare enough to not worry too much about it)

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <Box position="absolute" top={OFFSET_BY_INDEX} left={OFFSET_BY_INDEX}>
+      <Image
+        width={SIZE}
+        height={SIZE}
+        src={image.src}
+        srcSet={image.srcSet}
+        aria-label="Image"
+        preventRightClick
```
dzucconi, line 71: I would encourage us to not use this prop. I think the only place we're obligated to use it is on the main artwork page (and even then it's annoying).

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
```diff
@@ -0,0 +1,75 @@
...
+  })
+
+  return (
+    <Box position="absolute" top={OFFSET_BY_INDEX} left={OFFSET_BY_INDEX}>
+      <Image
+        width={SIZE}
+        height={SIZE}
+        src={image.src}
+        srcSet={image.srcSet}
+        aria-label="Image"
```
dzucconi, line 70: I don't think this needs a label. An `alt=""` sure or `role="presentation"`, maybe.

src/Apps/CollectorProfile/Routes/Saves2/Utils/prepareImageURLs.ts
dimatretyak: > I know it's possible but just should be rare enough to not worry too much about it

got it. I think then we don't have to worry about it (I was thinking of adding these changes as a safety net)

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dimatretyak: Clarified this question [here](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB?node-id=308:216953#357768111)

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dimatretyak: Done

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
gkartalis: 🤔 It would be better I think if we got all the data sorted from the backend side

src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <Image
+      width={SIZE}
+      height={SIZE}
+      src={image.src}
+      srcSet={image.srcSet}
+      alt=""
```
gkartalis, line 69: **question(non-blocking):** Should we pass something here? Like the title of the (artwork) image? Asking cause it might have impact on the SEO 🤔 (was reading [this](https://moz.com/learn/seo/alt-text))

gkartalis (commented):
Thanks for the tests ⭐ left one question

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dimatretyak: I don't think there will be any problems when scaling, since we specify specific [sizes](https://github.com/artsy/force/blob/42feb5c09093d6dff3a857e7b44dac9206a459a7/src/Apps/CollectorProfile/Routes/Saves2/Components/SavesItem.tsx#L30-L31) for the rail item

src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
dimatretyak: I don't think we need to worry about this, since this page is only available to logged in users (search bots will not have access to this page)

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
gkartalis: For now it might be okay since users won't have many wishlists but we might want to create a ticket to optimize this later 👌

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dzucconi: Yeah, I think this is fine. I guess in general I'm trying to understand if there's any underlying systematic logic. A lot of the layouts we get have elements size seemingly at random, whereas obviously there should be some site-wide consistency to how we approach this.

src/Apps/CollectorProfile/Routes/Saves2/Components/FourUpImageLayout.tsx
dzucconi: Yeah, no impact on SEO. And the thing to remember about `alt` tags is that you don't want to duplicate any information that's already labelling the node. And so for a lot of our content, images are essentially decorative, since we don't have the ability to describe the actual images. Good `alt` text is **hard**.

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const StackImage: FC<StackImageProps> = ({ url, index }) => {
+  const SIZE = [SMALL_IMAGE_SIZE, LARGE_IMAGE_SIZE]
+  const OFFSET_BY_INDEX = [
+    `${SMALL_IMAGE_OFFSET * index}px`,
+    `${LARGE_IMAGE_OFFSET * index}px`,
+  ]
+
+  if (url === null) {
```
anandaroop, line 42: **question (non-blocking):** Are we still planning to use the [custom icon](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB/Save-%2B-Wishlist?node-id=308%3A120308&t=e0MGCuUtSoRFTish-4) for missing images? If you're planning to add later, all good, just wanted to check. (Same question for the four-up version)

src/Apps/CollectorProfile/Routes/Saves2/Components/SavesItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+  isSelected,
+  imagesLayout,
+  item,
+  onClick,
+}) => {
+  const artworkNodes = extractNodes(item.artworksConnection)
+  const imageURLs = artworkNodes.map(node => node.image?.url ?? null)
+
+  return (
+    <Flex
```
anandaroop, line 26: **question (non-blocking):** Thoughts on keyboard accessibility for these items? (Or does it make more sense to think about that once we actually make these navigable...)

anandaroop (approved):
Looking great! A few thoughts and questions, no blockers…

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
nickskalkin: We could add param like `default_first: true` to the `collectionsConnection`, so that default collections go first, and others follow `sort` option

nickskalkin: created an empty ticket to refine later: https://artsyproduct.atlassian.net/browse/FX-4582

nickskalkin (approved):
All good! Left one tiny question

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -1,6 +1,53 @@
...
+    </Shelf>
+  )
+}
+
+export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContainer(
+  CollectorProfileSaves2Route,
+  {
+    me: graphql`
+      fragment CollectorProfileSaves2Route_me on Me {
+        collectionsConnection(first: 10) {
```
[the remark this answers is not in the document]
dimatretyak: yes, we can do that. now this query is here more for demo purposes and the logic it will be updated in the future in the future PRs

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dimatretyak: I'll clarify this moment with designer

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
dblandin: If we consider "Default Saves" to not be part of the normal collections response, either implicitly or via a query arg, we could query for it separately.

```graphql
{
  me {
    defaultSavesCollection: collection(id: "saved-artworks") {
      name
    }
    collectionsConnection(first: 10) {
      edges {
        node {
          name
        }
      }
    }
  }
}
```

src/Apps/CollectorProfile/Routes/Saves2/Components/StackedImageLayout.tsx
dimatretyak: good catch. I completely missed this point

dimatretyak: Added changes here: https://github.com/artsy/force/pull/11846/commits/52a38c93de6f4477885d3d32811f27c8acdac527

image

src/Apps/CollectorProfile/Routes/Saves2/Components/SavesItem.tsx
dimatretyak: I will add support for this in another PR

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
anandaroop: Yeah, I was thinking something like this too.

Since `saves` and `default` args are [already supported at the Gravity endpoint](https://github.com/artsy/gravity/blob/87acd815b88cbb19dc5839628fb3ba915fe6e005/app/api/v1/collections_endpoint.rb#L11-L12), my feeling is that we could add support for them in MP and leverage those as needed.

I've got [some WIP for MP here](https://github.com/artsy/metaphysics/compare/anandaroop/support-saves-and-default-args?expand=1), and this is what that might look like in action:

wip

PR #11847, opened 2023-01-30
fix: update experiment viewed tracking event

MounirDhahri:

This PR fixes CI breaking because of a tracking schema update. The event wasn't being used anywhere else, so I updated it to match the one on app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "superagent": "3.8.3"
   },
   "dependencies": {
-    "@artsy/cohesion": "^4.96.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -72,7 +72,7 @@
     "superagent": "3.8.3"
   },
   "dependencies": {
-    "@artsy/cohesion": "4.92.0",
+    "@artsy/cohesion": "^4.96.0",
```
dzucconi, line 75: This version should be pinned, actually (remove the `^`)

PR #11849, opened 2023-01-30
fix: Filter out null saved searches GRO-1525

jonallured:
After some exploration that I've documented in the Jira ticket, I'm left scratching my head but I can fix the problem by filtering out nulls. I think that's worth doing, right? I'm ccing the FX team here in case this rings any bells for them...

https://artsyproduct.atlassian.net/browse/GRO-1525

/cc @artsy/grow-devs @artsy/fx-devs

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
@@ -55,7 +55,7 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
   const [showDeleteModal, setShowDeleteModal] = useState(false)
   const [sort, setSort] = useState("CREATED_AT_DESC")
   const [loading, setLoading] = useState(false)
-  const alerts = extractNodes(me.savedSearchesConnection)
+  const alerts = extractNodes(me.savedSearchesConnection).filter(Boolean)
   const isEditMode = editAlertEntity !== null
 
   const closeEditForm = () => {


dzucconi:
This is probably fine but if we have a staff account with a reliable repro, can we just log in to it and see what the issue is? 

Like the problem here is that there's a null being returned in an array of otherwise normal alerts. I'd like to know how that's possible?

anandaroop (approved):
Yep, not blocking this fix or anything, but I (or someone else from FX I'm sure) would be happy to dig in further with Isabelle to see if we can figure out what's up.

dblandin:
Thanks for kicking this discussion off!

tldr: I think we should add a mapping entry [here in Metaphysics](https://github.com/artsy/metaphysics/blob/c3d1fe00438b7f4f8a63fbdc1738682f6818068d/src/schema/v2/searchCriteriaLabel.ts#L155-L170) for "Textile Arts". That should work around the need for this explicit filter here.

I was able to isolate the problem by generating an access token in Gravity (essentially impersonating Isabelle's Artsy user) and reproducing the query using a local GraphQL client (Insomnia).

```graphql
{
  me {
    email
    savedSearchesConnection(first: 10) {
      edges {
        node {
          labels {
            displayValue
          }
        }
      }
    }
  }
}
```

```json
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field SearchCriteriaLabel.displayValue."
    }
  ]
}
```

Note that there is a partial result returned by Metaphysics but it looks like we error pretty hard even if one node in the list is not resolved. Is there an opportunity to degrade a bit more gracefully?

The SearchCriteria type is defined [here in Gravity](https://github.com/artsy/gravity/blob/b82fac491b0b9847619ac079093f2fe326ae0fbc/app/graphql/types/search_criteria_type.rb) but these labels are added [here in Metaphysics](https://github.com/artsy/metaphysics/blob/c3d1fe00438b7f4f8a63fbdc1738682f6818068d/src/schema/v2/searchCriteriaLabel.ts#L63-L67) via stitching code (sorry).

We have a mapping of gene IDs to Medium Type [here in Metaphysics](https://github.com/artsy/metaphysics/blob/c3d1fe00438b7f4f8a63fbdc1738682f6818068d/src/schema/v2/searchCriteriaLabel.ts#L155-L170) which is intended to cover all of the options that would show up in the Medium artwork filter. I'm guessing that we're showing "Textile Arts" as a filter option either in Force or Eigen but because we're missing a mapping entry here, we're unable to fully resolve the filter labels via MP.

dzucconi:
@dblandin — it's more than that; there's missing entries for:

```
Mixed Media
Architecture
Fashion Design and Wearable Art
Textile Arts
Posters
Books and Portfolios
Other
```

Which yes, obviously we'll expand the mapping but I'm not entirely sure how it's expected to keep this list in sync between our services.

dblandin:
@dzucconi As far as I known, we knowingly do not expose all medium types as filter options in collector-facing artwork grids.

[This Notion page](https://www.notion.so/artsy/Adding-new-medium-types-to-Artsy-023c788ef7b1427296fba008d5ad2d74) goes into some context.

I believe this mapping used to accurately describe which filter options were exposed to users. Seems to still be accurate on Force but I'm wondering if the logic in Eigen has changed (maybe via [this PR](https://github.com/artsy/eigen/pull/7056)?) to include more of those partner-facing mediums.

[This comment](https://github.com/artsy/metaphysics/blob/c3d1fe00438b7f4f8a63fbdc1738682f6818068d/src/schema/v2/searchCriteriaLabel.ts#L149-L153) in particular points to the known difference between option in the Medium facet on artwork grids and the full list of artwork medium types as chosen in CMS.

dblandin:
Actually, this might be due to the "Create alert" flow on artwork pages. I'm not sure if we disable or alter the alert criteria when the artwork's medium type is not one of the available filters.

dzucconi:
Interesting, it seems like we should err on the side of inclusivity then though?

jonallured:
After some chatting we decided this was not a good enough solution and have transferred this ticket over to FX: https://artsyproduct.atlassian.net/browse/FX-4578

anandaroop:
> we knowingly do not expose all medium types as filter options in collector-facing artwork grids.
>
> [This Notion page](https://www.notion.so/artsy/Adding-new-medium-types-to-Artsy-023c788ef7b1427296fba008d5ad2d74) goes into some context.

This one might be more illuminating  — more than you ever wanted to know about this mediums-vs-genes indirection:

https://www.notion.so/artsy/Medium-types-vs-medium-filters-39310c0278d64675b755e6db46fec363

Each ❌ in that doc corresponds to the ones highlighted by Damon above. Interestingly there is a little caveat about "Textile Arts" there.  I'll add these details to the new FX ticket.

dzucconi:
😱

PR #11850, opened 2023-01-30
Update tests to match new nav item order

jonallured:
I'm grumpy that somehow https://github.com/artsy/force/pull/11844 was merged when it def failed on CI! This PR gets us back in working order.

/cc @artsy/grow-devs


jonallured:
This is what I get for going too fast - this was opened from my fork. 😕

PR #11858, opened 2023-01-31
fix(passport): fixes linked account redirect

dzucconi:
Surfaced this via Sentry: https://sentry.io/organizations/artsynet/issues/3588012322/?project=28316&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h

I guess the original passport code wasn't in strict mode so this started to error once it was moved over. Converting our passport implementation to TypeScript would be well worth the effort.

### DIFF
--- a/src/Server/passport/lib/app/lifecycle.js
+++ b/src/Server/passport/lib/app/lifecycle.js
@@ -156,8 +156,8 @@ module.exports.afterSocialAuth = provider =>
         err.response.body &&
         err.response.body.error === "Another Account Already Linked"
       ) {
-        alreadyLinkedParams = `?error=already-linked&provider=${providerName}`
-        return res.redirect(opts.settingsPagePath + alreadyLinkedParams)
+        msg = `${providerName} account previously linked to Artsy.`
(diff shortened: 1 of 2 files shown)


src/Server/passport/lib/app/lifecycle.js
```diff
@@ -156,8 +156,9 @@ module.exports.afterSocialAuth = provider =>
         err.response.body &&
         err.response.body.error === "Another Account Already Linked"
       ) {
-        alreadyLinkedParams = `?error=already-linked&provider=${providerName}`
-        return res.redirect(opts.settingsPagePath + alreadyLinkedParams)
+        return res.redirect(
+          `${opts.settingsPagePath}?error=already-linked&provider=${providerName}`
```
dzucconi, line 160: I don't think this redirect is going to actually _do_ anything... but it's better than 500ing...

dzucconi: OK I'll just fix this too.

dzucconi:
OK, now this redirects back to settings with an error param which is then surfaced via toast.

PR #11904, opened 2023-02-08
chore: restore smoke scripts

dzucconi:
Cherry picking @dimatretyak's change

### DIFF
--- a/scripts/smoke_test.sh
+++ b/scripts/smoke_test.sh
@@ -8,8 +8,13 @@ if [ ! -f ".env" ]; then
     cp .env.oss .env
 fi
 
-# start server in the background in development mode
-NODE_ENV=development nohup yarn start &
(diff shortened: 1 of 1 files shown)


scripts/smoke_test.sh
```diff
@@ -8,8 +8,13 @@ if [ ! -f ".env" ]; then
     cp .env.oss .env
 fi
 
-# start server in the background in development mode
-NODE_ENV=development nohup yarn start &
```
damassi, line 12: did this end up blowing stuff up?

dzucconi: Possibly; smoke tests just fail randomly ATM.

damassi: ah. be right back, eating shoe 🚪

PR #11914, opened 2023-02-10
feat(PRO-57): update routing logic for offers

mdole:
Route to order status page if the order is an offer and the offer requires review.

~~Waiting to merge until we've [decided](https://artsy.slack.com/archives/C9YNS4X32/p1676022301689529) if we want to deprecate the feature flag used in this file.~~ Also a drive-by update to remove the `conversational-buy-now` feature flag ([NX-3211](https://artsyproduct.atlassian.net/browse/NX-3211)). See the linked Slack thread for discussion

This PR solves [PRO-57]

[PRO-57]: https://artsyproduct.atlassian.net/browse/PRO-57?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NX-3211]: https://artsyproduct.atlassian.net/browse/NX-3211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -132,20 +132,29 @@ export const ReviewRoute: FC<ReviewProps> = props => {
...
-              })
-            )
-            // We cannot expect Eigen to respond all the time to messages sent from the webview
-            // a default fallback page is safer for old/broken Eigen versions
-            setTimeout(() => {
-              router.push(`/orders/${order.internalID}/status`)
-            }, 500)
-            return
+          if (order.mode === "OFFER") {
+            if (order.state === "IN_REVIEW") {
```
iskounen, line 136: Instead of using the order's state, what do you think about checking whether the order has a conversation ID or not? That way, we can continue to avoid using this state in Force. (We can also avoid refactoring this logic when we introduce a buyer status.)

src/Apps/Conversation/Components/Conversation.tsx
```diff
@@ -217,13 +217,14 @@ const Conversation: React.FC<ConversationProps> = props => {
...
-  // -Workaround Reply render resizing race condition
 
+  // -Workaround Reply render resizing race condition
   useEffect(() => {
     if (initialMount.current) scrollToBottom()
     const rect = scrollContainer.current?.getBoundingClientRect()
     setToastBottom(window.innerHeight - (rect?.bottom ?? 0) + 30)
-  }, [scrollContainer?.current?.clientHeight])
+  }, [scrollContainerCurrentHeight])
+
```
mdole, line 227: I'll come clean - this was to fix one `useEffect` error message, but there are still 5 other ones in this file 😬 I took the `--no-verify` approach since this wasn't related to the change I made and I'm not an effective enough React engineer at this moment to fix these quickly.

src/Apps/Conversation/Components/ConversationCTA.tsx
```diff
@@ -14,13 +14,11 @@ import { ReviewOfferCTA } from "./ReviewOfferCTA"
...
 import { MakeOfferOnInquiryButtonFragmentContainer } from "./MakeOfferOnInquiryButton"
 import { ConversationCTA_conversation$data } from "__generated__/ConversationCTA_conversation.graphql"
-import { useFeatureFlag } from "System/useFeatureFlag"
 import { themeGet } from "@styled-system/theme-get"
 
 interface ConversationCTAProps {
   conversation: ConversationCTA_conversation$data
-
-  openInquiryModal: ({ createsOfferOrder: boolean }) => void
+  openInquiryModal: (obj: { createsOfferOrder: boolean }) => void
```
iskounen, line 21: **question:** The `obj:` before the the curly braces is new in this side of the diff. Is that something added by a linter?

src/Apps/Conversation/Components/Conversation.tsx
damassi: I would say, unless you're certain about what items to go into the effect (to satisfy the linter) its best to avoid. There are circumstances where satisfying the hooks requirements can lead to infinite loops and whatnot. It's a rule that's best in place for new code. 

(Do feel free to fix other linter warnings / errors, however! The other ones are safe.)

src/Apps/Order/Routes/Review/index.tsx
iskounen: OK, I see the limitation of expecting a conversation ID to be available. I also think that we're limited by the fact that Exchange does not report this state back to users. We may need to check the 3 things here, like we have to do in Metaphysics: mode is offer, payment is wire, state is submitted.

src/Apps/Conversation/Components/Conversation.tsx
iskounen: **suggestion:** Should we undo this change if you're unsure about it?

src/Apps/Conversation/Components/ConversationCTA.tsx
mdole: nah this was an intentional change in response to a warning. the type declaration was incorrect before; it was declaring `createsOrderOffer` with type `any` and then aliasing it to the word `boolean`

src/Apps/Conversation/Components/Conversation.tsx
mdole: hm yeah now that I look at this again, pretty sure it should be reverted. I moved something that could change into a const; I think the net result is that even if it changes we wouldn't notice and re-fire the `useEffect` hook unless we also had a rerender.

Not 100% confident in my explanation there, which probably tells us enough 😄 so walked it back

src/Apps/Order/Routes/Review/index.tsx
mdole: good point - done in [acc19f3](https://github.com/artsy/force/pull/11914/commits/acc19f3e2f2c167f414c0ca51ad1ee0cfb6c97a5)

damassi:
This is minor! but note that Danger is flagging the PR title format:
```
thing[FOO]: description
```
should be 
```
thing(FOO): description
```

src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -156,8 +168,7 @@ export const ReviewRoute: FC<ReviewProps> = props => {
         }
 
         // Eigen redirects to the status page for non-Offer orders (must keep the user inside the webview)
-        // TODO: It must be only `if (isEigen) {` after removing the feature flag
```
iskounen, line 159: **praise:** to whomever left this comment!

iskounen (approved):

PR #11915, opened 2023-02-10
feat(auth): pre-fetches location

dzucconi:
This PR adds a `useQuery` hook, which lets us create a more portable query not tied to a specific piece of UI. Then this is used to fetch the `countryCode` for logged out users, in the `AppShell`, by the time it's pulled out in the sign up form, it's most likely cached and so we don't have to refetch it.


src/Apps/Components/AppShell.tsx
```diff
@@ -27,6 +28,7 @@ export const AppShell: React.FC<AppShellProps> = props => {
   useRunAuthIntent()
   useAuthValidation()
   useSocialAuthTracking()
+  useCountryCode()
```
damassi, line 31: More auth stuff 😛 Mentioned this in another PR, but can we roll all of these hooks into a single `useAuth` hook?

src/Components/AuthDialog/Hooks/useCountryCode.ts
```diff
@@ -0,0 +1,35 @@
...
+    requestLocation(ip: $ip) {
+      countryCode
+    }
+  }
+`
+
+export const useCountryCode = () => {
+  const { isLoggedIn } = useSystemContext()
+
+  const { data, loading, error } = useQuery<useCountryCodeQuery>({
```
damassi, line 18: Really nice 👍 gonna bring this over to a few apps today

src/Apps/Components/AppShell.tsx
dzucconi: I honestly don't really buy the argument to do this but I'm not opposed to it.

src/Components/AuthDialog/Hooks/useCountryCode.ts
dzucconi: Wondering if I should call this `useClientQuery`

src/Apps/Components/AppShell.tsx
dzucconi: Will just amend this PR

damassi: There were two auth hooks. then there was another auth related hook. and now another. Clearly, auth is a thing that grows and changes and thus should have a canonical source of reference for all setup-related things! Its general importance as a concept also warrants it. A single file that rolls it all up. Thanks for the update 👍

src/Components/AuthDialog/Hooks/useCountryCode.ts
damassi: hmmm

damassi: Yeah, i like that better actually.

src/Apps/Components/AppShell.tsx
dzucconi: Yeah OK, I'm convinced

dzucconi: Really struggling with the naming here: it's not any kind of setup and all of these things are only loosely related.

`useRunAuthIntent`: picks up any after auth actions that were set before authentication and runs them.
`useAuthValidation`: just checks to see if your current authentication is still valid and logs you out if it's not.
`useSocialAuthTracking`: picks up after social authentication to track a successful login or sign up
`useCountryCode`: pre-fetches the country code for the sign up form.

So one something that happens "before"; the others are things that happen "after".

`useAuthEnhancements`?

damassi: I'd say we just keep it simple with the name. `useSetupAuth` is general enough to contain loosely grouped things. Its also signal enough to add future things to that spot.

PR #11921, opened 2023-02-13
feat(TX-1053): Get versioned artwork image

AlicanAkyuz:

This PR solves [TX-1053]

### Description

This PR gets the artwork image, displayed in the order row in user purchases, from the artwork version, since this section can now include private sale orders.

Additionally, I made the artwork's name un-clickable for private sale orders, since users cannot go to the artwork page. 

web

[TX-1053]: https://artsyproduct.atlassian.net/browse/TX-1053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rquartararo (approved):
Looks good as well 🎉

PR #11956, opened 2023-02-17
fix: route inquiry wire offers to conversations

iskounen:
When an offer-order is initiated from an ongoing conversation, we want the buyer to be redirected to that conversation when the offer-order is submitted. This behavior changed for offer-orders initiated from an artwork page in #11914, and this PR updates that to consider the conversation-initiated offer-orders as well.


mdole (changes requested):
The change looks good, but we should add a spec or update the existing spec to make the expectation explicit! I.e. in t[he existing spec ](https://github.com/artsy/force/blob/dbc90cc89b209c21c6d9e7f8ecb9a6f0de8478bb/src/Apps/Order/Routes/__tests__/Review.jest.tsx#L693-L695) the `source` may be `artwork_page` by default, not sure, would be good to clarify

mdole (approved):
Nice work! 👏 ship it!

PR #11957, opened 2023-02-17
feat(error): cleans up error detail display

dzucconi:
Previously on server-side 500s we would not display the stack traces. On client-side errors we would (unintentionally).

Now:


mzikherman (dismissed):
Nice

kajatiger (approved):
Sweet!

joeyAghion (dismissed):

PR #11970, opened 2023-02-22
fix: event propagation for modal

dimatretyak:

This PR solves [PROJECT-XX]

### Description
This PR includes fix for the problem, which you can see in the demo video. This is due to the fact that the `ModalBase` component [uses](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Modal/ModalBase.tsx#L131) `createPortal` and we are facing [this problem](https://github.com/facebook/react/issues/11387) (example with [details](https://github.com/facebook/react/issues/11387#issuecomment-340127105))

### Problem
When you click on the modal content, the `click` event is propagated to parents and the `onClick` handlers are triggered. Navigation to the artwork page is performed in our case, since we render this modal inside the `Details` component and displays this modal when clicking on the heart icon 

### Demo
https://user-images.githubusercontent.com/3513494/220737267-84aaea3d-aa04-4b80-bb1c-c20669225f4b.mp4


dimatretyak:
Decided to use another approach: https://github.com/artsy/force/pull/11978

PR #11971, opened 2023-02-22
chore(GRO-1561): retires old auth frontend, mediator

dzucconi:
Closes [GRO-1561](https://artsyproduct.atlassian.net/browse/GRO-1561)

This PR removes the rollout related adapter/code and deletes the old auth dialog and associated routes. Additionally I've completely removed the `Mediator` (global event bus) — the only other thing relying on it was logout.

[GRO-1561]: https://artsyproduct.atlassian.net/browse/GRO-1561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Server/mediator.ts
```diff
@@ -1,82 +0,0 @@
-import EventEmitter from "eventemitter3"
```
dzucconi, line 1: We could remove this dependency once we get rid of https://github.com/artsy/force/blob/main/src/Utils/Events.ts — there's some legacy tracking stuff that's still using it apparently.

damassi: How this stuff works is pretty straightforward, if you want to swap it out with some kind of built in event pub sub: https://github.dev/artsy/force/blob/main/src/Apps/Artwork/Components/ArtistInfo.tsx#L37. Obvs, no need to do that here

damassi: And doesn't even really need to be a pub sub.. just call the analytics track event directly. It can be a direct function call, and delete this whole file

dzucconi: Yeah, that's what we'd want to do just update these to use the hooks

damassi (approved):

PR #11977, opened 2023-02-24
fix(CX-3360): swa final adjustments

dariakoko:

This PR solves [CX-3360]

### Description

Addressed QA comments
Can be tested here: https://swa-lp-redesign.artsy.net/sell

[CX-3360]: https://artsyproduct.atlassian.net/browse/CX-3360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
@@ -1,17 +1,22 @@
-import { Banner } from "@artsy/palette"
-import { Media } from "Utils/Responsive"
 import { RouterLink } from "System/Router/RouterLink"
+import { FullBleedBanner } from "Components/FullBleedBanner"
+import { useWindowSize } from "Utils/Hooks/useWindowSize"
+import { breakpoints } from "@artsy/palette"
 
 export const FooterBanner: React.FC = () => {
+  const { width } = useWindowSize()
+
   return (
-    <Media greaterThan="xs">
-      <Banner dismissable={false} variant="defaultLight" mx={[-2, -4]}>
-        Gallery or art dealer?
-        <RouterLink to="https://partners.artsy.net">
-          Become a partner
-        </RouterLink>
-        to access the world’s largest online marketplace.
-      </Banner>
-    </Media>
+    <>
+      {width > parseInt(breakpoints.sm, 10) && (
+        <FullBleedBanner dismissable={false} variant="defaultLight">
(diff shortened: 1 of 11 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
```diff
@@ -1,17 +1,22 @@
...
-      <Banner dismissable={false} variant="defaultLight" mx={[-2, -4]}>
-        Gallery or art dealer?
-        <RouterLink to="https://partners.artsy.net">
-          Become a partner
-        </RouterLink>
-        to access the world’s largest online marketplace.
-      </Banner>
-    </Media>
+    <>
+      {width > parseInt(breakpoints.sm, 10) && (
```
dariakoko, line 11: Using breakpoints because `<Media />` component breaks the banner

dariakoko: We should display the banner only on the web view

MounirDhahri (approved):
the-wolf-of-wall-street-clap

Thanks for creating the review app

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
olerichter00: Hmm, have you found out why this doesn't work with `Media`?

dzucconi: So: I'm assuming it 'breaks' it because of the `FullBleed` wrapper. Though: it's probably fixable and this isn't at all recommended as the way around it. You could do a few things: very simply use responsive props `display={["none", "block"]}` or use the `className` render props `{className => ...}` to merge with the `FullBleed` children.

dzucconi: The use of `useWindowSize` here is going to be breaking server-side rendering, for one.

MounirDhahri: Thanks for raising this @dzucconi - I addressed you comment [here](https://github.com/artsy/force/pull/12120)

PR #11978, opened 2023-02-24
spike: saving artworks to lists on other grids

dimatretyak:
The type of this PR is: **Spike**

Review app: [artwork-saving](https://artwork-saving.artsy.net/collector-profile/saves2)

### Demo for saves page
https://user-images.githubusercontent.com/3513494/221851486-8a2d409b-5248-4db1-8844-78227183418c.mp4

### Demo for artist artworks page
https://user-images.githubusercontent.com/3513494/221851557-7d23ae92-8eed-41bf-b1ab-43e26b9ae5b0.mp4


src/Components/Artwork/ManageArtworkForSaves.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {children}
+      </ManageArtworkForSaves.Provider>
+
+      {!!artworkEntityId && (
+        <SelectListsForArtworkModalQueryRender
+          artworkID={artworkEntityId}
+          onClose={clearArtworkId}
+          onSave={handleSaveListsForArtwork}
+        />
+      )}
```
dimatretyak, line 75: This approach to displaying the modal helps us solve [this](https://github.com/artsy/force/pull/11970) and some other problems

src/Components/Artwork/SaveButton/SaveArtworkToListsButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+export const SaveArtworkToListsButtonFragmentContainer = createFragmentContainer(
+  SaveArtworkToListsButton,
+  {
+    artwork: graphql`
+      fragment SaveArtworkToListsButton_artwork on Artwork {
+        id
+        internalID
+        is_saved: isSaved
```
gkartalis, line 97: I think we can change that to:
```suggestion
        isSaved
```

dimatretyak: if you don't mind, then I think it's better to do it in a separate PR and replace it in all places at once (e.g [here](https://github.com/artsy/force/blob/0177ef9df54e94bf39c8b084bc70320c903cd26a/src/Components/Artwork/FlatGridItem.tsx#L154), [here](https://github.com/artsy/force/blob/0177ef9df54e94bf39c8b084bc70320c903cd26a/src/Components/Artwork/SaveButton/SaveArtworkMutation.ts#L22), [here](https://github.com/artsy/force/blob/0177ef9df54e94bf39c8b084bc70320c903cd26a/src/Components/Artwork/SaveButton/SaveButton.tsx#L108), etc)

dimatretyak: although we can add these changes for this component now (I was afraid that we might break something somewhere. but after investigating the code a bit, I don't see any problems that we may encounter if we use `Issued` in the query)

damassi (commented):
Looks great 👍 My only question would be the toast inside the artwork, vs at the standard bottom right of the page, but that's probably been settled by design.

dzucconi:
@damassi the toast is top center in Force actually

damassi:
Ah, I thought that was only in a few spots, but most positions its at the bottom right (like in our other apps). Hm

anandaroop:
Just earlier today Ryan told me that design team has signed off on a global change to position the toasts in Force towards the bottom right of the viewport.

src/Components/Artwork/Details.tsx
```diff
@@ -246,6 +250,15 @@ export const Details: React.FC<DetailsProps> = ({
       return renderSaveButton(rest.artwork.internalID)
     }
 
+    if (isArtworksListEnabled && enableSaveButtonForLists) {
+      return (
+        <SaveArtworkToListsButtonFragmentContainer
```
anandaroop, line 255: Will this eventually replace the other button on L263?

anandaroop (approved):
LGTM

src/Components/Artwork/Details.tsx
dimatretyak: Yes, that's right

dimatretyak:
> Just earlier today Ryan told me that design team has signed off on a global change to position the toasts in Force towards the bottom right of the viewport

it is useful to know this. thank you for sharing this info!

src/Components/Artwork/ManageArtworkForSaves.tsx
```diff
@@ -0,0 +1,78 @@
...
+  (null as unknown) as ManageArtworkForSavesState
+)
+
+export const useManageArtworkForSavesContext = () => {
+  return useContext(ManageArtworkForSaves)
+}
+
+/**
+ *
+ * If `savedListId` was passed, it means the user is on the saves page
```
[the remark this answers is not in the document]
dimatretyak: Done: https://github.com/artsy/force/pull/11978/commits/6b4854904897736ef1c0959a67d7c2dd664d111f

PR #11983, opened 2023-02-24
chore(cleanup): Delete all dead code 

damassi:

### Description

Following work that @gkartalis did [over here in Eigen](https://github.com/artsy/eigen/pull/8266), this cleans up all the dead code in Force via [`ts-prune`](https://github.com/nadeesha/ts-prune)

Command run: 
```
npx ts-prune --project tsconfig.json | grep -v '(used in module)' | grep -v '__generated__'
```

Will add a command to `package.json` with some common exclude patterns that I noticed along the way so we can run this more regularly. 

Thanks for the amazing lib find @gkartalis, this work felt very good to do.


dzucconi:
Did the absolute path conversion have to happen for this to work? I'm assuming that's the cause of the spec failures?

damassi:
I ended up deleting a bunch of index files just to clean up future ts-prune output and I suspect that broke something -- had to jet today but will fix this in a bit!

damassi:
Yeah, see `subscription.unsubscribe()` in the [docs](https://relay.dev/docs/api-reference/fetch-query/#return-value). All it takes is to use the recommended form; the `toPromise` variant can't be cancelled.

PR #11994, opened 2023-02-27
perf: temporarily disable RelatedWorks rail

anandaroop:
The type of this PR is: **Perf**

This PR solves [PLATFORM-4980]

### Description

Temporarily suppresses the artwork page's `RelatedWorks` rail, which appears to be implicated in the increasing frequency of ES alerts. 

See Slack threads [here](https://artsy.slack.com/archives/CA8SANW3W/p1676999056487519) and [here](https://artsy.slack.com/archives/C9SATFLUU/p1677253352997539).

Effects can be monitored [here](https://artsy.slack.com/archives/C9SATFLUU/p1677533281286779?thread_ts=1677253352.997539&cid=C9SATFLUU).

[PLATFORM-4980]: https://artsyproduct.atlassian.net/browse/PLATFORM-4980?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -31,7 +31,7 @@ import { CascadingEndTimesBannerFragmentContainer } from "Components/CascadingEn
 import { UnlistedArtworkBannerFragmentContainer } from "Components/UnlistedArtworkBanner"
 import { useCallback, useEffect } from "react"
 import { ArtworkSidebarFragmentContainer } from "./Components/ArtworkSidebar/ArtworkSidebar"
-import { RelatedWorksQueryRenderer } from "Apps/Artwork/Components/RelatedWorks"
+// import { RelatedWorksQueryRenderer } from "Apps/Artwork/Components/RelatedWorks"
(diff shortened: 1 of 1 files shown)


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -230,7 +230,7 @@ export const ArtworkApp: React.FC<Props> = props => {
 
       <Spacer y={6} />
 
-      <RelatedWorksQueryRenderer slug={artwork.slug} />
+      {/* <RelatedWorksQueryRenderer slug={artwork.slug} /> */}
```
damassi, line 233: Maybe add a `// TODO` here giving some context?

anandaroop: Unclear what the TODO is! Need to confirm this has the desired effect, then [revisit](https://artsy.slack.com/archives/CA8SANW3W/p1677177993873459?thread_ts=1676999056.487519&cid=CA8SANW3W) for a solution. We should know more very soon.

damassi: np! Just didn't want this comment to be lost later should other things get picked up. Something like "This line seems to impact performance in ES" would be sufficient.

joeyAghion (dismissed):

PR #12005, opened 2023-02-28
feat(GRO-1559): implements progressive onboarding tips for artwork saves

dzucconi:
Closes [GRO-1559](https://artsyproduct.atlassian.net/browse/GRO-1559)

I'm keeping this behind [the same feature flag](https://unleash.artsy.net/projects/default/features/progressive-onboarding-artist) — though we can split this up if need be.

### DIFF
--- a/src/Components/Artwork/SaveButton/useSaveArtwork.ts
+++ b/src/Components/Artwork/SaveButton/useSaveArtwork.ts
@@ -40,6 +40,20 @@ export const useSaveArtwork = ({
                 slug: artwork.slug!,
                 is_saved: !isSaved,
               },
+              /**
+               * TODO: We don't _really_ need an optimistic response and
+               * it's too painful to alter all the consumers of this to pass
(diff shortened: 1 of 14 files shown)


src/Components/Artwork/SaveButton/useSaveArtwork.ts
```diff
@@ -40,6 +40,20 @@ export const useSaveArtwork = ({
...
+               *
+               * This hook shouldn't have to accept a stub `artwork` object either.
+               * Given just the artwork `id` we could fetch everything else.
+               */
+              me: {
+                id: "me",
+                counts: {
+                  savedArtworks: 0,
+                },
+              },
```
dzucconi, line 56: What this means in practice is that there's a slight delay before the tip disappears; let's see how it feels in Q/A. I can fix this it's just a bit of a pain considering the state of these components.

TMcMeans (approved):

PR #12008, opened 2023-03-01
feat: use order displayState instead of state

mdole:

### Description

Companion PR to https://github.com/artsy/eigen/pull/8275. Pretty sure this should cover all of the places we're using order `state` to determine what to display/how to route users where having state be `in_review` might cause things to break.

### DIFF
--- a/src/Apps/Order/redirects.tsx
+++ b/src/Apps/Order/redirects.tsx
@@ -110,8 +110,10 @@ const goToStatusIfNotAwaitingBuyerResponse = goToStatusIf(
   "Not currently awaiting buyer response"
 )
 
+// displayState is used here since an order may be in_review, in which case
+// it is still "submitted" to the user and we shouldn't redirect
 const goToStatusIfOrderIsNotSubmitted = goToStatusIf(
-  order => order.state !== "SUBMITTED",
+  order => order.displayState !== "SUBMITTED",
   "Order was not yet submitted"
 )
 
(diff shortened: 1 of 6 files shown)


src/Apps/Order/redirects.tsx
```diff
@@ -258,6 +260,7 @@ graphql`
     internalID
     mode
     state
```
iskounen, line 262: **question:** Can we remove this field, or does Force still need to know about it?

iskounen: I forgot to check the tests before writing this comment - it makes sense now that you're using both `state` and `displayState`.

iskounen (dismissed):

mdole:
@iskounen might as well! I had originally left that + one other instance where it _kinda_ made sense to use `state`, but `displayState` also works. Just updated.

I find it a little confusing now to know when we should use `state` vs. `displayState` 🤔 like I guess Force should mostly rely on `displayState`, especially any time we might need to handle `in_review` orders. But `displayState` also folds in some shipping logic...

For now, I don't think we're at too much risk of shooting ourselves in the feet, but something to keep an eye on for sure.

mdole:
We discussed on [Slack](https://artsy.slack.com/archives/C047EHC5CSK/p1678448408172429?thread_ts=1678446811.047119&cid=C047EHC5CSK) and via Zoom and decided to keep this simple for now and only change the places we knew we needed to make a change.

PR #12020, opened 2023-03-02
feat(artwork lists): add the edit action

anandaroop:

This PR solves [FX-4552]

Review app: https://edit-wishlist.artsy.net/collector-profile/saves2

### Description

Fleshes out the Artwork List area's contextual menu by implementing the "Edit list" functionality

### Desktop

https://user-images.githubusercontent.com/140521/222452515-c5889bc9-8a71-4005-9706-b25db64037a9.mov

### mWeb

https://user-images.githubusercontent.com/140521/222452528-e0e56968-5f94-4c5d-8f33-af0e5348c373.mov

[FX-4552]: https://artsyproduct.atlassian.net/browse/FX-4552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
@@ -90,9 +90,11 @@ export const EditSavesModal: React.FC<EditSavesModalProps> = ({
         values,
         errors,
         touched,
+        dirty,
         handleBlur,
         handleChange,
         isSubmitting,
+        isValid,
       }) => {
         return (
           <ModalDialog
@@ -124,7 +126,12 @@ export const EditSavesModal: React.FC<EditSavesModalProps> = ({
                 justifyContent={["space-between"]}
                 flexDirection={["column", "row-reverse"]}
               >
-                <Button type="submit" mb={[1, 0]} loading={!!isSubmitting}>
+                <Button
+                  type="submit"
+                  mb={[1, 0]}
+                  loading={!!isSubmitting}
+                  disabled={!dirty || !isValid}
+                >
                   {t("collectorSaves.editListModal.save")}
                 </Button>
 
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
@@ -1,0 +1,119 @@
+import {
+  Button,
+  Flex,
+  LabeledInput,
+  ModalDialog,
+  Spacer,
+  Text,
+} from "@artsy/palette"
+import EditIcon from "@artsy/icons/EditIcon"
+import { useTranslation } from "react-i18next"
+import { SavesArtworks_collection$data } from "__generated__/SavesArtworks_collection.graphql"
+import { Form, Formik, FormikHelpers } from "formik"
+import * as Yup from "yup"
+import { pick } from "lodash"
+
+interface EditSavesModalProps {
+  collection: SavesArtworks_collection$data
+  setIsEditModalOpen: React.Dispatch<React.SetStateAction<boolean>>
+}
+
+interface FormikValues {
+  name: string
+}
+
+const MAX_NAME_LENGTH = 40
+
+export const EditSavesModal: React.FC<EditSavesModalProps> = ({
+  collection,
+  setIsEditModalOpen,
+}) => {
+  const { t } = useTranslation()
+
+  const initialValues = pick(collection, "name")
+
+  const validationSchema = Yup.object().shape({
+    name: Yup.string()
+      .required(t("collectorSaves.editListModal.fields.name.required")) // TODO: confirm copy
+      .max(MAX_NAME_LENGTH),
+  })
+
+  const handleSubmit = (
+    values: FormikValues,
+    helpers: FormikHelpers<FormikValues>
+  ) => {
+    // TODO: replace placeholder with real mutation
+    setTimeout(() => {
+      alert("TK " + JSON.stringify(values))
+      helpers.setSubmitting(false)
+      setIsEditModalOpen(false)
+    }, 500)
+  }
+
+  return (
+    <Formik<FormikValues>
+      initialValues={initialValues}
+      validationSchema={validationSchema}
+      onSubmit={handleSubmit}
+    >
+      {({
+        values,
+        errors,
+        touched,
+        handleBlur,
+        handleChange,
+        isSubmitting,
+      }) => {
+        return (
+          <ModalDialog
+            title={t("collectorSaves.editListModal.title")}
+            onClose={() => {
+              setIsEditModalOpen(false)
+            }}
+            width={["100%", 700]}
+          >
+            <Form /* TODO: confirm implicit form submission */>
+              <LabeledInput
+                name="name"
+                value={values.name}
+                title={t("collectorSaves.editListModal.fields.name.label")}
+                label={<EditIcon />}
+                error={touched.name && errors.name}
+                maxLength={MAX_NAME_LENGTH}
+                required // TODO: confirm
+                onChange={handleChange}
+                onBlur={handleBlur}
+              />
+              <Spacer y={1} />
+              <Text variant="xs">
+                {t(
+                  "collectorSaves.createNewListModal.remainingCharactersCount",
+                  { count: MAX_NAME_LENGTH - values.name.length }
+                )}
+              </Text>
+              <Spacer y={2} />
+              <Flex
+                justifyContent={["space-between"]}
+                flexDirection={["column", "row-reverse"]}
+              >
+                <Button type="submit" mb={[1, 0]} loading={!!isSubmitting}>
+                  {t("collectorSaves.editListModal.save")}
+                </Button>
+
+                <Button
+                  variant="secondaryBlack"
+                  onClick={() => setIsEditModalOpen(false)}
+                  display={["block", "none"]} // TODO: intentionally hidden on desktop?
+                >
+                  {
+                    t("collectorSaves.editListModal.back") // TODO: "back" vs "cancel"?
+                  }
+                </Button>
+              </Flex>
+            </Form>
+          </ModalDialog>
+        )
+      }}
+    </Formik>
+  )
+}
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
@@ -18,7 +18,7 @@ import createLogger from "Utils/logger"
 
 interface EditSavesModalProps {
   collection: SavesArtworks_collection$data
-  setIsEditModalOpen: React.Dispatch<React.SetStateAction<boolean>>
+  onClose: () => void
 }
 
 interface FormikValues {
@@ -31,7 +31,7 @@ const logger = createLogger("EditSavesModal")
 
 export const EditSavesModal: React.FC<EditSavesModalProps> = ({
   collection,
-  setIsEditModalOpen,
+  onClose,
 }) => {
   const { t } = useTranslation()
 
@@ -80,7 +80,7 @@ export const EditSavesModal: React.FC<EditSavesModalProps> = ({
       })
     } finally {
       formikHelpers.setSubmitting(false)
-      setIsEditModalOpen(false)
+      onClose()
     }
   }
 
@@ -101,9 +101,7 @@ export const EditSavesModal: React.FC<EditSavesModalProps> = ({
         return (
           <ModalDialog
             title={t("collectorSaves.editListModal.title")}
-            onClose={() => {
-              setIsEditModalOpen(false)
-            }}
+            onClose={onClose}
             width={["100%", 700]}
           >
             <Form /* TODO: confirm implicit form submission */>
@@ -136,7 +134,7 @@ export const EditSavesModal: React.FC<EditSavesModalProps> = ({
 
                 <Button
                   variant="secondaryBlack"
-                  onClick={() => setIsEditModalOpen(false)}
+                  onClick={onClose}
                   display={["block", "none"]} // TODO: intentionally hidden on desktop?
                 >
                   {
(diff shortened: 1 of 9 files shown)


src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { SavesArtworks_collection$data } from "__generated__/SavesArtworks_collection.graphql"
+import { Form, Formik, FormikHelpers } from "formik"
+import * as Yup from "yup"
+import { pick } from "lodash"
+import { useMutation } from "Utils/Hooks/useMutation"
+import { graphql } from "react-relay"
+import { EditSavesModalMutation } from "__generated__/EditSavesModalMutation.graphql"
+import createLogger from "Utils/logger"
+
+interface Props {
```
damassi, line 21: This is #superminor - since props map directly to the component name, might as well keep things 1:1 -- `EditSavesModalProps`. then searches for `EditSavesModal` yield searches for the other and it remains simpler.

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  const initialValues = pick(collection, "name")
+
+  const validationSchema = Yup.object().shape({
+    name: Yup.string()
+      .required(t("collectorSaves.editListModal.fields.name.required")) // TODO: confirm copy
+      .max(MAX_NAME_LENGTH),
+  })
+
+  const { submitMutation } = useMutation<EditSavesModalMutation>({
```
damassi, line 48: The pattern we've started adopting around all of our relay codebases follows this pretty closely: https://github.com/artsy/force/blob/main/src/Apps/MyCollection/Routes/EditArtwork/Mutations/useCreateArtwork.ts#L5; we define it in a hook in its own file.

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useUpdateCollection.ts
```diff
@@ -0,0 +1,29 @@
...
+      mutation EditSavesModalMutation($input: updateCollectionInput!) {
+        updateCollection(input: $input) {
+          responseOrError {
+            __typename
+            ... on UpdateCollectionSuccess {
+              collection {
+                internalID
+                name
+              }
+            }
```
anandaroop, line 17: @dimatretyak I found that the the Artwork Lists rail gets an updated immediately thanks to the Relay store, as we discussed. Is that because the `name` is returned here by the mutation? 

(In other words: not an optimistic update, but an after-the-fact update based on the actual mutation result?)

anandaroop (commented):
I have a few small design Qs to double-check with Ryan but this is otherwise ready for review.

A few things I'll briefly note here but raise in Slack, and then revisit as needed.

- We have some duplication & divergence between the two forms (create vs edit). Would be nice to consolidate that logic. 

- There are some form-related best practices I'm wondering about as well, now that I'm back to writing forms for the first time in a long while!

- I'm seeing great adoption of i18n for shared strings, might be time for some more opinionated patterns to keep the shared strings from becoming too messy.

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useUpdateCollection.ts
dimatretyak: Yes, that's right

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                  {t("collectorSaves.editListModal.save")}
+                </Button>
+
+                <Button
+                  variant="secondaryBlack"
+                  onClick={() => setIsEditModalOpen(false)}
+                  display={["block", "none"]} // TODO: intentionally hidden on desktop?
+                >
+                  {
+                    t("collectorSaves.editListModal.back") // TODO: "back" vs "cancel"?
```
dimatretyak, line 143: Asked it [here](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB?node-id=630:162309#383926378). But IMHO it's better to use "Cancel" here

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                flexDirection={["column", "row-reverse"]}
+              >
+                <Button type="submit" mb={[1, 0]} loading={!!isSubmitting}>
+                  {t("collectorSaves.editListModal.save")}
+                </Button>
+
+                <Button
+                  variant="secondaryBlack"
+                  onClick={() => setIsEditModalOpen(false)}
+                  display={["block", "none"]} // TODO: intentionally hidden on desktop?
```
dimatretyak, line 140: As far as I can see by design, then yes, this is the desired behavior when this button will be hidden for desktop
Untitled – Figma 2023-03-03 14-09-00

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/__tests__/EditSavesModal.jest.tsx
```diff
@@ -0,0 +1,79 @@
...
+      expect.objectContaining({
+        variables: {
+          input: {
+            id: "foobar",
+            name: "Foo Bar",
+          },
+        },
+      })
+    )
+  })
```
dimatretyak, line 78: **praise:** nice test case 👌

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                name="name"
+                value={values.name}
+                title={t("collectorSaves.editListModal.fields.name.label")}
+                label={<EditIcon />}
+                error={touched.name && errors.name}
+                maxLength={MAX_NAME_LENGTH}
+                required // TODO: confirm
+                onChange={handleChange}
+                onBlur={handleBlur}
+              />
```
dimatretyak, line 120: I think we should trim spaces in the collection name. otherwise, this situation may happen. the same should be done for create list flow (cc: @nickskalkin)

https://user-images.githubusercontent.com/3513494/222706862-bce9d725-142c-461d-88be-36aaf8b38da6.mp4

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                )}
+              </Text>
+              <Spacer y={2} />
+              <Flex
+                justifyContent={["space-between"]}
+                flexDirection={["column", "row-reverse"]}
+              >
+                <Button type="submit" mb={[1, 0]} loading={!!isSubmitting}>
+                  {t("collectorSaves.editListModal.save")}
+                </Button>
```
dimatretyak, line 135: **suggestion**: Disable the "Save" button until the user has changed the name. We can use `dirty` value from formik for this (like [here](https://github.com/artsy/force/blob/a6331ebbe2687a36d87a848fc585db7167f6c878/src/Components/AuthDialog/Views/AuthDialogForgotPassword.tsx#L52))

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+        handleChange,
+        isSubmitting,
+      }) => {
+        return (
+          <ModalDialog
+            title={t("collectorSaves.editListModal.title")}
+            onClose={() => {
+              setIsEditModalOpen(false)
+            }}
+            width={["100%", 700]}
```
dimatretyak, line 107: I think we should open the modal window to the full screen (similar to the "create alert" modal. asked it [here](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB?node-id=466:149379#383949925)). To achieve this, we need to pass `dialogProps` and `m={0}` (like [here](https://github.com/artsy/force/blob/a6331ebbe2687a36d87a848fc585db7167f6c878/src/Apps/CollectorProfile/Routes/Saves2/Components/SelectListsForArtworkModal/SelectListsForArtworkModal.tsx#L177-L182)). Also we need to display the buttons at the bottom

| Before | After |
| before | after |

dimatretyak (dismissed):

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+            name: formikValues.name,
+          },
+        },
+        rejectIf: res => {
+          const result = res.updateCollection?.responseOrError
+
+          return result?.__typename === "UpdateCollectionFailure"
+            ? result.mutationError
+            : false
+        },
```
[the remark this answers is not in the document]
anandaroop: I see two problems with that approach:

- The smaller problem is that the above code doesn't pass TS, without some more type narrowing perhaps

- The bigger problem is that this would return a boolean instead of the actual error message, and thus swallows a potentially helpful user-facing error message

|With `!!responseOrError.mutationError` — <br/>generic error message|With `responseOrError.result.mutationError` — <br/>actual error message|
|---|---|
|bool|msg|

I actually think we should switch to _this_ error-handling pattern more broadly, so that we don't hide useful error messages.

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
anandaroop: Interesting, I'll wait for Ryan to chime in before making a change. 

(Currently, this works the same as the Create modal — we should change them all if this is approved.)

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+                onChange={handleChange}
+                onBlur={handleBlur}
+              />
+              <Spacer y={1} />
+              <Text variant="xs">
+                {t(
+                  "collectorSaves.createNewListModal.remainingCharactersCount",
+                  { count: MAX_NAME_LENGTH - values.name.length }
+                )}
+              </Text>
```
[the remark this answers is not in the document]
anandaroop: Agree that this should be shared, but I'll hold off on this change for now. 

I'm actually interested in even _more_ extensive sharing between the create and edit flow, so I'd like to address this all at once later.

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
anandaroop: I'm confirming [here](https://artsy.slack.com/archives/C9SATFLUU/p1677803444831519?thread_ts=1677803388.476209&cid=C9SATFLUU) (since this seems to differ from the Delete flow)

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
anandaroop: Heh, I [asked](https://artsy.slack.com/archives/C9SATFLUU/p1677803444831519?thread_ts=1677803388.476209&cid=C9SATFLUU) as well. Will wait for Ryan.

anandaroop (commented):
Thanks for the review!

I've responded to a couple of comments below, and preparing some commits now as well...

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
dimatretyak: Oh, sorry. didn't notice this message from you. Mark as resolved

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+  name: string
+}
+
+const MAX_NAME_LENGTH = 40
+
+const logger = createLogger("EditSavesModal")
+
+export const EditSavesModal: React.FC<EditSavesModalProps> = ({
+  collection,
+  setIsEditModalOpen,
```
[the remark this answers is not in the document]
anandaroop: Added in 8ed6849b8e

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+        message: t("collectorSaves.editListModal.success"),
+      })
+    } catch (error) {
+      logger.error(error)
+      sendToast({
+        variant: "error",
+        message: error.message ?? t("common.errors.somethingWentWrong"),
+      })
+    } finally {
+      formikHelpers.setSubmitting(false)
```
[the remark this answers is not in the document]
anandaroop: Added in 687d671a40

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
anandaroop: Added in 61cb84bf73

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/EditSavesModal.tsx
anandaroop: Added in 34534d6e97

anandaroop (commented):
Changes are addressed now

PR #7949, opened 2021-07-09
fix(PX-4310): artwork loading limit

kiryl-zubarau:
This PR fixes limited artwork loading and adds batch loading for the artwork carousel.

JIRA - https://artsyproduct.atlassian.net/browse/PX-4310

<details>
  <summary>Desktop</summary>

  https://user-images.githubusercontent.com/79979820/125110145-e2636300-e0ec-11eb-9769-dee05fdd1732.mov
</details>

<details>
  <summary>Mobile</summary>

  https://user-images.githubusercontent.com/79979820/125110534-5b62ba80-e0ed-11eb-941a-6ddcac76f1f4.mov
</details>

PR #12026, opened 2023-03-03
fix: PX-5458 - increase the number of images loaded in the partner artist rail

Mitch-Henson:

This PR solves [PX-5458]

### Description

Increases the number of images loaded to 99. This is because pagination was removed, causing feedback from partners. Shelf is currently unable to support dynamic pagination and Carousel is being deprecated. More info can be found in [this slack thread](https://artsy.slack.com/archives/CP9P4KR35/p1677777835014889).

Relevant PRs
- https://github.com/artsy/force/pull/7949
- https://github.com/artsy/force/pull/11276

https://user-images.githubusercontent.com/6280410/222759805-3698ca91-d42c-49b2-93e1-53a6eb923027.mp4


dzucconi:
I don't really think this is an acceptable solution to the issue. This loads too much data and renders too much DOM. The underlying problem should be solved: if you want to view _all_ of the works from an artist at a given then we should include a "View All" link to a page that lets you do that with filters which is an acceptable browsing experience.

`https://www.artsy.net/collect?artist_ids%5B0%5D=ARTIST_ID_HERE&partner_ids%5B0%5D=PARTNER_ID_HERE`

Mitch-Henson:
@dzucconi Thanks for the feedback! This was definitely a far from ideal solution and was merely a quick fix intended to solve the issues caused by the unannounced premature removal of an expected feature which caused numerous complaints to be raised by partners. I was relying on 2 things, the first being that not every artist has 99 artworks and secondly that the rails are dynamically loaded (as are the images in the rail I believe).

I have since added the view all button and am simultaneously asking for code review here as I am asking for design approval. This button will take collectors away from the partner page and make it easier for them to discover artworks not sold by the initial partner so we will see what the response is.

Mitch-Henson:
The acceptance check is failing on main and is unrelated to these PR changes. More info can be found in [this slack thread](https://artsy.slack.com/archives/CP9P4KR35/p1678116606765429).

mdole:
I'm :ceiling-cat: here so please feel free to ignore, but what about linking to the partner's artworks page filtered to that artist instead of the /collect page? Feels less jarring to me since then we're staying within the partner surface.

Example page for the partner + artist in your gif: https://www.artsy.net/partner/mla-gallery/works?artist_ids%5B0%5D=wifredo-lam

Mitch-Henson:
Thanks @mdole, that is a much better solution! Just updated

jpotts244 (approved):

PR #12038, opened 2023-03-06
feat(devx): add package script to pretty-print i18n translations

anandaroop:

### Description

Adds a little devx feature that I was missing while reviewing a PR just now.

We are seeing increasing adoption of the i18n pattern in Force. That adds some indirection/friction to the process of identifying a component by its user-facing labels, and then locating its source code.

This PR aims to make things easier by adding a `yarn translations` command that exposes the fully qualified path names for each i18n string, so that you can search for e.g. `collectorSaves.savesHeader.listCreated` and find the component in question.

Output is designed to be very skimmable by default:

basic

And also grep-friendly:

grep3

I skimmed https://www.i18next.com and didn't see anything like this included by default, but correct me if I'm wrong.

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -32,7 +32,7 @@ const { t } = useTranslation()
 ### Adding a static string
 
 1. Add the string to [translation.json](https://github.com/artsy/force/blob/main/src/System/i18n/locales/en-US/translation.json) following the Naming conventions.
-2. Use it in any component by calling ``{t`<path-for-the-key>`}`` (example: ``{t`navbar.shows`}``)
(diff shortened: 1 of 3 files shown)


damassi (approved):
Really nice @anandaroop

lidimayra:
So cool!!! 😻 👏 
Shall we also mention this on [docs/i18n](https://github.com/artsy/force/blob/main/docs/i18n.md)? Maybe just linking this PR there?

anandaroop:
Good idea, I added this to the i18n documentation:

Screen Shot 2023-03-07 at 9 49 13 AM

PR #12041, opened 2023-03-07
feat: modify relay store when artwork was saved/unsaved

dimatretyak:

This PR solves [PROJECT-XX]

### Description
* When artwork was saved only in "All saves" and the 💙 clicked, we should unsave artwork
* When artwork was saved only custom lists and the 💙 clicked, we should display the manage lists for artwork modal


src/Apps/CollectorProfile/Routes/Saves2/Components/SelectListsForArtworkModal/useUpdateCollectionsForArtwork.ts
```diff
@@ -29,5 +50,69 @@ export const useUpdateCollectionsForArtwork = () => {
...
+    custom: 0,
+    default: 0,
+  }
+
+  return listEntities.reduce((acc, list) => {
+    if (!list) {
+      return acc
+    }
+
+    const isDefault = list.name === "Saved Artwork"
```
dimatretyak, line 109: TODO Mention people
do we have the ability to return predefined `internalId` for "All Saves" list? For example, `all-saves'

dimatretyak: or is it better to do check in this way (if we can't add the changes 👆)

```javascript
// src/Apps/CollectorProfile/Routes/Saves2/Components/SelectListsForArtworkModal/SelectListsForArtworkModal.tsx
const savedCollection = me?.defaultSaves
const collections = savedCollection ? [savedCollection, ...nodes] : nodes
const { submitMutation } = useUpdateCollectionsForArtwork({
  artworkID: state.artwork!.id,
  allSavesListId: me?.defaultSaves?.internalID // 👈
})

// src/Apps/CollectorProfile/Routes/Saves2/Components/SelectListsForArtworkModal/useUpdateCollectionsForArtwork.ts
const addedCounts = getCountsByLists(response?.addedToCollections, allSavesListId)
const removedCounts = getCountsByLists(response?.removedFromCollections, allSavesListId)
...
const getCountsByLists = (lists: ListEntity) => {
  ...

  return listEntities.reduce((acc, list) => {
    ...
    const isDefault = list.internalID === allSavesListId
    ...
  }, defaultCounts)
}
```

dblandin: Can we use the default+saves boolean attributes instead of matching on the list name?

The default "All Saves" (currently "Saved Artwork") collection should consistently have `default=true` and `saves=false`.

```graphql
{
  me {
    collectionsConnection(first: 1, default: true, saves: false) {
      edges {
        node {
          name
          default
          saves
        }
      }
    }
  }
}
```

dimatretyak: totally forgot about these fields

dimatretyak: Addressed here: https://github.com/artsy/force/pull/12041/commits/e364c42066e4f296bc5a922faa247381f5ee6399

nickskalkin (dismissed):

src/Apps/CollectorProfile/Routes/Saves2/Components/SelectListsForArtworkModal/useUpdateCollectionsForArtwork.ts
```diff
@@ -29,5 +50,68 @@ export const useUpdateCollectionsForArtwork = () => {
         }
       }
     `,
+    updater: (store, data) => {
+      const artwork = store.get(artworkID)
```
anandaroop, line 54: Nice, TIL

PR #12045, opened 2023-03-07
fix(TX-1071): improve checkout flow tab accessibility

rquartararo:

This PR solves [TX-1071]

### Description

This PR fixes the tab selection for the checkout flow as the selectable fields inside the `Collapse` components can be selected even when they are hidden. 

This set a conditional `tabIndex` for input and buttons inside relevant Collapses. For Stripe components that don't support `tabIndex`, I set `display: "none"` when they are hidden, disabling the tab selection. 

More discussion on this issue[ here](https://artsy.slack.com/archives/C2TQ4PT8R/p1676472354876629). 

[TX-1071]: https://artsyproduct.atlassian.net/browse/TX-1071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/CreditCardPicker.tsx
+++ b/src/Apps/Order/Components/CreditCardPicker.tsx
@@ -339,6 +339,7 @@ export class CreditCardPicker extends React.Component<
             <Collapse open={this.needsAddress()}>
               <Spacer y={2} />
               <AddressForm
+                tabIndex={this.needsAddress() ? 0 : -1}
                 value={address}
                 errors={addressErrors}
                 touched={addressTouched}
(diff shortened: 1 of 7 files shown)


src/Apps/Order/Components/CreditCardPicker.tsx
```diff
@@ -339,6 +339,7 @@ export class CreditCardPicker extends React.Component<
             <Collapse open={this.needsAddress()}>
```
rquartararo, line 339: The `CreditCardPicker` needs to be converted to a functional component. I'm going to create a separate ticket for this.

AlicanAkyuz (approved):

PR #12054, opened 2023-03-09
feat(FX-4662): Saving an Artwork at Artwork detail level

dimatretyak:

This PR solves [FX-4662]


src/Components/Artwork/SaveButton/SaveArtworkToListsButton.tsx
```diff
@@ -18,79 +15,19 @@ const SaveArtworkToListsButton: FC<SaveArtworkToListsButtonProps> = ({
...
-      message: t(
-        "collectorSaves.saveArtworkToListsButton.artworkRemovedToast.message"
-      ),
-    })
-  }
-
-  const { handleSave } = useSaveArtwork({
-    isSaved,
+  const { state, savedListId } = useManageArtworkForSavesContext()
+  const { isSaved, onSave } = useSaveArtworkToLists({
```
dimatretyak, line 19: what do you think, maybe instead of using a custom hook, use [this approach](https://github.com/artsy/force/commit/e909c3054bccd1bd887cdf5ae4bfaea02b6fce55)? Advantages of this approach:
* all the logic is in one place
* all required query fields are requested in one place
* no duplication of logic

Disadvantages of the approach:
* it is more difficult to understand the code
*

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveToListsButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+    artwork,
+    contextModule: ContextModule.artworkImage,
+    customListsCount: artwork.customCollections?.totalCount ?? 0,
+    isSaved: !!artwork.isSaved,
+  })
+
+  return <SaveUtilButton isSaved={isSaved} onClick={onSave} />
+}
+
+export const ArtworkActionsSaveToListsButtonFragmentContainer = createFragmentContainer(
```
nickskalkin, line 25: hey @dimatretyak, just wanted to share my 2 cents, even though I don't know how I would do it better. My concern is that some data, like `title` for example is requested from here, but it is not explicitly used in this file, but in the custom hook. I am wondering if there is a way to declare which data is needed for custom hook functionality in `useSaveArtworkToLists` file itself. This way the code will become less error-prone I believe

nickskalkin (approved):
LGTM, left one non-blocking question

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveToListsButton.tsx
dimatretyak: > declare which data is needed

Here it describes what data is needed for `useSaveArtworkToLists` and if the user does not pass some data, then this will be reported (see screenshot).

Yes, I understand that this is not the best approach. Ideally, we should query this data in the hook itself (for example, using `useFragmentContainer`), but so far relay hooks are not used in Force. It is for this reason that a [second approach](https://github.com/artsy/force/pull/12054#discussion_r1132347013) has been added that solves this problem.
image

src/Components/Artwork/SaveButton/SaveArtworkToListsButton.tsx
anandaroop: Makes sense to me to have a hook, in order to share this behavior between the artwork detail actions as well as the artwork grid brick.

anandaroop (approved):

PR #12076, opened 2023-03-13
feat: add delete collection updater and redirect

gkartalis:

This PR solves [FX-4664]

### Description

- [x] delete collection updater and redirects to `router.replace("/collector-profile/saves2")` ([demo here](https://www.loom.com/share/aa616963b4384d5bb861742a192eb9d1))
- [x] create new list updater and updates ui ([demo here](https://www.loom.com/share/be3d70f27a4043799ce3a0701dfc1a72))

[FX-4664]: https://artsyproduct.atlassian.net/browse/FX-4664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -48,6 +48,25 @@ const logger = createLogger(
...
+
+//     const mutationPayload = store.getRootField("createCollection")
+
+//     const responseOrError = mutationPayload.getLinkedRecord("responseOrError")
+
+//     ConnectionHandler.insertEdgeAfter(otherSavesConnection!, responseOrError)
+
+//   }
+// }
+
```
gkartalis, line 69: wip updater not exactly 100% working

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -76,6 +104,7 @@ export const CreateNewListModal: React.FC<CreateNewListModalProps> = ({
         }
       }
     `,
+    // updater: onListAdded,
```
gkartalis, line 107: no need to check that for now, if we manage to get `@prependNode` directive to work we won't need this

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -90,9 +119,24 @@ export const CreateNewListModal: React.FC<CreateNewListModalProps> = ({
...
+
+    // console.log("[connectionID]", connectionID)
+
     try {
       const { createCollection } = await submitMutation({
-        variables: { input: { name: values.name } },
+        variables: {
+          input: { name: values.name },
+          // @ts-expect-error
+          connections: [connectionID],
```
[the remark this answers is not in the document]
gkartalis: the problem is that we are not exactly able to get the `"-hardcoded-Key-of-the me",` here because we don't have direct access to the `store` we have access to the `relayEnvironment.getStore()` but it is of type `Store` while the `store` coming from updaters i.e. has type of `RecordSourceSelectorProxy`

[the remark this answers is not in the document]
gkartalis: when I hardcoded the `"-hardcoded-Key-of-the me"` it works (but returns also an error) found the value of it with the help of relay dev tools
Screenshot 2023-03-13 at 18 30 01

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
dimatretyak: I think we should use the `updater` approach, since the user will have the opportunity to create a new list from the "[Select lists for this artwork](https://www.figma.com/file/wSV2KNz97Tyj8HkH2ZYAkB/Save-%2B-Lists?node-id=308%3A123937&t=3E13u44ClK91lojk-0)" modal (and for this reason it will be more difficult to get the `__id` from [connection](https://github.com/artsy/force/blob/f8e5c5fe4e5285cc7c08c96b4070c4d6cacf825d/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx#L90-L93) and further use of this id in `@prependNode`)

gkartalis: amazing, then I am going full on the updater, thanks for looking into it

dimatretyak: Some investigations: https://github.com/artsy/force/commit/ef04456e28265e2775a88939292da72f4fdb9e36

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -48,7 +47,38 @@ const logger = createLogger(
 
 const MAX_NAME_LENGTH = 40
 
-export const CreateNewListModal: FC<CreateNewListModalProps> = ({
+const onListAdded = (store, data) => {
```
nickskalkin, line 50: **suggestion**: I would extract useMutation to a separate file since mutation logic grew quite a bit

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -220,7 +254,7 @@ export const CreateNewListModal: FC<CreateNewListModalProps> = ({
   )
 }
 
-export const CreateNewListModalContainer: FC<CreateNewListModalContainerProps> = props => {
+export const CreateNewListModalContainer: React.FC<CreateNewListModalContainerProps> = props => {
```
nickskalkin, line 257: **question**: @dimatretyak I remember that you prefer using `FC` instead of `React.FC` and you asked me to change `React.FC` to `FC` in one of my PRs recently. I don't have strong opinion what to use, but let's all agree on something

nickskalkin (approved):

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
dimatretyak: yeah, agree with Nikita

dimatretyak, line 50: suggestion: add TS definitions for `store` and `data`

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
dimatretyak: @nickskalkin good catch. @gkartalis could you please use just `FC<CreateNewListModalContainerProps>`?

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/DeleteSavesModal.tsx
```diff
@@ -10,8 +11,25 @@ interface Props {
   onClose: () => void
 }
 
+const deleteCollectionUpdater = (store, collectionID) => {
```
dimatretyak, line 14: **suggestion**: Add TS for `store` and `collectionID`)

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/DeleteSavesModal.tsx
```diff
@@ -29,6 +47,7 @@ export const DeleteSavesModal: React.FC<Props> = ({ collection, onClose }) => {
         }
       }
     `,
+    updater: store => deleteCollectionUpdater(store, collection.internalID),
```
dimatretyak, line 50: we should pass `collection.id` here (not `collection.internalID`)

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
gkartalis: 🤷‍♂️ sure thing!

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/DeleteSavesModal.tsx
gkartalis: on it

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
dimatretyak: thanks!

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useDeleteCollection.ts
```diff
@@ -0,0 +1,58 @@
...
+    me,
+    "CollectorProfileSaves2Route_otherSaves"
+  )
+
+  if (!otherSavesConnection) {
+    return
+  }
+
+  ConnectionHandler.deleteNode(otherSavesConnection, collectionID)
+}
```
dimatretyak, line 32: **suggestion**: feel free to ignore if you don't like these changes. if the mutation failed, the updater will still be executed

```javascript
const deleteCollectionUpdater = (
  store: RecordSourceSelectorProxy<useDeleteCollectionMutation$data>,
  data: useDeleteCollectionMutation$data
) => {
  const { responseOrError } = data.deleteCollection ?? {}
  
  if (responseOrError?.__typename !== "DeleteCollectionSuccess") {
    return
  }
  
  const collectionID = responseOrError.collection?.id
  
  ...
}
```

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useCreateCollection.ts
```diff
@@ -0,0 +1,68 @@
...
+import { RecordSourceSelectorProxy } from "relay-runtime"
+import { useMutation } from "Utils/Hooks/useMutation"
+import {
+  useCreateCollectionMutation,
+  useCreateCollectionMutation$data,
+} from "__generated__/useCreateCollectionMutation.graphql"
+
+const onListAdded = (
+  store: RecordSourceSelectorProxy<useCreateCollectionMutation$data>,
+  data
```
dimatretyak, line 11: **suggestion**: `data: useCreateCollectionMutation$data`

dimatretyak (dismissed):
Great work and improvements!

nickskalkin (dismissed):

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useCreateCollection.ts
gkartalis: If I do that then store is not the correct type and we don't have access to `getRoot` and `getRootField` and we're getting this error

```
Property 'getRoot' does not exist on type 'useCreateCollectionMutation$data'.ts(2339)
```

gkartalis: ooooh, ignore my comment I realized that it was for data 📦

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -86,7 +86,11 @@ export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContai
           first: 30
           default: false
           saves: true
-        ) {
+        )
+          @connection(
+            key: "CollectorProfileSaves2Route_otherSaves"
+            filters: []
```
anandaroop, line 92: Is the empty `filters` argument needed?

anandaroop (approved):
LGTM

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
gkartalis: Good question, I should have left a comment (maybe its worth adding the comment to the codebase itself).

There was a gotcha here that @dimatretyak helped me with, I had the updater implementation without including this here and the result was that the ui was never updating. 

You can test it out if you comment that out and try to create a new list you'll see that the UI is not being updated. Not sure why but needs to be there.

@dimatretyak correct me if I am wrong, and do you think its worth adding a comment here?

gkartalis: On the delete flow you will see the update on the ui but I think it is because of the routing redirect that we do

dimatretyak: @anandaroop by default, all arguments and their values (in our case it is `first: 30`, `default: false`, `saves: true`) are used to generate the special "key" by which we can then find/identity the connections we need among others. 

And based on the 👆 information we have 2 approaches:
* specify `filters: []` and search for the desired connection **only** by the key (as George did it [here](https://github.com/artsy/force/blob/02d91ef25623a00cc69feff84b5984abb3b13391/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useCreateCollection.ts#L21))
* OR ignore `filters` and pass all arguments and their values to the `getConnection` (see example below). IMO This method is more complicated

```javascript
const otherSavesConnection = ConnectionHandler.getConnection(me, key, {
  first: 30,
  default: false,
  saves: true,
})
```

You can read more about it here: https://relay.dev/docs/guided-tour/list-data/updating-connections/#connection-identity-with-filters

anandaroop: Thanks for the explanation

PR #12083, opened 2023-03-14
style: move the see full results to the bottom

rajsam003:
The type of this PR is: Style

This PR solves [FX-4650]

|Before|After|
|---|---|
|image|image|

[FX-4650]: https://artsyproduct.atlassian.net/browse/FX-4650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -21,8 +21,6 @@ import qs from "qs"
 // eslint-disable-next-line no-restricted-imports
 import Autosuggest from "react-autosuggest"
 import { RelayRefetchProp, createRefetchContainer, graphql } from "react-relay"
-// eslint-disable-next-line no-restricted-imports
-import { data as sd } from "sharify"
 import styled from "styled-components"
 // eslint-disable-next-line no-restricted-imports
 import request from "superagent"
@@ -35,6 +33,7 @@ import { SearchInputContainer } from "./SearchInputContainer"
 import { useTranslation } from "react-i18next"
 import { ClassI18n } from "System/i18n/ClassI18n"
 import track from "react-tracking"
+import { getENV } from "Utils/getENV"
 
 const logger = createLogger("Components/Search/SearchBar")
 
@@ -148,7 +147,7 @@ export class SearchBar extends Component<Props, State> {
         if (error) {
           logger.error(error)
           return
-        } else if (performanceStart && sd.VOLLEY_ENDPOINT) {
+        } else if (performanceStart && getENV("VOLLEY_ENDPOINT")) {
           this.reportPerformanceMeasurement(performanceStart)
         }
         const { viewer } = this.props
@@ -191,7 +190,7 @@ export class SearchBar extends Component<Props, State> {
 
   reportPerformanceMeasurement = performanceStart => {
     const duration = performance.now() - performanceStart
-    const deviceType = sd.IS_MOBILE ? "mobile" : "desktop"
+    const deviceType = getENV("IS_MOBILE") ? "mobile" : "desktop"
 
     const metricPayload = {
       name: "autocomplete-search-response",
@@ -201,7 +200,7 @@ export class SearchBar extends Component<Props, State> {
     }
 
     request
-      .post(sd.VOLLEY_ENDPOINT)
+      .post(getENV("VOLLEY_ENDPOINT"))
       .send({
         metrics: [metricPayload],
         serviceName: "force",
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -18,10 +18,13 @@ import { Router } from "found"
 import { isEmpty } from "lodash"
 import { throttle } from "lodash"
 import qs from "qs"
+// eslint-disable-next-line no-restricted-imports
 import Autosuggest from "react-autosuggest"
 import { RelayRefetchProp, createRefetchContainer, graphql } from "react-relay"
+// eslint-disable-next-line no-restricted-imports
 import { data as sd } from "sharify"
 import styled from "styled-components"
+// eslint-disable-next-line no-restricted-imports
 import request from "superagent"
 import Events from "Utils/Events"
 import { get } from "Utils/get"
@@ -410,7 +413,7 @@ export class SearchBar extends Component<Props, State> {
     // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
     const edges = get(viewer, v => v.searchConnection.edges, [])
     // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-    const suggestions = [firstSuggestionPlaceholder, ...edges]
+    const suggestions = [...edges, firstSuggestionPlaceholder]
 
     return (
       <Autosuggest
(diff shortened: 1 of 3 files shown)


src/Components/Search/SearchBar.tsx
```diff
@@ -18,10 +18,13 @@ import { Router } from "found"
 import { isEmpty } from "lodash"
 import { throttle } from "lodash"
 import qs from "qs"
+// eslint-disable-next-line no-restricted-imports
 import Autosuggest from "react-autosuggest"
 import { RelayRefetchProp, createRefetchContainer, graphql } from "react-relay"
+// eslint-disable-next-line no-restricted-imports
 import { data as sd } from "sharify"
 import styled from "styled-components"
+// eslint-disable-next-line no-restricted-imports
```
rajsam003, line 27: I know it's not an ideal solution. Any better solution?

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -98,7 +105,13 @@ const DefaultSuggestion: React.FC<SuggestionItemProps> = ({
   const matches = match(display, query)
   const parts = parse(display, matches)
   const partTags = parts.map(({ highlight, text }, index) =>
-    highlight ? <strong key={index}>{text}</strong> : text
+    highlight ? (
+      <strong key={index} style={{ color: "blue" }}>
```
rajsam003, line 109: `<strong color="blue100">` tag is not working.  
I can replace it with <Text> but it's not simple like strong. Also, the alignments change. 
image

src/Components/Search/SearchBar.tsx
```diff
@@ -18,10 +18,11 @@ import { Router } from "found"
 import { isEmpty } from "lodash"
 import { throttle } from "lodash"
 import qs from "qs"
+// eslint-disable-next-line no-restricted-imports
 import Autosuggest from "react-autosuggest"
 import { RelayRefetchProp, createRefetchContainer, graphql } from "react-relay"
-import { data as sd } from "sharify"
 import styled from "styled-components"
+// eslint-disable-next-line no-restricted-imports
```
dimatretyak, line 25: **praise**: thank for these changes 🙏

src/Components/Search/SearchBar.tsx
```diff
@@ -410,7 +412,7 @@ export class SearchBar extends Component<Props, State> {
     // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
     const edges = get(viewer, v => v.searchConnection.edges, [])
     // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-    const suggestions = [firstSuggestionPlaceholder, ...edges]
+    const suggestions = [...edges, firstSuggestionPlaceholder]
```
dimatretyak, line 415: **question**: What do you think about renaming `firstSuggestionPlaceholder`? For example, in `lastSuggestionPlaceholder` or even better in `seeAllResultsPlaceholder`

dimatretyak: Also, I would rename all the related code (for example, as [here](https://github.com/artsy/force/blob/056073747177a3b77a2dd5bbbb068760dbb9b0ae/src/Components/Search/SearchBar.tsx#L341), [here](https://github.com/artsy/force/blob/056073747177a3b77a2dd5bbbb068760dbb9b0ae/src/Components/Search/SearchBar.tsx#L408) and [here](https://github.com/artsy/force/blob/056073747177a3b77a2dd5bbbb068760dbb9b0ae/src/Components/Search/Suggestions/SuggestionItem.tsx#L32))

src/Components/Search/Suggestions/SuggestionItem.tsx
dimatretyak: we can create the new styled component. something like this

```javascript
const Highlight = styled.strong`
`
...
const partTags = parts.map(({ highlight, text }, index) =>
  highlight ? <Highlight key={index}>{text}</Highlight> : text
)
```

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -40,12 +41,18 @@ export const FirstSuggestionItem: React.FC<SuggestionItemProps> = ({
...
+      mt={1}
+      borderTop="1px solid"
+      borderTopColor="black10"
       onClick={handleClick}
       to={href}
     >
-      <Text variant="sm">See full results for &ldquo;{query}&rdquo;</Text>
+      <Flex alignItems="center">
+        <Text variant="sm" mr={1}>
+          See full results for &ldquo;{query}&rdquo;
```
dimatretyak, line 52: we don't need to use `&ldquo;` and `&rdquo;` here (Just `See full results for {query}`)

dimatretyak: Also i think we need to highlight query (see screenshot)

for-sam

src/Components/Search/Suggestions/SuggestionItem.tsx
```diff
@@ -40,12 +41,18 @@ export const FirstSuggestionItem: React.FC<SuggestionItemProps> = ({
...
+        </Text>
+        <ChevronButton />
```
dimatretyak, line 54: **suggestion**: I think it's better to use `ChevronRightIcon` here. Something like this

```javascript
import ChevronRightIcon from "@artsy/icons/ChevronRightIcon"
...
<ChevronRightIcon height="14px" />
```

src/Components/Search/SearchBar.tsx
rajsam003: Good idea

src/Components/Search/SearchBar.tsx
```diff
@@ -400,19 +400,19 @@ export class SearchBar extends Component<Props, State> {
       value: term,
     }
 
-    const firstSuggestionPlaceholder = {
+    const seeAllResultsPlaceholder = {
       node: {
         displayLabel: term,
-        displayType: "FirstItem",
+        displayType: "SeeFullResultsSuggestionItem",
```
dblandin, line 406: **suggestion/blocking**: I think this has the risk of altering some analytics to track when a user has clicked on the "See full results" suggestion. For now, can we keep the displayType as `FirstItem` until we're able to touch base with @abhitip and align on a new value?

dblandin (changes requested):
LGTM overall but I left one blocking bit of feedback.

src/Components/Search/SearchBar.tsx
rajsam003: It makes sense. I am reverting back

dimatretyak: > reverting back

I think we can rename it in other places, but leave only the old value (`FirstItem`) here

dimatretyak (approved):
LGTM. Thanks for changes 🙏

src/Components/Search/SearchBar.tsx
dimatretyak: > I think this has the risk of altering some analytics to track when a user has clicked on the "See full results" suggestion

Yes, you're right

Sources - force-staging - Segment 2023-03-15 13-53-23

dblandin: Yeah, exactly. The code variables can be renamed/refactored as long as the analytics event remains consistent.

PR #12088, opened 2023-03-15
refactor: move AppToasts to bottom

gkartalis:
The type of this PR is: **Style**

This PR solves [FX-4666]

### Description

⚠️ Affects all Toasts in force (used in ~ 6 places as you see on the screenshot below

Screenshot 2023-03-15 at 11 54 50

- [x] Moves all toasts in force to the bottom instead of the top of the page that they appear.
- [x] [Review App](toasts.artsy.net) to thoroughly QA the surfaces that have toasts

|mWeb|Desktop|
|---|---|
|Screenshot 2023-03-15 at 11 52 05|Screenshot 2023-03-15 at 11 58 41|

[FX-4666]: https://artsyproduct.atlassian.net/browse/FX-4666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Components/AppToasts.tsx
```diff
@@ -1,27 +1,15 @@
...
 
   return (
-    <Box
-      position="fixed"
-      zIndex={Z.toasts}
-      top={accomodateNav ? height : 0}
-      left={0}
-      width="100%"
-    >
+    <Box position="fixed" zIndex={Z.toasts} bottom={4} left={0} width="100%">
```
dzucconi, line 12: 4 plus the 1 below equals an invalid value. Should just use the py below to vertically place

dimatretyak:
@anandaroop as far as I remember you said that the design team plans to move the toast messages to the bottom-right corner? Do you know if there is any feedback/final decision whether to do this?

src/Apps/Components/AppToasts.tsx
gkartalis: oh, gotcha

gkartalis:
> @anandaroop as far as I remember https://github.com/artsy/force/pull/11978#issuecomment-1448823586 that the design team plans to move the toast messages to the bottom-right corner? Do you know if there is any feedback/final decision whether to do this?

On the ticket the question was there but had a reply (from refinement if I remember correctly) to **keep it to center** (see the ticket questions section)

damassi:
Gosh, why do we have to do the toast in the center, as this tiny lil thing that's barely visible? Vs having it float right and match every other app / general toast idiom that one sees around? 

Can we raise this in the #design-system channel?

gkartalis:
Yep, we are gonna share with broader team (first shared it on this [slack thread](https://artsy.slack.com/archives/C9SATFLUU/p1678888732495149) to get feedback from our designer) and then we are gonna get feedback from the broader team

damassi:
It is worth gathering examples from Forque and Volt-v2 as well. Those are internal apps but the toasts work well as an example to refer to.

anandaroop (dismissed):
Nice! Change looks good to me.

Let's see how the review app / discussion shakes out.

dimatretyak (approved):

PR #12095, opened 2023-03-15
feat(TX-1079): update order app layout

rquartararo:

This PR solves [TX-1079] and [TX-1020]

### Description

This PR:
- Changes the Order App layout to the "LogoOnly" layout
- Deletes `MinimalNavBar` as it's deprecated and no longer being used

[TX-1079]: https://artsyproduct.atlassian.net/browse/TX-1079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TX-1020]: https://artsyproduct.atlassian.net/browse/TX-1020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Components/Layouts/LayoutLogoOnly.tsx
```diff
@@ -15,17 +14,15 @@ export const LayoutLogoOnly: FC<BaseLayoutProps> = ({ children }) => {
 
       <LayoutMain>
         <AppContainer>
-          <HorizontalPadding>
```
rquartararo, line 18: Removed this bc `LayoutMain` also has horizontal padding

dzucconi: It does? https://github.com/artsy/force/blob/main/src/Apps/Components/Layouts/Components/LayoutMain.tsx

src/Apps/Order/OrderApp.tsx
```diff
@@ -100,37 +99,33 @@ const OrderApp: FC<OrderAppProps> = props => {
...
-      <MinimalNavBar to="/" isBlank={isModal}>
-        <Title>Checkout | Artsy</Title>
-        <Meta
-          name="viewport"
-          content={
-            isEigen
-              ? "width=device-width, user-scalable=no"
-              : "width=device-width, initial-scale=1, maximum-scale=5 viewport-fit=cover"
-          }
+    <Layout variant={isModal ? "Blank" : "LogoOnly"}>
```
dzucconi, line 102: Neat

src/Apps/Components/Layouts/LayoutLogoOnly.tsx
rquartararo: oh.. its `LayoutContainterOnly` which I see is not inside the `LogoOnly` 🤔

dzucconi: This looks well aligned with the other layouts. Something else is likely causing the double padding, not the layout. I'd roll back this change. https://force-storybook.artsy.net/?path=/story/components-layouts--logo-only

rquartararo: Oh I see whats happening it's due to the routes being `ContainerOnly`.

dzucconi: Ah, yep. Generally speaking we'd want to apply the layout using the routes. Actually it hadn't occurred to me that anyone would want to use them otherwise. So it seems like you're going to have two `main` elements, with duplicated IDs, which you don't want.

dzucconi: (And duplicate toast containers)

If you wanted to use these outside of routing I suppose one thing you could do is maybe add a `LayoutNone` option (or whatever) that is literally just `children`.

rquartararo: Good idea i've added back the padding to `LogoOnly` and created a `LayoutNone` for the order routes

src/Apps/Order/orderRoutes.tsx
```diff
@@ -89,7 +89,7 @@ export const orderRoutes: AppRouteConfig[] = [
   {
     // TODO: Still need order2?
     path: "/order(2|s)/:orderID",
-    layout: "ContainerOnly",
+    layout: "None",
```
damassi, line 92: Might be worthwhile adding a comment above this first instance noting that we provide the layout elsewhere, as the app can appear within a modal or as a page. 

(Meaning, on first glance it seems confusing that we have an app without a layout, when it _seems_ to have a layout upon viewing the page.)

rquartararo: Yes, it is a bit confusing. Added a comment 👍

AlicanAkyuz (dismissed):

rquartararo:
@damon what do you mean by "real layout" in this case?

dzucconi:
Using a route-level layout that isn't `None` instead of at the component level.

rquartararo:
Yes, I agree with using a real layout at the route level. The biggest issue is with the modal view otherwise we could just route with `LogoOnly` and be done with it. I'm curious what design thinks about just keeping the modal view consistent and allowing the logo there. We only use the modal for responding to offers in conversations and I personally don't think it looks bad with the logo.

dzucconi:
Just extract the relevant parts and modal those though?

rquartararo:
Ah okay, I see what you were first suggesting now. Let me try it out.

rquartararo:
@dzucconi our order modal is an iframe using the order routes so I think this would have to involve creating new routes

dzucconi:
Ah, I would maybe bail out of this with the understanding that iframing is an obviously bad idea that needs to be addressed eventually

rquartararo:
I talked with Viviana and she thinks fixing the no logo on routing that we currently have is more of a priority than the logo inside the modal issue. With this in mind, I'm going to switch all order routes to `LogoOnly` and remove the layout from `OrderApp` for a consistent routing experience. We can handle to modal later when we start working on the conversations V2 in Force.

src/Apps/Order/__tests__/OrderApp.jest.tsx
```diff
@@ -668,19 +668,4 @@ describe("OrderApp", () => {
       "Sorry, the page you were looking for doesn’t exist at this URL."
     )
   })
-
-  it("redirects user to home page when Artsy logo is clicked", () => {
```
rquartararo, line 672: removing this test as the logo is no longer inside OrderApp.tsx

src/Apps/Components/Layouts/LayoutLogoOnly.tsx
```diff
@@ -16,13 +16,13 @@ export const LayoutLogoOnly: FC<BaseLayoutProps> = ({ children }) => {
       <LayoutMain>
         <AppContainer>
           <HorizontalPadding>
-            <Spacer y={4} />
+            <Spacer y={2} />
```
dzucconi, line 19: This isn't something that should be changed.

rquartararo: It was ticketed here: https://artsyproduct.atlassian.net/browse/TX-1079

dzucconi: This changes it for the entire app though.

dzucconi: The whole point of using a layout is consistency. I feel like we're missing something here

rquartararo: Yeah, I interpreted the ticket as updating the logo-only layout for everywhere on the app so the checkout is consistent with everything else. I'm going to clarify this.

rquartararo: Otherwise we need to create another layout

dzucconi: > Otherwise we need to create another layout

I would disagree with this :)

Why would we want an inconsistent experience in checkout? What's the underlying _intent_ behind this change? There's probably a better path forward.

dzucconi: For instance: I actually think this _should_ be `<Spacer y={[2, 4]} />` since we want that correspondence with the outer margins.

rquartararo: I definitely agree that the checkout should use the `LogoOnly` layout, but design wants the spacing changed. I talked with Viviana and she is okay with the 20px header spacing on the error pages / other pages using `LogoOnly`. So we can go ahead with this spacing change if it's okay with you.

dzucconi: RIP to this perfect correspondence:

rip

damassi: @rquartararo - We should push back on this. Perhaps Viviana isn't aware of the wider implications of this change?

dzucconi: I mean: I'd push back just because I think it looks nicer and we tend to use `[2, 4]` on _every_ page as the top margin. But the implications aren't _that_ wide. The layout is only used on error pages and I guess now the checkout. I do think it's going to look awkward on the error page and that if the goal is to fit the checkout on a single screen, misguided.

damassi: Its wide, because an error can throw from any page! so immediately that's obvious (and doesn't feel necessary).

rquartararo: Okay, Viviana agreed to table this change for now! I reverted it back to 40px. Let me know if all is good now and I can merge this.

dzucconi: 🙏 👍

PR #12100, opened 2023-03-16
feat(TX-1022): PS Conditions of sale page

AlicanAkyuz:

This PR solves [TX-1022]

### Description

This PR pulls and displays a new page in the PageApp, for private sales conditions of sale, subjected to authentication. Also updates the links in the checkout flow to point to this new page. 

#### Video 
... comins soon

### DIFF
--- a/src/Apps/Page/PageApp.tsx
+++ b/src/Apps/Page/PageApp.tsx
@@ -1,17 +1,27 @@
 import { Column, GridColumns, Spacer } from "@artsy/palette"
-import { FC, useMemo } from "react"
+import { FC, useMemo, useEffect } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { useSystemContext } from "System/useSystemContext"
+import { ContextModule } from "@artsy/cohesion"
+import { useAuthDialog } from "Components/AuthDialog"
 import { MetaTags } from "Components/MetaTags"
 import { useRouter } from "System/Router/useRouter"
 import { PageApp_page$data } from "__generated__/PageApp_page.graphql"
 import { PageHTML } from "./Components/PageHTML"
-import { TOP_LEVEL_PAGE_SLUG_ALLOWLIST } from "./pageRoutes"
+import {
+  TOP_LEVEL_PAGE_SLUG_ALLOWLIST,
+  PAGE_SLUGS_WITH_AUTH_REQUIRED,
+} from "./pageRoutes"
 
 interface PageAppProps {
   page: PageApp_page$data
 }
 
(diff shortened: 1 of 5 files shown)


rquartararo:
Is there a way to avoid the grey screen before the page loads?

src/Apps/Page/PageApp.tsx
```diff
@@ -25,14 +35,34 @@ const PageApp: FC<PageAppProps> = ({ page }) => {
...
+            return `${action} to view ${page.name}`
+          },
+          redirectTo: match.location.pathname,
+        },
+        analytics: {
+          contextModule: ContextModule.header,
+        },
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
AlicanAkyuz: I don't think using useCallback will help with that. It'll be the same story with it (that we want it to fire when page or user updates, but not when showAuthDialog or match updates).

AlicanAkyuz:
about the grey screen, I don't know a way to prevent it as long as we want to open the page in a new tab. let me know if you do!

src/Apps/Page/PageApp.tsx
```diff
@@ -25,14 +35,34 @@ const PageApp: FC<PageAppProps> = ({ page }) => {
...
 
-  const { match } = useRouter()
+  useEffect(() => {
+    if (PAGE_SLUGS_WITH_AUTH_REQUIRED.includes(page.internalID) && !user?.id) {
+      showAuthDialog({
+        mode: "SignUp",
+        options: {
+          title: mode => {
+            const action = mode === "SignUp" ? "Sign up" : "Log in"
+            return `${action} to view ${page.name}`
```
starsirius, line 45: This treatment is nice!

src/Apps/Page/PageApp.tsx
```diff
@@ -25,14 +35,34 @@ const PageApp: FC<PageAppProps> = ({ page }) => {
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [page, user])
 
   const canonical = TOP_LEVEL_PAGE_SLUG_ALLOWLIST.includes(match.params.id)
     ? `/${match.params.id}`
     : `/page/${match.params.id}`
 
   if (!page.content) return null
 
+  if (PAGE_SLUGS_WITH_AUTH_REQUIRED.includes(page.internalID) && !user?.id)
```
starsirius, line 63: #non-blocking. The logic repeats so potentially can be DRY'd with something like `if (auth_required)`.

PR #12106, opened 2023-03-16
feat: add to list functionallity to all rails

gkartalis:

This PR solves [FX-4683]

### Description

[Review App](https://artwork-list-rails.artsy.net/)
[Notion doc](https://www.notion.so/artsy/Artwork-lists-stories-tasks-443076e8acde4fc783abedc16b1fb438?pvs=4#b57762b7b8b94772869d58665733fbae)

- [x] add to list flow to all rails
- [x] exclude in auction artworks from add to list flow

[FX-4683]: https://artsyproduct.atlassian.net/browse/FX-4683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/GridItem.tsx
```diff
@@ -118,7 +118,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
           showHoverDetails={showHoverDetails}
           disableRouterLinking={disableRouterLinking}
           to={to}
-          enableSaveButtonForLists
+          enableSaveButtonForLists={!artwork.isInAuction}
```
gkartalis, line 121: excludes in auction works from the flow (for artwork grids)

src/Components/Artwork/ShelfArtwork.tsx
```diff
@@ -41,58 +44,61 @@ const ShelfArtwork: React.FC<ShelfArtworkProps> = ({
...
-      />
+        <Metadata
+          artwork={artwork}
+          hideSaleInfo={hideSaleInfo}
+          isHovered={isHovered}
+          contextModule={contextModule}
+          showSaveButton
+          disableRouterLinking
+          maxWidth="100%"
+          enableSaveButtonForLists={!artwork.isInAuction}
```
gkartalis, line 96: excludes in auction works from the flow (for artwork rails)

src/Components/Artwork/ShelfArtwork.tsx
```diff
@@ -17,6 +18,7 @@ export interface ShelfArtworkProps
   lazyLoad?: boolean
   onClick?: () => void
   width?: number[]
+  savedListId?: string
```
dimatretyak, line 21: The `savedListId` prop will never be used since the `ShelfArtwork` component is not used on the `collector-profile/saves2' page. For this reason we can delete this

src/Components/Artwork/ShelfArtwork.tsx
dimatretyak: **question**: maybe it makes sense to do [this check](https://github.com/artsy/force/blob/1716d72df521eacea2af9a32eeae8344dc17d8af/src/Components/Artwork/Details.tsx#L231) in the `Details` component itself, since now the new behavior of the 💙 button will be on all artwork grids/rails? Also we can safely remove `enableSaveButtonForLists` since there is no need for this prop anymore

dimatretyak (commented):
**question**: do we need to add changes for [FlatGridItem](https://github.com/artsy/force/blob/1716d72df521eacea2af9a32eeae8344dc17d8af/src/Components/Artwork/FlatGridItem.tsx#L24) also?

src/Components/Artwork/GridItem.tsx
gkartalis: should I also remove this from here as well?

gkartalis: @dimatretyak

src/Components/Artwork/ShelfArtwork.tsx
dimatretyak: hmm, good catch. perhaps checking `isAuctionArtwork` will be enough for our case

cc: @damassi

dimatretyak: I think it is better to check `isInAuction` (as it will be safer)

gkartalis: amazing, resolving this convo

src/Components/Artwork/GridItem.tsx
dimatretyak: yep

src/Components/Artwork/ShelfArtwork.tsx
dimatretyak: [Just for context] Since `isAuctionArtwork` will not always be `true` if the artwork is in an auction. For example, as in the attached video (if `isAuctionArtwork` is `true`, then there should be a **red** background. But in our example, the background is green even though the artwork is at auction)

dimatretyak:
I we navigate to the some artwork from viewing room (e.g. like [here](https://artwork-list-rails.artsy.net/viewing-room/massimodecarlo-thomas-grunfeld/artworks)), we got error

dimatretyak:
> @dimatretyak fixed

[Friend Reminder] Please push the fix to the review app branch 🙏

gkartalis:
(already done, might take ~10 minutes to get the update)

dimatretyak (approved):
Nice, looks great. You deserve it 🏅

Do you mind if you will merge this changes on Monday? To be safe and be calm on the weekend

gkartalis:
Makes total sense will merge on Monday! 

One concern:

The error was coming from the `DetailsFragmentContainer` not being wrapped with the `ManageArtworkForSavesProvider`. 

And we also have another unwrapped reference of `DetailsFragmentContainer` [here](https://github.com/artsy/force/blob/5986de88fcd19b1dc3c17cdb6becbfb3a5686ef4/src/Apps/Settings/Routes/Auctions/Components/SettingsAuctionsLotStanding.tsx#L68) but not quite sure where this lives in force so I can test in the review app.

gkartalis:
Looks like this is located on https://artwork-list-rails.artsy.net/settings/auctions but works fine since we don't display the save button on this surface (`showSaveButton=false`).

Tested with @dimatretyak 💯 

image

src/Components/Artwork/ShelfArtwork.tsx
anandaroop: **question (non-blocking):**  Does `isAuctionArtwork` still have value then, or should it be deprecated?

anandaroop (dismissed):
LGTM

PR #12110, opened 2023-03-16
fix(home): fixes image stretching; clean up

dzucconi:
As raised in this slack thread: https://artsy.slack.com/archives/C01ADJNCS5D/p1678979020271839

The image on the mobile card was getting stretched; this scales it proportionally along with some other clean up.

| Before | After |
| ----- | ----- |
| staging artsy net_(iPhone 12 Pro) |  localhost_5050_(iPhone 12 Pro) |


src/Apps/Home/Components/HomeContentCards/HomeContentCard.tsx
```diff
@@ -21,119 +21,157 @@ interface HomeContentCardProps {
...
   card,
   index,
   onClick,
 }) => {
-  const extras = card.extras || {}
+  const image = card.imageUrl
+    ? cropped(card.imageUrl, {
+        // 3:2
+        width: 500,
+        height: 333,
```
dzucconi, line 47: We can use a smaller image as well

src/Apps/Home/Components/HomeContentCards/HomeContentCard.tsx
```diff
@@ -21,119 +21,157 @@ interface HomeContentCardProps {
...
+  return (
+    <RouterLink
+      onClick={onClick}
+      display="block"
+      width="100%"
+      height="100%"
+      to={card.url!}
+      bg="black5"
+      textDecoration="none"
+      aria-label={`${card.title} - ${card.description}`}
```
dzucconi, line 60: Made these so that they only have a single link which has a label now for better accessibility. Previously it would just focus the buttons.

jonallured:
tyyyyyy 🙌

PR #12116, opened 2023-03-17
feat(FX-4689): Change the behavior of the 💙 button when the user is NOT logged in to the account

dimatretyak:

This PR solves [FX-4689]

Review app: [save-action-for-non-logged-user](https://save-action-for-non-logged-user.artsy.net/)

> **Warning**
> We may get "unexpected" behavior due to some features of Unleash.
> 
> For example, the "manage lists for artwork" modal may not open after the user logs in, and artwork will be saved to the default list. This is due to the fact that feature flag will always be "disabled" for non-logged users, because we use "UserIDs" and "Hosts" strategies. The problem will **NOT** be observed when using the "Standard on/off" strategy

#### Artwork was not previously saved to any list
https://user-images.githubusercontent.com/3513494/226399103-b1a83c6e-a429-4195-8a39-46ddc5335333.mp4

#### Artwork was previously saved only to the "All Saves" list
https://user-images.githubusercontent.com/3513494/226399224-56ac645a-419c-44ee-b63f-001435c1079f.mp4

#### Artwork was previously saved to any custom lists
https://user-images.githubusercontent.com/3513494/226399403-b87676eb-ad65-4b4f-a041-910c44e8a6c6.mp4

[FX-4689]: https://artsyproduct.atlassian.net/browse/FX-4689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/useArtworkLists.ts
```diff
@@ -0,0 +1,82 @@
...
+        action: {
+          label: t(
+            "collectorSaves.saveArtworkToLists.toast.artworkSaved.button"
+          ),
+          onClick: openSelectListsForArtworkModal,
+        },
+      })
+
+      return
+    }
```
dimatretyak, line 58: **question**: do you think it would be a good idea to move all this logic to [useSaveArtworkToLists](https://github.com/artsy/force/blob/3e1fc0c1a9d35d33e5b35b3e661cfa8e1b97cee7/src/Components/Artwork/SaveButton/useSaveArtworkToLists.ts#L30)?

src/Components/Artwork/SaveButton/useSaveArtworkToLists.ts
```diff
@@ -0,0 +1,129 @@
...
+    })
+  }
+
+  const saveToDefaultList = () => {
+    if (!relayEnvironment) {
+      return
+    }
+
+    // TODO: Use `useMutation` in the future
+    return SaveArtwork(
```
dimatretyak, line 76: copy-paster from here: https://github.com/artsy/force/blob/a11d421ff39822216c22d6409a52c5e11a9949aa/src/Components/Artwork/SaveButton/useSaveArtwork.ts#L32-L58

src/Components/Artwork/SaveButton/SaveArtworkToListsButton.tsx
```diff
@@ -15,19 +20,58 @@ const SaveArtworkToListsButton: FC<SaveArtworkToListsButtonProps> = ({
...
+      slug: artwork.slug,
+      title: `${artwork.title}, ${artwork.date}`,
+      imageURL: artwork.preview?.url ?? null,
+      isSavedToDefaultList,
+      isSavedToCustomLists,
+    },
   })
 
-  const handleClick = (
+  const getActionLabel = (action: ResultAction) => {
```
anandaroop, line 43: **suggestion (non-blocking):** A couple of readability thoughts…

Type this arg more narrowly, e.g. 

```ts
action: ResultAction.SavedToDefaultList | ResultAction.RemovedFromDefaultList
```

Make the mapping clearer:

```ts
  return {
    [ResultAction.SavedToDefaultList]: "Saved Artwork",
    [ResultAction.RemovedFromDefaultList]: "Removed Artwork",
  }[action]
```

src/Components/Artwork/useArtworkLists.ts
```diff
@@ -0,0 +1,82 @@
...
+    openSelectListsForArtworkModal,
+  } = useSaveArtworkToLists(options)
+
+  useEffect(() => {
+    if (value?.objectId !== artwork.internalID) {
+      return
+    }
+
+    if (artwork.isSavedToCustomLists) {
+      // Display select lists for artwork modal if arwork is saved to the custom lists
```
anandaroop, line 28:
```suggestion
      // Display select lists for artwork modal if artwork is saved to the custom lists
```

anandaroop (approved):
👍🏽

src/Components/Artwork/useArtworkLists.ts
```diff
@@ -0,0 +1,82 @@
...
+      openSelectListsForArtworkModal()
+    } else if (artwork.isSavedToDefaultList) {
+      // Display toast if artwork is already saved to the default list
+      showToastByAction(ResultAction.SavedToDefaultList)
```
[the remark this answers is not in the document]
dimatretyak: yes

src/Components/Artwork/useArtworkLists.ts
dimatretyak: thanks 🤝

dimatretyak:
> we could distinguish the purpose of these hooks better through their naming

@anandaroop yes, I fully agree with you. but to tell the truth, I can't think of anything better than `useSaveArtworkToLists` and `useArtworkLists` and therefore I am open to any suggestions 😅

src/Components/Artwork/useArtworkLists.ts
dimatretyak: > we could distinguish the purpose of these hooks better through their naming

yes, totally agree with you. But to tell the truth, I can't think of anything better than `useSaveArtworkToLists` and `useArtworkLists` and therefore I am open to any suggestions 😅

src/Components/Artwork/SaveButton/SaveArtworkToListsButton.tsx
dimatretyak: Done: https://github.com/artsy/force/pull/12116/commits/1e0ef77e8b017908c138718e73f6110b2190c126

gkartalis (approved):

PR #12120, opened 2023-03-20
fix: address sell footer review comment

MounirDhahri:

### Description

This PR address @dzucconi comment here https://github.com/artsy/force/pull/11977/files#r1140535183

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
@@ -1,24 +1,19 @@
-import { RouterLink } from "System/Router/RouterLink"
+import { Box, Text } from "@artsy/palette"
 import { FullBleedBanner } from "Components/FullBleedBanner"
-import { useWindowSize } from "Utils/Hooks/useWindowSize"
-import { breakpoints, Text } from "@artsy/palette"
+import { RouterLink } from "System/Router/RouterLink"
 
 export const FooterBanner: React.FC = () => {
-  const { width } = useWindowSize()
-
   return (
-    <>
-      {width > parseInt(breakpoints.sm, 10) && (
-        <FullBleedBanner dismissable={false} variant="defaultLight" mb={-4}>
-          <Text variant="xs">
-            Gallery or art dealer?{" "}
-            <RouterLink to="https://partners.artsy.net">
-              Become a partner
-            </RouterLink>{" "}
-            to access the world’s largest online marketplace.
-          </Text>
-        </FullBleedBanner>
-      )}
-    </>
+    <Box display={["none", "block"]}>
+      <FullBleedBanner dismissable={false} variant="defaultLight" mb={-4}>
+        <Text variant="xs">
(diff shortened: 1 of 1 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
```diff
@@ -1,24 +1,19 @@
...
-            <RouterLink to="https://partners.artsy.net">
-              Become a partner
-            </RouterLink>{" "}
-            to access the world’s largest online marketplace.
-          </Text>
-        </FullBleedBanner>
-      )}
-    </>
+    <Box display={["none", "block"]}>
+      <FullBleedBanner dismissable={false} variant="defaultLight" mb={-4}>
```
dzucconi, line 8: Might as well get rid of this external margin while you're in here: https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-external-margins

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
```diff
@@ -1,24 +1,19 @@
...
+        <Text variant="xs">
```
dzucconi, line 9: You can also remove this `Text` wrapper.

dzucconi: (It's already there: https://github.com/artsy/palette/blob/2f94a86cf74ba0226616d0474224245bb829eec2/packages/palette/src/elements/Banner/Banner.tsx#L32)

MounirDhahri: on it, thanks @dzucconi

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/FooterBanner.tsx
```diff
@@ -1,19 +1,18 @@
...
-          to access the world’s largest online marketplace.
-        </Text>
+      <FullBleedBanner dismissable={false} variant="defaultLight">
+        Gallery or art dealer?{" "}
+        <RouterLink to="https://partners.artsy.net">
+          Become a partner
+        </RouterLink>{" "}
+        to access the world’s largest online marketplace.
       </FullBleedBanner>
+      <Spacer y={-4} />
```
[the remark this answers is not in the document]
MounirDhahri: Pardon, I misunderstood your comment. I thought about this and I think keeping it this way is the right thing to do but I might be wrong, I am interested to know what you think.
In order to not remove it, I will need to either inject a property within `LayoutDefault` [here](https://github.com/artsy/force/blob/77e1b4c396b58804aff86d07458388b670a9d6f5/src/Apps/Components/Layouts/LayoutDefault.tsx#L30) or create a new layout for it. Both options do not seem right to me because this used only here, what do you think?

dzucconi (approved):
Thanks for this!

PR #12123, opened 2023-03-20
fix(TX-1015): include edition set Id when creating order after auth

AlicanAkyuz:

### Description

This PR includes edition set ID (when applicable) to the order creation mutation that is fired after auth.


kiryl-zubarau (approved):

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -183,6 +183,7 @@ const ArtworkSidebarCommerialButtons: React.FC<ArtworkSidebarCommercialButtonsPr
             action: "buyNow",
             kind: "artworks",
             objectId: artwork.internalID,
+            secondaryObjectId: selectedEditionSet?.internalID,
```
kiryl-zubarau, line 186: Non-blocking: can we use `editionSetId` instead of `secondaryObjectId`?

PR #12132, opened 2023-03-22
feat(FX-4690): Add ability to add auction artwork to the custom lists on artwork detail screen

dimatretyak:

This PR solves [FX-4690]

Review app: [toast-message-auction-artworks](https://toast-message-auction-artworks.artsy.net/)

### Demo
https://user-images.githubusercontent.com/3513494/227001230-19f08720-ccf4-4b0f-8221-fd7e79b5a58d.mp4

[FX-4690]: https://artsyproduct.atlassian.net/browse/FX-4690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
@@ -17,6 +17,9 @@ import {
   TextProps,
 } from "@artsy/palette"
 import { themeGet } from "@styled-system/theme-get"
+import { getENV } from "Utils/getENV"
+
+const isTest = getENV("NODE_ENV") === "test"
 
 interface UtilButtonProps {
   name:
@@ -131,7 +134,7 @@ const UtilButtonInnerText: React.FC<UtilButtonInnerTextProps> = ({
    * e.g. `screen.getByText("Label")` will return error,
    * since there will be 2 elements with "Label" text on the page.
    */
-  if (longestLabel && process.env.NODE_ENV !== "test") {
+  if (longestLabel && !isTest) {
     return (
       <Box position="relative">
         <VisibleText {...rest}>{label}</VisibleText>
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/__tests__/ArtworkActionsSaveButtonV2.jest.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/__tests__/ArtworkActionsSaveButtonV2.jest.tsx
@@ -11,6 +11,10 @@ import { ArtworkActionsSaveButtonV2_Test_Query } from "__generated__/ArtworkActi
 jest.unmock("react-relay")
 jest.mock("Components/Artwork/SaveButton/SaveArtworkMutation")
 
+jest.mock("Utils/getENV", () => ({
+  getENV: () => "test",
+}))
+
 describe("ArtworkActionsSaveButtonV2", () => {
   const mockSaveArtwork = SaveArtwork as jest.Mock
 


src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
```diff
@@ -126,7 +126,7 @@ const UtilButtonInnerText: React.FC<UtilButtonInnerTextProps> = ({
     return null
   }
 
-  if (longestLabel) {
+  if (longestLabel && process.env.NODE_ENV !== "test") {
```
dimatretyak, line 129: we hide the component that helps us calculate the maximum possible size of the button with a long text, since this functionality is not needed for tests and will add some problems (e.g. `screen.getByText("Label")` will return an error, since there will be 2 elements with the text "Label" on the page)

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButtonV2.tsx
```diff
@@ -0,0 +1,94 @@
...
+        isSaved={isSaved}
+        artwork={artwork}
+        onClick={handleSave}
+        canShowRegistrationPopover={!isSavedToCustomLists}
+      />
+    )
+  }
+
+  return (
+    <ProgressiveOnboardingSaveArtworkQueryRenderer>
```
dimatretyak, line 59: @dzucconi is it okay that I placed `ProgressiveOnboardingSaveArtworkQueryRenderer` here?

dzucconi: Yeah, this is fine. It was moved to avoid targeting auction artworks.

src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
```diff
@@ -126,7 +126,12 @@ const UtilButtonInnerText: React.FC<UtilButtonInnerTextProps> = ({
     return null
   }
 
-  if (longestLabel) {
+  /**
+   * We ignore this, since this functionality is not needed for tests and will add some problems
+   * e.g. `screen.getByText("Label")` will return error,
+   * since there will be 2 elements with "Label" text on the page.
+   */
+  if (longestLabel && process.env.NODE_ENV !== "test") {
```
gkartalis, line 134: **question:** shouldn't we have this `process.env.NODE_ENV !== "test"` check only for the [`HiddenText`](https://github.com/artsy/force/pull/12132/files#diff-c2ea95d31134e0d393b03cd0dcfd9aa3a56f6689c59652f20b949a531ee932b1R138-R140) part of the component?

src/Apps/Artwork/Components/ArtworkImageBrowser/__tests__/ArtworkActionsSaveButtonV2.jest.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+    expect(screen.getByText("Watch lot")).toBeInTheDocument()
+  })
+
+  describe("Save flow", () => {
+    beforeEach(() => {
+      mockSaveArtwork.mockImplementation(args => ({
+        saveArtwork: {
+          artwork: {
+            is_saved: true,
```
gkartalis, line 71: Should we also change the `is_saved` in the `SaveArtworkMutation` to match the [Artwork](https://github.com/artsy/force/pull/12132/files#diff-98dd7020596cdf678daa470d9dbfa754306be2db50b9f460c20e7b8064c5b179R63) query property (asking mostly to make sure that when we update via the mutation the store gets updated accordingly )

gkartalis (approved):
Left two questions but looks great!

src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
dzucconi: I would be genuinely surprised if this didn't crash on the client?

dimatretyak: > I would be genuinely surprised if this didn't crash on the client?

We already use this in some components (like [here](https://github.com/artsy/force/blob/main/src/Components/Truncator.tsx#L32) and [here](https://github.com/artsy/force/blob/main/src/Apps/Order/Routes/Payment/PaymentContext/OrderPaymentContext.tsx#L99)). everything seems to be ok

src/Apps/Artwork/Components/ArtworkImageBrowser/__tests__/ArtworkActionsSaveButtonV2.jest.tsx
dimatretyak: in this case, we need to return `is_saved` because such a field is returned by the `SaveArtwork` mutation [here](https://github.com/artsy/force/blob/606283dd2ba0dd3a68d7878a78c7419977e017f6/src/Components/Artwork/SaveButton/SaveArtworkMutation.ts#L22).

I understand that there is too much confusion due to the fact that somewhere `isSaved` is returned, somewhere `is_saved'. I plan to do this next week. sorry for the confusion 😅

src/Apps/Artwork/Components/ArtworkImageBrowser/UtilButton.tsx
dimatretyak: a good note about using `getENV`(I completely forgot about it). thanks again 🙏

anandaroop (approved):
🙏🏽

src/Apps/Artwork/Components/ArtworkImageBrowser/__tests__/ArtworkActionsSaveButtonV2.jest.tsx
dimatretyak: Renamed here: https://github.com/artsy/force/pull/12161

PR #12143, opened 2023-03-24
refactor: use window `history` for lists

dimatretyak:

This PR solves [PROJECT-XX]

Review app: [history-for-lists](https://history-for-lists.artsy.net/collector-profile/saves2)

### Description
The previous approach was based on the `found` API in combination with [useRouter](https://github.com/artsy/force/blob/main/src/System/Router/useRouter.tsx). The main disadvantage of this approach is that we made [the query to backend](https://github.com/artsy/force/blob/510b52a85950d42193d773ae659ecb33a467a6ca/src/Apps/CollectorProfile/collectorProfileRoutes.tsx#L196-L202) every time and received info about collections when navigating between collections (see video in _Before_ section). Also these new changes will allow us to add some improvements that will be added in a separate PR

#### Before
https://user-images.githubusercontent.com/3513494/227571218-16a7535e-323e-493d-934e-eb6cb8dd3082.mp4

#### After
https://user-images.githubusercontent.com/3513494/227571259-9f43768b-9cc5-42fa-a82a-7d6ee008a33c.mp4


dimatretyak:
Feel free to add any suggestions/comments

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -39,8 +48,30 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
...
+  )
+
+  useSavesHistoryEntity({
+    onChanged: onHistoryChanged,
+  })
+
+  const onDeleteCollection = () => {
+    setActiveCollectionId(savedCollectionId)
+    window.history.replaceState(null, "", "/collector-profile/saves2")
+  }
```
dimatretyak, line 65: Note that just calling `history.pushState()` or `history.replaceState()` won't trigger a `popstate` event (and this code [accordingly](https://github.com/artsy/force/blob/0393e301e64d6e1559cd286455220f637456ad34/src/Apps/CollectorProfile/Routes/Saves2/Hooks/useSavesHistoryEntity.ts#L12-L27)). The `popstate` event will be triggered by doing a browser action such as a **click** on the back or forward button (or **calling** `history.back()` or `history.forward()` in JavaScript). 

For this reason, these changes were added using context

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -39,8 +48,30 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
...
+    [savedCollectionId, setActiveCollectionId]
+  )
+
+  useSavesHistoryEntity({
+    onChanged: onHistoryChanged,
+  })
+
+  const onDeleteCollection = () => {
+    setActiveCollectionId(savedCollectionId)
+    window.history.replaceState(null, "", "/collector-profile/saves2")
```
dimatretyak, line 64: Feel free to add any suggestions/comments. Especially on the use of `window.history`

dzucconi:
Simply querying the backend shouldn't be an issue since the state should be stored in the Relay cache and returned immediately.

dimatretyak:
> Simply querying the backend shouldn't be an issue since the state should be stored in the Relay cache and returned immediately.

yes, but in case the query returns an error, we will get bad UX (for example, as [here](https://artsy.slack.com/archives/C9SATFLUU/p1679595792648279?thread_ts=1679594493.008869&cid=C9SATFLUU)) (although we have data in relay cache)

dzucconi:
I dunno, the problem: we get an error because there's an error — sounds like a valid problem to address? If the data isn't in the cache then it's going to have to be queried, client-side or server-side what does it matter? I'm not sure I understand the root cause here.

dimatretyak:
I've thought about your words and I think you're right. There is nothing wrong/bad with the fact that we will repeatedly query the list of collections when the user clicks on another collection in rail. And if the backend returns an error for one of these query, then we should not try to handle this case in any way.

Thank you for your feedback 🙏

dzucconi:
That said if this particular kind of error is common (which judging by some reports in Slack) it might be. Two things: 1) digging into the root cause of why it occurs, and 2) more elegant error handling, do seem warranted.

PR #12150, opened 2023-03-27
fix(CX-3581): adjust images and footer

dariakoko:

This PR solves [CX-3581]


MrSltun (dismissed):
✨✨

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/Footer.tsx
```diff
@@ -27,7 +28,11 @@ export const Footer: React.FC = () => {
     <>
       <Box px={[2, 4, 12]} mb={[4, 6, 6]}>
         <Text variant={["lg-display", "xl", "xxl"]} pr={0} textAlign="center">
-          Meet your new advisor. It’s Artsy.
+          Meet your new advisor.{" "}
+          <Media at="xs">
```
MounirDhahri, line 32: **suggestion:** I would suggest to avoid using Media here and instead use display none for bigger devices

MounirDhahri (commented):
LGTM, I only left one comment to avoid the usage of the Media.
I am surprised though that sizes for the images are wrong, I chose the ones from the design

MounirDhahri (approved):

PR #12153, opened 2023-03-27
refactor: Verify Your ID and Verify Your Email mutations

ovasdi:

This PR solves [PLATFORM-5038](https://artsyproduct.atlassian.net/browse/PLATFORM-5038)

### Description

This started as a `Fix` to address a problem in Web following this [PR](https://github.com/artsy/gravity/pull/16167), which enabled throttling of Identity Verification create requests. Upon testing throttling on the client, it became apparent that Force did not correctly handle the error coming from the backend. In fact, neither _Verify Your ID_ nor _Verify Your Email_ correctly errored on the client, _always_ displaying a _success message_, even when throttled. See below:

<details>
  <summary>Before changes</summary>
  
  before

</details>

As I looked for a way to fix the underlying issue I ended up changing course and refactoring the existing implementation. I chose to follow what seemed like a more recent pattern to execute mutations using the [`useMutation` hook](https://github.com/artsy/force/blob/7f0950971b7098e733231ff0172bf1a35af381a0/src/Utils/Hooks/useMutation.ts#L9). Having used this pattern in Forque and also seeing that this used in Force, I thought it would _simplify_ the existing implementation and address the problem. See client behavior after the change below:

<details>
  <summary>After changes</summary>

  after

</details>

Added more tests around these two flows.

If the fix is straightforward without this refactor and the preference is to keep the previous implementation, happy to close this PR. In that case, I would like to ask for advice on how to fix the previous implementation.

[PLATFORM-5038]: https://artsyproduct.atlassian.net/browse/PLATFORM-5038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Settings/Routes/EditProfile/Mutations/useVerifyID.ts
```diff
@@ -0,0 +1,31 @@
...
+export const useVerifyID = () => {
+  return useMutation<useVerifyIDMutation>({
+    mutation: graphql`
+      mutation useVerifyIDMutation(
+        $input: SendIdentityVerificationEmailMutationInput!
+      ) {
+        sendIdentityVerificationEmail(input: $input) {
+          confirmationOrError {
+            ... on IdentityVerificationEmailMutationSuccessType {
+              identityVerification {
```
ovasdi, line 14: Note that this refactor changed the mutation to use `identityVerification` field and _not_ `identityVerificationEmail`, which [has been deprecated](https://github.com/artsy/metaphysics/blob/bf6ff5cd64c9dae0e46d859657fb016e514984ac/src/schema/v2/me/sendIdentityVerficationEmailMutation.ts#L56-L60).

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
```diff
@@ -257,7 +230,33 @@ const SettingsEditProfileFields: React.FC<SettingsEditProfileFieldsProps> = ({
...
+
+                        sendToast({
+                          variant: "success",
+                          message: `ID verification link sent to ${
+                            me?.email ?? ""
+                          }.`,
+                          ttl: 6000,
+                        })
+                      } catch (error) {
+                        logger.error(error)
```
ovasdi, line 251: The previous implementation was setup to call [`captureException`](https://github.com/artsy/force/blob/7f0950971b7098e733231ff0172bf1a35af381a0/src/Apps/Settings/Routes/EditProfile/helpers/useHandleVerification.tsx#L39) directly.   

Adding `logger.error` to continue to send these errors to sentry. By my understanding, this should suffice but let me know if that isn't the case.

src/Apps/Settings/Routes/EditProfile/helpers/useHandleVerification.tsx
```diff
@@ -1,68 +0,0 @@
-import { captureException } from "@sentry/browser"
-import { verifyEmail } from "./verifyEmail"
-import { verifyID } from "./verifyID"
-import { useCallback, useState } from "react"
-import { useSystemContext } from "System/useSystemContext"
-import { Environment } from "relay-runtime"
-
-type verificationType = "ID" | "Email"
-
-const IDVerificationBlockingStates = ["passed", "failed", "watchlist_hit"]
```
ovasdi, line 10: I neglected to point out that these `states` are never returned to the client. By my testing, when any of those [states](https://github.com/artsy/gravity/blob/main/app/models/domain/identity_verification.rb#L41) are set on the `identity_verification` the endpoint returns [Unprocessable entity](https://github.com/artsy/gravity/blob/4b98054c2159d7397d53eba0350e4d77af3cf22a/app/api/v1/identity_verifications_endpoint.rb#L73). Before the refactor, the client was not displaying this error. It now does.

Screen Shot 2023-03-28 at 9 01 53 AM

ovasdi:
If there is someone else who should review these changes, please include them or let me know 🙏

damassi (approved):
The best kinds of PRs! LGTM 🚢

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
```diff
@@ -257,7 +230,33 @@ const SettingsEditProfileFields: React.FC<SettingsEditProfileFieldsProps> = ({
...
+                          rejectIf: res => {
+                            return res.sendIdentityVerificationEmail
+                              ?.confirmationOrError?.mutationError
+                          },
+                        })
+
+                        sendToast({
+                          variant: "success",
+                          message: `ID verification link sent to ${
+                            me?.email ?? ""
```
mc-jones, line 246: I think this is a good check to catch any unknown edge cases, but is there ever a scenario where we would expect to not have an email here?

mc-jones (approved):
Echo Chris's [sentiments](https://artsy.slack.com/archives/CA8SANW3W/p1680022449786799). Nicely done!

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
damassi: If this is due to type errors (don't think it is, but heads up), you can always use 
```
me?.email as string
```
to avoid using `any`

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
damassi: Yup! That's what the logger is there for

src/Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields.tsx
ovasdi: @mc-jones good point. I double-checked and there are no such scenarios on `User`. Removed chaining and coalescing in https://github.com/artsy/force/pull/12153/commits/db39b0601aaeec4a9c1e6e468c96eb01c08feaa3

@damassi no type errors. thanks for showing an alternative tho!

PR #12159, opened 2023-03-28
feat: search pills

rajsam003:
The type of this PR is: **TYPE**

This PR solves [PROJECT-XX]

### Description


src/Components/Search/SearchBarPills.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  @media (hover: none) {
+    display: none;
+  }
+`
+
+export const SearchBarPills = () => {
+  return (
+    <Flex p={2}>
+      <Carousel
```
dimatretyak, line 86: Faced with the problem that `Carousel` cannot be scrolled with the mouse /touchpad. Therefore, it may be better to use `Flex` with `overflow-x: scroll` and without chevron arrows. Please remind me to discuss this moment with you tomorrow 🙏

dimatretyak: or use the `Shelf` component, but adding the required changes ([the possibility of pagination by cell](https://github.com/artsy/palette/tree/dimatretyak/sandbox/shelf), it is possible to pass custom navigation buttons, etc)

src/Components/Search/SearchBarPills.tsx
```diff
@@ -0,0 +1,161 @@
...
+    display: none;
+  }
+`
+
+export const SearchBarPills = () => {
+  const handlePress = (e, props) => {
+    e.stopPropagation()
+    e.preventDefault()
+    props.onClick && props.onClick(e)
+  }
```
dimatretyak, line 106: **suggestion**:
// or `ClickableProps`
const handlePress = (props: ShelfNavigationProps) => (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
  event.stopPropagation()
  event.preventDefault()
  props.onClick?.(event)
}

<PreviousChevron {...props} onClick={handlePress(props)} />
```

File: src/Components/Shelf/paginate.ts [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+  const sum = values.reduce((acc, curr) => acc + curr, 0)
+
+  // No need to paginate
+  if (sum <= viewport) {
+    return [0]
+  }
+
+  const offsets = compound(values)
+  const lastOffset = offsets[offsets.length - 1]
+  const offsetToLastCell = lastOffset - viewport + 100
```
rajsam003, line 60: @dimatretyak I imported this file and added 100 to this. Only way it worked

dimatretyak: unfortunately, I don't think this is a working solution that will be applicable for all cases (not just search bar pills)

dimatretyak: if it's not difficult, can you please explain why exactly `100` is added here?

rajsam003: Yes its not a solution. I am just showing you where I made a change to make it work. We need the right value for this.

damassi:
@rajsam003 - is this still needed with https://github.com/artsy/force/pull/12276?

rajsam003:
> @rajsam003 - is this still needed with #12276?

No it not. 
I am closing it. Thanks

PR #12163, opened 2023-03-28
refactor(phase 1): standardise and use `ArtworkList` terminology

dimatretyak:

This PR solves [PROJECT-XX]

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -1,7 +1,7 @@
 import { Shelf, Spacer } from "@artsy/palette"
 import { SavesArtworksQueryRenderer } from "./Components/SavesArtworks"
 import { SavesHeader } from "./Components/SavesHeader"
-import { SavesItemFragmentContainer } from "./Components/SavesItem"
+import { ArtworkListItemFragmentContainer } from "./Components/ArtworkListItem"
 import { FC, useRef } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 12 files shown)


src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -79,7 +80,7 @@ export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContai
       fragment CollectorProfileSaves2Route_me on Me {
         defaultSaves: collection(id: "saved-artwork") {
           internalID
-          ...SavesItem_item
+          ...ArtworkListItem_item
         }
 
         otherSaves: collectionsConnection(
```
dimatretyak, line 86: **question**: do we also need to rename `defaultSaves` and `otherSaves`? For example, to`allSavesArtworkList` or `defaultArtworkList` and `otherArtworkLists` or `customArtworkLists`

dimatretyak: > My view is that they all should be renamed so that we have as much consistency as possible.

Got it. I'll do it

dimatretyak: > like the fragment name CollectorProfileSaves2Route_me

@anandaroop I'll rename it to `ArtworkListsRoute_me`. is it ok?

anandaroop: I guess Relay requires some strictness around naming with respect to components? I leave it up to you the best way / best sequence for doing this, but yes that name sounds fine to me.

PR #12164, opened 2023-03-28
feat: area-based image resizing for artwork shelf

dzucconi:
This PR picks up off of https://github.com/artsy/force/pull/11126 — we wound up pulling over all the placeholder work from that initial PR so this winds up being pretty easy to drop in.

| Before | After |
| ---- | ---- |
| www artsy net_artist_lee-krasner |  localhost_5050_artist_lee-krasner |
| staging artsy net_artist_jackson-pollock | localhost_5050_artist_jackson-pollock |
| staging artsy net_artist_julie-mehretu | localhost_5050_artist_julie-mehretu |

### DIFF
--- a/src/Components/Artwork/ShelfArtwork.tsx
+++ b/src/Components/Artwork/ShelfArtwork.tsx
@@ -6,36 +6,51 @@ import Metadata, { MetadataPlaceholder } from "Components/Artwork/Metadata"
 import { AuthContextModule } from "@artsy/cohesion"
 import { Box, Image, SkeletonBox } from "@artsy/palette"
 import { useHoverMetadata } from "Components/Artwork/useHoverMetadata"
-import { resized } from "Utils/resized"
 import { ManageArtworkForSavesProvider } from "Components/Artwork/ManageArtworkForSaves"
+import { maxWidthByArea, resized } from "Utils/resized"
+
+const DEFAULT_AREA = 200 * 200
+const DEFAULT_MAX_IMG_HEIGHT = 250
+const DEFAULT_MAX_WIDTH = 500
 
 export interface ShelfArtworkProps
   extends Omit<RouterLinkProps, "to" | "width"> {
+  // Target number of pixels for image to occupy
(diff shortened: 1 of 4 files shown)


src/Components/Artwork/ShelfArtwork.tsx
```diff
@@ -119,27 +133,43 @@ export const ShelfArtworkFragmentContainer = createFragmentContainer(
...
 export const ShelfArtworkPlaceholder: React.FC<ShelfArtworkPlaceholderProps> = ({
   index,
-  width = [150, 175, 200],
+  hideSaleInfo,
+  area = DEFAULT_AREA,
+  maxImageHeight = DEFAULT_MAX_IMG_HEIGHT,
   children,
 }) => {
+  const width = [275, 200, 300, 250][index % 4]
+  const height = [200, 300, 250, 275][index % 4]
```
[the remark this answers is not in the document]
dzucconi: "Random" looking values. They create placeholder boxes.

dzucconi:
One very nice side effect of this work is that we no longer have any content shift on artist pages. Previously some images would sometimes be larger than the rail placeholder and so the content would get pushed once it loads. Now we just set a max height and images are more predictably within that height.

src/Utils/__tests__/resized.jest.ts
```diff
@@ -64,3 +64,39 @@ describe("#resized", () => {
     expect(srcSet).toContain("quality=80")
   })
 })
+
+describe("maxWidthByArea", () => {
```
TMcMeans, line 68: I can't say that I've seen tests like these before. Would you call them unit specs specifically for UI elements?

TMcMeans: or is it better to say that this is a unit test specifically testing the output of the `maxWidthByArea` function?

dzucconi: Just a unit test for a pure function.

PR #12199, opened 2023-04-03
chore(TX-1116): update inline links

rquartararo:

This PR solves [TX-1116]

### Description

This PR adds an `inline` prop to `RouterLink` which gives RouterLink styles recently updated in the [design system](https://www.figma.com/file/gZNkyqLT8AU3T61tluVJyB/Artsy-3.1-Design-System?node-id=6726%3A4204&viewport=-2947%2C-4962%2C0.88&t=QTyt3NMhKQSmKDKZ-11). 

I've also removed the deprecated `noUnderline` with `textDecoration="none"` in our `RouterLinks`.  

[TX-1116]: https://artsyproduct.atlassian.net/browse/TX-1116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Article/Components/ArticleVideo.tsx
+++ b/src/Apps/Article/Components/ArticleVideo.tsx
@@ -178,7 +178,7 @@ const ArticleVideo: FC<ArticleVideoProps> = ({ article }) => {
               <Text variant="lg-display">
                 More in{" "}
                 {article.seriesArticle ? (
-                  <RouterLink to={article.seriesArticle.href}>
+                  <RouterLink inline to={article.seriesArticle.href}>
                     {article.seriesArticle.title}
                   </RouterLink>
                 ) : (
--- a/src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
+++ b/src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
@@ -16,7 +16,7 @@ const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
       <Box>
         <Text variant="sm">Questions?</Text>
 
-        <RouterLink to="/how-auctions-work" target="_blank">
+        <RouterLink inline to="/how-auctions-work" target="_blank">
           <Text variant="sm">How to bid on Artsy?</Text>
         </RouterLink>
       </Box>
@@ -24,9 +24,9 @@ const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
       <Box>
         <Text variant="sm">Contact Us</Text>
 
-        <a href="mailto:specialist@artsy.net">
+        <RouterLink inline to="mailto:specialist@artsy.net">
           <Text variant="sm">specialist@artsy.net</Text>
-        </a>
+        </RouterLink>
         <Text variant="sm">+1-845-582-3967</Text>
       </Box>
     </Join>
--- a/src/System/Router/RouterLink.tsx
+++ b/src/System/Router/RouterLink.tsx
@@ -22,13 +22,11 @@ export type RouterLinkProps = Omit<
      */
     to: string | null
     textDecoration?: ResponsiveValue<string>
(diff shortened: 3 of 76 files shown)


src/Apps/Article/Components/ArticleVideo.tsx
```diff
@@ -178,9 +179,12 @@ const ArticleVideo: FC<ArticleVideoProps> = ({ article }) => {
               <Text variant="lg-display">
                 More in{" "}
                 {article.seriesArticle ? (
-                  <RouterLink to={article.seriesArticle.href}>
+                  <Link
+                    as={RouterLink}
+                    href={article.seriesArticle.href || undefined}
```
dzucconi, line 184: This doesn't make sense that it'd work since by casting with `as` we'd accept props for `RouterLink` so `href` should be `to`

src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
```diff
@@ -16,17 +16,17 @@ const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
       <Box>
         <Text variant="sm">Questions?</Text>
 
-        <RouterLink to="/how-auctions-work" target="_blank">
+        <Link as={RouterLink} href="/how-auctions-work" target="_blank">
```
dzucconi, line 19: This shouldn't be a `RouterLink` since it's just opening in a new tab

dzucconi:
Yeah, I'm not so sure this would work as expected. I'm expecting all the instances of `RouterLink` to no longer client-side navigate here.

rquartararo:
@dzucconi Hmm I've been testing to see if they are client-side navigating by changing the background color in dev tools and seeing if color persists when navigating, and it does, so I assumed these were working. However, I now realize that may not be the best way to test this. What would you recommend?

src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
rquartararo: Ah yes I didn't consider the `target="_blank"` I will update these ones

dzucconi:
Maybe it is working then! @damassi I seem to remember something about the framework auto-detecting normal links. Do we even need to be using `RouterLink`?

damassi:
Yup, we check for links within things like html, etc, via [this bit of code](https://github.com/artsy/force/blob/main/src/System/Router/buildAppRoutes.tsx#L36-L40), which I adapted from a Gatsby plugin. 

Note that this was meant to just be used as a migration step, and that we should always be using RouterLink. Something like `interceptLinks` is brittle and while it works, we should stick to generally-accepted framework patterns (aka Next).

dzucconi:
In that case I'd suggesting wrapping `RouterLink` in a new component (ummmm `LinkRouterLink` 😭? ), then `@ts-ignore` the inevitable type error on the `to` prop.

damassi:
Haha please no `LinkRouterLink` in 81 files. Can we discuss this in #practice-web channel?

src/System/Router/RouterLink.tsx
```diff
@@ -78,6 +76,12 @@ export const RouterAwareLink: React.FC<LinkPropsSimple & RouterLinkMixinProps> =
       return defaultValidatorFn(prop) || routerLinkValidator(prop)
     },
   })`
+    :hover {
+      color: ${props => props.inline && themeGet("colors.blue100")};
```
rquartararo, line 80: @damassi `<Link>` wrapped around `<RouterLink>` internally still defaults to `RouterLink` styles. I ended up just conditionally adding the styles directly to RouterLink, but let me know if I misinterpreted what you were suggesting.

damassi (dismissed):

dzucconi (dismissed):

PR #12207, opened 2023-04-05
feat(FX-4717): Respond with a 404 when navigating to a non-existent list

dimatretyak:

This PR solves [FX-4717]

Notion card: [link](https://www.notion.so/artsy/Internal-Server-Error-when-trying-to-load-custom-list-that-doesn-t-exist-Should-404-488676f9749e42f28e322002f46cf74b?pvs=4)

Review app: [artwork-list-404](https://artwork-list-404.artsy.net/collector-profile/saves2/unknown-artwork-list)

### Demo
https://user-images.githubusercontent.com/3513494/230105690-153cc163-a174-4fee-9177-c7a90243f8ea.mp4

[FX-4717]: https://artsyproduct.atlassian.net/browse/FX-4717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -2,7 +2,7 @@ import { Shelf, Spacer } from "@artsy/palette"
 import { ArtworkListContentQueryRenderer } from "./Components/ArtworkListContent"
 import { ArtworkListsHeader } from "./Components/ArtworkListsHeader"
 import { ArtworkListItemFragmentContainer } from "./Components/ArtworkListItem"
-import { FC, useEffect, useRef } from "react"
+import { FC, useEffect, useMemo, useRef } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { useRouter } from "System/Router/useRouter"
 import { extractNodes } from "Utils/extractNodes"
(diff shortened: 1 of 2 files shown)


src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -54,6 +55,26 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
...
+      )
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [selectedArtworkListId]
+  )
+
+  if (!selectedArtworkList) {
+    throw new HttpError(404)
+  }
+
```
dimatretyak, line 77: When the artwork list has been successfully deleted, we also clear all info about the deleted artwork list from the relay store in [updater](https://github.com/artsy/force/blob/51fc7478ab56131f47c1c2245e9f89c4abf41c59/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/Mutations/useDeleteArtworkList.ts#L62) before [redirecting](https://github.com/artsy/force/blob/51fc7478ab56131f47c1c2245e9f89c4abf41c59/src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/DeleteArtworkListModal.tsx#L63) to "All Saves".

For this reason, 404 error will be displayed immediately (see attached video). To prevent this from happening, `useMemo` is used here


https://user-images.githubusercontent.com/3513494/230107210-cb2fcb22-f32e-4152-a062-29ed780a7623.mp4

dblandin (approved):
Nice. This LGTM!

Aside: I'm a bit surprised that we aren't using a singular query field `collection(id: "")` for this route but I suppose that's because of how the page is built to support quick switching between lists? If we were using a singular query field, I think we could trigger this behavior automatically by using `@principalField` in the query.

dimatretyak:
> I suppose that's because of how the page is built to support quick switching between lists

yep, you're right

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
anandaroop: Thanks for explaining 🙏🏽

anandaroop (approved):

PR #12223, opened 2023-04-10
fix(GRO-1649): cleans up session only when executing redirect

dzucconi:
Closes: [GRO-1649](https://artsyproduct.atlassian.net/browse/GRO-1649)

So in the case of 3rd party auth `redirectBack` would get called twice. This function would always delete the `session.redirectTo` key, so the second time it was called it was falling back to `"/"`.

This alters the code so that it only clears the session key after the redirect takes place or after it's done for real.

[GRO-1649]: https://artsyproduct.atlassian.net/browse/GRO-1649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Server/passport/lib/app/__tests__/redirectBack.jest.ts
dimatretyak, line 1: thank you for these test cases 🙏

dimatretyak (approved):
LGTM. Thank you for fixes!

dimatretyak (approved):

PR #12227, opened 2023-04-11
chore: Add artist ID to SaveCollectedArtwork event

olerichter00:
Addresses [CX-3611]

## Description

Add artist ID to SaveCollectedArtwork event.


MrSltun (approved):
🚀

PR #12232, opened 2023-04-12
fix(FX-4726): Ensure collections show accurate count of artworks

nickskalkin:

This PR solves [FX-4726]

### Description

Sometimes we see that the count of artworks within the Artwork Lists rail differs from the count of artworks displayed in the corresponding grid. This is because artworks are getting unpublished/deleted. I've added changes to Gravity/Metaphysics to allow querying only visible artworks by passing the `onlyVisible` property.

### Demo

- [Demo app](https://lists-count-visible-artworks.artsy.net/)

For testing purposes, you can 'un-publish' an artwork by executing the following code in Gravity console:

```ruby
Artwork.find('slug').update!(private: true, published: false)
```

[FX-4726]: https://artsyproduct.atlassian.net/browse/FX-4726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/AddArtworksModalContent.tsx
```diff
@@ -139,6 +139,7 @@ const AddArtworksModalContentPaginationContainer = createPaginationContainer(
           sort: { type: CollectionArtworkSorts, defaultValue: POSITION_DESC }
         ) {
         collection(id: "saved-artwork") {
+          artworksCount(onlyVisible: true)
           artworksConnection(first: $first, after: $after, sort: $sort)
             @connection(key: "AddArtworksModalContent_artworksConnection") {
             totalCount
```
dimatretyak, line 145: **question**: do we need query `totalCount` if `artworksCount` is used?

dimatretyak (approved):

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/AddArtworksModalContent.tsx
nickskalkin: Not needed indeed. I thought that it's required for pagination logic

PR #12239, opened 2023-04-13
[WIP]feat: Search Dropdown imporvements

rajsam003:
The type of this PR is: Feat

This PR solves [FX-4651]. It improves the search drop-down implementation

TODO: 

- [x] Add fx-force-search-dropdown-improvements feature flag in [Unleash](https://unleash.artsy.net/features)
- [x] Convert the class-based component to a new functional search component. 
- [ ] Investigate whether we can use [AutosuggestInput](https://palette-storybook.artsy.net/?path=/story/components-autocompleteinput--default) from palette instead of Autosuggest component from react-autosuggest

[FX-4651]: https://artsyproduct.atlassian.net/browse/FX-4651?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/SearchBar.tsx
+++ b/src/Components/Search/SearchBar.tsx
@@ -1,4 +1,4 @@
-import { Component, useContext } from "react"
+import { useCallback, useContext, useEffect, useState } from "react"
 import * as React from "react"
 import { Box, BoxProps } from "@artsy/palette"
 import { SearchBar_viewer$data } from "__generated__/SearchBar_viewer.graphql"
@@ -32,7 +32,7 @@ import { Media } from "Utils/Responsive"
 import { SearchInputContainer } from "./SearchInputContainer"
 import { useTranslation } from "react-i18next"
 import { ClassI18n } from "System/i18n/ClassI18n"
-import track from "react-tracking"
+import track, { useTracking } from "react-tracking"
 import { getENV } from "Utils/getENV"
 
 const logger = createLogger("Components/Search/SearchBar")
@@ -85,29 +85,36 @@ const Form = styled.form`
 `
 
 // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-@track(null, {
+const SearchBar = track(null, {
   dispatch: data => Events.postEvent(data),
-})
-export class SearchBar extends Component<Props, State> {
-  public input: HTMLInputElement
+})(({ router, relay, viewer }) => {
+  const [entityID, setEntityID] = useState(null)
+  const [entityType, setEntityType] = useState(null)
+  const [focused, setFocused] = useState(false)
+  const [term, setTerm] = useState(getSearchTerm(window.location))
+  const { trackEvent } = useTracking()
 
   // Once this is set, we don't ever expect to change it back. A click on a
   // descendant indicates that we're going to navigate away from the page, so
   // this behaviour  is acceptable.
-  private userClickedOnDescendant: boolean
+  let userClickedOnDescendant: boolean
 
-  private removeNavigationListener: () => void
+  const removeNavigationListener = useCallback(() => {
+    if (!router) {
+      return null
+    }
 
-  // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-  state = {
-    entityID: null,
-    entityType: null,
-    focused: false,
-    term: getSearchTerm(window.location),
-  }
+    return router.addNavigationListener(location => {
+      if (!location.pathname.startsWith("/search")) {
+        setTerm("")
+      }
+
+      return true
+    })
+  }, [router])
 
   // Throttled method to toggle previews.
-  throttledOnSuggestionHighlighted = ({ suggestion }) => {
+  const throttledOnSuggestionHighlighted = ({ suggestion }) => {
     if (!suggestion) return
 
     const {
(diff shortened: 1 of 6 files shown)


src/Components/Search/SearchBar.tsx
```diff
@@ -85,29 +76,36 @@ const Form = styled.form`
...
 // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-@track(null, {
+const SearchBar = track(null, {
   dispatch: data => Events.postEvent(data),
-})
-export class SearchBar extends Component<Props, State> {
-  public input: HTMLInputElement
+})(({ router, relay, viewer }) => {
+  const [, setEntityID] = useState(null)
+  const [, setEntityType] = useState(null)
```
rajsam003, line 83: After converting to a functional component, I noticed that we are not using this state anywhere.

src/Components/Search/SearchBar.tsx
```diff
@@ -116,79 +114,58 @@ export class SearchBar extends Component<Props, State> {
...
-    this.setState({
-      entityID,
-      entityType,
-    })
+    setEntityID(entityID)
+    setEntityType(entityType)
   }
 
-  @track((_props, _state, [query, hasResults]) => ({
+  track((_props, _state, [query, hasResults]) => ({
```
dimatretyak, line 121: what does this function do? Sorry if this is a stupid question 😅

src/Components/Search/SearchBar.tsx
```diff
@@ -410,72 +362,64 @@ export class SearchBar extends Component<Props, State> {
...
+          method="GET"
+          onSubmit={event => {
+            if (router) {
+              event.preventDefault()
+              const encodedTerm = encodeURIComponent(term)
+
+              // TODO: Reenable in-router push once all routes have been moved over
+              // to new app shell
+              // router.push(`/search?term=${this.state.term}`)
+              window.location.assign(`/search?term=${encodedTerm}`)
```
dimatretyak, line 405: can we use the suggested approach with `router.push'?

src/Components/Search/SearchBar.tsx
```diff
@@ -376,18 +331,15 @@ export class SearchBar extends Component<Props, State> {
     )
   }
 
-  renderInputComponent = props => <SearchInputContainer {...props} />
-
-  renderAutosuggestComponent(t, { xs }) {
-    const { term } = this.state
-    const { viewer } = this.props
+  const renderInputComponent = props => <SearchInputContainer {...props} />
```
dimatretyak, line 334: I think we can pass `SearchInputContainer` component for `renderInputComponent` prop. Like this

```javascript
<Autosuggest
  ...
  renderInputComponent={SearchInputContainer}
/>
```

src/Components/Search/SearchBar.tsx
```diff
@@ -376,18 +331,15 @@ export class SearchBar extends Component<Props, State> {
...
 
+  const renderAutosuggestComponent = (t, { xs }) => {
```
dimatretyak, line 336: can we split it into separate component?

src/Components/Search/SearchBar.tsx
```diff
@@ -354,10 +309,10 @@ export class SearchBar extends Component<Props, State> {
     )
   }
 
-  renderDefaultSuggestion = (edge, { query, isHighlighted }) => {
+  const renderDefaultSuggestion = (edge, { query, isHighlighted }) => {
```
dimatretyak, line 312: can we split it into separate component?

src/Components/Search/SearchBar.tsx
```diff
@@ -116,79 +114,58 @@ export class SearchBar extends Component<Props, State> {
...
-    )
+    throttle(throttledOnSuggestionHighlighted, 500, { leading: true })
 
-    // Clear the search term once you navigate away from search results
-    this.removeNavigationListener = this.props.router
-      ? this.props.router.addNavigationListener(location => {
-          if (!location.pathname.startsWith("/search")) {
-            this.setState({ term: "" })
-          }
+    return removeNavigationListener
```
dimatretyak, line 165: we need to subscribe to `NavigationListener` on mount and **un**subscribe on **un**mount. in the current implementation, you are trying to subscribe during the unmount stage

src/Components/Search/SearchBar.tsx
```diff
@@ -116,79 +114,58 @@ export class SearchBar extends Component<Props, State> {
...
+  }, [removeNavigationListener, router, throttledFetch])
```
dimatretyak, line 166: you don't need to pass `throttledFetch` in deps for `useEffect`. Otherwise, the `useEffect` body will be called every time the `throttledFetch` changes (in the old implementation, this was called only at the mount stage)


https://user-images.githubusercontent.com/3513494/231861951-0d0b925d-b82c-4f76-8e4e-90dc55853690.mp4

dimatretyak: and should we pass `router` in deps?

src/Components/Search/SearchBar.tsx
```diff
@@ -116,79 +114,58 @@ export class SearchBar extends Component<Props, State> {
...
-  }
+      )
+    },
+    [relay, viewer]
+  )
 
-  componentDidMount() {
-    this.throttledFetch = throttle(this.throttledFetch, 500, {
+  useEffect(() => {
+    throttle(throttledFetch, 500, {
```
dimatretyak, line 159: you are using `throttle` incorrectly here (see the examples in [the Lodash documentation](https://lodash.com/docs/4.17.15#throttle)). You can also find usage examples in Eigen (like [this](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Components/ArtworkFilter/Filters/KeywordFilter.tsx#L59), [this](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Scenes/Artwork/Components/ImageCarousel/FullScreen/DeepZoom/DeepZoomLevel.tsx#L244), [this](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Scenes/Search/Search2.tsx#L79), [this](https://github.com/artsy/eigen/blob/07007f404286907636c793c514cade5847c4fdef/src/app/Scenes/Search/components/SearchInput.tsx#L30) and etc)

dimatretyak (commented):
**suggestion**: as part of this task, add a new search dropdown component that will not display anything. Add all other logic in another PRs. For example:
* Display search results (first PR)
* Restore throttle logic (second PR)
* Track analytic events (third PR)
* etc

In this case, many small PRs are better than one large one. Otherwise, now there is a very big chance to miss a bug/not notice something and break some logic

rajsam003:
This PR was supposed to be Drafted. Sorry, my bad. 
Yes @dimatretyak I will create mini PR's. Thanks for reviewing :)

PR #12244, opened 2023-04-14
fix(CX-3619): change tabs text size

dariakoko:

This PR solves [CX-3619]

### Description

Adjust font size for Tabs in Collector Profile mweb to xs 

https://user-images.githubusercontent.com/36167539/231989099-af248376-c24c-4c43-ba51-b951080fd3f3.mov

[CX-3619]: https://artsyproduct.atlassian.net/browse/CX-3619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/CollectorProfileApp.tsx
+++ b/src/Apps/CollectorProfile/CollectorProfileApp.tsx
@@ -47,18 +47,24 @@ const CollectorProfileApp: React.FC<CollectorProfileAppProps> = ({
       <Spacer y={[0, 2]} />
 
       <RouteTabs fill>
-        <Tab to="/collector-profile/my-collection">My Collection</Tab>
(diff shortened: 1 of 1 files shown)


MrSltun (approved):
🌟🌟🌟🌟🌟

dzucconi:
Why does this keep happening. This is not correct. If you want to change this then we should change it for the entire design system. But I do not recommend that, tap areas should not be smaller on mobile.

dzucconi:
(I've rolled this back in 02af4ae80ff73fa2481e75906e3c5ec470fca4c3 just so we don't have a random one-off inconsistent tab bar for the time being.)

PR #12248, opened 2023-04-14
chore(FX-4735): Rename `All Saves` to `Saved Artworks`

dimatretyak:

This PR solves [FX-4735]

### Demo
| Before | After |
| before-1 | Content Cell | 
| before-2 | after-2 |


src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListEmptyState.tsx
```diff
@@ -11,9 +11,9 @@ interface ArtworkListEmptyStateProps {
 export const ArtworkListEmptyState: FC<ArtworkListEmptyStateProps> = ({
   me,
 }) => {
-  const allSavesArtworksCount = me.allSavesArtworkList?.artworksCount ?? 0
+  const savedArtworksCount = me.savedArtworksArtworkList?.artworksCount ?? 0
```
dimatretyak, line 14: I will be glad to any suggestions about the variable name 😅

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -135,7 +135,7 @@ export const CollectorProfileSaves2RouteFragmentContainer = createFragmentContai
   {
     me: graphql`
       fragment CollectorProfileSaves2Route_me on Me {
-        allSavesArtworkList: collection(id: "saved-artwork") {
+        savedArtworksArtworkList: collection(id: "saved-artwork") {
```
dblandin, line 138: **suggestion**: Maybe to avoid further churn we can call this "defaultCollection" or "standardCollection"?

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListEmptyState.tsx
```diff
@@ -78,6 +75,7 @@ const getText = (
 
   return {
     title: "Start curating your list of works",
-    description: "Add works from All Saves or add new artworks as you browse.",
+    description:
+      "Add works from Saved Artworks or add new artworks as you browse.",
```
dblandin, line 79: **suggestion**: Can we keep some of these strings in the newish [I18n translations file](https://github.com/artsy/force/blob/main/src/System/i18n/locales/en-US/translation.json)?

src/Apps/CollectorProfile/Routes/Saves2/Components/Actions/__tests__/EditArtworkListModal.jest.tsx
```diff
@@ -151,7 +151,7 @@ describe("EditArtworkListModal", () => {
           __typename: "UpdateCollectionSuccess",
           artworkList: {
             internalID: "artwork-list",
-            name: "All saves",
+            name: "Saved Artworks",
```
dblandin, line 154: **suggestion**: Because this is hard-coded, maybe we use something that's obviously a placeholder like `[default]` or `[standard]`? /shrug

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
dimatretyak: we recently made a small agreement on the use of uniform terminology ([link 1](https://artsy.slack.com/archives/C9SATFLUU/p1680012458469089?thread_ts=1680012130.077609&cid=C9SATFLUU), [link 2](https://www.notion.so/artsy/Artworks-list-I-think-we-ve-got-a-confusing-mix-of-terms-in-our-codebase-for-Artwork-Lists-collec-03de616986ab4cd4a839ec6aed989d3d))

@anandaroop what do you think about the suggested changes by Devon?

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListEmptyState.tsx
dimatretyak: we can do this, but we will need to make quick refactor. do you mind if I do this in a separate PR?

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
anandaroop: I'd be fine with the change here, or `defaultArtworkList` as well. 

I like the idea of sticking to the pattern where Relay-and-upstream continues to use the "connection" terminology, whereas client-side code aligns on "artwork list", [using the Relay query as the point where we "translate."](https://github.com/artsy/force/pull/12171#discussion_r1153273394)

nickskalkin (approved):

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
dimatretyak: I'm afraid this will add more confusion. for example, somewhere there will be a mention of "default artwork list", somewhere "saved artworks list". for this reason, I decided to leave only the current name.

In any case, thank you all very much for your participation and discussions 🙏

PR #12276, opened 2023-04-19
feat: add search pills

rajsam003:
The type of this PR is: feat

This PR solves [FX-4653]

- [x] Add AutoCompleteInput from palette
- [ ] Replace the mock options with fetched data. - We are working on this on this ticket-  https://artsyproduct.atlassian.net/browse/FX-4654

[FX-4653]: https://artsyproduct.atlassian.net/browse/FX-4653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+export const NewSearchInputPillsContainer = () => {
+  const pillsRef = useRef<HTMLDivElement>(null)
+  const [showPreviousChevron, setShowPreviousChevron] = useState<boolean>(false)
+  const [showNextChevron, setShowNextChevron] = useState<boolean>(true)
+
+  const scrollToLeft = () => {
+    if (pillsRef.current) {
+      pillsRef.current.scrollBy({
+        left: -100,
```
dimatretyak, line 77: we shouldn't always scroll to the same offset

dimatretyak (changes requested):

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
dimatretyak: see how this is implemented in the [Carousel](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Carousel/Carousel.tsx) and [Shelf](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Shelf/Shelf.tsx) components

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+        behavior: "smooth",
+      })
+    }
+  }
+
+  const handleScroll = () => {
+    if (pillsRef.current) {
+      const currentPosition = pillsRef.current.scrollLeft
+      const maxScroll =
+        pillsRef.current.scrollWidth - pillsRef.current.clientWidth
```
dimatretyak, line 96: please also remember that the user can resize the browser window and you have to recalculate this

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+      <PillsContainer
+        ref={pillsRef as any}
+        px={2}
+        py={2}
+        onScroll={handleScroll}
+      >
+        {PILLS_OPTIONS.map(pill => (
+          <Flex key={pill}>
+            <Pill mr={1}>{pill}</Pill>
+          </Flex>
```
dimatretyak, line 131: **question**: can we just use `Pill` without `Flex`?

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,203 @@
...
+        const pill = pills[i] as HTMLElement
+        if (
+          pill.offsetLeft + pill.offsetWidth <=
+          currentScroll + visibleWidth
+        ) {
+          const prevPill = pills[i - 1] as HTMLElement
+          scrollBy = -prevPill.offsetWidth
+          break
+        }
+      }
```
dimatretyak, line 122: **suggestion**: i think we can do this logic once when mounting the component or when receiving the `resize` event

dimatretyak: how do we do it for [Carousel](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Carousel/Carousel.tsx#L133-L144)

dimatretyak: Ah, maybe I misunderstood

dimatretyak: **question**: can we use the same approach with `useCursor` hook, as we do with the `Carousel` ([link 1](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Carousel/Carousel.tsx#L133-L144), [link 2](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Carousel/Carousel.tsx#L128)) and `Shelf` ([link 1](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Shelf/Shelf.tsx#L117), [link 2](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Shelf/Shelf.tsx#L156)) components?

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  return (
+    <Flex alignItems="center" bg="white">
+      <Flex position="absolute" left={0} zIndex={showPreviousChevron ? 0 : -1}>
+        <PreviousChevron onClick={scrollToLeft} left={2} />
+        <GradientBg placement="left" />
+      </Flex>
```
dimatretyak, line 183: **question**: should we unmount, instead of changing `zIndex` when `showPreviousChevron` is `false`? Like this

```javascript
{showPreviousChevron && (
  <Flex position="absolute" left={0} zIndex={1}>
    <PreviousChevron onClick={scrollToLeft} left={2} />
    <GradientBg placement="left" />
  </Flex>
)}
```

dimatretyak: the same for the next chevron button

src/Components/Search/NewSearch/NewSearchBarInputContainer.tsx
```diff
@@ -38,6 +39,7 @@ export const NewSearchBarInputContainer = ({ isXs }: { isXs: boolean }) => {
       spellCheck={false}
       options={value.length < 2 ? [] : OPTIONS}
       value={value}
+      header={<NewSearchInputPillsContainer />}
```
dimatretyak, line 42: 🔥🔥🔥

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,203 @@
...
+  scrollbar-width: none;
+
+  &::-webkit-scrollbar {
+    display: none;
+  }
+`
+
+const GRADIENT_BG_WIDTH = 30
+
+export const NewSearchInputPillsContainer = () => {
```
damassi, line 71: Lets be really diligent about refactoring out / deleting the "old" search when this launches, so that there's not components with "New" in the title all about. This problem is quite chronic in Eigen so want to be sure it doesn't spread to Force!  

Another approach would be to rename existing search to `LegacySearch` _right now_, and then this component, rather than being called `NewSearch`, can be renamed to `Search`. If this work isn't a spike then this feels like a safe approach.

damassi (commented):
Lets be really diligent about refactoring out / deleting the "old" search when this launches, so that there's not components with "New" in the name left all about. This problem is quite chronic in Eigen so want to be sure it doesn't spread to Force!

Another (suggested) approach would be to rename existing search to LegacySearch right now, and then this component, rather than being called NewSearch, can be renamed to Search. If this work isn't a spike then this feels like a safe approach.

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,203 @@
...
+        setShowNextChevron(true)
+      } else if (currentPosition === maxScroll) {
+        // Hide NextChevron when pills reaches the end
+        setShowPreviousChevron(true)
+        setShowNextChevron(false)
+      } else {
+        // Show both chevrons otherwise
+        setShowPreviousChevron(true)
+        setShowNextChevron(true)
+      }
```
dimatretyak, line 174: **suggestion**: i think we can combine all these checks. like this

```javascript
const isAtStart = currentPosition === 0
const isAtEnd = currentPosition === maxScroll

setShowPreviousChevron(!isAtStart)
setShowNextChevron(!isAtEnd)
```

dimatretyak:
> Lets be really diligent about refactoring out / deleting the "old" search when this launches, so that there's not components with "New" in the name left all about. This problem is quite chronic in Eigen so want to be sure it doesn't spread to Force!

> Another (suggested) approach would be to rename existing search to LegacySearch right now, and then this component, rather than being called NewSearch, can be renamed to Search

Sorry, it's my bad (I suggested to do this)

damassi:
It's all good! However y'all feel is best to proceed, proceed. Just be sure to file a ticket to remove the "New" language later, and rename old to Legacy, or to delete (or something of this sort 😄 )

dimatretyak:
> This would be caught by the type-checker, hopefully!

I was more referring to the following case:
1. We have `SearchItem` component
2. This component is used in the `SearchList`
3. We renamed `SearchItem` to `LegacySearchItem`
4. We created `SearchItem`, but with new design
5. We renamed `SearchList` to `LegacySearchList`, but **forgot** to update the name/import path from `SearchItem` to `LegacySearchItem`. And as a result, we will use the `SearchItem` with new design

damassi:
Sorry for the confusion, to be clear, using `New` in the name is totally fine 👍

dimatretyak:
> We just want to avoid, in principle, "NewFoo" components lingering the codebase after a feature has launched

Got it. thanks for the clarifications 🙏

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
dimatretyak: or better
1. get pills offsets on mount/resize event and keep it in local state (e.g. `offsets = [0, 100, 150, 250, 300]`)
2. keep index of current offset in local state
4. when user clicks the right chevron button -> scroll to the next offset (e.g. `scrollTo(offsets[currentOffsetIndex + 1])`)
5. when user clicks the left chevron button -> scroll to the prev offset (e.g. `scrollTo(offsets[currentOffsetIndex - 1])`)
6. when user scrolls - get nearest offset index and save it in local state (e.g. [like this](https://github.com/artsy/palette/blob/ffa2c198375e7853516cae20d7f6002db48b4b5c/packages/palette/src/elements/Shelf/Shelf.tsx#L110-L117))

You can find more examples/code in Shelf component: https://github.com/artsy/palette/blob/ffa2c198375e7853516cae20d7f6002db48b4b5c/packages/palette/src/elements/Shelf/Shelf.tsx#L133-L166

dimatretyak: Pushed some code with the proposed changes to the [dimatretyak/sandbox/scrollable-pills](https://github.com/artsy/force/compare/dimatretyak/sandbox/scrollable-pills) branch. This is not the final version yet and improvements are needed, but it is already close to what we want (as it seems to me)


https://user-images.githubusercontent.com/3513494/233492367-730580d3-d3f6-4aea-82c6-b10dbd64a52f.mp4

rajsam003: Thanks for the suggestions, Dima. I see what you mean. 
My use case is when the user clicks the next button the next pill should be fully visible. I know you are noticing the pill next to the previous button. Which is also good to take into account. But it would make more sense if the pill near the next button is visible when the next button is clicked and the pill near the previous button is visible when the previous button is clicked. 
What do you think?

rajsam003: I would love to pair on this with you :)

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
rajsam003: With ZIndex I wrote less code so I think its good like this. But if you have a better reason to have mount/unmount logic we can replace it 👍🏽.

dimatretyak: I think then instead of changing `zIndex` (usually a negative value is not used in Force) it is better to use [this approach](https://github.com/artsy/palette/blob/ffa2c198375e7853516cae20d7f6002db48b4b5c/packages/palette/src/elements/Shelf/ShelfNavigation.tsx#L29-L30)

dimatretyak (approved):

src/Components/Search/NewSearch/NewSearchInputPillsContainer.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const scrollToLeft = () => {
+    if (pillsRef.current) {
+      const pillsContainer = pillsRef.current
+      const pills = pillsContainer.children
+      const currentScroll = pillsContainer.scrollLeft
+      const visibleWidth = pillsContainer.offsetWidth
+      let scrollBy = 0
+
+      // Find the last visible pill and the next pill
+      for (let i = pills.length - 1; i >= 0; i--) {
```
dimatretyak, line 107: **question**: can we use `find` here instead of `for; break`?

PR #12280, opened 2023-04-19
feat: Segment Cookie Consent Manager

dzucconi:
### Review App

| Region | URL | Behavior |
| --- | --- | --- |
| US (sans California) | https://cookie-consent.artsy.net/ | Don't display |
| EU | https://cookie-consent.artsy.net/?geo=eu | Opt-in |
| California | https://cookie-consent.artsy.net/?geo=ca | Opt-out |
| Brazil | https://cookie-consent.artsy.net/?geo=br | Out-out |

### Overview

* We use Segment's `ConsentManagerBuilder` to load Segment integrations.
  * It will automatically handle any integrations which are enabled through Segment
  * For other things that introduce 3rd party cookies we include them as custom preferences then use the provider to expose those preferences and block or enable those features.
* We use timezone information to determine initial opt-in/out preferences and whether or not to display the banner to users
* We log an event for auditing to Segment when preferences are saved, it includes the tracking preferences as the `value`.

-----


src/server.ts
```diff
@@ -51,7 +50,6 @@ app.get(
  */
 
 app
-  .get("/set-tracking-preferences", setTrackingPreferences)
```
dzucconi, line 54: @damassi I'm aware of this and it will be dealt with.

src/Components/CookieConsentManager/CookieConsentManager.tsx
```diff
@@ -0,0 +1,255 @@
...
+}, {})
+
+export const CookieConsentManager: FC = () => {
+  const [mode, setMode] = useMode<
+    "Idle" | "Open" | "Allowing" | "Rejecting" | "Saving"
+  >("Idle")
+
+  return (
+    <ConsentManagerBuilder
+      shouldRequireConsent={() => true}
```
dzucconi, line 38: Right now, for ease of testing, I'm forcing this, but this is where we'd use Segments [in-eu](https://github.com/segmentio/in-eu) package.

src/Components/CookieConsentManager/categories.ts
```diff
@@ -0,0 +1,34 @@
...
+      "These cookies enable the website to provide enhanced functionality and personalisation. They may be set by us or by third party providers whose services we have added to our pages. If you do not allow these cookies then some or all of these services may not function properly.",
+  },
+  {
+    key: "performance",
+    name: "Performance Cookies",
+    description:
+      "These cookies allow us to count visits and traffic sources so we can measure and improve the performance of our site. They help us to know which pages are the most and least popular and see how visitors move around the site. If you do not allow these cookies we will not know when you have visited our site, and will not be able to monitor its performance.",
+  },
+]
+
```
artsyjian, line 27: OneTrust has two [additional categories](https://www.notion.so/artsy/Cookie-Consent-01dace52a765476ea4e9353260ce9a7f?pvs=4#deb05d1a4f2c45f89e7c5ad54b0089f5): `Targeting`, `Social Media`. I think the way OneTrust works is that if there's no cookie for a category, their  Pref Center does not show that category for toggling. But it looks like the intent here is to display all categories?

dzucconi: I suppose it's a decision to make. One benefit of displaying all categories is that we'd be able to imply consent for future destinations/integrations based on that category.

That said; where do we have advertising cookies on the site, since I see that category show up in the OneTrust dialog?

artsyjian: looks like EU pref center has `Targeting` as a sub-category of `Advertising`. so it's really `Targeting` cookie that will cause `Advertising` category to be shown?
Screen Shot 2023-04-20 at 10 22 44 AM

oof. so compilcated.

there are `Targeting` cookies in the [table](https://www.notion.so/artsy/e8b6666d69204f8c95c494b16c8d68d7?v=e7b597d613b14258844dde558432c50b&pvs=4). [`__gads`](https://www.notion.so/artsy/__gads-3e835fa674354af5844d77343fd648f4?pvs=4), for example, which is dropped on an [Article page](https://www.artsy.net/article/artsy-editorial-warhol-basquiat-ultimate-art-collab-joint-works-more).

dzucconi: Well, those are internal ads — we're not re-targeting anyone with that. We can of course add the advertising category and use it to hide or show those ads completely but I'm not so sure it's needed?

artsyjian:
> That leaves the only actual integration that is blocked or not to be Google Analytics, which we're categorizing as "Performance".
I guess. Sounds like a [bunch of the integrations in this table](https://www.notion.so/artsy/Amazon-S3-ee8598403d564e26adf51b82ef19fe4f?pvs=4) no longer exist?

dzucconi:
For example, let's say we deemed YouTube embeds to be a problem. We'd add a custom destination preference in this consent manager, lift the state of it up into a React context, then check for consent when including the embeds in articles. If the consent was denied we'd have to load a placeholder there, maybe with a link that let's you open this manager back up and adjust it.

Do YouTube embeds currently get blocked by OneTrust? I somehow doubt that's even possible.

artsyjian (dismissed):
lgtm

src/Server/middleware/bootstrapSharifyAndContextLocalsMiddleware.ts
```diff
@@ -15,7 +15,9 @@ export function bootstrapSharifyAndContextLocalsMiddleware(
 ) {
   const ua = req.get("user-agent") || ""
 
-  updateSharifyAndContext(res, "CURRENT_PATH", parse(req.url).pathname)
+  const parsed = parse(req.url)
+  updateSharifyAndContext(res, "CURRENT_PATH", parsed.pathname)
+  updateSharifyAndContext(res, "CURRENT_SEARCH", parsed.search)
```
damassi, line 20: is it possible to rename this to something more verbose, like `CURRENT_SEARCH_PARAMS` or `CURRENT_QUERY_PARAMS`? From the outside, it looks like something related to "search", and lots of devs aren't aware that search = query params

damassi: Its just that its common to pass over actual search stuff from the server to the client and so on. Disambiguating this would be helpful. `CURRENT_LOCATION_SEARCH`, though params communicates something more effectively, even if it needs to be parsed.

dzucconi: I'd be more inclined to just stick a fully qualified URL in here and then we just parse it on the client.

I can just refactor this hook to not run on the server though. I'll do that instead.

damassi: Either way 😄. Ambiguous ENV vars placed in sharify, etc, cause trouble through time, so as long as that's clarified a bit and / or removed.

src/server.ts
damassi: Are you sure that's sufficient to evade safari's clearing? According to their engineers, the only way to persist is if one sets a secure cookie on the server (i could be misunderstanding the domain point above)

dzucconi: Interesting and weird. Might as well do it then I suppose.

damassi: Interestingly, this seems to not be more widely known 🤷

dzucconi: Resolved in https://github.com/artsy/force/pull/12280/commits/871452a1a15c10847a50fe08e3ea1f9cdcdfe48a

src/Apps/Article/Components/ArticleAd/ArticleAd.tsx
```diff
@@ -30,21 +37,40 @@ export const ArticleAd: FC<ArticleAdProps> = memo(({ unit, size, ...rest }) => {
...
-        >
-          <AdSlot
-            adUnit={unit}
-            sizes={[[width, height]]}
-            onSlotIsViewable={displayedAd}
-          />
-        </Box>
+        {ready && (
+          <>
+            {isDestinationAllowed("Google Ads") ? (
```
damassi, line 42: is there any way we could roll `ready` up into `isDestinationAllowed` so that additional bit of (necessary) logic is transparent to future devs?

dzucconi: Actually, that's not desirable. Note what's actually happening here. When we aren't `ready` we don't display anything`, when we are we display either the button or the ad. That's because we don't want to display `"Manage cookies"` to US users. There's 3 states: loading, requires consent, content

damassi (approved):
One comment about simplifying the consumer API a bit to remove the need for `ready` (or make more clear), and another suggestion about adding some test coverage -- otherwise, amazing work here 👍

src/Components/CookieConsentManager/categories.ts
```diff
@@ -0,0 +1,165 @@
...
+    case SEGMENT_CATEGORIES.performance.includes(destination.category):
+      return "performance"
+    case SEGMENT_CATEGORIES.targeting.includes(destination.category):
+      return "targeting"
+    case SEGMENT_CATEGORIES.functional.includes(destination.category):
+      return "functional"
+    default:
+      return DEFAULT_CATEGORIZATION
+  }
+}
```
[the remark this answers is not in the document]
dzucconi: https://github.com/artsy/force/pull/12280/files/2f368ad13f09d20f1e265663edb8825bde56a22b#diff-d6c2ede592b983aab527b4ecc96ddce6b8047e6a0865d7337454174d5cd222dc implicitly tests this but yeah I'll add some more coverage.

src/Apps/Article/Components/ArticleAd/ArticleAd.tsx
damassi: Its just a bit hard to know how to use ready, should other areas be needed in the future. The description provided above doesn't quite translate into the code at a glance. 

If this is necessary complexity then no problem. And maybe it could be resolved by updating the name `ready` to something less general.

src/Components/CookieConsentManager/categories.ts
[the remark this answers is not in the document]
damassi: If you feel confident that the implicit coverage is enough, no problem! Code is all simple here, but there's some details that seem easy to break.

src/Apps/Article/Components/ArticleAd/ArticleAd.tsx
dzucconi: And actually, `ready` _is_ rolled into that function as well. In some cases that might be fine. I just expose `ready` for when it's not.

damassi: Cool, nothing blocking here as long as you think it's clear enough for the future 👍

dzucconi: I'll add a README

PR #12198, opened 2023-04-03
feat(GRO-1641): Updated navbar dropdowns

dzucconi:
Closes [GRO-1641](https://artsyproduct.atlassian.net/browse/GRO-1641)

https://user-images.githubusercontent.com/112297/229583405-39dfe904-a24a-44fb-955f-d855be5170a4.mp4

[GRO-1641]: https://artsyproduct.atlassian.net/browse/GRO-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12281, opened 2023-04-20
fix(settings): clean up

dzucconi:
This just pulls over some commits from the now closed https://github.com/artsy/force/pull/12198

And includes some additional house keeping.


src/Apps/Settings/SettingsApp.tsx
```diff
@@ -44,48 +41,20 @@ const SettingsApp: React.FC<SettingsAppProps> = ({ me, children }) => {
...
-  if (!isLoggedIn) {
-    return <MyCollectionRouteLoggedOutState />
-  }
-
   return (
     <>
       <MetaTags title="Settings | Artsy" />
 
-      <>
-        <Breadcrumbs py={[2, 1]}>
```
dzucconi, line 66: This was a bit of abuse of our design system: These aren't breadcrumbs, which leave a trail of the hierarchal relationship:



We also would never want to just overwrite all the styles to it.

What you seemed to be looking for is the `TopContextBar` which describes the current context under which a page might be nested. Like a singular breadcrumb. It's much easier to use and does not require messing with the styling.

src/Apps/Settings/SettingsApp.tsx
```diff
@@ -44,48 +41,20 @@ const SettingsApp: React.FC<SettingsAppProps> = ({ me, children }) => {
...
 
-        <Spacer y={4} />
-      </>
+      <Spacer y={[4, 6]} />
 
       <RouteTabs my={SETTINGS_ROUTE_TABS_MARGIN}>
-        {tabsWhenCollectorProfileEnabled.map(tab => {
+        {SETTINGS_TABS.map(tab => {
           return (
-            <RouteTab key={tab.url} to={tab.url} variant={["xs", "sm"]}>
```
dzucconi, line 88: Avoid overwriting styles coming from the design system. On mobile you want tap areas to be bigger, really, anyway.

src/Server/middleware/hardcodedRedirects.ts
```diff
@@ -71,6 +71,7 @@ const PERMANENT_REDIRECTS = {
   "/user/payments": "/settings/payments",
   "/user/alerts": "/settings/alerts",
   "/my-collection": "/collector-profile/my-collection",
+  "/settings/my-collection": "/collector-profile/my-collection",
```
dzucconi, line 74: Said redirect.

dariakoko (dismissed):

PR #12283, opened 2023-04-20
feat: alerts option in user dropdown

dzucconi:
This surfaces the alerts in the dropdown to allow us to guide through progressive onboarding.


damassi:
@dzucconi - did you pop that font-size commit off? Planning on pushing into another PR?

dzucconi:
I did: https://artsy.slack.com/archives/C01B2P6LJUU/p1681397838816459 — would make sense as it's own PR anyway.

PR #12292, opened 2023-04-24
feat: search pills backend data wire up

rajsam003:
The type of this PR is: Feat

This PR solves [FX-4654]

This PR is the continuation of the Search bar improvements PRs. It wires the UI with the backend data. As we are trying to replace the old SearchBar to New one it requires a lot of code changes. So we are excluding some features for now as in -
- On Pill click
- Tracking 
- Throttle functions
- Increase the height of the Search bar suggestion container. 

We will be working on them in future PR's. 

image

Paired with @nickskalkin 

[FX-4654]: https://artsyproduct.atlassian.net/browse/FX-4654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -0,0 +1,231 @@
...
+  return (
+    <AutocompleteInput
+      placeholder={isXs ? t`navbar.searchArtsy` : t`navbar.searchBy`}
+      spellCheck={false}
+      options={value.length < 2 ? [] : formattedOptions}
+      value={value}
+      onChange={handleChange}
+      onClear={() => setValue("")}
+      header={<NewSearchInputPillsFragmentContainer viewer={viewer} />}
+      // renderOption={option => <NewSearchBarSuggestion option={option} />}
```
dimatretyak, line 87: **question**: do we need it?

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -0,0 +1,231 @@
...
+  relay: RelayRefetchProp
+  router?: Router
+  viewer: NewSearchBarInput_viewer$data
+  isXs: boolean
+}
+
+const NewSearchBarInput: FC<NewSearchBarInputProps> = ({
+  isXs,
+  relay,
+  router,
```
dimatretyak, line 34: do we need it? If not, then we can remove it and `withSystemContext` [here](https://github.com/artsy/force/blob/32245276e87a474e36b1ed957608d2a7a30e9a26/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L114)

src/Components/Search/NewSearch/NewSearchBarFooter.tsx
```diff
@@ -0,0 +1,59 @@
...
+  showArtworksButton?: boolean
+  showAuctionResultsButton?: boolean
+}
+
+export const NewSearchBarFooter: React.FC<SuggestionItemProps> = ({
+  href,
+  isHighlighted,
+  query,
+}) => {
+  const handleClick = event => {
```
dimatretyak, line 23: please, don't forget to add TypeScript definitions for args

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: removed `router` for now, but I think we will need `withSystemContext` if we want to replicate [old search behaviour](https://github.com/artsy/force/blob/main/src/Components/Search/SearchBar.tsx#L481)

dimatretyak: > we will need `withSystemContext`

Yes, that's about what I thought. But it is better to use `useRouter` (as far as I understand, we will need `route`) instead of `withSystemContext` and for this reason I suggested to delete it for now

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -0,0 +1,228 @@
...
+  `
+)
+
+interface NewSearchBarInputQueryRendererProps extends BoxProps {
+  isXs: boolean
+}
+
+export const NewSearchBarInputQueryRenderer: FC<NewSearchBarInputQueryRendererProps> = props => {
+  const { relayEnvironment, searchQuery = "" } = useContext(SystemContext)
+  const isMounted = useDidMount(typeof window !== "undefined")
```
dimatretyak, line 159: I think it will be enough just to check `typeof window !== "undefined"` (or even better to use [isServer](https://github.com/artsy/force/blob/25bfc0789dc42bad14afcaea14c6e8cec80a968a/src/Utils/device.ts#L1)), since with client-side rendering the condition will always be `true` and there is no need to use `useDidMount`. @damassi or am I wrong?
 
Something like this
```javascript
export const NewSearchBarInputQueryRenderer = props => {
  ...

  if (isServer) {
    return <StaticSearchContainer searchQuery={searchQuery} {...props} />
  }

  return (...)
}
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -0,0 +1,228 @@
...
+ */
+const StaticSearchContainer: FC<{ searchQuery: string } & BoxProps> = ({
+  searchQuery,
+  ...rest
+}) => {
+  const { t } = useTranslation()
+  return (
+    <>
+      <Box display={["block", "none"]} {...rest}>
+        <SearchInputContainer
```
dimatretyak, line 214: **question**: do we need to use prev `SearchInputContainer` component here?

src/Components/Search/NewSearch/NewSearchInputPills.tsx
```diff
@@ -157,6 +166,10 @@ export const NewSearchInputPillsContainer = () => {
     }
   }
 
+  const onPillClick = (pill: PillType) => {
+    setSelectedPill(pill)
+  }
+
   return (
     <Flex alignItems="center" bg="white">
       <Flex position="absolute" left={0} opacity={showPreviousChevron ? 1 : 0}>
```
dimatretyak, line 175: @rajsam003 we need to change the `pointer-events` or remove/unmount component (as suggested [here](https://github.com/artsy/force/pull/12276#discussion_r1172844949)), otherwise it may be impossible to click on the "Top" pill due to the fact that the left chevron button is displayed and it will handle the click event (since we only change the `opacity` prop of the component. But we can still click on this chevron button)

https://user-images.githubusercontent.com/3513494/234380859-54cefebe-a100-4a3a-be93-17f448ce4f14.mp4

dimatretyak (approved):

src/Components/Search/NewSearch/NewSearchInputPills.tsx
anandaroop: Would [`visibility: hidden`](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility) help here? I thought maybe so, but it seems Flex doesn't support that 🤔

anandaroop (approved):

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: you mean that I should create a new `NewSearchInputContainer` or that we don't need it at all?

src/Components/Search/NewSearch/NewSearchInputPills.tsx
nickskalkin: I fixed this problem using a z-index property, @dimatretyak let's pair on refactoring in! I tried to follow `{showNextChevron && ()}` approach, but stumbled upon some issues

nickskalkin:
Merging this PR, since it was previously approved. There are still some pending questions, I will resolve them in a separate PR

PR #12308, opened 2023-04-26
feat(FX-4655): search - switch between pills

nickskalkin:

This PR solves [FX-4655]

### Description

- [Demo app](https://new-search-bar.artsy.net/) (stale)

[FX-4655]: https://artsyproduct.atlassian.net/browse/FX-4655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Search/NewSearch/constants.ts
```diff
@@ -65,3 +67,19 @@ export const PILLS: PillType[] = [
   SHOW_PILL,
   GALLERY_PILL,
 ]
+
+interface SearchEntityMap {
+  [key: string]: SearchEntity
+}
```
dimatretyak, line 73: JFYI we can use [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) from TS. Something like this

```javascript
export const ELASTIC_PILL_KEY_TO_SEARCH_ENTITY: Record<string, SearchEntity> = {...}
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -66,6 +71,25 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({
...
   }
 
+  const handlePillClick = (pill: PillType) => {
+    setSelectedPill(pill)
+
+    relay.refetch(
+      {
+        hasTerm: true,
+        term: value,
+        entities: ELASTIC_PILL_KEY_TO_SEARCH_ENTITY?.[pill.key],
```
dimatretyak, line 81: i think we need to pass array of entities here

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -66,6 +71,25 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({
     )
   }
 
+  const handlePillClick = (pill: PillType) => {
+    setSelectedPill(pill)
+
+    relay.refetch(
```
dimatretyak, line 77: **suggestion**: I think we can combine all refetch calls into one (feel free to ignore if you don't agree with the suggestions). Like this

```javascript
const refetch = (term: string, entity: SearchEntity) => {
  relay.refetch(
    {
      hasTerm: true,
      term,
      entities: [entity],
    },
    null,
    error => {
      if (error) {
        logger.error(error)
        return
      }
    }
  )
}

const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
  const term = event.target.value
  const entity = ELASTIC_PILL_KEY_TO_SEARCH_ENTITY?.[selectedPill.key]
  
  setValue(term)
  refetch(term, entity)
}

const handlePillClick = (pill: PillType) => {
  const entity = ELASTIC_PILL_KEY_TO_SEARCH_ENTITY?.[pill.key]
  
  setSelectedPill(pill)
  refetch(value, entity)
}
```

or even better like this
```javascript
useEffect(() => {
  const entity = ELASTIC_PILL_KEY_TO_SEARCH_ENTITY?.[selectedPill.key]
  
  relay.refetch(
    {
      hasTerm: true,
      term: value,
      entities: [entity],
    },
    null,
    error => {
      if (error) {
        logger.error(error)
        return
      }
    }
  )
}, [value, selectedPill])

const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
  setValue(event.target.value)
}

const handlePillClick = (pill: PillType) => {
  setSelectedPill(pill)
}
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -161,16 +201,18 @@ export const NewSearchBarInputQueryRenderer: FC<NewSearchBarInputQueryRendererPr
...
             ...NewSearchBarInput_viewer
-              @arguments(term: $term, hasTerm: $hasTerm)
+              @arguments(term: $term, hasTerm: $hasTerm, entities: $entities)
           }
         }
       `}
       variables={{
         hasTerm: false,
         term: "",
+        entities: null,
```
dimatretyak, line 215: **suggestion:** always to pass entities (feel free to ignore suggestions). Something like this

```javascript
const DEFAULT_PILL = TOP_PILL
const DEFAULT_PILL_ENTITY = ELASTIC_PILL_KEY_TO_SEARCH_ENTITY[DEFAULT_PILL.key]

...

variables={{
  hasTerm: false,
  term: "",
  entities: [DEFAULT_PILL_ENTITY],
}}
```

dimatretyak (approved):

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -74,26 +188,28 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({
       value={value}
       onChange={handleChange}
       onClear={() => setValue("")}
```
dimatretyak, line 190: **suggestion**: create reusable function for it. Something like this

```javascript
const clear = () => {
  setValue("")
}

const handleSelect = (option: any) => {
  ...
  clear()
  ...
}

...

return (
  <AutocompleteInput
    ...
    onClear={clear}
    ...
  />
)
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -1,70 +1,184 @@
...
   isXs,
   relay,
   viewer,
 }) => {
+  const tracking = useTracking()
   const { t } = useTranslation()
-  const [value, setValue] = useState("")
+  const [value, setValue] = useState(getSearchTerm(window.location))
+  const [selectedPill, setSelectedPill] = useState<PillType>(TOP_PILL)
+  const [searchIncrement, setSearchIncrement] = useState(0)
```
dimatretyak, line 75: **suggestion**: what do you think about `searchCounter` or `fetchCounter`?

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -1,70 +1,184 @@
...
+    refetch(value, ELASTIC_PILL_KEY_TO_SEARCH_ENTITY?.[pill.key])
+  }
+
+  const handleRedirect = () => {
+    setValue("")
+  }
+
+  const handleSubmit = () => {
+    if (value) {
+      window.location.href = `/search?term=${encodeURIComponent(value)}`
```
dimatretyak, line 157: **suggestion**: use `router.push` here. Something like this

```javascript
const { router } = useRouter()

...

const handleSubmit = () => {
  if (value) {
    router.push(`/search?term=${encodeURIComponent(value)}`)
  }
}
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -1,70 +1,184 @@
...
   }
+  const throttledSearchRequest = useMemo(
+    () => throttle(refetch, SEARCH_THROTTLE_DELAY, { leading: true }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  )
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setValue(event.target.value)
+    throttledSearchRequest(event.target.value)
```
dimatretyak, line 143: **question**: should we also pass the selected pill entity in the second argument?

dimatretyak, line 143: we should not refetch data if the user has entered **less** than 2 characters (as [here](https://github.com/artsy/force/blob/f85fdad385bcfae7c7fdeb016a2f49e59fa42bad/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L187))

dimatretyak: we will also have to cancel throttled search request if the user **has cleared** the search value/query or entered less than 2 characters

https://user-images.githubusercontent.com/3513494/236203551-53e77ae2-a629-420a-9bfc-6821b66fb596.mp4

src/Components/Search/NewSearch/NewSearchBarFooter.tsx
```diff
@@ -1,35 +1,40 @@
...
 }) => {
+  const tracking = useTracking()
+
   const handleClick = (
-    event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
+    _event: React.MouseEvent<HTMLAnchorElement, MouseEvent>
   ) => {
-    event.preventDefault()
+    tracking.trackEvent({
+      action_type: DeprecatedSchema.ActionType.SelectedItemFromSearch,
```
damassi, line 26: For new components, can we avoid using items from the deprecated analytics schema?

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -1,70 +1,184 @@
...
-  })
+  )
 
-  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
-    setValue(event.target.value)
+  useUpdateEffect(() => {
+    tracking.trackEvent({
+      action_type:
+        options.length > 0
+          ? DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
```
damassi, line 103: Ditto (here and elsewhere) - if these events aren't actually deprecated, could y'all open a PR and copy them into the canonical (current) analytics schema?

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -1,70 +1,184 @@
...
+        option.typename === "Artist"
+          ? `${option.href}/works-for-sale`
+          : option.href,
+      item_number: option.item_number,
+      item_type: option.item_type,
+      query: value,
+    })
+
+    setValue("")
+    window.location.href = option.href
```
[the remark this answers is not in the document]
damassi: Yep, for the record, anytime we use window.location.whatever, we lose all routing history (along with our routing session caches), and its quite a hit to performance. 

Always lean on the router for any kind of navigation case except for cases where things are so complicated that you literally have to reboot the computer (so to speak) -- aka, hard jumps between pages. (The one example I can think of where this is necessary is logging in and out.)

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: There is no TOP_PILL search entity, that's why I pass null when TOP pill is selected: https://github.com/artsy/metaphysics/blob/17e599f15acc745d39a3b6cf664b5dedbc66c78a/_schemaV2.graphql#L14667

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: Not sure if we need it here, since we store pill in the state, which we update when switching between pills. That's why when refetching data, pill should be up to date

nickskalkin: Seems for me that it's better to use debouncing instead of throttling in order to avoid firing search requests for previously typed strings. I also decreased debounce interval from 500 to 250, this way user experience feels more responsive, but we also do not fire extra requests if user types very fast.

Not sure if we need to dispose previous refetch requests. I think user has to be like incredibly fast clearing the input field, no?

src/Components/Search/NewSearch/NewSearchBarFooter.tsx
nickskalkin: Thanks! I've created a ticket to re-visit analytic events. For context, this component will be hidden behind a feature flag and won't be enabled after merging this PR. We have a couple of other tickets to implement before we can enable this feature (analytic events included)

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: Thanks! Same as here: https://github.com/artsy/force/pull/12308#discussion_r1185907851

src/Components/Search/NewSearch/NewSearchBarInput.tsx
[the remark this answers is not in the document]
nickskalkin: Not sure if we can get rid of `AutocompleteInputOption` easily, we will have to make changes in palette, and I'm not sure if I like this change. For now I went with using `router.push()` instead of `window.location`, but happy to re-iterate later

src/Components/Search/NewSearch/NewSuggestionItem.tsx
```diff
@@ -0,0 +1,176 @@
...
+    </SuggestionItemLink>
+  )
+}
+
+interface DefaultSuggestionProps {
+  option: SuggionItemOptionProps
+  query: string
+}
+
+const DefaultSuggestion: FC<DefaultSuggestionProps> = ({ option, query }) => {
```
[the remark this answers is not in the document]
nickskalkin: Yeah, better to separate them, will do 👍

src/Components/Search/NewSearch/NewSearchBarInput.tsx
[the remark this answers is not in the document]
nickskalkin: But I think we will have to do it... Clicking on quick nav links doesn't work, I think because of what you mentioned @dimatretyak

src/Components/Search/NewSearch/NewSearchBarInput.tsx
dimatretyak: we will always run `refetch` with empty entities when the search query changes (since `entity` is not passed [here](https://github.com/artsy/force/blob/f85fdad385bcfae7c7fdeb016a2f49e59fa42bad/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L143) and an [empty array](https://github.com/artsy/force/blob/f85fdad385bcfae7c7fdeb016a2f49e59fa42bad/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L111) is used by default)

https://user-images.githubusercontent.com/3513494/236439057-96d95e78-0c83-4542-a7dc-339298915026.mp4

nickskalkin: Weird, I will take a look, thank you

dimatretyak: > it's better to use debouncing instead of throttling

maybe you're right. but it seems to me that when using trolling, the UX will be more responsive and friendly (since the user will see the "fresh"/"relevant" search results for the query more often. for example, like Amazon or Wikipedia)

src/Components/Search/NewSearch/SearchInputContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <LabeledInput
+      ref={ref}
+      width="100%"
+      height={50}
+      label={
+        <SearchButton
+          type="submit"
+          onClick={event => {
+            // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
+            ;(event.target as HTMLElement).parentElement.blur()
```
dimatretyak, line 39: use this code`;(event.target as HTMLElement)?.parentElement?.blur()` and remove `@ts-expect-error` comment

src/Components/Search/NewSearch/NewSuggestionItem.tsx
```diff
@@ -0,0 +1,173 @@
...
+}) => {
+  const handleClick = event => {
+    // Stopping the event from propagating to prevent SearchBar from navigation to the main suggestion item url.
+    event.stopPropagation()
+    event.preventDefault()
+    onClick?.(event)
+
+    // TODO: Use routerLink instead of manual navigation?
+    // FIXME: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
+    window.location.assign(to)
```
dimatretyak, line 154: **suggestion**: as far as I understand we can use `router.push` now without any problems

https://user-images.githubusercontent.com/3513494/236444119-2f858c6e-4aab-4d73-b4c5-c2f2b1b47359.mp4

dimatretyak:
I think we should clear the search query when going to another page (as [implemented](https://github.com/artsy/force/blob/c578504f29a91dc1d6da1c54fa2ecf86672b7cb8/src/Components/Search/SearchBar.tsx#L173) in the previous search input)

https://user-images.githubusercontent.com/3513494/236444488-c3f7fc12-05a7-4d12-b8f4-23c78c065475.mp4

src/Components/Search/NewSearch/NewSearchBarInput.tsx
dimatretyak: **suggestion**: I think in this case it is better to pass an empty array instead of `null` (as it is done [here](https://github.com/artsy/force/blob/c578504f29a91dc1d6da1c54fa2ecf86672b7cb8/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L106)) so that there is the same logic everywhere

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,213 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
-          href={`/search?term=${encodeURIComponent(value)}`}
-          isHighlighted={false}
-          label={value}
           query={value}
+          href={`/search?term=${encodeURIComponent(value)}`}
+          onRedirect={handleRedirect}
+          currentIndex={options.length}
         />
       }
+      dropdownMaxHeight={`calc(100vh - ${DESKTOP_NAV_BAR_TOP_TIER_HEIGHT}px)`}
```
dimatretyak, line 218: **suggestion**: we need to add bottom offset


for-nikita-4

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,213 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+      action_type: DeprecatedSchema.ActionType.FocusedOnAutosuggestInput,
+    })
+  }
+
   return (
     <AutocompleteInput
       placeholder={isXs ? t`navbar.searchArtsy` : t`navbar.searchBy`}
       spellCheck={false}
       options={value.length < 2 ? [] : formattedOptions}
       value={value}
```
dimatretyak, line 190: when we navigate to another page using the "See full results..." button, the entered search query is not displayed

### Previous implementation
https://user-images.githubusercontent.com/3513494/236447498-519bd0ba-e93c-42f9-a5d4-1b7621512398.mp4

### Current implementation
https://user-images.githubusercontent.com/3513494/236447529-b43dd68b-b6ac-4b15-86d4-5fff85b530ef.mp4

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,213 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
         if (error) {
           logger.error(error)
           return
         }
+
+        if (performanceStart && getENV("VOLLEY_ENDPOINT")) {
+          reportPerformanceMeasurement(performanceStart)
+        }
+
+        setFetchCounter(prevCounter => prevCounter + 1)
```
dimatretyak, line 126: **suggestion**: Add a comment with details of what `fetchCounter` is for

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,213 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+  )
+
+  const clearSearchInput = () => {
+    setValue("")
+  }
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setValue(event.target.value)
+
+    if (event.target.value.length > 1)
```
dimatretyak, line 144: I think there should be a check for the presence of **2 characters or more**

dimatretyak, line 144: **suggestion**: create a helper method so that there is one source of truth

```javascript
const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ ... }) => {
  ...
  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
    const newValue = event.target.value
    
    setValue(newValue)

    if (shouldStartSearching(newValue))
      debouncedSearchRequest(event.target.value)
  }

  ...DESKTOP_NAV_BAR_TOP_TIER_HEIGHT

  return (
    <AutocompleteInput
      ...
      options={shouldStartSearching(value) ? formattedOptions : {}}
      ...
    />
  )
}

const shouldStartSearching = (value: string) => {
  return value.length >= 2
}
```

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,213 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
       footer={
         <NewSearchBarFooter
-          display={value}
-          href={`/search?term=${encodeURIComponent(value)}`}
-          isHighlighted={false}
-          label={value}
           query={value}
+          href={`/search?term=${encodeURIComponent(value)}`}
+          onRedirect={handleRedirect}
+          currentIndex={options.length}
```
dimatretyak, line 215: **suggestion**: use `index` instead of `currentIndex` (feel free to ignore if you disagree)

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,234 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+import {
+  NewSuggestionItem,
+  SuggionItemOptionProps,
+} from "./SuggestionItem/NewSuggestionItem"
+import { debounce } from "lodash"
+import { useTracking } from "react-tracking"
+import * as DeprecatedSchema from "@artsy/cohesion/dist/DeprecatedSchema"
+import { getENV } from "Utils/getENV"
+// eslint-disable-next-line no-restricted-imports
+import request from "superagent"
```
dzucconi, line 26: Just use `fetch` This is a restricted import for a reason — we want to remove it completely.

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,234 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+  // Clear the search term once you navigate away from search results
+  useMemo(() => {
+    router.addNavigationListener(location => {
+      if (!location.pathname.startsWith("/search")) {
+        setValue("")
+      }
+
+      return true
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 109: Why disable? If ever we do this we need a clear comment describing the reason why.

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,234 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+
+  useUpdateEffect(() => {
+    tracking.trackEvent({
+      action_type:
+        options.length > 0
+          ? DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
+          : DeprecatedSchema.ActionType.SearchedAutosuggestWithoutResults,
+      query: value,
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 120: Ditto

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -10,92 +9,234 @@ import { NewSearchBarInput_viewer$data } from "__generated__/NewSearchBarInput_v
...
+        }
+
+        setFetchCounter(prevCounter => prevCounter + 1)
       }
     )
   }
 
+  const debouncedSearchRequest = useMemo(
+    () => debounce(refetch, SEARCH_DEBOUNCE_DELAY),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 151: So debouncing in React hooks is non-trivial. This is papering over a bug. I'd just import the `use-debounce` package

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -190,31 +343,3 @@ export const NewSearchBarInputQueryRenderer: FC<NewSearchBarInputQueryRendererPr
...
-
-      <Box display={["none", "block"]} {...rest}>
-        <SearchInputContainer
-          placeholder={searchQuery || t`navbar.searchBy`}
-          defaultValue={searchQuery}
-        />
-      </Box>
-    </>
-  )
-}
```
dzucconi, line 220: So there's definitely a content shift happening that needs to get fixed before this get merged. Not sure if this is related but beware.

src/Components/Search/NewSearch/SearchInputContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+const SearchButton = styled(Clickable)`
+  width: 18px;
+  height: 18px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: 50%;
+
+  &:focus {
+    background: ${themeGet("colors.purple100")};
```
dzucconi, line 17: Purple!?

src/Components/Search/NewSearch/SuggestionItem/QuickNavigationItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+    event.stopPropagation()
+    event.preventDefault()
+    onClick?.(event)
+
+    // TODO: Use routerLink instead of manual navigation?
+    // FIXME: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
+    window.location.assign(to)
+  }
+
+  return <Pill onClick={handleClick} mt={1} mr={1} {...rest} />
```
dzucconi, line 25: 🚫  external margins

src/Components/Search/NewSearch/SearchInputContainer.tsx
nickskalkin: Just copied it over from [here](https://github.com/artsy/force/blob/e1566ca9ff8e117d7c350ecb5900e8ef0bcbdd73/src/Components/Search/SearchInputContainer.tsx#L17)

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: Yeah it is fixed already, review app was staled a bit

src/Components/Search/NewSearch/SearchInputContainer.tsx
nickskalkin: removed

src/Components/Search/NewSearch/SuggestionItem/QuickNavigationItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+}) => {
+  const handleClick = event => {
+    // Stopping the event from propagating to prevent SearchBar from navigation to the main suggestion item url.
+    event.stopPropagation()
+    event.preventDefault()
+    onClick?.(event)
+
+    // TODO: Use routerLink instead of manual navigation?
+    // FIXME: Using `window.location.assign(to)` instead of `router.push(to)` to prevent a bug where the search bar won't hide anymore.
+    window.location.assign(to)
```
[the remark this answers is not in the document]
nickskalkin: 👍 Will create a separate ticket

src/Components/Search/NewSearch/SuggestionItem/QuickNavigationItem.tsx
nickskalkin: Hey, didn't get it, could you elaborate?

nickskalkin:
Going to merge it (even though there are some bugs, I will address them in separate PRs)

src/Components/Search/NewSearch/SuggestionItem/QuickNavigationItem.tsx
dzucconi: https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-external-margins

PR #12309, opened 2023-04-26
chore(CX-3649): add redirect, href

dariakoko:
Co-authored with @MounirDhahri 

This PR solves [CX-3649]

### Description

[CX-3649]: https://artsyproduct.atlassian.net/browse/CX-3649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -78,110 +75,116 @@ export const ArtistAuctionResultItem: React.FC<Props> = props => {
...
+        textDecoration="none"
+        style={{ pointerEvents: user ? "auto" : "none" }}
+      >
+        <GridColumns>
+          <Column span={[4, 2]}>
+            <ResponsiveBox
+              aspectWidth={1}
+              aspectHeight={1}
+              maxWidth={130}
+              bg="black10"
```
dariakoko, line 89: Changes in this file displayed super weirdly. Actually we just wrapped `<GridColumns />` in `<RouterLink />`

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -78,110 +75,116 @@ export const ArtistAuctionResultItem: React.FC<Props> = props => {
...
-              >
-                <NoArtIcon height={24} width={24} fill="black60" />
-              </Box>
-            )}
-          </ResponsiveBox>
-        </Column>
+      <RouterLink
+        to={`/auction-result/${internalID}`}
+        textDecoration="none"
+        style={{ pointerEvents: user ? "auto" : "none" }}
```
MrSltun, line 81: Great solution!

**Nitpick:**
But I tested it with keyboard navigation and managed to get to the screen without being logged in

Screen recording: <video src="https://user-images.githubusercontent.com/20655703/235910503-40209486-cc1d-46b5-8d0a-d88747837897.mov" /> 


My suggestion is either use [`tabIndex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) as -1 so the keyboard doesn't focus on it (affects keyboard accessibility) or have the data hidden in the Auction Result screen if the user is not logged in

Using tabIndex example
```tsx
...
tabIndex={user ? 0 : -1}
...
```

MrSltun (dismissed):
Great work!! ✨
I just left a small nitpicky comment :)

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
MounirDhahri: good catch @MrSltun and thanks for the suggestion.
@dzucconi do you have suggestions on how we can add this href without making it tappable? And if actually we don't have to do that. Thank you

src/Apps/Auctions/EndingSoonAuctions/EndingSoonAuctions.tsx
```diff
@@ -13,7 +13,10 @@ interface EndingSoonAuctionsProps {
 const EndingSoonAuctions: React.FC<EndingSoonAuctionsProps> = ({ viewer }) => {
   return (
     <>
-      <MetaTags title="Auction Lots for You Ending Soon" />
+      <MetaTags
+        title="Auction Lots for You Ending Soon"
```
dzucconi, line 17: Generally we append `| Artsy` to these.

MounirDhahri (approved):

PR #12318, opened 2023-04-26
fix: set auction artwork initial sort from querystring

anandaroop:

This PR solves [FX-4744]

### Description

We found that auction artwork grids were not respecting an incoming `sort` param in the URL querystring, and instead always sorting by lot number (`sort: sale_position`) —

```
https://www.artsy.net/auction/artsy-auction-street-art-2?sort=-prices
                                                         ↑↑↑↑↑↑↑↑↑↑↑↑ ignored
```

This is because the default value of `sale_position` for the `sort` param was being set later in the request lifecycle, and was thus clobbering the value coming in from the querystring.

This PR hoists that default out of the later location (`AuctionArtworkFilter.tsx`) up into an earlier location (`auctionRoutes.tsx`) so that it can be correctly overriden by the URL param.

The grid is now sorted correctly…

✓ when there is no `sort` param in the URL _(defaults to `sort: sale_position`)_
✓ when there is a `sort` param in the URL _(the specified `sort` is used during the server render)_
✓ when a sort is chosen from the sort options menu  _(the specified `sort` is used during the client refetch)_

[FX-4744]: https://artsyproduct.atlassian.net/browse/FX-4744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Auction/auctionRoutes.tsx
+++ b/src/Apps/Auction/auctionRoutes.tsx
@@ -72,20 +72,26 @@ export const auctionRoutes: AppRouteConfig[] = [
       }
     `,
     prepareVariables: (params, props) => {
-      const initialFilterState = getInitialFilterState(
+      const initialFilterStateFromUrl = getInitialFilterState(
         props.location?.query ?? {}
       )
 
-      return {
+      const userSpecificFilterState = getArtworkFilterInputArgs(
+        props.context.user
+      )
+
+      const variables = {
         slug: params.slug,
         input: {
-          ...initialFilterState,
-          ...getArtworkFilterInputArgs(props.context.user),
+          ...initialFilterStateFromUrl,
+          ...userSpecificFilterState,
           saleID: params.slug,
           // FIXME: Understand why this is needed to view lots in `the-artist-is-present-a-benefit-auction-for-ukraine` while logged out
           priceRange: "*-*",
         },
       }
+
+      return variables
     },
     children: [
       { path: "" },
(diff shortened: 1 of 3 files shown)


src/Apps/Auction/auctionRoutes.tsx
```diff
@@ -72,20 +72,31 @@ export const auctionRoutes: AppRouteConfig[] = [
...
+      )
+
+      const variables = {
         slug: params.slug,
         input: {
-          ...initialFilterState,
-          ...getArtworkFilterInputArgs(props.context.user),
+          ...auctionFilterDefaults,
+          ...initialFilterStateFromUrl,
+          ...userSpecificFilterState,
```
anandaroop, line 92: This now represents the correct order of preference for the filter params.

src/Apps/Auction/auctionRoutes.tsx
```diff
@@ -72,20 +72,31 @@ export const auctionRoutes: AppRouteConfig[] = [
       }
     `,
     prepareVariables: (params, props) => {
-      const initialFilterState = getInitialFilterState(
+      const auctionFilterDefaults = {
+        sort: "sale_position",
+      }
```
dimatretyak, line 77: nice

dimatretyak:
noticed strange behavior when the user is **not** logged in. trying to understand the cause of the problem


https://user-images.githubusercontent.com/3513494/235003683-e10a81af-dab9-435c-b464-7a5aed37f445.mp4

dimatretyak:
ok, I checked on the [review app](https://auction-sort.artsy.net/) and everything works without problems. It looks like I'm having some problems with the local cache

https://user-images.githubusercontent.com/3513494/235006338-159bd9f6-4cc6-4d03-9c46-33c444b2ba0e.mp4

dimatretyak (approved):

anandaroop:
Thanks for verifying with the review app!

PR #12319, opened 2023-04-27
feat(NX-3525): render the salesforce chat bubble only on desktop

lean8086:

This PR solves [NX-3525]

### Description

- We do not want to render the chat bubble in mobile

cc @artsy/negotiate-devs

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkZendesk.tsx
+++ b/src/Apps/Artwork/Components/ArtworkZendesk.tsx
@@ -5,6 +5,7 @@ import { exceedsChatSupportThreshold } from "Utils/exceedsChatSupportThreshold"
 import { ArtworkZendesk_artwork$data } from "__generated__/ArtworkZendesk_artwork.graphql"
 import { getENV } from "Utils/getENV"
 import { SalesforceWrapper } from "Components/SalesforceWrapper"
+import { Media } from "Utils/Responsive"
 
 interface ArtworkZendeskProps {
   artwork: ArtworkZendesk_artwork$data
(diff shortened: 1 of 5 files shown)


araujobarret (approved):

lean8086:
One sec. I'll also hide the button in CSS, in case the component is loaded on desktop and then the window is resized.

src/Components/SalesforceWrapper.tsx
```diff
@@ -1,21 +1,27 @@
...
+        font-family: "ll-unica77", "Helvetica Neue", Helvetica, Arial, sans-serif; /* fonts.sans */
+        box-shadow: none;
+      }
+    
+      .embeddedServiceHelpButton .helpButton .uiButton:focus {
+        background: #1023D7; /* colors.blue100 */
+      }
+      
+      /* Make sure it doesn't render when loaded on desktop and then the window is resized */
+      @media (max-width: 767px) { /* mediaQueries.xs */
```
araujobarret, line 20: is it really necessary? Doesn't hurt having it but we don't need to cover a lot of edge cases from resizing Deskop<->Mobile, usually the ones doing it are us, not the final users.

araujobarret (approved):

src/Components/SalesforceWrapper.tsx
lidimayra: I'd say better safe than sorry on this one! Given the screenshots that partners share with us, very often they do not work with the window full size (and the same goes for some Artsy members as well, even non-engineers)

PR #12330, opened 2023-04-28
fix(NX-3530): hide SF chat after unmounting. Same as the zendesk component

lean8086:

This PR solves [NX-3530]

### Description

SF persisted after navigating to a different page. The component needs to be hidden when unmounting.

cc @artsy/negotiate-devs 

[NX-3530]: https://artsyproduct.atlassian.net/browse/NX-3530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/SalesforceWrapper.tsx
+++ b/src/Components/SalesforceWrapper.tsx
@@ -1,6 +1,7 @@
 import { getENV } from "Utils/getENV"
 import { useLoadScript } from "Utils/Hooks/useLoadScript"
 import { useAppendStylesheet } from "Utils/Hooks/useAppendStylesheet"
+import { useEffect } from "react"
 
 export const SalesforceWrapper: React.FC = () => {
   useAppendStylesheet({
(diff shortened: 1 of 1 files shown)


araujobarret (approved):
Looks good! I wonder if they start a conversation with some assistant and then move to another page if the assistant responds if we display the chat bubble again, but I don't consider this(if it happens) a blocker for this PR.

lean8086:
> Looks good! I wonder if they start a conversation with some assistant and then move to another page if the assistant responds if we display the chat bubble again, but I don't consider this(if it happens) a blocker for this PR.

That's why I'm also making sure the component is shown again [here](https://github.com/artsy/force/pull/12330/files#diff-05a58e6cba75db52b8d77d621bd3d519cdaa0b20356ef3f8fc859b435ac6cc3eR61). Good thing to keep an eye on, though.

PR #12333, opened 2023-04-28
fix(NX-3530): window is undefined on ssr

lean8086:

This PR solves [NX-3530]

### Description

Integration and smoke tests failing for `window` being undefined on the server.

cc @artsy/negotiate-devs 

[NX-3530]: https://artsyproduct.atlassian.net/browse/NX-3530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/SalesforceWrapper.tsx
+++ b/src/Components/SalesforceWrapper.tsx
@@ -58,7 +58,7 @@ export const SalesforceWrapper: React.FC = () => {
     },
   })
 
-  window.embedded_svc?.showHelpButton?.()
+  window?.embedded_svc?.showHelpButton?.()
 
   useEffect(() => {
     return () => {


lidimayra (approved):

jo-rs:
This isn't an issue on other `window` instances?

lean8086:
> This isn't an issue on other `window` instances of this file?

The problem appeared after introducing the show/hide functionality. The one on the `useEffect` will execute on the client, so the other one should be the only one with an issue. Thoughts?

src/Components/SalesforceWrapper.tsx
```diff
@@ -58,7 +58,7 @@ export const SalesforceWrapper: React.FC = () => {
     },
   })
 
-  window.embedded_svc?.showHelpButton?.()
+  window?.embedded_svc?.showHelpButton?.()
```
damassi, line 61: Saw this moved to the effect where its guaranteed to be defined, but note that the way to check this is
```
if (typeof window !== "undefined")
```
existential operator will still error out

PR #12343, opened 2023-05-02
Revert "feat(PRO-107): add tracking to settings IDV clicks"

iskounen:
Reverts artsy/force#12331

I tested this change on staging, and expected to see an event being fired in the console. Instead, I found the following error message, and thought it merited some more investigation:

Screenshot 2023-05-02 at 2 21 04 PM


iskounen:
It turned out that my ad blocker was preventing events from being sent, and the above error corresponds to a "too many requests" error from Jumio.

PR #12344, opened 2023-05-02
Revert "Revert "feat(PRO-107): add tracking to settings IDV clicks""

iskounen:
Reverts artsy/force#12343


damassi:
@iskounen - mind updating the PR description with the reason why we need to revert? Its always helpful to leave historical context so folks don't end up unintentionally repeating work (happens often with reverts)

mdole (approved):

PR #12346, opened 2023-05-03
Chore (PX-5561) - Add shipping disclaimer copy for Artsy Shipping Offers (web)

JanaeHijaz:
The type of this PR is: Chore

This PR solves [PX-5561]

### Description

As Artsy, 

I want to warn collectors that price might change for shipping when making an offer to an Artsy Shipping enabled artwork, so that they understand why shipping prices change after counteroffers are sent. 

This copy is updated in initial offer, accepting counter offer, and viewing accepted offer stages in Make Offer flow. 

Screenshot 2023-05-16 at 1 29 25 PM

Screenshot 2023-05-16 at 1 29 54 PM

[PX-5561]: https://artsyproduct.atlassian.net/browse/PX-5561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -220,8 +221,13 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
       />
       <Spacer y={2} />
+      {offer?.shippingTotal ? (
+        <Text variant="sm" color="black60">
+          *Estimate Only. Price may vary once offer is finalized.
+        </Text>
+      ) : null}
       <Text variant="sm" color="black60">
-        *Additional duties and taxes{" "}
+        † Additional duties and taxes{" "}
```
JanaeHijaz, line 230: Can't use `<sup>` here, and from what I've seen I'd have to use `lineHeight` instead but not working...

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -100,11 +100,11 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
       if (selectedShippingQuote) {
         return `${
           shippingQuoteDisplayNames[selectedShippingQuote.typeName]
-        } delivery`
+        } delivery*`
```
jpotts244, line 103: I'm wondering if this `*` is going to show for all orders and not just make offer ones?


As Im going through the flow I see the `*` here for buy now orders which Im not sure we want since Shipping is not really negotiable here and there is no footnote at the bottom
Screen Shot 2023-05-18 at 4 38 40 PM

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -100,11 +100,11 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
       }
     }
 
-    return "Shipping"
+    return "Shipping*"
```
jpotts244, line 107: Same comment as above, I dont think we want to show the `*` for a footnote on buy now orders correct? Buy nows are non-negotiable and shouldn't be expected to change

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
jpotts244: Palette has a `<Sup>` component that should work for us here
https://palette.artsy.net/elements/sup/#gatsby-focus-wrapper

If we do something like:
```
<Sup>†</Sup>Additional duties and taxes{" "}
```
^ it should evaluate as Superscript

jpotts244 (commented):
A few things here, 
- I think we want to not show these new elements across the board and only for Artsy Shipping Offers
- Let's wrap the daggers in Palette's `<Sup>` tags so we can have the placement handled automatically

jpotts244 (changes requested):
A few things here:

I think we want to not show these new elements across the board and only for Artsy Shipping Offers as those are the only cases where shipping is subject to change

Let's wrap the daggers in Palette's <Sup> tags so we can have the placement handled automatically

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -220,8 +231,16 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
       />
       <Spacer y={2} />
+      {offer?.shippingTotalCents &&
+      order.requestedFulfillment?.__typename === "CommerceShipArta" &&
+      order.mode === "OFFER" &&
+      order.displayState !== "PROCESSING" ? (
+        <Text variant="sm" color="black60">
+          *Estimate Only. Price may vary once offer is finalized.
+        </Text>
+      ) : null}
```
AlicanAkyuz, line 241: one useful syntax for this kind of cases where we want to render X if Y evaluates to true and don't render anything if it's false, is to treat JSX like a condition, like so: 
`{order.mode === "OFFER" && order.displayState !== "PROCESSING" && .... && your JSX}`

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -206,9 +214,12 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
         value={shippingDisplayAmount()}
         data-test="shippingDisplayAmount"
       />
-
       <Entry
-        label="Tax*"
+        label={
+          <>
+            Tax <Sup>†</Sup>
```
jpotts244, line 220: I think we still dont want to update to the dagger on regular old orders (like pickup of partner flat fee shipping) So we might need a few more conditionals here

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -220,8 +231,16 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
+      order.requestedFulfillment?.__typename === "CommerceShipArta" &&
+      order.mode === "OFFER" &&
+      order.displayState !== "PROCESSING" ? (
+        <Text variant="sm" color="black60">
+          *Estimate Only. Price may vary once offer is finalized.
+        </Text>
+      ) : null}
       <Text variant="sm" color="black60">
-        *Additional duties and taxes{" "}
+        <Sup>†</Sup> Additional duties and taxes{" "}
```
jpotts244, line 243: Same comment as above, these are shared across all order types so we probably dont want to show the cross unless there is a "shipping" footnote too

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
jpotts244: Im going to push some commits here!

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -162,6 +169,43 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
+      offer?.shippingTotalCents &&
+      order.requestedFulfillment?.__typename === "CommerceShipArta" &&
+      offerStillInNegotiation()
+    )
+  }
+
+  const offerStillInNegotiation = () => {
+    return (
+      order.mode === "OFFER" &&
+      ["PENDING", "SUBMITTED"].includes(order.displayState)
```
jpotts244, line 205: DisplayState check needed to not show this message on the `Details` view in conversions after everything has been approved

This screen: 
Screen Shot 2023-05-25 at 5 32 11 PM

jpotts244: This screen:
Screen Shot 2023-05-25 at 5 47 48 PM

jpotts244 (dismissed):

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
```diff
@@ -88,19 +88,26 @@ export const TransactionDetailsSummaryItem: FC<TransactionDetailsSummaryItemProp
...
+  const getShippingLabel = shippingQuoteType => {
+    const artsyShippingLabel =
+      shippingQuoteDisplayNames[shippingQuoteType] + " delivery"
+    const suffix = offerShippingCostSubjectToChange() ? "*" : ""
+    return artsyShippingLabel + suffix
+  }
+
   const shippingDisplayLabel = shippingNotCalculated => {
     if (shippingNotCalculated()) {
       return "Shipping**"
```
leamotta, line 100: wondering if we could address this ** here too? it would be something like 
```suggestion
      return "Shipping*"
```
here and in the message + a check for `shippingNotCalculate() || offerShippingCostSubjectToChange()` in the tax logic to show the dagger.
WDYT?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -790,8 +790,9 @@ export const ShippingRoute: FC<ShippingProps> = props => {
             <Collapse open={showArtsyShipping}>
               <Text variant="sm">Artsy shipping options</Text>
               <Text variant="xs" mb="1" color="black60">
-                All options are eligible for Artsy’s Buyer Protection policy,
-                which protects against damage and loss.
+                Please note that these are estimates and may change once offer
```
leamotta, line 793: should we check that the order is in offer mode here?

jpotts244: Great catch there 🦅

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
jpotts244: On it!

src/Apps/Order/Routes/Shipping/index.tsx
jpotts244: Added conditional logic for this help text

Buy Now experience:
Screen Shot 2023-05-26 at 8 58 47 AM

Make Offer experience:

Screen Shot 2023-05-26 at 8 57 38 AM

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
jpotts244: Adjusted -- Keeps the pattern going if there is ever 2 footnotes to use `*` and then `†`

Ex with shipping not calculated flow:
Screen Shot 2023-05-26 at 9 35 30 AM

src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx
jpotts244: Thanks! Refactoring made

PR #12354, opened 2023-05-04
chore: replaces icons

dzucconi:
Keep on keepin' on.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsWatchLotButton.tsx
+++ b/src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsWatchLotButton.tsx
@@ -1,4 +1,4 @@
-import { BellFillIcon, BellIcon, Popover, THEME } from "@artsy/palette"
+import { Popover, THEME } from "@artsy/palette"
 import { ArtworkAuctionRegistrationPanelFragmentContainer } from "Apps/Artwork/Components/ArtworkImageBrowser/ArtworkAuctionRegistrationPanel"
 import { UtilButton } from "./UtilButton"
 import { FC, useCallback, useEffect, useState } from "react"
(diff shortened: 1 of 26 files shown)


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsWatchLotButton.tsx
```diff
@@ -98,11 +100,6 @@ const ArtworkActionsWatchLotButton: FC<ArtworkActionsWatchLotButtonProps> = ({
   )
 }
 
-const UnfilledIcon = () => <BellIcon title="Unwatch lot icon" />
-const FilledIcon = () => (
-  <BellFillIcon title="Unwatch lot icon" fill="blue100" />
```
dzucconi, line 103: This is super odd to me: it appears we add these titles for specs? But they are legitimate titles?

Firstly: [just don't use them](https://www.a11yproject.com/posts/title-attributes/). If you have an icon only button use an `aria-label` on the clickable node. And never name anything 'Foo icon'.

src/Apps/Conversation/Components/UnreadMessagesToast.tsx
```diff
@@ -64,10 +63,18 @@ export const UnreadMessagesToast: React.FC<UnreadMessagesToastProps> = ({
...
   }, [
     hasScrolled,
+    // FIXME:
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     conversation?.lastMessageID,
+    // FIXME:
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     conversation?.fromLastViewedMessageID,
+    // FIXME:
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 77: 😭

PR #12359, opened 2023-05-06
add data- attributes to address form elements for tests

iskounen:
The type of this PR is: Chore

### Description

I would like to write an Integrity test to place an order that is shipped or billed to a country other than the United States, but there is no ID available on the address form's country select component to make that easy.

~I made the ID prop optional so that different forms can give it an ID that is consistent with other form elements (ie. `AddressForm_name`, `AddressForm_country`, etc.). I'd love to hear about whether that's a weird thing to do in React or not.~

I added a `data-test` attribute to the country selection element so that it can be easily selected. I also added `data-test` attributes to the rest of the elements on this form to be able to standardize this behavior.

### DIFF
--- a/src/Components/AddressForm.tsx
+++ b/src/Components/AddressForm.tsx
@@ -136,6 +136,7 @@ export const AddressForm: React.FC<AddressFormProps> = ({
           </Text>
           <CountrySelect
             tabIndex={tabIndex}
+            id="AddressForm_country"
             selected={
               lockCountryToOrigin || (lockCountriesToEU && !address.country)
                 ? shippingCountry
--- a/src/Components/AddressForm.tsx
+++ b/src/Components/AddressForm.tsx
@@ -136,7 +136,6 @@ export const AddressForm: React.FC<AddressFormProps> = ({
           </Text>
           <CountrySelect
             tabIndex={tabIndex}
-            id="AddressForm_country"
             selected={
               lockCountryToOrigin || (lockCountriesToEU && !address.country)
                 ? shippingCountry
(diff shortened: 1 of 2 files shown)


src/Components/AddressForm.tsx
```diff
@@ -136,6 +136,7 @@ export const AddressForm: React.FC<AddressFormProps> = ({
           </Text>
           <CountrySelect
             tabIndex={tabIndex}
+            id="AddressForm_country"
```
damassi, line 139: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

damassi: https://github.com/search?q=repo%3Aartsy%2Fforce%20data-test&type=code

AlicanAkyuz:
@iskounen I don't think it's weird to pass a test id to Select in this case. 

to make the prop more clear in terms of its purpose, I would generally rename it to something more explicit than `id`. but in this case, since rest of the props are [spread to Select](https://github.com/artsy/force/blob/db2abb721af904a14ade7bdff8f185ea30a19ced/src/Components/CountrySelect.tsx#L23), I think you could remove the new prop completely and just pass `data-test="AddressForm_country"` to `CountrySelect`.

and access it with `...find([data-test="AddressForm_country"])` in the tests.

AlicanAkyuz (dismissed):

iskounen:
@AlicanAkyuz and @damassi thanks a lot for your advice. Using a data attribute feels nicer, and taking advantage of prop spreading reduces the amount of noise.

I added `data-test` props to the other elements in this form so that I can standardize how Integrity access them.

PR #12361, opened 2023-05-08
fix: Hide install app on meet your new art advisor LP campaign

olerichter00:
Addresses [CX-3689]

## Description

Hide install app on meet your new art advisor LP campaign.

[CX-3689]: https://artsyproduct.atlassian.net/browse/CX-3689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Marketing/Routes/MarketingMeetArtAdvisorRoute.tsx
+++ b/src/Apps/Marketing/Routes/MarketingMeetArtAdvisorRoute.tsx
@@ -81,7 +81,7 @@ export const MarketingMeetArtAdvisorRoute: FC = () => {
           ]}
         />
 
-        {!!isEigen && (
+        {!isEigen && (
           <FullBleed bg={["black5", "transparent"]} py={[4, 0]} px={[2, 4]}>
             <GridColumns gridRowGap={4}>
               <Column span={12} textAlign="center">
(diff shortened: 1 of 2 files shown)


olerichter00:
> Looks great! But I think the solution is in the wrong file 😅
I think this is the right file: [MarketingMeetArtAdvisorRoute](https://github.com/artsy/force/blob/main/src/Apps/Marketing/Routes/MarketingMeetArtAdvisorRoute.tsx)

Monday 🤦 

Now it should be the correct page.

src/Apps/Marketing/Routes/MarketingMeetArtAdvisorRoute.tsx
```diff
@@ -78,26 +81,28 @@ export const MarketingMeetArtAdvisorRoute: FC = () => {
           ]}
         />
 
-        <FullBleed bg={["black5", "transparent"]} py={[4, 0]} px={[2, 4]}>
-          <GridColumns gridRowGap={4}>
-            <Column span={12} textAlign="center">
-              <Text variant="xl">Meet your new art advisor. It’s Artsy.</Text>
-            </Column>
+        {!!isEigen && (
```
MrSltun, line 84: We want it to show up if the client is not Eigen, right? 👀 
```suggestion
        {!isEigen && (
```

olerichter00: Exactly! I've committed your fix.

MrSltun (approved):

PR #12366, opened 2023-05-09
chore(icons): replaces remaining icons

dzucconi:
This will finally allow us to merge https://github.com/artsy/palette/pull/1283

@lilyfromseattle not to worry you — the migration in Volt won't be nearly as involved; I'll see if I can just take a quick pass at that later today.


src/Apps/BuyerGuarantee/Routes/BuyerGuaranteeIndex.tsx
```diff
@@ -308,7 +307,15 @@ export const BuyerGuaranteeIndex: FC = () => {
             millions of global businesses.
           </Text>
 
-          <PoweredByStripeIcon width={150} height={34} mt={2} />
+          <Spacer y={2} />
+
+          <img
+            src="https://files.artsy.net/images/PoweredByStripe_1.svg"
```
dzucconi, line 313: SVGs that don't belong in the system just get uploaded on tools.artsy.net and included via `img` tag.

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/Highlights.tsx
```diff
@@ -1,52 +1,47 @@
...
-        <Media greaterThan="xs">
-          <DollarCircleIcon height={50} width={50} />
-        </Media>
-        <Media at="xs">
-          <DollarCircleIcon height={30} width={30} />
-        </Media>
-      </>
+      <Box width={[30, 50]} height={[30, 50]}>
+        <img
+          src="https://files.artsy.net/images/DollarCircleIcon_1.svg"
```
dzucconi, line 9: These ones are really out of line tonally. Curious about their origin story.

Also, evergreen tweet: you generally don't need to be using `Media`

PR #12378, opened 2023-05-10
fix(PRO-130): correct routing after in_review offer

mdole:

This PR solves PRO-130

### Description

Turns out we were routing based on the _pending_ offer and not the response from the mutation.

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Review.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Review.jest.tsx
@@ -676,16 +676,16 @@ describe("Review", () => {
   })
 
   describe("in-review offers", () => {
-    const OfferOrderInReview = {
+    const OfferOrderToSubmit = {
       ...OfferOrderWithShippingDetails,
-      state: "IN_REVIEW",
+      state: "PENDING",
       internalID: "offer-order-id",
       impulseConversationId: null,
     }
 
     describe("from an artwork page", () => {
       const OfferOrderInReviewFromArtworkPage = {
-        ...OfferOrderInReview,
+        ...OfferOrderToSubmit,
         source: "artwork_page",
       }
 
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/__tests__/Review.jest.tsx
```diff
@@ -676,16 +676,16 @@ describe("Review", () => {
   })
 
   describe("in-review offers", () => {
-    const OfferOrderInReview = {
+    const OfferOrderToSubmit = {
       ...OfferOrderWithShippingDetails,
-      state: "IN_REVIEW",
+      state: "PENDING",
```
iskounen, line 681: **question:** Why did you change this fixture's state from `IN_REVIEW` to `PENDING`?

iskounen (commented):
Nice catch! I just have a question about the changes in the spec file.

src/Apps/Order/Routes/__tests__/Review.jest.tsx
mdole: Good q! Because it turns out I had misunderstood what this was mocking, which was why these tests were passing despite the behavior in production being incorrect.

Basically, this routing has access to two different order objects. One is `order`, [which is](https://github.com/artsy/force/blob/3a2bacc1f4855b5ff4e0c50307dfb739342b2b63/src/Apps/Order/Routes/Review/index.tsx#L124) the `pending` order that is being submitted at this time. The second is `orderOrError`, [which is](https://github.com/artsy/force/blob/3a2bacc1f4855b5ff4e0c50307dfb739342b2b63/src/Apps/Order/Routes/Review/index.tsx#L79) the response back from the submission mutation.

The reason this was broken in the first place was because we were performing our check on `order` instead of `orderOrError`, so we were always checking if the `pending` order was `in_review`.

This spec change corrects the same error for the spec. This is the pending order being submitted, while the result of the mutation is [what we pass](https://github.com/artsy/force/blob/3a2bacc1f4855b5ff4e0c50307dfb739342b2b63/src/Apps/Order/Routes/__tests__/Review.jest.tsx#L739) to `mockResolvedValue`

iskounen: We spend most of our time working in Exchange, where we can assume that an `order` object's state is always up-to-date, but in Force that data is detached from the database. Now it makes sense why we would have made this mistake. Thanks!

PR #12395, opened 2023-05-12
fix: Update SWA specialist key to match key in eigen

olerichter00:
## Description

This PR updates the SWA specialist key to match the key in [Eigen](https://github.com/artsy/eigen/blob/main/src/app/Scenes/SellWithArtsy/utils/useSWALandingPageData.tsx#L35). There is a typo in the key but it will be difficult to fix because Eigen fetches the data from S3 and old versions expect the key with the typo :/


MounirDhahri (approved):
👍

PR #12405, opened 2023-05-15
[WIP] feat: artist page auction tab empty

kajatiger:

This PR solves [GROW-1677]

### Description

Part of restructuring and polishing our new artist page are the tabs on the page. This PR adds what the user will see when visiting an artist page where the artist has zero auction results and still visits the auction results tab. 
Parts of the work are:

- [ ] adding a condition for the empty page
- [ ] adding a new empty page container
- [ ] adding the text 
- [ ] styling the text
- [ ] adding the button that redirects to the price database
- [ ] styling the button

figma design:

Bildschirmfoto 2023-05-11 um 11 27 04

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -46,6 +46,7 @@ import { KeywordFilter } from "./Components/KeywordFilter"
 import { MarketStatsQueryRenderer } from "./Components/MarketStats"
 import { SortSelect } from "./Components/SortSelect"
 import { TableSidebar } from "./Components/TableSidebar"
+import { EmptyArtistAuctionResults } from "./Components/EmptyArtistAuctionResults"
 
 const logger = createLogger("ArtistAuctionResults.tsx")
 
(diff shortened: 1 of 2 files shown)


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -214,141 +215,143 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
       0
     )
   }
+  if (results.length == 0) {
+    return <EmptyArtistAuctionResults></EmptyArtistAuctionResults>
+  } else {
+    return (
```
kajatiger, line 221: From this line down there are no changes except for the indentation (because of wrapping the view in the else block).

dzucconi: You don't need the `else`, in this case.

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
```diff
@@ -0,0 +1,32 @@
+import { Flex, Text, Button, GridColumns, Column } from "@artsy/palette"
+import { RouterLink } from "System/Router/RouterLink"
+
+export const EmptyArtistAuctionResults: React.FC = () => {
+  return (
+    <Flex>
```
dzucconi, line 6: This wrapping flex can be removed.

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <GridColumns width="100%" gridRowGap={2}>
+        <Column span={6} start={4}>
+          <Text variant="md" textAlign="center">
```
dzucconi, line 9: Can you drop a link to this Figma?

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
```diff
@@ -0,0 +1,32 @@
...
+        </Column>
+        <Column span={6} start={4}>
+          <Text variant="md" color="black60" textAlign="center">
+            The page will be updated once auction results are added. In the
+            meantime, check out the millions of auction results and art market
+            data.
+          </Text>
+        </Column>
+        <Column span={12} alignItems="center">
+          <Flex width="100%" justifyContent="center">
```
dzucconi, line 21: You can just `display=flex` the column, though I think you might just be looking for `mx="auto"` on the button.

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <RouterLink to="/price-database">
+              <Button variant="secondaryNeutral">
```
dzucconi, line 23: We would want to use `as={RouterLink} to="/price-database"` on the `Button` instead of wrapping it; this currently creates two tab-stops; one of which doesn't do anything. It will complain about a type error: it's safe to `// @ts-ignore` it

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -214,141 +215,143 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
       0
     )
   }
+  if (results.length == 0) {
+    return <EmptyArtistAuctionResults></EmptyArtistAuctionResults>
```
dzucconi, line 219: Self closing tag: `<EmptyArtistAuctionResults />`.

I would also change the name to stay prefixed by the parent: `ArtistAuctionResultsEmptyState`

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const EmptyArtistAuctionResults: React.FC = () => {
+  return (
+    <Flex>
+      <GridColumns width="100%" gridRowGap={2}>
+        <Column span={6} start={4}>
+          <Text variant="md" textAlign="center">
+            There are no auction results for this artist at this time.
+          </Text>
+        </Column>
+        <Column span={6} start={4}>
```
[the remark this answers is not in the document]
kajatiger: I tried with the break, but actually when I looked back at the figma it looks like there is no break at all. So I will just have the two text properties right underneath each other.

src/Apps/Artist/Routes/AuctionResults/Components/EmptyArtistAuctionResults.tsx
kajatiger: I don't quite understand this. Does the `display=flex` in the column mean, I can get rid of the wrapping `Flex`? And what would the `mx=auto` do on the button? I will try it out a bit but I don't really know what I am doing.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
kajatiger: Can I leave it in for readability reasons or is this looking weird for the JS people?

dzucconi (approved):
😗👌

PR #12406, opened 2023-05-15
fix(alerts): fixes children type

dzucconi:
Closes [FX-4771](https://artsyproduct.atlassian.net/browse/FX-4771)

I had the type wrong for children which was causing the render props function to not be called.

[FX-4771]: https://artsyproduct.atlassian.net/browse/FX-4771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dblandin (dismissed):
Thank you!! 💯

dzucconi:
OK, I fixed the immediate bug but am seeing we actually don't render the alert button when logged in until that counts query is loaded. All this indirection with the higher order component is due to the query hook not updating cached values. I'm going to see if I can refactor a bit to get the button rendering consistently during loading before merging.

dzucconi:
OK, fixed the underlying bug by converting that weird counts wrapper component into an actual HOC (shudder). But having a lot of difficulty mocking for some reason. Will pick this back up in the morning.

dzucconi:
OK, there we go.

PR #12407, opened 2023-05-15
feat(TX-1168): update add new address button + collector profile cleanup

rquartararo:
The type of this PR is: **Feature/Refactor**

This PR solves [TX-1168]

### Description

This PR changes the "add a new address" button on the shipping page to a clickable text button from a standard button. 

I also went ahead and removed the `inCollectorProfile` logic and components from `SavedAddresses.tsx` as it is no longer used. 

Screenshot 2023-05-15 at 4 12 02 PM

[TX-1168]: https://artsyproduct.atlassian.net/browse/TX-1168?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/SavedAddresses.tsx
+++ b/src/Apps/Order/Components/SavedAddresses.tsx
@@ -1,20 +1,7 @@
-import {
-  Box,
-  Button,
-  RadioGroup,
-  BorderedRadio,
-  Spacer,
-  Flex,
-  Text,
-  Separator,
-  BorderBox,
-  Join,
-  Clickable,
-} from "@artsy/palette"
+import { Button, RadioGroup, BorderedRadio, Spacer } from "@artsy/palette"
 import { useEffect, useState } from "react"
 import * as React from "react"
 import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
-import styled from "styled-components"
 import { SavedAddresses_me$data } from "__generated__/SavedAddresses_me.graphql"
 import { AddressModal, ModalDetails } from "Apps/Order/Components/AddressModal"
 import { CommitMutation } from "Apps/Order/Utils/commitMutation"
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Components/__tests__/SavedAddresses.jest.tsx
```diff
@@ -61,7 +61,7 @@ describe("Saved Addresses", () => {
           addressConnection: mockAddressConnection,
         }),
       })
-      const button = wrapper.find("Button[data-test='shippingButton']")
+      const button = wrapper.find("[data-test='shippingButton']").first()
```
JanaeHijaz, line 64: Curious about why `.first()` needs to be added here? I understand vaguely that .first() reduces all of the matched nodes and then wraps only the first node. Is this needed when checking a clickable link vs. a button? @rquartararo

rquartararo: Yes, so there are multiple ways to go about this. The `data-test='shippingButton'` selector has two nodes because one of them is the `AddAddressButton` component, and the other is the button that the component renders. You can console.log the wrapper content using `wrapper.debug()` to see this more clearly. I could have also done `"SavedAddresses__AddAddressButton[data-test='shippingButton']"` to specifically select the component without using `first()`, as we did with the previous `Button` component. Not entirely sure which way is preferred as I've seen this done both ways in Force.

rquartararo: One benefit of using `first()` in this case is that if we ever change the component name, we won't have to update the test again.

src/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -168,112 +155,23 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
     })
   }
 
-  const collectorProfileAddressItems = addressList.map((address, index) => {
```
AlicanAkyuz, line 171: 🙏

AlicanAkyuz (approved):
LGTM! 

and big thanks for removing those unused stuff 🥇

PR #12410, opened 2023-05-16
feat(FX-4660): mWeb new search overlay - results list

nickskalkin:

This PR solves [FX-4660]

### Description

- [Review app](https://mweb-search.artsy.net/)

### Demo

https://github.com/artsy/force/assets/3934579/78793c16-5e61-403a-964f-509c83a1c45d

[FX-4660]: https://artsyproduct.atlassian.net/browse/FX-4660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -13,13 +12,17 @@ import createLogger from "Utils/logger"
...
-import { useTracking } from "react-tracking"
-import * as DeprecatedSchema from "@artsy/cohesion/dist/DeprecatedSchema"
+import { useDebounce, useDebouncedValue } from "Utils/Hooks/useDebounce"
 import { MobileSearchBar_viewer$data } from "__generated__/MobileSearchBar_viewer.graphql"
+import { OverlayBase } from "Components/Search/NewSearch/Mobile/OverlayBase"
+import { SearchResultsListQueryRenderer } from "Components/Search/NewSearch/Mobile/SearchResultsList"
 
 const logger = createLogger("Components/Search/NewSearch/Mobile")
 
+const scrollToTop = () => {
```
damassi, line 22: Can we use this? https://github.com/artsy/force/blob/main/src/Utils/Hooks/useJump.tsx

damassi: Also @dzucconi - do you think `useJump` might be less discoverable than `useScrollTo`? There's also a few other `useScrollToThing` hooks in the utils/hooks folder

dzucconi: useJump is useful if we need to accommodate sticky items and/or nav. I don't see why we can't simply `window.scrollTo(0, 0)` here?

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+    return <NoResults query={query} mt={50} mx={20} />
+  }
+
+  tracking.trackEvent({
+    action_type:
+      options.length > 0
+        ? DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
+        : DeprecatedSchema.ActionType.SearchedAutosuggestWithoutResults,
+    query: query,
+  })
```
damassi, line 57: side-effects should be called from `useEffect`

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+      })}
+
+      {relay.hasMore() && <InfiniteScrollSentinel onNext={handleLoadMore} />}
+
+      {!relay.hasMore() && (
+        <SeeFullResults query={query} onClick={handleOnRedirect} />
+      )}
+
+      {isLoading && (
+        <Flex width="100%" minHeight="60px" alignItems="center">
```
dzucconi, line 103: Hardcoded height is odd

damassi (commented):
Are there plans to add any tests?

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+          first: $first
+          after: $after
+        )
+          @include(if: $hasTerm)
+          @connection(key: "SearchResultsList_searchConnection") {
+          edges {
+            node {
+              displayLabel
+              href
+              imageUrl
```
dzucconi, line 136: It doesn't look like any of these properties are being used in this component? Is there a missing fragment container?

src/Components/Search/NewSearch/Mobile/Overlay.tsx
damassi: Cool, np then 👍

src/Components/Search/NewSearch/Mobile/SearchResultsList/NoResults.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const NoResults: FC<NoResultsProps> = ({ query, ...rest }) => {
+  const { t } = useTranslation()
+
+  return (
+    <Flex flexDirection="column" {...rest}>
+      <Text alignContent="center" textAlign="center">
+        {t("searchApp.noResults.sorryMessage")} “<Highlight>{query}</Highlight>”
+      </Text>
+      <Text color="black60" textAlign="center" mt={10}>
```
dzucconi, line 18:
```suggestion
      <Text color="black60" textAlign="center" mt={1}>
```

src/Components/Search/NewSearch/Mobile/SearchResultsList/SeeFullResults.tsx
```diff
@@ -0,0 +1,43 @@
...
+          <SearchIcon width={22} height={22} />
+        </Flex>
+
+        <Spacer x={1} />
+
+        <Text
+          variant="sm"
+          mr={1}
+          overflow="hidden"
+          style={{ textOverflow: "ellipsis" }}
```
dzucconi, line 36: You can just use the `ellipsis` prop

src/Components/Search/NewSearch/Mobile/SearchResultsList/SeeFullResults.tsx
```diff
@@ -0,0 +1,43 @@
...
+          alignItems="center"
+        >
+          <SearchIcon width={22} height={22} />
+        </Flex>
+
+        <Spacer x={1} />
+
+        <Text
+          variant="sm"
+          mr={1}
```
dzucconi, line 34: Probably should be padding on the container?

src/Components/Search/NewSearch/utils/formatOptions.ts
```diff
@@ -0,0 +1,37 @@
...
+        }) ?? "",
+      imageUrl: option.imageUrl!,
+      showArtworksButton: !!option.statuses?.artworks,
+      showAuctionResultsButton: !!option.statuses?.auctionLots,
+      href: option.href!,
+      typename: option.__typename,
+      item_number: index,
+      item_type: option.displayType!,
+    }
+  })
```
dzucconi, line 36: Not sure why we would do something like this; build fragment containers instead.

dzucconi: (This kind of thing makes refactors extremely difficult)

src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -87,37 +83,57 @@ export const Overlay: FC<OverlayProps> = ({ viewer, relay, onClose }) => {
...
-          value={value}
-          placeholder={t`navbar.searchArtsy`}
-          label={<SearchIcon fill="black60" aria-hidden mr={-10} size={18} />}
-          onChange={handleValueChange}
-          // px={2}
+    <OverlayBase
+      onClose={onClose}
+      header={
+        <>
+          <Box mt={-15}>
```
dzucconi, line 104: What's this for?

src/Components/Search/NewSearch/utils/formatOptions.ts
damassi: @dzucconi - mind elaborating a bit on what this would ideally look like? Like just move this logic into a component and tie the props to the relay generated type?

dzucconi: To be clear: there should be a fragment container for the search result items, and those components can handle this formatting themselves.

src/Components/Search/NewSearch/Mobile/SearchResultsList/InfiniteScrollSentinel.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box } from "@artsy/palette"
+import { useIntersectionObserver } from "Utils/Hooks/useIntersectionObserver"
+import { FC } from "react"
+
+interface InfiniteScrollSentinelProps {
+  onNext: () => void
+}
+
+export const InfiniteScrollSentinel: FC<InfiniteScrollSentinelProps> = ({
```
dimatretyak, line 9: **suggestion**: make it reusable. since we already have components with the same code and logic ([component 1](https://github.com/artsy/force/blob/145c983c61757b89ba1dc7c864d8da57e2d6035e/src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/ArtworksInfiniteScrollSentinel.tsx#L9), [component 2](https://github.com/artsy/force/blob/145c983c61757b89ba1dc7c864d8da57e2d6035e/src/Components/Notifications/NotificationsListScrollSentinel.tsx#L8), [component 3](https://github.com/artsy/force/blob/145c983c61757b89ba1dc7c864d8da57e2d6035e/src/Apps/Article/Components/ArticleInfiniteScroll.tsx#L92-L99))

src/Components/Search/NewSearch/Mobile/SearchResultsList/NoResults.tsx
dimatretyak, line 18: **suggestion**: instead of `mt={10}` use `mt={1}` (I think it's the same thing)

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+      {isLoading && (
+        <Flex width="100%" minHeight="60px" alignItems="center">
+          <Spinner position="relative" />
+        </Flex>
+      )}
+    </>
+  )
+}
+
+const SearchResultsListPaginationContainer = createPaginationContainer(
```
dimatretyak, line 111: **question**: should we support pagination for mobile view? As far as I can see by [design in Figma](https://www.figma.com/file/i5mJhlbbC4XEvvrwpImjGp/Search-Experience---Ideas-%2B-Mocks?type=design&node-id=336-44450&t=dpbI9xlBp8o3WPDo-0), instead of pagination, we should display the "See full results" button

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+            onRedirect={handleOnRedirect}
+            key={index}
+          />
+        )
+      })}
+
+      {relay.hasMore() && <InfiniteScrollSentinel onNext={handleLoadMore} />}
+
+      {!relay.hasMore() && (
+        <SeeFullResults query={query} onClick={handleOnRedirect} />
```
dimatretyak, line 99: **question**: It seems to me a bit illogical to show the "See Full results" button when we are already displaying all the results (since `hasMore()` is false)

src/Components/Search/NewSearch/Mobile/SearchResultsList/NoResults.tsx
dimatretyak: Duplicate: https://github.com/artsy/force/pull/12410#discussion_r1195446834

src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -87,37 +83,57 @@ export const Overlay: FC<OverlayProps> = ({ viewer, relay, onClose }) => {
...
+            <LabeledInput
+              mx={2}
+              ref={inputRef}
+              value={inputValue}
+              placeholder={t`navbar.searchArtsy`}
+              label={
+                <SearchIcon fill="black60" aria-hidden mr={-10} size={18} />
```
dimatretyak, line 111: **suggestion**: do we really want (and need) to use `mr={-10}` here?

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: How would you fix it? I just took the value from [SuggestedItemLink](https://github.com/artsy/force/blob/main/src/Components/Search/NewSearch/SuggestionItem/SuggestionItemLink.ts#L14) so it looks consistent.

nickskalkin:
> Are there plans to add any tests?

@damassi yes, definitely. I don't write tests right away, first I want to pass code review. I'm still not comfortable around Force, there is a high chance that you will ask me to re-write a lot of code and therefore tests

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: I pass these properties to the [formatOptions](https://github.com/artsy/force/pull/12410/files#diff-236782da27733a3ff4c1063878a65fb8dcd8e43348e55fde1e5319b57b9bd884R16) helper function.

src/Components/Search/NewSearch/utils/formatOptions.ts
nickskalkin: Yeah, makes sense, I will refactor this

src/Components/Search/NewSearch/Mobile/MobileSearchBar.tsx
```diff
@@ -91,9 +91,7 @@ export const MobileSearchBarQueryRenderer: FC<MobileSearchBarQueryRendererProps>
         }
       `}
       variables={{
-        hasTerm: false,
         term: "",
-        entities: [],
```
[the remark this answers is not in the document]
nickskalkin: I think we need this data there, no? It's used in `searchConnection` [here](https://github.com/artsy/force/blob/145c983c61757b89ba1dc7c864d8da57e2d6035e/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L215)

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+      )}
+
+      {isLoading && (
+        <Flex width="100%" minHeight="60px" alignItems="center">
+          <Spinner position="relative" />
+        </Flex>
+      )}
```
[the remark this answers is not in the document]
nickskalkin: Could you elaborate a bit? Better because we don't need extra state in the component or are there other reasons?

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+  query,
+  onClose,
+}) => {
+  const tracking = useTracking()
+  const [isLoading, setIsLoading] = useState(false)
+
+  const options = extractNodes(viewer.searchConnection)
+
+  if (!relay.isLoading() && options.length === 0) {
+    return <NoResults query={query} mt={50} mx={20} />
```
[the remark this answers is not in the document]
nickskalkin: I think so, at least this way it is close to design specs

Screenshot 2023-05-17 at 10 44 30

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+        ? DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
+        : DeprecatedSchema.ActionType.SearchedAutosuggestWithoutResults,
+    query: query,
+  })
+
+  const formattedOptions: SuggionItemOptionProps[] = formatOptions(
+    options as SearchNodeOption[]
+  )
+
+  const handleLoadMore = () => {
```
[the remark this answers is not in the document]
nickskalkin: No, for desktop we always have 7 results + "See full results" row

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+      {relay.hasMore() && <InfiniteScrollSentinel onNext={handleLoadMore} />}
+
+      {!relay.hasMore() && (
+        <SeeFullResults query={query} onClick={handleOnRedirect} />
+      )}
+
+      {isLoading && (
+        <Flex width="100%" minHeight="60px" alignItems="center">
+          <Spinner position="relative" />
```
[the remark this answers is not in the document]
nickskalkin: By default Spinner's position is `absolute`. It will look like this:

Screenshot 2023-05-17 at 10 49 34

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: following Dima's suggestion here: https://github.com/artsy/force/pull/12410#discussion_r1195445815

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: but will be refactored because of [this thread](https://github.com/artsy/force/pull/12410#discussion_r1195450367)

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: There is an "Advanced Options" [section](https://www.figma.com/file/i5mJhlbbC4XEvvrwpImjGp/Search-Experience---Ideas-%2B-Mocks?type=design&node-id=336-22330&t=EyiCog9nve22KiD1-0) which suggests implementing infinite scrolling. In this case it probably doesn't make sense to implement "See full results" link like you said here: https://github.com/artsy/force/pull/12410#discussion_r1195476108

I will remove "See full results" then

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
nickskalkin: I think you are right, I will remove "See full results" link like said [here](https://github.com/artsy/force/pull/12410#discussion_r1196179439)

src/Components/Search/NewSearch/Mobile/OverlayBase.tsx
```diff
@@ -1,32 +1,80 @@
...
+
+const Close = styled(Clickable)`
+  color: ${themeGet("colors.black100")};
+
+  &:focus,
+  &.focus-visible {
+    outline: none;
+    color: ${themeGet("colors.black60")};
+  }
+`
```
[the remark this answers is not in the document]
dimatretyak: yep, sure

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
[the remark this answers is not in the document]
dimatretyak: Got it. In that case I think it would be better to use one of the spacing values from [here](https://palette.artsy.net/atoms/spacing/#all-spacing-values). E.g. instead of `mt={50}` use `mt={4}` or `mt={6`}, and `mx={2}` instead of `mx={20}`

damassi:
> I don't write tests right away

I'm the same way @nickskalkin. No point writing tests if the code isn't stable yet. 

But yah, as there is a lot here, as long as there's a plan to return at the end, all good 👍

dimatretyak (dismissed):

src/Components/Search/NewSearch/Mobile/Overlay.tsx
dzucconi: I know I say this _a lot_ but really: please don't fight against what Palette proscribes. If the spec is intentional then frequently it is an oversight. It may be the case that we need a labelled input that has an icon that is flush with the edge but if that was the case then if the component is updated this instance will break.

src/Components/Search/NewSearch/useTrackSearchResults.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const useTrackSearchResults = (options: Options) => {
+  const tracking = useTracking()
+
+  // We use fetchCounter together with useUpdateEffect to track updating search query
+  const [fetchCounter, setFetchCounter] = useState(0)
+
+  const searchResultsUpdated = useCallback(() => {
+    setFetchCounter(prevCounter => prevCounter + 1)
+  }, [setFetchCounter])
```
dzucconi, line 22: Why wouldn't you just call the trackEvent here, why the indirection with the counter state?

dimatretyak: * We should track `ActionType.SearchedAutosuggestWithResults` if there are results for the query
* We should track `ActionType.SearchedAutosuggestWithoutResults` if there are **no** results for the query
* Refetch `callback` returns only `error` (see attached screenshot), and we have no way to determine if there are results for the query or not. For this reason, `searchResultsUpdated` is called, which updates the `counter` and tracks [the event](https://github.com/artsy/force/blob/5f925aa83d2fbd4f1e043ac80b7b7b75fac072c6/src/Components/Search/NewSearch/useTrackSearchResults.ts#L28-L35) with the current results for the query

for-damon

dimatretyak: Tried [this approach](https://github.com/artsy/force/commit/a85466e2afd760ebecce1a746eef922e46c29d36). But unfortunately it does **not** work correctly (events are duplicated)

https://github.com/artsy/force/assets/3513494/853bdfb9-591a-449b-88c9-73ad48951c7b

nickskalkin:
We've chatted with @dzucconi yesterday and decided to merge this PR. There are some code review suggestions that I do not know how to resolve, we've agreed to merge this PR in any case and follow-up on them later

PR #12422, opened 2023-05-18
fix(price-database): implements a working autocomplete

dzucconi:
I saw this mentioned a few times in Slack so just figured I'd do a quick fix while I was working with autocompletes. I am not entirely sure if this has been in this state for a while. Code gets a lot simpler. It could still use a debounce and "no results" state but it at least works now.

The page has some layout issues but it'd be nice to find a Figma spec for it before adjusting.

cc @artsy/collector-experience


olerichter00 (dismissed):
Thanks 🙏

mbransfield (approved):

dzucconi:
One exciting thing is we can finally drop `react-autosuggest` once the new nav search lands

PR #12430, opened 2023-05-20
chore(GRO-1671): update meta tags based on the selected tab on the Artist page

pam-:

This PR solves [GRO-1671]

### Description

* uses the `ArtistTabEnumType` set in MP to easily differentiate the meta tags content based on the selected tab.
* removes `computeTitle` helper since all the data is now set in MP.

**About**
about-meta
  
**Artworks**
artworks-meta
  
**Auction Results**
auction-results-meta


kajatiger:
The CI is breaking with some weird problem in the image build (end of file error) 🤔

data/schema.graphql
```diff
@@ -1966,6 +1966,12 @@ type ArtistStatuses {
   shows: Boolean
 }
 
+enum ArtistPage {
+  ABOUT
+  ARTWORKS
+  AUCTION_RESULTS
+}
+
```
dzucconi, line 1974: I would wait for the automation to pull these schema changes over.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -58,6 +58,25 @@ interface AuctionResultsProps {
   artist: ArtistAuctionResults_artist$data
 }
 
+interface AuctionMetaTagsProps {
+  meta: {
+    title: string
+    description: string
+  }
+}
```
dzucconi, line 66: We don't manually type props like this. You'd want to use a fragment container. I would probably just inline this in the parent component; there's nothing really here.

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -215,13 +215,30 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
     )
   }
 
+  const AuctionMetaTags = ({ meta }) => {
```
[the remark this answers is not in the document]
pam-: ooh aiite

PR #12432, opened 2023-05-22
fix(artist): fixes empty state for filters on auction-results

dzucconi:
Closes [GRO-1725](https://artsyproduct.atlassian.net/browse/GRO-1725)

We can check the `counts.auctionResults` to display the full empty component; when there's 0 results we can just leverage the existing filter empty component.

A slight rough patch here is that it doesn't handle resetting the keyword correctly:

https://github.com/artsy/force/assets/112297/b703816e-83cb-4fa5-ab13-fbbce55e63a6

The main reason this is an issue is because we're debouncing the query by debouncing setting the keyword. This is a bit backwards: we want to actually debounce the query execution for all filters. The filter context state should be a true representation. I'm not going to fix that in this PR however.

[GRO-1725]: https://artsyproduct.atlassian.net/browse/GRO-1725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -1,6 +1,14 @@
 import { ContextModule, Intent } from "@artsy/cohesion"
 import * as DeprecatedAnalyticsSchema from "@artsy/cohesion/dist/DeprecatedSchema"
-import { Box, Column, GridColumns, Join, Spacer, Text } from "@artsy/palette"
+import {
+  Box,
+  Column,
+  Flex,
+  GridColumns,
+  Join,
+  Spacer,
+  Text,
+} from "@artsy/palette"
 import { allowedAuctionResultFilters } from "Apps/Artist/Utils/allowedAuctionResultFilters"
 import {
   paramsToCamelCase,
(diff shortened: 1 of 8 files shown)


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -274,62 +283,70 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
 
           <Spacer y={[2, 0]} />
 
+          {results.length === 0 && (
+            <ArtworkGridEmptyState
+              onClearFilters={() => {
+                resetFilters?.()
```
dzucconi, line 289: The erroneous nullable typing on all these filter contexts is super frustrating.

src/Apps/Artist/Routes/AuctionResults/__tests__/ArtistAuctionResults.jest.tsx
```diff
@@ -514,12 +514,16 @@ describe("AuctionResults", () => {
   })
 })
 
+// FIXME: Should not be using fixtures
```
dzucconi, line 517: 😞

kajatiger (approved):
It's a bit of a hacky way to reset the filters and I have no idea how this is usually done, but it definitely fixes the dead end problem, thank you for doing that so quickly. Do we have a desing for this case or is it obvious that we just use the same empty state for the no results case?

dzucconi:
> It's a bit of a hacky way to reset the filters and I have no idea how this is usually done
It's just a function that resets the filter context state; that's not hacky. The thing that's a problem is one of the filters debounced the wrong thing.

> Do we have a desing for this case or is it obvious that we just use the same empty state for the no results case?
Same empty state as the filters since it's a filter view.

PR #12433, opened 2023-05-22
fix(artist): fixes content shift for market signals component

dzucconi:
Just something easy to fix I noticed when looking at this page. There is currently a content shift due to an out of date placeholder state.

This PR cleans that up and fixes a few other things, including the wrong up/down arrows, which have now been replaced with the appropriate icons.


src/Apps/Artist/Routes/AuctionResults/Components/MarketStats.tsx
```diff
@@ -264,13 +324,12 @@ export const MarketStatsQueryRenderer: React.FC<{
       render={({ props, error }) => {
         if (error) {
           onRender(false)
```
dzucconi, line 326: This `onRender` stuff is weird and could be simplified. Some other day.

PR #12440, opened 2023-05-23
fix(artwork lists): improve a11y by making focus visible

anandaroop:

This PR solves [FX-4722]

Review app: https://fx-4722-a11y.artsy.net/

### Description

Restores the focus styles that were missing when you navigated the Artwork Lists rail via keyboard.

This required moving the styling from the `Flex` to the containing `RouterLink`, and giving it `display: block` to fix the collapsed focus outline problem.

Both of the screencaps below show the results of keyboard navigation: navigating with `Tab` and selecting with `Enter`.

|Before|After|
|---|---|
| <video src="https://github.com/artsy/force/assets/140521/90cceef1-2b85-4ae2-bdb5-1fa1b5eeaebc" /> | <video src="https://github.com/artsy/force/assets/140521/fef166a7-7134-4744-bb03-5e17e86a60a9" /> |

The Artwork List cards _were_ in fact already focusable and tabbable, as you can see in the **Before** screencap above.

But they didn't display the customary `focus-visible` outline until the `display: block` fix, visible in the **After** screencap.

[FX-4722]: https://artsyproduct.atlassian.net/browse/FX-4722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListItem.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListItem.tsx
@@ -4,10 +4,11 @@ import { StackedImageLayout } from "./Images/StackedImageLayout"
 import { FC } from "react"
 import { useTranslation } from "react-i18next"
 import { createFragmentContainer, graphql } from "react-relay"
-import { RouterLink } from "System/Router/RouterLink"
+import { RouterLink, RouterLinkProps } from "System/Router/RouterLink"
 import { extractNodes } from "Utils/extractNodes"
 import { ArtworkListItem_item$data } from "__generated__/ArtworkListItem_item.graphql"
 import { BASE_SAVES_PATH } from "Apps/CollectorProfile/constants"
+import styled from "styled-components"
 
 interface ArtworkListItemProps {
   isSelected?: boolean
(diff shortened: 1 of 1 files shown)


src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListItem.tsx
```diff
@@ -92,3 +87,18 @@ export const ArtworkListItemFragmentContainer = createFragmentContainer(
...
+}
+
+const ArtworkListItemLink = styled<ArtworkListItemLinkProps>(RouterLink)`
+  /* always */
+  border-radius: 10px;
+  margin-top: 2px; // otherwise top borders get cut off
+  display: block; // otherwise the child element collapses without a visible focus outline
+
+  /* when this list is currently being viewed (isSelected) */
+  border: solid 1px ${props => (!!props.isSelected ? "black" : "transparent")};
```
dimatretyak, line 102: **question**: did you clarified black border color with Ryan?

dimatretyak: if not, what do you think about using `outline-offset: 0` here? It will look something like this


https://github.com/artsy/force/assets/3513494/0f660997-86a9-4954-84e6-d6c8418e2594

dblandin (approved):
LGTM! 👍🏼

dimatretyak (approved):

PR #12446, opened 2023-05-24
chore(deps): removes unused dependencies

dzucconi:
Just removing some obviously no longer used dependencies.

Also noticing a bunch where there's essentially a single import that's easily replaced with something else we already use.


package.json
```diff
@@ -203,7 +199,6 @@
     "yup": "0.32.6"
   },
   "devDependencies": {
-    "@artsy/antigravity": "0.3.0",
```
dzucconi, line 206: 👋

package.json
```diff
@@ -284,7 +278,6 @@
     "babel-plugin-module-resolver": "3.1.0",
     "babel-plugin-relay": "^14.1.0",
     "babel-plugin-styled-components": "1.12.0",
-    "benv": "3.3.0",
```
dzucconi, line 287: 👋

PR #12447, opened 2023-05-24
chore: add error message event tracking for payment step

rquartararo:

Paired with @iskounen 

This PR adds the new `errorMessageViewed` [event](https://github.com/artsy/cohesion/pull/431) to the payments order route whenever a user views an error dialog. 

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -237,6 +237,17 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
       if (result?.type === "invalid_form") return
 
       if (result?.type === "error") {
+        trackEvent({
+          action: ActionType.errorMessageViewed,
+          context_owner_type: OwnerType.ordersPayment,
+          context_owner_id: props.order.internalID,
+          title: result.error,
+          message:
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -299,6 +333,18 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
+        action: ActionType.errorMessageViewed,
+        context_owner_type: OwnerType.ordersPayment,
+        context_owner_id: props.order.internalID,
+        title: "An error occurred",
+        message:
+          "Something went wrong. Please try again or contact order@artsy.net",
+        error_code: null,
+        flow: "user sets wire transfer as payment method",
+      })
+
```
AlicanAkyuz, line 347: suggestion for future: to prevent duplication and make future updates easier, we could wrap the call to trackEvent in a function with static key/value pairs (action, owner type and id) and call that function with the dynamic values (title, message, etc.).

AlicanAkyuz: oh it'd have been even better if we could share that function among the routes. 

_I think_ there's no direct way to that though, since each route is loaded separately. perhaps @damassi knows a way.

PR #12448, opened 2023-05-25
feat: add filter for location to artist auction results

kajatiger:

This PR solves [GRO-1738]

### Description

Part of the bigger Artist Page makeover is to enhance the filter options for the auction results tab. A new filter we want to add is the location filter. This is what the figma design looks like:

locationfigma

This is what the local changes look like:

https://github.com/artsy/force/assets/15628617/6c4e187e-3fc0-4109-8029-7ec0f5909211


src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/LocationFilter.tsx
```diff
@@ -0,0 +1,57 @@
...
+            const props = {
+              key: index,
+              onSelect: (selected: boolean) => {
+                toggleSelection(selected, name)
+              },
+              my: 1,
+              selected: locations?.includes(name),
+              testID: `location-filter-${name}`,
+            }
+            // eslint-disable-next-line react/jsx-key
```
dzucconi, line 50: You'll need to provide a `key`

src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/LocationFilter.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { FilterExpandable } from "Components/ArtworkFilter/ArtworkFilters/FilterExpandable"
+
+export const locationMap = [
+  { displayName: "New York", name: "New York" },
+  { displayName: "Shanghai", name: "Shanghai" },
+  { displayName: "London", name: "London" },
+  { displayName: "Paris", name: "Paris" },
+  { displayName: "Vienna", name: "Vienna" },
+  { displayName: "Amsterdam", name: "Amsterdam" },
+]
```
dzucconi, line 17: Wouldn't we want to be pulling these from the actual results? I don't think we want a hardcoded list?

kajatiger:
I'll close this PR for now, since the whole backend implementation is not decided on yet.

PR #12454, opened 2023-05-25
fix: corrects content shift in some instances of systemqueryrenderer usage

dzucconi:
I noticed that in the new artist header there was a subtle content shift that would happen around the follow buttons. Turns out that this was due to our usage of a sentinel div for lazy loading of `SystemQueryRenderer`.

The fix here is to attempt to forward the ref to the placeholder. If we can't then we fall back to the sentinel method.

I've also replaced the usage of the transitive dep `react-lazy-load-image-component` and for bonus points removed its erroneous usage on the collection pages.


src/Apps/Collect/Routes/Collection/index.tsx
```diff
@@ -123,23 +122,17 @@ export const CollectionApp: React.FC<CollectionAppProps> = props => {
           counts={collection.artworksConnection?.counts as Counts}
         />
 
-        {/* HOTFIX FIXME: This rail was causing an error if included in SSR render
-              pass and so it was deferred to the client.
-
-              See: https://github.com/artsy/force/pull/6137
-          */}
-        {collection.linkedCollections.length === 0 &&
-          typeof window !== "undefined" && (
```
dzucconi, line 132: This also would have been causing a server-side rendering mismatch.

src/Apps/Collect/Routes/Collection/index.tsx
```diff
@@ -123,23 +122,17 @@ export const CollectionApp: React.FC<CollectionAppProps> = props => {
...
-            <LazyLoadComponent threshold={1000}>
```
dzucconi, line 133: This didn't do anything to begin with since it's a fragment container. There was a waypoint usage in that component used for tracking, maybe that's what it originally was doing.

src/Apps/Collect/Routes/Collection/index.tsx
damassi: I remember this -- for whatever reason which made zero sense at all and took ages to track down -- would cause a full on crash somewhere in the router sometimes due to some kind of subtle race condition. This was way long ago and could definitively have been fixed somewhere along the way by us just moving everything into a proper SPA and so on and so forth

PR #12459, opened 2023-05-26
feat: Users and visitors sees hero units GRO-1687

jonallured:
This PR sits on top of https://github.com/artsy/metaphysics/pull/5019 and migrates force from Braze content cards to our own home-grown hero units. It is meant to be a visual noop where only the source of the homepage content is altered but the user experience does not change at all.

Looks like this:

Screenshot 2023-05-26 at 12 45 32 PM

My plan is to create a review app for QA next week but once this PR is merged we will be migrated away from Braze for this stuff.

Note: we get to _remove_ a lib! We'll no longer need the braze web sdk.

## Review App

https://jonallured-hero-units.artsy.net

https://artsyproduct.atlassian.net/browse/GRO-1687

/cc @artsy/grow-devs @kathytafel @isakelaris


src/Apps/Home/__tests__/HomeApp.jest.tsx
```diff
@@ -34,6 +34,9 @@ describe("HomeApp", () => {
         featuredEventsOrderedSet: orderedSet(id: "example") {
           ...HomeApp_featuredEventsOrderedSet
         }
+        heroUnitsConnection(first: 10) {
+          ...HomeApp_heroUnitsConnection
+        }
```
jonallured, line 39: Hey @dzucconi I'm getting a type error in this test but I can't figure out what else it wants me to do? Shouldn't the relay generator update the test with the right shape? Is there something I'm missing?

jonallured: Here's a link to the error:

https://app.circleci.com/pipelines/github/artsy/force/48962/workflows/7099df50-d67d-424e-b72a-24965647b1ac/jobs/367169

dzucconi: What's the type error?

dzucconi:
Looks good

src/Apps/Home/__tests__/HomeApp.jest.tsx
dzucconi: Ah, that's just because it can be nullable. We should probably alter the type signature of setupTestWrapper to assert that it'll be there but for now you can either ts-ignore it or null check for the props. If you search for `Component: props =>` you'll see what I mean.

jonallured:
Ok merging!

PR #12464, opened 2023-05-30
fix: Check for presence of shippingTotalCents and not numeric value

jpotts244:

This PR solves [PROJECT-XX]

### Description

Fix for the rogue `0` that gets returned as part of order screens if the order is PICKUP (aka shipping is `0.00`)

Screen Shot 2023-05-30 at 9 36 12 AM


Mitch-Henson (approved):
Ahh nice find ‼️

PR #12492, opened 2023-06-05
Revert "fix(usesizeandposition): ensures that if node is added after hook is set up; we still get geometry"

dzucconi:
Reverts artsy/force#12489


dzucconi:
Not sure why this is locking up the fair page; will investigate after reverting/deploying

PR #12498, opened 2023-06-06
chore(docs): move cookie consent docs to /docs

damassi:
This moves the cookie consent docs over to the root level `/docs` folder for better discoverability.


dzucconi:
Let's hold off on any changes to this PR to ease our ability to revert. Will re-open in a few days.

PR #12502, opened 2023-06-07
feat(artwork lists): improve visibility of artwork list content on smaller screens

anandaroop:

This PR solves [FX-4802]  & [FX-4803]

Review app: https://fx-artwork-list-scroll.artsy.net/collector-profile/saves

## Description

Based on internal beta feedback we want to address the usability of the Artwork Lists area on small screens.

This PR introduces some selective auto-scroll behavior, and aims to be as helpful and unobtrusive as possible.

Therefore an artwork list's artwork grid content will be scrolled into view when and only when:

- the artwork list content is **out of view** (i.e. below the bottom edge of the viewport https://github.com/artsy/force/pull/12502/commits/90f6ee13f7b0c49127e8c88c8ebecd5616166be8)

- the user is **expecting** to see new content (i.e. they have interacted with the Artwork Lists rail https://github.com/artsy/force/pull/12502/commits/845ee47a4e89817bd468d31aefc92e38a673707d)

- the user is likely **unaware** there may be more content (i.e. we have not auto-scrolled the grid content into view at least once already https://github.com/artsy/force/pull/12502/commits/50ad5746630ac5b0c134368888667e23c0c767db)

## Examples

|large screen|small screen, needs scroll|small screen, already auto-scrolled|
|---|---|---|
|On a large enough screen, no action is necessary, so we do nothing|This is the case where we assume action is necessary, so we scroll the content into view|Since we have already scrolled content into view previously, we avoid over-assertively doing it again. |
|<video src="https://github.com/artsy/force/assets/140521/315e88e6-891b-4f01-83f8-8216a26acc1c" />|<video src="https://github.com/artsy/force/assets/140521/6b747f1c-371c-4678-88d1-2d166d973ab3" />|<video src="https://github.com/artsy/force/assets/140521/543551b7-959d-4c31-a6a3-6d9a7c74511d" />|

(Previously: [Loom](https://www.loom.com/share/a4b8c81b6f9244ffbe301e33dbba554f) with older version of review app.)

[FX-4802]: https://artsyproduct.atlassian.net/browse/FX-4802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FX-4803]: https://artsyproduct.atlassian.net/browse/FX-4803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -15,6 +15,8 @@ import { MetaTags } from "Components/MetaTags"
 import { Jump } from "Utils/Hooks/useJump"
 import { ArtworkListVisibilityProvider } from "Apps/CollectorProfile/Routes/Saves2/Utils/useArtworkListVisibility"
 
+export const ARTWORK_LIST_SCROLL_TARGET_ID = "ArtworkListScrollTarget"
+
 interface CollectorProfileSaves2RouteProps {
   me: CollectorProfileSaves2Route_me$data
 }
@@ -88,7 +90,7 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
         }
       />
 
-      <Jump id="AboveArtworkListShelf" />
+      <Jump id={ARTWORK_LIST_SCROLL_TARGET_ID} />
 
       <Spacer y={4} />
 
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -12,6 +12,7 @@ import { ActionType, OwnerType, ViewedArtworkList } from "@artsy/cohesion"
 import { AnalyticsContext } from "System/Analytics/AnalyticsContext"
 import { HttpError } from "found"
 import { MetaTags } from "Components/MetaTags"
+import { Jump } from "Utils/Hooks/useJump"
 
 interface CollectorProfileSaves2RouteProps {
   me: CollectorProfileSaves2Route_me$data
@@ -85,6 +86,8 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
           me?.savedArtworksArtworkList?.artworksConnection?.totalCount ?? 0
         }
       />
+
+      <Jump id="AboveArtworkListShelf" />
 
       <Spacer y={4} />
 
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -13,6 +13,7 @@ import { AnalyticsContext } from "System/Analytics/AnalyticsContext"
 import { HttpError } from "found"
 import { MetaTags } from "Components/MetaTags"
 import { Jump } from "Utils/Hooks/useJump"
+import { ArtworkListVisibilityProvider } from "Apps/CollectorProfile/Routes/Saves2/Utils/useArtworkListVisibility"
 
 interface CollectorProfileSaves2RouteProps {
   me: CollectorProfileSaves2Route_me$data
@@ -78,7 +79,7 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
   }
 
   return (
-    <>
+    <ArtworkListVisibilityProvider>
       <MetaTags title="Saves | Artsy" pathname="collector-profile/saves" />
 
       <ArtworkListsHeader
@@ -112,7 +113,7 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
         initialPage={(page as unknown) as number}
         initialSort={sort}
       />
-    </>
+    </ArtworkListVisibilityProvider>
   )
 }
 
--- a/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
@@ -104,8 +104,6 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
         })}
       </Shelf>
 
-      <Spacer y={4} />
-
       <ArtworkListContentQueryRenderer
         listID={selectedArtworkListId}
         initialPage={(page as unknown) as number}
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListArtworksGrid.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListArtworksGrid.tsx
@@ -145,6 +145,9 @@ const ArtworkListArtworksGrid: FC<ArtworkListArtworksGridProps> = ({
 
   return (
     <>
+      {/* Note: this <Jump> element is also used as a sentinel by the
+      shouldScroll() helper in the ArtworkListContent component  */}
+
       <Jump id="artworksGrid" />
 
       <ArtworkListArtworksGridHeader />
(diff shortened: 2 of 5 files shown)


src/Apps/CollectorProfile/CollectorProfileApp.tsx
```diff
@@ -37,45 +73,101 @@ const CollectorProfileApp: React.FC<CollectorProfileAppProps> = ({
...
+      justifyContent={"space-between"}
+      height={"24em"}
+      position="fixed"
+      top="4em"
+      right="4em"
+      bg="#ffffffcc"
+      border="solid 2px blue"
+      p={2}
+      zIndex={1000}
+      style={{ boxShadow: "0 0 20px #99c" }}
```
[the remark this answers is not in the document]
dzucconi: (Oh, it's a demo!)

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListArtworksGrid.tsx
```diff
@@ -145,6 +145,9 @@ const ArtworkListArtworksGrid: FC<ArtworkListArtworksGridProps> = ({
 
   return (
     <>
+      {/* Note: this <Jump> element is also used as a sentinel by the
+      shouldScroll() helper in the ArtworkListContent component  */}
+
       <Jump id="artworksGrid" />
```
anandaroop, line 151: I used this as the element for which to check visibility, because it was already available. 

However we may want to add a new sentinel element instead, a little further down and closer to the artwork grid, if we want to make it a bit more likely that we scroll content into view.

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
```diff
@@ -34,14 +36,47 @@ const sortOptions: SortOptions = [
...
+  useEffect(() => {
+    const shouldScroll =
+      isContentOutOfView() &&
+      artworkListItemHasBeenTouched &&
+      !artworkListHasBeenScrolled
+
+    if (shouldScroll) {
+      jumpTo("AboveArtworkListShelf")
+      setArtworkListHasBeenScrolled()
+    }
```
anandaroop, line 72: The gist of it

src/Apps/CollectorProfile/Routes/Saves2/Utils/useArtworkListVisibility.tsx
```diff
@@ -0,0 +1,51 @@
...
+    /** To be called once any Artwork List card has been selected */
+    setArtworkListItemHasBeenTouched: () => setTouched(true),
+
+    /** True if we have already auto-scrolled the Artwork List page */
+    artworkListHasBeenScrolled: scrolled,
+
+    /** To be called once we have auto-scrolled the Artwork List page */
+    setArtworkListHasBeenScrolled: () => setScrolled(true),
+  }
+}
```
anandaroop, line 51: This hook lets components nested within `ArtworkListVisibilityProvider` access and set these flags.

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
```diff
@@ -34,14 +36,47 @@ const sortOptions: SortOptions = [
...
   const artworkList = me.artworkList!
   const counts: Counts = {
     artworks: artworkList.artworks?.totalCount ?? 0,
   }
 
+  useEffect(() => {
+    const shouldScroll =
+      isContentOutOfView() &&
+      artworkListItemHasBeenTouched &&
+      !artworkListHasBeenScrolled
```
dimatretyak, line 67: I think we should reset these values (`artworkListItemHasBeenTouched`, `artworkListHasBeenScrolled`), otherwise scrolling won't work when clicking on another artwork list and scrolling happened before


https://github.com/artsy/force/assets/3513494/6017f68f-3293-48ab-8636-d9896c8612af

src/Apps/CollectorProfile/Routes/Saves2/CollectorProfileSaves2Route.tsx
```diff
@@ -86,6 +88,8 @@ const CollectorProfileSaves2Route: FC<CollectorProfileSaves2RouteProps> = ({
         }
       />
 
+      <Jump id="AboveArtworkListShelf" />
```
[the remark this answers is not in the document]
anandaroop: Good idea. 60633f0ece8f1b16e55d16b1396470c1552ab3e9

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
```diff
@@ -34,14 +36,47 @@ const sortOptions: SortOptions = [
 ]
 const defaultSort = sortOptions[0].value
 
+function isContentOutOfView() {
+  const element = document.querySelector("#JUMP--artworksGrid")
```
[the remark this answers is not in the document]
anandaroop: Yeah, this is safer. Also in 60633f0ece8f1b16e55d16b1396470c1552ab3e9

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
anandaroop: Yeah, I added `artworkListHasBeenScrolled` deliberately to prevent auto-scrolling more than once — but if we don't want to prevent it, I can just revert 50ad5746630ac5b0c134368888667e23c0c767db and we won't see the behavior above.

(`artworkListItemHasBeenTouched` was also intentionally set once based on [this feedback](https://artsy.slack.com/archives/C9SATFLUU/p1686155549672359?thread_ts=1686154582.647879&cid=C9SATFLUU) from @dblandin. I think this is working as intended?)

I'll leave this for now, in case there's more feedback.

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
```diff
@@ -34,14 +36,47 @@ const sortOptions: SortOptions = [
...
+  if (element === null) return false
+
+  let { top } = element.getBoundingClientRect()
+  let viewportH = window.innerHeight || document.documentElement.clientHeight
```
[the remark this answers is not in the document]
anandaroop: 039e5d97bd89b8fa80588a2468533b162061969f

src/Apps/CollectorProfile/Routes/Saves2/Utils/useArtworkListVisibility.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const ArtworkListVisibilityContext = createContext<ArtworkListVisibility>({
+  touched: false,
+  setTouched: (_value: boolean) => {},
+  scrolled: false,
+  setScrolled: (_value: boolean) => {},
+})
+
+export const ArtworkListVisibilityProvider: React.FC = ({ children }) => {
+  const [touched, setTouched] = useState<boolean>(false)
```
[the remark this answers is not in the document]
anandaroop: 1a306cc04e40fc991a3426edadd9ca87a31a0719

anandaroop (commented):
Addressed most feedback. Haven't figured out a good testing strategy for this viewport/scrolly stuff, but I'm taking suggestions 😅. I'll leave open a bit longer before merging.

src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListContent.tsx
dimatretyak: > I think it would be a good idea to prevent that auto-scroll behavior until a user has interacted with the rail

I think Devon meant not to do scrolling on the initial render (when the user just navigated to the screen, but did not click on any artwork list card in the rail.) But every time when user clicks on the artwork list card in the rail, we should do scrolling

@dblandin is it correct? Did I understand you correctly?

anandaroop: I guess I had in mind the kind of accessibility advice that cautions against  [excessive movement that is not controlled by the user](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html), but in fact: this interaction does not violate that, and arguably is controlled by the user. So maybe I'm being too cautious here.

I'll revert the `!artworkListHasBeenScrolled` requirement in https://github.com/artsy/force/commit/50ad5746630ac5b0c134368888667e23c0c767db, and then merge 👍🏽

anandaroop: Doh! I forgot to actually commit that revert, added via follow-up PR in #12517

PR #12507, opened 2023-06-07
chore(TX-1229): add error message tracking for review route

rquartararo:

This PR solves [TX-1229]

### Description

This PR adds the new `errorMessageViewed` [event](https://github.com/artsy/cohesion/pull/431) to the review order route whenever a user views an error dialog.

[TX-1229]: https://artsyproduct.atlassian.net/browse/TX-1229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Review/index.tsx
+++ b/src/Apps/Order/Routes/Review/index.tsx
@@ -58,6 +58,22 @@ export const ReviewRoute: FC<ReviewProps> = props => {
   const productId = extractNodes(props.order.lineItems)[0].artwork?.internalID
   const artworkVersion = extractNodes(props.order.lineItems)[0]?.artworkVersion
 
+  const trackErrorMessageEvent = (
+    title: string,
+    message: string | undefined,
+    code?: string
+  ) => {
+    return trackEvent({
+      action: ActionType.errorMessageViewed,
+      context_owner_type: OwnerType.ordersReview,
+      context_owner_id: props.order.internalID,
+      title: title,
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Review/index.tsx
```diff
@@ -93,6 +109,8 @@ export const ReviewRoute: FC<ReviewProps> = props => {
           .handleCardAction(orderOrError.actionData.clientSecret)
           .then(result => {
             if (result.error) {
+              trackErrorMessageEvent("An error occurred", result.error.message)
```
iskounen, line 112: [`handleCardAction`](https://stripe.com/docs/js/payment_intents/handle_card_action) is a Stripe API call, which is expected to return an error code in addition to an error message, so I think we can populate that here too.

(The same goes for the [`confirmCardSetup`](https://stripe.com/docs/js/setup_intents/confirm_card_setup) call below.)

iskounen (commented):
Thank you for doing this work. I didn't realize how many things could go wrong on the review page.

I have one suggestion about error codes, but otherwise this looks good.

src/Apps/Order/Routes/Review/index.tsx
rquartararo: Good catch will add that thanks!

PR #12509, opened 2023-06-08
feat: Show only active hero units by default GRO-1755

jonallured:
This PR sits on top of https://github.com/artsy/metaphysics/pull/5072 and changes the default behavior of the homepage to show only active hero units by default. This should ease confusion around why inactive hero units are showing up for admins.

Then I also added support for a query param called `show_all` and when this is set to true, then a signed in admin can preview any inactive hero units that they are working on.

https://artsyproduct.atlassian.net/browse/GRO-1755

/cc @artsy/grow-devs


dzucconi:
Is there an actual need to support this?

jonallured:
Yes the reason to do this is otherwise admins have no way of verifying their hero unit work.

PR #12510, opened 2023-06-08
feat(cookie-consent): implements cookie consent manager

dzucconi:
Alright, take two. 🎬
Original PR: https://github.com/artsy/force/pull/12280

To summarize: the problem I ran into, and why this was reverted, was that, previously, we were conditionally returning out a "different" set of children, sans context. This somehow broke routing: links would work fine but it was pressing browser back would change the URL without actually navigating the page. Here I just thread children through unconditionally and then conditionally render only the banner/dialog.

This PR _only_ adds the code, and does not yet enable it. So it's safe to merge. I'll make a separate PR with a review app again to enable it. (Review app because I'll be adding some specs covering basic back/forward navigation to ensure routing is OK and won't merge until everything is green, flaky or not).

### DIFF
--- a/src/Components/CookieConsentManager/CookieConsentManager.tsx
+++ b/src/Components/CookieConsentManager/CookieConsentManager.tsx
@@ -1,0 +1,119 @@
+import { FC, ReactNode } from "react"
+import { ConsentManagerBuilder } from "@segment/consent-manager"
+import { getENV } from "Utils/getENV"
+import { useDidMount } from "@artsy/palette"
+import {
+  CUSTOM_DESTINATIONS,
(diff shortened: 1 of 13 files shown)


src/Components/CookieConsentManager/CookieConsentManager.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const handleClose = () => {
+    setMode("Idle")
+  }
+
+  return (
+    <CookieConsentManagerProvider openConsentManager={handleManage}>
+      {children}
+
+      {isMounted && (
```
dzucconi, line 50: This is the only real change.

PR #12515, opened 2023-06-09
chore(FX-4718): it's hard to click on the buttons if the toast message is displayed

dimatretyak:

Review app: [improvement-for-toast](https://improvement-for-toast.artsy.net/)

This PR solves [FX-4718]

### Description
There may be situations when it will be difficult to click on the buttons if the toast message is displayed. The problem is due to the fact that toast messages have paddings that block clicks. If the button is in this zone, the click will not be processed

image

### Demo
| Before | After |
| <video src="https://github.com/artsy/force/assets/3513494/77463d51-8d10-48f7-817f-a465faaabb03" /> | <video src="https://github.com/artsy/force/assets/3513494/f81ad660-125d-4ee2-9488-9b2f82ca526c" /> | 

[FX-4718]: https://artsyproduct.atlassian.net/browse/FX-4718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Components/AppToasts.tsx
+++ b/src/Apps/Components/AppToasts.tsx
@@ -20,7 +20,9 @@ export const AppToasts: React.FC = () => {
       <AppContainer px={[2, 4]} py={4}>
         <GridColumns>
           <Column span={4} start={5}>
-            <Toasts />
+            <Box style={{ pointerEvents: "auto" }}>
+              <Toasts />
+            </Box>
           </Column>
         </GridColumns>
       </AppContainer>
--- a/src/Apps/Components/AppToasts.tsx
+++ b/src/Apps/Components/AppToasts.tsx
@@ -9,7 +9,14 @@ export const AppToasts: React.FC = () => {
   if (toasts.length === 0) return null
 
   return (
-    <Box position="fixed" zIndex={Z.toasts} bottom={0} left={0} width="100%">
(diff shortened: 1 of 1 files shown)


src/Apps/Components/AppToasts.tsx
```diff
@@ -9,7 +9,14 @@ export const AppToasts: React.FC = () => {
...
 
   return (
-    <Box position="fixed" zIndex={Z.toasts} bottom={0} left={0} width="100%">
+    <Box
+      position="fixed"
+      zIndex={Z.toasts}
+      bottom={0}
+      left={0}
+      width="100%"
+      style={{ pointerEvents: "none" }}
```
dimatretyak, line 18: if we add this, then clicking on toast (for example, to hide the toast message) will be ignored. for this reason, this solution is not suitable

dimatretyak: fixed [here](https://github.com/artsy/force/pull/12515/commits/cd4b389e218d5226900900e81e27c529a3437f46) by using another `pointer-events` for children content

anandaroop (approved):
Works for me 👍🏽

PR #12522, opened 2023-06-12
fix(CX-3780): show artwork image inside my collection when gemini processing fails

MounirDhahri:

This PR solves [CX-3780]

### Description
We are currently saving the large version of the image as image url assuming the latter always exists. This is not the case though when Gemini fails. We are updating the logic now to parse the available image versions and use the first biggest image available.

**Next step:** Introduce the same fix to **Eigen**

| Before                                                                              | After                                                                               |
|-------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|
| https://github.com/artsy/force/assets/11945712/4941487c-c6b6-4883-8ca8-aa391f52f7c2 | https://github.com/artsy/force/assets/11945712/3fa28e9c-5fd6-45a8-bf62-a51857076eca |

Thanks for reporting this @pvinis 

[CX-3780]: https://artsyproduct.atlassian.net/browse/CX-3780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtworkStep.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtworkStep.tsx
@@ -58,24 +58,9 @@ export const MyCollectionArtworkFormArtworkStep: React.FC<MyCollectionArtworkFor
     // Initialize photos with artwork images
 
     const photos = artwork.images?.map(image => {
-      const imageVersionsSortedBySize = [
-        "main",
-        "normalized",
-        "larger",
-        "large",
-        "medium",
-        "small",
-      ]
-
-      const imageVersion = imageVersionsSortedBySize.find(version =>
-        image?.imageVersions?.includes(version)
-      )
-
       return {
         name: "Automatically added",
-        url: !!imageVersion
-          ? image?.imageURL?.replace(":version", imageVersion)
-          : null,
+        url: image.url,
         id: image?.internalID,
         size: image?.width,
       }
--- a/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtworkStep.tsx
+++ b/src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtworkStep.tsx
@@ -57,12 +57,29 @@ export const MyCollectionArtworkFormArtworkStep: React.FC<MyCollectionArtworkFor
 
     // Initialize photos with artwork images
 
-    const photos = artwork.images?.map(image => ({
-      name: "Automatically added",
-      url: image?.imageURL?.replace(":version", "large"),
-      id: image?.internalID,
-      size: image?.width,
-    }))
+    const photos = artwork.images?.map(image => {
+      const imageVersionsSortedBySize = [
+        "main",
+        "normalized",
+        "larger",
+        "large",
+        "medium",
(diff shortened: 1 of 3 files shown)


dariakoko (approved):

src/Apps/MyCollection/Routes/EditArtwork/Components/MyCollectionArtworkFormArtworkStep.tsx
```diff
@@ -57,12 +57,29 @@ export const MyCollectionArtworkFormArtworkStep: React.FC<MyCollectionArtworkFor
...
+
+      const imageVersion = imageVersionsSortedBySize.find(version =>
+        image?.imageVersions?.includes(version)
+      )
+
+      return {
+        name: "Automatically added",
+        url: !!imageVersion
+          ? image?.imageURL?.replace(":version", imageVersion)
+          : null,
```
olerichter00, line 78: Could we pass the list of versions directly in the GrqphQL quer instead ([example](https://github.com/artsy/force/blob/b64d9edf38df65a3a2733799771cd2f09c131e4d/src/Apps/Article/Components/Sections/ArticleSectionImageCollection/ArticleSectionImageCollectionImage.tsx#L66))? I think this case should be supported 🤔

MounirDhahri: didn't know we can do that! does this pattern also return the first one available in the provided array?

olerichter00: Yes, I think it does exactly what we need. It goes through `imageVersions` and looks for the first version that exists ([code](https://github.com/artsy/metaphysics/blob/eb9a1362323222a5bd22089c4b798f1442af4a89/src/schema/v2/image/normalize.ts#L24)).

You should be able to get the image URL like this:

```graphql
image {
  url(
    version: [
        "main",
        "normalized",
        "larger",
        "large",
        "medium",
        "small"
    ]
  )
}
```

MounirDhahri: great, thanks @olerichter00. I will update

MounirDhahri: it seems like the version string is not supported within the Image interface in MP. I don't believe it makes sense to add support to it in this context, what do you think?

olerichter00: I updated the query in `MyCollectionArworkSearch.tsx` and it appears to be working correctly.
```
images {
                  height
                  internalID
                  isDefault
                  url(version: ["large", "tall"])
                  width
                }
```

MounirDhahri: I used `imageUrl` initially , i can try this out as well 👀

MounirDhahri: worked! thanks

PR #12530, opened 2023-06-14
fix(FX-4730): fixes incorrect validation message when creating / updating artwork lists

nickskalkin:
[Jira ticket](https://artsyproduct.atlassian.net/browse/FX-4730)

This PR solves [FX-4730]

### Description

- Aligns error message handling between Create / Edit flows
- Uses correct metaphysics field for displaying error message

### Demo

https://github.com/artsy/force/assets/3934579/8729f9a1-c0f4-4d29-8331-6a1101595573

[FX-4730]: https://artsyproduct.atlassian.net/browse/FX-4730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
@@ -73,9 +73,9 @@ export const CreateNewListModal: React.FC<CreateNewListModalProps> = ({
         },
         rejectIf: response => {
           const result = response.createCollection?.responseOrError
-          const errorMessage = result?.mutationError?.message
+          const error = result?.mutationError
 
-          return !!errorMessage
+          return error
         },
       })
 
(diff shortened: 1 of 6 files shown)


src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
```diff
@@ -90,10 +90,19 @@ export const CreateNewListModal: React.FC<CreateNewListModalProps> = ({
       trackAnalyticEvent(artworkListId)
     } catch (error) {
       logger.error(error)
-      helpers.setFieldError(
-        "name",
-        error.message ?? t("common.errors.somethingWentWrong")
+
+      // use generic error message by default
+      let errorMessage = t("common.errors.somethingWentWrong")
```
anandaroop, line 95: **question (non-blocking):** In the case where there's a top-level error instead of field-level, we'd always show this generic error message, right? 

Do we feel the need to display a specific error message if present? 

(Thinking of the top-level errors that we might receive [here for example](https://github.com/artsy/metaphysics/pull/5084/files#diff-c5ff3880188b0505c2138083d5f374a8651de974328c22385c0ba50eeecb6844R95) — it may be the case that this is not a relevant concern in these modals, in which case fine 👍🏽 )

anandaroop (approved):
Looks good, just one q…

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModal.tsx
nickskalkin: > In the case where there's a top-level error instead of field-level, we'd always show this generic error message, right?

yes, you are right.

For now, I would avoid exposing any unmoderated exceptions coming from Gravity. I've made a few experiments sending requests that fail with different kinds of errors, and it seems for me that sometimes the exposed error messages are too verbose for a user:

For instance, this is what Gravity returns when random exception occures:

```
  "body": {
    "type": "other_error",
    "message": "exception class/object expected",
    "backtrace": [
      "..."
    ]
  }
```

PR #12531, opened 2023-06-14
chore: Update empty message for no auction results GRO-1747

jonallured:
Very minor update to the copy on the auctions tab when there are no results.

### before

Screenshot 2023-06-14 at 3 12 37 PM

### after

Screenshot 2023-06-14 at 3 12 40 PM

https://artsyproduct.atlassian.net/browse/GRO-1747

/cc @artsy/grow-devs


kajatiger:
Looks good! I don't understand why the smoke test is failing and am rerunning it to see if this works. 🤷

jonallured:
Ok I rebased on top of latest main which includes the timeout that @pam- added so I think this should go green!

PR #12536, opened 2023-06-15
fix(artist): fixes list logic

dzucconi:
Saw this one [come up in Slack.](https://artsy.slack.com/archives/C03N12SR0RK/p1686797097773909)

Fixes the list logic to insert the `and` in the correct location when the list is expanded.


dzucconi:
cc @jpoczik

TMcMeans:
funny enough I believe someone wanted the "and" to be in the weird order before so that why it was built that way. Either way, nice quick fix!

PR #12540, opened 2023-06-16
feat: add error code to the review step's dead-end event

iskounen:
A tracking event is fired when buyers encounter a generic "Something went wrong..." error message while submitting orders on the Review step. This event doesn't currently include an error code property because we weren't sure whether one would be available when this error occurs.

This PR adds an optional error code property to these "dead-end" events so that we can see if any information is available to help us better map occurrences of these errors to their causes.


rquartararo (approved):
Looks good thanks for adding this 👍

PR #12541, opened 2023-06-16
fix: artsy shipping quote selection

leamotta:

### Description

This PR intends to solve 2 issues:
- Requesting new quotes whenever the price changes if the collector decides to go back to the offer step and make a lower or higher value offer.
Before:

https://github.com/artsy/force/assets/7518671/49a3baef-d346-4a55-8aaf-1f091bf2016e

After:

https://github.com/artsy/force/assets/7518671/2ab6e424-16d6-4be6-8eef-2bf4dc3b40bf

- The broken behaviour when toggling between pickup and shipping, caused by a faulty check on `onSelectShippingOption` where the code assumed that the shipping option would be changed after the [setShippingOption](https://github.com/artsy/force/compare/leamotta/artsy-shipping-effect?expand=1#diff-e0afcd1142a7fdd8a803a66f3f2490eb8db1d25fe9d66bd889fefd4852937ec6R553) line. However this is not the case and I solved it by using useEffect.
Before:

https://github.com/artsy/force/assets/7518671/31be6f00-0064-4dcc-ad60-94dc420a6c37

After:

https://github.com/artsy/force/assets/7518671/3b415a0d-2efe-4c99-8afb-a47fe38d5519

Props @gkartalis for teaching me about useEffect.

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -182,7 +182,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
   }
 
   useEffect(() => {
-    if (checkIfArtsyShipping() && !isCreateNewAddress() && !shippingQuoteId) {
+    if (checkIfArtsyShipping() && !isCreateNewAddress()) {
       selectShipping()
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -530,6 +524,18 @@ export const ShippingRoute: FC<ShippingProps> = props => {
     setShippingQuoteId(undefined)
   }
 
+  useEffect(() => {
+    if (
+      addressList &&
+      addressList.length > 0 &&
```
AlicanAkyuz, line 530: **non-blocking**: you could achieve these 2 statements with `addressList?.length`

AlicanAkyuz (approved):
LGTM! thank you for the fix.

PR #12543, opened 2023-06-20
feat(FX-4804): Confirmation step in Create Alert modal flow

nickskalkin:

This PR solves [FX-4804]

### Description

- [Review app](https://alerts-confirmation-step.artsy.net/)

https://github.com/artsy/force/assets/3934579/c8273d0e-0da7-450a-bceb-117df1b25a26

[FX-4804]: https://artsyproduct.atlassian.net/browse/FX-4804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -30,7 +30,6 @@ interface ArtworkGridItemProps extends React.HTMLAttributes<HTMLDivElement> {
   to?: string | null
   savedListId?: string
   renderSaveButton?: (artworkId: string) => React.ReactNode
-  saveOnlyToDefaultList?: boolean
 }
 
 export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
@@ -46,7 +45,6 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
   to,
   savedListId,
   renderSaveButton,
-  saveOnlyToDefaultList,
   ...rest
 }) => {
   const localImage = useLocalImage(artwork.image)
@@ -122,7 +120,6 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
           disableRouterLinking={disableRouterLinking}
           to={to}
           renderSaveButton={renderSaveButton}
-          saveOnlyToDefaultList={saveOnlyToDefaultList}
         />
       </div>
     </ManageArtworkForSavesProvider>
--- a/src/Components/Artwork/GridItem.tsx
+++ b/src/Components/Artwork/GridItem.tsx
@@ -30,6 +30,7 @@ interface ArtworkGridItemProps extends React.HTMLAttributes<HTMLDivElement> {
   to?: string | null
   savedListId?: string
   renderSaveButton?: (artworkId: string) => React.ReactNode
+  saveOnlyToDefaultList?: boolean
 }
 
 export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
@@ -45,6 +46,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
   to,
   savedListId,
   renderSaveButton,
+  saveOnlyToDefaultList,
   ...rest
 }) => {
   const localImage = useLocalImage(artwork.image)
@@ -120,6 +122,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
           disableRouterLinking={disableRouterLinking}
           to={to}
           renderSaveButton={renderSaveButton}
+          saveOnlyToDefaultList={saveOnlyToDefaultList}
         />
       </div>
     </ManageArtworkForSavesProvider>
--- a/src/Components/Artwork/Metadata.tsx
+++ b/src/Components/Artwork/Metadata.tsx
@@ -25,6 +25,7 @@ export interface MetadataProps
   showSaveButton?: boolean
   to?: string | null
   renderSaveButton?: (artworkId: string) => React.ReactNode
+  saveOnlyToDefaultList?: boolean
 }
 
 export const Metadata: React.FC<MetadataProps> = ({
@@ -40,6 +41,7 @@ export const Metadata: React.FC<MetadataProps> = ({
   showHoverDetails,
   showSaveButton,
   renderSaveButton,
+  saveOnlyToDefaultList,
   ...rest
 }) => {
   return (
@@ -61,6 +63,7 @@ export const Metadata: React.FC<MetadataProps> = ({
         showSaveButton={showSaveButton}
         contextModule={contextModule}
         renderSaveButton={renderSaveButton}
+        saveOnlyToDefaultList={saveOnlyToDefaultList}
       />
     </LinkContainer>
   )
--- a/src/Components/Artwork/Metadata.tsx
+++ b/src/Components/Artwork/Metadata.tsx
@@ -25,7 +25,6 @@ export interface MetadataProps
   showSaveButton?: boolean
   to?: string | null
   renderSaveButton?: (artworkId: string) => React.ReactNode
-  saveOnlyToDefaultList?: boolean
 }
 
 export const Metadata: React.FC<MetadataProps> = ({
@@ -41,7 +40,6 @@ export const Metadata: React.FC<MetadataProps> = ({
   showHoverDetails,
   showSaveButton,
   renderSaveButton,
-  saveOnlyToDefaultList,
   ...rest
 }) => {
   return (
@@ -63,7 +61,6 @@ export const Metadata: React.FC<MetadataProps> = ({
         showSaveButton={showSaveButton}
         contextModule={contextModule}
         renderSaveButton={renderSaveButton}
-        saveOnlyToDefaultList={saveOnlyToDefaultList}
       />
     </LinkContainer>
   )
--- a/src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
+++ b/src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
@@ -8,7 +8,7 @@ export const ConfirmationModalHeader: FC = () => {
   const { pills } = useSavedSearchAlertContext()
 
   return (
-    <Flex flexDirection="column" mt={-2}>
+    <Flex flexDirection="column">
       <Join separator={<Spacer y={2} />}>
         <Text variant="sm-display" color="black60">
           {t("createAlertModal.confirmationStep.weWillNotifyYou")}
--- a/src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
+++ b/src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
@@ -1,0 +1,22 @@
+import { Text, Flex, Join, Separator, Spacer } from "@artsy/palette"
+import { SavedSearchAlertPills } from "Components/SavedSearchAlert/Components/SavedSearchAlertPills"
+import { useSavedSearchAlertContext } from "Components/SavedSearchAlert/SavedSearchAlertContext"
+import { FC } from "react"
+
+export const ConfirmationModalHeader: FC = () => {
+  const { pills } = useSavedSearchAlertContext()
+
+  return (
+    <Flex flexDirection="column" mt={-2}>
+      <Join separator={<Spacer y={2} />}>
+        <Text variant="sm-display" color="black60">
+          We will notify you when new works get added to Artsy.
+        </Text>
+        <Flex flexWrap="wrap" mx={-0.5}>
+          <SavedSearchAlertPills items={pills} onDeletePress={() => {}} />
+        </Flex>
+        <Separator />
+      </Join>
+    </Flex>
+  )
+}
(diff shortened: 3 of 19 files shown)


src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useSavedSearchAlertContext } from "Components/SavedSearchAlert/SavedSearchAlertContext"
+import { FC } from "react"
+import { useTranslation } from "react-i18next"
+
+export const ConfirmationModalHeader: FC = () => {
+  const { t } = useTranslation()
+  const { pills } = useSavedSearchAlertContext()
+
+  return (
+    <Flex flexDirection="column" mt={-2}>
```
nickskalkin, line 11: I know that we are against not following palette-driven conventions, but still wanted to ask if this change is OK.

| With mt (as in design specs) | Without mt |
|---|---|
|  with_top_margin | without_top_margin |

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useTranslation()
+  const savedSearch = me?.savedSearch
+
+  return (
+    <Flex flexDirection={["column", "row"]}>
+      <Button
+        width="100%"
+        mr={[0, 1]}
+        mb={[1, 0]}
+        // @ts-ignore
```
nickskalkin, line 26: Is there a better way to do it without `ts-ignore`? I've took inspiration from here: https://github.com/artsy/force/blob/main/src/Apps/CollectorProfile/Routes/Saves2/Components/ArtworkListEmptyState.tsx#L36

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface ConfirmationStepFooterQueryRendererProps {
+  searchCriteriaId: string
+}
+
+export const ConfirmationStepFooterQueryRenderer: FC<ConfirmationStepFooterQueryRendererProps> = props => {
+  return (
+    <SystemQueryRenderer<ConfirmationStepFooterQuery>
+      placeholder={<ContentPlaceholder />}
+      // Temporary workaround internalID is requested because there is a bug in Metaphysics. If a user's field is not requested, the
+      // query returns null for savedSearch.
```
nickskalkin, line 56: I will create a bug ticket for this

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
```diff
@@ -0,0 +1,104 @@
...
+            counts {
+              total
+            }
+            ...ArtworkGrid_artworks
+          }
+        }
+      `}
+      variables={{
+        input: {
+          first: 10,
```
nickskalkin, line 59: in ticket:

> 10 works for all devices except tablet, in which case we’d display 20.

Couldn't find anything similar to it in the codebase. What would be a correct way to determine if user uses tablet?

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
```diff
@@ -0,0 +1,104 @@
...
+          artworksCount,
+        })}
+      </Text>
+      <Text variant="sm-display" color="black60">
+        {t("createAlertModal.confirmationStep.seeOurTopPicks")}
+      </Text>
+
+      <Spacer y={2} />
+
+      <ArtworkGrid artworks={artworksConnection!} columnCount={2} />
```
nickskalkin, line 38: I might need your help here. Grid doesn't render correctly in the confirmation modal for some reason (mWeb). I cannot figure out how to fix it properly

https://github.com/artsy/force/assets/3934579/272dc1f3-4e80-4358-9770-d532ac9ca2d9

src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
dzucconi: It appears that there's something else going on. There's only a 20px padding on that title to begin with. So I am still against this change, it just looks like you need to remove a margin elsewhere.

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+      }}
+    />
+  )
+}
+
+const ContentPlaceholder: FC = () => {
+  return (
+    <Flex flexDirection={["column", "row"]}>
+      <SkeletonBox height={50} width="100%" mr={[0, 1]} />
+      <SkeletonBox height={50} width="100%" ml={[0, 1]} />
```
dzucconi, line 97: If they are buttons then just use disabled buttons with the appropriate labels.

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
```diff
@@ -0,0 +1,93 @@
...
+const ConfirmationStepFooter: FC<ConfirmationStepFooterProps> = ({ me }) => {
+  const { t } = useTranslation()
+  const savedSearch = me?.savedSearch
+
+  return (
+    <Flex flexDirection={["column", "row"]}>
+      <Button
+        width="100%"
+        mr={[0, 1]}
+        mb={[1, 0]}
```
dzucconi, line 35: Just use `gap` on flex instead of these margins

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
dzucconi: A workaround for what?

dzucconi: Oh, probably the stitched field. I've run into this before.

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
dzucconi: That component is a bit of a mess and maybe overkill. Can you just use the `Masonry` component?

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
dzucconi: Not a thing. Not sure what would make tablet's different than desktop :)

dzucconi: I mean you *could* detect a combination of screen dimensions and touch. But we shouldn't, I think the design is just specified oddly.

src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
nickskalkin: I think title has m={2} ([code](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx#L63)) and header has p={2} ([code](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/ModalDialog/ModalDialogContent.tsx#L83)), which gives 40 in total.

Going to remove `mt={-2}` though, maybe padding issue will be fixed in palette somewhen?

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
nickskalkin: I've just remembered that this is not a requirement from designer, rather something we came up with during eng refinement. We though that 10 artworks won't fill modal fully on tables in vertical mode. I've just checked on my ipad and all looks fine, so let's use 10 for all devices

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
nickskalkin: Masonry has the same problem, unfortunately..

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
nickskalkin: TIL gap 👍

src/Components/SavedSearchAlert/Components/ConfirmationModalHeader.tsx
dzucconi: Ah, yep. I did not know this `header` prop existed. Definitely incorrect. I will fix today.

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
dzucconi: It's *so* useful!

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
nickskalkin: [Temporarily changed](https://github.com/artsy/force/pull/12543/commits/ec582e8748d78601d27ed65f5057d2ace4ae809d) to GridColumns which works fine, but completely differs from design specs. 

Still looking for a proper solution for the ArtworksGrid

dzucconi: I can take a look at this today as well

src/Components/Artwork/GridItem.tsx
```diff
@@ -120,6 +122,7 @@ export const ArtworkGridItem: React.FC<ArtworkGridItemProps> = ({
           disableRouterLinking={disableRouterLinking}
           to={to}
           renderSaveButton={renderSaveButton}
+          saveOnlyToDefaultList={saveOnlyToDefaultList}
```
dzucconi, line 125: Really wish we just set up a simple artwork grid provider to avoid the props drilling

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
```diff
@@ -0,0 +1,101 @@
...
+  return (
+    <Flex flexDirection={["column", "row"]} gap={1}>
+      <Button width="100%" onClick={() => redirect(savedSearch?.href!)}>
+        {t("createAlertModal.confirmationStep.seeAllMatchingWorks")}
+      </Button>
+
+      <Button
+        width="100%"
+        variant="secondaryBlack"
+        onClick={() => redirect("/settings/alerts")}
```
dzucconi, line 39: These just be RouterLinks using the `as` prop

nickskalkin:
Should be ready for review, the only un-resolved issue is this one: https://github.com/artsy/force/pull/12543#discussion_r1235267245

nickskalkin:
I would make header section non-sticky. If there are many selected filters, scrollable area is too small:

src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx
anandaroop: @nickskalkin I take it that the answer was that there isn't a better way right now? (I was [also wondering](https://artsy.slack.com/archives/CP9P4KR35/p1687547580653299?thread_ts=1687543308.378649&cid=CP9P4KR35) about `<Button as={…} />` recently)

src/Components/Artwork/Metadata.tsx
```diff
@@ -25,6 +25,7 @@ export interface MetadataProps
   showSaveButton?: boolean
   to?: string | null
   renderSaveButton?: (artworkId: string) => React.ReactNode
+  saveOnlyToDefaultList?: boolean
```
damassi, line 28: Can avoid all of the passing via https://github.com/artsy/force/blob/c9a511068a3870c78756eb80fc877f93eccc5622/src/Components/ArtworkGrid/ArtworkGridContext.tsx#L21

anandaroop (dismissed):
Looks good to me 👍🏽 

General question: do we want to track interactions with the new modal? (Can be a follow-up if needed)

anandaroop:
One more observation from the review app —

Not sure this was specified in the design, but is this the behavior we want for clicks on the artwork grid in the modal?

Or do we want to dismiss the modal? (If so maybe we can use [the onClick callback](https://github.com/artsy/force/pull/12543/files#diff-b8d8f73a39d3fd3a331a6c98de65d7a4147e427847ea0991e3d599b84efd8542R42) for that)

https://github.com/artsy/force/assets/140521/05443449-6f10-4c7a-96f7-705e65ea9091

nickskalkin:
@anandaroop thanks, I think def should close the modal, I overlooked 😄

nickskalkin:
@anandaroop should be fixed now!

src/Components/Artwork/Metadata.tsx
nickskalkin: thanks @damassi ! Fixed

anandaroop (approved):

src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
nickskalkin: btw @gkartalis pointed to how Masonry is used together with GridColumns: https://github.com/artsy/force/blob/2f7217c3daa9bb8f2d8164408384d3ceec5cb787/src/Apps/Artist/Routes/Consign/Components/ArtistConsignFAQ.tsx#L26

I've re-used the same approach and it worked! Going to read about css grids this weekend...

and props to @gkartalis ❤️

PR #12552, opened 2023-06-22
fix: upgrades palette, fixes modal dialog headers

dzucconi:
https://github.com/artsy/palette/pull/1302 removed the doubled up spacing. There was some hacks around it. This removes those hacks and includes some minor clean up.


src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModalHeader.tsx
```diff
@@ -1,21 +0,0 @@
...
-export const CreateNewListModalHeader: FC<CreateNewListModalHeaderProps> = ({
-  artwork,
-}) => {
-  return (
-    <>
-      <ArtworkModalHeaderInfo artwork={artwork} />
-      <Spacer y={2} />
-    </>
-  )
-}
```
dzucconi, line 21: I have no clue why this wrapper component existed.

src/Apps/CollectorProfile/Routes/Saves2/Components/SelectArtworkListsModal/SelectArtworkListsHeader.tsx
```diff
@@ -38,23 +38,22 @@ export const SelectArtworkListsHeader: FC = () => {
...
 
-      <Spacer y={2} />
-
       {state.recentlyAddedList && (
-        <Message
-          variant="success"
-          title={t(
-            "collectorSaves.selectArtworkLists.header.recentlyAddedList.title"
-          )}
-          mx={-2}
```
dzucconi, line 49: I would expect a negative margin to have a comment explaining why it exists, since it's by default working around something that probably exists for a good reason.

dzucconi:
Looking at your PR @nickskalkin it does seem like a reasonable usage for that screen size. So, I guess I can get https://github.com/artsy/palette/pull/1302 merged. I can just restore these to use the `header` instead of fighting to get rid of it. 🤷

nickskalkin:
@dzucconi yeah I would vote to merge palette's PR instead. I'm not comfortable making decisions like making artwork info not sticky without discussing it with Ryan first.

src/Apps/CollectorProfile/Routes/Saves2/Components/SelectArtworkListsModal/SelectArtworkListsHeader.tsx
nickskalkin: Could this be to open modal full-screen?

| Before | Your changes |
|---|---|
| Screenshot 2023-06-22 at 22 05 33 | Screenshot 2023-06-22 at 22 05 44 |

src/Apps/CollectorProfile/Routes/Saves2/Components/CreateNewListModal/CreateNewListModalHeader.tsx
nickskalkin: There were a bit more logic before, then it got trimmed and we've decided to leave this component be

dzucconi: Yeah, having the external margin dangling off the end here is a no go

src/Apps/CollectorProfile/Routes/Saves2/Components/SelectArtworkListsModal/SelectArtworkListsHeader.tsx
dzucconi: This is a different margin, actually. This one is for `Message`.

dzucconi: Though, it seems to me that external dialog margins are not something that should be customized. If there's a problem we should change it in Palette.

dzucconi:
K, this should be fine to merge now.

PR #12557, opened 2023-06-26
fix: sort past shows by end date

Mitch-Henson:

### Description

This PR comes off the back of a complaint where the past shows rail is showing only a single gallery even though there are multiple galleries in the specific city. As the results are paginated and sorted by gallery name, they are very one sided when a gallery has a large number of closed events.

The fix means we now sort by end date and should mean that galleries are more varied, especially when certain galleries are no longer with Artsy, yet still have a large amount of past shows and a name which was favoured for sorting.

### DIFF
--- a/src/Apps/Shows/Routes/ShowsCity.tsx
+++ b/src/Apps/Shows/Routes/ShowsCity.tsx
@@ -281,7 +281,11 @@ export const ShowsCityRefetchContainer = createRefetchContainer(
             }
           }
         }
-        pastShows: showsConnection(first: 18, status: CLOSED) {
+        pastShows: showsConnection(
+          first: 18
+          status: CLOSED
+          sort: END_AT_DESC
+        ) {
           edges {
             node {
               internalID
--- a/src/__generated__/ShowsCityQuery.graphql.ts
+++ b/src/__generated__/ShowsCityQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<fe54b9babd3dd4f4b0329be3890566a2>>
+ * @generated SignedSource<<0cf3d6cfb3b96cd38a5469614c2f124e>>
  * @lightSyntaxTransform
  * @nogrep
  */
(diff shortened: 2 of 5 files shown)


leamotta (approved):
LGTM! Tagging a few FX friends to get their +1 since they technically own the page for now

nickskalkin (approved):
👍

dblandin:
Looks good to me but it might be useful to cross-link to a Jira ticket and/or Slack thread where this was discussed so that context isn't lost.

Mitch-Henson:
Good idea @dblandin. The slack thread is [here](https://artsy.slack.com/archives/C9ZJYFDNF/p1687269449631419)

Mitch-Henson:
@dblandin I believe that it gets the sort option from the [sort_order method](https://github.com/artsy/gravity/blob/8e34dcd62d798400aa259e4bcf41b9152e252e96/app/api/util/sort_helpers.rb#L7) which is called in the [paginate_and_sort method](https://github.com/artsy/gravity/blob/8e34dcd62d798400aa259e4bcf41b9152e252e96/app/api/util/misc_helpers.rb#L15).

Happy to make further changes to the page, though I'd want to double check any changes with a designer. As of Wednesday, our team no longer has a designer so I will speak with Joh to see what we can do!

dblandin:
> I believe that it gets the sort option from the [sort_order method](https://github.com/artsy/gravity/blob/8e34dcd62d798400aa259e4bcf41b9152e252e96/app/api/util/sort_helpers.rb#L7) which is called in the [paginate_and_sort method](https://github.com/artsy/gravity/blob/8e34dcd62d798400aa259e4bcf41b9152e252e96/app/api/util/misc_helpers.rb#L15).

That does look like where an explicit `sort` option is passed in, but because the previous behavior did not apply an explicit sort, I'm still wondering what the _default_ sort would have been. Seems unlikely that it would have been Gallery Name without that being specified somewhere and instead might have defaulted to a Rails/Mongoid/MongoDB default ([example](https://www.mongodb.com/docs/manual/reference/method/cursor.sort/#return-in-natural-order)) or something similar.

Not super important to have 100% clarity on but I thought it was strange to hear that it was sorting by gallery name without being able to see where that was specified.

PR #12558, opened 2023-06-26
chore(NX-3560): Setup Auction Support Chat (Salesforce)

lidimayra:

This PR solves [NX-3560]

### Description

When displaying the chat bubble, in case it is an auction, display the chat bubble that points to Auction Support instead of Collector Support.

I also took the chance to completely removed the Zendesk references, but I'm submitting the cleanup on a separate PR to avoid it affecting the progress of the Chat Bubble for Actions, which has a higher priority.

cc @artsy/negotiate-devs 

[NX-3560]: https://artsyproduct.atlassian.net/browse/NX-3560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/SalesforceWrapper.tsx
```diff
@@ -33,6 +57,7 @@ export const SalesforceWrapper: React.FC = () => {
     )}/embeddedservice/5.0/esw.min.js`,
     removeOnUnmount: true,
     onReady: () => {
+      window.embedded_svc.settings.defaultMinimizedText = "Chat"
```
lidimayra, line 60: Taking the chance to handle this here, but it is actually part of [NX-3562](https://artsyproduct.atlassian.net/browse/NX-3562)

(I checked with Allie, and there is no problem is releasing the `Chat with expert` > `Chat` changes in different dates for the different apps)

src/Server/setup_sharify.ts
```diff
@@ -80,10 +80,13 @@ sharify.data = _.extend(
     "SALESFORCE_CHAT_INSTANCE_URL",
     "SALESFORCE_CHAT_HELP_URL",
     "SALESFORCE_CHAT_ORG_ID",
-    "SALESFORCE_CHAT_EMBEDDED_SERVICE_NAME",
+    "SALESFORCE_CHAT_EMBEDDED_SERVICE_COLLECTOR_NAME",
```
araujobarret, line 83: I have no idea what name fits here, it's more a question to clarify, is `COLLECTOR` a good name to differentiate from `AUCTION`?

lidimayra: Good question. This is because now we'll have two different Salesforce services being used on artsy.net.
So far we were using the "Collector Support Chat" for everything.

With these changes, we have "Collector Support Chat" and "Auction Support Chat" being used depending on the scenario.

araujobarret: thanks for explaining, so maybe we can make it clearer since in both cases they work for the collector but 1 for auction scope and the other for the general scope.
maybe `GENERAL`/`COMMON` instead of `COLLECTOR`?

lidimayra: I just wonder how confusing could it be for future maintainers to use a different terminology than just matching the same we're already using on Salesforce, which is the actual source of truth for this stuff...

araujobarret: Was more curious about other possibilities, but let's move with this and re-visit later if we need it.

araujobarret (approved):

src/Server/setup_sharify.ts
lidimayra: Cool, thank you!! 💜

PR #12567, opened 2023-06-28
feat(FX-4648): add missing analytic tracking for new search behaviour

nickskalkin:

This PR solves [FX-4648]

### Description

Adds missing analytics tracking for the new search bar / mWeb overlay.

- [Review app](https://search-tracking.artsy.net/)

You can see analytics events in the network tab in you browser (I filter by `v1/t`, which filters out everything not related to tracking)

### Other notes / questions

- [ ] Should we track scrolling (pagination) on mWeb?

There are a few differences compared to prod / weird things which I noticed during testing:

-  [Desktop] On prod, when clicking on the footer ("See full results" link), the `item_type` field in analytics props is `FirstItem` at the moment. I think this happened because before "See full results" were the first item in the list, but now is the last. We've renamed it to "Footer" to reflect the reality, I hope that's ok (I think that's ok, but we need to account for that when querying data, cc @abhitip)
- Previous search bar sends a deprecated [SelectedItemFromSearch](https://github.com/artsy/cohesion/blob/main/src/DeprecatedSchema/DeprecatedValues.ts#L128) event, whilst the new one sends the [modern one](https://github.com/artsy/cohesion/blob/main/src/Schema/Events/Search.ts#L50). The only noticeable difference is the event name (from payload to segment), it was "Selected item from search" before, now it is "selectedItemFromSearch"
- "context_module" was "Header" before, not it is "topTab" by default, don't' know whether this is an important change. Initially it was introduced [here](https://github.com/artsy/force/pull/12456/files#diff-33b96796781e80e736c2a98608313cddc065b962bcef1c6d5eeb1f8e837640f1R182)
- Same as above, old deprecated "Focused on search input" event was changed to a new  "focusedOnSearchInput"
 
[FX-4648]: https://artsyproduct.atlassian.net/browse/FX-4648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -34,14 +36,22 @@ interface OverlayProps {
...
+  const [fetchCounter, setFetchCounter] = useState(0)
 
   useEffect(() => {
     inputRef.current?.focus()
+
+    tracking.trackEvent({
+      action_type: ActionType.focusedOnSearchInput,
+      context_module: selectedPill.analyticsContextModule,
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nickskalkin, line 54: why disable linter: we want to track focus on the input only once, at the same time we want to send current pill's context info. When selecting another pill - this effect shouldn't be executed again

src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -34,14 +36,22 @@ interface OverlayProps {
 
 export const Overlay: FC<OverlayProps> = ({ viewer, relay, onClose }) => {
   const { t } = useTranslation()
+  const tracking = useTracking()
   const inputRef = useRef<HTMLInputElement | null>(null)
   const [selectedPill, setSelectedPill] = useState<PillType>(TOP_PILL)
   // TODO: Parse value from url
   const [inputValue, setInputValue] = useState("")
   const disablePills = !shouldStartSearching(inputValue)
+  const [fetchCounter, setFetchCounter] = useState(0)
```
nickskalkin, line 45: Works together with [this piece of code](https://github.com/artsy/force/pull/12567/files#diff-fa4cc42dc68b59e2279b658912183f1658324589a3f28404e27e7255023ba273R60).

This is similar to how it was implemented in the [desktop version of the new search bar](https://github.com/artsy/force/blob/98567f256f5aab0006ebf371144cff93f2e8253e/src/Components/Search/NewSearch/NewSearchBarInput.tsx#L48C10-L48C22).

In short, relay's `refetch` doesn't allow to get response data in the callback function.

```js
relay.refetch(..., error => {
  if (error) { ... }
  
  // here there is no known to me way to get response data
})
```

We've managed to overcome this obstacle by combining `fetchCounter` state and `useEffect` to make sure that tracking event is sent only when component is re-rendered because of new data refetch.

Let me know if you know of a better solution, I would be extremely happy to refactor this one.

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -32,9 +38,27 @@ const SearchResultsList: FC<SearchResultsListProps> = ({
...
+      tracking.trackEvent({
+        action_type:
+          options.length > 0
+            ? DeprecatedSchema.ActionType.SearchedAutosuggestWithResults
+            : DeprecatedSchema.ActionType.SearchedAutosuggestWithoutResults,
+        context_module: selectedPill.analyticsContextModule,
+        query: query,
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nickskalkin, line 59: same explanation as [here](https://github.com/artsy/force/pull/12567/files#r1245177229), I don't want to execute this effect function on query or selectedPill change, only manually when updating fetchCounter

src/Components/Search/NewSearch/Mobile/Overlay.tsx
```diff
@@ -67,6 +77,8 @@ export const Overlay: FC<OverlayProps> = ({ viewer, relay, onClose }) => {
           }
 
           scrollToTop()
+          // trigger useEffect to send tracking event
+          setFetchCounter(prevCounter => prevCounter + 1)
```
dzucconi, line 81: Just pass a callback and use that for tracking. The state is unnecessary — in general avoid effects if you can.

src/Components/Search/NewSearch/Mobile/Overlay.tsx
nickskalkin: solved the other way, really there was no need for this counter...

src/Components/Search/NewSearch/Mobile/Overlay.tsx
nickskalkin: solved!

src/Components/Search/NewSearch/Mobile/Overlay.tsx
dzucconi: For comments like this, you would want to leave them in the code itself.

src/Components/Search/NewSearch/Mobile/SearchResultsList.tsx
```diff
@@ -18,11 +18,16 @@ import {
 import { InfiniteScrollSentinel } from "Components/InfiniteScrollSentinel"
 import { NoResults } from "Components/Search/NewSearch/Mobile/SearchResultsList/NoResults"
 import { ContentPlaceholder } from "Components/Search/NewSearch/Mobile/SearchResultsList/ContentPlaceholder"
+import { useTracking } from "react-tracking"
+import * as DeprecatedSchema from "@artsy/cohesion/dist/DeprecatedSchema"
```
dzucconi, line 22: Why are we still using the `DeprecatedSchema` here — can't we update cohesion?

nickskalkin: addressed, thanks 👍

src/Components/Search/NewSearch/Mobile/Overlay.tsx
nickskalkin: done

nickskalkin:
> I don't see an item_type field in the search events cohesion

@abhitip you are right, I think I've got confused:

- Current search bar implementation sends deprecated event with the same name, but different fields, like item_id, item_type, etc. ([code](https://github.com/artsy/force/blob/main/src/Components/Search/SearchBar.tsx#L265))
- I though that the new event has the same shape, but seems like it doesn't. I see that you've added this new event to cohesion [long ago](https://github.com/artsy/cohesion/commit/8a475e0985a5adf427fc27dd202e9b25a192a4fd#diff-ee70df3a6cb0738382827014e641e2491d82c625fcd0596f7f1075da3454b724R50), but apparently it is not used by web app (at least I cannot find any piece of code sending `selectedItemFromSearch` event). Now I'm wondering if this is ok to use this event instead of an old one? Should we change the shape of it by adding fields like `item_number`, etc. or is it ok to send like it is?

abhitip:
Ah yes let's use cohesion and we can use it as-is!

nickskalkin:
Hey, going to merge (I've addressed initial feedback from @dzucconi)

PR #12571, opened 2023-06-28
Revert "chore: Remove zendesk references"

lidimayra:
Reverts artsy/force#12559

Relates to the issue described here: https://artsy.slack.com/archives/C9RK0BLEP/p1687970942807569

I can reproduce it locally and I noticed that by reverting this PR, the issue doesn't happen.

I'm reverting it for now, so we can at least activate Salesforce chat back again.


araujobarret (approved):

damassi:
@lidimayra - this can be closed, yah?

PR #12572, opened 2023-06-28
fix: Make SalesforceWrapper safer

lidimayra:

### Description

Avoid hitting errors when the expected settings are not present. For ref, see:
https://artsy.slack.com/archives/C9RK0BLEP/p1687972679348039?thread_ts=1687970942.807569&cid=C9RK0BLEP

### DIFF
--- a/src/Apps/Order/__tests__/OrderApp.jest.tsx
+++ b/src/Apps/Order/__tests__/OrderApp.jest.tsx
@@ -678,7 +678,7 @@ describe("OrderApp", () => {
       mockGetENV.mockImplementation(() => ({ SALESFORCE_CHAT_ENABLED: true }))
       const props = getProps()
       const subject = getWrapper({
-        props: { ...props, location: { query: { isModal: true } } },
+        props: { ...props, match: { location: { query: { isModal: true } } } },
       })
 
       expect(subject.find("SalesforceWrapper")).toHaveLength(0)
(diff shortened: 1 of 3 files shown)


src/Components/SalesforceWrapper.tsx
```diff
@@ -57,27 +57,29 @@ export const SalesforceWrapper: React.FC<SalesforceWrapperProps> = ({
     )}/embeddedservice/5.0/esw.min.js`,
     removeOnUnmount: true,
     onReady: () => {
-      window.embedded_svc.settings.defaultMinimizedText = "Chat"
-      window.embedded_svc.settings.enabledFeatures = ["LiveAgent"]
-      window.embedded_svc.settings.entryFeature = "LiveAgent"
+      if (window.embedded_svc?.settings?.enabledFeatures) {
```
damassi, line 60: I think if we check against settings that should enable all of the rest of the props below to be guarded safely:

```
if (window.embedded_svc?.settings) {
  ...
}
```

It doesn't look like there's anything deeper than `settings.foo`, so all good 👍

joeyAghion (dismissed):

damassi (dismissed):
LGTM 👍

src/Apps/Order/__tests__/OrderApp.jest.tsx
```diff
@@ -678,7 +678,7 @@ describe("OrderApp", () => {
       mockGetENV.mockImplementation(() => ({ SALESFORCE_CHAT_ENABLED: true }))
       const props = getProps()
       const subject = getWrapper({
-        props: { ...props, location: { query: { isModal: true } } },
+        props: { ...props, match: { location: { query: { isModal: true } } } },
```
damassi, line 681: Code in the component is pulling from `match`, so not sure how this ever passed before 🤷

lidimayra: wow, that's interesting!! Thank you for identifying it and fixing it!! 🙇‍♀️

PR #12575, opened 2023-06-29
fix(NX-3562): Remove duplicated component

lidimayra:
[NX-3562]

### Description

We're calling SalesforceWrapper twice inside ArtworkChatBubble (mistakenly done during #12559). This commit removes one of them.

[NX-3562]: https://artsyproduct.atlassian.net/browse/NX-3562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cc @artsy/negotiate-devs

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkChatBubble.tsx
+++ b/src/Apps/Artwork/Components/ArtworkChatBubble.tsx
@@ -39,12 +39,15 @@ const ArtworkChatBubble: FC<ArtworkChatBubbleProps> = ({ artwork }) => {
     return null
   }
 
-  return getENV("SALESFORCE_CHAT_ENABLED") ? (
+  if (!getENV("SALESFORCE_CHAT_ENABLED")) {
+    return null
+  }
+
+  return (
     <Media greaterThan="xs">
-      <SalesforceWrapper />
       <SalesforceWrapper isInAuction={isInAuction} />
     </Media>
-  ) : null
+  )
 }
 
 export const ArtworkChatBubbleFragmentContainer = createFragmentContainer(


src/Apps/Artwork/Components/ArtworkChatBubble.tsx
```diff
@@ -39,12 +39,15 @@ const ArtworkChatBubble: FC<ArtworkChatBubbleProps> = ({ artwork }) => {
...
+    return null
+  }
+
+  return (
     <Media greaterThan="xs">
-      <SalesforceWrapper />
       <SalesforceWrapper isInAuction={isInAuction} />
     </Media>
-  ) : null
+  )
```
araujobarret, line 50: good catch, I thought it was intentional before due to some funkiness of SF 👍

araujobarret (approved):

src/Apps/Artwork/Components/ArtworkChatBubble.tsx
```diff
@@ -39,12 +39,15 @@ const ArtworkChatBubble: FC<ArtworkChatBubbleProps> = ({ artwork }) => {
     return null
   }
 
-  return getENV("SALESFORCE_CHAT_ENABLED") ? (
+  if (!getENV("SALESFORCE_CHAT_ENABLED")) {
+    return null
+  }
```
damassi, line 44: Thanks for removing the ternary here. Quick early exits (no matter how many there are, one after the other) make code way more readable 👍

lidimayra: Thank you actually, I only did this because I saw you [doing it](https://github.com/artsy/force/pull/12572/commits/9cda3c85384f255b3399f0509cd021989cd23cec) and it really seemed much easier to read ❤️

PR #12576, opened 2023-06-29
fix(FX-4827): fixes new search bar for small screens

nickskalkin:

This PR solves [FX-4827]

### Description

Currently search bar is broken on small devices: if search dropdown is bigger than display area - dropdown is "flipped" (displayed above the search input). This is something we want to avoid, and it became possible to enforce "un-flippable" behaviour after https://github.com/artsy/palette/pull/1303 was merged.

Apart from that, I've changed the "See full results..." item from being a regular option to a footer. Now if dropdown area doesn't fit into the screen, footer is still visible (options area becomes scrollable).

| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/4036fb6a-1b4c-4b34-93c3-003f6fe70c33" /> | <video src="https://github.com/artsy/force/assets/3934579/2abcf8a5-2734-4ca8-bc8c-689c2ad9c798" />|

Props to @dzucconi for the fix in palette ❤️ 

[FX-4827]: https://artsyproduct.atlassian.net/browse/FX-4827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/NewSearch/NewSearchBarInput.tsx
+++ b/src/Components/Search/NewSearch/NewSearchBarInput.tsx
@@ -48,6 +48,7 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
   const [fetchCounter, setFetchCounter] = useState(0)
   const { router } = useRouter()
   const encodedSearchURL = `/search?term=${encodeURIComponent(value)}`
+  const [forceRerender, setForceRerender] = useState(0)
 
   const options = extractNodes(viewer.searchConnection)
   const formattedOptions: SuggestionItemOptionProps[] = options.map(
@@ -171,7 +172,7 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
   }
 
   const handleClickOnFooter = () => {
-    clearSearchInput()
+    setForceRerender(prevCounter => prevCounter + 1)
   }
 
   const handleFocus = () => {
@@ -184,6 +185,7 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
   return (
     <AutocompleteInput
       id="SearchBarInput"
+      key={`autocomplete-input-${forceRerender}`}
       placeholder={t`navbar.searchBy`}
       spellCheck={false}
       options={shouldStartSearching(value) ? formattedOptions : []}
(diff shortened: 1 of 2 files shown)


src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -182,6 +171,10 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
     redirect(option.href)
   }
 
+  const handleClickOnFooter = () => {
+    setForceRerender(prevCounter => prevCounter + 1)
```
nickskalkin, line 175: One issue that I've introduced in this PR and not aware how to handle it properly: when clicking on the "See full results" item, user is redirected to a /search?term=query page and **dropdown should be closed**.

For some reason, when clicking on the footer component, dropdown area is not being closed (I think close event is not being dispatched in the palette's component). Is there a way to trigger close somehow? Or maybe I should add a new argument to the AutocompleteInput component in palette so that it will be triggering "close" event when clicking on footer? @dzucconi

dzucconi: `setForceRerender` looks like a hack; I think any time you're including these state counters we're papering over some other issue.

nickskalkin: oh, cool, I didn't know this existed! And seems like it was added recently for this particular case: https://github.com/artsy/palette/pull/1296

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -191,6 +179,7 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
 
   return (
     <AutocompleteInput
+      id="SearchBarInput"
```
anandaroop, line 182: **question:** Is this `id` actually needed? (Or was it meant to be a test id? Since this UI element is present globally, could potentially cause an id collision.)

anandaroop (dismissed):
Looks great! One minor point below…

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: Thanks, not needed for sure. I even removed it [last week](https://github.com/artsy/force/pull/12576/commits/344592a45ee9c6f66f6d4776b23b025fd97acd6c) but seems like I forgot to push changes

PR #12579, opened 2023-06-29
chore: adds search example using autocomplete input

dzucconi:
@nickskalkin was wondering how to use a fragment container with the `AutocompleteInput` — the answer is: it just works. Use the `matchConnection` instead of the `searchConnection` (to be deprecated), however.

https://github.com/artsy/force/assets/112297/f1de9123-5c1e-4069-8907-dc799421e47b

### DIFF
--- a/src/Apps/Example/Routes/Search/SearchRoute.tsx
+++ b/src/Apps/Example/Routes/Search/SearchRoute.tsx
@@ -1,0 +1,73 @@
+import { AutocompleteInput } from "@artsy/palette"
+import { useClientQuery } from "Utils/Hooks/useClientQuery"
+import { extractNodes } from "Utils/extractNodes"
+import { compact, debounce } from "lodash"
+import { FC, useMemo, useState } from "react"
+import { graphql } from "react-relay"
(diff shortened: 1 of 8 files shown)


src/Apps/Example/Routes/Search/SearchRoute.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const options = compact(
+    extractNodes(data?.matchConnection).map(node => {
+      if (!node) return null
+
+      return {
+        // Internally, `AutocompleteInput` requires a display `text` and a `value`
+        value: node.value!,
+        text: node.text!,
+        // Pass on the node for the fragment container
+        artist: node,
```
dzucconi, line 27: ☝️

dzucconi:
Might as well have a pattern for debounce since it's surprisingly hard to do for React inputs.

damassi (dismissed):

nickskalkin:
wow, thanks a lot @dzucconi! I'll try to refactor components I wrote when I will have some spare time!

PR #12583, opened 2023-07-04
feat: update empty state for alerts

dblandin:
Provide a more useful empty state design for the alerts section, guiding collectors towards creating their first alert.

Jira: [FX-4826](https://artsyproduct.atlassian.net/browse/FX-4826)

## Screenshots

| Before | After |
|--------|--------|
| Screenshot 2023-07-04 at 11 48 10 | Screenshot 2023-07-04 at 11 47 45 |
| staging artsy net_settings_alerts(iPhone 12 Pro) | localhost_4000_settings_alerts(iPhone 12 Pro) |


nickskalkin (changes requested):
Cool! The only thing that I would change, see how Filter icon is not aligned with other icons:



I was able to fix it locally by making [these changes](https://github.com/artsy/force/commit/6bcd3f4041c8b9fb48868cc40705adf339cd57ef).

dblandin:
@nickskalkin Good catch! I just pushed an update to the icon rendering which should resolve the size/alignment issues. Updated the screenshots as well.

nickskalkin (approved):

PR #12590, opened 2023-07-07
[#FF] refactor: bank account auto-selection hook

iskounen:
This PR is an attempt to understand how the payment route works by refactoring the hook that is responsible for picking a bank account to be selected when either `US_BANK_ACCOUNT` or `SEPA_DEBIT` is the selected payment method.

I want to refactor it for the following reasons:

1. I don't think that this hook should do anything when `CREDIT_CARD` or `WIRE_TRANSFER` is selected. (It currently updates the `bankAccountSelection` state property to `{ type: "new" }`)
1. I think that this logic would be easier to read if it were in simple if-else statements.
1. I think that the `BankAccountPicker` component should house this logic.

I would also like to see if I can remove the `// eslint-disable-next-line` by resolving the warning that it otherwise generates.

### DIFF
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -1,4 +1,4 @@
-import { FC } from "react"
+import { FC, useEffect } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { StripeError } from "@stripe/stripe-js"
 
@@ -61,6 +61,31 @@ export const BankAccountPicker: FC<Props> = props => {
   }
 
   const { submitMutation: setPaymentMutation } = useSetPayment()
+
+  useEffect(() => {
+    if (bankAccountSelection) {
+      return
+    }
+
+    if (selectedPaymentMethod === "SEPA_DEBIT") {
+      setBankAccountSelection({ type: "new" })
+    } else if (selectedPaymentMethod === "US_BANK_ACCOUNT") {
+      if (order.bankAccountId) {
+        setBankAccountSelection({
+          type: "existing",
+          id: order.bankAccountId,
+        })
+      } else if (bankAccountsArray.length > 0) {
+        setBankAccountSelection({
+          type: "existing",
+          id: bankAccountsArray[0]?.internalID!,
+        })
+      } else {
+        setBankAccountSelection({ type: "new" })
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [bankAccountSelection, selectedPaymentMethod])
 
   const handleContinue = async () => {
     setBalanceCheckComplete(false)
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -27,10 +27,7 @@ interface Props {
 }
 
 export const BankAccountPicker: FC<Props> = props => {
-  const {
-    me: { bankAccounts },
-    order,
-  } = props
+  const { me, order } = props
 
   const {
     bankAccountSelection,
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -62,6 +62,31 @@ export const BankAccountPicker: FC<Props> = props => {
...
+      } else if (bankAccountsArray.length > 0) {
+        setBankAccountSelection({
+          type: "existing",
+          id: bankAccountsArray[0]?.internalID!,
+        })
+      } else {
+        setBankAccountSelection({ type: "new" })
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
iskounen, line 87: I would like to see how I can remove this. In its current form, the hook depends on `order` and `bankAccountsArray`.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -101,21 +97,6 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
-    if (!bankAccountSelection && selectedPaymentMethod) {
-      const bankAccountsArray =
-        selectedPaymentMethod !== "SEPA_DEBIT"
-          ? extractNodes(me.bankAccounts)
-          : []
-
-      setBankAccountSelection(
-        getInitialBankAccountSelection(order, bankAccountsArray)
-      )
-    }
```
iskounen, line 114: This is more concise than what I am proposing, but it makes it harder to read this hook an understand what it is doing.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -101,21 +97,6 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
 
   const artworkVersion = extractNodes(order.lineItems)[0]?.artworkVersion
 
-  useEffect(() => {
-    if (!bankAccountSelection && selectedPaymentMethod) {
```
iskounen, line 105: We are not guarding against non-bank account payment methods, so this hook is updating state for payment methods that don't need it.

src/Apps/Order/Components/BankAccountPicker.tsx
dzucconi: Generally, you want to just include everything it specifies.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -62,6 +62,31 @@ export const BankAccountPicker: FC<Props> = props => {
...
+        })
+      } else if (bankAccountsArray.length > 0) {
+        setBankAccountSelection({
+          type: "existing",
+          id: bankAccountsArray[0]?.internalID!,
+        })
+      } else {
+        setBankAccountSelection({ type: "new" })
+      }
+    }
```
dzucconi, line 86: Could you unnest all these conditions?

src/Apps/Order/Components/BankAccountPicker.tsx
rquartararo: Is there an issue with including `order` and `bankAccountsArray` in the dependency array?

rquartararo: The `if (bankAccountSelection) { return }` safeguards the user's selection from being overridden on rerender, so it may be safe to add them.

src/Apps/Order/Components/BankAccountPicker.tsx
```diff
@@ -62,6 +59,64 @@ export const BankAccountPicker: FC<Props> = props => {
 
   const { submitMutation: setPaymentMutation } = useSetPayment()
 
+  /**
+   * This hook is responsible for setting the bankAccountSelection state
+   * properties that are used by this component to display an appropriate
+   * payment details section to users.
+   */
+  useEffect(() => {
```
araujobarret, line 67: That's a good refactor, in Order app a lot of routes `index` files have tons of useEffect and other callbacks making it super chunky, and hard to read and maintain.
See how clearer it is right now, thanks for going down this rabbit hole!

PR #12597, opened 2023-07-11
Revert "[#FF] refactor: bank account auto-selection hook"

iskounen:
Reverts artsy/force#12590 because I noticed two issues while testing with "closed account" and "insufficient funds" accounts on staging:

1. After connecting the bank accounts and continuing to the Review step, I was correctly returned to the Payment step and shown an error message. However, the bank account that I had just connected was not visible in the list of bank accounts.
2. When I select another bank account from the list after recovering from the error in (1) the form is automatically re-submitted. The submission process takes several seconds, and when it is complete I see the Review step with the bank account from (1) selected.

I would like to test this again without these changes to confirm whether this is a regression or not.

### DIFF
--- a/src/Apps/Order/Components/BankAccountPicker.tsx
+++ b/src/Apps/Order/Components/BankAccountPicker.tsx
@@ -1,4 +1,4 @@
-import { FC, useEffect } from "react"
+import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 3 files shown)


iskounen:
I've confirmed that this behavior exists without my changes in #12590 so I will re-merge those changes and write a bug ticket for this behavior.

PR #12599, opened 2023-07-11
feat(AMBER-8): paginated VRs

mdole:

This PR solves [AMBER-8]

### Description

Converts from 2 non-paginated components for current and upcoming VRs and a page-based paginated component for past VRs -> 3 "show more" paginated components so that all VRs can be loaded if the partner has more than 12 current or upcoming.

partner-vrs

[AMBER-8]: https://artsyproduct.atlassian.net/browse/AMBER-8?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Partner/Components/PartnerViewingRooms/PartnerViewingRoomsGrid.tsx
```diff
@@ -0,0 +1,136 @@
...
+const PartnerViewingRoomsGrid: React.FC<PartnerViewingRoomsGridProps> = ({
+  sectionTitle,
+  viewingRoomsConnection,
+  relay,
+}) => {
+  const viewingRoomNodes = extractNodes(
+    viewingRoomsConnection.viewingRoomsConnection
+  )
+
+  const [loading, setLoading] = useState(false)
```
araujobarret, line 30: Relay pagination already has an internal loading state, you don't need your own, you can use `relay.isLoading()` in your pagination CTA.

mdole: Tried it out - we still need the `useState` because `relay.isLoading()` returns `true` when the _previous_ call to `relay.loadMore()` [is still pending](https://relay.dev/docs/v10.1.3/pagination-container/#isloading), and returns `false` otherwise.

So if our loading state is just `loading={relay.isLoading()}`, we'll show the spinner forever after it's clicked the first time. And if it's `loading={!relay.isLoading()}`, it'll show a spinner forever on page load.

But this did make me notice that we should be checking in the click handler if we're still waiting on the previous call to resolve and stop trying to load more if so. Updated that!

Mitch-Henson (approved):
Not a relay expert myself but it is looking good to me!

PR #12602, opened 2023-07-11
feat: make address verification query (EMI-1242)

iskounen:
The type of this PR is: Feat
This PR solves [EMI-1242]

### Description

This PR connects the Shipping step to the newly implemented address verification API to give users a warning (or suggestions) if their address is not recognized by the Smarty API.

This work is currently behind `"address_verification"` Unleash feature flag.

[EMI-1242]: https://artsyproduct.atlassian.net/browse/EMI-1242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Components/CheckoutAddress.tsx
```diff
@@ -97,8 +101,49 @@ export const CheckoutAddress: FC<{
...
+        address,
+      },
+      {
+        fetchPolicy: "network-only",
+      }
+    ).toPromise()
+
+    console.log({ response })
+
+    const { firstLine, secondLine } = formatAddress({ name, ...address })
```
dzucconi, line 146: We use Yup to validate. Here's an example of an async validator: https://github.com/artsy/force/blob/main/src/Components/PhoneNumberInput/validatePhoneNumber.ts

dzucconi: See https://github.com/artsy/force/tree/main/src/Components/PhoneNumberInput#validatephonenumber for how to include one in the validation schema

iskounen: Thanks for the resources! For context, there's two types of validation/verification happening in this form: field validation and address verification. We'll be showing inline error messages like "a postal code is required" when field validation fails, but we'll be opening a modal dialog when address verification fails.

Would you recommend that we use Yup to initiate address verification once field validation passes?

modal dialog

dzucconi: Ah, that looks like it should be it's own fragment container since all this data is tied to UI. But you'll also want to conditionally transition the dialog into this state. You *could* just transition the dialog into that state as a processing step then transition out of it if the query comes back valid or what have you.

dzucconi: But yes, as you say: Yup is only used for form validation and this is a distinct kind of validation not tied to fields.

dzucconi: I mean it sounds like what should happen is that the verification should be a field on the mutation, you request it as part of the mutation and if needed you transition into this dialog

iskounen: I like the idea of Exchange being responsible for address verification (via Gravity or directly) when the shipping method/address are saved to the order. If verification succeeds, Force won't need to behave any differently than it does today. If verification fails, Force will be notified in the response from the `SetShipment` mutation.

@erikdstock and @fladson I see that you considered this approach in the [tech plan](https://www.notion.so/artsy/Address-Verification-Checkout-flow-51ae42901d094faf8b8c18f6c61de875?pvs=4#e294d7b1847b4602b29c344d97265be4). I'm not proposing we change approaches for this iteration, but for the sake of discussion, I'm curious whether you're still interested in that approach or not.

erikdstock: I added a comment on that section [here](https://www.notion.so/Address-Verification-Checkout-flow-51ae42901d094faf8b8c18f6c61de875?d=462d2c67714f4eb7a347be872ea99a8b&pvs=4#87ae84a74051418d838351057a1860d1). I think it's worth revisiting.

src/Apps/Debug/DebugApp.tsx
```diff
@@ -8,6 +12,46 @@ export const DebugApp: React.FC<{}> = () => {
       <Title>Baseline</Title>
 
       <Text variant="sm-display">Baseline</Text>
+      {/* TODO: REMOVE THIS BEFORE MERGE */}
```
erikdstock, line 15: for this file we can just checkout the `DebugApp.tsx` app from main, but it was a great TIL

src/Components/AddressForm.tsx
```diff
@@ -123,7 +123,7 @@ export const AddressForm: React.FC<AddressFormProps> = ({
           placeholder="Full name"
           title={billing ? "Name on card" : "Full name"}
           autoCorrect="off"
-          value={address.name}
+          value={value?.name}
```
iskounen, line 126: This change is what allows the address form to be updated when the user selects a suggested address after verification. Before this change, the address form was only updated from user input. After this change, the parent component (`ShippingRoute`) can also tell the address form which values to show.

PR #12608, opened 2023-07-13
Extract address verification functionality from address form

erikdstock:
Update: Paired with @dzucconi to update this to use a query renderer. The solution below is no longer applicable but can be seen in the first commit.

This is a PR against the branch for #12602 with one major step extracted out: Removing address verification logic from the address form and using it to wrap the address form (any address form) instead. I would like to pair with someone to clean up some likely messy code parts, make sure i'm using react hooks correctly, fix up types, etc. I've been pairing on this ticket with @iskounen.

There is a lot to improve in here, particularly around **types** (esp working with the metaphysics schema, which is still malleable since this query is not used yet - some mixed in nulls and undefineds), **validations** (we don't want a verified address to fail validations at a later step), and **react hooks** (I'm feeling a little rusty on best practices and safe assumptions WRT when to `useState` and when to trust a stateful variable will have been set in procedural code vs having to rely on `useEffect` hooks).

_Thanks in advance_

In short, the extracted `AddressVerificationManager` works like this:
- Takes the form as a render prop (not even sure if this is still a popular pattern)
- Passes `verifyAddress()` and `resetModal()` to the render prop
- If a form, the wrapped renderProp should `verifyAddress()` on submit, `.then` handle the final values and probably `resetModal()`.
- `verifyAddress()`
  - Receives the form values, setting any non-address values aside for later
  - calls the metaphysics verification endpoint
  - presents the user with a modal to help them choose the best address (unless it is perfectly verified, unless it proceeds to the last step)
  - waits for the user to choose an address
  - returns a promise that resolves, adding any non-address values back in, when the final chosen address is available

The implemented component would look something like [this](https://github.com/artsy/force/compare/iskounen/feat/address-verification-query...iskounen+erik/feat/address-verification-query?expand=1#diff-e0afcd1142a7fdd8a803a66f3f2490eb8db1d25fe9d66bd889fefd4852937ec6R746-R766):
```tsx
<AddressVerificationManager
  render={({ verifyAddress, resetModal }) => (
    <SomeAddressForm
      onSubmit={values => {
        verifyAddress(values).then(finalAddress => {
          onAddressChange(finalAddress)
          resetModal()
        })
      }}
    />
  )}
/>
```


src/Apps/Order/Components/AddressVerificationManager.tsx
```diff
@@ -0,0 +1,357 @@
...
+  recommended = "Recommended",
+  user_address = "What you entered",
+}
+
+export const AddressVerificationManager: FC<{
+  render: React.FC<{
+    verifyAddress: (address: VerifiableAddress) => React.ReactNode
+    resetModal: () => void
+  }>
+}> = ({ render }) => {
```
ansor4, line 61: :ceiling_cat: What's the upside of using a function param? Is this possible by passing something like `({ children })` instead?

If it is, then it could simplify some of the code below if I noted.

src/Apps/Order/Components/AddressVerificationManager.tsx
```diff
@@ -0,0 +1,357 @@
...
+
+    addressOptions = [
+      ...visibleSuggestedAddresses,
+      { ...inputAddress, key: "userAddress", recommended: false },
+    ]
+  }
+
+  return (
+    <>
+      {render({ verifyAddress, resetModal })}
```
ansor4, line 216: If we pass `children` as props instead, this line could just be `{children}`.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -742,9 +743,27 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                 Delivery address
               </Text>
               {isAddressVerificationEnabled ? (
-                <CheckoutAddress
-                  userCountry={props.me.location?.country || "United States"}
-                  onChange={onAddressChange}
+                <AddressVerificationManager
```
ansor4, line 746: If we're to use `children` as props for `AddressVerificationManager`, then instead of passing this `render` function that returns a react component, we could just nest it:

```.jsx
<AddressVerificationManager>
	<CheckoutAddress
		userCountry={props.me.location?.country || "United States"}
		onChange={ e => {} }
		onSubmit={values => {
			verifyAddress(values).then(finalAddress => {
			onAddressChange(finalAddress)
			resetModal()
			})
		}}
	/>
</AddressVerificationManager>
```

ansor4 (commented):
Overall I think the direction makes sense! Not sure I understand the why we need to pass a `render` function as opposed to using `children` components, but I'm not close to this code!

src/Apps/Order/Components/AddressVerificationManager.tsx
erikdstock: The reason (using this approach) was that the children/render component needs to accept a prop created in this components body, so it can only be passed in after that is created. However, Damon pointed out a much more elegant solution to me using a `QueryRenderer`.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -0,0 +1,309 @@
...
+    onClose()
+  }, [onClose])
+
+  useEffect(() => {
+    if (addressOptions.length > 0) {
+      setSelectedAddressKey(addressOptions[0].key)
+    }
+  }, [addressOptions])
+
+  // Handling possibly incorrectly (by MP) possible null values
```
dzucconi, line 98: I would just fix the types in Metaphysics

erikdstock: yes, i agree. i'll make a ticket.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -0,0 +1,309 @@
...
+  const setAddressOptions = (options: AddressOption[]) => {
+    console.log("options", options)
+    setAddressOptions2(options)
+  }
+  const setSelectedAddressKey = (key: string) => {
+    console.log("key", key)
+    setSelectedAddressKey2(key)
+  }
+
+  const chooseAddress = useCallback(() => {
```
TMcMeans, line 75: **question [non-blocking]:** 

Got a quick question, so it looks like choosing the selected address is based on string matching by address key, right? Would there be any possibility of key collisions here? Like, could two different addresses wind up with the same key property and maybe cause some unexpected behavior?

iskounen: I don't think that there is a risk of key collision because each suggestion gets a unique key based on its position in the Smarty API response that we get from Gravity. In the `useEffect` hook that builds up the `addressOptions` state variable, we set the user-entered address key to `userAddress` and the suggested address keys to `suggestedAddress-0`, `suggestedAddress-1`, etc.

iskounen:
I merged this back into #12602 to have a shorter path to `main`.

src/Apps/Order/Components/AddressVerificationFlow.tsx
TMcMeans: Ah okay, so the addresses should stay consistent as long as the Smarty API's response order doesn't change between calls and the userAddress is assigned a different key than the suggestedAddress. Thanks for explaining this :)

PR #12609, opened 2023-07-14
Future friday: pending orders

erikdstock:
The type of this PR is: **FUTURE FRIDAY**

### Description

This is a future friday spike to see what showing pending orders on the order history page could look like. The implementation is minimal. Other notes:

- As I am re-using our order history items which means some expected values do not appear.
- I think some orders may have a display state of pending already on because they are pending acceptance from the user's perspective. Things like this could be replaced in a list item specific to a pending orders section.
- private sales orders would need to be filtered based on whether the invoice is finalized.
- other 'action required' orders could also be included - for example failed payment so that artsy users can see that information front and center, but ideally this would use a query based on a failed payment state which does not exist today.
- we could use a similar query in the nav to show a dot on the nav item in the presence of any pending/action required orders. The dot could go away by canceling or allowing such orders to expire.
- currently i deduplicate these orders based on `artworkId-editionSetId`.
- We would probably want more nuanced querying and logic eventually showing recent abandoned orders along with pending ones and that sort of thing.

cc @artsy/emerald-devs 

image

### DIFF
--- a/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchases.tsx
+++ b/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchases.tsx
@@ -1,6 +1,6 @@
 import { useState, FC } from "react"
 import { RelayRefetchProp, createRefetchContainer, graphql } from "react-relay"
-import { Join, Spacer, Message } from "@artsy/palette"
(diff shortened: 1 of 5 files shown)


damassi:
Going to close! Can revive if the work picks back up

PR #12622, opened 2023-07-19
feat: use two k8s deployments

artsyjian:

[PHIRE-111]

### Description

[We noticed](https://artsy.slack.com/archives/CA8SANW3W/p1689785076024629?thread_ts=1689359985.499589&cid=CA8SANW3W) a situation where 9 Spot instances were interrupted together, taking down most of prod `force-web` pods, causing elevated latency for ~5 minutes. This is not acceptable.

This PR is to try the idea of using two deployments: `force-web1`, `force-web2`.

`web1` runs on on-demand EC2 instances only, so its pods are guaranteed to exist.

`web2`, as the current single-deployment, can run on on-demand or Spot but prefers Spot. Its pods can be all on Spot and therefore can all go away.

With this setup, a baseline of pods (`web1`) is guaranteed, so the service may be degraded by Spot interruptions but it will never completely go down.

[PHIRE-88]: https://artsyproduct.atlassian.net/browse/PHIRE-88?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PHIRE-111]: https://artsyproduct.atlassian.net/browse/PHIRE-111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -7,7 +7,7 @@ metadata:
     layer: application
     component: web
     app.kubernetes.io/version: production
-  name: force-web1
+  name: force-web
 spec:
   replicas: 3
   strategy:
@@ -27,11 +27,11 @@ spec:
         layer: application
         component: web
         app.kubernetes.io/version: production
-      name: force-web1
+      name: force-web
       namespace: default
     spec:
       containers:
-      - name: force-web1
+      - name: force-web
         env:
         - name: PORT
           value: '5000'
@@ -97,7 +97,7 @@ metadata:
     layer: application
     component: web
     app.kubernetes.io/version: production
-  name: force-web2
+  name: force-web-spotty
 spec:
   strategy:
     rollingUpdate:
@@ -116,11 +116,11 @@ spec:
         layer: application
         component: web
         app.kubernetes.io/version: production
-      name: force-web2
+      name: force-web-spotty
       namespace: default
     spec:
       containers:
-      - name: force-web2
+      - name: force-web-spotty
         env:
         - name: PORT
           value: '5000'
@@ -195,13 +195,13 @@ spec:
 apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
-  name: force-web2
+  name: force-web-spotty
   namespace: default
 spec:
   scaleTargetRef:
     apiVersion: apps/v1
     kind: Deployment
-    name: force-web2
+    name: force-web-spotty
   minReplicas: 1
   maxReplicas: 17
   behavior:
--- a/hokusai/production.yml
+++ b/hokusai/production.yml
@@ -6,7 +6,6 @@ metadata:
     app: force
     layer: application
     component: web
-    spot-allowed: false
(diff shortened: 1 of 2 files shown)


hokusai/production.yml
```diff
@@ -1,12 +1,107 @@
...
+metadata:
+  labels:
+    app: force
+    layer: application
+    component: web
+    spot-allowed: false
+    app.kubernetes.io/version: production
+  name: force-web1
+spec:
+  replicas: 3
```
joeyAghion, line 13: This^ would be the bare minimum scale during spot disruptions?

joeyAghion (approved):
As discussed a little, this starts to get a little messy, but I see the appeal. Is this an experiment? Would we extend it to Metaphysics and maybe other critical services?

One thought (that you're not going to like 😆): I'd prefer more descriptive labels instead of `force-web1`/`2`. Should we call them `-on-demand` and `-spot`? Or maybe just `force-web` and `force-web-spot` to match how node tiers are labeled?

hokusai/production.yml
artsyjian: yes. there's a trade-off here. higher replica count means less degradation but also less Spot utilization. lower count means higher Spot utilization but more degradation.

artsyjian:
> Should we call them `-on-demand` and `-spot`?

That's what I had in mind at first, but `-spot` for the 2nd deployment is not accurate, as the deployment allows both on-demand and spot. Can't think of a name that captures that.

artsyjian:
> Is this an experiment? Would we extend it to Metaphysics and maybe other critical services?

We can use this  as a pilot. If it works well for Force, we can extend it to MP/Gravity/Gemini.

joeyAghion:
> Can't think of a name that captures that.

Yeah I see what you mean. We could tolerate that the name isn't precise (but keep it short, like `force-web-spot`), or be more explicit like `force-web-spot-preferred`. Or keep it like this. 🤷

artsyjian:
> be more explicit like `force-web-spot-preferred`. Or keep it like this. shrug

how about `force-web-spotty`?

joeyAghion:
Actually I love that but can't tell if you're joking. 😆

artsyjian:
> Actually I love that but can't tell if you're joking. laughing

dead serious. committed.

artsyjian:
cc @mzikherman - change of plan, going back to using Spot in staging, but now with a baseline deployment (1 pod) on on-demand nodes.

joeyAghion:
By the way, if we used `-web1`/`-web2`, we'd have to follow up and manually delete the `force-web` deployment, right? This avoids that, which is especially nice if we make similar changes to other projects.

artsyjian:
deleted `force-web` hpa in prod.

PR #12624, opened 2023-07-19
feat(EMI-1285): add tracking to address validation on checkout

TMcMeans:

This PR solves [EMI-1285]

### Description

The goal of this PR is to add analytics tracking to specific events associated with address validation during checkout. Here are the list of events to be scheduled with their appropriate data properties: https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0

This PR relates to https://github.com/artsy/cohesion/pull/446 && https://github.com/artsy/cohesion/pull/443

@artsy/emerald-devs 

TODO

- [x] add tracking onClose
- [x] fix duplicate analytics calls firing

[EMI-1285]: https://artsyproduct.atlassian.net/browse/EMI-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Components/AddressModal.tsx
```diff
@@ -112,7 +112,11 @@ export const AddressModal: React.FC<Props> = ({
       {show && (
         <ModalDialog
           title={title}
-          onClose={handleModalClose}
+          onClose={() => {
+            // TODO: tracking for modal close
+
+            handleModalClose()
+          }}
```
iskounen, line 119: **question:** Do we want to track how often users cancel out of this modal?

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -109,6 +114,18 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
     } else {
       if (verificationStatus === "VERIFIED_WITH_CHANGES") {
         setModalType(ModalType.SUGGESTIONS)
+
+        trackEvent({
+          action_type: "validationAddressViewed",
+          context_module: "OrdersShipping",
+          context_page_owner_type: "orders-shipping",
+          context_page_owner_id: "57e60c68-a198-431e-8a02-6ecb01e3a99b",
```
iskounen, line 122: **suggestion:** I'm not too excited about having to pass an order ID to this component just to populate this event. Perhaps we can fire this event on the [`onClose` and `onChosenAddress` props](https://github.com/artsy/force/blob/ce6b42bdef12bc07be7d432ec89507324a73a63e/src/Apps/Order/Routes/Shipping/index.tsx#L769-L777) that are passed in from the parent component?

src/Apps/Order/Components/AddressModal.tsx
TMcMeans: there's some chatter on slack about this and @daytavares agrees that we should make another event to handle tracking the modal closing action.

src/Apps/Order/Components/AddressVerificationFlow.tsx
TMcMeans: hey, so I totally understand not wanting to pass the ids as props to the AddressVerificationFlow, but looking through the flow there are some distinctions that need to be made for each tracking event. We want to be able to log an event for each modalType based on user action (whether they see the review or suggestions modals) so that's why I included them here.

TMcMeans: I was able to useSystemContext() and useAnalyticsContext() to get access to the user_id & ownerrType so I believe we're all good here!

TMcMeans:
Am currently investigating duplicate tracking events at the same time and haven't gotten to the bottom of that yet!

src/Apps/Order/Components/AddressModal.tsx
TMcMeans: yup I think I have what you suggest covered here... not tracking a modal close on button press but tracking exactly which buttons were pressed

src/Apps/Order/Components/AddressVerificationFlow.tsx
TMcMeans: Figured this out with some help from @rquartararo- the `contextPageOwnerSlug` coming from systemContext() provides the order-id and that's what I'll use here.

TMcMeans:
> Am currently investigating duplicate tracking events firing at the same time and haven't gotten to the bottom of that yet!

The dupe problem has been solved!

damassi:
@TMcMeans - that danger semantic commit error will be fixed if you format your title right -- almost had it:
```
feat(ticket-number): description of thing
```

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -166,13 +211,50 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+                  context_page_owner_id: contextPageOwnerSlug,
+                  user_id: user?.id,
+                  subject: "Check your delivery address",
+                  option: selectedAddressKey.includes("suggestedAddress")
+                    ? "Recommended"
+                    : "What you entered",
+                  label: "Back to Edit",
+                })
+              }
+              handleClose()
```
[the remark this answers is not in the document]
starsirius: From the [spreadsheet](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0), looks like we want to track the `clickedCloseValidationAddress` event from all sources, e.g. button clicks, X, and clicking away. So we might want to make `handleClose` aware of the source.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -75,6 +83,14 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
   const handleClose = () => {
     setModalType(null)
+    trackEvent({
+      action_type: "clickedCloseValidationAddress",
+      context_module: ContextModule.ordersShipping,
+      context_page_owner_type: OwnerType.ordersShipping,
+      context_page_owner_id: contextPageOwnerSlug,
+      option: "null",
```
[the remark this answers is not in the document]
TMcMeans: I think you're right, it should be non-string null type not string "null."

src/Apps/Order/Components/AddressVerificationFlow.tsx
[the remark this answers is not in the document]
starsirius: Actually I just realized the 2 modals handle close differently. The "review and confirm" modal [does not call `handleClose`](https://github.com/artsy/force/blob/78cbd94f05418f4b819913df648b81a29f8cf7c5/src/Apps/Order/Components/AddressVerificationFlow.tsx#L187), but [the other one does](https://github.com/artsy/force/blob/78cbd94f05418f4b819913df648b81a29f8cf7c5/src/Apps/Order/Components/AddressVerificationFlow.tsx#L132). We might want to keep them consistent.

Also if I understand correctly the [`onClose` prop](https://github.com/artsy/palette/blob/917516d7205fe8f03b4a0d3455fc8fb4c251ad8f/packages/palette/src/elements/Modal/ModalBase.tsx#L74) on `ModalDialog` is only called when the modal is dismissed when clicking away, the "X" and the escape key. To track the modal close from the modal buttons, we might need to implement it separately. I haven't touched tracking code for a while, but let me know if I misunderstood things here.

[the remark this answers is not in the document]
TMcMeans: In working with this code I was a little confused as to why we would want to track the modal closing on button clicks considering that pressing any of the button options effectively closes the modal entirely. I'm not sure how useful that data will be if that action happens no matter what button is pressed? Pressing X or clicking away are different actions, so it makes sense to include tracking in the onClose().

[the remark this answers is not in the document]
TMcMeans: To resolve this, I think @rquartararo point about refactoring the `handleClose()` method to accept an optional argument specifying the source of the action makes sense here to maintain consistency.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -75,6 +83,14 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+      label: "X",
```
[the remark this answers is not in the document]
TMcMeans: in this case, if I consider both pressing "X" and clicking away from the modal the same action, is it okay to specify null as the `option` and `label`? The other reason why I am conflating the two is that its becoming difficult to figure out where to make the distinction in` handleClose()`. I honestly can't tell when the modal x button is being pressed or clicked away from in the code.

src/Apps/Order/Components/AddressVerificationFlow.tsx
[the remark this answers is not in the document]
TMcMeans: Also, to deal with the inconsistencies with calling handleClose() I am investigating why one modal type calls `onClose()` but the other calls `handleClose()`

TMcMeans:
There were some conflicts that I resolved in the github visual editor. I was as surgical as possible, and am hoping that I didn't make any mistakes in fixing the conflicts overall.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -166,13 +244,30 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+                  user_id: user?.id,
+                  subject: "Confirm your delivery address",
+                  option: selectedAddressKey.includes("suggestedAddress")
+                    ? "Recommended"
+                    : "What you entered",
+                  label: "Use This Address",
+                })
+              }
+
+              chooseAddress()
```
starsirius, line 269: I see inconsistent order of `chooseAddress` and `trackEvent` between here and the `REVIEW_AND_CONFIRM` below. I assume we'll want to leave tracking at the last step of the handling?

starsirius:
It's looking great!! 💪

src/Apps/Order/Components/AddressVerificationFlow.tsx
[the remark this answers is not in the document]
daytavares: Agreed! So if option and label are `null` , I can assume the user didn't click on the buttons but closed the modal?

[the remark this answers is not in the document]
TMcMeans: yes, I made an assumption and chose `null` as the qualifier here, hope that's okay!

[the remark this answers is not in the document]
daytavares: Adding [here](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0) to make sure we are documenting it.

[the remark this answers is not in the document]
TMcMeans: Sorry, actually I chose "X" but let me know if that is inconsistent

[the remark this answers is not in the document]
TMcMeans: I can switch both option and label to `null` instead of `X` to reflect the documentation. Plus, `null` makes it easier to determine that is represents both an X and click away.

src/Apps/Order/Components/AddressVerificationFlow.tsx
iskounen: My preference has been to `trackEvent` before triggering the event handler (`chooseAddress`, `onClose`, etc.) so that we can capture the user's intention before performing an action that has side-effects. I am worried about the case where that action errors-out or the side-effects cause the tracking event to not be fired.

I don't feel strongly about this preference as long as we are consistent, so I agree that the call order of one or the other should be updated.

iskounen (approved):
LGTM

src/Apps/Order/Components/AddressVerificationFlow.tsx
TMcMeans: I'm going to maintain the order that @iskounen is talking about here to be able to capture the user intent one line of code before the user action is taken for all tracking calls in this PR before merging. This can be easily reversible if not an ideal setup also, and I can QA on staging to make sure. 

IMO it does make more sense to track AFTER use action, but aren't there also scenarios where this can lead to inconsistent if the user action is inconsistent?? 

However, @damassi or @dzucconi wondering if either of you have preferences on where in the logic flow tracking calls should be placed in Force, before OR after user actions?? Any input on this?

src/Apps/Order/Components/AddressVerificationFlow.tsx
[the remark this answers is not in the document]
TMcMeans: My most recent commit reverted option and label for modal closing tracking actions to `null` so going to resolve this conversation if no objections.

src/Apps/Order/Components/AddressVerificationFlow.tsx
dzucconi: I've used enough checkouts with ad blockers installed to understand that tracking calls should always happen after any logic executes. In our case this shouldn't matter but to me it makes sense that you'd only want to track something once it happened.

damassi: For something like a click event I would do something like this:

```ts
<Button onClick={() => {
  tracking.trackEvent(...)
  
  doSomethingElse()
}}>Click Me</Button>
```

Then we can be sure that the tracking event fired before anything else happened. I wouldn't overthink it too much or code too defensively here. If something fails in `doSomethingElse` it could mess up the stats, but that's a bug and hopefully it would be caught and addressed quickly and averaged out as a deviation.

TMcMeans: I've got tracking being called directly after button clicks and when onClose() handlers are being called exactly like how you have it described @damassi, but to @dzucconi's point it's important to make sure to track an action after it was taken regardless. Thank you for the input!

damassi: Ah yes, @dzucconi does make a very good point about ad blockers etc. I'm sure its all good here tho

PR #12625, opened 2023-07-19
feat(alerts): update confirmation messages for various cases

anandaroop:

This PR solves [ONYX-148]

### Description

Per the discussion on ONYX-15 this updates the copy for the various cases that might be encountered in the Alert confirmation modal: many matches, few matches, one match, no matches.

|Many|Few|One|None|
|---|---|---|---|
|many|few|one||

[ONYX-148]: https://artsyproduct.atlassian.net/browse/ONYX-148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/SavedSearchAlert/Components/__tests__/ConfirmationModalHeader.jest.tsx
+++ b/src/Components/SavedSearchAlert/Components/__tests__/ConfirmationModalHeader.jest.tsx
@@ -41,7 +41,9 @@ describe("ConfirmationModalHeader", () => {
     )
 
     expect(
-      screen.getByText("We will notify you when new works get added to Artsy.")
+      screen.getByText(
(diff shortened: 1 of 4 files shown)


anandaroop:
question (non-blocking): A separate thing I was wondering about @nickskalkin — do you think we should leave the **See all matching works** button in place always? My feeling was that it only really provided value in the first (many matches) case, therefore on Eigen I suppressed it for the other cases.

nickskalkin:
@anandaroop I think we should hide "Se all matching works" button conditionally like you did in Eigen, I will prepare a PR now, thank you!

nickskalkin (approved):

PR #12627, opened 2023-07-20
fix(onyx-147): fix impossible to navigate to a See full results option using keyboard arrows

nickskalkin:

This PR solves [ONYX-147]

### Description

https://github.com/artsy/force/assets/3934579/be85378c-6d55-494b-911e-d995a79b37c7

[ONYX-147]: https://artsyproduct.atlassian.net/browse/ONYX-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/NewSearch/NewSearchBarFooter.tsx
+++ b/src/Components/Search/NewSearch/NewSearchBarFooter.tsx
@@ -11,14 +11,12 @@ interface SuggestionItemProps {
   href: string
   query: string
   selectedPill: PillType
-  onClick: () => void
 }
 
 export const NewSearchBarFooter: FC<SuggestionItemProps> = ({
   href,
   query,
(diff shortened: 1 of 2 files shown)


nickskalkin:
There is a new "bug" I am introducing in this PR: when user clicks on the "See full results..." option - search input field is cleared, but should still include typed text. This probably happens because of this [resetUI()](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx#L188) call in palette. @dzucconi you think we could make this call optional based on some argument passed to `AutocompleteInputOption` component?

PR #12629, opened 2023-07-20
fix: only perform address verification once

iskounen:
The type of this PR is: Fix
This PR solves [EMI-1302]

### Description

This PR updates the address verification flow to only happen once. Previously, users could close the address verification modal and try again with a modified address. However, that creates a lot of noise and racks up additional usage charges from Smarty, so we are limiting it to 1 address verification per visit to the Shipping step of the checkout flow.

**Note:** If the user proceeds to the Payment step and returns to the Shipping step to edit their address, verification will be performed again.

[EMI-1302]: https://artsyproduct.atlassian.net/browse/EMI-1302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -107,10 +107,14 @@ export const ShippingRoute: FC<ShippingProps> = props => {
 
   // true if we want to verify addresses for this order
   const isAddressVerificationEnabled = useFeatureFlag("address_verification")
-  // true if the current address has been verified
-  const [addressVerified, setAddressVerified] = useState<boolean>(false)
   // true if the current address needs to be verified
-  const [verifyAddress, setVerifyAddress] = useState<boolean>(false)
+  const [addressNeedsVerification, setAddressNeedsVerification] = useState<
+    boolean
+  >(false)
+  // true if the current address has been verified
+  const [addressHasBeenVerified, setAddressHasBeenVerified] = useState<boolean>(
+    false
+  )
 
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -768,6 +768,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                   }}
                   onClose={() => {
                     setVerifyAddress(false)
+                    setAddressVerified(true)
```
erikdstock, line 771: #minor & #non-blocking
I think these two stateful variables `verifyAddress` and `addressVerified` ([here](https://github.com/artsy/force/blob/52019141b7534d4c9fbf69c334858947be56bfc1/src/Apps/Order/Routes/Shipping/index.tsx#L110-L113)) are a little bit confusing, if only in their naming. They look like they are the inverse of each other (like they could be called `shouldVerifyAddress` and `didVerifyAddress`), and for the latter it's not clear to me if this means that we called for verification or that it was successfully verified. I wonder if we could simplify this down to one var or make the naming more clear.

TMcMeans: agreed, I was slightly confused about how each method functions based on naming as well.

iskounen: I like the `shouldVerifyAddress` and `didVerifyAddress` variation.

PR #12636, opened 2023-07-21
chore: test address verification

iskounen:
The type of this PR is: Chore
This PR solves [EMI-1299]

### Description

This PR adds tests to the work we did in #12608 to add address verification to the Shipping step of the checkout flow.


src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -0,0 +1,38 @@
...
+      @relay_test_operation {
+      verifyAddress(address: $address) {
+        ...AddressVerificationFlow_verifyAddress
+      }
+    }
+  `,
+})
+
+describe("AddressVerificationFlow", () => {
+  it(`renders the "Check your delivery address" variant when verificaiton is "NOT_FOUND"`, async () => {
```
TMcMeans, line 21: **nitpick [blocking]:** 

There is a typo in the test description: "verificaiton" should be "verification"

TMcMeans (approved):
got one or two small nitpicks but otherwise LGTM 👍🏾 🥇

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -0,0 +1,38 @@
...
+      VerifyAddressType: () => ({
+        verificationStatus: "NOT_FOUND",
+      }),
+    })
+
+    await screen.findByText("Check your delivery address")
+
+    expect(
+      screen.getByText(
+        "The address you entered may be incorrect or incomplete. Please check it and make any changes necessary."
```
TMcMeans, line 32: **question [non-blocking]:** 

One more thing I noticed, are you expecting all elements to be present in the DOM immediately after rendering? If so, it would be consistent to use getByText() for all of them. However, if some elements might not be immediately available in the DOM, using findByText() for those would be more appropriate in that case.

iskounen: You're right - the text that I'm expecting here is not available on the initial render. They are added in the `useEffect` hook that fires after that initial render. I have added an `await screen.findByText(...)` for these tests to wait for that hook to fire before making these assertions.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -15,7 +15,7 @@ import {
 import { useCallback, useEffect, useState } from "react"
 
 interface AddressVerificationFlowProps {
-  verifiedAddressResult: AddressVerificationFlow_verifiedAddressResult$data
+  verifyAddress: AddressVerificationFlow_verifyAddress$data
```
iskounen, line 18: I renamed this while trying to get the test wrapper to render this component. I observed that it was expecting the name of the prop to match the name of the GraphQL query, so I assumed that was a convention and it worked.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -47,19 +47,19 @@ enum AddressSuggestionRadioButton {
...
-  const [selectedAddressKey, setSelectedAddressKey2] = useState<string | null>(
+  const [selectedAddressKey, setSelectedAddressKey] = useState<string | null>(
     null
   )
 
-  const setSelectedAddressKey = (key: string) => {
-    setSelectedAddressKey2(key)
-  }
+  const emptyDiv: JSX.Element = (
+    <div data-testid="emptyAddressVerification"></div>
```
iskounen, line 61: This allows our tests to await the initial render when the `useEffect` hooks are called, which in turn allows them to make assertions about changes made by those hooks.

TMcMeans:
Just noticed that after I merged my tracking code pr that there are some merge conflicts, let me know if I can help resolve them. Since this is mostly spec code w/ the exception to the changes to `verifyAddress` in the flow, I think we may also want to mock out tracking calls in the tests as well.

erikdstock:
closed as this work was integrated with #12689

PR #12638, opened 2023-07-21
fix(toasts): adjust position when cookie consent banner is displayed

anandaroop:

This PR solves [ONYX-161]

### Description

Since `Toasts` are now bottom-aligned in Force there is an edge case interaction between the `Toast` and `CookieConsentBanner` components.

## Before

Note that in the right-hand window, the cookie banner prevents the toasts from being visible.

https://github.com/artsy/force/assets/140521/68eecfd0-1ef7-4c87-b81c-c21be00d434f

## After

Change to the toasts’ z-index ensures they are above the banner.

Screenshot 2023-07-21 at 5 11 23 PM

[ONYX-141]: https://artsyproduct.atlassian.net/browse/ONYX-141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ONYX-161]: https://artsyproduct.atlassian.net/browse/ONYX-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Components/constants.ts
+++ b/src/Apps/Components/constants.ts
@@ -2,6 +2,6 @@ export const Z = {
   dropdown: 200,
   globalNav: 100,
   popover: 75,
-  toasts: 50,
+  toasts: 150,
   footer: 1,
 }


src/Apps/Components/AppToasts.tsx
```diff
@@ -29,3 +32,16 @@ export const AppToasts: React.FC = () => {
...
+ * Check if Toasts' bottom offset needs to be adjusted due to presence of
+ * cookie consent banner.
+ */
+function useBottomOffset(): number {
+  const { isDisplayable } = useConsentRequired()
+
+  if (isServer || !isDisplayable) return 0
+
+  const bannerElement = document.querySelector("#cookie-consent-banner")
+  return bannerElement?.getBoundingClientRect()?.height ?? 0
```
anandaroop, line 46: This approach seemed simple enough, but lmk if you have any concerns.

src/Apps/Components/AppToasts.tsx
```diff
@@ -29,3 +32,16 @@ export const AppToasts: React.FC = () => {
...
 }
+
+/**
+ * Check if Toasts' bottom offset needs to be adjusted due to presence of
+ * cookie consent banner.
+ */
+function useBottomOffset(): number {
+  const { isDisplayable } = useConsentRequired()
+
+  if (isServer || !isDisplayable) return 0
```
dzucconi, line 43: The use of `isServer` here may cause a server/client mismatch. We should set the height in a `useEffect`

dzucconi (dismissed):
This works for me. I was going to suggest just to adjust the z-indexes so that the toasts appeared over the consent banner. But this is fine, just adjust so that the height is set client-side.

anandaroop:
>  I was going to suggest just to adjust the z-indexes

Yep, this is better, should have gone with [my first instinct](https://artsy.slack.com/archives/C05EQL4R5N0/p1689948421425869?thread_ts=1689948390.787009&cid=C05EQL4R5N0) lol

PR #12640, opened 2023-07-24
fix(onyx-145): update the search input field when scrolling through options using keyboard arrows

nickskalkin:

This PR solves [ONYX-145]

### Description

https://github.com/artsy/force/assets/3934579/83a5e67c-f937-47ed-a2ea-9d194159be08

[ONYX-145]: https://artsyproduct.atlassian.net/browse/ONYX-145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/NewSearch/NewSearchBar.tsx
+++ b/src/Components/Search/NewSearch/NewSearchBar.tsx
@@ -3,19 +3,23 @@ import { NewSearchBarInputQueryRenderer } from "Components/Search/NewSearch/NewS
 import { Media } from "Utils/Responsive"
 import { MobileSearchBarQueryRenderer } from "./Mobile/MobileSearchBar"
 import { FC } from "react"
+import { getSearchTerm } from "Components/Search/NewSearch/utils/getSearchTerm"
+import { isServer } from "Server/isServer"
 
 interface NewSearchBarProps {
   onClose: () => void
 }
 
 export const NewSearchBar: FC<NewSearchBarProps> = ({ onClose }) => {
+  const urlSearchTerm = isServer ? undefined : getSearchTerm(window?.location)
+
   return (
     <Box flex={1}>
       <Media at="xs">
         <MobileSearchBarQueryRenderer onClose={onClose} />
       </Media>
       <Media greaterThan="xs">
-        <NewSearchBarInputQueryRenderer />
+        <NewSearchBarInputQueryRenderer term={urlSearchTerm} />
       </Media>
     </Box>
   )
--- a/src/Components/Search/NewSearch/NewSearchBar.tsx
+++ b/src/Components/Search/NewSearch/NewSearchBar.tsx
@@ -3,15 +3,14 @@ import { NewSearchBarInputQueryRenderer } from "Components/Search/NewSearch/NewS
 import { Media } from "Utils/Responsive"
 import { MobileSearchBarQueryRenderer } from "./Mobile/MobileSearchBar"
 import { FC } from "react"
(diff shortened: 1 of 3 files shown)


src/Components/Search/NewSearch/NewSearchBar.tsx
```diff
@@ -3,19 +3,23 @@ import { NewSearchBarInputQueryRenderer } from "Components/Search/NewSearch/NewS
...
 import { FC } from "react"
+import { getSearchTerm } from "Components/Search/NewSearch/utils/getSearchTerm"
+import { isServer } from "Server/isServer"
 
 interface NewSearchBarProps {
   onClose: () => void
 }
 
 export const NewSearchBar: FC<NewSearchBarProps> = ({ onClose }) => {
+  const urlSearchTerm = isServer ? undefined : getSearchTerm(window?.location)
```
nickskalkin, line 14: This change is not related to the PR's title. It fixes the case when user navigates to the `search?term=something` page and click on the search input. Search should be already pre-populated at this point.

https://github.com/artsy/force/assets/3934579/ac1b48f0-f259-406e-afc3-279b44533b03

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -168,8 +168,8 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
       query: value,
     })
 
-    clearSearchInput()
-    redirect(option.href)
+    resetValue()
+    window.location.href = option.href
```
dzucconi, line 172: Why is there a hard page load? Shouldn't this be using our router?

(Also does cmd+click still work with these?)

src/Components/Search/NewSearch/NewSearchBarInput.tsx
```diff
@@ -184,9 +184,9 @@ const NewSearchBarInput: FC<NewSearchBarInputProps> = ({ relay, viewer }) => {
       placeholder={t`navbar.searchBy`}
       spellCheck={false}
       options={shouldStartSearching(value) ? formattedOptions : []}
-      value={value}
+      defaultValue={value}
```
dzucconi, line 187: 👍

src/Components/Search/NewSearch/NewSearchBar.tsx
dzucconi: I really need to change the name of `isServer` to: `__DO_NOT_USE_OUTSIDE_OF_EFFECT_isServer__` or something. You can't call this in a render function; will cause a mismatch

dzucconi: Or better, I guess I can wrap this in a hook.

src/Components/Search/NewSearch/NewSearchBarInput.tsx
nickskalkin: > (Also does cmd+click still work with these?)

Actually no! It doesn't work on current prod implementation as well, but would really nice to have. I'll create a ticket for that

dzucconi: Yeah, it can be a separate ticket but basically you just detect if the meta key is pressed in the `event` and just return early if it is.

dzucconi (approved):

MounirDhahri:
LGTM

PR #12652, opened 2023-07-26
fix(EMI-1321): address verify only on create

leamotta:

https://artsyproduct.atlassian.net/browse/EMI-1321

### Description

As a first step in the address verification flow, we should only verify addresses when the user has none and they are creating their first one. If the user already has at least an address in our system, we won't validate existing or new ones they create.

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -224,8 +224,8 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       isCreateNewAddress()
     ) {
       /**
-       * Setting verifyAddress to true will cause the address verification flow
-       * to be initiated on this render.
+       * Setting addressNeedsVerification to true will cause the address
+       * verification flow to be initiated on this render.
        */
       setAddressNeedsVerification(true)
       return
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -113,7 +113,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
   >(false)
   // true if the current address has been verified
   const [addressHasBeenVerified, setAddressHasBeenVerified] = useState<boolean>(
-    true
+    false
   )
 
   const [shippingOption, setShippingOption] = useState<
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -218,10 +218,14 @@ export const ShippingRoute: FC<ShippingProps> = props => {
   }
 
   const onContinueButtonPressed = async () => {
-    if (isAddressVerificationEnabled && !addressHasBeenVerified) {
+    if (
+      isAddressVerificationEnabled &&
+      !addressHasBeenVerified &&
+      isCreateNewAddress()
```
starsirius, line 224: #non-blocking. This seems sufficient for now to fix the issue, but I'm a bit worried a regression can happen in the future when the implementation of this page changes.

For example, clicking the "Add a new address" button is another way to create a new address. It's not an issue here because it's in a modal, and users always save the address first before they can click continue. If it were implemented as another inline form, then I think checking if the address is new isn't sufficient to prevent verification from happening for that form.

I think we need a way to enable address verification only when the address was submitted from that particular form. It's still not perfect, but I wonder if additionally checking [`showAddressForm`](https://github.com/artsy/force/blob/be039d10ffcacda5cca1a2d9e95c57aea9204f24/src/Apps/Order/Routes/Shipping/index.tsx#L765) (which is true when the particular form is opened) makes it safer.

Screen Shot 2023-07-26 at 9 53 15 PM

starsirius (dismissed):
Had a non-blocking comment. Thank you for jumping in to help! 🙇 

Merging https://github.com/artsy/force/pull/12654 created merge conflicts and there is failing tests. Feel free to merge when they are resolved!

src/Apps/Order/Routes/Shipping/index.tsx
leamotta: I agree this solution doesn't scale to a scenario where we want to validate addresses when the user already has at least one stored, syncing with @iskounen we agreed that for the first iteration we'd validate only in the scenario where the user has no addresses.

For the suggestion, I believe adding a check on `showAddressForm` here won't help, since by the time the user hits "Save and Continue" it wouldn't be displaying anymore. Maybe adding the address validation logic to the "Save" button in the modal would be a good solution, where we check the address as soon as the user wants to save it. We can discuss and create another ticket for that.

PR #12660, opened 2023-07-28
fix: avoid double triggering address verification modal

starsirius:

This PR solves [EMI-1311](https://artsyproduct.atlassian.net/browse/EMI-1311)

### Description

The address verification modal gets triggered again after user choosing an address from it. Setting states is async. Checking if the verification needs to run right after the async set states will get the old value of the states, which triggers the address verification again.

Since on choosing an address we are ready to proceed to the next step, this directly move on to finalizing fulfillment, instead of checking again if address verification needs to run.

#### Before

double-modal-before

#### After

double-modal

[EMI-1311]: https://artsyproduct.atlassian.net/browse/EMI-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -228,6 +228,17 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       : addressVerificationIntlEnabled
   }
 
+  // Save shipping info on the order. If it's Artsy shipping and a quote hasn't
+  // been selected, this renders the quotes for user to select and finalize
+  // again.
+  const finalizeFulfillment = async () => {
+    if (checkIfArtsyShipping() && !!shippingQuoteId) {
+      selectShippingQuote()
+    } else {
+      selectShipping()
+    }
+  }
+
   const onContinueButtonPressed = async () => {
     if (
       isAddressVerificationEnabled() &&
@@ -242,11 +253,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       return
     }
 
-    if (checkIfArtsyShipping() && !!shippingQuoteId) {
-      selectShippingQuote()
-    } else {
-      selectShipping()
-    }
+    finalizeFulfillment()
   }
 
   const selectShipping = async (editedAddress?: MutationAddressResponse) => {
@@ -793,7 +800,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                     setAddressNeedsVerification(false)
                     setAddressHasBeenVerified(true)
                     setAddress({ ...address, ...chosenAddress })
-                    onContinueButtonPressed()
+                    finalizeFulfillment()
                   }}
                 />
               )}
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -793,7 +800,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                     setAddressNeedsVerification(false)
                     setAddressHasBeenVerified(true)
                     setAddress({ ...address, ...chosenAddress })
-                    onContinueButtonPressed()
+                    finalizeFulfillment()
```
starsirius, line 803: Alternatively we can use `useEffect` to wait for `addressHasBeenVerified` state to be updated. However, I decided to avoid it because [we might not need it](https://react.dev/learn/you-might-not-need-an-effect).

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -228,6 +228,17 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       : addressVerificationIntlEnabled
   }
 
+  // Save shipping info on the order. If it's Artsy shipping and a quote hasn't
+  // been selected, this renders the quotes for user to select and finalize
+  // again.
```
starsirius, line 233: This step is actually pretty complicated (and so hard to name it) and can use some refactor. I'm documenting it first and would like to do another followup.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -1315,10 +1323,14 @@ describe("Shipping", () => {
     describe("with address verification enabled", () => {
       beforeAll(() => {
         ;(useFeatureFlag as jest.Mock).mockImplementation(
-          (featureName: string) => featureName === "address_verification"
+          (featureName: string) => featureName === "address_verification_us"
```
starsirius, line 1326: The test update is not relevant to the change, and I'll need to follow up with a proper test.

This uses the correct feature flag name, and after correcting, tests are failing. It turns out the mock is leaking to other tests so I reset them.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -242,11 +253,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       return
     }
 
-    if (checkIfArtsyShipping() && !!shippingQuoteId) {
-      selectShippingQuote()
-    } else {
-      selectShipping()
-    }
+    finalizeFulfillment()
```
leamotta, line 256: nice!

leamotta (approved):

iskounen (approved):
Elegant solution to an ugly problem :)

erikdstock (approved):
Looks good, thank you!

PR #12664, opened 2023-07-28
chore: prefers main over normalized

dzucconi:
[DIA-80](https://artsyproduct.atlassian.net/browse/DIA-80)

Some context here: https://artsy.slack.com/archives/C2TQ4PT8R/p1690386360841419

Since `main` has an upper limit on its dimensions, it should be more performant on images that exceed `5120px` on a single side. This is also plenty sufficient for our current use cases to pull resizes from.

[DIA-80]: https://artsyproduct.atlassian.net/browse/DIA-80?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anandaroop:
🎉

PR #12673, opened 2023-07-31
fix: do not proceed automatically after choosing a verified address

starsirius:

This PR solves [EMI-1357](https://artsyproduct.atlassian.net/browse/EMI-1357), [EMI-1320](https://artsyproduct.atlassian.net/browse/EMI-1320)

### Description

In https://github.com/artsy/force/pull/12660, we fixed the double address verification modal issue by synchronously proceeding to finalizing fulfillment step. However, setting the chosen address as a state is also async. Finalizing fulfillment synchronously will be using the old address. As a result, although the recommended address seemed to be populated on the form, the old address is actually used in the `setShipping` mutation and eventually persisted.

I explored some options such as passing down the chosen address to `finalizeFulfillment` or entirely relying on states and `useEffect` for automatically proceeding after choosing an address, but they are all deeply involved given the amount of states and hooks on this page. Instead of adding more complexity for this 🐛  and introducing more hidden issues (like incorrectly requesting Arta quotes), I decided to not automatically advance to the next step.

This will require users to click "Save and continue" after selecting a verified address, and I created a [followup ticket](https://artsyproduct.atlassian.net/browse/EMI-1359) to revisit. We might want to also refactor this page to simplify the implementation.

#### Before and after

<details><summary>Before</summary>
<p>

recommended-address-not-persisted (1)

</p>
</details> 

<details><summary>After</summary>
<p>

persist-address

</p>
</details> 

[EMI-1357]: https://artsyproduct.atlassian.net/browse/EMI-1357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMI-1320]: https://artsyproduct.atlassian.net/browse/EMI-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/AddressForm.tsx
```diff
@@ -175,7 +175,7 @@ export const AddressForm: React.FC<AddressFormProps> = ({
           id="AddressForm_addressLine2"
           placeholder="Apt, floor, suite, etc."
           title="Address line 2 (optional)"
-          value={value?.addressLine2}
+          value={value?.addressLine2 || ""}
```
starsirius, line 178: Pushed a quick fix for the [non-removed value](https://artsyproduct.atlassian.net/browse/EMI-1320) issue.

starsirius:
Re-assigning this to @leamotta to help while folks are out/unavailable. Thank you! 🙏

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -814,7 +814,6 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                     setAddressNeedsVerification(false)
                     setAddressHasBeenVerified(true)
                     setAddress({ ...address, ...chosenAddress })
```
damassi, line 816: This actually came up earlier today. Have you tried 
```
setAddress(address => ({
  ...address,
  ...chosenAddress
}))
```

Using the callback form for setting state should ensure that the previous state is updated correctly. (See [this thread](https://artsy.slack.com/archives/C2TQ4PT8R/p1690812245052079?thread_ts=1690811471.733219&cid=C2TQ4PT8R))

leamotta: This didn't work in my testing so I'm going with the proposed solution and we can revisit in the follow up ticket

leamotta (approved):
LGTM!

src/Apps/Order/Routes/Shipping/index.tsx
damassi: Thanks for giving it a shot 👍

PR #12677, opened 2023-08-01
feat: add verified representatives to Artist header

egdbear:

This PR solves [DIA-66]

### Description
It adds Represented By Gallery to Artist Page Header. 

Screenshot 2023-08-01 at 12 21 06
Screenshot 2023-08-01 at 12 20 56

cc @artsy/diamond-devs 

[DIA-66]: https://artsyproduct.atlassian.net/browse/DIA-66?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Components/ArtistHeader/ArtistHeader2.tsx
+++ b/src/Apps/Artist/Components/ArtistHeader/ArtistHeader2.tsx
@@ -12,7 +12,7 @@ import {
   Spacer,
   Text,
 } from "@artsy/palette"
-import * as React from "react"
+import { Fragment } from "react"
 import { ContextModule } from "@artsy/cohesion"
 import { createFragmentContainer, graphql } from "react-relay"
 import { FollowArtistButtonQueryRenderer } from "Components/FollowButton/FollowArtistButton"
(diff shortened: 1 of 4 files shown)


src/Apps/Artist/Components/ArtistHeader/ArtistHeader2.tsx
```diff
@@ -303,4 +317,19 @@ const formatFollowerCount = (n: number) => {
...
+    return (
+      <Fragment key={rep.slug}>
+        {index > 0 && ", "}
+        <RouterLink to={`/partner/${rep.slug}`} color="black100">
+          {rep.name}
+        </RouterLink>
+      </Fragment>
+    )
+  })
+}
```
dzucconi, line 333: This function is unnecessary, just inline this. If you wanted to extract it would be a component and fragment container. But not needed in this case.

dzucconi (commented):
What does this look like with a lot of galleries? What’s the worst case scenario?

egdbear:
> What does this look like with a lot of galleries? What’s the worst case scenario?


I don't think we will have more than 3 gallery representatives in that case we could add MAX constant depending on the viewport. 

This is how it looks in _worst case_

Screenshot 2023-08-01 at 13 29 42
Screenshot 2023-08-01 at 13 29 37

dzucconi (approved):
👍

PR #12688, opened 2023-08-02
feat: show error message when setting payment by setup intent fails (EMI-1284)

iskounen:
The type of this PR is: Feat
This PR solves [EMI-1284]

### Description

This is a companion PR to artsy/exchange#1733 that displays an error message to users if it doesn't pass backend validation in Exchange. (See the companion PR in Exchange for more context.)

[EMI-1284]: https://artsyproduct.atlassian.net/browse/EMI-1284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/index.tsx
+++ b/src/Apps/Order/Routes/Payment/index.tsx
@@ -136,7 +136,9 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
   const displayLoading =
     isProcessingRedirect ||
     isSavingPayment ||
-    (!!match?.location?.query?.setup_intent && !bankAccountHasInsufficientFunds)
+    (!!match?.location?.query?.setup_intent &&
+      !bankAccountHasInsufficientFunds &&
+      !paymentSetupErrorCode)
 
   let routeSteps
   if (order.mode === "OFFER") {
@@ -341,12 +343,12 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
     ) {
       handlePaymentStepComplete()
     } else if (paymentSetupErrorCode && !isPaymentSetupSuccessful) {
-      setIsSavingPayment(false)
-      props.router.push(`/orders/${props.order.internalID}/payment`)
+      // setIsSavingPayment(false)
+      // props.router.push(`/orders/${props.order.internalID}/payment`)
 
-      const title = "Unsupported country"
+      const title = "Choose another payment method"
       const message =
-        "We are unable to process payments from SEPA accounts in your country at the moment."
+        "The bank account you entered is not denominated in EUR. Please select another payment method and try again."
       const error_code = paymentSetupErrorCode
 
       trackEvent({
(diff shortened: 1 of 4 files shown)


src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -339,13 +340,37 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
         selectedPaymentMethod === "SEPA_DEBIT")
     ) {
       handlePaymentStepComplete()
+    } else if (paymentSetupErrorCode && !isPaymentSetupSuccessful) {
+      setIsSavingPayment(false)
+      props.router.push(`/orders/${props.order.internalID}/payment`)
```
iskounen, line 345: I did this to reset the form when an error occurs (otherwise the loading screen was being shown indefinitely). There must be a better way...

erikdstock: I wonder if a single `resetForm` helper makes sense here

fladson (approved):

iskounen:
@starsirius this is a WIP because the copy in the error dialog needs to be updated, and the error recovery needs to be completed (ie. what happens after the error message is displayed).

src/Apps/Order/Routes/Payment/index.tsx
iskounen: That did the trick!! Thank you.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -339,13 +342,42 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
         selectedPaymentMethod === "SEPA_DEBIT")
     ) {
       handlePaymentStepComplete()
+    } else if (paymentSetupErrorCode && !isPaymentSetupSuccessful) {
```
starsirius, line 345: In case there is an error but no error code, we might still want to display the generic error message. Maybe this can just be
```suggestion
    } else if (!isPaymentSetupSuccessful) {
```

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -129,13 +129,16 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
...
   } = useStripePaymentBySetupIntentId(order.internalID)
 
   // SavingPaymentSpinner is rendered and PaymentContent is hidden when true
   const displayLoading =
     isProcessingRedirect ||
     isSavingPayment ||
-    (!!match?.location?.query?.setup_intent && !bankAccountHasInsufficientFunds)
+    (!!match?.location?.query?.setup_intent &&
+      !bankAccountHasInsufficientFunds &&
+      !paymentSetupErrorCode)
```
starsirius, line 141: It's not immediately clear what condition this represents. Can we make a comment about it?

starsirius (dismissed):
Had a few non-blocking comments. Feel free to merge when ready! 🎉  Re-assigning it back to you!

PR #12689, opened 2023-08-03
chore: EM-1355 - send address verified by

erikdstock:

cc @artsy/emerald-devs 
This PR solves [EM-1295] and mostly completes  [EM-1355]. It includes an update of the changes in https://github.com/artsy/force/pull/12636

### Description

With this change the `address_verified_by` argument is sent to [exchange's `setShipping` mutation](https://github.com/artsy/exchange/blob/b9079eb9d9bb5b157fb55f8d67b950e9f3ab2a73/app/graphql/mutations/set_shipping.rb#L30), triggering the backend address verification job.

It also fixes some types and cleans up some development code.

To accomplish this I updated our boolean `addressHasBeenVerified` state variable to be a `null | "ARTSY" | "USER"` `addressVerifiedBy` variable - its presence means that it has been verified, and editing the form after saving sets the address as verified by the user (since we don't perform verification more than once without a reload).

**Note/followup:** After the address has a verified-by value in exchange, it [will not be re-set](https://github.com/artsy/exchange/blob/b9079eb9d9bb5b157fb55f8d67b950e9f3ab2a73/app/services/shipping/verify_address.rb#L6). Thus if a user goes back and edits their address neither this nor the payload will be updated. Fixed in https://github.com/artsy/exchange/pull/1735

### Steps to test manually
1. Have an exchange staging console open
2. Begin an order checkout, removing all saved addresses
3. Enter a completely wrong US address for the delivery address and see the address verification modal - the version with no choices.
4. Click to keep the address
5. Click save & continue: 

**Find the order in the exchange staging console, see that the shipping address has been set along with `shipping_address_verified_by: "USER"` but no `shipping_address_verified_payload`.**

6. Reload the page
7. Enter a new address that will get suggestions, (401 broadway, 26th floor, new york, NY 10013 with the floor number on line 2 usually works).
8. Select the suggested address ['use this address']
9. Click save & continue again

**Reload the order in the exchange staging console, see that the shipping address has been updated along with `shipping_address_verified_by: "ARTSY"` and a `shipping_address_verified_payload`.**

10. On the shipping page with an address that has already been saved as verified by ARTSY (e.g. an Arta order that has not had its shipping option selected yet), edit the form to change something and save it again.

**Reload the order in exchange once more and see that the address has updated and is again verified by USER.**

Other expected behavior:
- User closes either modal without actively clicking the primary address select button: `address_verified_by USER`
- address verification not enabled: `address_verified_by: nil`

<details><summary>Saved address verification payload in exchange for an ARTSY-verified address</summary>

image

</details>

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -217,6 +217,8 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
   if (verificationStatus === "VERIFIED_NO_CHANGE")
     return <div data-testid="emptyAddressVerification"></div>
 
+  if (addressOptions.length === 0) return null
+
   if (modalType === ModalType.SUGGESTIONS) {
     return (
       <ModalDialog title="Confirm your delivery address" onClose={handleClose}>
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -18,7 +18,10 @@ import { useSystemContext } from "System/SystemContext"
 import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
 
-export type AddressVerifiedBy = "USER" | "ARTSY"
+export enum AddressVerifiedBy {
+  USER = "USER",
+  ARTSY = "ARTSY",
+}
 type AddressOptionKey = "userInput" | string
 
 interface AddressVerificationFlowProps {
@@ -80,7 +83,10 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
       option => option.key === selectedAddressKey
     )
 
-    const verifiedBy = selectedAddressKey === "userInput" ? "USER" : "ARTSY"
+    const verifiedBy =
+      selectedAddressKey === "userInput"
+        ? AddressVerifiedBy.USER
+        : AddressVerifiedBy.ARTSY
 
     if (selectedAddress) {
       setModalType(null)
@@ -156,7 +162,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
     }
 
     if (verificationStatus === "VERIFIED_NO_CHANGE") {
-      const verifiedBy = "ARTSY"
+      const verifiedBy = AddressVerifiedBy.ARTSY
       onChosenAddress(verifiedBy, inputOption.address as AddressValues, true)
     } else {
       if (verificationStatus === "VERIFIED_WITH_CHANGES") {
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -1,7 +1,7 @@
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
 import { createFragmentContainer, graphql } from "react-relay"
 import { AddressVerificationFlowQuery } from "__generated__/AddressVerificationFlowQuery.graphql"
-import { AddressVerificationFlow_verifiedAddressResult$data } from "__generated__/AddressVerificationFlow_verifiedAddressResult.graphql"
+import { AddressVerificationFlow_verifyAddress$data } from "__generated__/AddressVerificationFlow_verifyAddress.graphql"
 import {
   BorderedRadio,
   Box,
@@ -22,10 +22,11 @@ export enum AddressVerifiedBy {
   USER = "USER",
   ARTSY = "ARTSY",
 }
+
 type AddressOptionKey = "userInput" | string
 
 interface AddressVerificationFlowProps {
-  verifiedAddressResult: AddressVerificationFlow_verifiedAddressResult$data
+  verifyAddress: AddressVerificationFlow_verifyAddress$data
   onChosenAddress: (
     verifiedBy: AddressVerifiedBy,
     address: AddressValues,
@@ -61,7 +62,7 @@ enum AddressSuggestionRadioButton {
 }
 
 const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
-  verifiedAddressResult,
+  verifyAddress,
   onChosenAddress,
   onClose,
 }) => {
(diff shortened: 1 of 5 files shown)


src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -51,33 +62,36 @@ enum AddressSuggestionRadioButton {
 }
 
 const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
-  verifiedAddressResult,
+  verifyAddress,
```
erikdstock, line 65: This was one of the changes in https://github.com/artsy/force/pull/12636 which i thought were worthwhile.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -18,9 +18,20 @@ import { useSystemContext } from "System/SystemContext"
 import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
 
+export enum AddressVerifiedBy {
+  USER = "USER",
+  ARTSY = "ARTSY",
+}
+
+type AddressOptionKey = "userInput" | string
```
erikdstock, line 26: I'm not sure if this guarantees anything.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -203,6 +214,9 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
     contextPageOwnerSlug,
   ])
 
+  if (verificationStatus === "VERIFIED_NO_CHANGE")
+    return <div data-testid="emptyAddressVerification"></div>
+
   if (addressOptions.length === 0) return null
```
erikdstock, line 220: This part differs a bit from [how #12636 handled it](https://github.com/artsy/force/pull/12636/files#diff-a65d5b9bf3f3ba0af862f0c53c8ea2e0be243adf95b86daacdb768a05e3f538aR127-R129). I'm not sure which is right.

erikdstock, line 220: This got me to thinking about a bigger question: Do we need to do so much setup code in the [big useEffect here](https://github.com/artsy/force/pull/12689/files#diff-a65d5b9bf3f3ba0af862f0c53c8ea2e0be243adf95b86daacdb768a05e3f538aR158)? Or can we just load this all  outside of an effect on first render? I think this data should be available when the component is mounted.

erikdstock (commented):
Some comments.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -113,13 +119,17 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-  const [addressHasBeenVerified, setAddressHasBeenVerified] = useState<boolean>(
-    false
-  )
+
+  // Presence of addressVerifiedBy indicates that the address has been verified
+  // via the address verification flow.
+  const [
+    addressVerifiedBy,
+    setAddressVerifiedBy,
+  ] = useState<AddressVerifiedBy | null>(null)
```
erikdstock, line 132: feels like an opportunity for simplification here in the TODO.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -0,0 +1,264 @@
...
+    it('calls onChosenAddress with the suggested address when "Use This Address" is clicked', async () => {
+      renderWithRelay(
+        {
+          VerifyAddressType: () => mockResult,
+        },
+        undefined,
+        {
+          onChosenAddress: mockOnChosenAddress,
+          onClose: mockOnClose,
+        }
```
damassi, line 203: Nice, i've never seen these extra arguments used 👌

fladson (approved):

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -95,9 +109,9 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
   const handleBackToEdit = () => {
     const option =
-      selectedAddressKey === "suggestedAddress"
+      selectedAddressKey && selectedAddressKey.includes("suggestedAddress")
```
starsirius, line 112: 🤔  Was this a bug?

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -324,14 +334,19 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
 
+      const mutationInput: CommerceSetShippingInput = {
+        id: props.order.internalID,
+        fulfillmentType: isArtsyShipping ? "SHIP_ARTA" : shippingOption,
+        shipping: shipToAddress,
+        phoneNumber: shipToPhoneNumber,
+      }
+      if (addressVerifiedBy) {
+        mutationInput.addressVerifiedBy = addressVerifiedBy
+      }
```
starsirius, line 345: 👍

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -565,6 +580,12 @@ export const ShippingRoute: FC<ShippingProps> = props => {
 
     setShippingQuotes(null)
     setShippingQuoteId(undefined)
+
+    // If the address has already been verified and the user is editing the form,
+    // consider this a user-verified address (perform verification only once).
+    if (addressVerifiedBy) {
+      setAddressVerifiedBy(AddressVerifiedBy.USER)
+    }
```
starsirius, line 588: Good catch!

Unrelated to this PR: I feel the shipping step has become so complicated to map out what happens at what point. We should probably follow up and refactor/simplify things.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Looks like it, but would only affect tracking and still looks like it could be improved. i figured that could come after tests are in place.

src/Apps/Order/Routes/Shipping/index.tsx
erikdstock: oh, maybe i should add a test for this. Hoping we can do this in a followup on tests for the shipping step, but i noticed that still uses enzyme and i'm not sure of the current opinion of investing more there.

PR #12693, opened 2023-08-03
fix(ONYX-128): update Create Alert button with icon and add grid layout

tam-kis:
The type of this PR is: Fix 

This PR solves [ONYX-128](https://artsyproduct.atlassian.net/browse/ONYX-128)

### Description
Update Create Alert buttons to have the `BellStrokeIcon`. 
Add grid for title and button layout.

Screenshot 2023-08-03 at 15 07 50

[ONYX-128]: https://artsyproduct.atlassian.net/browse/ONYX-128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/ArtworkAuctionCreateAlertHeader.tsx
+++ b/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/ArtworkAuctionCreateAlertHeader.tsx
@@ -1,4 +1,4 @@
-import { Box, Text } from "@artsy/palette"
+import { Column, GridColumns, Text } from "@artsy/palette"
 import * as React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { ArtworkAuctionCreateAlertHeader_artwork$data } from "__generated__/ArtworkAuctionCreateAlertHeader_artwork.graphql"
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/ArtworkAuctionCreateAlertHeader.tsx
```diff
@@ -35,15 +35,18 @@ const ArtworkAuctionCreateAlertHeader: React.FC<ArtworkAuctionCreateAlertHeaderP
...
-    <Box py={12}>
-      <Text variant="lg" textAlign="center">
-        Bidding for <i>{artwork.title?.trim()}</i>
-        {artistName} has ended.
-      </Text>
-      <Box mt={2} mx="auto" width={["100%", 209]}>
+    <GridColumns py={6}>
+      <Column span={12}>
+        <Text variant="lg" textAlign="center">
+          Bidding for <i>{artwork.title?.trim()}</i>
```
nickskalkin, line 41: **suggestion**:  I would move the text to [translations](https://github.com/artsy/force/blob/main/src/System/i18n/locales/en-US/translation.json).

tam-kis: Should we also then update the button text then? (we have Create Alert in the translation file) 
I can add this to the next PR since I will have more text there.

nickskalkin: Yes, I think that we should add all new texts as translations

nickskalkin: I'm ok to go either way, but we need to be aligned which approach to take. I am asking to add translations because:

1. I was asked to do so many times.
2. I though there was an agreement to do it. 

Regarding backend projects @mzikherman, we had an RFC to adopt internationalization, and resolution was to, in fact, to [adopt it](https://github.com/artsy/README/issues/421#issuecomment-976407182).

I can't find an RFC for Force, maybe we should start one? 🤷

PR #12694, opened 2023-08-03
chore(hot-reloading): Bump related libs to latest

damassi:

### Description

There were a few issues reported with hot reloading. Was able to verify that it was leading to full reloads, and after bumping all of the libs to latest and running `yarn clean && yarn start` confirmed a fix.


dzucconi:
🫶

PR #12700, opened 2023-08-07
chore: add jest-watch-typeahead

erikdstock:

This PR adds jest-watch-typeahead which gives previews for what files will run when using pattern or test name matching with `jest. --watch`. It's a nicety but not necessary and may have been considered and rejected in the past. This is what it looks like:


damassi (approved):
Interesting -- i always thought this was built into the latest version of Jest? If not, no problem, and 👍 👍

PR #12703, opened 2023-08-08
feat(emi-1286): handle errors using the orError approach

fladson:

This PR solves [EMI-1286]

### Description

This is a followup of https://github.com/artsy/metaphysics/pull/5179.
The goal here is to update the code to use the `...orError` response format in order to handle errors more easily.

[EMI-1286]: https://artsyproduct.atlassian.net/browse/EMI-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -135,15 +135,23 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
   // Handling possibly incorrectly (by MP) possible null values
   const suggestedAddresses =
-    verifiedAddressResult.suggestedAddresses ||
+    verifiedAddressResult.verifyAddressOrError?.suggestedAddresses ||
     ([] as NonNullable<
-      NonNullable<AddressVerificationFlow_verifiedAddressResult$data>
+      NonNullable<
+        NonNullable<AddressVerificationFlow_verifiedAddressResult$data>
+      >["verifyAddressOrError"]
     >["suggestedAddresses"])
-  const inputAddress = verifiedAddressResult.inputAddress as NonNullable<
-    NonNullable<AddressVerificationFlow_verifiedAddressResult$data>
+  const inputAddress = verifiedAddressResult.verifyAddressOrError
+    ?.inputAddress as NonNullable<
+    NonNullable<
+      NonNullable<AddressVerificationFlow_verifiedAddressResult$data>
+    >["verifyAddressOrError"]
   >["inputAddress"]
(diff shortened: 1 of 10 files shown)


fladson:
1. Getting the code up to date with the local metaphysics schema ✅  SCR-20230808-maf

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -135,15 +135,23 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+      NonNullable<AddressVerificationFlow_verifiedAddressResult$data>
+    >["verifyAddressOrError"]
   >["inputAddress"]
-  const verificationStatus = verifiedAddressResult.verificationStatus as NonNullable<
-    AddressVerificationFlow_verifiedAddressResult$data
+  const verificationStatus = verifiedAddressResult.verifyAddressOrError
+    ?.verificationStatus as NonNullable<
+    NonNullable<
+      AddressVerificationFlow_verifiedAddressResult$data
+    >["verifyAddressOrError"]
```
damassi, line 154: I know this code was mostly all here before, but do all of these verbose types need to be repeated? Also, `suggestedAddresses` is different than `inputAddress` and `verificationStatus`, so casting in this way (seems) to be incorrect. 

As `verifiedAddressResult.verifyAddressOrError` is a union underneath, you can use the `Extract` typescript util helper to pull out the desired "slice" of the union:

```tsx
type SuccessType = Extract<SomeType, { __typename: 'SomeTypeSuccess' }>
type ErrorType = Extract<SomeType, { __typename: 'SomeTypeError' }>

const foo: SuccessType = ...
```

This might help here, but not too sure!

fladson: Thanks for the review.
Those are being handled at https://github.com/artsy/force/pull/12689/files, so I will wait until that's merged to do a rebase here.

fladson:
Okay, I got the modal to render when there's some error but the form is still not proceeding to the next step because we have this:
```
if (!props?.verifyAddress) {
          return <div>... loading ... </div>
        }
```

SCR-20230809-frn


@erikdstock I would love some suggestions on how to proceed from here.
Looks like the logic is dependant on data that is returned from the success cases, so I need some more help changing it to send the form input when the user still chooses to send unverified data.

erikdstock:
I'll take over the rebase to bring this up to date and ready to merge - thanks @fladson!

erikdstock:
I've resolved the merge conflicts here and completed the PR in #12723; going to close this one but keep the branch for review & comparison.

PR #12714, opened 2023-08-09
refactor(filters): avoids props drilling alerts/pills

dzucconi:
So the main goal of this PR is to remove the `FilterPillsSection` prop from the `ArtworkFilters` component.

The updated filter layout calls for separating out the filters which are displayed and the "Create Alert" button. This makes sense logically as well: why should these things be conflated in code. Additionally it's unnecessary to thread this component and state through all of these layers.

-----

I'm, generally, a bit confused about this piece of UI. Some of the state is threaded through as props, others pieces use state that's provided by a context provider. I wind up using this context provider to provide the entity down to the "Create Alert" button. I'm not really sure why the context isn't the complete source of truth. It's pretty odd to, in all of these different call sites, to have to pull things out of both the alerts and filters context, then apply a bunch of functions to them to massage them into the things you're looking for. Seems to me like the alerts context could just talk to filter's context and then simply provide all of this through a single hook.


src/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -37,32 +34,23 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
...
   }
 
-  const defaultPills: FilterPill[] = [
-    {
-      isDefault: true,
-      value: artist.internalID,
-      displayValue: artist.name ?? "",
-      field: "artistIDs",
-    },
-  ]
```
dzucconi, line 64: We're removing this notion of a 'default pill' from the filter section; though it remains to be displayed in the alert dialog.

The biggest puzzle to me is why we were manually providing this as a prop, when the dialog just pulls it out of the filter state. 🤷

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
```diff
@@ -170,8 +170,10 @@ const SavedSearchAlertEditForm: React.FC<SavedSearchAlertEditFormProps> = ({
 
               <Box>
                 <Text variant="xs">Filters</Text>
+
                 <Spacer y={2} />
-                <Flex flexWrap="wrap" mx={-0.5}>
```
dzucconi, line 174: The `ActiveFilterPills` component suffered from having external margins, then all the call sites had to negate them using negative margin.

Just use `gap`.

src/Components/ArtworkFilter/ArtworkFilterSavedSearchAlertContextProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+} from "Components/SavedSearchAlert/SavedSearchAlertContext"
+import { getSearchCriteriaFromFilters } from "Components/SavedSearchAlert/Utils/savedSearchCriteria"
+import { FC } from "react"
+
+interface ArtworkFilterSavedSearchAlertContextProviderProps
+  extends Omit<SavedSearchAlertContextProviderProps, "criteria"> {}
+
+/**
+ * A simple component that just wraps the `SavedSearchAlertContextProvider` in order
+ *  to omit the `criteria` prop, which is inferred from the filter context
```
dzucconi, line 14: The criteria this provides isn't used. The provider pushes a copy of it into it's own state, but keeps the name the same. I believe it's so that your alert criteria and search criteria can diverge. Which is fine. But it is confusing.

src/Components/SavedSearchAlert/Components/ActiveFilterPills.tsx
```diff
@@ -1,30 +1,19 @@
...
 
   return (
-    <Flex
-      flexWrap="wrap"
-      mx={-PILL_HORIZONTAL_MARGIN_SIZE}
-      mb={4}
-      data-testid="artworkGridFilterPills"
-    >
+    <Flex flexWrap="wrap" data-testid="artworkGridFilterPills" gap={1}>
       <SavedSearchAlertPills items={pills} onDeletePress={removePill} />
```
dzucconi, line 17: This is the only usage of this component currently; probably worth inlining but I'm trying to keep the touch here minimal.

dblandin (approved):
Looks like a solid refactor to me 👍🏼 

Happy to help QA via a review app or on staging once this merges.

src/Components/ArtworkFilter/ArtworkFilterCreateAlert.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <ProgressiveOnboardingAlertCreate>
+          {({ onSkip: createSkip }) => (
+            <ProgressiveOnboardingAlertReady>
+              {({ onSkip: readySkip }) => (
+                <Button
+                  variant="secondaryBlack"
+                  size="small"
+                  Icon={BellStrokeIcon}
+                  onClick={() => {
+                    createSkip()
```
araujobarret, line 56: nit: maybe `onCreateSkip` `onReadySkip`

nickskalkin (approved):

PR #12719, opened 2023-08-10
add tests for tracking on address verification

erikdstock:

This PR solves [EMI-1299] by adding tests for address verification - builds off the work of #12636

### DIFF
--- a/src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
+++ b/src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
@@ -16,6 +16,10 @@ const mockInputAddress = {
   postalCode: "10013",
   country: "US",
 }
+const componentProps = {
+  onClose: mockOnClose,
+  onChosenAddress: mockOnChosenAddress,
+}
 
 jest.unmock("react-relay")
 
@@ -82,9 +86,13 @@ describe("AddressVerificationFlow", () => {
     }
 
     it("displays the correct content", async () => {
-      renderWithRelay({
-        VerifyAddressType: () => mockResult,
-      })
+      renderWithRelay(
+        {
+          VerifyAddressType: () => mockResult,
+        },
+        undefined,
+        componentProps
+      )
 
       await screen.findByText("Check your delivery address")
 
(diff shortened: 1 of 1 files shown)


damassi (approved):

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -96,16 +104,36 @@ describe("AddressVerificationFlow", () => {
...
         undefined,
+        componentProps
+      )
+      await screen.findByText("Check your delivery address")
+
+      expect(trackEvent).toHaveBeenCalledTimes(1)
+      expect(trackEvent).toHaveBeenCalledWith({
+        action_type: "validationAddressViewed",
+        context_module: "ordersShipping",
+        context_page_owner_id: undefined,
```
iskounen, line 121: When we track checkout dead-ends, we set this field to the order ID. Should that also be the case here? (cc: @TMcMeans in case you've discussed this with @daytavares)

iskounen, line 121: This field is set the order ID for other events that we track in the checkout flow. Why is it `undefined` for this event?

iskounen (commented):
Event tracking out-of-scope for this PR, but these tests highlight that we aren't setting the `context_page_owner_id` and `user_id` fields of these events, so I'm curious why that is. Both seem to be helpful and available fields, so any context would be appreciated.

(cc: @TMcMeans in case you and @daytavares have discussed omitting these fields)

erikdstock:
> Event tracking out-of-scope for this PR, but these tests highlight that we aren't setting the `context_page_owner_id` and `user_id` fields of these events, so I'm curious why that is.

Yes, one reason i wanted to get these in was to specify the behavior explicitly (and confirm it in this review). Thanks for highlighting :)

erikdstock:
I'm going to merge this as-is and create a followup ticket to clarify tracking questions raised in [Adam's comment.](https://github.com/artsy/force/pull/12719#pullrequestreview-1572175081)

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
TMcMeans: hey, adding in a comment after merge to say that there are ways to mock out useTracking and force the data you want for testing. using a` trackingSpy` or something else like that in Force.

TMcMeans:
@erikdstock @iskounen mentioning that I'm going to take over this ticket you wrote after our chat earlier to do an audit of the tracking calls being mocked in the test: https://artsyproduct.atlassian.net/browse/EMI-1392

PR #12720, opened 2023-08-10
fix(EMI-1392): address verification flow tracking spec audit

TMcMeans:

This PR solves [EMI-1392]

### Description

The goal of this PR is to do an audit of the mocked tracking calls in the tests and fix any errors in expected values, mocks or empty fields in the `AddressVerificationFlow` spec. 

@artsy/emerald-devs 

[EMI-1392]: https://artsyproduct.atlassian.net/browse/EMI-1392?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -160,7 +160,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
       context_module: ContextModule.ordersShipping,
       context_page_owner_type: OwnerType.ordersShipping,
       context_page_owner_id: contextPageOwnerSlug,
-      user_id: user?.id,
+      user_id: userId,
       subject: "Check your delivery address",
       option,
       label: "Back to Edit",
@@ -174,7 +174,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
       context_module: ContextModule.ordersShipping,
       context_page_owner_type: OwnerType.ordersShipping,
       context_page_owner_id: contextPageOwnerSlug,
-      user_id: user?.id,
+      user_id: userId,
       subject: "Check your delivery address",
       label: "Edit Address",
     })
@@ -334,7 +334,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
                     context_module: ContextModule.ordersShipping,
                     context_page_owner_type: OwnerType.ordersShipping,
                     context_page_owner_id: contextPageOwnerSlug,
-                    user_id: user?.id,
+                    user_id: userId,
                     subject: "Confirm your delivery address",
                     option: selectedAddressKey.includes("suggestedAddress")
                       ? "Recommended"
@@ -373,7 +373,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
                   context_module: ContextModule.ordersShipping,
                   context_page_owner_type: OwnerType.ordersShipping,
                   context_page_owner_id: contextPageOwnerSlug,
-                  user_id: user?.id,
+                  user_id: userId,
                   subject: "Check your delivery address",
                   label: "Use This Address",
                 })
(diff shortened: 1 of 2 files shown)


TMcMeans:
During an audit of the tracking mocks in the spec, I encountered an infinite loop warning. This was specifically tied to a `useEffect()` that utilized the `user` object from SystemContext, in its dependency array. The user object's reference was updating with every render of the component. This triggered the useEffect() on each render because it detected a change in its dependency, even though the actual data within the user object remained consistent. 

To fix this, I removed the direct reference to the user obj from the useEffect() and assigned it to a variable. Added this newly created variable to the useEffect() dependency array under the assumption is that by doing this useEffect() will only be triggered by real changes to the user instead of superficial references updates.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -21,14 +23,38 @@ const componentProps = {
...
-  Component: AddressVerificationFlowFragmentContainer,
+  Component: props => {
+    return (
+      <AnalyticsContext.Provider
+        value={{
+          contextPageOwnerSlug: "example-order-id",
+        }}
+      >
+        <AddressVerificationFlowFragmentContainer {...(props as any)} />
+      </AnalyticsContext.Provider>
```
erikdstock, line 55: Can we use something like [this](https://github.com/artsy/force/blob/e1c9df1422973db7d5e913c10c58080b2efa18fa/src/Apps/Consign/Routes/MarketingLanding/Components/__tests__/HowItWorksSteps.jest.tsx#L8-L12)?

erikdstock (commented):
Overall looks good - I searched `*.jest.tsx` files for other uses of the analytics context and found an example that mocks it out similar to how we do with `useTracking`, so that might be useful here, but non-blocking!

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -159,6 +159,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
     >["suggestedAddresses"])
   const inputAddress = verifyAddress.inputAddress as AddressVerificationFlow_verifyAddress$data["inputAddress"]
   const verificationStatus = verifyAddress.verificationStatus as AddressVerificationFlow_verifyAddress$data["verificationStatus"]
+  const userId = user?.id
```
dzucconi, line 162: This would still change on every render. I'm surprised that this doesn't manifest the bug in the spec but this wouldn't fix it if it's an actual bug.

dzucconi: If you wanted to do this you would use a ref. But that's probably going to still be a bug.

Better would be just to remove these effects.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
dzucconi: Yep

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: My theory here was that for some reason the `user` object is changing identity and because of how the dependency array evaluates objects the effect is being triggered again. I've only seen this in tests, so could it be part of our test setup?

dzucconi: Yeah, it's definitely part of the test setup. It's unclear to me whether or not this is something that's possible to drive to in the browser — though that's sort of why I think you should just get rid of the effects: easier to reason about.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
TMcMeans: I like this way to do it, is it just preference or a newer way to mock/stub out analytics context?

src/Apps/Order/Components/AddressVerificationFlow.tsx
TMcMeans: I agree w/ you @dzucconi but think that removing the useEffects() should go in a follow up PR.

PR #12723, opened 2023-08-11
feat: handle gravity errors in metaphysics verify address query resolver

erikdstock:
Must be merged with https://github.com/artsy/metaphysics/pull/5179
Completes [EMI-1286]

This is a rebase of #12703. I put it in a separate PR for comparison in case i missed anything in the rebase to avoid force pushing over that work. The functionality is to handle an error from our address verification service and fall back to our confirmation modal without suggestions. Must be tested against https://github.com/artsy/metaphysics/pull/5179 locally and merged together, as it includes a major schema change used only by this component.

<details><summary>Summary of changes from #12703 </summary>

- rebased the @fladson's branch off of `main`, doing my best to fix merge conflicts
- added tests for the new behavior
- removed a `...loading...` message used in development
- added functionality to fall back to an address based on the user's input in case a gravity error meant that the formatted `inputAddress` lines were not available.

</details> 

Questions:
- Do we want to track viewing this kind of modal differently? Currently it is tracked as any other verification without suggested addresses.
- Tracking code is becoming duplicative. We could move this out into a hook that runs only when the modal renders and eliminate the large `useEffect()` hook which adds a lot of complexity - but that could be a followup.
- The Query Renderer passes an `error` prop that we aren't handling which is different from our schema error. I want to understand how that works so that if it ever pops up we don't block address selection completely.

[EMI-1286]: https://artsyproduct.atlassian.net/browse/EMI-1286?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -218,7 +218,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
         context_page_owner_id: contextPageOwnerSlug,
         user_id: userId,
         flow: "user adding shipping address",
-        subject: "Check your delivery address",
+        subject: "Check your delivery address (error)",
         option: "review and confirm",
       })
     } else {
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -17,6 +17,7 @@ import { useTracking } from "react-tracking"
 import { useSystemContext } from "System/SystemContext"
 import { ContextModule, OwnerType } from "@artsy/cohesion"
 import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
+import { Address } from "Components/AddressForm"
 
 type VerifyAddressSuccessType = Extract<
   AddressVerificationFlow_verifyAddress$data["verifyAddressOrError"],
@@ -73,6 +74,18 @@ interface AddressOption {
 enum AddressSuggestionRadioButton {
   recommended = "Recommended",
   user_address = "What you entered",
+}
+
+const formValues = (address: Address) => {
+  return {
+    address: {
+      addressLine1: address.addressLine1,
+      addressLine2: address.addressLine2,
+      country: address.country,
+      city: address.city,
+      postalCode: address.postalCode,
+    },
+  }
 }
 
 const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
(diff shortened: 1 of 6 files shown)


src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -152,60 +187,81 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+      trackEvent({
+        action_type: "validationAddressViewed",
+        context_module: ContextModule.ordersShipping,
+        context_page_owner_type: OwnerType.ordersShipping,
+        context_page_owner_id: contextPageOwnerSlug,
+        user_id: userId,
+        flow: "user adding shipping address",
+        subject: "Check your delivery address",
+        option: "review and confirm",
+      })
```
iskounen, line 223: I think this is a great way of handling backend errors for this flow.

One suggestion I would make is to differentiate the tracking event from the regular `REVIEW_AND_CONFIRM` instance so that we can tell how many are caused by poor user input v.s. backend errors.

iskounen (dismissed):

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Agreed!

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -392,28 +462,23 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
   return (
     <SystemQueryRenderer<AddressVerificationFlowQuery>
       variables={{ address }}
-      render={({ props, error }) => {
-        if (error) {
-          // TODO: we need better error handling
-          console.error(error)
-          return null
-        }
-
```
erikdstock, line 401: I'm not sure about merging this - I haven't seen an error manifest here in testing (which confused me) but if it did happen the entire address submission would be blocked. Could we return a fake unavailable value as a fallback in this case?

iskounen: We could mock `props` to look like a `REVIEW_AND_CONFIRM` situation...

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -66,7 +78,30 @@ enum AddressSuggestionRadioButton {
...
 }
 
+const fallbackFromFormValues = (address: AddressValues): AddressOption => {
+  return {
+    key: USER_INPUT,
+    recommended: false,
+    lines: [
+      address.addressLine1,
+      address.addressLine2 || null,
+      `${address.city}, ${address.region} ${address.postalCode}`,
```
erikdstock, line 88: This could be more clever (array with filter/join plus the comma thing) but given it's an error case I'm inclined to leave it for now.

src/Apps/Order/Components/AddressVerificationFlow.tsx
iskounen: I take this back. I think that we should act like nothing happened if a backend error is raised.

When this error occurs, we haven't been able to make any determination about the quality of a user's input, so showing the `REVIEW_AND_CONFIRM` dialog is equivalent to us telling a buyer that we don't recognize a legitimate address.

In that scenario I think it's better to abort the address verification flow and allow the user to take their chances with the address they've entered.

src/Apps/Order/Components/AddressVerificationFlow.tsx
iskounen: Clever code is usually more expensive to read and change, so I'm pretty happy with this :)

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Yes, I'm going to ticket this thread for followup since I don't even know how to reproduce it in staging for testing.

PR #12733, opened 2023-08-12
fix: preserve suggested address after editing

starsirius:

This PR solves [EMI-1388](https://artsyproduct.atlassian.net/browse/EMI-1388)

### Description

Editing address after accepting suggested address resets form with original user inputs. This was due to the `initialState` in `useState` is only used in the initial render and ignored in re-render. This PR also adds a test to cover this scenario. More details in thread.

<details><summary>Before</summary>
<p>
persist-address-old

</p>
</details> 

<details><summary>After</summary>
<p>

persist-address-new

</p>
</details> 

[EMI-1388]: https://artsyproduct.atlassian.net/browse/EMI-1388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/AddressForm.tsx
+++ b/src/Components/AddressForm.tsx
@@ -2,6 +2,7 @@ import { Column, GridColumns, Text, Spacer, Input } from "@artsy/palette"
 import { CountrySelect } from "Components/CountrySelect"
 import * as React from "react"
 import { CreateTokenCardData } from "@stripe/stripe-js"
+import { isEqual } from "lodash"
 
 export interface Address {
   name: string
@@ -69,7 +70,15 @@ export const AddressForm: React.FC<AddressFormProps> = ({
   touched,
   tabIndex,
 }) => {
-  const [address, setAddress] = React.useState({ ...emptyAddress, ...value })
+  const addressFromProp = { ...emptyAddress, ...value }
+  const [address, setAddress] = React.useState(addressFromProp)
+  const [prevValue, setPrevValue] = React.useState(value)
+
+  if (!isEqual(value, prevValue)) {
+    setPrevValue(value)
+    setAddress(addressFromProp)
+  }
+
   const [key, setKey] = React.useState<keyof Address>()
 
   const changeEventHandler = (key: keyof Address) => (
--- a/src/Components/__tests__/AddressForm.jest.tsx
+++ b/src/Components/__tests__/AddressForm.jest.tsx
@@ -1,0 +1,42 @@
+import { screen, render, fireEvent } from "@testing-library/react"
+import { AddressForm } from "Components/AddressForm"
+import { FC, useState } from "react"
+
+describe("AddressForm", () => {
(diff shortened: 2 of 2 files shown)


src/Components/__tests__/AddressForm.jest.tsx
```diff
@@ -0,0 +1,42 @@
...
+      "Street address"
+    ) as HTMLInputElement
+    expect(line1.value).toBe("Before update")
+
+    fireEvent.click(screen.getByText("Update"))
+    expect(line1.value).toBe("After update")
+
+    const city = screen.getByPlaceholderText("City")
+    fireEvent.change(city, { target: { value: "New York" } })
+    expect(line1.value).toBe("After update")
```
starsirius, line 40: This test [reproduces the issue](https://app.circleci.com/pipelines/github/artsy/force/51173/workflows/c20d9ce9-3454-4a80-8007-6bf27c743441/jobs/383361).

Screen Shot 2023-08-11 at 10 05 49 PM

src/Components/AddressForm.tsx
```diff
@@ -69,7 +70,15 @@ export const AddressForm: React.FC<AddressFormProps> = ({
   touched,
   tabIndex,
 }) => {
-  const [address, setAddress] = React.useState({ ...emptyAddress, ...value })
```
starsirius, line 72: The address form in shipping uses the [`value`](https://github.com/artsy/force/blob/c3b5a8df47e2dac2ece9590c2d26af5b9d669125/src/Apps/Order/Routes/Shipping/index.tsx#L848) prop to render input values. The address verification flow [updates the address](https://github.com/artsy/force/blob/c3b5a8df47e2dac2ece9590c2d26af5b9d669125/src/Apps/Order/Routes/Shipping/index.tsx#L841) after accepting a suggested address, and it triggers re-rendering of the address form with the updated `value` prop. However, the initial state in `useState` here only gets used in the initial render and not in re-render. `address` becomes out-of-sync with `value` and remains the original address, although accepting the suggested address passes an updated `value` prop.

This follows the doc to [update states when props change](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).

starsirius: I wonder if there is a better way to not control form data from both internal state and props passed in. 🤔

erikdstock: Looks great! I wonder if there is an edge case here where an updated address does not pass our form's validations. It would be good to have an idea of what cases this is possible in. The main one I am aware of is places that don't use postal codes (I think the UAE is one). This probably wouldn't affect us for US release but something I'd like to understand better.

starsirius: Good callout! We should make sure the basic form validation does not conflict with Smarty address verification.

PR #12736, opened 2023-08-14
chore: Tracking updates, refactor address verification flow to avoid duplicative tracking

erikdstock:

This PR updates our tracking code to match the schema specified in [EMI-1402](https://artsyproduct.atlassian.net/browse/EMI-1402). In the process I noticed that our Order app does not have an analytics context provider even though it is specified in our analytics schema, so I added one here for the **shipping step only**. If this was the right thing to do we should consider following up with others.

In addition, over the weekend I had the nagging sense that we could simplify some of our address verification flow code. Both @TMcMeans and myself had issues with tests getting stuck in infinite loops or worker processes failing to exit, so hopefully this would reduce those issues.

[EMI-1402]: https://artsyproduct.atlassian.net/browse/EMI-1402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -100,22 +104,151 @@ const fallbackFromFormValues = (address: AddressValues): AddressOption => {
   }
 }
 
+const useAddressVerificationTracking = () => {
```
erikdstock, line 107: created this to DRY up some repeated verification calls and remove hooks which were only needed for tracking.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -100,22 +104,151 @@ const fallbackFromFormValues = (address: AddressValues): AddressOption => {
...
+  const suggestedAddresses = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.suggestedAddresses
+  const inputAddress = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.inputAddress
+  const verificationStatus = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.verificationStatus
+
+  const hasError = Boolean(error)
+
+  if (hasError) {
```
erikdstock, line 183: Previously this whole section ran in a hook, but it depends on props which don't currently change after load. If we changed the behavior to run multiple verifications on a flow then our use of the `didLoad` ref would have to change to allow updates, but this code should now be stable.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -100,22 +104,151 @@ const fallbackFromFormValues = (address: AddressValues): AddressOption => {
...
 const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
   verificationInput,
   verifyAddress,
   onChosenAddress,
   onClose,
 }) => {
-  const [modalType, setModalType] = useState<ModalType | null>(null)
-  const [addressOptions, setAddressOptions] = useState<AddressOption[]>([])
+  const [showModal, setShowModal] = useState(false)
+  const didLoad = useRef(false)
```
erikdstock, line 162: This is used to make sure the initial modal loading and tracking happens only once.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -169,117 +293,27 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
   }
 
   const handleEditAddress = () => {
-    trackEvent({
-      action_type: "clickedValidationAddress",
-      context_module: ContextModule.ordersShipping,
-      context_page_owner_type: OwnerType.ordersShipping,
-      context_page_owner_id: contextPageOwnerSlug,
-      user_id: userId,
+    trackClickedModal({
```
erikdstock, line 295: (May need to expand diff) - worth noting here that `handleBackToEdit` and `handleEditAddress` are the same button for the two different modals and have nearly identical behavior.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
```diff
@@ -164,7 +186,7 @@ describe("AddressVerificationFlow", () => {
         context_page_owner_id: "example-order-id",
         context_page_owner_type: "orders-shipping",
         flow: "user adding shipping address",
-        option: "review and confirm",
+        option: "review_and_confirm",
```
erikdstock, line 189: I made this change as a convenience to pass the value directly. We could keep it the other way if we prefer.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -100,22 +104,150 @@ const fallbackFromFormValues = (address: AddressValues): AddressOption => {
...
+        )
+        didLoad.current = true
+      } else {
+        didLoad.current = true
+
+        trackViewedModal({
+          option: verificationPath,
+          subject: modalTitles[verificationPath]!,
+        })
+      }
```
erikdstock, line 240: I noticed that our existing behavior (migrated here) was to track `option` as the path/type of modal (`suggestions | review and confirm`), but when closing the modal we track the [`option` as which address the user had selected at that moment](https://github.com/artsy/force/pull/12736/files#diff-a65d5b9bf3f3ba0af862f0c53c8ea2e0be243adf95b86daacdb768a05e3f538aR278-R292). Note that `subject` also indicates the kind of modal here via its title. I'm not sure if this inconsistency matters.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -100,22 +104,150 @@ const fallbackFromFormValues = (address: AddressValues): AddressOption => {
...
+      setShowModal(true)
```
erikdstock, line 241: I `setShowModal(true)` even if no modal will be displayed because our tests look for an empty div in those cases :shrug:

erikdstock (commented):
tour of changes:

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Resolved in 23c495d48cb3eb67d2971453372fb84a5742bdae

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -130,19 +185,15 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
         : AddressVerifiedBy.ARTSY
 
     if (selectedAddress) {
-      setModalType(null)
+      setShowModal(false)
       onChosenAddress(verifiedBy, selectedAddress.address, false)
     }
   }, [addressOptions, onChosenAddress, selectedAddressKey])
```
erikdstock, line 191: Not sure if this is meaningful, but since they are different handlers I thought i would specify the case where the user dismisses the modal vs clicking a button to edit their address.

erikdstock, line 191: Not sure if this is meaningful, but since they are different handlers I thought i would specify the case where the user dismisses the modal vs clicking a button to edit their address.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -130,19 +185,15 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
-    trackEvent({
-      action_type: "clickedCloseValidationAddress",
-      context_module: ContextModule.ordersShipping,
-      context_page_owner_type: OwnerType.ordersShipping,
-      context_page_owner_id: contextPageOwnerSlug,
-      option: null,
-      label: null,
+    trackClickedModal({
+      label: "close modal",
+      subject: modalTitles[verificationPath],
```
erikdstock, line 196: Not sure if this is meaningful, but since they are different handlers I thought i would specify the case where the user dismisses the modal vs clicking a button to edit their address.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -481,3 +431,80 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
...
+      addressLine2: address.addressLine2,
+      region: address.region,
+      country: address.country,
+      city: address.city,
+      postalCode: address.postalCode,
+    },
+  }
+}
+
+const useAddressVerificationTracking = () => {
```
TMcMeans, line 464: non-blocking, but just wanted to point out that moving the tracking to a different place may cause dupes. As long as tracking only runs once within a useEffect() there shouldn't be any issues.

erikdstock: I think we should be ok here - because tracking calls have been removed from the previous version (`useEffect` with a big chain of if/elses that calls side effects) the code is now more explicitly structured to run the tracking calls only on initial load and when a user selects an address. So my hope is that this will actually be safer.

src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx
erikdstock: reverted

src/DevTools/setupTestWrapper.tsx
```diff
@@ -86,12 +86,12 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
     manualEnvControl?: boolean,
-    componentProps?: {}
+    componentProps?: {},
+    mockedEnv: MockEnvironment = createMockEnvironment()
   ): RenderWithRelay => {
-    const env = createMockEnvironment()
     const TestRenderer = () => (
       <QueryRenderer<T>
-        environment={env}
+        environment={mockedEnv}
```
erikdstock, line 94: After some wrestling I have got the whole form working in tests - it required a small adjustment to our rendering harness. Interested in feedback on this approach, as after implementing it I realized it looks similar to how our enzyme-based `renderWithRelay` did it.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -199,21 +205,135 @@ const testMe: ShippingTestQuery$rawResponse["me"] = {
...
+            // return MockPayloadGenerator.generate(operation, {
+            //   VerifyAddressMutationType: () => addressVerificationResult,
+            // })
+            return {
+              data: {
+                verifyAddress: {
+                  verifyAddressOrError: addressVerificationResult,
+                },
+              },
+            }
```
erikdstock, line 320: After finally getting the query renderer wrapping the Address Verification Flow to use my mocked env I still couldn't get the payload generator to respect my mock resolver, so i just passed data explicitly :/

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -199,21 +205,135 @@ const testMe: ShippingTestQuery$rawResponse["me"] = {
   },
 }
 
+const pushMock = jest.fn()
+let isCommittingMutation
+const relayEnv = createMockEnvironment()
+
+const testWrapperConfig = {
+  Component: (props: any) => (
+    <MockBoot relayEnvironment={relayEnv}>
```
erikdstock, line 214: This makes sure our system provider gets the mocked relay env and doesn't create its own. Really hard to debug.

erikdstock (commented):
After some struggling this afternoon and a break I finally got tests working with `@testing-library/react`. Interested in feedback on that especially.

src/Components/__tests__/Utils/addressFormTLHelpers.ts
```diff
@@ -0,0 +1,85 @@
...
+    address.region || ""
+  )
+
+  fireChangeEvent(
+    screen.getAllByPlaceholderText(
+      "Add phone number including country code"
+    )[0],
+    address.phoneNumber || ""
+  )
+}
```
damassi, line 49: We should co-locate this file with the address form file/test. Here it is buried and prematurely abstracted, when it only has to do with one thing (the address form)

src/DevTools/setupTestWrapper.tsx
damassi: we should keep things consistent with how we do it below, or update the below so that it is consistent with whats here

src/Components/__tests__/Utils/addressFormTLHelpers.ts
erikdstock: I was trying to mirror the structure of the current tests and helpers, but maybe a bigger break or new file is a good idea.

damassi: If you stick with colocation in the place that the helper is being called from (until said helper is used > 3 times), and avoid abstraction as much as humanly possible, you'll always win 😄

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -199,21 +204,295 @@ const testMe: ShippingTestQuery$rawResponse["me"] = {
...
 
+const pushMock = jest.fn()
+let isCommittingMutation
+const relayEnv = createMockEnvironment()
+
+// This config extracted so it can be used with both enzyme (setupTestWrapper)
+// and react-testing-library (setupTestWrapperTL) tests.
+// We need to pass in our own relayEnv to both <MockBoot /> and
+// renderWithRelay so that downstream QueryRenderers can use the test
+// environment. Otherwise SystemContext will create a new [real] environment.
```
damassi, line 215: #superminor - You can write multiline comments in VSCode by typing 
```
/** <hit enter>
```

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -199,21 +204,295 @@ const testMe: ShippingTestQuery$rawResponse["me"] = {
...
+        ...Shipping_order
+      }
+      me {
+        ...Shipping_me
+      }
+    }
+  `,
+}
+
+describe("Shipping [@testing-library/react]", () => {
```
damassi, line 248: Is the `[@testing-library/react]` string necessary?

If we're going to migrate this stuff, lets just migrate everything all at once and not mix testing libraries inside the same file. Mixing like this makes the the chaotic order app situation worse for little gain (since there's no real need to refactor tests or change approaches). 

We should either refactor it all, or stick with the existing model and then refactor everything later all at once.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -199,21 +204,295 @@ const testMe: ShippingTestQuery$rawResponse["me"] = {
...
+  const trackEvent = jest.fn()
+  beforeAll(() => {
+    ;(useTracking as jest.Mock).mockImplementation(() => ({
```
damassi, line 251:
```ts
const mockUseTracking = useTracking as jest.Mock
....
mockUseTracking.mockImplementation(...)
```

src/DevTools/setupTestWrapper.tsx
erikdstock: Do you mean match [this format](https://github.com/artsy/force/blob/f043d0db90852bca50854df7fabf98b871ee77f4/src/DevTools/setupTestWrapper.tsx#L129-L139) for falling back to creating the env, or matching all the positional arguments in both? Because they are positional i think this could be a huge diff on an already big PR.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
erikdstock: yeah, sometimes you start out thinking you are writing a single line comment and things just get out of hand.

erikdstock (commented):
On second thought, having checked my new e-e tests against the old code - these aren't catching any errors yet that weren't caught in the local `AddressVerificationFlow` tests.
The new tests do illustrate in a relatively contained way how to complete the whole end-end flow in jest, but I am going to pull them out into a POC branch that can [inform EMI-1413.](https://artsyproduct.atlassian.net/browse/EMI-1413) This diff is already big enough to deserve manual testing, and I don't want to set us down a messy path prematurely. Thanks for your feedback @damassi!

erikdstock:
Closing in favor of #12778. The spike on tests using testing-library has been moved to [this branch](https://github.com/artsy/force/compare/erik.SPIKE-shipping-react-testing-library-spike?expand=1) and I've left a comment on [EMI-1413](https://artsyproduct.atlassian.net/browse/EMI-1413) where it belongs.

PR #12745, opened 2023-08-16
fix: Extra save & continue click investigation fix

erikdstock:

Resolves [EMI-1410] and _I think _ [EMI-1397] as well.
This PR solves a bug we currently hack around, creating a negative user experience. Paired with @damassi and @iskounen on it.
cc @artsy/emerald-devs 

### Description

In https://github.com/artsy/force/pull/12684 and #12673 we changed the behavior of our address verification and submission flow to require an additional button click. Initially the thought was that we had a race condition with numerous `useState` variable settings, causing an old address to be sent immediately after confirmation.

While that fix solved the problem, we think the problem was actually that the `onChosenAddress` prop, an anonymous arrow function, was closuring the `finalizeFulfillment()` function, which itself was closuring the `setShipping()` function, which at the time the component was mounted contained an old address.

Both fixes accomplish the same thing - not calling the stale `finalizeFulfillment` closured into the onChosenAddress prop. We initially fixed this by checking for presence of `addressVerifiedBy` (it is null until set by the flow), and sometimes this worked, but other times it triggered an error from simultaneous mutations. Thus to illustrate the problem and potential fix we created a new state variable.

Lacking tests at the shipping route level i think this reserves at least one more round of manual testing for variations on:
- ship vs arta ship
- perfect address vs suggestions vs no suggestions

[EMI-1410]: https://artsyproduct.atlassian.net/browse/EMI-1410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1397]: https://artsyproduct.atlassian.net/browse/EMI-1397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -39,8 +39,7 @@ interface AddressVerificationFlowProps {
   verifyAddress: AddressVerificationFlow_verifyAddress$data
   onChosenAddress: (
     verifiedBy: AddressVerifiedBy,
-    address: AddressValues,
-    saveAndContinue: boolean
+    address: AddressValues
   ) => void
   onClose: () => void
   /* used only as a fallback if verification is unavailable */
@@ -131,7 +130,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
     if (selectedAddress) {
       setModalType(null)
-      onChosenAddress(verifiedBy, selectedAddress.address, true)
+      onChosenAddress(verifiedBy, selectedAddress.address)
     }
   }, [addressOptions, onChosenAddress, selectedAddressKey])
 
@@ -210,11 +209,11 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
     if (hasError) {
       const fallbackOption = fallbackFromFormValues(verificationInput)
       const verifiedBy = AddressVerifiedBy.USER
-      onChosenAddress(verifiedBy, fallbackOption.address, true)
+      onChosenAddress(verifiedBy, fallbackOption.address)
     } else {
       if (verificationStatus === "VERIFIED_NO_CHANGE") {
         const verifiedBy = AddressVerifiedBy.ARTSY
-        onChosenAddress(verifiedBy, inputOption.address, true)
+        onChosenAddress(verifiedBy, inputOption.address)
       } else {
         if (verificationStatus === "VERIFIED_WITH_CHANGES") {
           setModalType(ModalType.SUGGESTIONS)
@@ -449,8 +448,7 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
   address: AddressValues
   onChosenAddress: (
     verifiedBy: AddressVerifiedBy,
-    address: AddressValues,
-    saveAndContinue: boolean
+    address: AddressValues
   ) => void
   onClose: () => void
 }> = ({ address, onChosenAddress, onClose }) => {
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -131,7 +131,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
     if (selectedAddress) {
       setModalType(null)
-      onChosenAddress(verifiedBy, selectedAddress.address, false)
+      onChosenAddress(verifiedBy, selectedAddress.address, true)
     }
   }, [addressOptions, onChosenAddress, selectedAddressKey])
 
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -850,7 +850,7 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                     setAddress(address => {
                       return { ...address, ...chosenAddress }
                     })
-                    // trigger finalizeAddres() via useEffect
+                    // trigger finalizeFulfillment() via useEffect
                     setReadyToSaveVerifiedAddress(true)
                   }}
                 />
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -131,6 +131,10 @@ export const ShippingRoute: FC<ShippingProps> = props => {
     setAddressVerifiedBy,
   ] = useState<AddressVerifiedBy | null>(null)
 
+  const [readyToSaveVerifiedAddress, setReadyToSaveVerifiedAddress] = useState(
+    false
+  )
+
   const [shippingOption, setShippingOption] = useState<
     CommerceOrderFulfillmentTypeEnum
   >(getShippingOption(props.order.requestedFulfillment?.__typename))
@@ -241,7 +245,9 @@ export const ShippingRoute: FC<ShippingProps> = props => {
   // Save shipping info on the order. If it's Artsy shipping and a quote hasn't
   // been selected, this renders the quotes for user to select and finalize
   // again.
+  console.log({ renderCycle: address })
   const finalizeFulfillment = async () => {
+    console.log({ finalizeFulfillment: address })
     if (checkIfArtsyShipping() && !!shippingQuoteId) {
       selectShippingQuote()
     } else {
@@ -686,6 +692,12 @@ export const ShippingRoute: FC<ShippingProps> = props => {
     return false
   }
 
+  useEffect(() => {
+    if (readyToSaveVerifiedAddress) {
+      finalizeFulfillment()
+    }
+  }, [readyToSaveVerifiedAddress])
+
   const renderArtaErrorMessage = () => {
     return (
       <Text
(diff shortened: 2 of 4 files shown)


src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -131,7 +131,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
 
     if (selectedAddress) {
       setModalType(null)
-      onChosenAddress(verifiedBy, selectedAddress.address, false)
+      onChosenAddress(verifiedBy, selectedAddress.address, true)
```
erikdstock, line 134: this change doesn't matter to the fix.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -686,6 +692,12 @@ export const ShippingRoute: FC<ShippingProps> = props => {
     return false
   }
 
+  useEffect(() => {
+    if (readyToSaveVerifiedAddress) {
+      finalizeFulfillment()
+    }
+  }, [readyToSaveVerifiedAddress])
+
```
erikdstock, line 700: We wouldn't want to merge this as-is because this function does other things and the dependency array is intentionally missing `finalizeFulfillment`. That function could be added but then several more would need to be wrapped in `useCallback()`

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -838,8 +850,11 @@ export const ShippingRoute: FC<ShippingProps> = props => {
                   ) => {
                     setAddressNeedsVerification(false)
                     setAddressVerifiedBy(verifiedBy)
-                    setAddress({ ...address, ...chosenAddress })
-                    saveAndContinue && finalizeFulfillment()
+                    setAddress(address => {
+                      return { ...address, ...chosenAddress }
+                    })
+                    setReadyToSaveVerifiedAddress(true)
+                    // saveAndContinue && finalizeFulfillment()
```
erikdstock, line 857: by the time this finalizeFulfillment has executed it has been redefined due to the changes above, but the stale function value persists in the closure of the function passed as the `onChosenAddress` prop.

mzikherman:
I haven't touched/look at this code at all - but from reading the PR description, and esp the last sentence:

> That said this are of the code has lots of competing updates and interactions, including other known bugs.

Do you think it might make sense to work on a re-build of the order flow/app? (it may or may not come with design updates). My thinking is if the code has become very complex and unwieldy, even though iterative bug-fixes are _usually_ better than a full rebuild and a rebuild can come with its _own_ bugs - maybe a rebuild by a now more experienced team can come up with an overall better code-base?

To be clear, I'm specifically referring to the Force order/checkout flow - _not_ necessarily anything related to Exchange or more backend/mutation updates.

src/Apps/Order/Routes/Shipping/index.tsx
damassi: I don't think its imperative that we pass in finalize fulfillment and then callbackize everything. We know what we're doing -- check to see if that boolean state has changed -- and so its fine to add a exaustive deps ignore (imo)

damassi: Assuming this change is verified by @starsirius (or someone else) we should really get it shipped asap since this UX issue certainly has the potential to kill a sale.

damassi:
It def needs to be rebuilt; we've been talking about this for a long time now and various folks has expressed intent / interest, but it never goes anywhere significant. Hopefully this changes with the new team setup. But that said this code is complex and delicate and we can't refactor it until someone spends some time putting together a detailed control-flow diagram that maps each transition. And from there, use a formal state machine.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -241,7 +245,9 @@ export const ShippingRoute: FC<ShippingProps> = props => {
   // Save shipping info on the order. If it's Artsy shipping and a quote hasn't
   // been selected, this renders the quotes for user to select and finalize
   // again.
+  console.log({ renderCycle: address })
```
damassi, line 248: @erikdstock - can we remove these logs?

src/Apps/Order/Routes/Shipping/index.tsx
starsirius: Nice! So we should probably not rely on any data that's not passed in from the function arguments. Guess using a function in the set function above is more reliable in the same way.

starsirius, line 857: Then I think we can remove the whole `saveAndContinue` argument with the proposed approach?

src/Apps/Order/Routes/Shipping/index.tsx
damassi: Agree that passing an arg could work, but threading state through is likely more complex vs setting (this) state to trigger an effect, since this is scoped to exactly one location setting the state, and nothing else needs to change.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -831,15 +844,13 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-                  ) => {
+                  onChosenAddress={(verifiedBy, chosenAddress) => {
                     setAddressNeedsVerification(false)
                     setAddressVerifiedBy(verifiedBy)
-                    setAddress({ ...address, ...chosenAddress })
-                    saveAndContinue && finalizeFulfillment()
+                    setAddress(address => {
+                      return { ...address, ...chosenAddress }
+                    })
+                    setReadyToSaveVerifiedAddress(true)
```
damassi, line 853: Can we add a comment above this line describing exactly what happens in terms of triggering the button in the effect? That way there's no ambiguity.

damassi (dismissed):
One request for a small comment above `setReadyToSaveVerifiedAddress`, otherwise LGTM once 🍏 👍

erikdstock:
Alright, this appears to be working - thanks for the reviews so far. It also appears to fix the bug in [EM-1397](https://artsyproduct.atlassian.net/browse/EMI-1397).

starsirius:
Still need to do more thorough QA, and here are my notes so far:

- Accepting a recommended address when an order already had a quote does not re-quote or persist the recommended address. This happens when going back to the shipping step second time in the checkout flow. Still need to verify if this is the same in production currently.

edit-address-not-persisted

- This fix doesn't fix [EM-1397](https://artsyproduct.atlassian.net/browse/EMI-1397) fully even though it sends the desired `verifiedBy` parameter. It's a separate issue from what this PR is solving, so I'll add a note too there. See the recording: it passes the correct `verifiedBy` param when finalizing fulfillment, but leaves the app in an incorrect state, which can lead to future issues.

verified_by


I think it still makes sense to merge this PR if no blocking issues, and in the same time, I'd like to discuss if we can have better test coverage to catch regressions. It's really hard to have to manually test all scenarios. 😓

erikdstock:
@starsirius thanks for testing! Agree we need tests for this flow at the route level- I wonder if this justifies its own ticket so we can prioritize it appropriately.

starsirius:
@erikdstock Yes. Tests can be its own task and not blocking for this PR. I'll try to verify the major flows are covered and merge!

starsirius:
OK I tested again the major flows and wrote up the test cases (expand below). It may be helpful as an outline for the future improved tests.

<details><summary>Arta shipping test cases</summary>
<p>

- Suggestion
  - Correct modal shows up
  - Selecting the suggested address populates the form (secondary CTA skipped)
  - Selecting the suggested address continues to quote selection
  - Selecting the suggested address calls the set shipping mutation with the corrected address
  - Clicking save and continue again proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address
- No suggestion
  - Correct modal shows up
  - Selecting edit address allows me to edit (secondary CTA skipped)
  - Clicking save and continue calls the set shipping mutation with the original address and populates the quotes
  - Clicking save and continue again proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address
- Perfect address
  - No modal shows up
  - Clicking save and continue calls the set shipping mutation with the original address and populates the quotes
  - Clicking save and continue again proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address

</p>
</details> 

<details><summary>Partner shipping test cases</summary>
<p>

- Suggestion
  - Correct modal shows up
  - Selecting the suggested address calls the set shipping mutation with the corrected address (secondary CTA skipped)
  - Selecting the suggested address proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address
- No suggestion
  - Correct modal shows up
  - Selecting edit address allows me to edit (secondary CTA skipped)
  - Clicking save and continue calls the set shipping mutation with the right address
  - Clicking save and continue proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address
- Perfect address
  - No modal shows up
  - Clicking save and continue calls the set shipping mutation with the original address
  - Clicking save and continue proceeds to the payment step
  - Navigating back to the shipping step shows the suggested address

</p>
</details> 

 I'm going to merge this and we'll follow up separately for

- The issues discovered in https://github.com/artsy/force/pull/12745#issuecomment-1685546255.
- Better test coverage

PR #12746, opened 2023-08-17
fix: add spacing for Market Stats for mobile view

egdbear:

### Description

I noticed Market Stats needs a bit spacing for mobile view

Screenshot 2023-08-17 at 12 39 31

Screenshot 2023-08-17 at 12 39 36

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -240,7 +240,7 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
 
       <Jump id="marketSignalsTop" />
 
-      <Spacer y={2} />
+      <Spacer y={[2, 0]} />
 
       <MarketStatsQueryRenderer
         id={artist.internalID}
--- a/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
@@ -240,6 +240,8 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
 
       <Jump id="marketSignalsTop" />
 
+      <Spacer y={2} />
+
       <MarketStatsQueryRenderer
         id={artist.internalID}
         onRendered={handleMarketStatsRendered}
--- a/src/Apps/Artist/Routes/Overview/ArtistOverviewRoute.tsx
+++ b/src/Apps/Artist/Routes/Overview/ArtistOverviewRoute.tsx
@@ -1,4 +1,4 @@
-import { Flex } from "@artsy/palette"
+import { Flex, Spacer } from "@artsy/palette"
 import * as React from "react"
 import { Title, Meta } from "react-head"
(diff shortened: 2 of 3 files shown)


dzucconi (changes requested):
Good catch but: two things:
* [The margin should be in the parent](https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-external-margins)
* We wouldn't use a non-token value for spacing: https://palette-storybook.artsy.net/?path=/story/theme--spacing

egdbear:
> Good catch but: two things:
> 
> * [The margin should be in the parent](https://github.com/artsy/force/blob/main/docs/best_practices.md#avoid-external-margins)
> * We wouldn't use a non-token value for spacing: https://palette-storybook.artsy.net/?path=/story/theme--spacing

makes sense!

src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResults.tsx
```diff
@@ -240,6 +240,8 @@ const AuctionResultsContainer: React.FC<AuctionResultsProps> = ({
 
       <Jump id="marketSignalsTop" />
 
+      <Spacer y={2} />
```
dzucconi, line 243: This winds up being a weird one. Actually the about tab has the same issue; the reason being is that the actual parent — the app with tabs: https://github.com/artsy/force/blob/4331d87e14354f15419e682ec609f9e897d65d8f/src/Apps/Artist/ArtistApp.tsx#L68 doesn't provide any margin on mobile but does on desktop. This seems to be because of the filter bar structure on mobile requiring 0 margin.

So to go forward with this:

I would change this to `<Spacer y={[2, 0]} />` — margins collapse so it's technically not a problem on desktop but just trying to be a bit more explicit about what we are targeting maybe (if, for instance this margin were to ever exceed `4`.) And then add this same thing to the about tab.

This winds up kind of breaking the 'no external margins' rule since this one dangles off the top here. I think it's OK in this instance because this is a route, not a component.

egdbear: Oh wow. Good spot for the about tab! I was testing it always against Artworks 😅

dzucconi (approved):

src/Apps/Artist/Routes/Overview/ArtistOverviewRoute.tsx
```diff
@@ -45,6 +45,8 @@ const ArtistOverviewRoute: React.FC<ArtistOverviewRouteProps> = ({
       <Meta name="title" content={title} />
       <Meta name="description" content={description} />
 
+      <Spacer y={[2, 0]} />
```
egdbear, line 48: This seems the parent component, right @dzucconi ? 😅

dzucconi: Sort of; for our purposes let's say yes — but really `ArtistApp` would be. It's just kind of simpler to target it here rather than swap out different margins per-route in the actual parent. This is still an external margin (there's no UI above this in this component) but it's the lesser evil.

egdbear: 🫠

PR #12748, opened 2023-08-18
feat(ONYX-130): add shelf for suggested artworks

tam-kis:
The type of this PR is: Feat

This PR solves [ONYX-130](https://artsyproduct.atlassian.net/browse/ONYX-130)

### Description

Adds Shelf for suggested artworks based on criteria.

Screenshot 2023-08-18 at 13 11 55

[ONYX-130]: https://artsyproduct.atlassian.net/browse/ONYX-130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
+++ b/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
@@ -31,7 +31,7 @@ export const SuggestedArtworksShelf: FC<SuggestedArtworksShelfProps> = ({
         <Flex
           flexDirection="row"
           flexWrap="wrap"
-          gap={1}
+          gap={2}
           justifyContent="center"
         >
           {artworks.map(artwork => (
--- a/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
+++ b/src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
@@ -6,7 +6,7 @@ import {
   SuggestedArtworksShelfQuery,
   SuggestedArtworksShelfQuery$data,
 } from "__generated__/SuggestedArtworksShelfQuery.graphql"
-import { Media } from "Utils/Responsive"
+
 import {
   ShelfArtworkFragmentContainer,
   ShelfArtworkPlaceholder,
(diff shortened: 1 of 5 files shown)


nickskalkin (dismissed):
Looks good! I would only make [this change](https://github.com/artsy/force/pull/12748/files#r1298631777)

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
```diff
@@ -0,0 +1,98 @@
...
+  artworksConnection: SuggestedArtworksShelfQuery$data["artworksConnection"]
+}
+
+export const SuggestedArtworksShelf: FC<SuggestedArtworksShelfProps> = ({
+  artworksConnection,
+}) => {
+  const artworks = extractNodes(artworksConnection)
+
+  return (
+    <Flex flexDirection="row" flexWrap="wrap" gap={2} justifyContent="center">
```
dzucconi, line 29: Why aren't we using an actual `Shelf` component? What this never overflows?

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <SuggestedArtworksShelf
+            artworksConnection={relayProps.artworksConnection}
+          />
+        )
+      }}
+    />
+  )
+}
+
+const ContentPlaceholder: FC = () => {
```
dzucconi, line 88: Would use a consistent naming prefix: `SuggestedArtworksShelfPlaceholder`

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
```diff
@@ -0,0 +1,98 @@
...
+  return (
+    <Skeleton>
+      <Flex flexDirection="row" flexWrap="wrap" gap={1} justifyContent="center">
```
dzucconi, line 91: This gap is different than the one above (again probably should just use `Shelf`)

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/SuggestedArtworksShelf.tsx
tam-kis: Yes, it doesn't overflow - it just displays the first 5 suggested artworks. There is a "See more" button underneath it that opens modal where we see all of the artworks.

dzucconi: What does it look like when it wraps? I think these components are variable width, sized based on area; so for instance with some landscapes in here?

tam-kis: The idea was to wrap them underneath (centered) but now that you mention it - it sounds better to have them scale instead of wrapping, I will suggest it!

dzucconi: That will probably wind up being quite difficult to achieve using the current components. It would probably make sense to remove them completely, as they overflow. Which would be easier to achieve.

tam-kis: Removing the wrap keeps the default behaviour where it scales on smaller screens, but I will discuss with design as well and update in the next PR.

PR #12750, opened 2023-08-18
chore(DIA-64): Update curated collections in nav

jo-rs:

This PR solves [DIA-64]

### Description
Updates in curated collections in nav
Screenshot 2023-08-18 at 15 22 17
Screenshot 2023-08-18 at 15 22 08

[DIA-64]: https://artsyproduct.atlassian.net/browse/DIA-64?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/NavBar/menuData.ts
+++ b/src/Components/NavBar/menuData.ts
@@ -37,12 +37,7 @@ export const ARTISTS_SUBMENU_DATA: MenuLinkData = {
       },
       {
         text: "Trending Now",
-        href: "/collection/trending-this-week",
-      },
-      {
-        text: "Artsy Vanguard Artists",
-        href: "/collection/the-artsy-vanguard",
-        dividerBelow: true,
+        href: "/collection/trending-this-week/new-this-week",
       },
       {
         text: "Curator's Picks",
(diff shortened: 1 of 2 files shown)


src/Components/NavBar/menuData.ts
```diff
@@ -137,6 +92,14 @@ export const ARTISTS_SUBMENU_DATA: MenuLinkData = {
...
         },
       },
+      // Empty column
+      {
+        text: "",
+        menu: {
+          title: "",
+          links: [],
+        },
+      },
```
jo-rs, line 102: Open to other suggestions here

jonallured: 🤷 seems legit

jonallured (approved):

src/Components/NavBar/menuData.ts
```diff
@@ -37,70 +37,25 @@ export const ARTISTS_SUBMENU_DATA: MenuLinkData = {
...
-            {
-              text: "Rising Artists of Asia & Its Diaspora",
-              href: "/collection/rising-artists-of-asia-and-its-diaspora",
+              text: "Curators’ Picks: Emerging",
+              href: "/collection/curators-picks-emerging",
             },
             {
-              text: "Women Artists Now",
-              href: "/collection/women-artists-to-watch",
+              text: "Curators’ Picks: Blue-Chip",
```
araujobarret, line 52: is it like this or `Curator's`?
Line 43 is in this form and here it's in a different form

jo-rs: Good question 🤔  
It's like that on the ticket, but maybe @jpoczik can confirm?
It's a bit odd that the main header would refer to an individual curator, but the sub links to the collective. But, I'm not a native speaker 🤷

jonallured: I believe it should match what you see when you go to the page:

https://www.artsy.net/collection/curators-picks-blue-chip?metric=in

Which looks like this:

```
Curators’ Picks: Blue-Chip
```

jpoczik: yes confirming what jon says that it's from multiple curators so **Curators’ Picks: Blue-Chip** is correct

PR #12751, opened 2023-08-18
feat(alerts): use generated displayName in create and edit forms

anandaroop:

This PR solves [ONYX-241]

Co-authored-by: @dblandin 

## Description

Wires up the new `previewSavedSearch.displayName` field from https://github.com/artsy/metaphysics/pull/5186 into the Alert creating and editing flows.

This feature is guarded by the [`onyx_force-fallback-to-generated-alert-names`](https://unleash.artsy.net/projects/default/features/onyx_force-fallback-to-generated-alert-names) feature flag, which is based on a user id strategy, so you might have to opt in to see these changes.

---

### When the feature flag is enabled

The MP-backed display name is shown (and updated realtime) as the placeholder in the alert form:

|Artist|Artwork|
|---|---|
|<video src="https://github.com/artsy/force/assets/140521/64c43a1c-c1f8-4cb2-abb7-cecd0a252602" />|<video src="https://github.com/artsy/force/assets/140521/0b4362b4-fb38-44af-b946-a8fb2a0bfe69" />|

---

### When the feature flag is disabled

Functions as before, i.e. the old-style placeholder is shown:

|Artist|Artwork|
|---|---|
|<video src="https://github.com/artsy/force/assets/140521/fc70c391-6259-44e2-be2f-84aa9673e215" />|<video src="https://github.com/artsy/force/assets/140521/62cc686c-8bab-44fe-acef-7bcf59492948" />|

[ONYX-241]: https://artsyproduct.atlassian.net/browse/ONYX-241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
+++ b/src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
@@ -18,22 +18,21 @@ export const SavedSearchAlertNameInputQueryRenderer: FC = () => {
 
   const debouncedSetCriteriaState = useMemo(
     () => debounce(setCriteriaState, 200),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
     []
   )
 
   useEffect(() => {
     debouncedSetCriteriaState(criteria)
-  }, [criteria])
+  }, [criteria, debouncedSetCriteriaState])
 
   return (
     <SystemQueryRenderer<SavedSearchAlertNameInputQuery>
       lazyLoad
       query={graphql`
-        query SavedSearchAlertNameInputQuery($attributes: PreviewSavedSearchAttributes!) {
-          previewSavedSearch(
-            attributes: $attributes
-          ) {
+        query SavedSearchAlertNameInputQuery(
+          $attributes: PreviewSavedSearchAttributes!
+        ) {
+          previewSavedSearch(attributes: $attributes) {
             displayName
           }
         }
@@ -61,7 +60,6 @@ export const SavedSearchAlertNameInput: FC<SavedSavedAlertNameInputProps> = ({
   const { values, errors, handleChange, handleBlur } = useFormikContext<
     SavedSearchAlertFormValues
   >()
-
 
   useEffect(() => {
     if (placeholder) {
--- a/src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
+++ b/src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
@@ -1,0 +1,66 @@
+import { Input, Text } from "@artsy/palette"
+import { SavedSearchAlertFormValues } from "Components/SavedSearchAlert/types"
+import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
+import { useFormikContext } from "formik"
+import { FC } from "react"
+import { graphql } from "react-relay"
+import { SavedSearchAlertNameInputQuery } from "__generated__/SavedSearchAlertNameInputQuery.graphql"
+
+interface SavedSavedAlertNameInputProps {
+  placeholder?: string
+}
+
+export const SavedSearchAlertNameInputQueryRenderer: FC = () => {
+  return (
+    <SystemQueryRenderer<SavedSearchAlertNameInputQuery>
+      lazyLoad
+      // placeholder={}
+      query={graphql`
+        query SavedSearchAlertNameInputQuery {
+          previewSavedSearch(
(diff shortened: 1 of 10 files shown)


src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
```diff
@@ -163,20 +164,20 @@ const SavedSearchAlertEditForm: React.FC<SavedSearchAlertEditFormProps> = ({
...
-                }
-                value={values.name}
-                onChange={handleChange("name")}
-                onBlur={handleBlur("name")}
-                error={errors.name}
-                maxLength={75}
-              />
+              {isFallbackToGeneratedAlertNamesEnabled ? (
+                <SavedSearchAlertNameInputQueryRenderer />
+              ) : (
```
anandaroop, line 169: Might be better to just use the new component here (as we [did](https://github.com/artsy/force/pull/12751/files#diff-5b7f8a8cb9b3b800721b3117460dd2172c5525dbc121b85fad4aa6ec17fe8542R139) in the create modal) instead of branching based on the ff.  But doing so caused some test failures that I didn't have time to resolve. We can either try to address that before merging, or update the tests when we remove the branching.

src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
```diff
@@ -0,0 +1,103 @@
...
+      render={({ props, error }) => {
+        if (error) {
+          console.error(error)
+        }
+
+        return (
+          <SavedSearchAlertNameInputFragmentContainer
+            // @ts-expect-error "No overload matches this call" TODO: why?
+            previewSavedSearch={props?.viewer?.previewSavedSearch}
+          />
```
anandaroop: Another issue I wasn't able to resolve yet. Seemed like a straightforward refactor to use this fragment container, but perhaps I missed something.

anandaroop (commented):
I called out a couple of issues below that I couldn't resolve before handing this off, but this should nonetheless be mergeable.

If anyone wants to timebox an attempt at these two improvements feel free, else I'm happy to pick this up on my return.

src/Components/SavedSearchAlert/Components/SavedSearchAlertNameInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+      variables={{ attributes: criteriaState }}
+      render={({ props, error }) => {
+        if (error) {
+          console.error(error)
+        }
+
+        return (
+          <SavedSearchAlertNameInputFragmentContainer
+            // @ts-expect-error "No overload matches this call" TODO: why?
+            previewSavedSearch={props?.previewSavedSearch}
```
olerichter00, line 46: It seems like the fragment container expects `previewSavedSearch` to not be undefined.
```
Warning: createFragmentSpecResolver: Expected prop `previewSavedSearch` to be supplied to `Relay(SavedSearchAlertNameInput)`, but got `undefined`. Pass an explicit `null` if this is intentional. 
```

Now even passing `null` results in a type error 🤔

olerichter00: To solve the issue, we could change the line to 

```suggestion
            // @ts-expect-error "No overload matches this call" TODO: why?
            previewSavedSearch={(props?.previewSavedSearch ?? null)!}
```

 or remove the fragment container and add the field `displayName` to the query and pass the value directly to `SavedSearchAlertNameInput` like this:

```suggestion
            // @ts-expect-error "No overload matches this call" TODO: why?
            displayName={props?.previewSavedSearch?.displayName}
```

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
olerichter00: I guess, it's okay to leave it like this and clean up the code when we remove the following.

olerichter00:
@anandaroop I have rebased the PR and added some improvements/fixes in https://github.com/artsy/force/pull/12751/commits/af1d067c0bfa52f3156041074445e2b564aa0ea8

anandaroop:
Thanks for the spec fix @olerichter00! I've QA'd this one more time locally, with and without the feature flag enabled, and all still looks good. Going to merge!

PR #12762, opened 2023-08-23
feat: add sale end year filter

egdbear:

This PR solves [DIA-99]

### Description

Adds Sale Year filter for Auction Results page.

cc @artsy/diamond-devs 
image

[DIA-99]: https://artsyproduct.atlassian.net/browse/DIA-99?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/SaleEndYearFilter.tsx
+++ b/src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/SaleEndYearFilter.tsx
@@ -24,15 +24,18 @@ export const SaleEndYearFilter: React.FC = () => {
     value: c?.name,
   }))
 
-  React.useEffect(() => {
-    if (saleEndYear === null) {
-      setFilter?.("saleEndYear", new Date().getFullYear())
-    }
-    if (saleStartYear === null && options.length > 0) {
-      setFilter?.("saleStartYear", parseInt(options[0].value))
-    }
-  }, [saleEndYear, saleStartYear]) // eslint-disable-line react-hooks/exhaustive-deps
+  // All options less than the end year
+  const startOptions = options.filter(
+    option =>
+      parseInt(option.value) <=
+      (saleEndYear || parseInt(options[options.length - 1]?.value))
+  )
 
+  // All options greater than the start year
+  const endOptions = options.filter(
+    option =>
+      parseInt(option.value) >= (saleStartYear || parseInt(options[0]?.value))
+  )
   return (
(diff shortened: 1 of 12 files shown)


src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilters/SaleEndYearFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const options = (
+    aggregations
+      ?.find(aggregation => aggregation.slice === "LOTS_BY_SALE_YEAR")
+      ?.counts.filter(c => c !== null) || []
+  ).map(c => ({
+    text: c?.name,
+    value: c?.name,
+  }))
+
+  React.useEffect(() => {
```
egdbear, line 27: Not sure if this makes sense here, I'm looking for _suggestions_  😄

dzucconi: Why can't this logic happen on the select callbacks?

egdbear: Do you have an example? 

```
          onSelect={year =>
                saleStartYear === null && options.length > 0
                  ? setFilter?.("saleStartYear", parseInt(options[0].value))
                  : setFilter?.("saleStartYear", parseInt(year))
              }
```

This won't work as expected :|

dzucconi: I suppose I don't follow exactly what this is doing without pulling it; I can take a look in a bit. If the effect works that's fine with me but they are much harder to reason about and frequently introduce bugs when performing something directly in callbacks is more clear.

dzucconi:
OK — so I checked this out.

We need some clarification on the behavior here. If I select an end year; presumably the start year should only show years UP TO that year, right?

I think we should actively prevent invalid inputs.

egdbear:
> OK — so I checked this out.
We need some clarification on the behavior here. If I select an end year; presumably the start year should only show years UP TO that year, right?
I think we should actively prevent invalid inputs.

Makes sense!

egdbear:
hey @dzucconi this is ready for 👀

dzucconi (approved):
👍

PR #12764, opened 2023-08-23
Revert "refactor(analyticscontext): rework analytics context provider"

dzucconi:
Reverts artsy/force#12747

Trialing a revert where some Integrity suites started consistently failing after this was initially merged, but annoyingly its proving difficult to debug Integrity. At some point, we may merge this revert and see if that fixes Integrity. If so, we'll know something here is being odd, if not it was something else that changed.


dzucconi:
@damassi just opening this to get this locally. Will update if we decide to actually merge.

PR #12770, opened 2023-08-24
fix(artist): removes verified reps for time being

dzucconi:
Removing code that touches this schema so we can deploy


egdbear:
✅ ✅ ✅ ✅ ✅

egdbear (approved):

PR #12775, opened 2023-08-24
Chore: Reinstate temporary verified rep line

jonallured:
After much pain I'm back with a PR that reinstates the temporary verified rep line that we were using to QA this data on the new Artist page header. Looks like this:

Screenshot 2023-08-24 at 11 15 59 AM

/cc @artsy/diamond-devs


jonallured:
Merging!

PR #12776, opened 2023-08-24
fix(analyticscontext): ensures updated context on route changes

dzucconi:
OK — what do we think about this. Basically: solve the same problem but without state or effects.

The problem is two-fold: we need to update the analytics context on route changes and we need to merge in a value lower in the tree. So instead of a single context that listens to the route change: two separate contexts that just accept props and a single hook that merges them.

Ultimately you don't need to know anything new as a dev: just pull in `AnalyticsContextProvider` in your app on entity pages and provide the `internalID`. And where needed use the `useAnalayticsContext` hook to pull out the values.

This shouldn't have any issues with Integrity because it is essentially the same as before: a static context.

### DIFF
--- a/src/Apps/Article/Components/ArticleBody.tsx
+++ b/src/Apps/Article/Components/ArticleBody.tsx
@@ -25,8 +25,7 @@ import { ArticleAd } from "./ArticleAd/ArticleAd"
 import { ArticleSectionFragmentContainer } from "./ArticleSection"
 import { ArticleSectionAdFragmentContainer } from "./ArticleSectionAd"
 import { OPTIMAL_READING_WIDTH } from "./Sections/ArticleSectionText"
-import { AnalyticsContext } from "System/Analytics/AnalyticsContext"
-import { OwnerType } from "@artsy/cohesion"
+import { AnalyticsContextProvider } from "System/Analytics/AnalyticsContext"
 import { ArticleNewsSourceFragmentContainer } from "./ArticleNewsSource"
 import { TopContextBar } from "Components/TopContextBar"
 
@@ -38,13 +37,7 @@ const ArticleBody: FC<ArticleBodyProps> = ({ article }) => {
   const centered = article.layout === "FEATURE" || article.layout === "NEWS"
 
   return (
-    <AnalyticsContext.Provider
-      value={{
-        contextPageOwnerId: article.internalID,
-        contextPageOwnerSlug: article.slug!,
-        contextPageOwnerType: OwnerType.article,
-      }}
-    >
+    <AnalyticsContextProvider contextPageOwnerId={article.internalID}>
       <ArticleContextProvider articleId={article.internalID}>
         {article.layout === "STANDARD" && (
           <FullBleed bg="black5" p={1}>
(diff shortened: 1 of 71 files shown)


src/Apps/Artist/ArtistApp.tsx
```diff
@@ -28,13 +23,7 @@ const ArtistApp: React.FC<ArtistAppProps> = ({ artist, children }) => {
...
       <ArtistMetaFragmentContainer artist={artist} />
 
-      <AnalyticsContext.Provider
-        value={{
-          contextPageOwnerId: artist.internalID,
-          contextPageOwnerSlug,
-          contextPageOwnerType,
-        }}
-      >
+      <Analytics contextPageOwnerId={artist.internalID}>
```
damassi, line 26: @dzucconi - Now that we're here doing this, maybe we could improve the api of this a bit even further... thoughts on 

```tsx
<Analytics internalID={artist.internalID}>
```

Nobody understands the contextPage redshift whatever idioms. if its always gonna be an `internalID`, lets just make that the prop. Ditto for `slug`

dzucconi: I'm fine with that; we still want to pull them out though as `contextPageId` etc. It's confusing because in this case it means context in the normal sense of the word, not the React sense.

damassi: Totally, just thinking that having `contextPageOwnerId` littered everywhere on init code (which actually just means `internalID` to the end user of the component) feels a bit better. What the user does with the data returned from the hook later is deliberate and intentional so yah, however that should be is fine

damassi: (ignore me!)

src/Apps/Article/Components/ArticleBody.tsx
```diff
@@ -25,8 +25,7 @@ import { ArticleAd } from "./ArticleAd/ArticleAd"
 import { ArticleSectionFragmentContainer } from "./ArticleSection"
 import { ArticleSectionAdFragmentContainer } from "./ArticleSectionAd"
 import { OPTIMAL_READING_WIDTH } from "./Sections/ArticleSectionText"
-import { AnalyticsContext } from "System/Analytics/AnalyticsContext"
-import { OwnerType } from "@artsy/cohesion"
+import { Analytics } from "System/Analytics/AnalyticsContext"
```
damassi, line 28: One last minor thing and then 👍 to merge: 
```
import { Analytics } from "System/Analytics/Analytics"
```
Keep it 1:1, more findable to intellisense and less confusing with the context-context

dzucconi: What?

dzucconi: I'm leavin this one alone; there's other exports here as well like `useAnalyticsContext`

dzucconi: Heh, leaving it named `AnalyticsInferredContext`

damassi (approved):
Thanks for pushing this through 👍

dzucconi:
LET'S SEE WHAT INTEGRITY HAS TO SAY ABOUT THIS @mzikherman

PR #12778, opened 2023-08-24
chore: Tracking corrections, refactor address verification flow

erikdstock:
Resolves [EM-1402]

cc @artsy/emerald-devs 
### Description

This PR updates our address verification flow with 🔒[corrected tracking values](https://artsy.slack.com/archives/C02JHHHKP5K/p1693242656034689) and refactors to remove unnecessary `useState/useEffect` hooks, which may be contributing to some infinite loops we've seen in testing. 

This PR replaces #12736 which grew too big, so I am reopening with a clean slate.

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -12,7 +12,7 @@ import {
   Spacer,
   Text,
 } from "@artsy/palette"
-import { useCallback, useEffect, useState } from "react"
+import { useCallback, useEffect, useRef, useState } from "react"
 import { useTracking } from "react-tracking"
 import { useSystemContext } from "System/SystemContext"
 import { ContextModule, OwnerType } from "@artsy/cohesion"
@@ -55,14 +55,18 @@ export interface AddressValues {
   postalCode: string
 }
 
-enum ModalType {
+enum FlowType {
+  ERROR_IMMEDIATE_CONFIRM = "error_immediate_confirm",
+  IMMEDIATE_CONFIRM = "immediate_confirm",
   SUGGESTIONS = "suggestions",
   REVIEW_AND_CONFIRM = "review_and_confirm",
 }
 
-const modalTitles: Record<ModalType, string> = {
-  [ModalType.SUGGESTIONS]: "Confirm your delivery address",
-  [ModalType.REVIEW_AND_CONFIRM]: "Check your delivery address",
+const modalTitles: Record<FlowType, string | null> = {
+  [FlowType.SUGGESTIONS]: "Confirm your delivery address",
+  [FlowType.REVIEW_AND_CONFIRM]: "Check your delivery address",
+  [FlowType.ERROR_IMMEDIATE_CONFIRM]: null,
+  [FlowType.IMMEDIATE_CONFIRM]: null,
 } as const
 
 interface AddressOption {
@@ -153,17 +157,93 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
   onChosenAddress,
   onClose,
 }) => {
-  const [modalType, setModalType] = useState<ModalType | null>(null)
-  const [addressOptions, setAddressOptions] = useState<AddressOption[]>([])
-  const [
-    selectedAddressKey,
-    setSelectedAddressKey,
-  ] = useState<AddressOptionKey | null>(null)
+  const [showModal, setShowModal] = useState(false)
+  const didLoad = useRef(false)
+
+  let modalType: FlowType = FlowType.REVIEW_AND_CONFIRM
+  let addressOptions: AddressOption[] = []
 
   const {
     trackViewedModal,
     trackClickedModal,
   } = useAddressVerificationTracking()
+
+  const error = (verifyAddress.verifyAddressOrError as VerifyAddressErrorType)
+    ?.mutationError
+  const suggestedAddresses = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.suggestedAddresses
+  const inputAddress = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.inputAddress
+  const verificationStatus = (verifyAddress.verifyAddressOrError as VerifyAddressSuccessType)
+    ?.verificationStatus
+
+  const hasError = Boolean(error)
+
+  if (hasError) {
+    modalType = FlowType.ERROR_IMMEDIATE_CONFIRM
+  } else {
+    const inputOption: AddressOption = {
+      ...(inputAddress as any),
+      key: USER_INPUT,
+      recommended: false,
+    }
+    if (verificationStatus === "VERIFIED_NO_CHANGE") {
+      modalType = FlowType.IMMEDIATE_CONFIRM
+      addressOptions = [inputOption]
+    } else {
+      if (verificationStatus === "VERIFIED_WITH_CHANGES") {
+        modalType = FlowType.SUGGESTIONS
+
+        const suggestedOptions = (suggestedAddresses || [])
+          .slice(0, 1)
+          .map(({ address, lines }: any, index) => ({
+            key: `suggestedAddress-${index}`,
+            recommended: index === 0,
+            lines: lines,
+            address: address,
+          }))
+        addressOptions = [...suggestedOptions, inputOption]
+      } else {
+        addressOptions = [inputOption]
+        modalType = FlowType.REVIEW_AND_CONFIRM
+      }
+    }
+  }
+
+  const [
+    selectedAddressKey,
+    setSelectedAddressKey,
+  ] = useState<AddressOptionKey | null>(addressOptions[0]?.key)
+
+  // perform only once on mount
+  useEffect(() => {
+    if (!didLoad.current) {
+      if (modalType === FlowType.ERROR_IMMEDIATE_CONFIRM) {
+        const verifiedBy = AddressVerifiedBy.USER
+        const fallbackOption = fallbackFromFormValues(verificationInput)
+        onChosenAddress(verifiedBy, fallbackOption.address)
+        didLoad.current = true
+      } else if (modalType === FlowType.IMMEDIATE_CONFIRM) {
+        onChosenAddress(AddressVerifiedBy.ARTSY, addressOptions[0].address!)
+        didLoad.current = true
+      } else {
+        didLoad.current = true
+
+        trackViewedModal({
+          option: modalType,
+          subject: modalTitles[modalType]!,
+        })
+      }
+      setShowModal(true)
+    }
+  }, [
+    didLoad,
+    addressOptions,
+    modalType,
+    verificationInput,
+    onChosenAddress,
+    trackViewedModal,
+  ])
 
   const chooseAddress = useCallback(() => {
     if (!selectedAddressKey) return
@@ -178,15 +258,15 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
         : AddressVerifiedBy.ARTSY
 
     if (selectedAddress) {
-      setModalType(null)
+      setShowModal(false)
       onChosenAddress(verifiedBy, selectedAddress.address)
     }
   }, [addressOptions, onChosenAddress, selectedAddressKey])
 
   const handleClose = () => {
     trackClickedModal({
-      option: null,
-      label: "close icon",
+      label: "close modal",
+      subject: modalTitles[modalType],
     })
     onClose()
   }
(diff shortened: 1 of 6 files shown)


src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -195,90 +113,129 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+      onChosenAddress(AddressVerifiedBy.ARTSY, addressOptions[0].address!)
+      didLoad.current = true
+    } else {
+      setShowModal(true)
+      didLoad.current = true
+
+      trackViewedModal({
+        subject: modalTitle!,
+      })
+    }
```
erikdstock, line 173: a previous large useEffect that was running unpredictably has been mostly moved into the imperative outer layer of the code. Instead, we put only the actual side effects (which should only occur once) in here, and make sure it only loads once.

Maybe the `didLoad.current = true` line should just move below L172.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -479,3 +439,73 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
...
+          context_module: ContextModule.ordersShipping,
+          context_page_owner_type: OwnerType.ordersShipping,
+          context_page_owner_id: contextPageOwnerId,
+          user_id: userId,
+          flow: "user adding shipping address",
+          subject,
+        } as ValidationAddressViewed)
+      },
+      [contextPageOwnerId, trackEvent, userId]
+    ),
```
erikdstock, line 485: tracking logic moved into this local helper to clean up the main component. I'm not sure if all these `useCallback()`s are necessary.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -195,90 +113,129 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+    subject: string
+    option?: string
+  }) => {
+    trackClickedModal({
+      subject,
+      option,
+      label,
+    })
+    onClose()
+  }
```
erikdstock, line 217: Realized that closing the modal is a different tracking event from choosing an address so i need to disambiguate this bit.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -757,208 +758,215 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
+                    <BorderedRadio value="SHIP" label="Shipping" />
+                    <BorderedRadio
+                      value="PICKUP"
+                      label="Arrange for pickup (free)"
+                      data-test="pickupOption"
+                    >
+                      <Collapse open={shippingOption === "PICKUP"}>
+                        <Text variant="xs" color="black60">
+                          After your order is confirmed, a specialist will
+                          contact you to coordinate pickup.
```
erikdstock, line 799: 🤔

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -757,208 +758,215 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-            flexDirection="column"
-            style={isCommittingMutation ? { pointerEvents: "none" } : {}}
-          >
-            {/* TODO: Make RadioGroup generic for the allowed values,
+    <AnalyticsContext.Provider
+      value={{
+        contextPageOwnerId: order.internalID,
+        contextPageOwnerType: OwnerType.ordersShipping,
+      }}
+    >
```
erikdstock, line 766: Need to double check that this is still necessary after #12776 which just merged.

erikdstock (commented):
Small comments, I'll push followups shortly.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -195,90 +113,129 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
     }
+  }
+
+  const modalTitle = modalTitles[verificationPath]
+
+  const [selectedAddressKey, setSelectedAddressKey] = useState<
+    AddressOptionKey | undefined
+  >(addressOptions[0]?.key)
+
+  // perform only once on mount
```
damassi, line 155: This comment isn't actually accurate, since there are dependencies in the `useEffect` array. If you want it to only perform once, need to explicitly use `[]`

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -479,3 +439,73 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
...
+        trackEvent({
+          action: ActionType.clickedValidationAddressOptions,
+          context_module: ContextModule.ordersShipping,
+          context_page_owner_type: OwnerType.ordersShipping,
+          context_page_owner_id: contextPageOwnerId,
+          user_id: userId,
+          subject,
+          ...(typeof subject !== "undefined" && { subject }),
+          ...(typeof option !== "undefined" && { option }),
+          ...(typeof label !== "undefined" && { label }),
```
damassi, line 505: Can this be accomplished in a less ambiguous way?

src/Apps/Order/Routes/Shipping/index.tsx
damassi: This is no longer valid as of that PR, as we've changed the API

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Ah, i guess in this case i don't need the ref - i thought there was a risk of double-firing events without it and was hesitant to disable the exhaustive deps rule but i can do it if you say its ok.

damassi: If you want something to fire once, the effect fires on mount -- and if there are no deps to track, it will never fire again. The linter doesn't know your intent, so it will always warn.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Moved down, then removed completely because there was nothing to wait for before loading - firing once on mount (`[]` deps array) does the trick.

src/Apps/Order/Routes/Shipping/index.tsx
erikdstock: Thanks. I think I updated for that - Let me know if i missed something

damassi: Looks good!

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -479,3 +426,90 @@ export const AddressVerificationFlowQueryRenderer: React.FC<{
...
+    ),
+
+    trackClosedModal: useCallback(
+      ({ label, subject }: { label: string; subject: string }) => {
+        const event: ClickedCloseValidationAddressModal = {
+          action: ActionType.clickedCloseValidationAddressModal,
+          context_module: ContextModule.ordersShipping,
+          context_page_owner_type: OwnerType.ordersShipping,
+          context_page_owner_id: contextPageOwnerId!,
+          option: "close",
```
starsirius, line 506: "close" is not a valid option from the [spreadsheet](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0). Looks like this needs to be dynamic.

Screen Shot 2023-08-25 at 1 46 12 PM

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -195,90 +114,114 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+  return !showModal ? null : (
     <ModalDialog
       width={["100%", 590]}
       height={["100vh", "auto"]}
       m={[0, 2]}
-      title={modalTitles[modalType]}
-      onClose={handleClose}
+      title={modalTitle!}
+      onClose={() => {
+        handleCloseModal({ label: "close modal", subject: modalTitle! })
```
starsirius, line 221: If I remember correctly, the label should be `null` when the modal is closed via clicking "X", clicking away, or hitting <ESC>. "close modal" is not a valid label from the [spreadsheet](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0).

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -317,7 +260,12 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
           <Spacer y={4} />
           <Flex width="100%" justifyContent="space-between">
             <Button
-              onClick={handleBackToEdit}
+              onClick={() => {
+                handleCloseModal({
```
starsirius, line 264: If I understand correctly from the [spreadsheet](https://docs.google.com/spreadsheets/d/1c65vugkF2C4YuXwBXno8aqsBh4B2fTwxiLmgY_MhjWc/edit#gid=0), we want to track both the `ClickedValidationAddressOptions` and `ClickedCloseValidationAddressModal` event here.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -328,13 +276,8 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
               onClick={() => {
                 if (selectedAddressKey) {
-                  trackEvent({
-                    action_type: "clickedValidationAddress",
-                    context_module: ContextModule.ordersShipping,
-                    context_page_owner_type: OwnerType.ordersShipping,
-                    context_page_owner_id: contextPageOwnerSlug,
-                    user_id: userId,
-                    subject: "Confirm your delivery address",
+                  trackSelectedAddress({
```
starsirius, line 279: Same as https://github.com/artsy/force/pull/12778/files?diff=unified&w=0#r1305979017.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -372,13 +315,9 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
             <Button
               onClick={() => {
-                trackEvent({
-                  action_type: "clickedValidationAddress",
-                  context_module: ContextModule.ordersShipping,
-                  context_page_owner_type: OwnerType.ordersShipping,
-                  context_page_owner_id: contextPageOwnerSlug,
-                  user_id: userId,
-                  subject: "Check your delivery address",
+                trackSelectedAddress({
```
starsirius, line 318: Same as https://github.com/artsy/force/pull/12778/files?diff=unified&w=0#r1305979017.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -389,7 +328,15 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
               Use This Address
             </Button>
             <Spacer x={1} />
-            <Button onClick={handleEditAddress} flex={1}>
+            <Button
+              onClick={() => {
+                handleCloseModal({
```
starsirius, line 333: Same as https://github.com/artsy/force/pull/12778/files?diff=unified&w=0#r1305979017.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -195,90 +114,114 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+  useEffect(() => {
+    if (verificationPath === VerificationPath.ERROR_IMMEDIATE_CONFIRM) {
+      const fallbackOption = fallbackFromFormValues(verificationInput)
+      onChosenAddress(AddressVerifiedBy.USER, fallbackOption.address)
+    } else if (verificationPath === VerificationPath.IMMEDIATE_CONFIRM) {
+      onChosenAddress(AddressVerifiedBy.ARTSY, addressOptions[0].address!)
+    } else {
+      setShowModal(true)
+
+      trackViewedModal({
```
starsirius, line 166: Question: firing the viewed event right after setting a state seems to be a few steps early than the modal is rendered to the user. Is this generally safe?

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Ah, i was looking into the palette docs and couldn't find a handler for clicking outside the modal so i decided to just put this for now. I also ran into an issue with some of these (like `option` where the cohesion scheme didn't permit null even though it was the specified value.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -206,11 +206,7 @@ class ShippingTestPage extends OrderAppTestPage {
   }
 }
 
-describe("Shipping [testing-library]", () => {
-  describe("US address verification", () => {})
-})
-
-describe("Shipping [enzyme]", () => {
+describe("Shipping", () => {
```
erikdstock, line 209: Removed these empty tests that i accidentally merged recently 🙊

erikdstock (commented):
I met with Dayane to get final clarification on tracking behaviors we want here. In short:
- Track dismissing modal as `clickedCloseValidationAddressModal` with option as `null` and label as the 'back to edit/edit address' button text, or when closing via `x` as `null`. This label value as `null` violates the cohesion schema and will be ticketed as a followup, using coercions for now.
- Track choosing an address as `clickedValidationAddressOption`
- use proper action types and rename the key from `action_type` to `action`

Summing up the changes to behavior as viewed by analytics:
1. action_type has been replaced with the correct key action
2. action_type: clickedValidationAddress is now the corrected action: clickedValidationAddressOptions
3. Closing modal via X behavior should be the same
4. Closing modal via 'back to edit' was previously action_type: clickedValidationAddress  (see 2 above) - now it will be action: clickedCloseValidationAddress with label 'Back to Edit'

This is ready for review and merge pending the comments i've left below.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Confirmed behavior with dayane and updated per her request!

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: resolved in latest commit [9bb308f](https://github.com/artsy/force/pull/12778/commits/9bb308f0717dc996fc0314135e65404ad56d3a85) after working with dayane.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Correct, fixed with a type coercion for now, will ticket updating cohesion to permit this.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: Updated to simply fire once on mount with `[]`.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: If we come back to this and want to add values that are not present immediately on render (eg tracking a value that is null to start but then changed with a `set_` we would need the additional `useRef` that I removed in that commit 👆 . Otherwise this is the right way to do it according to web-practice.

starsirius: Yeah I wasn't sure if there will be cases that modal is not actually rendered, or more impression tracking fired than necessary when react batches state update. Sounds like it's the recommended way. 👍 

Linking to the [discussion](https://artsy.slack.com/archives/C2TQ4PT8R/p1693247057861649) for completeness.

PR #12789, opened 2023-08-25
feat(search): focus search input on forward slash

dzucconi:
This is just one of those things I keep trying to do and being vaguely frustrated when it doesn't work. I'm a little unclear on the origin of this keyboard shortcut but it is pervasive at this point.

### DIFF
--- a/src/Components/Search/NewSearch/SearchBarInput.tsx
+++ b/src/Components/Search/NewSearch/SearchBarInput.tsx
@@ -1,5 +1,12 @@
 import { AutocompleteInput, useUpdateEffect } from "@artsy/palette"
-import { ChangeEvent, FC, useCallback, useState } from "react"
+import {
+  ChangeEvent,
+  FC,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react"
 import { useTranslation } from "react-i18next"
 import { createRefetchContainer, graphql, RelayRefetchProp } from "react-relay"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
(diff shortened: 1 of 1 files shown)


src/Components/Search/NewSearch/SearchBarInput.tsx
```diff
@@ -177,8 +191,33 @@ const SearchBarInput: FC<SearchBarInputProps> = ({
...
+    const handleKeyUp = ({ target, key }: KeyboardEvent) => {
+      if (!ref.current || key !== "/") return
+
+      const tag =
+        (target && (target as HTMLElement).tagName.toLowerCase()) || ""
+
+      // Ignore if an input has focus
+      if (["input", "textarea", "select"].includes(tag)) {
+        return
+      }
```
anandaroop, line 207: **question (non-blocking):** Does a `select` fire keyup events?

anandaroop (approved):
Nice!

>  I'm a little unclear on the origin of this keyboard shortcut 

Dare I say `vi`? 

Feels like a geeky shortcut to include in a general-audience product, but Gmail has had it forever and I've often wished we did too 👍🏽

src/Components/Search/NewSearch/SearchBarInput.tsx
dzucconi: Yep, you can tab into one then start typing which will select corresponding options.

dblandin (approved):
Nice 💯

nickskalkin (dismissed):

damassi:
Can we merge this @dzucconi ?

PR #12807, opened 2023-08-30
fix(dia-103): ensures top context bar only links to internal routes

dzucconi:
Closes [DIA-103](https://artsyproduct.atlassian.net/browse/DIA-103)

This is just a minor clean up related to the ticket wherein one could alter the query string to execute a script on click. This also now prevents someone from linking to an external site.

In the process I noticed some code that didn't actually work and uncovered some bugs in the consignment flow.

[DIA-103]: https://artsyproduct.atlassian.net/browse/DIA-103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
@@ -208,15 +208,16 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
   return (
     <>
       {isFirstStep && !artworkId ? (
+        // FIXME: Inappropriate component for this; don't replicate browser back button
         <TopContextBar
           displayBackArrow
           hideSeparator
           onClick={() => router.go(-1)}
-          redirectTo="/sell"
         >
           Back
         </TopContextBar>
       ) : (
+        // FIXME: Should not have a generic "Back" label that links to a specific location
         <TopContextBar displayBackArrow hideSeparator href={backTo}>
           Back
         </TopContextBar>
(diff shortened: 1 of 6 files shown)


src/Components/TopContextBar.tsx
```diff
@@ -39,14 +47,14 @@ export const TopContextBar: React.FC<TopContextBarProps> = ({
...
               onClick,
             }
-          : {})}
-        {...(redirectTo
-          ? {
-              as: RouterLink,
-              textDecoration: "none",
-              onClick,
-            }
-          : {})}
```
dzucconi, line 49: This didn't actually _do_ anything with the string; but even if it did, we wouldn't want to do this. I'm really not sure what the original intent was, because we have an `href` prop to begin with. Since it was kind of acting as just a way to trigger the `onClick` handler, I've corrected and added real support for that below.

src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
```diff
@@ -208,15 +208,16 @@ export const ContactInformation: React.FC<ContactInformationProps> = ({
   return (
     <>
       {isFirstStep && !artworkId ? (
+        // FIXME: Inappropriate component for this; don't replicate browser back button
         <TopContextBar
           displayBackArrow
           hideSeparator
           onClick={() => router.go(-1)}
```
dzucconi, line 215: I think if you ever see this in our codebase it's probably a sign that something is wrong. We don't want to replicate the browser back button.

mzikherman (approved):
Thanks for the fix!

dariakoko (commented):
Thank you!

PR #12810, opened 2023-08-30
feat: report destination_could_not_be_geocoded error (EMI-1332)

iskounen:
The type of this PR is: Feat
This PR solves [EMI-1332]

This PR is a follow-up to artsy/exchange#1763 that displays an error message for the error that is newly being returned from Exchange when a user submits a non-existing shipping address.

It also makes use of the new `getErrorDialogCopy` function created in #12803 to house the new copy. That function is still being tested out as I work on checkout dead-ends so any feedback on it is appreciated.

[EMI-1332]: https://artsyproduct.atlassian.net/browse/EMI-1332?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Utils/getErrorDialogCopy.tsx
+++ b/src/Apps/Order/Utils/getErrorDialogCopy.tsx
@@ -4,6 +4,7 @@ import React from "react"
 const SUPPORT_EMAIL = "orders@artsy.net"
 
 export enum ErrorDialogs {
+  DestinationCouldNotBeGeocoded = "destination_could_not_be_geocoded",
   CurrencyNotSupported = "currency_not_supported",
 }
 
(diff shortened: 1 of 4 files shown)


src/Apps/Order/Utils/getErrorDialogCopy.tsx
```diff
@@ -4,6 +4,7 @@ import React from "react"
 const SUPPORT_EMAIL = "orders@artsy.net"
 
 export enum ErrorDialogs {
+  DestinationCouldNotBeGeocoded = "destination_could_not_be_geocoded",
   CurrencyNotSupported = "currency_not_supported",
 }
```
erikdstock, line 9: non-blocking: Similar to my comment here https://github.com/artsy/force/pull/12803#discussion_r1310658734, we could make these values a graphql enum for free always up-to-date types.

erikdstock (commented):
Looks good to me!

PR #12813, opened 2023-08-31
fix: Alert edit form scroll bug

olerichter00:
Addresses [ONYX-283]

## Description

This fixes the scroll bug in the Saved Search Alert edit form. There are two nested scroll views. The issue is that the user cannot scroll to the end of the form component when focusing on the form scroll view.

Screenshot 2023-08-31 at 13 40 58

[ONYX-283]: https://artsyproduct.atlassian.net/browse/ONYX-283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
@@ -27,7 +27,6 @@ import { MetaTags } from "Components/MetaTags"
 import { SavedSearchAlertsEmptyResults } from "./Components/SavedSearchAlertsEmptyResults"
 import { SavedSearchAlertEditFormDesktop } from "./Components/SavedSearchAlertEditFormDesktop"
 import { Sticky } from "Components/Sticky"
-import { useNavBarHeight } from "Components/NavBar/useNavBarHeight"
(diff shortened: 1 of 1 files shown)


src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -186,17 +184,12 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
                     borderLeftColor="black15"
                   >
                     <Sticky bottomBoundary="#content-end">
-                      <Box
-                        overflowY="auto"
-                        maxHeight={`calc(100vh - ${desktop}px)`}
-                      >
```
olerichter00, line 192: I guess, this is not needed and the form itself doesn't need a scrollbar.

dariakoko (approved):
Looks gooood
Have you tested it with mWeb?

olerichter00:
> Have you tested it with mWeb?

Good point. I've tested it on mWeb and it actually even uses another component for the layout 👍

PR #12818, opened 2023-08-31
fix: remove sale year params

egdbear:

### Description

I noticed when including `saleEndYear` and `saleStartYear` as url params, that we were sending  to Metaphysics string as parameter from these variables.
This PR remove the argument support


jonallured (approved):
oops!

PR #12823, opened 2023-08-31
feat: [EMI-1409] Address autocomplete: initial implementation

erikdstock:

This PR implements a basic address autocomplete using our autocomplete provider. Before merging it will be hidden behind a feature flag with basic unit tests in place. The hope is that we can enable it relatively soon and begin testing in staging to find and iterate on the hopefully minor papercuts which follow.

This PR solves [EMI-1409] and most of [EMI-1376](https://artsyproduct.atlassian.net/browse/EMI-1376)

### Description

This PR adds a hook that can be used to query for address suggestions and suggest them to a user with the Palette AutocompleteInput component. We will continue iterating on the functionality as part of the autocomplete epic.

Notes:
- in the current, draft state, the code is messy due to some implementation being copy-pasted from smarty docs
- we used `fetch` instead of the javascript sdk supplied by smarty because importing it breaks force.
- when accepting an address with sub-entries (generally apt numbers) the react hook submits a special query for the sub-entries, but in order to see them you have to click back into the input. So far, we haven't been able to fix this, but we aren't sure this feature will be part of the initial scope.
- For now autocomplete requests are debounced at around 700ms, which makes the form seem less responsive than it would be on release. This is to reduce the number of requests we make while testing. When you want to test your suggestions, stop typing.

[EMI-1409]: https://artsyproduct.atlassian.net/browse/EMI-1409?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMI-1376]: https://artsyproduct.atlassian.net/browse/EMI-1376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
+import { AutocompleteInputOptionType } from "@artsy/palette"
+import { Address } from "Components/AddressForm"
+import { useDebounce } from "Utils/Hooks/useDebounce"
+import { getENV } from "Utils/getENV"
+import { useCallback, useState } from "react"
+
+// NOTE: Due to the format of this key (a long string of numbers that cannot be parsed as json)
+// This key must be set in the env as a json string like SMARTY_EMBEDDED_KEY_JSON_JSON='{ "key": "xxxxxxxxxxxxxxxxxx" }'
```
erikdstock, line 8: Add to .env.example maybe? Not necessary for development though.
```suggestion
// This key must be set in the env as a json string like SMARTY_EMBEDDED_KEY_JSON_JSON={ "key": "xxxxxxxxxxxxxxxxxx" }
```

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+const smartyCreds = getENV("SMARTY_EMBEDDED_KEY_JSON")
+console.log({ smartyCreds })
+const key = smartyCreds.key
```
erikdstock, line 11: had a weird time extracting the key, but in retrospect this may have been another issue.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+  const fetchSuggestions = useCallback(
+    async (searchParam: string, selectedParam?: string) => {
+      console.log({ key })
+      if (!key) return null
+      let url = `https://us-autocomplete-pro.api.smarty.com/lookup?key=${encodeURIComponent(
+        key
+      )}&prefer_ratio=3&search=${encodeURIComponent(searchParam)}`
+
+      if (selectedParam) {
+        url += `&selected=${encodeURIComponent(selectedParam)}`
```
erikdstock, line 43: We can do this more gracefully - see https://stackoverflow.com/questions/35038857/setting-query-string-using-fetch-get-request

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+  address: Omit<Address, "name">
+  entries: number
+}
+
+export const useAddressAutocomplete = () => {
+  const [result, setResult] = useState<SuggestionJSON[]>([])
+
+  // Add feature flag checks including country, etc here later
+  // console.log({ key, kind: typeof key })
+  const enabled = !!key
```
erikdstock, line 32: we should put a check somewhere for this that can incorporate US vs international, feature flags, etc. I was thinking we would just expose it here and clients can render a fallback plain input if autocomplete is not supported.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+      if (!key) return null
+      let url = `https://us-autocomplete-pro.api.smarty.com/lookup?key=${encodeURIComponent(
+        key
+      )}&prefer_ratio=3&search=${encodeURIComponent(searchParam)}`
+
+      if (selectedParam) {
+        url += `&selected=${encodeURIComponent(selectedParam)}`
+      }
+
+      console.log({ url })
```
erikdstock, line 46: lots of these to remove.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+    },
+    [fetchSuggestions]
+  )
+
+  // debounce requests
+  const debouncedFetchForAutocomplete = useDebounce({
+    // high debounce for testing to not eat up our credits
+    delay: 700,
+    callback: fetchForAutocomplete,
+  })
```
erikdstock, line 85: Avoid hammering smarty's autocomplete api on every keypress - right now the value is very high to avoid blowing through promo credits in testing.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+      suggestion.secondary +
+      secondaryExtraInformation +
+      " " +
+      suggestion.city +
+      ", " +
+      suggestion.state +
+      " " +
+      suggestion.zipcode
+    )
+  }
```
erikdstock, line 108: easy refactor, adapted from [buggy code example in the smarty docs.](https://www.smarty.com/docs/cloud/us-autocomplete-pro-api#pro-http-response-status)

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const autocompleteOptions: Array<AddressSuggestion> = result.map(
+    suggestion => {
+      const text = buildAddressText(suggestion)
```
erikdstock, line 112: smarty json could benefit from a type. it's quite simple.

src/Components/AddressForm.tsx
```diff
@@ -74,6 +85,12 @@ export const AddressForm: React.FC<AddressFormProps> = ({
   const [address, setAddress] = React.useState(addressFromProp)
   const [prevValue, setPrevValue] = React.useState(value)
 
+  const { autocompleteOptions, fetchForAutocomplete } = useAddressAutocomplete()
+
+  React.useEffect(() => {
+    console.log({ autocompleteOptions })
+  }, [autocompleteOptions])
```
erikdstock, line 92: remove

src/Components/AddressForm.tsx
```diff
@@ -84,6 +101,13 @@ export const AddressForm: React.FC<AddressFormProps> = ({
   const changeEventHandler = (key: keyof Address) => (
     ev: React.FormEvent<HTMLInputElement>
   ) => {
+    const shouldFetch = !!(key === "addressLine1")
```
erikdstock, line 104: unnecessary `!!`

src/Components/AddressForm.tsx
```diff
@@ -167,15 +193,46 @@ export const AddressForm: React.FC<AddressFormProps> = ({
...
+          id="AddressForm_addressLine1"
+          placeholder="Street address"
+          title="Address line 1"
+          value={value?.addressLine1}
+          onChange={changeEventHandler("addressLine1")}
+          error={getError("addressLine1")}
+          data-test="AddressForm_addressLine1"
+        /> */}
+        {/* TODO: Make this work with autocomplete input */}
+        <AutocompleteInput<AddressSuggestion>
```
erikdstock, line 207: Should probably render this as a fallback if autocomplete is not enabled - otherwise the input has a 🔍 in it. Alternately we could make the AutocompleteInput icon overridable via props.

erikdstock (commented):
Notes to self for cleanup.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,135 @@
...
+      return {
+        text,
+        value: text,
+        entries: suggestion.entries,
+        address: {
+          addressLine1: suggestion.street_line,
+          addressLine2: suggestion.secondary,
```
erikdstock, line 119: We observed that `secondary` should actually be appended to `street_line` as is common with things like apartment numbers.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
erikdstock: There should also be a comma before the city. Something like 
```ts
const buildingAddress = suggestion.street_line
if (suggestion.secondary) buildingAddress += ` ${suggestion.secondary}`

return [`${buildingAddress}, ${suggestion.city}`, suggestion.state, suggestion.zipcode].join(" ")
```

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const useAddressAutocomplete = () => {
+  const [result, setResult] = useState<SuggestionJSON[]>([])
+
+  // Add feature flag checks including country, etc here later
+  // console.log({ key, kind: typeof key })
+
+  const isAPIKeyPresent = !!key
+  const isFeatureFlagEnabled = !!useFeatureFlag("address_autocomplete_us")
+  const isFeatureEnabled = isAPIKeyPresent && isFeatureFlagEnabled
```
erikdstock, line 36: When calling the hook, we might want to pass the search country in to select which api to use - and right now only US would be enabled.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
erikdstock: The logic here could be shared with the logic in our buildAddressText code.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
dzucconi: I don't think this hook debounces correctly. Verify but you should probably use `useMemo` in conjunction with lodash's `debounce`.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,145 @@
...
+  const isUSAddress = address.country === "US"
+  const isAPIKeyPresent = !!apiKey
+  const isFeatureFlagEnabled = !!useFeatureFlag("address_autocomplete_us")
+  const isAddressAutocompleteEnabled =
+    isAPIKeyPresent && isFeatureFlagEnabled && isUSAddress
+
+  useEffect(() => {
+    if (result.length > 0 && !isUSAddress) {
+      setResult([])
+    }
```
dzucconi, line 40: Why wouldn't we just do this in the callback itself. This doesn't need to be an effect as far as I can see.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
```diff
@@ -0,0 +1,145 @@
+import { AutocompleteInputOptionType } from "@artsy/palette"
+import { Address } from "Components/AddressForm"
+import { useFeatureFlag } from "System/useFeatureFlag"
+import { useDebounce } from "Utils/Hooks/useDebounce"
+import { getENV } from "Utils/getENV"
+import { useCallback, useEffect, useState } from "react"
+
+// NOTE: Due to the format of this key (a long string of numbers that cannot be parsed as json)
+// This key must be set in the env as a json string like SMARTY_EMBEDDED_KEY_JSON={ "key": "xxxxxxxxxxxxxxxxxx" }
+const { key: apiKey } = getENV("SMARTY_EMBEDDED_KEY_JSON") || { key: "" }
```
dzucconi, line 10: This isn't going to parse the JSON

erikdstock: It doesn't, the json gets parsed in [config.ts](https://github.com/artsy/force/blob/1fcd1bb6bc35a36c89f894723d40b4bd21b08a11/src/Server/config.ts#L140-L153). Because the smarty key is a long string of exclusively numbers and because of the peculiarity of JSON parsing, this string passed to json.parse gets interpreted as a number and rounded - but it is not a number, it is a key. So the workaround to this is to instead give it something else to parse, a JSON object, which will become a javascript object, with an actual string value inside.

source: [Stack Overflow](https://stackoverflow.com/questions/15689790/parse-json-in-javascript-long-numbers-get-rounded)

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
erikdstock: The case here is when a user changes their country and then goes back - we don't want clicking into the address field to show them old suggestions. For now this is exclusively a US feature, so it's just US or not US, and if it isn't, we want to zero out the suggestions.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
dzucconi: Yowch

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
erikdstock: Maybe a followup for later - currently the line 2 input specifically says apt number even though it is not.

OTOH, doing this right would help us avoid the address verification modal if we ignore the zipcode +4 for VERIFIED_NO_CHANGES results in the future.

Probably a followup for design since this line 2 label/placeholder text is very standard, though wrong, but the proposal would be:
- stick secondary info on line 1 where the verification api expects it
- update verification api to consider a zip code that is correct but lacking the zip+4 as still verified without changes
- force receives result and proceeds without showing the user a modal.

iskounen (commented):
The latest batch of changes look good.

The only (minor) concern I have is that useAddressAutocomplete is responsible for determining whether autocompletion should be happening. I think that AddressForm should be responsible for determining that to simplify the code a bit.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -38,6 +34,11 @@ export const useAddressAutocomplete = (
   ) => void
 } => {
   const [result, setResult] = useState<ProviderSuggestion[]>([])
+
+  // NOTE: Due to the format of this key (a long string of numbers that cannot be parsed as json)
+  // This key must be set in the env as a json string like SMARTY_EMBEDDED_KEY_JSON={ "key": "xxxxxxxxxxxxxxxxxx" }
+  const { key: apiKey } = getENV("SMARTY_EMBEDDED_KEY_JSON") || { key: "" }
```
iskounen, line 40: @erikdstock @TMcMeans I moved this line into this function so that I could write tests with and without the API key. When this line was at the top of the file, I was only able to mock the environment variable once per test file.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -74,7 +75,7 @@ export const useAddressAutocomplete = (
       const json = await response.json()
       return json
     },
-    []
+    [apiKey]
```
erikdstock, line 78: since the ENV should never change i think it is technically ok to leave this out and disable the linter.

erikdstock:
> The only (minor) concern I have is that useAddressAutocomplete is responsible for determining whether autocompletion should be happening. I think that AddressForm should be responsible for determining that to simplify the code a bit.

@iskounen I'm sympathetic to this but maybe need more information. Another side effect of the current setup is that it makes sure autocomplete results are reset when the country changes, which prevents a behavior of showing old results when you click back into the line 1 input, but if you are ok with this for a first iteration I'm totally open to revisiting.

src/Apps/Order/Routes/Shipping/useAddressAutocomplete.ts
erikdstock: I tested once more - it only requests when you stop typing (and have entered at least 5 chars). Ignore my bad typing and the content of the logging here - each one log is coming immediately before the `fetch(url)` line above, and appears to be nearly 1s after i stop typing- so I do think it is working.

PR #12833, opened 2023-09-06
Migrate alerting for missing OwnerTypes into refactored AnalyticsContext

erikdstock:
Paired on this with @TMcMeans, if we've missed some critical consideration or otherwise don't want to merge that it's totally ok.

### Description
Following up on [https://github.com/artsy/force/pull/12776](https://github.com/artsy/force/pull/12776/files#diff-91d000540513622973ef9d699e2cfd5ae4444c21ef601128c970314655d42f5eR113-R117), which fixed and refactored a lot of our analytics context, this PR:

- migrates the `OwnerType orders was not part of @artsy/cohesion's schema` messaging to our `useMemo` hook where it is set
- removes references to a deprecated method that did an earlier version of the same thing
- fixes a bug in our alerting code which was testing the keys rather than the values of the OwnerType enum.

### DIFF
--- a/src/System/Analytics/AnalyticsContext.tsx
+++ b/src/System/Analytics/AnalyticsContext.tsx
@@ -1,6 +1,9 @@
 import { FC, ReactNode, createContext, useContext, useMemo } from "react"
 import { OwnerType, PageOwnerType } from "@artsy/cohesion"
 import { camelCase } from "lodash"
+import { getENV } from "Utils/getENV"
+
+const SHOW_ANALYTICS_CALLS = getENV("SHOW_ANALYTICS_CALLS")
 
 const AnalyticsContext = createContext<{
   contextPageOwnerId?: string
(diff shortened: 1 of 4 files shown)


dzucconi:
Thanks for this!

TMcMeans:
lgtm!!!

TMcMeans (approved):

src/System/Analytics/AnalyticsContext.tsx
```diff
@@ -63,9 +66,16 @@ export const AnalyticsContextProvider: FC<AnalyticsContextProviderProps> = ({
...
-    return (path
+    const ownerType = (path
       ? pathToOwnerType(path) || formatType(_type) // Fallback for unsupported routes
       : "Unknown") as PageOwnerType
+
+    if (!Object.values(OwnerType).includes(ownerType) && SHOW_ANALYTICS_CALLS) {
+      console.warn(
+        `OwnerType "${ownerType}" is not part of @artsy/cohesion's schema.`
+      )
+    }
```
erikdstock, line 77: It is a shame we can't test `PageOwnerType` rather than `OwnerType` - I think this is because [it is a union](https://github.com/artsy/cohesion/blob/main/src/Schema/Values/OwnerType.ts#L172) ('x' | 'y' | 'z') rather than an enum, and enums [act like objects at runtime](https://www.typescriptlang.org/docs/handbook/enums.html) so can interact with normal javascript code.

I wish there were a similar typescript utility to `Pick<>` for these.

TMcMeans: thanks for walking through and helping me understand why this doesn't work btw!

PR #12844, opened 2023-09-07
chore(EMI-1448): wrap up tests for address autocomplete hook

TMcMeans:

This PR solves [EMI-1448]

### Description

The goal of this PR is to wrap up the tests for the autocomplete hook. I have one test that is currently skipped while there is still conversation with design on debouncing/throttling calls to the Smarty API while the user key inputs. 

@artsy/emerald-devs 

[EMI-1448]: https://artsyproduct.atlassian.net/browse/EMI-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
+++ b/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
@@ -15,7 +15,9 @@ let mockFetch: jest.Mock
 
 describe("useAddressAutocomplete", () => {
   const setupHook = (...args: Parameters<typeof useAddressAutocomplete>) => {
-    const result = renderHook(() => useAddressAutocomplete(...args))
+    const result = renderHook(props => useAddressAutocomplete(props), {
+      initialProps: args[0],
+    })
     return result
   }
 
@@ -41,10 +43,15 @@ describe("useAddressAutocomplete", () => {
 
   afterAll(() => {
     jest.resetAllMocks()
+    jest.resetModules()
   })
 
   beforeEach(() => {
     jest.clearAllMocks()
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
   })
 
   describe("when the US address autocomplete feature flag is enabled", () => {
@@ -132,10 +139,7 @@ describe("useAddressAutocomplete", () => {
       })
 
       it("returns the correct single-line text for an address with multiple entries", async () => {
-        jest.clearAllMocks()
-
-        // Override the mock just for this example to include multiple entries??
-        global.fetch = jest.fn().mockResolvedValue({
+        mockFetch.mockResolvedValue({
           json: jest.fn().mockResolvedValue({
             suggestions: [
               {
@@ -174,8 +178,8 @@ describe("useAddressAutocomplete", () => {
         await waitFor(() => result.current.autocompleteOptions.length > 0)
 
         rerender({ country: "UK" })
+
         expect(result.current.autocompleteOptions).toEqual([])
-        // :(
       })
 
       it("resets the suggestions without fetching when the search term is too short", async () => {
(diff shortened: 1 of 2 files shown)


TMcMeans:
For the two tests that are failing, I am currently trying to debug an issue with state leakage coming from one test affecting the others. I believe this has to do with the way the mocks are being set. I'm experimenting with using afterEach() to reset some of these mocks, but its not effective.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -45,10 +45,6 @@ export const useAddressAutocomplete = (
   const isAddressAutocompleteEnabled =
     isAPIKeyPresent && isFeatureFlagEnabled && isUSAddress
 
-  // // TODO: Remove this console log after testing
-  // console.log({ isAPIKeyPresent, isFeatureFlagEnabled, isUSAddress })
-
-  // reset suggestions if the country changes
```
iskounen, line 51: I think that it would be OK to keep this comment about resetting the suggestions.

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
```diff
@@ -108,18 +115,111 @@ describe("useAddressAutocomplete", () => {
...
+
+        act(() => {
+          result.current.fetchForAutocomplete({ search: "40" })
+        })
+
+        expect(result.current.autocompleteOptions).toEqual([])
+        expect(mockFetch).not.toHaveBeenCalled()
+      })
+
+      it.skip("debounces calls to Smarty API", async () => {
```
iskounen, line 205: This test looks good, but it's being skipped so I'm not sure if it passes or not.

TMcMeans: Testing the debouncing is good to do for this hook for sure, but it actually isn't passing unfortunately :( 

Also, I have it as skipped because there was a follow up conversation w/ design about reconsidering how the UI handles the key inputs so this test is TBD. @erikdstock wrote another ticket as a follow up here: https://artsyproduct.atlassian.net/browse/EMI-1459

src/Components/Address/useAddressAutocomplete.ts
TMcMeans: sure, I'll add it back in

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
iskounen: Got it. I was missing out on that context 👍

TMcMeans: I agree, but the existing set up for adding fake jest timers may be useful for updating the tests once someone gets around to it? No opinion either way though!

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
```diff
@@ -41,12 +43,17 @@ describe("useAddressAutocomplete", () => {
 
   afterAll(() => {
     jest.resetAllMocks()
+    jest.resetModules()
```
erikdstock, line 46: What does this get us?

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
erikdstock: Maybe it makes sense to remove rather than skip for now, but then you could immediately begin on the debounce ticket and just paste it right back in?

erikdstock (dismissed):
No blockers from me - I think i agree with adam on removing the skipped test, then adding it back when we add our debounce behavior (can literally start from this PR or reverting the commit to remove the test)

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
TMcMeans: `jest.resetModules()` resets the module registry and clears all cached module imports so that you get a fresh copy of the imported module in each test. It's useful for when a module has internal state that you don't want persisted. I was experimenting with it when I thought there was a state leakage problem happening between the tests or some kind of side effects causing other tests to fail. 

The test still pass without it though, so I might just remove it.

PR #12848, opened 2023-09-08
chore(EMI-1308): run yarn relay-compiler

fladson:

This PR solves [EMI-1308]

### Description

After merging https://github.com/artsy/metaphysics/pull/5250/files I did a QA on staging and didn't see the new exposed field there (see screenshot below), this is adding the missing generated files changes that are needed for it to work.
SCR-20230908-joi

[EMI-1308]: https://artsyproduct.atlassian.net/browse/EMI-1308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -382,6 +382,7 @@ export const AddressVerificationFlowFragmentContainer = createFragmentContainer(
               }
             }
             verificationStatus
+            addressVerificationId
           }
           ... on VerifyAddressFailureType {
             mutationError {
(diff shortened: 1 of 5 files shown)


damassi:
@fladson - note that the generated files are _only_ for typescript type-checking. They have no bearing on actual runtime code, to the point where all generated files could be omitted via gitignore and the app would still function. (I think the issue was that missing field in the graphql query.)

fladson:
> @fladson - note that the generated files are _only_ for typescript type-checking. They have no bearing on actual runtime code, to the point where all generated files could be omitted via gitignore and the app would still function. (I think the issue was that missing field in the graphql query.)

Thanks for the explanation.

PR #12834, opened 2023-09-06
[spike] refactor: EMI-1414 New address form

erikdstock:
Temporary PR to finish up a spike ticket from some time ago, no review required yet.

This PR creates a general address form that can be used as-is or composed using a context provider (essentially a Formik wrapper) and any custom inputs. In addition the wrapper can accept custom form validation and `onSubmit` logic. It is part of [EMI-1414] and may also enable [EMI-1418].

Included in there is some early play in the DebugApp context convincing myself it could work for different kinds of address forms, followed by a spike on dropping it into the shipping route. The shipping route currently uses a lot of formik-like useState hooks, `[addressTouched, setAddressTouched]` that sort of thing

[EMI-1414]: https://artsyproduct.atlassian.net/browse/EMI-1414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1418]: https://artsyproduct.atlassian.net/browse/EMI-1418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12852, opened 2023-09-08
feat: generic address form context and helpers

erikdstock:
This is split off #12834 and completes [EMI-1414] by creating a new address form context that can be used to place a similar address form (or just base address form logic) in multiple areas of the app.

- AddressFormWrapper is a formik context that expects address values to be present, but can be extended with a wider type
- Supports custom validation logic
- Bring your own fields or import and provide container for a default set of them

Added tests demonstrating usage, but there seem to be some quirks working with formik and @testing-library/react. Need to revisit this.

See the bigger discussion below on whether this is really necessary. Off the top of my head, enhancements that move the needle of justification will be:
- Prop to limit available countries ([current implementation leaks shipping dependencies](https://github.com/artsy/force/blob/066271522363386edfebcd585c253a1fbdc0a297/src/Components/Address/AddressForm.tsx#L77-L80))
- Add props to enable autocomplete and [maybe] address verification

TODO:
- [ ] Finish tests
- [ ] Demonstrate implementation in a PR to this PR

[EMI-1414]: https://artsyproduct.atlassian.net/browse/EMI-1414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+export const DEFAULT_VALIDATION_SCHEMA = Yup.object().shape({
+  name: Yup.string().required("Name is required"),
+  country: Yup.string().required("Country is required"),
+  addressLine1: Yup.string().required("Address is required"),
+  city: Yup.string().required("City is required"),
+  region: Yup.string().required("Region is required"),
+  postalCode: Yup.string().required("Postal Code is required"),
+})
+
+// For debugging
```
damassi, line 41: Stuff that is secondary to the primary concern of the component should be moved to the bottom of the file or to another file

src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  initialValues: Partial<T>
+}
+
+/**
+ * A Formik context that provides a default validation schema and initial values
+ * for an address form. Provide a custom type extending `AddressFormValues` to
+ * add additional fields, along with other Formik props. The children should
+ * formik-compatible contain inputs for any AddressFormValues and other fields.
+ */
+export const AddressFormWrapper = <
```
damassi, line 79: Can this just be named `AddressForm`? That its a wrapper is more of an implementation detail

src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  ...props
+}: AddressFormWrapperProps<T>): JSX.Element => {
+  let validationRule: Pick<FormikConfig<T>, "validate" | "validationSchema">
+  if (validationSchema) {
+    validationRule = { validationSchema }
+  } else if (validate) {
+    validationRule = { validate }
+  } else {
+    validationRule = { validationSchema: DEFAULT_VALIDATION_SCHEMA }
+  }
```
damassi, line 95: Whats the difference between `validationSchema` and `validate` and the default case? 

Couldn't we just do something with standard default props, like 
```tsx
const AddressForm = ({ validationSchema = DEFAULT_VALIDATION_SCHEMA, ... }) => { ... } 
```

damassi (commented):
This is starting to make a bit more sense, but still having a little trouble understanding why the form needs to be abstracted like this since all of the functionality is basically a default formik wrapper that receives some address props. Do we have situations where the address form needs to change from one place to the next? 

Ideally, the simplest expression of this would simply be 

```
<AddressForm />
```
It doesn't need to take any props, or any other kinds of validation since the needs around validating an address are consistent across the board. Internal, it would look like:

```
const AddressForm = ({ onChange, onSubmit }) => {
  return (
    <Formik validationSchema={addressValidationSchema} onChange={onChange} onSubmit={onSubmit}>
      <AddressInputs /> 
    </Formik>
  )
}
```
That kind of thing. And AddressInputs doesn't even need to be exported; its just referenced internally. Basically, just removing everything that isn't strictly necessary. If there's no need to configure it, just keep things vanilla and fixed and as simple as possible. 

(Of course, i could be missing the wider applications of things, but this is mostly to fight against premature abstraction lest its specifically needed.)

src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+ */
+export const AddressFormWrapper = <
+  T extends AddressFormValues = AddressFormValues
+>({
+  initialValues = {} as Partial<T>,
+  children,
+  validate,
+  validationSchema,
+  ...props
+}: AddressFormWrapperProps<T>): JSX.Element => {
```
damassi, line 87: Do we _need_ generics here?

src/Components/Address/__tests__/NewAddressForm.jest.tsx
```diff
@@ -0,0 +1,169 @@
...
+    <AddressFormWrapper<AddressFormValues>
+      initialValues={EMPTY_ADDRESS}
+      onSubmit={mockOnSubmit}
+      {...props}
+    >
+      <Form>
+        <AddressInputs />
+        <Button type="submit">Submit</Button>
+      </Form>
+    </AddressFormWrapper>
```
damassi, line 26: Using this as an example, its hard to see why this is better than just doing 
```jsx
<Formik ...>
  <Form>
    <AddressInputs>
  </Form>
</Formik>
```

Just because, all that `AddressFormWrapper` does is contain some default props, but it doesn't contain `<Form>`, it doesn't contain the form inputs, it doesn't contain submit buttons or handlers or anything like that. So all it appears to be is an intermediary-layer abstraction over a single other component (`<Formik />`) that basically does the exact same thing. 

If `AddressFormWrapper` contained more _behavior_ then I could understand its utility, but without that I would rather just have a file called `AddressFormInputs` and `addressFormValidation` and use those where need be. _OR_, if indeed the address form is a consistent, stable single thing, merge all of those together into one component as I mentioned elsewhere in this PR.

src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const {
+    values,
+    errors,
+    touched,
+    status,
+    handleChange,
+    handleBlur,
+  } = useFormikContext<AddressFormValues>()
+  return (
+    <GridColumns data-testid="address-inputs">
```
damassi, line 146: #non-blocking, but following the principle of 1:1, its best to keep this stuff all consistent. If this component is called `AddressInputs`, it follows that the `data-testid` should also equal `AddressInputs`. Then when searching for one thing, you find the other, and so forth. `address-inputs` is 1:2.

src/Components/Address/NewAddressForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+      <AddressInputs
+        endContent={
+          <Button type="submit" variant="primaryBlack" width="50%">
+            Save and Continue
+          </Button>
+        }
+      />
+    </Form>
+  )
+}
```
damassi, line 66: Is the debug component necessary? I don't see us wiring up helper components like this often. And a dev can just console.log the formik state in a one-liner from the context or whatever.

erikdstock: yeah, this can go away.

damassi:
Thanks for the overview, really helpful 👍 I'll ping you in slack in a bit we can go over a few different angles just so I fully understand. This isn't blocking any work, right?

damassi:
@dzucconi - if you have a sec, curious about your thoughts here

erikdstock:
Looking over https://github.com/artsy/force/pull/9248 to see if that is a better base for this work. I think the key difference to extract would be the layouts and the nesting of the address inside formik `values.attributes`.

damassi:
Yeah @erikdstock - looking over the form in that PR, that's exactly what i'm talking about in my comments -- there's nothing abstracted there. Its just a form that takes props and inside is a normal formik component doing totally normal things. Even just duplicating the shipping address form and using it for your needs would be totally fine. Duplication is ok if the code is simple. The point is that its easy for others (than yourself) to understand years down the line.

erikdstock:
Closing for now, the POC worked but it makes more sense to tackle things in a different order. Thanks for the reviews and convo @damassi!

PR #12853, opened 2023-09-11
fix(EMI-1344): Save Offer note and price in offer flow

MrSltun:

This PR solves [EMI-1344]

### Description
This PR updates the offer page to pre-fill the form with the user's custom note and offer if available

### Video

| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/20655703/1a441ae7-a40d-4383-8e9b-3a140db7d2d3" />	 | <video src="https://github.com/artsy/force/assets/20655703/bb7e6639-40a0-4414-9586-372903e7829f" />	 |	

cc @artsy/emerald-devs

### DIFF
--- a/src/Apps/Order/Routes/Offer/index.tsx
+++ b/src/Apps/Order/Routes/Offer/index.tsx
@@ -1,4 +1,4 @@
-import { FC, useState } from "react"
+import { FC, useEffect, useState } from "react"
 import * as DeprecatedAnalyticsSchema from "@artsy/cohesion/dist/DeprecatedSchema"
 import { Button, Flex, Message, Spacer, Text } from "@artsy/palette"
 import { Offer_order$data } from "__generated__/Offer_order.graphql"
@@ -53,6 +53,19 @@ export const OfferRoute: FC<OfferRouteProps> = ({
     value: "",
   })
   const [offerValue, setOfferValue] = useState(0)
+
+  useEffect(() => {
+    if (order.myLastOffer?.note) {
+      if (!order.myLastOffer.note.startsWith("I sent an offer for")) {
+        setOfferNoteValue({
+          exceedsCharacterLimit: false,
+          value: order.myLastOffer.note,
+        })
+      }
+    }
+    // need this to run only once
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
 
   const onOfferInputFocus = () => {
     trackEvent({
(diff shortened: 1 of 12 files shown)


src/Apps/Order/Routes/Offer/index.tsx
```diff
@@ -54,6 +54,19 @@ export const OfferRoute: FC<OfferRouteProps> = ({
...
+  useEffect(() => {
+    if (order.myLastOffer?.note) {
+      if (!order.myLastOffer.note.startsWith("I sent an offer for")) {
+        setOfferNoteValue({
+          exceedsCharacterLimit: false,
+          value: order.myLastOffer.note,
+        })
+      }
+    }
+    // need this to run only once
```
iskounen, line 66: Since this value only needs to be set once to initialize the state variable, can we put this logic into a function called `initialNoteValue(note)` and use it when we call `useState` instead?

```js
const [offerNoteValue, setOfferNoteValue] = useState({
  exceedsCharacterLimit: false,
  value: initialNoteValue(order.myLastOffer?.note),
})
```

iskounen: Something similar is happening in a `useEffect` hook in the PriceOptions component, and I think that logic could also be simplified by performing that initialization in the `useState` call rather than an effect.

src/Apps/Order/Utils/offerUtils.ts
```diff
@@ -0,0 +1,127 @@
...
+      value: minPriceRange,
+      description: "Low-end of range (lower chance of acceptance)",
+    },
+  ]
+
+  return getRangeDetails.map((rangePrice, idx) => ({
+    key: `price-option-${idx}`,
+    value: rangePrice.value!,
+    description: rangePrice.description,
+  }))
```
iskounen, line 100: Since there are only 3 options (and we don't foresee adding more), this method would be easier to read if it just returned lines 82-93 with the `key` field added to each and the `value` field populated inline.

I also think that the `key` for each option can be more meaningful (ie. price-option-min instead of price-option-2).

iskounen (dismissed):
LGTM 🎉 

Thanks for keeping those `useEffect` hooks simple. We've been working in the Shipping step lately, and the amount of logic that is happening in `useEffect` hooks has made it difficult to work with, so I appreciate you effort to keep those slim.

- I have one non-blocking suggestion for readability
- I also wonder why there are so many unrelated generated files in this PR - maybe something to double-check

PR #12855, opened 2023-09-11
fix(EMI-1342): jumping issue on iPhones in Offer page

MrSltun:

This PR solves [EMI-1342]

### Description
This PR fixes the double-jumping that happens in the offer screen in Safari iOS (tested with Safari mobile view and it works fine)

### Videos
| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/20655703/552c60aa-7492-41c4-8e1d-0b17dd56b089" /> | <video src="https://github.com/artsy/force/assets/20655703/3a6a69e8-8edd-4d1a-a416-2f3e4d7c6532" />	 | 

[EMI-1342]: https://artsyproduct.atlassian.net/browse/EMI-1342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/PriceOptions.tsx
+++ b/src/Apps/Order/Components/PriceOptions.tsx
@@ -12,6 +12,7 @@ import { PriceOptions_order$data } from "__generated__/PriceOptions_order.graphq
 import { appendCurrencySymbol } from "Apps/Order/Utils/currencyUtils"
 import { useTracking } from "react-tracking"
 import { Jump, useJump } from "Utils/Hooks/useJump"
+import { Device, useDeviceDetection } from "Utils/Hooks/useDeviceDetection"
 
 export interface PriceOptionsProps {
   onChange: (value: number) => void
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/PriceOptions.tsx
```diff
@@ -185,7 +187,8 @@ export const PriceOptions: React.FC<PriceOptionsProps> = ({
                     onChange={setCustomValue}
                     onFocus={() => {
                       onFocus()
-                      jumpTo("price-option-custom")
+                      if (device !== Device.iPhone)
+                        jumpTo("price-option-custom")
```
MrSltun, line 191: I added the check here instead of inside `useJump` since I assume it works fine in other places (haven't checked)

Should we add it to the [useJump](https://github.com/artsy/force/blob/main/src/Utils/Hooks/useJump.tsx) hook instead?

TMcMeans (commented):
**question [non-blocking]:** 

I had one small question about whether the fix should apply to other iOS devices like iPads or is the issue exclusive to iPhones? Otherwise, LGTM :) 

MrSltun:
@TMcMeans Good question! I actually tested it on the iPod and iPad simulators and they work fine, iPad doesn't need a change, but iPod does, and it's covered with the implementation in this PR :)

src/Apps/Order/Components/PriceOptions.tsx
dzucconi: No need to add this to the hook

PR #12867, opened 2023-09-13
Refactor shipping tests

starsirius:

This PR solves [EMI-1413](https://artsyproduct.atlassian.net/browse/EMI-1413)

### Description

In preparation of bigger refactor and/or new features of the shipping step, I want to invest in tests to prevent us from future regressions while iterating on the implementation. This PR refactors the current tests and does mainly 2 things:

- Re-organize the tests and split them by the 3 fulfillment methods. See [the first commit](00d6d3beb37577e42e9c1e010b29422011f0e306) for the skeleton of the new structure.
- Convert the tests from Enzyme to Testing Library. I hope the philosophy of TL will
  - encourage us to write better implementation (such as accessibility), and
  - focus on how users interact with the products (rather than implementation details), so that the tests can remain relevant even when we swap out the implementation.

The diff is huge and it's better the review commit by commit. The process is basically

- Duplicate the test to a TL one with a better structure
- Port Enzyme tests to TL in the new file one by one
- Then promote the TL file when tests are all migrated over

I'll comment inline for things I learned along the way, and please let me know if I can make reviewing this PR easier. 💪 

cc @artsy/emerald-devs 

[EMI-1413]: https://artsyproduct.atlassian.net/browse/EMI-1413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -91,53 +88,25 @@ jest.mock("relay-runtime", () => ({
...
-const ArtsyShippingDomesticFromGermanyOrder: ShippingTestQuery$rawResponse["order"] = {
-  ...UntouchedBuyOrderWithArtsyShippingDomesticFromGermany,
-  __typename: "CommerceBuyOrder",
-  mode: "BUY",
-  internalID: "1234",
-  id: "1234",
-}
-
-const ArtsyShippingInternationalFromGermanyOrder: ShippingTestQuery$rawResponse["order"] = {
-  ...UntouchedBuyOrderWithArtsyShippingInternationalFromGermany,
```
starsirius, line 126: These customizations seem unnecessary, so I just used the untouched ones to simplify fixture setup.

src/Apps/Order/Routes/__tests__/Shipping.TL.jest.tsx
```diff
@@ -217,35 +217,32 @@ describe("Shipping", () => {
...
-                  ...validAddress,
-                  phoneNumber: "",
-                },
-              },
-            },
-          })
+
+        let mutationArg = mockCommitMutation.mock.calls[0][0]
+        expect(mutationArg.mutation.default.operation.name).toEqual(
+          "SetShippingMutation"
```
starsirius: Previously we only check the mutation variables, which can be confusing sometimes. The new pattern here uses the [`mock` property](https://jestjs.io/docs/mock-functions#mock-property), so we can more explicitly check both the mutation name and the variables.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -254,270 +219,220 @@ describe("Shipping", () => {
...
-              id: "1234",
-              fulfillmentType: "SHIP",
-              phoneNumber: address.phoneNumber,
-              shipping: {
-                ...address,
-                phoneNumber: "",
-              },
+        expect(mockCommitMutation).toHaveBeenCalledTimes(2)
+
+        let mutationArg = mockCommitMutation.mock.calls[0][0]
```
starsirius, line 315: Previously we only checked the mutation variables, which can be confusing sometimes. The new pattern here uses the [`mock` property](https://jestjs.io/docs/mock-functions#mock-property), so we can more explicitly check both the mutation name and the variables.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -152,68 +121,64 @@ const emptyTestMe: ShippingTestQuery$rawResponse["me"] = {
...
-    await this.update()
+      ],
+      pageInfo,
+    },
   }
+)
+
+const saveAndContinue = async () => {
+  await userEvent.click(screen.getByText("Save and Continue"))
+  await flushPromiseQueue()
```
starsirius, line 175: Ideally when writing tests I shouldn't need to know if/where I should wait, which is more like implementation details. Wrapping it in a helper, like this, helps encapsulate things but I still need to wait explicitly here and there. Perhaps it's inevitable in unit tests. An alternative would be like Cypress that the tests internally _waits_ for some UI updates.

erikdstock: can you use something like `await waitFor(() => {expect(screen.getByText("My success message").toBeInTheDocument())})`? This lets you define an expectation as part of your action, which could be good if the action has a consistent result.

erikdstock: Another option since this is just a click helper could be to wait for the loading indicator to go away, since that should always happen.

src/Components/__tests__/Utils/addressForm.ts
```diff
@@ -58,3 +60,25 @@ export const fillAddressForm = (component: any, address: Address) => {
     value: address.phoneNumber!,
   })
 }
+
+export const fillAddressFormTL = (address: Address) => {
+  const name = screen.getByPlaceholderText("Full name")
+  const country = screen.getByRole("combobox")
```
starsirius, line 66: These non-ideal queries suggest implementation improvements!

src/Components/__tests__/Utils/addressForm.ts
```diff
@@ -58,3 +60,25 @@ export const fillAddressForm = (component: any, address: Address) => {
...
+  const addressLine1 = screen.getByPlaceholderText("Street address")
+  const addressLine2 = screen.getByPlaceholderText("Apt, floor, suite, etc.")
+  const city = screen.getByPlaceholderText("City")
+  const region = screen.getByPlaceholderText("State, province, or region")
+  const postalCode = screen.getByPlaceholderText("ZIP/postal code")
+  const phoneNumber = screen.getAllByPlaceholderText(
+    "Add phone number including country code"
+  )[0]
+
+  userEvent.paste(name, address.name)
```
starsirius, line 76: TL can be slow, and it's a tradeoff we will decide. Here is an example I changed to `paste` from `type` (which is significantly slower). In this case we don't really care about each key stroke so I think it's acceptable. If we do need to consider each key, such as autocomplete use cases, we might need to accept the slowness.

See [a5782ec](https://github.com/artsy/force/pull/12867/commits/a5782ec99010954c22cfc6d2b5abeff0632e601e) for more details.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
starsirius: Yeah, but my question was to not worry about needing to wait or not for each action. I see the [`findBy`](https://testing-library.com/docs/queries/about#types-of-queries) queries perform retries and that's more similar to Cypress, and potentially can be an alternative.

PR #12873, opened 2023-09-13
fix(EMI-1459) use throttle instead of debounce for address autocomplete

TMcMeans:

This PR solves [EMI-1459]

### Description

The goal of this PR is to change the usage of debounce to implementing Lodash's throttle function in the `useAddressAutocomplete` hook to stagger multiple query requests to the Smarty API on user key press.

Note: Thoroughly testing the throttle's behavior was challenging. Although Lodash's throttle can use `setTimeout`, it also employs timestamps for finer control over function execution. This makes it less straightforward to mock using Jest's timer mocks, which are primarily designed to work with `setInterval()` and `setTimeout()`.

@artsy/emerald-devs 

Sep-15-2023 12-21-21

[EMI-1459]: https://artsyproduct.atlassian.net/browse/EMI-1459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
+++ b/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
@@ -6,10 +6,13 @@ import {
 import { useFeatureFlag } from "System/useFeatureFlag"
 import { getENV } from "Utils/getENV"
 import { waitFor } from "@testing-library/react"
+import { throttle } from "lodash"
 
 jest.mock("react-tracking")
 jest.mock("System/useFeatureFlag")
 jest.mock("Utils/getENV")
+
+jest.mock("lodash/throttle", () => jest.fn(fn => fn))
 
 let mockFetch: jest.Mock
 
@@ -195,6 +198,20 @@ describe("useAddressAutocomplete", () => {
 
         expect(result.current.autocompleteOptions).toEqual([])
         expect(mockFetch).not.toHaveBeenCalled()
+      })
+
+      it("throttles calls to the Smarty API", async () => {
+        const { result } = setupHook({ country: "US" })
+
+        act(() => {
+          result.current.fetchForAutocomplete({ search: "401 Broadway" })
+        })
+
+        expect(throttle).toHaveBeenCalledTimes(1)
+        expect(throttle).toHaveBeenCalledWith(expect.any(Function), 500, {
+          leading: true,
+          trailing: true,
+        })
       })
     })
 
--- a/src/Components/Address/useAddressAutocomplete.ts
+++ b/src/Components/Address/useAddressAutocomplete.ts
@@ -1,9 +1,11 @@
 import { AutocompleteInputOptionType } from "@artsy/palette"
 import { Address } from "Components/Address/AddressForm"
 import { useFeatureFlag } from "System/useFeatureFlag"
-import { useDebounce } from "Utils/Hooks/useDebounce"
 import { getENV } from "Utils/getENV"
 import { useCallback, useEffect, useState } from "react"
+import { throttle } from "lodash"
+
+const THROTTLE_DELAY = 500
 
 export type ProviderSuggestion = {
   city: string
@@ -75,6 +77,15 @@ export const useAddressAutocomplete = (
     [apiKey]
   )
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const throttledFetchSuggestions = useCallback(
+    throttle(fetchSuggestions, THROTTLE_DELAY, {
+      leading: true,
+      trailing: true,
+    }),
+    []
+  )
+
   const fetchForAutocomplete = useCallback(
     // these are the parameters to the Smarty API call
     async ({ search, selected }: { search: string; selected?: string }) => {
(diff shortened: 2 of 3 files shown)


src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
```diff
@@ -196,6 +199,16 @@ describe("useAddressAutocomplete", () => {
...
+
+      it("throttles calls to the Smarty API", async () => {
+        const { result } = setupHook({ country: "US" })
+
+        act(() => {
+          result.current.fetchForAutocomplete({ search: "401 Broadway" })
+        })
+
+        expect(throttle).toHaveBeenCalled()
+      })
```
erikdstock, line 211: In order to really test this we need to ask whether it was called with the private `fetchSuggestions` function defined inside the hook and used by `fetchForAutocomplete` and `fetchSecondarySuggestions`, ideally also that it was only called once (we aren't creating multiple throttles which would make it not work). I was the one who recommended this test, but in short I don't know if it's really possible to test if we are using `_.throttle`.

erikdstock: We could assert something like 
```ts
expect(throttle.mock.calls.length).toEqual(1)
expect(throttle.mock.calls[0]).toEqual([expect.anything(), 500, {leading: true, trailing: true}])
```

But I think deleting this could also be fine.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -75,6 +75,15 @@ export const useAddressAutocomplete = (
     [apiKey]
   )
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const throttledFetchSuggestions = useCallback(
+    throttle(fetchSuggestions, 2000, {
```
erikdstock, line 80: nitpick: throttle delay should be 500

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -75,6 +75,15 @@ export const useAddressAutocomplete = (
...
+      leading: true,
+      trailing: true,
+    }),
+    [fetchSuggestions, throttle]
```
erikdstock, line 84: non-blocking: this should technically be fine, but `[]` might be even safer. We only want it to execute once, and you have to look at this code to convince yourself it will do that.

src/Components/Address/useAddressAutocomplete.ts
erikdstock: non-blocking: Would it make sense to rename `fetchSuggestions` `_fetchSuggestions` to make clear it is private and should never be called? Or we could not even declare it, just pass it directly to throttle. def not blocking.

erikdstock (commented):
A few minor comments but I think this is just about ready! For reference on our troubles testing the throttling behavior using jest's fake timers, this is my current understanding:
- `_.throttle()` [passes its 2nd arg (wait, the delay) to debounce](https://github.com/lodash/lodash/blob/master/throttle.js#L52-L67)
- `_.debounce()` [uses `requestAnimationFrame()` instead of `setTimeout()`](https://github.com/lodash/lodash/blob/master/debounce.js#L102-L108) unless you [explicitly pass `wait=0`](https://github.com/lodash/lodash/blob/master/debounce.js#L79). If we did that it wouldn't be much of a throttle.
- [jest doesn't mess with raf](https://github.com/jestjs/jest/issues/5147#issuecomment-353316538), which [works outside of setTimeout for performance reasons](https://stackoverflow.com/questions/38709923/why-is-requestanimationframe-better-than-setinterval-or-settimeout), and setTimeout/setInterval are jest's fake timers' main concerns.

src/Components/Address/useAddressAutocomplete.ts
TMcMeans: a throttle delay of 500 would mean half a second delay between requests, is that okay?

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
TMcMeans: I'll include these assertions and see how things go but will delete as a backup if I can't get things to pass. Overall, I think its okay to keep this test because we aren't testing lodash but just that its being called.

src/Components/Address/useAddressAutocomplete.ts
erikdstock: yep, i think we may want to make it smaller in time - Could you make it a constant at the top of this file?

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
TMcMeans: Based on how the test is written I think it should be ` expect(throttle).toHaveBeenCalledWith([
    expect.any(Function), // <-- Change here
    500,
    { leading: true, trailing: true },
  ])`

because it actually returns an anonymous function instead of anything. Also I believe Jest's `expect.anything()` matcher doesn't match against a function type but only against other non-`null` and non-`undefined` values. Jest has other matchers like `expect.any(Function)` to match against expecting a function as the first arg.

src/Components/Address/useAddressAutocomplete.ts
TMcMeans: I'm not sure what you mean here, can you provide more context?

TMcMeans: are you saying to use an empty array instead of deps?

erikdstock (approved):

PR #12878, opened 2023-09-14
feat: add tracking events to errors in new payment route

iskounen:
This PR adds event tracking to the "new payment" route which users visit when fixing a failed payment for an order.

The goal of this work is to better understand what dead-ends users encounter when re-submitting orders after fixing failed payments. We currently know that it is possible for this to fail if the seller's stripe account is deactivated before the buyer gets a chance to fix a failed payment. There may be other cases we are missing.

### DIFF
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -23,8 +23,10 @@ import {
 import { get } from "Utils/get"
 import { BuyerGuarantee } from "Apps/Order/Components/BuyerGuarantee"
 import { createStripeWrapper } from "Utils/createStripeWrapper"
-import { ContextModule, OwnerType } from "@artsy/cohesion"
+import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
 import { OrderRouteContainer } from "Apps/Order/Components/OrderRouteContainer"
+import { useTracking } from "react-tracking"
+import { getErrorDialogCopy } from "Apps/Order/Utils/getErrorDialogCopy"
 
 export const ContinueButton = props => (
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -206,6 +226,12 @@ export const NewPaymentRoute: FC<NewPaymentProps & StripeProps> = props => {
   const handleFixFailedPaymentError = async (code: string) => {
     switch (code) {
       case "capture_failed": {
+        trackErrorMessageEvent(
+          "Charge failed",
+          "Payment has been declined. Please contact your card provider or bank institution, then press “Continue” again. Alternatively, use another payment method.",
+          code
+        )
```
erikdstock, line 233: Ah, i was thinking of a big switch error handler like this recently. I'll bet we could make this ever more general than just the new payment step.

erikdstock: Question: This message string is duplicated directly below in `dialog.showErrorDialog()`. If we want to track errors we are showing to users could a single function do both?

erikdstock (commented):
Looks good with one question about duplication, no blockers

src/Apps/Order/Routes/NewPayment/index.tsx
iskounen: Yeah, it may be time to start consolidating a lot of this tracking + dialog code. We rushed the initial batch of tracking events so that we could start seeing data about errors in the checkout flow, but this code is starting to become cumbersome.

@rquartararo and I tried to refactor the dialog to do its own data tracking early on in this project, and got blocked trying to convert the dialog into a functional component (to use the tracking hook). Again, our goal was to see data ASAP so we didn't pursue it, but now would be a good time to revisit that work.

PR #12898, opened 2023-09-18
Fix: EMI-1457 - show autocomplete input by default so hidden element styles are applied properly

erikdstock:
The type of this PR is: fix

This PR resolves [EMI-1457] by adding a brief loading state to the autocomplete input. While loading (and server-side) the address form renders the input, but disabled, then swaps it out if autocomplete is not available. This has the effect of making the 🔎  disappear, a small papercut and possible followup in palette.

[EMI-1457]: https://artsyproduct.atlassian.net/browse/EMI-1457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
@@ -47,6 +47,12 @@ jest.mock("Utils/Hooks/useMatchMedia", () => ({
   __internal__useMatchMedia: () => ({}),
 }))
 
+// TODO: Turning this on causes 1 test to fail expectedly and another to also fail. What is going on here?
+const TRIGGER_LEAKY_FAILURES = false
+const triggerLeakyFailure = () => {
+  expect(TRIGGER_LEAKY_FAILURES).toBe(false)
+}
+
 jest.mock("@artsy/palette", () => {
   return {
     ...jest.requireActual("@artsy/palette"),
@@ -277,8 +283,8 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        expect(screen.getByRole("combobox")).toHaveValue("US")
-        expect(screen.getByRole("combobox")).toBeDisabled()
+        expect(screen.getByTestId("AddressForm_country")).toHaveValue("US")
+        expect(screen.getByTestId("AddressForm_country")).toBeDisabled()
       })
 
       it("sets and enables country select if artwork only ships domestically and is in the EU", async () => {
@@ -296,8 +302,8 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        expect(screen.getByRole("combobox")).toHaveValue("IT")
-        expect(screen.getByRole("combobox")).toBeEnabled()
+        expect(screen.getByTestId("AddressForm_country")).toHaveValue("IT")
+        expect(screen.getByTestId("AddressForm_country")).toBeEnabled()
       })
 
       it("sets shipping on order and saves address on user", async () => {
@@ -307,7 +313,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockCommitMutation).toHaveBeenCalledTimes(2)
@@ -346,7 +352,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         userEvent.click(
           screen.getByRole("checkbox", { name: /Save shipping address/ })
         )
@@ -378,7 +384,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockCommitMutation).toHaveBeenCalledTimes(2)
@@ -403,7 +409,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockShowErrorDialog).toHaveBeenCalledWith()
@@ -416,7 +422,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockShowErrorDialog).toHaveBeenCalledWith()
@@ -431,7 +437,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockShowErrorDialog).toHaveBeenCalledWith({
@@ -450,7 +456,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockShowErrorDialog).toHaveBeenCalledWith({
@@ -469,7 +475,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        fillAddressForm(validAddress)
+        await fillAddressForm(validAddress)
         await saveAndContinue()
 
         expect(mockShowErrorDialog).toHaveBeenCalledWith({
@@ -573,7 +579,7 @@ describe("Shipping", () => {
             Me: () => meWithoutAddress,
           })
 
-          fillAddressForm({
+          await fillAddressForm({
             name: "Erik David",
             addressLine1: "401 Broadway",
             addressLine2: "",
@@ -596,7 +602,7 @@ describe("Shipping", () => {
             Me: () => meWithoutAddress,
           })
 
-          fillAddressForm({
+          await fillAddressForm({
             name: "Erik David",
             addressLine1: "401 Broadway",
             addressLine2: "",
@@ -672,7 +678,7 @@ describe("Shipping", () => {
               screen.queryByText("This field is required")
             ).not.toBeInTheDocument()
 
-            fillAddressForm(validAddress)
+            await fillAddressForm(validAddress)
             userEvent.clear(screen.getByPlaceholderText("Street address"))
 
             await userEvent.click(screen.getByText("Save and Continue"))
@@ -718,7 +724,7 @@ describe("Shipping", () => {
               relayEnv
             )
 
-            fillAddressForm(validAddress)
+            await fillAddressForm(validAddress)
             await userEvent.click(screen.getByText("Save and Continue"))
 
             const mutation = env.mock.getMostRecentOperation()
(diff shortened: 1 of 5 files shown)


src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -23,13 +23,17 @@ export interface AddressAutocompleteSuggestion
   entries: number
 }
 
+interface ServiceAvailability {
+  loaded: boolean
+  enabled?: boolean
+}
```
erikdstock, line 29: I made loaded/enabled two separate properties to not get loose with the meaning of 'enabled'.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -54,37 +72,35 @@ export const useAddressAutocomplete = (
     }
   }, [isUSAddress, result.length])
 
-  const fetchSuggestions = useCallback(
-    async ({ search, selected }: { search: string; selected?: string }) => {
-      const params = {
-        key: apiKey,
-        search: search,
+  const fetchSuggestions = useMemo(() => {
+    const throttledFetch = throttle(
```
erikdstock, line 76: switched from `useCallback` to `useMemo()` and dropped the intermediate function declaration in favor of building it inside the body to `useMemo`. Chose `useMemo` because of an error from the linter saying that i should pass an inline function, and not recognizing that `throttle(() => function i had moved in from the external fetchSuggestions...` was pretty much inline. In practice i think this should work the same, but that's the reason.

erikdstock: With useCallback however, tests did start failing (even when i used an empty dependency array). The point of this test was to assert `throttle()` is only called once so we don't start creating duplicate throttles, which would break the functionality:

image

dzucconi: Yep using memo for both debounce and throttle is preferrable.

src/Components/Address/AddressForm.tsx
```diff
@@ -193,9 +193,11 @@ export const AddressForm: React.FC<AddressFormProps> = ({
         )}
       </Column>
       <Column span={12}>
-        {isAddressAutocompleteEnabled ? (
+        {/* Render the autocomplete input optimistically on the server to make sure we send applicable styles to the client */}
```
dzucconi, line 196: I don't think this comment makes sense. The reason the styles weren't working was because of an SSR mismatch.

src/Components/Address/useAddressAutocomplete.ts
dzucconi: `enabled` + `ready`? I suppose there's not really any "loading" happening.

dzucconi (approved):
LGTM. Not sure why passing `disabled` would cause the icon to disappear though? We can investigate.

erikdstock:
> LGTM. Not sure why passing `disabled` would cause the icon to disappear though? We can investigate.

Sorry @dzucconi the icon disappears if, after loading, autocomplete is not available (today, for 99% of users, after launch for non-US orders)

dzucconi:
Ah, I see.

src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -54,37 +72,35 @@ export const useAddressAutocomplete = (
     }
   }, [isUSAddress, result.length])
 
-  const fetchSuggestions = useCallback(
-    async ({ search, selected }: { search: string; selected?: string }) => {
-      const params = {
-        key: apiKey,
-        search: search,
+  const fetchSuggestions = useMemo(() => {
```
TMcMeans, line 75: (non-blocking question): I'm trying to wrap my head around the difference between switching from `useCallback` to `useMemo`.

`useCallback` returns a memoized version of the function, while `useMemo` returns a memoized result, right? Why is it preferable to optimize  the result of the throttled function, and not the instance itself? I'm still trying to wrap my head around this. Or is this switch really only to prevent linter errors?

TMcMeans: Might be answering my own question in realizing that `useMemo()` does memoize a new instance of throttle because that's what is returned as the result.

TMcMeans (dismissed):
Since the functionality of the throttle remains the same because `useMemo` returns a memoized version of the throttle function and it also prevents linter errors and the test are still passing it LGTM 👍🏾 🥇

TMcMeans:
One last question (non-blocking just looking for context): just making sure I understand that by adding a loading state to the input we can help ensure that the initial render from the server and the subsequent client-side renders are consistent? And that allows for more reliable CSS styling?

src/Components/Address/useAddressAutocomplete.ts
dzucconi: Kind of other way around: `useCallback` returns a memoized version of the function but it will be calling `debounce/throttle` every single time it's called, so you'd get a new copy every time you call it. The calls would then be simply delayed by the specified time rather than throttled or debounced. `useMemo` will memoize it as a _value_ a single time.

dzucconi:
It manifests as something different every time. It's not specific to the `VisuallyHidden` component it's just that some of the classes were just gone.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -47,6 +47,12 @@ jest.mock("Utils/Hooks/useMatchMedia", () => ({
   __internal__useMatchMedia: () => ({}),
 }))
 
+// TODO: Turning this on causes 1 test to fail expectedly and another to also fail. What is going on here?
+const TRIGGER_LEAKY_FAILURES = false
+const triggerLeakyFailure = () => {
+  expect(TRIGGER_LEAKY_FAILURES).toBe(false)
+}
+
```
erikdstock, line 55: I tried to identify tests causing additional failures but unfortunately this was the only one i could find.

erikdstock (commented):
Well, the test failures on this were HECK to debug, and I don't think i fully did it, but the tests pass now.

There were 24 failures at the previous commit d28f619072:
<details><summary>Test failure picture</summary>

image

</details> 

<details><summary>Test failure text output</summary>

```
FAIL  src/Apps/Order/Routes/__tests__/Shipping.jest.tsx (77.76 s)
  Shipping
    with partner shipping
      with no saved address
        ✓ shows an active offer stepper if it's an offer order (693 ms)
        ✓ renders fulfillment selection if artwork is available for pickup (289 ms)
        ✓ does not render fulfillment selection if artwork is not available for pickup (119 ms)
        ✕ sets and disables country select if artwork only ships domestically and is not in the EU (165 ms)
        ✕ sets and enables country select if artwork only ships domestically and is in the EU (112 ms)
        ✕ sets shipping on order and saves address on user (144 ms)
        ✕ sets shipping on order but does not save address if save address is not checked (144 ms)
        ✕ routes to payment screen after mutation completes (142 ms)
        ✓ shows the button spinner while loading the mutation (287 ms)
        ✕ shows an error when there is an error from the server (160 ms)
        ✕ shows an error when there is a network error (152 ms)
        ✕ shows an error when there is a missing_country error from the server (152 ms)
        ✕ shows an error when there is a missing_region error from the server (151 ms)
        ✕ shows an error when there is a destination_could_not_be_geocodederror from the server (154 ms)
        ✓ pre-populates address form for order with already persisted shipping info (63 ms)
        ✕ resets shipping for order with already persisted shipping info (131 ms)
        form validations
          ✓ does not submit an empty form (205 ms)
          ✓ does not submit an incomplete form (396 ms)
          ✓ requires some fields (118 ms)
          ✕ requires a phone number (171 ms)
          ✕ allows a missing postal code and state/province if the selected country is not US or Canada (171 ms)
          ✓ only shows validation erros on touched inputs before submission (441 ms)
          ✓ shows all validation erros including untouched inputs after submission (454 ms)
        address verification
          with US enabled and international disabled
            ✕ triggers basic form validation before address verification (186 ms)
            ✕ triggers the flow for US address after clicking continue (183 ms)
            ✕ does not triggers the flow for international address after clicking continue (186 ms)
          with US disabled and international enabled
            ✕ does not triggers tthe flow for US address after clicking continue (188 ms)
            ✕ triggers the flow for international address after clicking continue (191 ms)
      with saved addresses
        ✓ does not show the new address form (134 ms)
        ✓ lists the addresses and renders the add address option (1354 ms)
        ✓ sets shipping with the first saved address and phone number when user submits the form directly (121 ms)
        ✓ sets shipping with the selected saved address and phone number (619 ms)
        address verification
          with address verification enabled
            ✓ does not trigger the flow (125 ms)
        editing address
          ✓ opens a modal with the address prepopulated (606 ms)
          ✓ updates the address after submitting the modal form (1254 ms)
    with Artsy shipping
      with no saved address
        ✕ sets shipping on order, selects shipping quote, and save address on user (218 ms)
        ✕ shows an error if Arta doesn't return shipping quotes (214 ms)
        ✕ removes saved address if save address is deselected after fetching shipping quotes (213 ms)
      with saved addresses
        Artsy shipping international only
          with artwork located in the US
            ✓ sets shipping on order if the collector is in the EU (96 ms)
            ✓ does not set shipping on order if the collector is in the US (613 ms)
          with artwork located in Germany
            ✓ does not set shipping on order if the collector is in the EU (759 ms)
            ✓ sets shipping on order if the collector is in the US (115 ms)
        Artsy shipping domestic only
          with artwork located in Germany
            ✓ sets shipping on order if the collector is in Germany (158 ms)
            ✓ does not set shipping on order if the collector is in the US (678 ms)
          with artwork located in the US
            ✓ does not fetch or show shipping quotes if the collector is in the EU (641 ms)
            with the collector in the US
              ✓ sets shipping with the default address on load (118 ms)
              ✕ shows shipping quotes for the default address on load (27152 ms)
              ✕ sets shipping on order, shows shipping quotes and saves the pre-selected quote (245 ms)
              ✓ selects a different shipping quote and saves it (270 ms)
              ✕ keeps the submit button enabled after selecting a shipping quote (28226 ms)
              ✓ routes to payment screen after saving shipping option (234 ms)
              ✓ reloads shipping quotes after editing the selected address (1082 ms)
              ✓ does not reload shipping quotes after editing a non-selected address (1457 ms)
    with pickup
      ✓ shows an empty phone number input with saved addresses (768 ms)
      ✓ sets pickup on order (547 ms)
      ✓ disables submission without a phone number (725 ms)
```

I fixed this by doing 2 things:
- fixing a country selector that was using role rather than `data-testid` (I also corrected other cases using `data-test` which RTL doesn't see)
- Making the `fillAddressForm` wait for the address line 1 input to be enabled, since it is disabled for an instant on load.

Then i went back and checked every fixed test (or so i thought) and removed the fix. In one case, it caused a second failure in the following test. I marked that one test with a temporary helper to trigger the failure for some future fixer.

But the problem remains that by my count i only fixed 20 tests, so 4 failures are unaccounted for. That said, hopefully by the time I post this the tests will have passed 🤷 .

</details>

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -1091,7 +1101,7 @@ describe("Shipping", () => {
         // expect(screen.getByRole("radio", { name: /White Glove/ })).toBeVisible()
         // expect(screen.getByRole("radio", { name: /Rush/ })).toBeVisible()
         // expect(screen.getByRole("radio", { name: /Premium/ })).toBeVisible()
-
+        triggerLeakyFailure()
```
starsirius, line 1104: I have a theory about the leak: The [mockCommitMutation](https://github.com/artsy/force/blob/161e26e1bdd2416d83cf7f4bd084c73288bf0c85/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx#L80) was mocked above but we [didn't reset it properly](https://github.com/artsy/force/blob/161e26e1bdd2416d83cf7f4bd084c73288bf0c85/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx#L188) after each test. So when a test fails here, the uncleared mock would leak to the subsequent test(s) and cause it to fail too.

Maybe we should do `jest.resetAllMocks()` instead of `jest.clearAllMocks()`.

PR #12905, opened 2023-09-19
refactor(alerts): use eligibility flag from MP instead of computing locally

anandaroop:

This PR solves [ONYX-363]

Follow-up to https://github.com/artsy/metaphysics/pull/5296

> **Warning**
> Depends on that ☝🏽 PR being merged, and then having the schema updated here in Force.
> (Then we can drop the first commit in this PR before merging.)

### Description

We have conditional logic to determine whether an artwork is eligible to have an alert created from it. This logic has multiplied throughout client code paths in Force and Eigen.

This addresses that tech debt by relying on the new `isEligibleToCreateAlert` flag computed in Metaphysics, instead of computing this eligibility locally.


nickskalkin (approved):
👍

dblandin (approved):
💜

PR #12913, opened 2023-09-20
fix(images): ensure minimum quality for lightbox artwork images

anandaroop:

This PR solves [ONYX-365]

### Description

This comes out of the [investigation](https://artsy.slack.com/archives/C045UP78DG8/p1694371647491209) into the color casts that afflict many artwork images.

This PR represents the _smallest_, most targeted fix for the problem — getting rid of the pink/green hues that appear on the primary artwork surface by explicitly requesting the artwork images at a quality (`85`) that has been demonstrated to sidestep the problem.

|Before|After|
|---|---|
|bf|af|

### Caveats

I'd consider it to _generally_ be an antipattern to have custom quality overrides sprinkled by developers throughout client apps, so I tried to avoid this by making the change [upstream in Metaphysics](https://github.com/artsy/metaphysics/blob/2ecff4865bd7d0a7f9e91a5663917e1fa36de705/src/schema/v2/image/services/config.ts#L4-L5).

But I learned that wouldn't quite be the systemic fix I was looking for, due to the variety of ways in which we have historically requested images in Force and Eigen.

Namely, a pungent mixture that includes…

- raw image urls from Metaphysics

- `cropped` and `resized` urls from Metaphysics’ `Image` type, which has the [quality defaults](https://github.com/artsy/metaphysics/blob/2ecff4865bd7d0a7f9e91a5663917e1fa36de705/src/schema/v2/image/services/config.ts#L4-L5) cited above

- `cropped` and `resized` urls generated locally via utility functions
  - which in turn rely on local `crop` and `resize` functions
    - which rely on the @artsy/img library that was meant(?) to ease switching image processing backends, and which has its own [quality defaults](https://github.com/artsy/img/blob/790f030d872f299d6b7c3c616eb1302ecf0b77d4/src/services.ts#L7)

- library `Image` components from palette & palette-mobile
  - one of which (palette-mobile) also knows how to create Gemini+Cloudfront resized urls, with its own [quality defaults](https://github.com/artsy/palette-mobile/blob/fbd1428d074599ef295cea67faa6b089cb362b1b/src/utils/createGeminiUrl.ts#L10)

I'd say it's still very do-able to increase the default quality across the board, if we want to do so, but I wanted to ship a solution to the most persistent complaint before going down that rabbit hole. (And if we do we can come back and undo this change, and be governed by the new defaults.)

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkLightbox.tsx
+++ b/src/Apps/Artwork/Components/ArtworkLightbox.tsx
@@ -120,6 +120,7 @@ export const ArtworkLightboxFragmentContainer = createFragmentContainer(
           isDefault
           placeholder: url(version: ["small", "medium"])
           fallback: cropped(
+            quality: 85
             width: 800
             height: 800
             version: ["main", "normalized", "larger", "large"]
@@ -130,6 +131,7 @@ export const ArtworkLightboxFragmentContainer = createFragmentContainer(
             srcSet
           }
           resized(
+            quality: 85
             width: 800
             height: 800
             version: ["main", "normalized", "larger", "large"]
--- a/src/__generated__/ArtworkImageBrowserLarge_Test_Query.graphql.ts
+++ b/src/__generated__/ArtworkImageBrowserLarge_Test_Query.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<6939d27f5e12f88efd24ae5fbaa9c383>>
+ * @generated SignedSource<<62eed7c06394a79a5d460954891e9450>>
  * @lightSyntaxTransform
  * @nogrep
  */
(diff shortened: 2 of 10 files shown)


joeyAghion (commented):
I'm OK with these narrow changes, but defer to @dzucconi about whether to decide this more centrally instead.

Metaphysics includes `img`, so theoretically I'd expect it _could_  skip applying its own defaults and defer to the lib instead. Not sure that's needed though, if the quality effects are most conspicuous in only certain locations that Force/Eigen can decide.

dzucconi:
This looks good to me and seems like a reasonable approach should solve the main source of complaints. I wouldn't be opposed to globally increasing quality but as you point out: it would be annoying. And I don't think thumbnails are a source of complaints.

anandaroop:
Wanted to check in on the [Cloudfront metrics](https://us-east-1.console.aws.amazon.com/cloudfront/v3/home?region=us-east-1#/monitoring/distribution/E24BF4TNI7YUZ1) since this went out. Cursor highlights when it [hit prod](https://app.circleci.com/pipelines/github/artsy/force/52807/workflows/e3e0b83d-048c-453e-8fd9-c9d0f0a10b8d) — nothing there to alarm me.

Screenshot 2023-09-22 at 12 20 05 PM

PR #12914, opened 2023-09-21
feat(EMI-1283): make submit order modal more deterministic

fladson:

This PR solves [EMI-1283]

### Description
The idea here is to send the order id to avoid checking if the slug from the current artwork page is the same as the user's last order. This slug check is causing issues with specs that are running concurrently, making the modal to not appear sometimes.
SCR-20230921-gt3

[EMI-1283]: https://artsyproduct.atlassian.net/browse/EMI-1283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun (approved):
Looks good to me! 🌟

PR #12925, opened 2023-09-22
Make Collapse a local component; be more defensive animating

starsirius:
The type of this PR is: **Refacor**

### Description

In https://github.com/artsy/palette/pull/1316, I'd like to improve the `Collapse` element in Palette for test flakiness. And I learned we should deprecate it from Palette. So this PR:

- Port it over to Force as a local component.
- Apply [the same fix](https://github.com/artsy/force/pull/12925/commits/f2c9fc3d3a73525d470e853b1a4b37625e8fd0a8) to make the animation a bit more defensive.

cc @artsy/emerald-devs


dzucconi (approved):
Thank you!

PR #12926, opened 2023-09-22
fix: Show artist gallery reps even without insights DIA-173

jonallured:
This PR updates the cocktail of booleans that we use to show and hide things in the new ArtistHeader component. I added a new boolean that indicates whether we have a third column - could be because we have gallery reps or could be because we have insights or both!

Then I just updated a few call sites to use that instead. I checked with a couple artists to verify that rendering works as expected:

<details><summary>screenshots</summary>
Screenshot 2023-09-22 at 10 59 42 AM
Screenshot 2023-09-22 at 10 59 45 AM

</details> 

https://artsyproduct.atlassian.net/browse/DIA-173

/cc @artsy/diamond-devs


dzucconi (approved):
🙏

PR #12931, opened 2023-09-23
refactor: Refactor shipping route

erikdstock:

This PR refactors the shipping route to lean heavily on relay and formik context to simplify our logic. It's quite large, but it currently works in manual testing for limited cases.

I did my best to maintain a consistent interface and UI, but tests will require some updates - at the very least to change the way we interact with inputs in order to accomodate formik's quirks ([eg](https://stackoverflow.com/a/65444080/4409186)).

It's likely we will want to build this as a review app and make further PRs to this branch. Currently a lot of old code is commented out because I want to be able to verify all behavior (especially tracking) has been migrated before deleting.

The basic changes are:
- Most logic from Shipping/index.tsx dealing with addresses has been moved into FulfillmentDetailsForm.
- Refactoring across the shipping step, preserving the original intent and meaning.
- Some helpers are no longer used. One big helper in fulfillmentDetailsForm contains some of the most complicated logic and might deserve special consideration.

Roughly the new structure is:
```tsx
relayProps = { order, me }
<ShippingFragmentContainer {...relayProps} >
  {/* fulfillment values vs shipping quote step kept in state; shipping quotes still managed on this component */ }
  <FulfillmentDetailsForm {...relayProps} >
    {/* Most shipping logic moved from parent into this new formik-based component */}
    <ShippingInputs | PickupInputs
      {/* 
        These inputs render inside the parent component, could be extracted
        if needed, but aren't particularly complicated 
      */}
    >
    <SavedAddresses me={relayProps.me}
      {/* 
        Slight refactorings to these components for managing user addresses.
        Selecing one sets the formik values for the address
      */}
    />
  </FulfillmentDetailsForm>
  {/* The existing, mostly untouched shipping quote logic remains in ShippingFragmentContainer. */}
</ShippingFragmentContainer>
```

Notes & Caveats
- I kept the single button for both address + shipping quote steps as in the current style, but considered having a separate button for each form. Not only would this be easier to do, but it could address a UI complaint when checking out on mobile.
- All forms should now support the richer phone number + country code input (not yet implemented)
- There is probably still some things to clean up, including work I brought in from previous spikes that should be deleted or moved out of new files.

<details><summary>Demos</summary>

</details> 

## Known TODOs:
**Features**
- [ ] Make the shipping route step smart about which form should be active at load time - fulfillment details or shipping quotes
- [ ] Shipping route should know to clear its previous values if the user goes back from the quotes to the shipping step
- [ ] Bring in phone number input to all address forms (country code is present in form values, but not sent to exchange)
- [ ] Migrate tracking calls
- [ ] restrict initial addresses and disable saved addresses based on shippable countries
**Code quality & testing**
- [ ] Make tests work
- [ ] Consolidate types of different addresses
- [ ] Code organization and consolidation
  - [ ] Separate shipping and pickup form layouts into separate components
  - [ ] Delete CheckoutAddress.tsx (this was an earlier attempt, and never used)
  - [ ] Delete unused imports, and delete their exports if we find they are no longer used at all
- [ ] Address numerous todos in code
- [ ] Consider an error boundary or other consolidated error handling


src/Apps/Order/Components/SavedAddresses.tsx
```diff
@@ -1,90 +1,101 @@
...
+  addressID?: string
+) => {
+  return (
+    (addressID && getAddressByID(addressList, addressID)) ||
+    getDefaultUserAddress(addressList)
+  )
 }
 
 const SavedAddresses: React.FC<SavedAddressesProps> = props => {
-  const { trackEvent } = useTracking()
```
erikdstock, line 63: SavedAddresses, SavedAddressItem and AddressModal are responsible for creating, editing and deleting user addresses. In addition when a user selects an address it sets all address values in the FulfillmentDetailsForm formik context.

src/Apps/Order/Mutations/useShippingOperations.ts
```diff
@@ -0,0 +1,121 @@
...
+  UpdateUserAddressMutation$data,
+  UserAddressAttributes,
+} from "__generated__/UpdateUserAddressMutation.graphql"
+import { useCallback } from "react"
+
+export type SaveFulfillmentDetailsResponse = NonNullable<
+  SetShippingMutation$data["commerceSetShipping"]
+>["orderOrError"]
+
+export const useShippingOperations = (
```
erikdstock, line 25: This hook may not be necessary. I also noticed that the settings page has its own mutations for updating user addresses, and wonder if these should eventually be joined.

src/Apps/Order/Routes/Shipping/FulfillmentDetailsForm.tsx
```diff
@@ -0,0 +1,830 @@
...
+    ? euOrigin
+      ? "EU"
+      : shippingCountry
+    : null
+  return { lockShippingCountryTo }
+}
+
+// Get initial values for the form using what may already be saved
+// on the order and the user's saved addresses.
+const getInitialValues = (
```
erikdstock, line 126: The bulk of the shipping route (index.tsx) logic has been moved into this file, with the most important logic in this rather large helper. It's main role is to provide the initial values to formik when the address form loads.

src/Apps/Order/Routes/Shipping/FulfillmentDetailsForm.tsx
```diff
@@ -0,0 +1,830 @@
...
+                  data-testid="savedAddressesCollapse"
+                  open={showAddressForm === "saved_addresses"}
+                >
+                  {
+                    // TODO: Maybe this stays in index? We render an error message
+                    // here if shipping quotes are present but empty.
+                    // isArtsyShipping &&
+                    //   shippingQuotes &&
+                    //   shippingQuotes.length === 0 &&
+                    //   renderArtaErrorMessage()
```
erikdstock, line 374: May need to pass this error message in from the parent Shipping Route (unless this form is to know about shipping quotes) - if so it can probably go below the Text element directly above.

src/Apps/Order/Routes/Shipping/FulfillmentDetailsForm.tsx
```diff
@@ -0,0 +1,830 @@
...
+                            {lockShippingCountryTo === "EU"
+                              ? "Continental Europe shipping only."
+                              : "Domestic shipping only."}
+                          </Text>
+                        </>
+                      )}
+                    </Column>
+                    <Column span={12}>
+                      {!autocomplete.loaded || autocomplete.enabled ? (
+                        <AutocompleteInput<AddressAutocompleteSuggestion>
```
erikdstock, line 480: I've been keeping this input up to date with changes in its current in-development usage in AddressForm (eg #12898)

src/Apps/Order/Routes/Shipping/FulfillmentDetailsForm.tsx
```diff
@@ -0,0 +1,830 @@
...
+            </Collapse>
+            {/* <Button
+              type="submit"
+              loading={formikProps.isSubmitting || undefined}
+              variant="primaryBlack"
+              width="50%"
+              // disabled={formikProps.isValid}
+            >
+              Save and Continue (shipping)
+            </Button> */}
```
erikdstock, line 718: Removed the local button and am passing the handler into a ref because in the current behavior the shipping form displays a continue button in two separate places depending on breakpoint. Personally I would advocate for treating these as 2 separate steps, with a unique button that appears below each form.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -112,308 +118,241 @@ export interface ShippingProps {
   isCommittingMutation: boolean
 }
 
-export const ShippingRoute: FC<ShippingProps> = props => {
-  const { trackEvent } = useTracking()
-  const { relayEnvironment } = useSystemContext()
+const getOrderFulfillmentDetails = (
```
erikdstock, line 121: This massive diff is going to be easier to review once all the commented out code and unused imports are gone, but personally I don't think it is worth the effort to try to understand the logic. Instead I would pay attention to the new changes here (minus commented old code) and rely on tests and observed behavior.

src/Apps/Order/Utils/shippingUtils.ts
```diff
@@ -1,16 +1,18 @@
...
 import {
   UpdateUserAddressMutation$data,
   UserAddressAttributes,
 } from "__generated__/UpdateUserAddressMutation.graphql"
 import { NEW_ADDRESS } from "Apps/Order/Components/SavedAddresses"
-import {
-  CommerceOrderFulfillmentTypeEnum,
-  SetShippingMutation$data,
-} from "__generated__/SetShippingMutation.graphql"
+import { SetShippingMutation$data } from "__generated__/SetShippingMutation.graphql"
```
erikdstock, line 10: There are probably some helpers in here no longer used, and some shared logic in the other files that could be bubbled up to helpers. Not only specific to this file, there are also some overlapping types that could probably be consolidated, mostly around varying address shapes.

src/Components/Address/__tests__/utils.jest.tsx
```diff
@@ -0,0 +1,238 @@
+import { Address } from "Components/Address/AddressForm"
+import { getYupAddressSchema } from "Components/Address/utils"
+
+describe.skip("getYupAddressSchema (local phone validation)", () => {
```
erikdstock, line 4: This and the file it is testing might be candidates for deletion/consolidation elsewhere.

erikdstock (commented):
Brief tour

src/Apps/Order/Components/SavedAddresses.tsx
iskounen: **question:** I didn't expect this refactor to affect saved addresses. Does FulfillmentDetailsForm need to know about saved addresses because it is now responsible for submitting the fulfillment details regardless of whether they are new or existing?

src/Apps/Order/Mutations/useShippingOperations.ts
```diff
@@ -0,0 +1,121 @@
...
+          onSuccess,
+          onError
+        )
+      },
+      [logger.error, relayEnvironment]
+    ),
+
+    // TODO: these aren't used bc SavedAddresses doesn't have ShippingProps.
+    // is this file needed?
+    createUserAddress: useCallback(
```
erikdstock, line 73: this whole file was a carryover form an earlier attempt and probably isn't necessary.

src/Apps/Order/Components/AddressModal.tsx
```diff
@@ -108,185 +121,184 @@ export const AddressModal: React.FC<Props> = ({
...
+              createUserAddress(
+                relayEnvironment,
+                addressInput,
+                handleSuccess,
+                handleError,
+                me,
+                closeModal
+              )
+            } else {
+              if (modalAction.type === "editUserAddress") {
```
dzucconi, line 181: I would very much encourage you to flatten out all these conditionals into short circuiting returns (no nested, no else)

src/Apps/Order/Components/AddressModal.tsx
```diff
@@ -108,185 +121,184 @@ export const AddressModal: React.FC<Props> = ({
...
+                      data-test="countryDropdown"
+                    />
+                  </Box>
+                  <Flex
+                    flexDirection="column"
+                    style={{
+                      width: "100%",
+                    }}
+                  >
+                    <Box height="100%"></Box>
```
dzucconi, line 242: ?

src/Apps/Order/Components/AddressModal.tsx
damassi: https://preslav.me/2023/09/22/ditch-that-else/

src/Apps/Order/Components/SavedAddresses.tsx
erikdstock: Yes: Considering fulfillment details as a step separate from the shipping quotes, the `SavedAddresses` component acts as a single input to set the entire address object. _Inside_ the component are controls to create/update/delete saved addresses, which in turn affect the available options for the `FulfillmentDetailsForm`. This is how it worked before, just with that logic localized with its behavior in a more predictable way.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -112,308 +118,241 @@ export interface ShippingProps {
...
+export const ShippingRoute: FC<ShippingProps> = props => {
+  const { trackEvent } = useTracking()
+  const shippingOperations = useShippingOperations(props, logger)
 
-  const [shippingOption, setShippingOption] = useState<
-    CommerceOrderFulfillmentTypeEnum
-  >(getShippingOption(props.order.requestedFulfillment?.__typename))
+  const [activeStep, setActiveStep] = useState<
+    "fulfillment_details" | "shipping_quotes"
+  >("fulfillment_details")
```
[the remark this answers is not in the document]
erikdstock: Thanks - I've made a note of this in a QA doc in the emerald notion space.

src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -780,6 +589,155 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
+        const orderResult = orderOrError?.order!
+
+        if (
+          formValues.fulfillmentType === FulfillmentType.SHIP &&
+          formValues.attributes.saveAddress
+        ) {
+          await shippingOperations.createUserAddress(formValues.attributes)
+        }
+
+        const orderResultRequiresArtsyShipping = getOrderRequiresArtsyShipping(
```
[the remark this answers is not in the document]
erikdstock: oh, great idea!

src/Apps/Order/Components/AddressModal.tsx
erikdstock: Good points - this was the last thing I did to get all the pieces working together again.

src/Components/Address/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+    addressLine1: Yup.string().required("Address is required"),
+    city: Yup.string().required("City is required"),
+    region: Yup.string().required("Region is required"),
+    postalCode: postalCodeValidator,
+    phoneNumber: phoneValidator,
+    phoneNumberCountryCode: Yup.string().required(
+      "Phone Number Country Code is required"
+    ),
+  })
+}
```
erikdstock, line 84: Todo: Ended up not using this bit yet, but it might be worth keeping around as we are going to go back to simple phone validation for this first refactor to avoid scope creep. It does have (WIP) unit tests tests which although they are really just testing we use yup correctly could be useful to isolate from the rest of our address form logic. Maybe they would just test the yup schema directly though - depends if we want to keep this factory method for incremental adoption of the better phone input.

Slightly related comment: I noticed that the phone validation query fires on every address change in any field, can we limit this?

src/Apps/Order/Components/AddressModal.tsx
```diff
@@ -66,23 +78,32 @@ const SERVER_ERROR_MAP: Record<string, Record<string, string>> = {
...
 
-  const title = modalDetails?.addressModalTitle
-  const createMutation =
-    modalDetails?.addressModalAction === "createUserAddress"
+  const [createUpdateError, setCreateUpdateError] = useState<string | null>(
+    null
+  )
+  const [showDialog, setShowDialog] = useState<boolean>(false)
+  if (!relayEnvironment) return null
+  if (!modalAction) return null
```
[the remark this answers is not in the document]
erikdstock: I _think_ the way it should work (based on previous behavior and my updates which tried to preserve that) is that the type is `ModalAction | null`, but not optional. Null means there is no active address edit/create happening. If this isn't the case we should update it.

erikdstock:
Ongoing notes:
- Must bring in new functionality from #12940.
- I've created a notion doc for tracking ongoing work here for the time being. It's linked in the @artsy/emerald-devs team QA docs section on notion.

erikdstock:
Emerald formally prioritized this work for our next sprint. In order to make it digestible I am going to open a new PR with this work behind a feature flag. That will require a little more restructuring to isolate the changes. cc @damassi

PR #12935, opened 2023-09-25
fix: Adjust My Collection artwork grid columns to match other grids

olerichter00:
## Description

Adjust My Collection artwork grid columns to match other grids.


olerichter00:
Just realized, it's actually looking like other artwork grids without filters ([example](https://www.artsy.net/new-for-you)). Although the artwork images are smaller than on pages with filters ([example](https://www.artsy.net/artist/barbara-kruger)).

PR #12929, opened 2023-09-22
Fix leaky tests

starsirius:
The type of this PR is: **Test**

### Description

In https://github.com/artsy/force/pull/12898#discussion_r1330888144, we suspected the mock was not cleared properly which caused the leaky tests issue. This PR verified it and fixed it by resetting the particular mocked commit mutation.

cc @artsy/emerald-devs

PR #12939, opened 2023-09-25
Add address verification tests

starsirius:

This PR solves [EMI-1413](https://artsyproduct.atlassian.net/browse/EMI-1413)

### Description

This adds tests to the shipping step for address verification flow. They might feel more like integration tests but given the number of regressions we saw in developing address verification, we think it's still important to cover the major flows.

The expanded test cases are:
```
Shipping
  with partner shipping
    with no saved address
      address verification
        with US enabled and international disabled
          ✓ uses recommended address
          ✓ goes back and edits address after verification
  with Artsy shipping
    with no saved address
      address verification
        with US enabled and international disabled
          ✓ uses recommended address
          ✓ goes back and edits address after verification
```

cc @artsy/emerald-devs 

[EMI-1413]: https://artsyproduct.atlassian.net/browse/EMI-1413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
@@ -179,6 +179,65 @@ const saveAndContinue = async () => {
   await flushPromiseQueue()
 }
 
+const recommendedAddress = {
+  addressLine1: "401 Broadway Suite 25",
+  addressLine2: null,
+  city: "New York",
+  region: "NY",
+  postalCode: "10013",
+  country: "US",
+}
+
+const verifyAddressWithSuggestions = async (relayEnv, input, suggested) => {
+  const addressLines = addr => {
+    return [
+      addr.addressLine1,
+      addr.addressLine2,
+      `${addr.city}, ${addr.region} ${addr.postalCode}`,
+      addr.country,
+    ]
+  }
+
+  const verificationResult = {
+    __typename: "VerifyAddressType",
+    verificationStatus: "VERIFIED_WITH_CHANGES",
+    suggestedAddresses: [
+      {
+        lines: addressLines(recommendedAddress),
+        address: recommendedAddress,
+      },
+    ],
+    inputAddress: {
+      lines: addressLines(validAddress),
+      address: validAddress,
+    },
+  }
+
+  const mutation = relayEnv.mock.getMostRecentOperation()
+  expect(mutation.request.node.operation.name).toEqual(
+    "AddressVerificationFlowQuery"
+  )
+  expect(mutation.request.variables).toEqual({
+    address: {
+      addressLine1: input.addressLine1,
+      addressLine2: input.addressLine2,
+      city: input.city,
+      region: input.region,
+      postalCode: input.postalCode,
+      country: input.country,
+    },
+  })
+
+  relayEnv.mock.resolveMostRecentOperation(operation => {
+    return MockPayloadGenerator.generate(operation, {
+      VerifyAddressType: () => verificationResult,
+    })
+  })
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -179,27 +179,82 @@ const saveAndContinue = async () => {
...
-
-  beforeAll(() => {
     ;(useTracking as jest.Mock).mockImplementation(() => ({
       trackEvent: jest.fn(),
     }))
     ;(useFeatureFlag as jest.Mock).mockImplementation(() => false)
   })
 
+  afterEach(() => {
+    jest.restoreAllMocks()
```
starsirius, line 255: Unrelated changes to this PR. It's a [followup]([891b251](https://github.com/artsy/force/pull/12939/commits/891b2519762ca77dcc624c54f9bc0c02280ea9d0)) from https://github.com/artsy/force/pull/12929 and I learned `restore` is the right way to do.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -179,27 +179,82 @@ const saveAndContinue = async () => {
...
+const recommendedAddress = {
+  addressLine1: "401 Broadway Suite 25",
+  addressLine2: null,
+  city: "New York",
+  region: "NY",
+  postalCode: "10013",
+  country: "US",
+}
+
+const verifyAddressWithSuggestions = async (relayEnv, input, suggested) => {
```
starsirius, line 191: There was discussion about rspec-like shared example/context on Slack. I started by this basic helper only for DRYing the address verification step. The step can be plugged in (or removed from) the flow as the [example](https://github.com/artsy/force/pull/12939/files#diff-24b4bdbd70487ae4bc0352d30cf15dad7d22c4ea225bc9dec6ebd9eb8b23de99R1208-R1212) below.

erikdstock: I'm in favor of trying this out, let's just be intentional about seeing it through.

non-blocking: As a matter of convention, since both this and `fillAddressForm` include expectations, would it be worth making clear that there is expectation behavior attached? We could just change the name, or we could look into [custom matchers.](https://jestjs.io/blog/2018/05/29/jest-23-blazing-fast-delightful-testing#custom-asynchronous-matchers)

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -1077,6 +1178,215 @@ describe("Shipping", () => {
...
+            mutationArg = mockCommitMutation.mock.calls[2][0]
+            expect(mutationArg.mutation.default.operation.name).toEqual(
+              "SelectShippingOptionMutation"
+            )
+            expect(mutationArg.variables).toEqual({
+              input: {
+                id: "2939023",
+                selectedShippingQuoteId: "1eb3ba19-643b-4101-b113-2eb4ef7e30b6",
+              },
+            })
```
erikdstock, line 1371: non-blocking: `expect(mockCommitMutation).toHaveBeenNthCalledWith(3, {...})` is also an option here. [Example](https://github.com/artsy/force/blob/4566d16b6a3fea1aff1c3022855a3baed6ed8d2b/src/Apps/Order/Components/__tests__/AddressVerificationFlow.jest.tsx#L329)

erikdstock (approved):
LGTM, thank you!

2 comments, 0 blocking.

PR #12941, opened 2023-09-26
fix(onyx-396): fixes search pagination issue when previous search results appear

nickskalkin:

This PR solves [ONYX-396]

### Description

Steps to reproduce:

- Type "kaws" in search input, press enter
- Type "warhol" in search input, press enter
- Under artworks grid, click on the "next" link

Outcome before this fix:

- Kaws's artworks are displayed in the grid

Expected outcome:

- Warhol's artworks should be displayed

[ONYX-396]: https://artsyproduct.atlassian.net/browse/ONYX-396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Search/Routes/SearchResultsArtworks.tsx
+++ b/src/Apps/Search/Routes/SearchResultsArtworks.tsx
@@ -11,6 +11,7 @@ import {
 } from "Components/ArtworkFilter/ArtworkFilterContext"
 import { useSystemContext } from "System/useSystemContext"
 import { SearchResultsArtworksFilters } from "Apps/Search/Components/SearchResultsArtworksFilters"
+import { useEffect, useState } from "react"
 
 interface SearchResultsRouteProps {
   viewer: SearchResultsArtworks_viewer$data
@@ -19,11 +20,20 @@ interface SearchResultsRouteProps {
 export const SearchResultsArtworksRoute: React.FC<SearchResultsRouteProps> = props => {
   const { match } = useRouter()
   const { userPreferences } = useSystemContext()
+  const [searchFitlerKey, setSearchFilterKey] = useState(
+    match.location.query.term
+  )
   const { viewer } = props
   const { sidebar } = viewer
 
+  useEffect(() => {
+    // refresh artwork filter on query change
+    setSearchFilterKey(match.location.query.term)
+  }, [match.location.query.term])
+
   return (
     <ArtworkFilter
+      key={searchFitlerKey}
       mt={4}
       viewer={viewer}
       filters={match.location.query}


nickskalkin:
props to @damassi for the implementation suggestion 🙏

dblandin (approved):
Awesome 💯

MounirDhahri:
Thanks for the fix

src/Apps/Search/Routes/SearchResultsArtworks.tsx
```diff
@@ -19,11 +20,20 @@ interface SearchResultsRouteProps {
...
   const { sidebar } = viewer
 
+  useEffect(() => {
+    // refresh artwork filter on query change
+    setSearchFilterKey(match.location.query.term)
+  }, [match.location.query.term])
+
   return (
     <ArtworkFilter
+      key={searchFitlerKey}
```
anandaroop, line 36: **question:** It would not have occurred to me to use a `key` outside of a list context; would be helpful to understand the rationale for this. I take it that this forces the component to essentially be remounted when the url (and therefore the key) changes, but why is this necessary, especially given the presence of L39 below? 🤔

anandaroop (commented):
Thanks for fixing @nickskalkin  (and for the suggestion @damassi).

Just one q below, to better understand what's going on here…

src/Apps/Search/Routes/SearchResultsArtworks.tsx
```diff
@@ -19,11 +20,20 @@ interface SearchResultsRouteProps {
 export const SearchResultsArtworksRoute: React.FC<SearchResultsRouteProps> = props => {
   const { match } = useRouter()
   const { userPreferences } = useSystemContext()
+  const [searchFitlerKey, setSearchFilterKey] = useState(
```
joeyAghion, line 23: Typo

src/Apps/Search/Routes/SearchResultsArtworks.tsx
damassi: When you reset the key, it will also also refreshes all internal state. No new key, state internal to components via `useState` is preserved across renders. But as soon as a parent key changes, each of those calls get recomputed. Note that this can be expensive. 

There's the question of whether `match.location.subFooVariables` as returned from `useRouter` is internally triggering its own state change when those values update. I would expect this to be the case, as you mention @anandaroop! but maybe that doesn't.

PR #12954, opened 2023-09-27
fix(start): Add --kill-all to exit when something errors out

damassi:

### Description

This (attempts) to fix our pipeline to prevent relay errors from slipping in by checking for errors in commands and then exiting. If relay isn't working, we should immediately fail so that devs can fix their relay command. 

Not sure if this is the best fix, but lets give it a shot and see if it doesn't create more problems.


dzucconi:
👍

PR #12962, opened 2023-09-28
feat [EMI-1480]: disable autocomplete secondary suggestions by default

erikdstock:
This PR makes the secondary suggestion feature of our autocomplete hook disabled by default. The reasoning here is that we need a strategy for handling apartment/unit/suite information in a [more specific way than "address line 1/2"](https://www.smarty.com/articles/address-line-2-meaning), and until then this experience will give a bad UX when automatically filling suggestions.

Change
The change is to disable this functionality in the hook unless it receives a specific parameter. We can turn this back on and refine it later.

User facing changes unless the secondary suggestions are explicitly enabled:
- No more `(3 entries)` when an address has sub-entries.
- Secondary information that doesn't rest in a nested entries like above is stripped out. This way we won't autofill that input (they still will, but we aren't contributing to the negative experience).
- Addresses are deduplicated based on their text representation.

~~The change will _not_ prevent us from showing a user addresses with additional unit numbers, so we should check if that part is still a big enough pain point that we want to go for a deeper change in how we present suggestions~~ (eg we could remove secondary information from the suggestions and then filter for unique addresses) **(update: I had to do this).**

<details><summary>Before and after (plus the `(entries)` were removed):</summary>

Before (note in this screenshot the `(entries)` has already been removed): 
image
After:
image

</details> 

<details><summary>Behavior documented in specs:</summary>
<pre>
 PASS  src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
  useAddressAutocomplete
    when the US address autocomplete feature flag is enabled
      fetching for autocomplete suggestions
        ✓ does not fetch for a query of less than 3 characters (2 ms)
        ✓ fetches top-level suggestions from the API with correct parameters (2 ms)
        ✓ sets line 2 to an empty string if the value is missing (3 ms)
        ✓ resets the suggestions when the country changes (4 ms)
        ✓ resets the suggestions without fetching when the search term is too short (2 ms)
        ✓ throttles calls to the Smarty API (2 ms)
        fetching secondary suggestions enabled
          ✓ returns the first 5 results in the correct shape (61 ms)
        fetching secondary suggestions not enabled
          ✓ returns simplified, deduplicated results without considering secondary information (54 ms)
          ✓ returns the correct single-line text for an address with multiple entries (3 ms)
          ✓ returns the first 5 results in the correct shape (52 ms)
          ✓ returns the correct single-line text for an address without multiple entries (3 ms)
      fetching secondary suggestions not enabled
        ✓ fetchSecondaySuggestions() does nothing (1 ms)
      fetching secondary suggestions enabled
        ✓ fetches secondary suggestions from the API with correct parameters (52 ms)
        ✓ returns the correct single-line text for an address with multiple entries (2 ms)
    enabled
      feature flag is enabled
        API key is availaible
          selected country is US
            ✓ returns true (1 ms)
          selected country is not US
            ✓ returns false (1 ms)
        API key is missing
          selected country is US
            ✓ returns false
          selected country is not US
            ✓ returns false (1 ms)
      feature flag is disabled
        API key is availaible
          selected country is US
            ✓ returns false (1 ms)
          selected country is not US
            ✓ returns false
        API key is missing
          selected country is US
            ✓ returns false (1 ms)
          selected country is not US
            ✓ returns false (1 ms)
</pre>
</details>

### DIFF
--- a/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
+++ b/src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
@@ -17,6 +17,7 @@ jest.mock("lodash/throttle", () => jest.fn(fn => fn))
 let mockFetch: jest.Mock
 
 type HookProps = Parameters<typeof useAddressAutocomplete>
+let hookOptions = { enableSecondarySuggestions: false }
 
 describe("useAddressAutocomplete", () => {
   const setupHook = (...args: HookProps) => {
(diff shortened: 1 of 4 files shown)


fladson (approved):

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
```diff
@@ -16,10 +16,13 @@ jest.mock("lodash/throttle", () => jest.fn(fn => fn))
...
+type HookProps = Parameters<typeof useAddressAutocomplete>
+let hookOptions = { enableSecondarySuggestions: false }
+
 describe("useAddressAutocomplete", () => {
-  const setupHook = (...args: Parameters<typeof useAddressAutocomplete>) => {
-    const result = renderHook(props => useAddressAutocomplete(props), {
-      initialProps: args[0],
+  const setupHook = (...args: HookProps) => {
+    const result = renderHook(props => useAddressAutocomplete(...props), {
+      initialProps: args,
```
erikdstock, line 25: one thing is i got a little mixed up here since i have to pass two args to the hook now. as a result the way i pass them looks a little weird.

PR #12963, opened 2023-09-28
feat: Dry run of testing horizontal filters

jonallured:
This PR adds an invisible change to the artist page by adding a couple meta tags. It uses a feature I setup in Unleash so that we could do a dry run of testing for the horizontal filters. The idea is to let this sit on production for a bit and then we can verify that the data we want to analyze is flowing through the data pipeline as expected.

/cc @artsy/diamond-devs @leodmz

### DIFF
--- a/src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
+++ b/src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
@@ -7,6 +7,10 @@ import { Title, Meta } from "react-head"
 import { useRouter } from "System/Router/useRouter"
 import { useJump } from "Utils/Hooks/useJump"
 import { ArtistWorksForSaleEmptyFragmentContainer } from "Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty"
+import {
+  useFeatureVariant,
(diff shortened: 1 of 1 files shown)


src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
```diff
@@ -34,11 +38,26 @@ const ArtistWorksForSaleRoute: React.FC<ArtistWorksForSaleRouteProps> = ({
...
+  const variantName = variant?.name || "unknown"
+
+  const { trackFeatureVariant } = useTrackFeatureVariant({
+    experimentName,
+    variantName,
+  })
+
+  useEffect(() => {
+    trackFeatureVariant()
+  })
```
dzucconi, line 52:
```tsx
useEffect(trackFeatureVariant, [trackFeatureVariant])
```

dzucconi: (This is going to run on every render, otherwise)

jonallured: Thanks, fixed! 🙏

PR #12971, opened 2023-09-29
chore [EMI-1447]: add a test for autocomplete functionality (wip: only works alone)

erikdstock:
The type of this PR is: **Chore** and it partially completes [EMI-1447]
TODOs
- [ ] Fix the issues documented below with these tests
- [ ] Add these new tests to shipping refactor (Shipping2) routes - pending ok if they can't be immediately enabled, but ticket it in checkout flow `epic/refactor-shipping` label.

This is a test for the basic address autocomplete functionality in the shipping route. It only works when you run it alone - any tests that appear above it will cause it to fail due to `mockFetch()` not being called. For that reason the test is currently skipped.

I paired with @starsirius and we were able to fix this issue by [removing the `_.throttle()` usage in the autocomplete hook](https://github.com/artsy/force/pull/12971#:~:text=_.throttle%20call-,296cd7e,-Revert%20%22poc%3A%20fix). Before that I was sure that I was mocking `fetch` wrong. It's not clear why this is happening, but if you log the private, memoized, throttled `fetchSuggestions` right before it is called, then run the spec with two tests that exercise autocomplete (eg, duplicate the existing spec, or use a tracking one when that gets implemented), you can see that `fetchSuggestions` is a function in the first spec and undefined in each subsequent spec.

<details><summary>example showing broken spec with missing memoized function - removing `_.throttle` call fixes this</summary>

image

</details> 

[EMI-1447]: https://artsyproduct.atlassian.net/browse/EMI-1447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
@@ -26,7 +26,7 @@ import {
   updateAddressSuccess,
 } from "Apps/Order/Routes/__fixtures__/MutationResults/saveAddress"
 import { ShippingTestQuery$rawResponse } from "__generated__/ShippingTestQuery.graphql"
-import { screen } from "@testing-library/react"
+import { screen, waitFor } from "@testing-library/react"
 import { useTracking } from "react-tracking"
 import { useFeatureFlag } from "System/useFeatureFlag"
 import {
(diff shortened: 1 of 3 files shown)


erikdstock:
Closing - finished by #13189

PR #12973, opened 2023-10-02
fix(alerts): don't include current artwork in suggested inventory

anandaroop:

This PR solves [ONYX-397]

### Description

When saving an alert, we use the alert's criteria to display potentially matching inventory.

If the alert is being saved from an _artwork_, we don't want to redundantly show that very artwork again as part of that inventory — a problem that is especially apparent when the alert criteria only match a handful of works.

This change therefore allows the current artwork to be passed in the `excludeArtworkIDs` param that is already supported by the `artworksConnection` field.

|Before|After|
|---|---|
|repeated|not|

Also: found some more dead code, imported only in tests. Removed in [1e740d0](https://github.com/artsy/force/pull/12973/commits/1e740d0dabdb208ffa9a5f5865381282494b06d1).

[ONYX-397]: https://artsyproduct.atlassian.net/browse/ONYX-397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/SavedSearchAlert/ConfirmationArtworksGrid.tsx
```diff
@@ -127,6 +128,7 @@ export const ConfirmationArtworksGridQueryRenderer: FC<ConfirmationArtworksGridQ
           first: NUMBER_OF_ARTWORKS_TO_SHOW,
           sort: "-published_at",
           forSale: true,
+          excludeArtworkIDs: excludeArtworkIDs ?? [],
```
anandaroop, line 131: Unsure how to add a meaningful test for this, but open to ideas.

MounirDhahri (approved):
LGTM! thanks Roop for adding this fast

dblandin (approved):
💯 — very handy to have that `excludeArtworkIDs` query arg available!

nickskalkin:
belated feedback: Another approach could be to use new [savedSearch.suggestedArtworksConnection](https://github.com/artsy/metaphysics/blob/3b74b13ae7cc85113a1fbdb77677eca54a85a7f8/src/schema/v2/artwork/index.ts#L1541) field here. This query field encapsulates all logic related to building search criteria input field, including `exclude_ids`.

PR #12980, opened 2023-10-03
chore(EMI-1446): add address autocomplete tracking

rquartararo:

This PR solves [EMI-1446]

### Description

This adds tracking for the address autocomplete according to https://github.com/artsy/cohesion/pull/457

[EMI-1446]: https://artsyproduct.atlassian.net/browse/EMI-1446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Address/AddressForm.tsx
+++ b/src/Components/Address/AddressForm.tsx
@@ -15,7 +15,13 @@ import {
   useAddressAutocomplete,
 } from "Components/Address/useAddressAutocomplete"
 import { useTracking } from "react-tracking"
-import { ActionType, ContextModule, OwnerType } from "@artsy/cohesion"
+import {
+  ActionType,
+  ContextModule,
+  EditedAutocompletedAddress,
+  OwnerType,
+  SelectedItemFromAddressAutoCompletion,
+} from "@artsy/cohesion"
 import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
 
 const ENABLE_SECONDARY_SUGGESTIONS = false
@@ -175,13 +181,15 @@ export const AddressForm: React.FC<AddressFormProps> = ({
   const autocompleteRef = React.createRef<HTMLInputElement>()
 
   const trackAutoCompleteEdit = field => {
-    trackEvent({
+    const event: EditedAutocompletedAddress = {
       action: ActionType.editedAutocompletedAddress,
       context_module: ContextModule.ordersShipping,
-      context_page_owner_type: OwnerType.ordersShipping,
-      context_page_owner_id: contextPageOwnerId,
+      context_owner_type: OwnerType.ordersShipping,
+      context_owner_id: contextPageOwnerId,
       field: field,
-    })
+    }
+
+    trackEvent(event)
   }
 
   return (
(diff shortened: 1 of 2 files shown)


src/Components/Address/useAddressAutocomplete.ts
```diff
@@ -154,15 +159,26 @@ export const useAddressAutocomplete = (
...
         setResult(finalSuggestions.slice(0, 5))
+
+        trackEvent({
+          action: ActionType.addressAutoCompletionResult,
+          context_module: ContextModule.ordersShipping,
+          context_page_owner_type: "orders-shipping",
+          context_page_owner_id: contextPageOwnerId,
+          input: search,
+          suggested_addresses_results: finalSuggestions.length,
+        })
```
erikdstock, line 170: nit: Can we add `as TheCohesionType` types to all of the params we pass to `trackEvent()`? This surfaces errors in our interfaces:
image

erikdstock (commented):
Looks good with one minor comment to make sure we stick to the cohesion schema.

TMcMeans (dismissed):
LGTM 👍🏾 🥇

src/Components/Address/useAddressAutocomplete.ts
erikdstock: Update: In the case above i think our cohesion schema is actually inconsistent (see the doc comment vs the interface [here](https://github.com/artsy/cohesion/pull/457/files#diff-ee70df3a6cb0738382827014e641e2491d82c625fcd0596f7f1075da3454b724R249-R260)). I'd still like to add the `as TheType` bits to help us catch things like this.

TMcMeans: is the goal of adding` TheCohesionType` as a type to the params in trackEvent() to enforce type checking to make sure that the values compile as expected?

src/Components/Address/AddressForm.tsx
```diff
@@ -112,6 +121,10 @@ export const AddressForm: React.FC<AddressFormProps> = ({
     }
     setKey(key)
     onChangeValue(key, ev.currentTarget.value)
+
+    if (userSelectedAddressOption && key !== "name" && key !== "phoneNumber") {
+      trackAutoCompleteEdit(key)
```
erikdstock, line 126: Sorry, one more nit here - can we only track this once per selected address? 
image

erikdstock: ah, maybe that makes sense - adding secondary info would mean something different than fixing the zip code. Your call - either track it the first time or once for each field, and we can confirm data's preference later (cc @daytavares)

src/Components/Address/AddressForm.tsx
```diff
@@ -127,13 +123,13 @@ export const AddressForm: React.FC<AddressFormProps> = ({
     onChangeValue(key, ev.currentTarget.value)
 
     if (
-      !userEditedSelectedAddress &&
-      userSelectedAddressOption &&
+      selectedAddressOption.option &&
+      !selectedAddressOption.edited &&
```
erikdstock: nice :+1:

erikdstock (commented):
Looking good - can we type the events as discussed in the [#web-practice slack thread here](https://artsy.slack.com/archives/C2TQ4PT8R/p1696363649011089?thread_ts=1696361425.671659&cid=C2TQ4PT8R)? Then ok to merge from me.

Edit: As i posted this i see the types were added. Thanks!

erikdstock (approved):
Thanks - Good to merge once we confirm these types with @daytavares!

src/Components/Address/useAddressAutocomplete.ts
daytavares: Created the [PR](https://github.com/artsy/cohesion/pull/465)

PR #12981, opened 2023-10-04
fix(router): Ensure page transitions execute only when page is fully done loading 

damassi:

### Description

This addresses a long-standing and difficult to fix issue involving very unacceptable UX layout jank when transitioning between routes that have different route layouts. 

**The problem:**
- User navigates to an artwork page, with a default layout
- User clicks "Make Offer", which transitions to the Order app, with a `LogoOnly` layout
- Immediately upon click, the page layout changes and the logo-only appears, but the main content stays because the page is still fetching the query
- When fetch is done, main content updates to the latest

I've attempted to fix this about 20 times! But ran into it yet again today while working on migrating convos work from V2 over to force and thought I'd give it another shot. 

**The solution**
- Globally export the `isFetching` setState hook so that it can be used from outside a react component 
- Add a new `loadingIndicatorMiddleware` to our router
- Tap into the lower-most router lifecycle actions to trigger loading and not loading states
  -  `NAVIGATE` and `UPDATE_LOCATION` should always set fetching to true
  - `RESOLVE_MATCH` should set loading to false 
- When these actions match, update the state
- Cached pages resolve so quickly as to never show loader

Obviously, this fix isn't entirely ideal and open to other approaches, but at least it does afford us the ability to clean up all of the extreme weirdness that was present in `RenderStatus`. And tapping into the root router lifecycle events _does_ feel more correct. However, there's some hidden complexity added here. 

**Before:**

https://github.com/artsy/force/assets/236943/f90f573c-d484-4876-aaed-cbea0d55f817

**After:**

https://github.com/artsy/force/assets/236943/db37ecf1-9d88-4510-b702-3437ba81667c

### DIFF
--- a/src/Apps/Components/Layouts/index.tsx
+++ b/src/Apps/Components/Layouts/index.tsx
@@ -1,8 +1,10 @@
+import { usePrevious } from "@artsy/palette"
 import { LayoutBlank } from "Apps/Components/Layouts/LayoutBlank"
 import { LayoutContainerOnly } from "Apps/Components/Layouts/LayoutContainerOnly"
 import { LayoutDefault } from "Apps/Components/Layouts/LayoutDefault"
 import { LayoutLogoOnly } from "Apps/Components/Layouts/LayoutLogoOnly"
 import { LayoutNavOnly } from "Apps/Components/Layouts/LayoutNavOnly"
+import { useSystemContext } from "System/SystemContext"
 import { FC, ReactNode } from "react"
 
 export interface BaseLayoutProps {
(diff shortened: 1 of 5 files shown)


src/System/Router/RenderStatus.tsx
```diff
@@ -56,25 +43,20 @@ export const RenderPending = () => {
...
-  const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+  const { isFetching } = useSystemContext()
 
   if (isFetching) {
-    timeoutRef.current = setTimeout(() => setFetching?.(false), 0)
     return undefined
   }
 
-  if (timeoutRef.current) clearTimeout(timeoutRef.current)
-
```
dzucconi, line 69: YES

src/Apps/Components/Layouts/index.tsx
```diff
@@ -24,6 +26,17 @@ export const LAYOUTS = {
 export type LayoutVariant = keyof typeof LAYOUTS
 
 export const Layout: FC<LayoutProps> = ({ variant = "Default", children }) => {
+  const { isFetching } = useSystemContext()
+
+  // If we're fetching, we want to render the previous layout and not execute
+  // the new one right away.
+  const previousVariant = usePrevious(variant)
```
dzucconi, line 33: YES

src/System/SystemContext.tsx
```diff
@@ -93,6 +95,10 @@ export const SystemContextProvider: FC<SystemContextProps> = ({
     getUser(props.user!)
   )
 
+  // Globally export the fetch toggle so that we can access it from the router's
+  // loadingIndicatorMiddleware actions
+  setRouteFetching = useMemo(() => setFetching, [setFetching])
```
dzucconi, line 100: Clever

dzucconi (approved):
Nice work; this feels so much better.

PR #12987, opened 2023-10-04
feat(devx): allow configurability of RTL output level

anandaroop:
Welcome to the bike shed  🚲 🏠  🎨🖌️  

Follow-up to https://github.com/artsy/force/pull/12984 which kicked off a discussion about the detailed output of React Testing Library.

This PR:

- retains the quieter output configured above^ when running in CI
- allows developers to opt-in to this behavior locally via env var, rather than making it the default
- adds `yarn jest:terse` so you don't have to remember the env var

See [this comment thread](https://github.com/artsy/force/pull/12984#discussion_r1346424738), and particularly the output [examples here](https://github.com/artsy/force/pull/12984#discussion_r1346458121) to see my reasoning for this change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,9 @@
 # -- Display analytics calls in browser console
 # SHOW_ANALYTICS_CALLS=true
 #
+# -- Hide detailed debugging output from react-testing-library
+# TERSE_RTL_OUTPUT=true
+#
 
 # ------------------------
 # Localhost Environment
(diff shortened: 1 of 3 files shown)


erikdstock:
What if the env var was available as a `yarn jest:something` command? Then you don't have to think about an env var.

It's weird that RTL doesn't support this natively.

damassi (approved):
This works, and is even better 👍

PR #13004, opened 2023-10-10
fix: updates `Artist#image` to `Artist#coverArtwork`

dzucconi:
Re: [DIA-192](https://artsyproduct.atlassian.net/browse/DIA-192)

The `Artist#coverArtwork` is replacing the `Artist#image` field, which will be deprecated.

This updates the, widely used, artist cell and artist entity header. The search dropdown is doing it's own thing (it returns a non-artist type with it's own `imageUrl` and I need to come up with an approach there.

[DIA-192]: https://artsyproduct.atlassian.net/browse/DIA-192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


araujobarret (approved):
Looks good, we do the same in Eigen but I'll investigate if we still have more of it there 👍

PR #13010, opened 2023-10-11
fix: get timezone from luxon when Intl timezone is undefined [WIP]

MounirDhahri:
The type of this PR is: fix

This PR fixes an issue where multiple users in Seoul where not able to access artsy.net due to Timezone being returned as `undefined`. This seems to be a fresh issue affecting some chrome versions and I expect it to be solved in few weeks. 

Review app link: fix-broken-artsy-app-in-seoul.artsy.net


araujobarret (approved):
😍 seems like a super legit fix, thanks a lot for looking into this! 💪

PR #13013, opened 2023-10-11
fix: offer input value

leamotta:

### Description

Solves https://artsy.slack.com/archives/C03GX7A4KV0/p1697034020143039
Since we're using the value of the OfferInput to properly format and display, we need to make sure that is being passed on by the parent component. This worked for the normal offer flow but it wasn't being set on the rest use cases for that component.

https://github.com/artsy/force/assets/7518671/d4a4b465-4eca-4696-ab94-50dba6c2462c


iskounen (approved):
My preference would be to set the value to `undefined` in `onChange` when the value becomes blank so that we can differentiate when the user enters 0 versus erases the value. However, we don't accept any offers of 0 so I think that this is OK.

PR #13017, opened 2023-10-12
feat: numeric input mode for auth code

leamotta:

### Description

While working on #13011 I noticed we could use the numeric input mode for the auth code, to make the authentication flow easier on mobile web by triggering the numeric keyboard on that input


gkartalis (approved):

MrSltun (approved):
🔢

PR #13020, opened 2023-10-12
feat: implements a transitioning panel

dzucconi:
https://github.com/artsy/force/assets/112297/394dba7d-0514-4d95-835b-7e5b570ff303

The transition animation isn't the hard part here, rather it's maintaining a natural 'order' to the panels. If you're navigating to a new panel from the home, it should be oriented on the right, if you're navigating back then it should be oriented on the left. I support N panels here in the event we need more and they will appear on the left or right depending on their ordering within the provider. You also need to be able to trigger navigation from inside panels so that's where the context provider comes in.

We don't animate changes in modal size. Theoretically we could but this adds a lot more complexity. We may not be adjusting size at all, it turns out. I suspect this will appear a lot less strange when it contains real content.

### DIFF
--- a/src/Components/TransitionPanel.tsx
+++ b/src/Components/TransitionPanel.tsx
@@ -1,0 +1,111 @@
+import { Box } from "@artsy/palette"
+import {
+  Children,
+  FC,
+  createContext,
+  isValidElement,
+  useContext,
+  useRef,
+  useState,
+} from "react"
+
+const TransitionPanelContext = createContext<{
+  active: number
+  navigateTo: (index: number) => void
+}>({
+  active: 0,
+  navigateTo: () => {},
+})
+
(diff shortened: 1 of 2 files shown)


tam-kis (approved):
I would be fine with this as I will be adding it to the new "Add Filters" feature and look more into these transitions with real content. Will let you know if anything comes up 👍

damassi:
Ah @dzucconi - what good timing 😄 We're actually going to need this on the very next project we're gonna work on in Volt. Can this be moved to Palette?

dzucconi:
Ah interesting — @damassi. I'm going to wait until @tam-kis gives it a whirl with a real use-case, unless it's pressing?

damassi:
Nope, not pressing! We're still figuring out designs on our end.

PR #13021, opened 2023-10-12
feat(EMI-1377): scroll to address errors and arta shipping options in checkout

rquartararo:

This PR solves [EMI-1377]

@artsy/emerald-devs 

### Description

When a user clicks "save and continue" on shipping and payment steps, it will scroll up to the form validation errors on the address forms or phone number input.

In the case of ARTA shipping, it will scroll to the shipping options if one hasn't been selected. 

This prevents users on mobile from clicking "save and continue" and not being able to know why they are not proceeding to the next step. 

### Videos

https://github.com/artsy/force/assets/50849237/beebd825-3d25-4eb1-aab9-079bdaa0967a

[EMI-1377]: https://artsyproduct.atlassian.net/browse/EMI-1377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/__tests__/Payment.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Payment.jest.tsx
@@ -92,8 +92,11 @@ jest.mock("Apps/Order/Utils/commitMutation", () => ({
 }))
 const trackEvent = jest.fn()
 
-const scrollIntoViewMock = jest.fn()
-window.HTMLElement.prototype.scrollIntoView = scrollIntoViewMock
+const mockJumpTo = jest.fn()
+jest.mock("Utils/Hooks/useJump", () => ({
+  useJump: () => ({ jumpTo: mockJumpTo }),
+  Jump: () => null,
+}))
 
 const testOrder: PaymentTestQuery$rawResponse["order"] = {
   ...BuyOrderWithShippingDetails,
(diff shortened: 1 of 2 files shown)


iskounen:
Thank you for the video examples - the scrolling looks great. One thing that caught my attention is that the screen size of the video examples is long enough to show the whole form, so the error will be in view regardless of whether it happens at the top of the form (ie. Full Name or Address Line 1) or at the bottom of the form (ie. Postal Code or Phone Number).

If a validation error occurs at the bottom of the form with a smaller screen size, will the scrolling behavior only show the top of the form?

If not, we may need to add some more refs (ie. `paymentDetailsTopRef`,  `paymentDetailsMiddleRef`,  `paymentDetailsBottomRef`) to make sure users don't scroll to another dead area of the form.

iskounen:
I think that we will end up needing a separate ref on the phone number field, but I don't want to optimize too early. I'd be OK to test this out on a few devices when it hits the main branch and then figure out how we should address smaller screens (or if we even need to).

damassi (commented):
Note that we have this [`useJump` hook](https://github.com/artsy/force/blob/main/src/Utils/Hooks/useJump.tsx) that we can use for this! It takes into account a few different environmental conditions. There are examples around the codebase on how it's used.

src/Apps/Order/Routes/Payment/index.tsx
```diff
@@ -221,7 +222,7 @@ export const PaymentRoute: FC<PaymentRouteProps> = props => {
       const result = await CreditCardPicker?.current?.getCreditCardId()
 
       if (result?.type === "invalid_form") {
-        paymentDetailsRef.current?.scrollIntoView({ behavior: "smooth" })
+        jumpTo("paymentDetailsTop", { behavior: "smooth" })
```
damassi, line 225: 👌

iskounen (approved):
LGTM

PR #13022, opened 2023-10-13
fix(sell): fix hard jump between sell and sell/submission, and fix broken back button to /contact-info

damassi:

### Description

Noticed that the Sell button on /sell page was doing hard jumps between pages, circumventing our SPA router and impacting performance. This was due to the button pointing to a URL which on the server then performs a server-side redirect. Instead, just add a proper route that matches `/sell/submission`. This is to preserve the SEO of having a canonical `sell/submission` link, vs updating the button to point directly at `sell/submission/contact-info`. 

Additionally, this fixes the back button that goes back to contact info, as it currently force redirects you to artwork details, which is confusing. I'm not sure entirely of the history behind this code, but a back button shouldn't ever be subject to redirect rules that prevent one from going back!

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/Utils/redirects.ts
+++ b/src/Apps/Consign/Routes/SubmissionFlow/Utils/redirects.ts
@@ -5,8 +5,8 @@ import { redirects_submission$data } from "__generated__/redirects_submission.gr
 import {
   getArtworkDetailsFormInitialValues,
   SubmissionType,
-} from "../ArtworkDetails/Components/ArtworkDetailsForm"
-import { getUploadPhotosFormInitialValues } from "../UploadPhotos/UploadPhotos"
+} from "Apps/Consign/Routes/SubmissionFlow/ArtworkDetails/Components/ArtworkDetailsForm"
+import { getUploadPhotosFormInitialValues } from "Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos"
 import {
   artworkDetailsValidationSchema,
   uploadPhotosValidationSchema,
@@ -37,7 +37,8 @@ const checkArtworkDetailsFormValid = redirectToIf(
     )
 )
 
-const checkUploadPhotosFormValid = redirectToIf(
+// TODO: Is this needed as a rule to go back to edit contact info?
+export const checkUploadPhotosFormValid = redirectToIf(
   id => `/sell/submission/${id}/upload-photos`,
   submission =>
     !uploadPhotosValidationSchema.isValidSync(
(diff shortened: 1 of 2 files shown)


src/Apps/Consign/Routes/SubmissionFlow/Utils/redirects.ts
```diff
@@ -65,8 +66,12 @@ export const redirects = {
...
       rules: [
         checkSubmissionExist,
-        checkArtworkDetailsFormValid,
-        checkUploadPhotosFormValid,
+        // TODO: Is this needed? A user should be able to edit their contact
+        // info independent of all else. With these rules enabled, the back
+        // button is broken.
+
+        // checkArtworkDetailsFormValid,
+        // checkUploadPhotosFormValid,
```
damassi, line 74: I would imagine the only check here to go to `/sell/some-consign-id/contact-info` would be the presents of a submission. Not sure about these other steps, or how this might impact other areas of the code.

mzikherman:
There's actually a ticket kicking around Onyx describing a broken back button for this page! It's  https://artsyproduct.atlassian.net/browse/ONYX-279

damassi:
That ticket is actually incorrect; the more likely case is that this button is totally broken 😄 In this case -- and there are multiple back button / redirect rules -- the back button takes you back but then redirects you forward due to some seemingly-unnecessary redirect rule.

As a side note, I was mildly suprised about the complexity of that back button code, which seems to be duplicated across all three steps. Stuff like this would ideally be caught in code review:

Screenshot 2023-10-13 at 12 03 20 PM

mzikherman:
Lol that screenshot is insane

damassi:
Yah don't mean to call anyone out with it, only to note that we're likely dealing with a confluence of issues here. There's complex, unnecessarily duplicated code that is all likely a bit different, there's server-side router redirects with certain rules that don't seem to apply to certain steps, and theres some shared UI thats logically displayed depending on the page context. 

Ideally this can all be simplified, but at the very least lets allow our users to go back and edit their submission contact info! (And also give this code a thorough audit.)

damassi:
@dariakoko - can you check out this PR? Would like to merge if ok.

dariakoko (approved):
Thank you for the changes @damassi

Your changes fix the issue with the back button. While testing I found two more issues (not related to your changes): 
1. The browser back button behaviour feels wrong - it can close the form without asking the user if he wants to continue. Not sure if the confirmation before exiting is necesarry, but in my opinion it's nice to have
2. **[error]** Finish step 1, go to step 2, go back to step 1 and to the step 2 again. Submit step 2 - you see error message. I think it could be related to the fact that at this moment the submission already exists and when going from step 1 to step 2 we consider this, but when going from step 2 to 3 we don't.

PR #13025, opened 2023-10-16
Refactor: Checkout flow shipping step

erikdstock:

Completes [EMI-1431]

This PR lays the groundwork for refactoring the shipping step of the checkout flow, primarily using Formik to handle address form state and eliminating a lot of intermediate + useState variables along the way. There is still a lot of room for improvement, both in refactoring and functionality.

This is feature flagged in unleash, so will not be exposed to users yet.

The structure and main logic is here, but before taking out of draft I need to:
- [x] Create/make public a document I'm working on to track known issues and quirks in existing behavior that I don't want to miss. That document is [here](https://www.notion.so/artsy/Shipping-route-refactor-Ongoing-e89b381c73424a9f8c0c78822e9aa2fd). I will use it to ticket real followup work in Jira.
- [x] Mark failing tests in Shipping2.jest.tsx pending.
- [x] Go once more over for cleanup.

The PR is ready. Because it is feature flagged it should be safe to merge and iterate on. I will follow up with a code review & tour of the changes.

[EMI-1431]: https://artsyproduct.atlassian.net/browse/EMI-1431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Address/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const usPostalCodeRegexp = /^\d{5}(-\d{4})?$/
+
+export const postalCodeValidator = Yup.string().when("country", {
+  is: country => country === "US",
+  then: Yup.string()
+    .required("ZIP code is required")
+    .matches(usPostalCodeRegexp, "Invalid postal code"),
+  otherwise: Yup.string().required("Postal code is required"),
+})
```
erikdstock, line 57: Some cleanup to do here - I don't think I am using anything except this validator in the current version.

src/Apps/Order/orderRoutes.tsx
```diff
@@ -20,12 +21,28 @@ const OfferRoute = loadable(
...
+
+const ShippingRoute = props => {
+  const newShippingRoute = useFeatureFlag("refactor-shipping-route")
+
+  if (newShippingRoute) {
+    return <ShippingRoute2 {...props} />
+  }
+
+  return <ShippingRouteLegacy {...props} />
+}
```
erikdstock, line 45: Feature flag code.

src/Apps/Order/Routes/Shipping2/shippingUtils.ts
```diff
@@ -0,0 +1,229 @@
...
+
+// Interface to separate local values from what we want to expose
+// in the react context.
+interface OrderData extends ShippingContextProps {
+  shippingQuotesEdges: ShippingQuotesConnectionEdges
+  selectedShippingQuoteId?: string
+}
+
+// Compute and memoize data from the saved order.
+export const useLoadComputedData = (props: ShippingProps): OrderData => {
```
erikdstock, line 167: The structure of Shipping2/shippingUtils.ts is similar to the existing shippingUtils.ts, but with much fewer exports. Most things about inferring the state of the order are contained in this one giant hook. This is used to always reflect current order state from the relay-provided order prop instead of maintaining our own copies in the shipping route, and was one of the main challenges i found in the refactor.

src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
```diff
@@ -0,0 +1,276 @@
...
+const getBestAvailableAddress = (
+  addressList: SavedAddressType[],
+  addressID?: string,
+  availableShippingCountries?: string[]
+) => {
+  return (
+    (addressID && getAddressByID(addressList, addressID)) ||
+    getDefaultUserAddress(addressList, availableShippingCountries)
+  )
+}
```
damassi, line 56: #superminor - can these helper functions be moved below the primary concern (`const SavedAddress = ... `). Its nice to not have to wade through intermediate stuff before getting to the part that's most likely to change in the future.

damassi: Or, since there's already a shippingUtils file, it can just go there

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -0,0 +1,887 @@
...
+    }
+    return "fulfillment_details"
+  }, [isArtsyShipping, savedFulfillmentType, selectedShippingQuoteId])
+
+  return {
+    savedOrderData: orderData,
+    initialValues,
+    targetStep,
+  }
+}
```
damassi, line 199: Can we move these two big functions down below `ShippingRoute`?

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -0,0 +1,887 @@
...
+        trackEvent({
+          action: ActionType.errorMessageViewed,
+          context_owner_type: OwnerType.ordersShipping,
+          context_owner_id: props.order.internalID,
+          title: "Invalid address",
+          message:
+            "There was an error processing your address. Please review and try again.",
+          error_code: error.code,
+          flow: "user submits a shipping option",
+        })
```
damassi, line 291: Can we move all of the tracking function calls used throughout the shipping file into their own `useShippingTracking` file? Something like this: https://github.com/artsy/force/blob/main/src/Apps/Auction/Hooks/useAuctionTracking.ts

iskounen (dismissed):
I have some minor questions, but it looks like the new work is protected by a feature flag and ready for iteration 🎉

damassi (approved):
Amazing work on this @erikdstock 👍 Looks much much better. 

I have some suggestions about moving just about _all_ of the state into a shipping form `useReducer` that's connected to the new context, so that any bit of usage is centrally controlled and contained, but that can be in a follow-up pr.  

Lets chat about it once this is merged 👍

src/Apps/__tests__/Fixtures/Order.ts
```diff
@@ -49,6 +49,7 @@ const OrderArtworkNodeWithoutShipping = {
     href: "/partner/a-g",
   },
   pickup_available: true,
+  pickupAvailable: true,
```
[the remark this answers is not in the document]
erikdstock: A test was failing and I traced it to this. I will double check and remove if it's not necessary.

src/Components/CountrySelect.tsx
```diff
@@ -319,10 +320,13 @@ const EU_COUNTRY_SELECT_OPTIONS = [
...
 ]
 
-export const ALL_COUNTRY_SELECT_OPTIONS = [
-  ...COUNTRY_SELECT_OPTIONS,
-  ...EU_COUNTRY_SELECT_OPTIONS,
-]
+export const ALL_COUNTRY_SELECT_OPTIONS = uniqBy(
+  [...COUNTRY_SELECT_OPTIONS, ...EU_COUNTRY_SELECT_OPTIONS],
+  "value"
+)
```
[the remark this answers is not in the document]
erikdstock: Yes, the EU countries are included in the country select options. We also have a package that includes different country lists - we don't use these different lists consistently, so I did this for now to make sure we had the most complete list possible.

src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
erikdstock: Moved to the bottom for now. I'm hesitant to use the shippingUtils file for things that aren't shared across multiple components at this point but open to leaning on it more later or others making that decision.

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -0,0 +1,1053 @@
...
+  props: ShippingProps
+): {
+  savedOrderData: ReturnType<typeof useLoadComputedData>
+  initialValues: {
+    fulfillmentDetails: FulfillmentValues
+  }
+  targetStep: ShippingRouteStep
+} => {
+  const orderData = useLoadComputedData(props)
+  const initialValues = getInitialValues(props, orderData)
```
erikdstock, line 792: This hook (hook so it can take advantage of other hooks) loads all the initial order into a shape that we might decide to change. I tried to split things up logically - savedOrderData ends up in the context for other orders, similar to how we use a lot of helper functions in our [current shippingUtils file](https://github.com/artsy/force/blob/dcc35900f3e9e8fb291cd6779b7f993d689544f8/src/Apps/Order/Utils/shippingUtils.ts). This just makes it a little more clear and lets those values change more predictably when the relay order changes. So this last 300 lines or so are pretty important.

erikdstock (commented):
Assorted comments that never sent. This PR is ready for squash and merge.

src/Apps/Order/Routes/Shipping2/index.tsx
damassi: Doing it in this PR (which would be very quick to implement) could be a quick way to clean up a lot of secondary noise that's sprinkled throughout. 

Def not blocking, just a suggestion to avoid follow-ups for trivial (yet valuable) readability stuff in such a complex PR that could likely be returned to as a reference.

erikdstock:
> I have some suggestions about moving just about _all_ of the state into a shipping form `useReducer` that's connected to the new context, so that any bit of usage is centrally controlled and contained, but that can be in a follow-up pr.
> 
> Lets chat about it once this is merged 👍

Merging now 🚀 . 

I'm totally down for talking about useReducer and other enhancements. It's been a really long time since I had to think in those terms (pre-hooks) so I'm not in the best position to take that up on my own. In a world without priorities I'd ask people to just pick an area of logic in here and check their understanding/look for improvements.

PR #13036, opened 2023-10-17
fix(EMI-1377): default to address form jump

rquartararo:

This PR solves [EMI-1377]

### Description

Follow up to https://github.com/artsy/force/pull/13021

@iskounen Pointed out that we want to default to scroll to the address form instead of the phone number input in the case where both of these are invalid. 

[EMI-1377]: https://artsyproduct.atlassian.net/browse/EMI-1377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping/index.tsx
+++ b/src/Apps/Order/Routes/Shipping/index.tsx
@@ -274,16 +274,16 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       hasError: invalidPhoneNumber,
     } = validatePhoneNumber(phoneNumber)
 
+    if (invalidPhoneNumber) {
+      setPhoneNumberError(phoneNumberError!)
+      setPhoneNumberTouched(true)
+      !invalidAddress && jumpTo("phoneNumberTop", { behavior: "smooth" })
+    }
+
     if (invalidAddress) {
       setAddressErrors(addressErrors!)
       setAddressTouched(touchedAddress)
       jumpTo("deliveryAddressTop", { behavior: "smooth" })
-    }
-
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Shipping/index.tsx
```diff
@@ -274,18 +274,18 @@ export const ShippingRoute: FC<ShippingProps> = props => {
       hasError: invalidPhoneNumber,
     } = validatePhoneNumber(phoneNumber)
 
+    if (invalidPhoneNumber) {
+      setPhoneNumberError(phoneNumberError!)
+      setPhoneNumberTouched(true)
+      !invalidAddress && jumpTo("phoneNumberTop", { behavior: "smooth" })
```
iskounen, line 280: This works, but the logic may be easier to read if instead of adding a `!invalidAddress &&` here, you use an if-else statement to decide where to jump to:

```
if (invalidAddress) {
  setAddressErrors(addressErrors!)
  setAddressTouched(touchedAddress)
  jumpTo("deliveryAddressTop", { behavior: "smooth" })
} else if (invalidPhoneNumber) {
  setPhoneNumberError(phoneNumberError!)
  setPhoneNumberTouched(true)
  jumpTo("phoneNumberTop", { behavior: "smooth" })
}
```

iskounen: Oh, never mind. Then the state hooks won't all get updated.

iskounen (approved):

PR #10379, opened 2022-06-23
feat(FX-4027): Add keyword filter to artist page's artwork grid

nickskalkin:
The type of this PR is: **Feature**

This PR solves [FX-4027]

### Description

As a collector
When I visit the artist page’s artwork grid
I see a new keyword filter
And when I enter a keyword
the grid is filtered down to match that query

I've created a feature toggle in Unleash, it's called `artist-artwork-grid-keyword-search` and currently enabled for development and staging

### Demo

* [Demo app](https://artwork-grid-keyword-filter.artsy.net)
* [Banksy grid](https://artwork-grid-keyword-filter.artsy.net/artist/banksy/works-for-sale?sort=-default_trending_score)

https://user-images.githubusercontent.com/3934579/175893281-c8b0e340-644b-4087-9d80-0496e926e4dd.mp4

[FX-4027]: https://artsyproduct.atlassian.net/browse/FX-4027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13052, opened 2023-10-20
test: extract shared RTL patterns for artwork filter specs

anandaroop:

Decided to split this from https://github.com/artsy/force/pull/13041#discussion_r1366195014 so that I can get this on `main` and then include it in that PR.

### Description

This takes some patterns for working with artwork filters in a react-testing-library friendly way and extracts them into shared helpers.

This was [introduced](https://github.com/artsy/force/pull/7629#discussion_r639742956) in the original RTL spike for the `SizeFilter` spec, [copied over](https://github.com/artsy/force/pull/10379) to the `KeywordFilter` spec and now will be [used](https://github.com/artsy/force/pull/13041/commits/686e2e5ed163b117f6ab35f015c250e604d50f3b) for the `ArtistSeriesFilter` spec as well.

Seems like a good time to extract and document this shareable code.

### DIFF
--- a/src/Components/ArtworkFilter/ArtworkFilters/__tests__/Utils.tsx
+++ b/src/Components/ArtworkFilter/ArtworkFilters/__tests__/Utils.tsx
@@ -1,0 +1,149 @@
+import {
+  RenderOptions,
+  render as originalRender,
+  screen,
+} from "@testing-library/react"
+import {
+  ArtworkFilterContextProps,
+  ArtworkFilterContextProvider,
+  useArtworkFilterContext,
+} from "Components/ArtworkFilter/ArtworkFilterContext"
+import { ReactElement } from "react"
+
+/**
+ * A test helper that can be used to render components that depend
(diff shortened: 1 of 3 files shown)


src/Components/ArtworkFilter/ArtworkFilters/__tests__/Utils.tsx
```diff
@@ -0,0 +1,149 @@
...
+import {
+  ArtworkFilterContextProps,
+  ArtworkFilterContextProvider,
+  useArtworkFilterContext,
+} from "Components/ArtworkFilter/ArtworkFilterContext"
+import { ReactElement } from "react"
+
+/**
+ * A test helper that can be used to render components that depend
+ * on an enclosing `ArtworkFilterContext`.
```
anandaroop, line 15: Tried to get some good intellisense going…

Screenshot 2023-10-20 at 11 28 31 AM

anandaroop:
Going for it, so I can keep #13041 moving…

PR #13060, opened 2023-10-24
fix(EMI-1336): don't show cookie banner for Eigen

leamotta:

This PR solves [EMI-1336]

### Description

Cookie consent banner shows up when opening the checkout flow in the app. Introducing a `isEigen` check to avoid showing it in those scenarios.
Note: I found an Eigen check [in the past implementation](https://github.com/artsy/force/pull/12280/files#diff-62f5fc92583c63dc0ef6698014cb81e6d7db0a22c0bf70bd599482fd876785deL7)

[EMI-1336]: https://artsyproduct.atlassian.net/browse/EMI-1336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi (approved):

gkartalis (approved):
great catch! Will add a case around that in the Mobile QA

PR #13061, opened 2023-10-24
Refactor [EMI-1513]: Shipping2 address verification

erikdstock:

This PR continues the shipping route refactor by fixing most address verification specs. It does _not_ include the saved addresses specs because they are complicated by other factors around testing shipping quotes, but the functionality should be there.

### DIFF
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -268,7 +268,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
             <Button
               // prevent a parent formik context from treating this as a submit button
               type="button"
-              onClick={e => {
+              onClick={() => {
                 handleCloseModal({
                   label: "Back to Edit",
                   subject: modalTitle!,
@@ -285,7 +285,7 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
               onClick={e => {
                 // Because this is a type="submit" button, it will trigger the formik submit handler
                 // if it is inside a form. This is not the case here. We want to make this behavior
-                // more explicitly handled by the parent context.
+                // more explicitly handled by this component's `chooseAddress()` logic.
                 e.preventDefault()
                 if (selectedAddressKey) {
                   trackSelectedAddress({
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -283,9 +283,10 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
             <Button
               disabled={!(selectedAddressKey && selectedAddressKey.length > 0)}
               onClick={e => {
-                // Because this is a type="submit" button, it will trigger the formik submit handler
-                // if it is inside a form. This is not the case here. We want to make this behavior
-                // more explicitly handled by this component's `chooseAddress()` logic.
+                // Because a <button> is treated as type="submit" by default, it
+                // will trigger the formik submit handler if it is inside a formik
+                // <Form />. We want to make this behavior more explicitly handled
+                // by this component's `chooseAddress()` logic.
                 e.preventDefault()
                 if (selectedAddressKey) {
                   trackSelectedAddress({
--- a/src/Apps/Order/Components/AddressVerificationFlow.tsx
+++ b/src/Apps/Order/Components/AddressVerificationFlow.tsx
@@ -266,7 +266,9 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
           <Spacer y={4} />
           <Flex width="100%" justifyContent="space-between">
             <Button
-              onClick={() => {
+              // prevent a parent formik context from treating this as a submit button
+              type="button"
+              onClick={e => {
                 handleCloseModal({
                   label: "Back to Edit",
                   subject: modalTitle!,
@@ -280,7 +282,11 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
             <Spacer x={1} />
             <Button
               disabled={!(selectedAddressKey && selectedAddressKey.length > 0)}
-              onClick={() => {
+              onClick={e => {
+                // Because this is a type="submit" button, it will trigger the formik submit handler
+                // if it is inside a form. This is not the case here. We want to make this behavior
+                // more explicitly handled by the parent context.
+                e.preventDefault()
                 if (selectedAddressKey) {
                   trackSelectedAddress({
                     subject: modalTitle!,
--- a/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
+++ b/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
@@ -95,7 +95,9 @@ export const AddressModal: React.FC<Props> = ({
     null
   )
   const [showDialog, setShowDialog] = useState<boolean>(false)
-  const { shipsFrom } = useShippingContext()
+  const {
+    savedOrderData: { shipsFrom },
+  } = useShippingContext()
   if (!relayEnvironment) return null
   if (!modalAction) return null
 
--- a/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
+++ b/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
@@ -95,9 +95,7 @@ export const AddressModal: React.FC<Props> = ({
     null
   )
   const [showDialog, setShowDialog] = useState<boolean>(false)
-  const {
-    savedOrderData: { shipsFrom },
-  } = useShippingContext()
+  const shippingContext = useShippingContext()
   if (!relayEnvironment) return null
   if (!modalAction) return null
 
@@ -105,7 +103,11 @@ export const AddressModal: React.FC<Props> = ({
   const initialAddress =
     modalAction.type === "editUserAddress"
       ? modalAction.address
-      : { country: shipsFrom, internalID: undefined, isDefault: false }
+      : {
+          country: shippingContext.savedOrderData.shipsFrom,
+          internalID: undefined,
+          isDefault: false,
+        }
 
   const handleModalClose = () => {
     closeModal()
--- a/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
+++ b/src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
@@ -104,7 +104,7 @@ export const AddressModal: React.FC<Props> = ({
     modalAction.type === "editUserAddress"
       ? modalAction.address
       : {
-          country: shippingContext.savedOrderData.shipsFrom,
+          country: shippingContext.computedOrderData.shipsFrom,
           internalID: undefined,
           isDefault: false,
         }
--- a/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
+++ b/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
@@ -27,7 +27,6 @@ import {
   Formik,
   FormikErrors,
   FormikHelpers,
-  FormikProps,
   FormikTouched,
   useFormikContext,
 } from "formik"
@@ -49,11 +48,11 @@ import { RouterLink } from "System/Router/RouterLink"
 
 export const ADDRESS_VALIDATION_SHAPE = {
   addressLine1: Yup.string().required("Street address is required"),
-  addressLine2: Yup.string(),
+  addressLine2: Yup.string().nullable(),
   city: Yup.string().required("City is required"),
   postalCode: postalCodeValidator,
   region: Yup.string().when("country", {
-    is: country => country === "US",
+    is: country => ["US", "CA"].includes(country),
     then: Yup.string().required("State is required"),
     otherwise: Yup.string(),
   }),
@@ -97,43 +96,19 @@ export interface ShipValues {
 export type FulfillmentValues = ShipValues | PickupValues
 
 export interface FulfillmentDetailsFormProps {
-  active: boolean
   onSubmit: (
     values: FulfillmentValues,
     formikHelpers?: FormikHelpers<FulfillmentValues>
   ) => void | Promise<any>
-  setFulfillmentFormHelpers: React.Dispatch<
-    React.SetStateAction<
-      Pick<
-        FormikProps<FulfillmentValues>,
-        "handleSubmit" | "isValid" | "values"
-      >
-    >
-  >
-  initialFulfillmentValues: FulfillmentValues
+
   me: FulfillmentDetailsForm_me$data
   order: FulfillmentDetailsForm_order$data
-  renderMissingShippingQuotesError: boolean
 }
 
 type AddressFormMode = "saved_addresses" | "new_address" | "pickup"
 
-export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
-  active,
-  initialFulfillmentValues,
-  ...props
-}) => {
-  // // Current we do this from index (to refresh shipping quotes)
-  // const {
-  //   fulfillmentType: savedFulfillmentType,
-  //   fulfillmentDetails: savedFulfillmentDetails,
-  // } = useShippingContext()
-  // useEffect(() => {
-  //   if (savedFulfillmentType === FulfillmentType.SHIP) {
-  //     props.onSubmit(savedFulfillmentDetails)
-  //   }
-  // // eslint-disable-next-line react-hooks/exhaustive-deps
-  // }, [])
+export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
+  const { initialValues } = useShippingContext()
 
   const addressVerificationUSEnabled = !!useFeatureFlag(
     "address_verification_us"
@@ -153,19 +128,14 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
       values.fulfillmentType === FulfillmentType.SHIP &&
       !hasSavedAddresses &&
       enabledForAddress &&
-      addressVerificationCount < 1
+      values.attributes.addressVerifiedBy === null
     )
-    // && addressFormMode === "new_address"
   }
 
   // trigger address verification by setting this to true
   const [verifyAddressNow, setVerifyAddressNow] = useState<boolean>(false)
-  const [addressVerificationCount, setAddressVerificationCount] = useState<
-    number
-  >(0)
 
   const handleVerificationComplete = async () => {
-    setAddressVerificationCount(addressVerificationCount + 1)
     setVerifyAddressNow(false)
   }
 
@@ -180,19 +150,13 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
 
   return (
     <Formik<FulfillmentValues>
-      initialValues={initialFulfillmentValues}
+      initialValues={initialValues.fulfillmentDetails}
       validationSchema={VALIDATION_SCHEMA}
       onSubmit={handleSubmit}
     >
       <FulfillmentDetailsFormLayout
-        renderMissingShippingQuotesError={
-          props.renderMissingShippingQuotesError
-        }
         order={props.order}
-        onSubmit={props.onSubmit}
         me={props.me}
-        setFulfillmentFormHelpers={props.setFulfillmentFormHelpers}
-        active={active}
         verifyAddressNow={verifyAddressNow}
         addressVerificationComplete={handleVerificationComplete}
       />
@@ -309,29 +273,40 @@ export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
 interface LayoutProps
   extends Pick<
     FulfillmentDetailsFormProps,
-    | "order"
-    | "me"
-    | "setFulfillmentFormHelpers"
-    | "active"
-    | "renderMissingShippingQuotesError"
-    | "onSubmit"
+    "order" | "me"
+    // | "onSubmit"
   > {
   verifyAddressNow: boolean
   addressVerificationComplete: () => Promise<void>
 }
 
 const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
-  const { active, onSubmit: submitWithoutFormik } = props
+  const {
+    step,
+    savedOrderData: {
+      isArtsyShipping,
+      shippingQuotes,
+      lockShippingCountryTo,
+      fulfillmentType: savedFulfillmentType,
+      selectedSavedAddressId: savedSelectedAddressId,
+    },
+    helpers: {
+      fulfillmentDetails: { setFulfillmentFormHelpers },
+    },
+  } = useShippingContext()
+  const active = step === "fulfillment_details"
+
+  const renderMissingShippingQuotesError = !!(
+    isArtsyShipping &&
+    shippingQuotes &&
+    shippingQuotes.length === 0
+  )
+
   const firstArtwork = extractNodes(props.order.lineItems)[0]!.artwork!
 
   const savedAddresses = compact(
     extractNodes(props.me?.addressConnection) ?? []
   )
-  const {
-    lockShippingCountryTo,
-    fulfillmentType: savedFulfillmentType,
-    selectedSavedAddressId: savedSelectedAddressId,
-  } = useShippingContext()
 
   const availableFulfillmentTypes: FulfillmentType[] = firstArtwork.pickupAvailable
     ? [FulfillmentType.PICKUP, FulfillmentType.SHIP]
@@ -353,7 +328,7 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
   // Pass some key formik bits up to the shipping route
   useEffect(() => {
     if (active) {
-      props.setFulfillmentFormHelpers({
+      setFulfillmentFormHelpers({
         handleSubmit: handleSubmit,
         isValid: isValid,
         values: values,
@@ -378,7 +353,7 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
     }
     await setValues(newValues)
     await props.addressVerificationComplete()
-    submitWithoutFormik(newValues)
+    formikContext.submitForm()
   }
 
   const addressFormMode: AddressFormMode =
@@ -490,7 +465,7 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
           <Text variant="lg-display" mb="1">
             Delivery address
           </Text>{" "}
-          {props.renderMissingShippingQuotesError && (
+          {renderMissingShippingQuotesError && (
             <ArtaMissingShippingQuoteMessage />
           )}
           {/* SAVED ADDRESSES */}
--- a/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
+++ b/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
@@ -281,9 +281,9 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
   const active = shippingContext.step === "fulfillment_details"
 
   const renderMissingShippingQuotesError = !!(
-    shippingContext.savedOrderData.isArtsyShipping &&
-    shippingContext.savedOrderData.shippingQuotes &&
-    shippingContext.savedOrderData.shippingQuotes.length === 0
+    shippingContext.computedOrderData.isArtsyShipping &&
+    shippingContext.computedOrderData.shippingQuotes &&
+    shippingContext.computedOrderData.shippingQuotes.length === 0
   )
 
   const firstArtwork = extractNodes(props.order.lineItems)[0]!.artwork!
@@ -470,9 +470,9 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
             data-testid="addressFormCollapse"
             open={
               addressFormMode === "new_address" ||
-              (shippingContext.savedOrderData.fulfillmentType ===
+              (shippingContext.computedOrderData.fulfillmentType ===
                 FulfillmentType.SHIP &&
-                !shippingContext.savedOrderData.selectedSavedAddressId)
+                !shippingContext.computedOrderData.selectedSavedAddressId)
             }
           >
             <GridColumns>
@@ -508,22 +508,22 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
                     setFieldValue(`attributes.country`, selected)
                   }
                   disabled={
-                    !!shippingContext.savedOrderData.lockShippingCountryTo &&
-                    shippingContext.savedOrderData.lockShippingCountryTo !==
+                    !!shippingContext.computedOrderData.lockShippingCountryTo &&
+                    shippingContext.computedOrderData.lockShippingCountryTo !==
                       "EU"
                   }
                   euShippingOnly={
-                    shippingContext.savedOrderData.lockShippingCountryTo ===
+                    shippingContext.computedOrderData.lockShippingCountryTo ===
                     "EU"
                   }
                   data-testid="AddressForm_country"
                 />
-                {shippingContext.savedOrderData.lockShippingCountryTo && (
+                {shippingContext.computedOrderData.lockShippingCountryTo && (
                   <>
                     <Spacer x={0.5} y={0.5} />
                     <Text variant="xs" color="black60">
-                      {shippingContext.savedOrderData.lockShippingCountryTo ===
-                      "EU"
+                      {shippingContext.computedOrderData
+                        .lockShippingCountryTo === "EU"
                         ? "Continental Europe shipping only."
                         : "Domestic shipping only."}
                     </Text>
--- a/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
+++ b/src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
@@ -108,7 +108,7 @@ export interface FulfillmentDetailsFormProps {
 type AddressFormMode = "saved_addresses" | "new_address" | "pickup"
 
 export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
-  const { initialValues } = useShippingContext()
+  const shippingContext = useShippingContext()
 
   const addressVerificationUSEnabled = !!useFeatureFlag(
     "address_verification_us"
@@ -135,7 +135,7 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
   // trigger address verification by setting this to true
   const [verifyAddressNow, setVerifyAddressNow] = useState<boolean>(false)
 
-  const handleVerificationComplete = async () => {
+  const handleVerificationComplete = () => {
     setVerifyAddressNow(false)
   }
 
@@ -150,7 +150,7 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
 
   return (
     <Formik<FulfillmentValues>
-      initialValues={initialValues.fulfillmentDetails}
+      initialValues={shippingContext.initialValues.fulfillmentDetails}
       validationSchema={VALIDATION_SCHEMA}
       onSubmit={handleSubmit}
     >
@@ -158,7 +158,7 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
         order={props.order}
         me={props.me}
         verifyAddressNow={verifyAddressNow}
-        addressVerificationComplete={handleVerificationComplete}
+        onAddressVerificationComplete={handleVerificationComplete}
       />
     </Formik>
   )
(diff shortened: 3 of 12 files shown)


src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -46,13 +46,18 @@ import {
   ShippingAddressFormValues,
 } from "Apps/Order/Routes/Shipping2/shippingUtils"
 import { RouterLink } from "System/Router/RouterLink"
+import { log, trace } from "console"
```
erikdstock: before I remove this: I learned that importing `log` from console can help when jest for some reason isn't logging your logs.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -134,23 +138,40 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
...
+    number
+  >(0)
+
+  const handleVerificationComplete = async () => {
+    setAddressVerificationCount(addressVerificationCount + 1)
+    setVerifyAddressNow(false)
+  }
 
   const handleSubmit = (values, helpers) => {
     if (shouldVerifyAddressOnSubmit(values)) {
```
erikdstock, line 173: This section is the core of the functional changes to make address verification work right. `hasSavedAddresses` is now calculated twice in this file - it might be worth adding it (or the saved addresses) to the Shipping Context (bottom of index file) but because that depends on a relay object i hesitated so far.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -393,12 +433,27 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
...
+              "city",
+              "region",
+              "postalCode",
+              "country",
+            ]) as ShipValues["attributes"]
+          }
+          onClose={handleCloseVerification}
+          onChosenAddress={handleChooseAddress}
+        />
+      )}
```
erikdstock, line 456: Moved because the modal doesn't need children.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -198,13 +200,17 @@ const verifyAddressWithSuggestions = async (relayEnv, input, suggested) => {
...
+
   const verificationResult = {
     __typename: "VerifyAddressType",
     verificationStatus: "VERIFIED_WITH_CHANGES",
     suggestedAddresses: [
       {
-        lines: addressLines(recommendedAddress),
-        address: recommendedAddress,
+        lines: addressLines(suggestedAddress),
+        address: suggestedAddress,
```
erikdstock, line 213: possibly unnecessary optimization - the result could always just be input address + line1: "suggested-change"

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -669,7 +678,7 @@ describe("Shipping", () => {
           await saveAndContinue()
 
           expect(
-            screen.queryByText("This field is required")
+            screen.queryByText(/[\w\s]is required/)
```
erikdstock, line 681: required messages are no customized per field

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -728,19 +737,23 @@ describe("Shipping", () => {
...
             await userEvent.click(screen.getByText("Save and Continue"))
 
-            expect(screen.getByText("This field is required")).toBeVisible()
-            expect(env.mock.getAllOperations()).toHaveLength(0)
+            await waitFor(() => {
+              expect(
+                screen.getByText("Street address is required")
+              ).toBeVisible()
+              expect(env.mock.getAllOperations()).toHaveLength(0)
+            })
```
erikdstock, line 753: `await waitFor(() => somethingAsyncThatMightFail)` is a good first choice before `flushPromiseQueue`

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -781,13 +794,15 @@ describe("Shipping", () => {
               recommendedAddress
             )
 
-            mockCommitMutation.mockResolvedValueOnce(
-              settingOrderShipmentSuccess
-            )
+            mockCommitMutation
+              // .mockResolvedValueOnce(settingOrderShipmentSuccess)
```
erikdstock, line 798: I think i commented this out while testing something else and forgot to put it back, but the tests pass. Maybe it is already mocked by default.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -781,13 +794,15 @@ describe("Shipping", () => {
...
+              .mockResolvedValueOnce(updateAddressSuccess)
             // Clicking "Use This Address" on verification modal automatically
             // sets shipping on the order and proceeds to the next step.
-            userEvent.click(screen.getByText("Use This Address"))
-            expect(mockCommitMutation).toHaveBeenCalledTimes(1)
+            await userEvent.click(screen.getByText("Use This Address"))
+            await waitFor(() => {
+              expect(mockCommitMutation).toHaveBeenCalledTimes(2)
+            })
```
erikdstock, line 805: It's possible this assertion had to change because the system is better at waiting now - the extra request is due to the user address being saved.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -829,29 +845,47 @@ describe("Shipping", () => {
...
                 shipping: {
-                  ...validAddress,
+                  ...initialAddress,
+                  city: "New York: the big apple",
                   phoneNumber: "",
                 },
               },
             })
+            const addressMutationArg = mockCommitMutation.mock.calls[1][0][1]
+            expect(addressMutationArg.mutation.default.operation.name).toEqual(
```
erikdstock, line 878: strange: The nesting or args changes between 1st and 2nd call. I don't know why but i am pretty sure i've seen this before in the same file.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -913,24 +949,37 @@ describe("Shipping", () => {
...
             ])
             await userEvent.click(screen.getByText("Save and Continue"))
 
-            const mutation = env.mock.getMostRecentOperation()
-            expect(mutation.request.node.operation.name).toEqual(
-              "AddressVerificationFlowQuery"
+            await verifyAddressWithSuggestions(env, inputAddress, {
+              ...inputAddress,
+              addressLine1: "<recommended change>",
+            })
```
erikdstock, line 961: As mentioned above this might be enough to not need to supply a recommended address to this helper.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -1211,6 +1266,8 @@ describe("Shipping", () => {
               recommendedAddress
             )
 
+            // TODO: Arta quotes don't appear in the updated order relay object.
+            // Do we need to do something to refresh the relay store in testing?
             mockCommitMutation
```
erikdstock, line 1271: Didn't fix address verification tests for saved addresses because it can't find the shipping quotes. Left note here for later but it could be a red herring.

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -304,14 +304,16 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
           requiresArtsyShipping = requiresArtsyShippingTo(addressValues.country)
+
           fulfillmentMutationValues = {
             id: props.order.internalID,
             fulfillmentType: requiresArtsyShipping
               ? "SHIP_ARTA"
               : FulfillmentType.SHIP,
             phoneNumber,
-            shipping: addressValues,
+            shipping: { ...addressValues, phoneNumber: "" },
```
erikdstock, line 314: fixed test that assumes we send phone number in two places, one empty. I looked at exchange and this seems safe to ignore/fix but didn't want to bundle that into other changes on this refactor.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -1089,6 +1090,16 @@ describe("Shipping", () => {
...
+      it.only("set shippinglike above with trcking", async () => {
+        renderWithRelay({
+          CommerceOrder: () => order,
+          Me: () => meWithAddresses,
+        })
+
+        await saveAndContinue()
+
+        expect(mockTrackEvent.mock.calls).toEqual({})
+      })
```
erikdstock, line 1102: forgot to remove this. I was discussing with @rquartararo how we might backfill tests for analytics that are missing as part of the shipping2 work - write a failing test, see what actually gets tracked, verify it makes sense, copy into old test, copy into new test.

erikdstock (commented):
Tour of changes, some comments.

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -266,7 +266,9 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
           <Spacer y={4} />
           <Flex width="100%" justifyContent="space-between">
             <Button
-              onClick={() => {
+              // prevent a parent formik context from treating this as a submit button
+              type="button"
+              onClick={e => {
```
damassi, line 271: Is this comment accurate? Without `e.preventDefault()` it should still behave as a button

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -280,7 +282,11 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
             <Spacer x={1} />
             <Button
               disabled={!(selectedAddressKey && selectedAddressKey.length > 0)}
-              onClick={() => {
+              onClick={e => {
+                // Because this is a type="submit" button, it will trigger the formik submit handler
```
damassi, line 286: This button isn't `type=submit` tho; the one up above is

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -164,16 +185,16 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
...
           props.renderMissingShippingQuotesError
         }
         order={props.order}
+        onSubmit={props.onSubmit}
         me={props.me}
         setFulfillmentFormHelpers={props.setFulfillmentFormHelpers}
         active={active}
         verifyAddressNow={verifyAddressNow}
-        addressVerificationComplete={() => setVerifyAddressNow(false)}
+        addressVerificationComplete={handleVerificationComplete}
```
damassi, line 197: `addressVerificationComplete` => `onAddressVerificationComplete`

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -164,16 +185,16 @@ export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = ({
       onSubmit={handleSubmit}
     >
       <FulfillmentDetailsFormLayout
-        onSubmit={props.onSubmit}
         renderMissingShippingQuotesError={
           props.renderMissingShippingQuotesError
         }
         order={props.order}
+        onSubmit={props.onSubmit}
```
damassi, line 192: is there any reason this can't be called from `handleSubmit`? Having two divergent `onSubmit` callbacks defined (one in formik props, one here)  is a bit confusing and i can imagine lead to subtle race conditions later. Can we roll this up?

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -291,16 +312,16 @@ interface LayoutProps
...
+    | "onSubmit"
   > {
   verifyAddressNow: boolean
-  addressVerificationComplete: () => void
+  addressVerificationComplete: () => Promise<void>
 }
 
 const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
-  const { active } = props
+  const { active, onSubmit: submitWithoutFormik } = props
```
damassi, line 324: The way this branches between the formik submit handler and here is a bit confusing

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -341,6 +362,25 @@ const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
...
+      attributes: {
+        ...values.attributes,
+        ...chosenAddress,
+        addressVerifiedBy: verifiedBy,
+      },
+    }
+    await setValues(newValues)
+    await props.addressVerificationComplete()
+    submitWithoutFormik(newValues)
+  }
```
damassi, line 382: Side note: It would be great if all of these function callbacks, vs being spread throughout components and passed down as props and calling to parents and so on (which creates a sort of tri-dimensional brittleness) could be rolled up into the context and centralized as actions, or stored within a hook. its not just state stuff that benefits from centralization, but also actions. We have access to formik values and actions from the useFormikContext hook. An action could be plucked from the form context and used directly within the query renderer, no props passing required. 

Basically, the more things that are passed as props, the more complex. The more one centralizes within the new Context and / or leans on hooks, the easier this will be able to change and modify in the future. 

If you're around I can help walk you through this setup; we have an opportunity to centralize 👍 (The best way to think about react in 2023 is: for every passed prop, subtract one point from the game 😄)

damassi (commented):
Left a few comments -- none blocking! But just notes and a nudge that there are opportunities here to simplify and consolidate.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
iskounen: Now that these messages are customized, I think we should use specific error messages (instead of a regex) in each test.

For example, here we want to make sure that neither "Postal code is required" nor "State/Region/Provice is required" is displayed.

src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx
```diff
@@ -684,7 +693,7 @@ describe("Shipping", () => {
           userEvent.type(name, "First Last")
           userEvent.clear(name)
 
-          expect(screen.getByText("This field is required")).toBeInTheDocument()
+          expect(screen.getByText(/[\w\s]is required/)).toBeInTheDocument()
```
iskounen, line 696: This is another example where I'd love to assert the customized error message. Only "Full name is required" should be in the document, so this would mask failures if "Postal code is required" is in the document instead.

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock: I think I responded to this point above (lacking a `type` attribute, it is treated as `submit`). Why preventDefault vs stopPropagation is what formik cares about I do not know.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: yes i agree and don't like this. it previously seemed impossible but let me take another look.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: I was following [what I understood the convention to be](https://stackoverflow.com/a/60048346/4409186), which is:
- expose props from your component the same way html does, with `on*`
- When making your own functions to pass as props, use `handle*`

So you might have a component that accepts an `onSubmit` prop, and passes that into a button: `<Button onClick={props.onSubmit}>Submit</Button>`, or you might have to add a little more to it:
```ts
const handleSubmit = () => {
}

File: src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx (hunk 19 above)
Comment (erikdstock) [reply src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx:382]: We talked about this in person and identified a few things to look at. Will update with some changes here.

File: src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx (hunk 5 above)
Comment (erikdstock) [reply src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx:681]: Sadly you are right. Will update.

File: src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx (hunk 11 above)
Comment (erikdstock) [reply src/Apps/Order/Routes/Shipping2/Shipping2.jest.tsx:1271]: Briefly talked with @damassi about this. No definite answer yet.

**Review by erikdstock — COMMENTED** [review]
Respond to feedback, fixing other bits now.

File: src/Apps/Order/Routes/Shipping2/ShippingContext.ts [hunk 21]
```diff
@@ -1,20 +1,342 @@
...
+      selectedShippingQuoteId: any
+    }
+  }
+  step: ShippingRouteStep
+  helpers: {
+    fulfillmentDetails: FulfillmentHelpers & {
+      setFulfillmentFormHelpers: (helpers: FulfillmentHelpers) => void
+    }
+  }
 }
```
erikdstock, line 54: Major change here is to the context, which is now much less stringent about what it includes (it includes a lot more, and other types are simpler). Thanks to this i've moved all context loading stuff from index.tsx to this file. I wonder if it is worth adding comments to give us more insight in our vscode type definitions on hover for some of them.

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -552,22 +529,11 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-        "fulfillmentType"
-      ),
-    [savedOrderData]
-  )
-
+  const maybeShippingQuotesConnectionEdges =
+    order.lineItems?.edges?.[0]?.node?.shippingQuoteOptions?.edges
+  const shippingQuotesConnectionEdges = maybeShippingQuotesConnectionEdges
+    ? compact(maybeShippingQuotesConnectionEdges)
+    : ([] as NonNullable<typeof maybeShippingQuotesConnectionEdges>)
```
erikdstock, line 536: a good amount of deletions in this file due to simplifying the context, but we still manually pull this stuff apart because the recipient of this prop on L534 is shared with the old shipping route.

erikdstock (commented):
Pushed 6feea3de173cc5bf9340ea998d493e868acc58f5 which makes our context a little more full of data. Ready for merge after i eat lunch.

src/Apps/Order/Components/AddressVerificationFlow.tsx
damassi: My point is that the comment is incorrect; mind updating?

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
```diff
@@ -95,7 +95,9 @@ export const AddressModal: React.FC<Props> = ({
     null
   )
   const [showDialog, setShowDialog] = useState<boolean>(false)
-  const { shipsFrom } = useShippingContext()
+  const {
+    savedOrderData: { shipsFrom },
+  } = useShippingContext()
```
damassi, line 100: It is helpful for understanding origins to avoid destructuring everywhere. It might feel cleaner and more fun to do this, but in complex code -- in the future -- its important to know where things originate from. Ie, 

```
const shippingContext = useShippingContext()
...
if (shippingContext.savedOrderData.shipsFrom) {
  ...
}
```
yields an enormous amount of contextual info. It says: data comes from context; data is from savedOrderData; and it is the shipsFrom field. Where `shipsFrom` alone says nothing at a glance. The more complex the code, the more important immediately-at-a-glance metadata like this is becomes, even if it might not feel that way as the author.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -93,64 +96,52 @@ export interface ShipValues {
...
-  //   fulfillmentDetails: savedFulfillmentDetails,
-  // } = useShippingContext()
-  // useEffect(() => {
-  //   if (savedFulfillmentType === FulfillmentType.SHIP) {
-  //     props.onSubmit(savedFulfillmentDetails)
-  //   }
-  // // eslint-disable-next-line react-hooks/exhaustive-deps
-  // }, [])
+export const FulfillmentDetails: FC<FulfillmentDetailsFormProps> = props => {
+  const { initialValues } = useShippingContext()
```
damassi, line 111: blocking: Here and elsewhere -- now that even more data is being stored in the context -- lets update all context-based call-sites to read:

```tsx
const shippingContext = useShippingContext()
...
foo(shippingContext.initialValues)
```

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -93,64 +96,52 @@ export interface ShipValues {
...
+      values.attributes.addressVerifiedBy === null
     )
-    // && addressFormMode === "new_address"
   }
 
   // trigger address verification by setting this to true
   const [verifyAddressNow, setVerifyAddressNow] = useState<boolean>(false)
 
+  const handleVerificationComplete = async () => {
+    setVerifyAddressNow(false)
```
damassi, line 139: There are no async operations inside of here -- thoughts?

src/Apps/Order/Components/AddressVerificationFlow.tsx
erikdstock:
```suggestion
                // Because a <button> is treated as type="submit" by default, it will trigger the formik submit handler
```

src/Apps/Order/Components/AddressVerificationFlow.tsx
```diff
@@ -280,7 +282,11 @@ const AddressVerificationFlow: React.FC<AddressVerificationFlowProps> = ({
...
+                // if it is inside a form. This is not the case here. We want to make this behavior
+                // more explicitly handled by this component's `chooseAddress()` logic.
```
erikdstock, line 288:
```suggestion
                // if it is inside a form. We want to make this behaviour more explicitly handled 
                // by this component's `chooseAddress()` logic.
```

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: alright, will do

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
damassi: Your explanation up above is totally correct and how it is done. 

However, the implementation in this PR is not following the guidance here and lacks appropriate naming metadata. 

For example, imagine you've never seen this code before. 

```
addressVerificationComplete
```
What comes  to mind here? Is it a `boolean` value? is it a function? Callback code that is passed **downward** in the react should always be prefixed with `on`, so that -- at a glance -- all ambiguity is reduced. 

Following the above exercise, lets start over. Imagine you're seeing this code for the first time: 

```
onAddressVerificationComplete
```
Is there any question whatsoever what this thing is? Its obviously a function callback. 

And so:

```
onAddressVerificationComplete={handleAddressVerificationComplete}
```

Local functions are `handleFoo`, which are passed props prefixed with `onFoo`; internal to the the child component, it then becomes clear what's going on without needed to read around all the code.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -288,29 +273,40 @@ export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
...
   extends Pick<
     FulfillmentDetailsFormProps,
-    | "order"
-    | "me"
-    | "setFulfillmentFormHelpers"
-    | "onSubmit"
-    | "active"
-    | "renderMissingShippingQuotesError"
+    "order" | "me"
+    // | "onSubmit"
```
damassi, line 277: 🐟

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -288,29 +273,40 @@ export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
...
+      isArtsyShipping,
+      shippingQuotes,
+      lockShippingCountryTo,
+      fulfillmentType: savedFulfillmentType,
+      selectedSavedAddressId: savedSelectedAddressId,
+    },
+    helpers: {
+      fulfillmentDetails: { setFulfillmentFormHelpers },
+    },
+  } = useShippingContext()
```
damassi, line 296: blocking: Avoid destructuring, especially at multiple levels deep. It is a trap, both in terms of understandability as well as safety: nested destructuring will blow up runtime code if some parent prop comes back null -- regardless of how it is typed. 

The above is a general comment, applied to javascript in general. But in terms of the shipping context -- just don't do it for the reasons discussed elsewhere 🙏

damassi: Additionally, this kind of stuff:
```
fulfillmentType: savedFulfillmentType
```
or 
```
selectedSavedAddressId: savedSelectedAddressId,
```
is complexity. Why? because it is 1:2. You have variable a, which is named `fullfilmentType`. But now, we have `savedFulfillmentType`, aliased to `fulfilmentType`. If it is _actually_ `savedFulfillmentType`, then change the name in the context, and keep things 1:1. 

Aliasing in general should be avoided, as a rule. If there are variable clashes going on,  there's an elegant solution right at ones fingertips, which both reduces complexity  and increases comprehensibility of a system:

```tsx
const shippingContext = useShippingContext()
const foo = "hi"

if (foo !== shippingContext.foo) {
  return null
}
```

src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
```diff
@@ -47,8 +47,7 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
     null
   )
   const {
-    selectedSavedAddressId,
-    availableShippingCountries,
+    savedOrderData: { availableShippingCountries, selectedSavedAddressId },
```
damassi, line 50: no destructuring

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  addressWithFallbackValues,
+  getDefaultUserAddress,
+  FulfillmentType,
+  ShippingAddressFormValues,
+} from "Apps/Order/Routes/Shipping2/shippingUtils"
+import { SavedAddressType } from "Apps/Order/Utils/shippingUtils"
+import { extractNodes } from "Utils/extractNodes"
+import { ALL_COUNTRY_CODES, EU_COUNTRY_CODES } from "Components/CountrySelect"
+import { COUNTRIES_IN_EUROPEAN_UNION } from "@artsy/commerce_helpers"
+// TODO: Duplicated list ^
```
damassi, line 24:
```
// FIXME: Duplicated list
import { COUNTRIES_IN_EUROPEAN_UNION } from "@artsy/commerce_helpers"
```

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+    availableShippingCountries: string[]
+    requiresArtsyShippingTo: (shipTo: string) => boolean
+    selectedSavedAddressId: string | null
+    fulfillmentDetails: FulfillmentValues["attributes"] | null
+    fulfillmentType: FulfillmentValues["fulfillmentType"] | null
+    isArtsyShipping?: boolean
+    shippingQuotes?: Array<{ id: string; isSelected: boolean }>
+  }
+  initialValues: {
+    fulfillmentDetails: FulfillmentValues
```
damassi, line 43: is it `fulfilmentDetails` or `FulfilmentValues`? 

Again, 1:2 complexity. 

Every new variable in this PR needs to be audited for 1:1 correctness 🙏 The only way we'll reduce complexity in the shipping route is by literally reducing complexity.

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+    shippingQuotes: {
+      selectedShippingQuoteId: any
+    }
+  }
+  step: ShippingRouteStep
+  helpers: {
+    fulfillmentDetails: FulfillmentHelpers & {
```
damassi, line 50: is it `fulfilmentDetails`, or `FulfulmentHelpers`?

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
 export const ShippingContext = createContext<ShippingContextProps>({} as any)
 
 export const useShippingContext = () => {
   return useContext(ShippingContext)
 }
+
+// Get information from the order and user, savedOrderData (shared, computed context about order)
+// initial values for forms, and the current step in the shipping route
+// values for forms
+export const useLoadOrder = (props: ShippingProps): ShippingContextProps => {
```
damassi, line 65: if we're not doing any actual "fetching", then we should rename this to be less indirect. `useOrder`, or `useOrderProps`, or `useComputedOrder`

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  const savedOrderData = useLoadComputedData(props)
```
damassi, line 66: ditto here. Is this just `orderData`? And are we just computing it? 

```
const orderData = useComputedOrderData(props)
```

"Load" = fetching = async = remote server. No fetch, no load in the name. 

But also be sure that there is a clear enough difference with this shipping context hook (which would ideally be its own file in `Shipping2/hooks`, along with any other complex hooks (like `useLoadComputedData`).

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  const initialValues = getInitialValues(props, savedOrderData)
+
+  const {
+    fulfillmentType: savedFulfillmentType,
```
damassi, line 70: 1:2 violation

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+
+    isArtsyShipping,
+  } = savedOrderData
```
damassi, line 73:
```tsx
if (savedOrderData.fulfulmentType === "foo") {  
  // metatdata FTW
}
```

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  )
+}
+
+const getSavedFulfillmentData = (
+  order: ShippingProps["order"],
+  me: ShippingProps["me"]
+): {
+  fulfillmentType: FulfillmentType
+  isArtsyShipping: boolean
+  fulfillmentDetails: FulfillmentValues["attributes"]
```
damassi, line 227: `fulfilmentValueAttributes` -> 1:1

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+      },
+    } as any,
+  })
+  return {
+    ...fulfillmentFormHelpers,
+    setFulfillmentFormHelpers,
+  }
+}
+
+const getInitialValues = (
```
damassi, line 129: There's so much stuff here in this file that requires tests. We need to create a folder and structure things appropriately: 
```
src/Apps/Order/Routes/Shipping2/ShippingContext/ShippingContext.tsx
src/Apps/Order/Routes/Shipping2/ShippingContext/utils/foo.ts
src/Apps/Order/Routes/Shipping2/ShippingContext/utils/__tests__/foo.test.ts
src/Apps/Order/Routes/Shipping2/ShippingContext/hooks/useFoo.ts
src/Apps/Order/Routes/Shipping2/ShippingContext/hooks/__tests__/useFoo.test.ts
```

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  props: ShippingProps,
+  orderData: ShippingContextProps["savedOrderData"]
+): ShippingContextProps["initialValues"] => {
+  const { me } = props
+  const {
+    fulfillmentType: savedFulfillmentType,
+    fulfillmentDetails: savedFulfillmentDetails,
```
damassi, line 136: 1:2

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+  } = orderData
+  const selectedShippingQuote =
+    (shippingQuotes && shippingQuotes.find(quote => quote.isSelected)) || null
+
+  const initialShippingQuotes = {
+    selectedShippingQuoteId: selectedShippingQuote?.id,
+  }
+  if (savedFulfillmentType) {
+    return {
+      shippingQuotes: initialShippingQuotes,
```
damassi, line 147: 1:2

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+      fulfillmentDetails: {
+        fulfillmentType: savedFulfillmentType,
```
damassi, line 149: 1:2

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: oh yeah, you are right. the prop is named wrong.

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: this is preserved from the legacy route - i just noticed the duplicated collections (via [artsy/address_helper](https://github.com/artsy/address_helper/blob/main/lib/address_helper.rb)) and wanted to note it here. Ultimately I think we should consolidate on this but the comment is just calling out but preserving an existing inconsistency.

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: `fulfillmentDetails` refers to what the form/component is named, while `FulfillmentValues` is the interface of the form values

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: again, a type name, namespaced by the area the helpers come from (picturing a series of steps to complete). Happy to iterate on this over time though

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
damassi: if `FulfilmentValues` is the type for `fulfilmentDetails`, how does that make any sense? One doesn't typically write `foo: Bar`; one writes `foo: Foo` and `bar: Bar`, and this is particularly true in highly sensitive areas that deal with money.

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: yes i agree

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
damassi: Seeing the follow-up component below, this makes more sense 👍

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
damassi: This makes sense now 👍

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
damassi: I understand, but comments typically don't go below the line of code that one is referring to

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,342 @@
...
+      fulfillmentDetails,
+      selectedSavedAddressId,
+    }
+  }
+  // Should never happen. Log it?
+  return null
+}
+
+// Compute and memoize data from the saved order.
+export const useLoadComputedData = (
```
damassi, line 283: We should remove `load` from the name here as it is not actually fetching anything

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: The function is intended to calculate the values for the shipping context. It is now called `useComputeOrderContext()`. Similar changes in other places

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -286,31 +271,26 @@ export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
...
+  onAddressVerificationComplete: () => void
 }
 
 const FulfillmentDetailsFormLayout = (props: LayoutProps) => {
-  const { active } = props
+  const shippingContext = useShippingContext()
+  const active = shippingContext.step === "fulfillment_details"
+
+  const renderMissingShippingQuotesError = !!(
+    shippingContext.computedOrderData.isArtsyShipping &&
```
damassi, line 284: `computedOrderData` 👍 👍 now here we go

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
```diff
@@ -1,20 +1,338 @@
...
   return useContext(ShippingContext)
 }
+
+// Get information from the order and user, computedOrderData (shared, computed context about order)
+// initial values for forms, and the current step in the shipping route
+// values for forms
+export const useComputeOrderContext = (
+  props: ShippingProps
+): ShippingContextProps => {
+  const computedOrderData = useComputeOrderData(props)
```
damassi, line 67: 👍

damassi (approved):
This looks much better and is def an improvement 👍 Thank you so much for bearing with me on all of this feedback. 

Now that things are consolidated inside of the context and centralized here I think we have a solid place to launch from for further work.

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
damassi: Totally. I think, too, now that everything is in this single file all all immediately referenceable (without needing to jump all around) we'll be able to do some further organization and from there and then, once its just right, break it apart (if need be).

iskounen (approved):
I read your PR more to learn than to critique, but I think your ShippingContext succeeds in decluttering the Shipping route and making it easer to find where data comes from. Congrats on getting more specs to pass 🎉

src/Apps/Order/Routes/Shipping2/ShippingContext.ts
erikdstock: Ticketed in https://artsyproduct.atlassian.net/browse/EMI-1535.

PR #13078, opened 2023-10-28
feat(EMI-1512): add order tracking helper for new shipping route

rquartararo:

This PR solves [EMI-1512]

@artsy/emerald-devs 

### Description

@erikdstock assigning you as I will be out on Monday. 

This adds a tracking helper for the checkout flow, starting with the new shipping route. We will be adding order tracking calls for all other routes to the helper in follow-up tickets. 

[EMI-1512]: https://artsyproduct.atlassian.net/browse/EMI-1512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
+++ b/src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
@@ -13,13 +13,6 @@ import {
 import createLogger from "Utils/logger"
 import { SavedAddressItem } from "Apps/Order/Routes/Shipping2/SavedAddressItem2"
 import { extractNodes } from "Utils/extractNodes"
-import { useTracking } from "react-tracking"
-import {
-  ActionType,
-  ClickedShippingAddress,
-  ContextModule,
-  OwnerType,
-} from "@artsy/cohesion"
 import { themeGet } from "@styled-system/theme-get"
 
 import { useShippingContext } from "Apps/Order/Routes/Shipping2/ShippingContext"
@@ -29,6 +22,7 @@ import {
   addressWithFallbackValues,
   getDefaultUserAddress,
 } from "Apps/Order/Routes/Shipping2/shippingUtils"
+import { useOrderTracking } from "Apps/Order/Utils/useOrderTracking"
 
 export const NEW_ADDRESS = "NEW_ADDRESS"
 const PAGE_SIZE = 30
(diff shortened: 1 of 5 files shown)


src/Apps/Order/Routes/Shipping2/SavedAddresses2.tsx
```diff
@@ -68,6 +61,11 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
     selectedAddressID && getAddressByID(addressList, selectedAddressID)
   const selectedAddressPresent = !!selectedAddress
 
+  const {
+    clickedShippingAddress,
+    clickedAddNewShippingAddress,
+  } = useOrderTracking()
```
damassi, line 67: [blocking]: We should avoid destructuring here (and elsewhere in the PR) 🙏 I mentioned this in @erikdstock's  previous refactor PR, but for complex code that is in the process of a refactoring pass we need all of the helpful clarifying metadata we can get. Something like
```tsx
tracking.clickedShippingAddress()
```
tells us very precisely whats going on, where a bare invocation like 
```
clickedShippingAddress()
```
could mean many things at a glance, and future devs might get confused thinking its some kind of event handler that performs further operations. 

An alternative (that preserves destructuring) would be to rename the functions to 
```tsx
const {
  trackClickedShippingAddress
  trackAddNewShippingAddress
} = useOrderTracking() 
```

Though I would encourage the first approach over this.

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -158,14 +135,10 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-          action: ActionType.errorMessageViewed,
-          context_owner_type: OwnerType.ordersShipping,
-          context_owner_id: props.order.internalID,
-          title: "Invalid address",
-          message:
-            "There was an error processing your address. Please review and try again.",
+        errorMessageViewed({
           error_code: error.code,
+          title: "Missing information",
+          message: "Please complete all required fields.",
```
iskounen, line 141: **issue:** This title and message don't match what's displayed to the user in the error dialog.

src/Apps/Order/Utils/useOrderTracking.tsx
```diff
@@ -0,0 +1,98 @@
...
+          context_module: ContextModule.ordersShipping,
+          context_page_owner_type: "orders-shipping",
+          subject: newShippingQuoteId,
+          context_page_owner_id: contextPageOwnerId,
+        }
+
+        trackEvent(payload)
+      },
+
+      errorMessageViewed: ({
```
iskounen, line 46: **nitpick:** `displayedErrorMessage` may be a better name to match the other methods in this hook.

erikdstock: If we are not destructuring and saying something like `orderTracking.errorMessageViewed()` then this makes sense to me for nice intuitive mapping. No strong opinions though.

erikdstock (commented):
Others comments look good - thanks for taking this on! I think i would call the result of the hook `orderTracking` just to not confuse with the generic `useTracking()` hook. But that is super minor.

damassi (approved):
👍

PR #13082, opened 2023-10-30
fix(my-collection): Remove non-null assertion crasher

damassi:

### Description

Fixes a bug that was [noted in the bugs and feedback channel](https://artsy.slack.com/archives/C03N12SR0RK/p1698683435339269). 

The issue arose because we were using `!`, which assumes that the code there is always safe. We should never use `!` (we can use `?` instead); but even so, best to guard and exit early.  As a general rule, UI not appearing is (data is null) is better than a crash.


olerichter00 (commented):
Thanks for the fix 🙏

PR #13083, opened 2023-10-30
feat(eslint): Enable no-non-null-assertion rule

damassi:

### Description

Thank you @gkartalis for the suggestion to add this to our `estlint` 🙏 Looking at our eslint file, found that it was disabled, probably due to some migration at some point. Turned it back on!

Screenshot 2023-10-30 at 10 00 05 AM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -90,7 +90,7 @@ module.exports = {
     "react/no-unescaped-entities": 0,
     "@typescript-eslint/no-empty-function": 0,
     "@typescript-eslint/no-empty-interface": 0,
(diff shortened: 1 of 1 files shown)


gkartalis (approved):
woohoooo!

damassi:
Lol, it really can be SO EASY @gkartalis 😭

gkartalis:
Oh wait, just saw that in eigen there is another one https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-asserted-optional-chain.md 

didn't know about that, maybe we should enable both?

PR #13085, opened 2023-10-31
feat(dia-295): filter q/a

dzucconi:
Re: [DIA-295](https://artsyproduct.atlassian.net/browse/DIA-295)
[Q/A board](https://www.notion.so/artsy/14381bc3646e40a6a92ac9b1a0776960?v=6c29026ec39b41b6941567d3a688c589)

This tackles all the issues sans 2:
* When I check “artists you follow,” “Andy Warhol” appears as a filter pill
  * This has been in production since perhaps when this feature was introduced.
* Bug with the two instances of the price range filter keeping in sync
  * Going to PR that one separately because I do not understand why it's doing what it was doing in the first place.

[DIA-295]: https://artsyproduct.atlassian.net/browse/DIA-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/ArtworkFilter/index.tsx
+++ b/src/Components/ArtworkFilter/index.tsx
@@ -426,7 +426,7 @@ export const BaseArtworkFilter: React.FC<
                   </ArtworkFilterCreateAlert>
 
                   <Pill Icon={FilterIcon} size="small" onClick={handleOpen}>
-                    All filters
+                    All Filters
                   </Pill>
 
                   <ArtworkFiltersQuick />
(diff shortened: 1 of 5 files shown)


src/Components/ArtworkFilter/index.tsx
```diff
@@ -461,7 +497,7 @@ export const BaseArtworkFilter: React.FC<
 
             {children || (
               <ArtworkFilterArtworkGrid
-                filtered_artworks={viewer.filtered_artworks!}
```
damassi, line 464: Props @gkartalis

PR #13100, opened 2023-11-02
fix(pricerange): respond to external value changes; adds debounced callback

dzucconi:
Closes [DIA-309](https://artsyproduct.atlassian.net/browse/DIA-309)

The problem we were seeing was that, if you updated the price range within the quick filter dropdown, then opened up the all filters drawer, that price range filter wasn't in sync with the state.

This is because the price range component keeps local state and only sets it based on initial props, not responding to any props updates. We can't make it a controlled component because the artwork filter context is naive, firing off API requests on every change, the slider dragging is chatty, and I have no desire to touch the filter code.

So, this PR:
* Adds an `onDebouncedUpdate` internally to the slider so that we can just use it since that's generally what you want with this type of input
* Responds to external updates to the value that's passed in to update its local state
* Updates all call sites and removes extraneous debouncing.

[DIA-309]: https://artsyproduct.atlassian.net/browse/DIA-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Form/PriceRange.tsx
+++ b/src/Components/Alert/Components/Form/PriceRange.tsx
@@ -1,6 +1,5 @@
 import { Expandable, Spacer } from "@artsy/palette"
 import { FC } from "react"
-
 import { useAlertContext } from "Components/Alert/Hooks/useAlertContext"
 import { PriceRange } from "Components/PriceRange/PriceRange"
 import {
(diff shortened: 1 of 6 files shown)


src/Components/Alert/Components/Form/PriceRange.tsx
```diff
@@ -34,7 +33,7 @@ export const PriceRangeFilter: FC<PricaRangeFilterProps> = ({
 
       <PriceRange
         priceRange={state.criteria.priceRange ?? DEFAULT_PRICE_RANGE}
-        onPriceRangeUpdate={handlePriceRangeUpdate}
+        onUpdate={handlePriceRangeUpdate}
```
dzucconi, line 36: These alert ranges DO debounce updates to the context apparently, so we use the un-debounced callback, so that we don't have stacked debouncing.

PR #13103, opened 2023-11-02
fix(artist): corrects empty condition; guards on extraneous fetches of subcomponents

dzucconi:
Closes [DIA-311](https://artsyproduct.atlassian.net/browse/DIA-311)

Few issues uncovered here:
* One of the conditions for rendering the empty state was checking for featured artworks, which was removed, so if an artist had any artworks but nothing else, the page would appear empty
* We did not ever wrap any of these sub-components in the empty check which is two problems: unsure what to remove if the sub-components change and we were fetching for _all_ of those sub-components, even when we knew they weren't there. That lead to an inaccurate skeleton state for the page and extraneous requests to Metaphysics.
* The last two components on the page didn't have conditions factored into the empty state, so on artists where we might have related artists + related categories, we wouldn't display anything.

[DIA-311]: https://artsyproduct.atlassian.net/browse/DIA-311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jonallured (dismissed):
nice one 🙏

PR #13106, opened 2023-11-06
feat: Add Medium Alert Filter

olerichter00:
Addresses [ONYX-480]

## Description

This PR adds the Medium filter to the new Alert Filter modal.

It also moves the filter logic, which we would probably need to duplicate between the different filters, into a base filter component.

Screenshot 2023-11-07 at 13 17 13

[ONYX-480]: https://artsyproduct.atlassian.net/browse/ONYX-480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
+++ b/src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
@@ -1,0 +1,73 @@
+import { Checkbox, Column, GridColumns, Spacer, Text } from "@artsy/palette"
+import { FC } from "react"
+import { useAlertContext } from "Components/Alert/Hooks/useAlertContext"
+import { ShowMore } from "Components/Alert/Components/Filters/ShowMore"
+import { SearchCriteriaAttributeKeys } from "Components/SavedSearchAlert/types"
+
+interface BaseAlertFilterProps {
+  criteriaKey: SearchCriteriaAttributeKeys
+  description?: string
+  expanded?: boolean
+  label: string
+  options: { value: string; name: string }[]
+}
+
+export const BaseAlertFilter: FC<BaseAlertFilterProps> = ({
+  criteriaKey,
+  description,
+  expanded = false,
+  label,
+  options,
+}) => {
+  const { state, dispatch } = useAlertContext()
+
+  const toggleSelection = (selected, name) => {
+    let updatedValues = (state.criteria[criteriaKey] || []) as string[]
+
+    if (selected) {
+      updatedValues = [...updatedValues, name]
+    } else {
+      updatedValues = updatedValues.filter(item => item !== name)
+    }
+
+    dispatch({
+      type: "SET_CRITERIA_ATTRIBUTE",
+      payload: { key: criteriaKey, value: updatedValues },
+    })
+  }
+
+  return (
+    <>
+      <Text variant="sm-display" pb={1}>
+        {label}
+      </Text>
+
+      {!!description && (
+        <Text variant="xs" pb={1} mt={2}>
+          {description}
+        </Text>
+      )}
+
+      <Spacer y={2} />
+
+      <GridColumns>
+        <ShowMore expanded={expanded}>
+          {options.map(({ name, value }, index) => {
+            return (
+              <Column span={6} key={index}>
+                <Checkbox
+                  onSelect={selected => toggleSelection(selected, value)}
+                  selected={(state.criteria[criteriaKey] as
+                    | string[]
+                    | null)?.includes(value)}
+                >
+                  {name}
+                </Checkbox>
+              </Column>
+            )
+          })}
+        </ShowMore>
+      </GridColumns>
+    </>
+  )
+}
--- a/src/Components/Alert/Components/Filters/Medium.tsx
+++ b/src/Components/Alert/Components/Filters/Medium.tsx
@@ -1,51 +1,14 @@
-import { Checkbox, Column, GridColumns, Spacer, Text } from "@artsy/palette"
 import { FC } from "react"
-import { useAlertContext } from "Components/Alert/Hooks/useAlertContext"
 import { MEDIUM_OPTIONS } from "Components/ArtworkFilter/ArtworkFilters/MediumFilter"
-import { ShowMore } from "Components/Alert/Components/Filters/ShowMore"
+import { BaseAlertFilter } from "Components/Alert/Components/Filters/BaseAlertFilter"
 
 export const Medium: FC = () => {
-  const { state, dispatch } = useAlertContext()
-
-  const toggleSelection = (selected, name) => {
-    let updatedValues = state.criteria.additionalGeneIDs || []
-
-    if (selected) {
-      updatedValues = [...updatedValues, name]
-    } else {
-      updatedValues = updatedValues.filter(item => item !== name)
-    }
-
-    dispatch({
-      type: "SET_CRITERIA_ATTRIBUTE",
-      payload: { key: "additionalGeneIDs", value: updatedValues },
-    })
-  }
-
   return (
-    <>
-      <Text variant="sm-display" pb={1}>
-        Medium
-      </Text>
-
-      <Spacer y={2} />
-
-      <GridColumns>
-        <ShowMore expanded={false}>
-          {MEDIUM_OPTIONS.map(({ name, value }, index) => {
-            return (
-              <Column span={6} key={index}>
-                <Checkbox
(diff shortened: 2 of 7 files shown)


src/Components/Alert/Components/Filters/Medium.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const allowedMediums = mediums?.counts?.length
+    ? mediums.counts
+    : MEDIUM_OPTIONS
+
+  const intersected = intersection(
+    additionalGeneIDs,
+    allowedMediums.slice(INITIAL_ITEMS_TO_SHOW).map(({ value }) => value)
+  )
+  const hasBelowTheFoldMediumFilter = intersected.length > 0
+  const resultsSorted = sortResults(additionalGeneIDs, allowedMediums)
```
olerichter00, line 48: For now, I have implemented the same sorting logic as the one used in the filter panel. I have opened a [Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1699277137885669) to clarify on whether we should move the selected options to the top or not.

src/Components/Alert/Components/Filters/ShowMore.tsx
```diff
@@ -0,0 +1,39 @@
...
+import * as React from "react"
+
+interface ShowMoreProps {
+  initial?: number
+  expanded?: boolean
+}
+
+export const INITIAL_ITEMS_TO_SHOW = 6
+
+export const ShowMore: React.FC<ShowMoreProps> = ({
```
olerichter00, line 12: I have created a new `ShowMore` for the two-column `GridColumn` design.

src/Components/Alert/Components/Filters/Medium.tsx
olerichter00: I will update this to not move selected options to the top, as discussed on [Slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1699277137885669).

src/Components/Alert/Components/Filters/ShowMore.tsx
olerichter00: [Palette's ShowMore ](https://palette.artsy.net/elements/showmore/#gatsby-focus-wrapper)component looks a bit different and cannot be used with `GridColumn` at the moment. Therefore, I thought it would be better to create a custom component instead of modifying the component from Palette to work in different places.

src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+interface BaseAlertFilterProps {
+  criteriaKey: SearchCriteriaAttributeKeys
+  description?: string
+  expanded?: boolean
+  label: string
+  options: { value: string; name: string }[]
+}
+
+export const BaseAlertFilter: FC<BaseAlertFilterProps> = ({
```
olerichter00, line 15: I have created this base filter which contains all the logic and can be used like this:

```
    <BaseAlertFilter
      label="Rarity"
      criteriaKey="attributionClass"
      options={ATTRIBUTION_CLASS_OPTIONS}
    />
```

olerichter00: I'm happy to hear suggestions for a better name 🙃

olerichter00: This abstraction should work for all filters, except colors and price range.

dblandin: **suggestion**: Maybe `QuickMultipleSelectAlertFilter` to communicate that this is a quick/convenient way to spin up a filter as long as the filter:
- shares the same `criteriaKey` (not applicable for "ways to buy")
- is multiple select (checkbox UX)
- assumes that the data persisted to the alert context
- persisted data takes the form of `key: values[]`

src/Components/Alert/Components/Filters/__tests__/Medium.jest.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const MediumFilterTestComponent = () => {
+    filterContext = useArtworkFilterContext()
+    alertContext = useAlertContext()
+
+    return <Medium />
+  }
+
+  const getWrapper = (contextProps = {}, initialCriteria = {}) => {
+    return mount(
+      <ArtworkFilterContextProvider {...contextProps}>
```
dblandin, line 29: **question**: Do we need this wrapping `ArtworkFilterContextProvider` component here? Unless we're specifically testing a component that makes use of the context (which shouldn't be the case with these new alert components), I don't think it adds much value in the test.

src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
```diff
@@ -0,0 +1,73 @@
+import { Checkbox, Column, GridColumns, Spacer, Text } from "@artsy/palette"
+import { FC } from "react"
+import { useAlertContext } from "Components/Alert/Hooks/useAlertContext"
+import { ShowMore } from "Components/Alert/Components/Filters/ShowMore"
+import { SearchCriteriaAttributeKeys } from "Components/SavedSearchAlert/types"
+
+interface BaseAlertFilterProps {
+  criteriaKey: SearchCriteriaAttributeKeys
+  description?: string
+  expanded?: boolean
```
dblandin, line 10: **suggestion**: Based on conversations in Slack, perhaps we defer the addition of this `expanded` prop until we need it?

src/Components/Alert/Components/Filters/__tests__/Medium.jest.tsx
olerichter00: I'm testing whether updating the alert filter affects the filter state of the `ArtworkFilterContext` [here](https://github.com/artsy/force/pull/13106/files#diff-157c5645f7fc64324b20c7393e897dab23f33efdeffdf304190ffd09ad467eb1R51) to make sure only the state in the alert context gets changed.

I thought it might be good to have it here because it's one of the requirements.

src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
olerichter00: I've updated the name.

src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
olerichter00: Hmm, I would be better leaving it, given that we will need the "Show More" logic when implementing other filters, and removing the code would be more work than leaving it.

But yeah, you're right. I would not have added this logic if if I had known about the new requirements before working on the ticket.

But I'm happy to remove it if you feel strongly about it.

dblandin: Sure, let's leave it in for now — we can revisit later.

I wonder how many of those filters will utilize "Show more" given the questions around selected filters below the fold or repositioning those selected filters to the top of the list.

We also need to figure out whether we can port this proposed pattern for app over to Desktop Web — which would also obviate the need for a "Show more" implementation.

Screenshot 2023-11-07 at 14 39 19

src/Components/Alert/Components/Filters/__tests__/Medium.jest.tsx
dblandin: Gothcha, I don't consider this blocking but in that case would it be possible to move this wrapping to only the test that exercises that requirement? Otherwise it feels a bit misleading to have it included for every test in the file.

I'd also support eventually moving this kind of assertion up a level once we're exercising a higher-level flow and not filter-specific — but also not blocking.

dblandin (approved):
LGTM!

dariakoko (approved):

src/Components/Alert/Components/Filters/BaseAlertFilter.tsx
olerichter00: Yeah, it might be that we won't need the "Show More" pattern at all in the end.

src/Components/Alert/Components/Filters/__tests__/Medium.jest.tsx
olerichter00: This is a good argument. We don't need the assertion in every filter test. I removed it in https://github.com/artsy/force/pull/13112

PR #13109, opened 2023-11-06
RFC: DX: Helper for expecting and handling relay operations

erikdstock:

This PR creates a helper which can be used with `renderWithRelayTL` or on its own to `await` a single operation to be called,  [optionally] assert on what the operation is and its variables, and provide a mockResolver object or error to resolve/reject the mutation. It solves a number of problems:

- Awkward reliance on `tick()`, `flushPromiseQueue`, `wrapper.update()` to try and shepherd a string of mutations along
- Verbose and difficult-to-parse setup/assertion code in tests
- Reliance on jest mocks and deeply nested mutation names/variables when inspecting the mockRelayEnv directly.

The code itself is pretty simple and documented, and the helper can be used inside a relay test wrapper or alone, providing your own mock relay env as a first parameter. Its usage with the former looks like this:
```ts
import { expectAndHandleOperationWithEnv } from "DevTools/mockRelayExpectations"

userEvent.click(screen.findByText("Submit"))

await expectAndHandleOperationWithEnv(mockRelayEnv, {
  name: "UpdateUserDefaultAddressMutation",
  variables: { input: { addressID: "123" } },
  mockResolvers: {
   UpdateUserDefaultAddressMutation: () => ({ internalID: "123", isDefault: true }),
  }
 )

await waitFor(() => getByText("Your default address has been updated"))
```

With the test wrapper and a rejected query: 
```tsx
const { expectAndHandleOperation } = renderWithRelay()

userEvent.click(screen.findByText("Submit"))

await expectAndHandleOperation({
  name: "UpdateUserDefaultAddressMutation",
  variables: { input: { addressID: "" } },
  error: new Error("address not supplied")
 )

await waitFor(() => getByText("oops"))
```


src/DevTools/mockRelayExpectations.ts
```diff
@@ -0,0 +1,72 @@
...
+ *  }
+ * )
+ *
+ * await expectAndHandleOperationWithEnv(mockRelayEnv, {
+ *   name: "grantAdminPrivilegesMutation",
+ *   error: new Error("You do not have permission to perform this action"),
+ * })
+ * ```
+ */
+export async function expectAndHandleOperationWithEnv(
```
damassi, line 44: Can we name this something a little gentler? The name is saying what this is doing, but we're generally always "handling stuff with relay operations". What would be a friendly, terse way to express what this is doing? `resolveMutations` maybe? 

And we don't want assertions buried in lib code. Lib code _supports_ assertions, and we do what we can to make that as easy as possible. But when we start mixing the two together, it soon becomes brittle and hard to understand later. 

The main named export should be 1:1 with the file name.

src/DevTools/mockRelayExpectations.ts
```diff
@@ -0,0 +1,72 @@
...
+  } = operationConfig
+  let operation: OperationDescriptor = {} as OperationDescriptor
+  await waitFor(() => {
+    operation = mockRelayEnv.mock.getMostRecentOperation()
+  })
+  if (expectedName) {
+    expect(operation.request.node.operation.name).toEqual(expectedName)
+  }
+  if (expectedVariables) {
+    expect(operation.request.variables).toMatchObject(expectedVariables)
```
damassi, line 61: no assertions in lib code

src/DevTools/mockRelayExpectations.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  if ("mockResolvers" in handler) {
```
damassi, line 64: no need to scoop things up into `...handler`. Assert directly against the thing. 

```tsx
if (mockResolvers) {
...
}
if (error) {
  ... 
}
```

src/DevTools/setupTestWrapper.tsx
```diff
@@ -56,13 +66,23 @@ type SetupTestWrapper<T extends OperationType> = {
...
     })
   })
 
-	expect(screen.queryByText("Hello world")).toBeInTheDocument()
+  // Manually resolve pending mutations:
+  expect(screen.queryByText("Hello world")).toBeInTheDocument()
+  userEvent.click(screen.getByText("Click me"))
+
+  await expectAndHandleOperation({
+    name: "clickIncrementMutation",
```
damassi, line 80: Why do we need the name? The name should be derived from the operation itself (does relay test utils return this metadata?) and returned to be asserted in actual test code

src/DevTools/setupTestWrapper.tsx
```diff
@@ -56,13 +66,23 @@ type SetupTestWrapper<T extends OperationType> = {
...
       description: "Hello world!"
     })
   })
 
-	expect(screen.queryByText("Hello world")).toBeInTheDocument()
+  // Manually resolve pending mutations:
+  expect(screen.queryByText("Hello world")).toBeInTheDocument()
+  userEvent.click(screen.getByText("Click me"))
+
+  await expectAndHandleOperation({
```
damassi, line 79: `resolveMutations`

src/DevTools/setupTestWrapper.tsx
```diff
@@ -110,7 +133,11 @@ export const setupTestWrapperTL = <T extends OperationType>({
       return MockPayloadGenerator.generate(operation, mockResolvers)
     })
 
-    return { ...view, env }
+    return {
+      ...view,
+      env,
+      expectAndHandleOperation: buildExpectAndHandleOperation(env),
```
damassi, line 139:
```tsx
const resolveMutations = setupMutationResolver(env)
...
return {
   ...view,
   env,
   resolveMutations
}
```

damassi:
This addition def seems harmless, as it's an add-on to our existing flow and doesn't impact the existing API. That said, I still don't exactly understand what the problem is here. We do a lot of mutations, both in Force as well as in Eigen, using a pretty vanilla relay API that folks haven't ever really had a problem with (afaik). Have you thoroughly researched those use-cases before diving into this? If this makes the mutation testing process a bit smoother, though, than I'm all for it. 

Maybe @dzucconi has some thoughts about it, and @araujobarret. Also, @MounirDhahri has dealt with the Eigen mutation side of things a lot, so he would have a lot of context there about how things are handled.

Regarding the overall API, I'm wondering why we need so much additional surface area. For example, the existing api for our relay helper is:

```tsx
const { renderWithRelay } = setupTestWrapper({
  query
  component,
  variables
})
```

And then, later:

```
renderWithRelay(mockResolvers)
```

All of the things we need to resolve mutation responses are encoded right in here. We have mockResolvers, we have a query, we have variables. (We don't have errors, but that can be added in the return). In order to create _multiple_ instances of setup code, its as easy as wrapping it within a function:

```tsx
const setup = (props: SetupTestWrapperProps) => {
  return setupTestWrapper({
    Component: Foo,
    ...props
  })
}
```

And then in each test, can easily run 
```tsx
const { resolveMutations } = setup({ variables: { someMutationThing }})

await resolveMutations(mockResolvers)
```

This then leverages the existing api (with perhaps some minor internal modifications) without expanding introducing new requirements. Just like with `renderWithRelay`, the only arguments that it takes are the mock resolvers. It keeps things carefully aligned with existing patterns. 

Thoughts on the API suggestion above?

src/DevTools/mockRelayExpectations.ts
```diff
@@ -0,0 +1,72 @@
...
+ * })
+ * ```
+ */
+export async function expectAndHandleOperationWithEnv(
+  mockRelayEnv: MockEnvironment,
+  operationConfig: OperationConfig
+) {
+  const {
+    name: expectedName,
+    variables: expectedVariables,
```
damassi, line 50: We need to really try hard to resist the urge to 1:2 everything (as discussed at length in your other big refactor PR). `name` is `name`, `variables` is `variables`. Every unnecessarily aliased thing is complexity in purest form.

dzucconi:
This seems like it's the only useful bit?

```ts
  await waitFor(() => {
    mockRelayEnv.mock.getMostRecentOperation()
  })
```

We wouldn't want to automatically set up assertions or anything of that sort.

damassi:
Yeah, looking around in Eigen [this is how it used to be done](https://github.com/artsy/eigen/blob/ede0c64152702a63de92c69a93db9b618724f7ee/src/app/Scenes/SavedSearchAlert/EditSavedSearchAlert.tests.tsx#L103-L116). 

However, `resolveMostRecentRelayOperation` has been deprecated, and rolled up into the relay testing util, [which seems pretty ideal](https://github.com/artsy/eigen/blob/ede0c64152702a63de92c69a93db9b618724f7ee/src/app/utils/tests/setupTestWrapper.tsx#L129-L143). 

Screenshot 2023-11-06 at 12 37 21 PM

Thoughts on setting things up similar to how they are in Eigen?

erikdstock:
So this is just a proposal and I'm not married to it but let me try to illustrate my points.
1. The helper is called `expect...` because it has an expectation and that is by design. It has an expectation so that developers are encouraged to make explicit their assumptions in tests. Another way I wrote this in an earlier version was something like this:
```ts
const mockRelayOperation = (mockResolvers) => {
}
const firstRequest =

File: src/DevTools/mockRelayExpectations.ts (hunk 7 above)
Comment (erikdstock) [reply src/DevTools/mockRelayExpectations.ts:50]: responded below

File: src/DevTools/setupTestWrapper.tsx (hunk 4 above)
Comment (erikdstock) [reply src/DevTools/setupTestWrapper.tsx:80]: yes, it's available inside, though i re-wrote it now to be optional and only match if provided. The helper is already an expectation (if no operation happens it will throw an error), and I though it semantically useful to let the developer make their intent and other expectations explicit:
- i expect there to be an operation
- i expect it to be this operation
- i expect it to send this input
- it should resolve/reject with these values i care about

**Comment by damassi** [comment]
Thanks for explaining a bit more. I want to dig a bit deeper into the Eigen example and try to suss out where it might be lacking. Looking [at this example](https://github.com/artsy/eigen/blob/ede0c64152702a63de92c69a93db9b618724f7ee/src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx#L42-L70), we get a really clean, low-information API. How might this be improved upon?

**Comment by erikdstock** [comment]
> This seems like it's the only useful bit?
> 
> ```ts
>   await waitFor(() => {
>     mockRelayEnv.mock.getMostRecentOperation()
>   })
> ```
> 
> We wouldn't want to automatically set up assertions or anything of that sort.

The thing is that `await waitfor(() => operatio = mockEnv.mock.getMostRecentOperation())` already acts as an assertion - if there is no operation it throws an error. This method is just adding additional checks to that, which help with readability and intent.

File: src/DevTools/setupTestWrapper.tsx (hunk 4 above)
Comment (araujobarret) [reply src/DevTools/setupTestWrapper.tsx:80]: I'm kinda super against this(wanted to emphasize really hard here  😂), I had a nightmare refactoring these codes in the order app tests to unblock the relay upgrade mostly because of a lot of these abstractions involving this approach expecting a specific name from a mutation.
I will come with a decent explanation why tomorrow

**Comment by erikdstock** [comment]
> Thanks for explaining a bit more. I want to dig a bit deeper into the Eigen example and try to suss out where it might be lacking. Looking [at this example](https://github.com/artsy/eigen/blob/ede0c64152702a63de92c69a93db9b618724f7ee/src/app/Components/Artist/RelatedArtistsRailCell.tests.tsx#L42-L70), we get a really clean, low-information API. How might this be improved upon?

For reference, this helper is [here](https://github.com/artsy/eigen/blob/main/src/app/utils/tests/setupTestWrapper.tsx#L129-L135):
```ts
    const mockResolveLastOperation = (mockResolvers: MockResolvers) => {
      act(() => {
        env.mock.resolveMostRecentOperation((operation) => {
          return MockPayloadGenerator.generate(operation, mockResolvers)
        })
      })
    }
```

What mine adds is extra flexibily to handle with mockResolvers _or_ reject, and extra args to assert on the operation name/vars/etc without needing to fiddle around deep inside the operation object (which could be improved more with custom matchers).

I think i've made my case and recognize there is some mixture of mock/assertion that feels wrong (copilot agrees with yall this is a weakness). My arg is that since `getMostRecentOperation` itself asserts that an operation is taking place, might as well give ourselves the ability to document/assert what we think it is right there rather than 5 lines later.

But no hard feelings if we don't like this! Thanks for considering.

damassi:
Its not about liking or not liking! Its about understanding tradeoffs, and getting to the bottom of things. 

> What mine adds is extra flexibility to handle with mockResolvers or reject

If you look closer, there's also `mockRejectLastOperation` returned on the very next line which is also returned, in addition to `mockResolveLastOperation`. And `mockResolveLastOperation` accepts a `mockResolvers`. So unless I'm misunderstanding, this Eigen pattern achieves both of these aims? 

> and extra args to assert on the operation name/vars/etc without needing to fiddle around deep inside the operation object (which could be improved more with custom matchers).

Referring to this bit:

```tsx
expect(env.mock.getMostRecentOperation().request.node.operation.name).toBe(name)
```

The `request.node.operation.name` is the concern? 

I can understand why it might feel nice to come up with a pattern that abstracts upon this, but two things:

1) With relay, we literally want: the request, the node, the operation, and then the name. We don't necessarily need an entirely new API for something this literal, which is more things to learn. Abstraction should be avoided unless it is necessary.
2) But if we still wanted abstraction, we _can_ abstract this in a simple fashion:

```tsx
const getOperationName = () => {
  return env.mock.getMostRecentOperation().request.node.operation.name
}

return { ...view, env, mockResolveLastOperation, mockRejectLastOperation, getOperationName }

// in test 

mockResolveLastOperation()
expect(getOperationName()).toBe('foo')
```

Simple getter. No abstraction necessary other than a function that shorthands a long object chain.

What other bits might be needed here, following this same pattern? We can get far with getters like this, while keeping complexity and information density low.

src/DevTools/mockRelayExpectations.ts
damassi: I'm not seeing what you're referring to. If there's a response, its typically helpful for reviewers to provide a direct link to said response.

erikdstock:
I think you have some points here. In using this pattern in a few more tests (just local helpers, not using this PR) I noticed that even when they ran correctly I was getting lint errors because my tests did not have explicit expectations.

Since I'm doing this as I refactor a lot of tests that previous skipped relay entirely, my goal is to get it right in a way that is consistent and not mysterious (why did they flush here, tick there, etc) but if I can do that without introducing a whole new pattern, great. I'll start out with what was linked in eigen and see if that works.

One thought - what if we took [eigen's implmentation](https://github.com/artsy/force/pull/13109#issuecomment-1797006638) and returned `{operation, operatioName, operationVariables}`? It would remove the need for extra getters and give us a nearly identical implementation to what eigen already does (only adding the return value and promisifying). Assertion logic could continue to live outside:
```ts
const mockResolveLastOperation = async (mockResolvers: MockResolvers) => {
  const operation = await waitFor(() => env.mock.getMostRecentOperation())
  await act(() => {
    env.mock.resolveMostRecentOperation((operation) => {
      return MockPayloadGenerator.generate(operation, mockResolvers)
    })
  })
  const operationName = operation.request.node.operation.name
  const operationVariables = operation.request.variables
  return { operation, operationName, operationVariables }
}

const adminRequest = await mockResolveMostRecentOperation({User: () => {admin: true})
expect(adminRequest.operationName).toBe("adminMutation")
```
This would hide a little bit of boilerplate but let us assert in a clean way.

damassi:
I think that's a fine compromise 🤝 That function as it currently is is a fire-and-forget thing. But with this suggested change, if a user wanted additional info, then could pluck from the return value and it would just be better.

Thank you for working thorough these thoughts 👍

damassi:
Ah, actually @erikdstock - looking closer, we definitely don't want to include `waitFor` in there, as that's crossing the test/lib-code boundary. I think using a standard await is sufficient.

araujobarret:
he meant `waitFor` should not be used outside of a test file using React Testing Library/Enzyme.
also act doesn't need the `await` we just :
```jsx
act(() => {
  env.mock.resolveMostRecentOperation((operation) => {
    return MockPayloadGenerator.generate(operation, mockResolvers)
  })
})
```
re the attempt to use `waitFor`, I don't think it's a good pattern, if the operation is not ready at this point in time then we cannot force it in a way to always return the other fields you want.

A suggestion could be you use a more functional approach if you really need them
```jsx
const mockResolveLastOperation = (mockResolvers: MockResolvers) => {
  let _operation: ReadOnly<OperationDescriptor>
  act(() => {
    env.mock.resolveMostRecentOperation((operation) => {
      _operation = operation
      return MockPayloadGenerator.generate(operation, mockResolvers)
    })
  })
  const operationName = _operation?.request.node.operation.name
  const operationVariables = _operation?.request.variables
  return { operation, operationName, operationVariables }
}

const adminRequest = await mockResolveMostRecentOperation({User: () => {admin: true})
expect(adminRequest.operationName).toBe("adminMutation")
```

Again, I don't see the need to add more stuff to the TestWrapper internally to expose the `operationName` and `operationVariables`, I'd rather have some utility like:
```jsx
export const getLastOperationProps = (env: RelayMockEnvironment) => {
  const operation = env.mock.getMostRecentOperation()
  return {
    operationName: operation.request.node.operation.name,
    operationVariables: operation.request.variables,
  }
}

// in the test
expect(getLastOperationProps(env).operationName).toBe("adminMutation")
```
☝️ but again, I don't really see the verbosity gain soo much in doing it

damassi:
Thanks @araujobarret, yep exactly. Since we do assert against name all over, I think it would be helpful to roll up in the way you outlined in your example, vs another snippet. Its like, `mockResolveLastOperation` resolves things, and then returns common metadata that we often assert against in tests. its just a couple lines of code that do seem appropriately placed, versus wrapping up into another file where it could be lost and we end up reinventing the wheel again. @araujobarret, do you have any strong objections?

erikdstock:
OK, pushing updates that come closer to this:
- Added `getOperationProps(operation)`
- Removed expectation code
- Added `getOperationProps()` method, but am not getting _the_ operation from it because this can fail.

I kept `await act()` because I have never understood `act()` that well, but GH copilot _insists_ that `await` is useful here in making sure updates are applied to the react dom. It also insists that act will work with `env.mock.resolve(operation)` even though that only refers to a promise, not explicitly triggering any react updates. I'm not saying we should trust copilot though.

<details><summary>copilot chat that could be wrong</summary>

image
image
image

</details>

src/DevTools/setupTestWrapper.tsx
```diff
@@ -110,7 +132,14 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
     })
 
-    return { ...view, env }
+    return {
+      ...view,
+      env,
+      mockResolveMostRecentOperation: mockResolveMostRecentOperationWithEnv(
+        env
+      ),
+      mockRejectMostRecentOperation: mockRejectMostRecentOperationWithEnv(env),
```
damassi, line 141: [blocking]: Sorry if this wasn't clear, but we should be copying Eigen's approach exactly, to a T, minus the few additions we discussed in the final comments. We don't need new files, or new `apply` functions, or new types, or even comments, (other than a basic usecase example). and so on. 

It also might be helpful to close this PR and open up a fresh one.

src/DevTools/mockRelayOperations.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Returns an object containing the operation, its name and variables.
+ */
+export const mockResolveMostRecentOperation = async (
+  mockRelayEnv: MockEnvironment,
+  mockResolvers: MockPayloadGenerator.MockResolvers
+): Promise<MockedOperationResult> => {
+  const operation = await waitFor(() =>
+    mockRelayEnv.mock.getMostRecentOperation()
+  )
+  await act(() => {
```
damassi, line 39: And as @araujobarret said quite clearly, and contrary to the Bot, we don't need to to `await` act.

erikdstock:
Alright, I'm gonna close this for now and limit my changes to the test updates I actually developed this approach for without complex helpers

It's true that I did post some copilot chats as I was trying to work through this, but I also posted links to other sources showing how act can be used as a promise. It does introduce a promise to a test that uses it, but tests touching relay operations are always async anyways. Anyhow, disagree and commit.

damassi:
Eigen-based follow-up here: https://github.com/artsy/force/pull/13111

damassi:
And to be clear for the record: 

> tests touching relay operations are always already async. 

This is false. Tests touching relay operations are _not_ always already async. There's actually not a single `async` reference in our relay test utils at all.

PR #13111, opened 2023-11-07
feat(relay-test-utils): Add support for mutations

damassi:

### Description

Following Eigen, this allows for better mutation testing support from directly within our common testing patterns. 

### Example

```tsx
it("follows the artist", () => {
  const { mockResolveLastOperation } = renderWithRelay({ Artist: () => artist })

  fireEvent.press(screen.getByText("Follow"))

  mockResolveLastOperation({
    Artist: () => ({
      isFollowed: true,
    }),
  })

  expect(screen.getByText("Following")).toBeOnTheScreen()
})

it("unfollows the artist", () => {
  const { mockResolveLastOperation } = renderWithRelay({
    Artist: () => ({
      isFollowed: true,
    }),
  })

  fireEvent.press(screen.getByText("Following"))

  const { operationName, operationVariables } = mockResolveLastOperation({ 
    Artist: () => ({ 
      isFollowed: false 
    }) 
  })

  expect(screen.getByText("Follow")).toBeOnTheScreen()
  expect(operationName).toBe("foo")
  expect(operationVariables).toBe("bar")
})
```

### DIFF
--- a/src/DevTools/setupTestWrapper.tsx
+++ b/src/DevTools/setupTestWrapper.tsx
@@ -1,8 +1,12 @@
-import { render, RenderResult } from "@testing-library/react"
+import { act, render, RenderResult } from "@testing-library/react"
 import { mount } from "enzyme"
 import * as React from "react"
-import { QueryRenderer } from "react-relay"
-import { GraphQLTaggedNode, OperationType } from "relay-runtime"
+import { QueryRenderer, Variables } from "react-relay"
+import {
+  GraphQLTaggedNode,
+  OperationDescriptor,
+  OperationType,
+} from "relay-runtime"
 import {
   createMockEnvironment,
   MockPayloadGenerator,
@@ -77,7 +81,19 @@ type RTLRenderResult = RenderResult<
   typeof import("@testing-library/dom/types/queries"),
   HTMLElement
 >
-type RenderWithRelay = RTLRenderResult & { env: MockEnvironment }
+
+type RenderWithRelay = RTLRenderResult & {
+  env: MockEnvironment
+  mockResolveLastOperation: (
+    mockResolvers: MockResolvers
+  ) => {
+    operation: OperationDescriptor
+    operationName: string
+    operationVariables: Variables
+  }
+  mockRejectLastOperation: (error: Error) => void
+}
+
 export const setupTestWrapperTL = <T extends OperationType>({
   Component,
   query,
@@ -89,6 +105,7 @@ export const setupTestWrapperTL = <T extends OperationType>({
     mockedEnv?: ReturnType<typeof createMockEnvironment>
   ): RenderWithRelay => {
     const env = mockedEnv ?? createMockEnvironment()
+
     const TestRenderer = () => (
       <QueryRenderer<T>
         environment={env}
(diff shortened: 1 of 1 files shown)


src/DevTools/setupTestWrapper.tsx
```diff
@@ -110,7 +127,32 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
+
+      act(() => {
+        env.mock.resolveMostRecentOperation(relayOperation => {
+          operation = relayOperation
+          return MockPayloadGenerator.generate(operation, mockResolvers)
+        })
+      })
+
+      // Cast here to get around use-before-assign rule
+      const _operation = operation as OperationDescriptor
```
damassi, line 141: Not sure how else to do this @araujobarret

erikdstock: If you look at the [relay v14 docs](https://relay.dev/docs/v14.0.0/guides/testing-relay-components/#relaymockenvironment-api-overview) it says that `mockEnvironment.resolveMostRecentOperation()` is actually a sugar for ` env.mock.resolve(env.mock.getMostRecentOperation())`. So I would suggest just calling these methods proceduratlly:
```ts
const operation = env.mock.getMostRecentOperation()
act(() => {
  env.mock.resolve(operation, MockPayloadGenerator.generate(operation, mockResolvers))
})
return operation, etc

File: src/DevTools/setupTestWrapper.tsx [hunk 2]
```diff
@@ -77,7 +81,19 @@ type RTLRenderResult = RenderResult<
...
+
+type RenderWithRelay = RTLRenderResult & {
+  env: MockEnvironment
+  mockResolveLastOperation: (
+    mockResolvers: MockResolvers
+  ) => {
+    operation: OperationDescriptor
+    operationName: string
+    operationVariables: Variables
+  }
```
erikdstock, line 93: After working with this API a bit while fixing tests, I dropped the extra names from here and started using `getOperationProps(operation)` instead. It felt cleaner but I worried it would create confusion later, while putting these in one single definition as carlos suggested makes clear where they come from without relying on intermediate helper logic. I think the `props` naming can get a little overloaded but otherwise it does end up feeling cleaner for me if we don't mind extra imports (also, i was doing this outside the setupTestWrapper context).

erikdstock (commented):
some comments but no blockers here. thanks for considering.

src/DevTools/setupTestWrapper.tsx
damassi: This is a nice solution to the Q, updated

src/DevTools/setupTestWrapper.tsx
damassi: For something this exceedingly simple, I'd rather leave everything consolidated around one single single thing. If it grows and actually becomes complex, then we consider breaking things apart, but not prematurely. 

```tsx
const { operationName, operationVariables } = mockResolveLastOperation({ ... })

expect(operationName).toBe("foo")
expect(operationVariables).toBe(bar)
```

araujobarret (approved):
Easy to use and a simple implementation 👍

PR #13114, opened 2023-11-08
fix: Move "Saved Search preview" fetching logic into alert context to avoid flashing

olerichter00:
Addresses [ONYX-507]

## Description

This PR fixes the display name and criteria pill flashing in the new Create Alert flow by moving the "Saved Search Preview" fetching logic to the alert context.

I'm replacing the two `QueryRenderer` with one `fetchQuery` call that is triggered by a `useEffect` hook inside of `AlertProvider`. It fetches and updates the display name and labels (saved search preview) whenever the alert criteria change.

https://github.com/artsy/force/assets/4691889/313dc9ca-9d55-4928-87c3-bf20ec18ad78

[ONYX-507]: https://artsyproduct.atlassian.net/browse/ONYX-507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Alert/Hooks/useAlertContext.tsx
```diff
@@ -22,11 +28,23 @@ type Settings = {
...
+  labels?: (Label | null)[]
+  displayName: string
+}
+
 export type State = {
   settings: Settings
   criteria: SearchCriteriaAttributes
   searchCriteriaID?: string
   currentArtworkID?: string
+  preview: Preview
```
dblandin, line 47: **suggestion / non-blocking**: Instead of defining new types, would it make sense to have this be of type `PreviewSavedSearch` as exposed via the Metaphysics schema? Might also make this nullable so that it can unset initially but then set as soon as we get the first MP response.

src/Components/Alert/Hooks/useAlertContext.tsx
```diff
@@ -188,6 +223,42 @@ export const AlertProvider: FC<AlertProviderProps> = ({
...
+                  value
+                }
+              }
+            }
+          }
+        `,
+        { attributes: criteriaState as PreviewSavedSearchAttributes }
+      )?.toPromise()
+
+      dispatch({
```
dblandin, line 250: **praise**: This is a great approach!!

dblandin (approved):
Looking good!!

src/Components/Alert/Hooks/useAlertContext.tsx
```diff
@@ -188,6 +211,48 @@ export const AlertProvider: FC<AlertProviderProps> = ({
     basis.current = state
   }, [state])
 
+  // Fetching the display name and labels for alert name input placeholder and the criteria pills, when the criteria changes.
+
+  const { debouncedValue: debouncedCriteria } = useDebouncedValue({
+    value: state.criteria,
+    delay: 200,
+    settings: { leading: true },
+  })
```
olerichter00, line 220: Debouncing the preview fetch with `leading` set to `true` in order to get the result as fast as possible and prevent too many requests.

src/Components/Alert/Hooks/useAlertContext.tsx
olerichter00: Good idea! I could extract the type like this:

```
type PreviewSavedSearch = NonNullable<
  useAlertContextPreviewQuery["response"]["viewer"]
>["previewSavedSearch"]
```

dblandin: Nice! That looks good to me 👍🏼

src/Components/Alert/Components/Form/AlertNameInput.tsx
```diff
@@ -1,78 +1,21 @@
-import { FC, useEffect, useState } from "react"
+import { FC } from "react"
```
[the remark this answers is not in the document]
olerichter00: I don't think there is any difference 🤔

src/Components/Alert/Components/CriteriaPills.tsx
```diff
@@ -103,17 +54,12 @@ const CriteriaPills: FC<CriteriaPillsProps> = ({
...
-    `,
-  }
-)
+const CriteriaPillsPlaceholder: FC = () => {
+  return (
+    <>
+      <SkeletonBox width={150} height={30} />
+      <SkeletonBox width={80} height={30} />
+      <SkeletonBox width={100} height={30} />
+    </>
```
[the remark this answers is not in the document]
olerichter00: That could work as well 👍 We would just need to get the correct width somehow (`150`, `80`, `100` or random).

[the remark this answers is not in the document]
dariakoko: With random, or the same value for all components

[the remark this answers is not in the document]
olerichter00: Hmm, it would actually be longer because we have a character limit per line:

```
    <>
      {[...Array(3)].map((_, index) => (
        <SkeletonBox
          width={Math.floor(Math.random() * (150 - 80 + 1)) + 80}
          height={30}
          key={index}
        />
      ))}
    </>
```

PR #13115, opened 2023-11-08
chore: Port mutation helpers to enzyme setupTestWrapper, return destructured wrapper

erikdstock:
This PR follows #13111 to add our mutation helper support to tests using the legacy enzyme `setupTestWrapper`. In order to do this I changed that file's `return wrapper` to a destructurable `return { wrapper, ...otherStuff }`, matching the return interface of our current preferred `setupTestWrapperTL`.

### DIFF
--- a/src/DevTools/setupTestWrapper.tsx
+++ b/src/DevTools/setupTestWrapper.tsx
@@ -121,12 +121,6 @@ export const setupTestWrapperTL = <T extends OperationType>({
       />
     )
 
-    const view = render(<TestRenderer />)
-
-    env.mock.resolveMostRecentOperation(operation => {
-      return MockPayloadGenerator.generate(operation, mockResolvers)
-    })
-
     const mockResolveLastOperation = (mockResolvers: MockResolvers) => {
       const operation = env.mock.getMostRecentOperation()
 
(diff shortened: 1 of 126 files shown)


src/DevTools/setupTestWrapper.tsx
```diff
@@ -197,7 +231,7 @@ export const setupTestWrapper = <T extends OperationType>({
 
     wrapper.update()
 
-    return wrapper
+    return { wrapper, env, mockResolveLastOperation, mockRejectLastOperation }
```
erikdstock, line 234: This file contains the only functional changes.

src/DevTools/setupTestWrapper.tsx
```diff
@@ -137,18 +131,30 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
+      const operationVariables = operation.request.variables
+
+      return { operation, operationName, operationVariables }
     }
 
+    const view = render(<TestRenderer />)
+
+    env.mock.resolveMostRecentOperation(operation => {
+      return MockPayloadGenerator.generate(operation, mockResolvers)
+    })
```
erikdstock, line 156: Changes to the RTL setup:
- `mockRejectLastOperation()` returns the operation and its name/vars just like resolve version does
- operation does not need safe navigation operator
- (minor) render view after defining helpers

erikdstock (commented):
Linking to only substantive changes; the rest were done using various find/replace/`grep`/`sed` and `type-check` whack-a-mole hacks.

damassi:
Awesome change, thanks so much @erikdstock

PR #13120, opened 2023-11-09
chore: EMI-1552 Shipping with no saved addresses

erikdstock:

This PR continues the shipping refactor with a focus on shipped orders with no saved addresses.

Completes [EMI-1521], [EMI-1522] and [EMI-1563].

Because of a change in how the new shipping route works to remove extra `useState` variables, this PR required us to update how we send mutations from the shipping route. In addition, the existing tests had to be updated to remove our mocks of the (artsy) `commitMutation` helpers. The result is a big diff but a less complex route overall.

The changes are split in 3 main pieces (prior to any feedback)
- [x] 24428ac9e7aa44db92f25eb053f625e06426815f - Refactor to use `useMutation` - drops legacy mutations, puts mutation helpers in shipping context, creates a support folder to make our context object easier to understand
- [x] 31970911a18b9f6de0efac830b724cd2a728c204 - Updates AddressModal2 tests to use new mutation helpers for testing. In order to do this I needed to use the `setupTestWrapper` helper, which is a little awkward since this helper assumes you are rendering a fragment container. I just fetch and ignore some data. After that it works perfectly.
- [x] Update our shipping route tests to drop commitMutation and fix tests.

The last commit(s) will finish the PR, but the first one in particular includes the major changes to implementation.

[EMI-1521]: https://artsyproduct.atlassian.net/browse/EMI-1521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1522]: https://artsyproduct.atlassian.net/browse/EMI-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1563]: https://artsyproduct.atlassian.net/browse/EMI-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -39,11 +39,11 @@ import { useFeatureFlag } from "System/useFeatureFlag"
 import { SavedAddressesFragmentContainer as SavedAddresses } from "Apps/Order/Routes/Shipping2/SavedAddresses2"
 import { usePrevious } from "Utils/Hooks/usePrevious"
 import { Collapse } from "Apps/Order/Components/Collapse"
-import { useShippingContext } from "Apps/Order/Routes/Shipping2/ShippingContext"
+import { useShippingContext } from "Apps/Order/Routes/Shipping2/support/ShippingContext"
```
damassi, line 42: Never seen a "support" directory concept in force before, so lets not introduce new patterns

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -45,17 +44,13 @@ import {
...
-  FulfillmentType,
-  ShippingAddressFormValues,
-} from "Apps/Order/Routes/Shipping2/shippingUtils"
+  useComputeShippingContext,
+} from "Apps/Order/Routes/Shipping2/support/ShippingContext"
+
+import { FulfillmentType } from "Apps/Order/Routes/Shipping2/support/shippingUtils"
 import { usePrevious } from "Utils/Hooks/usePrevious"
 import { useOrderTracking } from "Apps/Order/Utils/useOrderTracking"
+import { log, warn } from "console"
```
damassi, line 53: console.log, console.warn; we don't pluck these anywhere else, so we should start now

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -206,30 +201,11 @@ export const ShippingRoute: FC<ShippingProps> = props => {
...
-      )
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [props.me.id, relayEnvironment]
-  )
-
+  const {
+    saveFulfillmentDetails: saveFulfillmentDetailsMutation,
+    createSavedAddress: createSavedAddressMutation,
+    selectShippingQuote: selectShippingQuoteMutation,
```
damassi, line 207: 1:2. I'm opening a piggy bank

src/Apps/Order/Routes/Shipping2/support/ShippingContext/ShippingContext.ts
```diff
@@ -61,10 +54,10 @@ export const useShippingContext = () => {
 // Get information from the order and user, computedOrderData (shared, computed context about order)
 // initial values for forms, and the current step in the shipping route
 // values for forms
-export const useComputeOrderContext = (
+export const useComputeShippingContext = (
   props: ShippingProps
 ): ShippingContextProps => {
-  const computedOrderData = useComputeOrderData(props)
+  const computedOrderData = useProcessOrderData(props)
```
damassi, line 60: 1:2. Another coin dropped in the slot

src/Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/index.tsx
```diff
@@ -0,0 +1,23 @@
+import { useCreateSavedAddress } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useCreateSavedAddress"
+import { useUpdateSavedAddress } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useUpdateSavedAddress"
+import { useDeleteSavedAddress } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useDeleteSavedAddress"
+import { useSaveFulfillmentDetails } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useSaveFulfillmentDetails"
+import { useSelectShippingQuote } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useSelectShippingQuote"
+import { useUpdateUserDefaultAddress } from "Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useUpdateUserDefaultAddress"
+
+export const useShippingMutations = () => {
```
damassi, line 8: a) we don't use index files; b) main named export should be the name of the file, so that its discoverable; c) if we can import mutations from the mutations hook, we don't need to pass it into the context. Just import the hook directly and use that where necessary

src/Apps/Order/Routes/Shipping2/index.tsx
erikdstock: ah yes, that was from earlier debugging and i already removed it - got advice from SO that sometimes when console output doesn't appear in jest you can hack around it (not sure how) by importing from console (but maybe logger would work just as well).

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: ah, i think i just brought that over from rails apps. i'm not stuck on the name, just wanted to start bringing some order to this folder, maybe not necessary right now as you said

src/Apps/Order/Routes/Shipping2/support/ShippingContext/ShippingContext.ts
erikdstock: making note to consider mimicking [similar helper naming in volt](https://github.com/artsy/volt/blob/main/app/javascript/apps/OrderPage/helpers.ts#L265)

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
damassi: Not necessary, but if it were, `utils` go in `Utils`, Context lives in root, `Components` live in components, hooks go in `Hooks`. We don't need any other folders (though `Queries` and `Mutations` could be considered an exception). If there's ever any question, the `Apps/Auction` app is canonical.

src/Apps/Order/Routes/Shipping2/support/ShippingContext/ShippingContext.ts
erikdstock: changed to `useParseOrderData/parsedOrderData`.

src/Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/index.tsx
erikdstock: updated to import the hooks individually. I'd like to make sure they don't re-render since their args don't change but that can wait.

src/Apps/Order/Routes/Shipping2/support/ShippingContext/ShippingContext.ts
damassi: I don't see that in the diff?

src/Apps/Order/Routes/Shipping2/support/ShippingContext/mutations/useSaveFulfillmentDetails.ts
```diff
@@ -0,0 +1,38 @@
+import { useMutation } from "Utils/Hooks/useMutation"
+import { graphql } from "react-relay"
+import { useSaveFulfillmentDetailsMutation } from "__generated__/useSaveFulfillmentDetailsMutation.graphql"
+
+export const useSaveFulfillmentDetails = () => {
```
damassi, line 5: 👏 (tyyy)

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
```diff
@@ -18,11 +18,15 @@ import { AddressModalFields } from "Components/Address/AddressModalFields"
...
 import { ADDRESS_VALIDATION_SHAPE } from "Apps/Order/Routes/Shipping2/FulfillmentDetails"
 import { SavedAddressType } from "Apps/Order/Utils/shippingUtils"
-import { useShippingContext } from "Apps/Order/Routes/Shipping2/support/ShippingContext"
-import { addressWithFallbackValues } from "Apps/Order/Routes/Shipping2/support/shippingUtils"
+import { useShippingContext } from "Apps/Order/Routes/Shipping2/ShippingContext"
+import { addressWithFallbackValues } from "Apps/Order/Routes/Shipping2/shippingUtils"
 import { useCreateSavedAddressMutation$data } from "__generated__/useCreateSavedAddressMutation.graphql"
 import { useUpdateSavedAddressMutation$data } from "__generated__/useUpdateSavedAddressMutation.graphql"
 import createLogger from "Utils/logger"
+import { useCreateSavedAddress } from "Apps/Order/Routes/Shipping2/shippingContextHelpers/mutations/useCreateSavedAddress"
```
damassi, line 26: As mentioned previously, the pattern we have in Force can be [gleaned very precisely from the example app](https://github.com/artsy/force/tree/main/src/Apps/Example) and the [best practices doc](https://github.com/artsy/force/blob/main/docs/best_practices.md). 

- mutations go in `Routes/Shipping2/Mutations/useWhateverMutation.ts` so that they are clearly discoverable and follow the rest of Force
- misc utilities go in `Shipping2/Utils/whateverUtilName.ts`

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
```diff
@@ -118,13 +126,11 @@ export const AddressModal: React.FC<Props> = ({
...
-        variables: {
-          input: { userAddressID: addressID },
-        },
-      })
-      .catch(logger.error)
+    return deleteSavedAddress({
+      variables: {
+        input: { userAddressID: addressID },
+      },
+    }).catch(logger.error)
```
damassi, line 133: we should be using `try` / `catch` everywhere, vs vanilla promises

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
erikdstock: ah sorry, missed this.

erikdstock: Just noting here that I don't think the hooks or their filenames should end in mutation as that causes a naming clash with generated types 
image

damassi: No prob! Its more folder structure; naming is flexible 👍

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+  isDefault: false,
+}
+
+let testAddressModalProps: Props
+
+let mockRelayEnv: ReturnType<typeof createMockEnvironment>
+
+beforeEach(() => {
+  mockRelayEnv = createMockEnvironment()
+  ;(useSystemContext as jest.Mock).mockImplementation(() => {
```
damassi, line 54: `mockUseSystemContext`

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+      isLoggedIn: true,
+      relayEnvironment: mockRelayEnv,
+    }
+  })
+})
+
+const { getWrapper } = setupTestWrapper<AddressModal2TestQuery>({
+  // Cast props here because we don't actually need the query, just the relay
+  // test wrapper. -
+  // props must be passed via getWrapper's componentProps arg
```
damassi, line 66: No need to leave a comment here about this

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
+import {
+  AddressModal,
+  Props,
```
damassi, line 3: `AddressModalProps`

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+    testAddressModalProps = {
+      onSuccess: jest.fn(),
+      modalAction: {
+        type: AddressModalActionType.EDIT_USER_ADDRESS,
+        address: savedAddress,
+      },
+
+      closeModal: jest.fn(),
+    }
+    ;(useSystemContext as jest.Mock).mockImplementation(() => {
```
damassi, line 94: `mockUseSystemContext`

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+      return {
+        user: { lab_features: [] },
+        isLoggedIn: true,
+        relayEnvironment: mockRelayEnv,
+      }
```
damassi, line 99: This doesn't need to be duplicated up above in a top-level `beforeEach`, and here in a `describe` `beforeEach`; lets just keep the one inside of the describe block

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+    })
+  })
+
+  it("renders EditModal with the title, input fields and buttons", () => {
+    const { wrapper } = getWrapper({}, testAddressModalProps, mockRelayEnv)
```
damassi, line 104: must we absolutely pass this relayEnv? can't we just assign it as a default arg in the `getWrapper` function definition?

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+    expect(wrapper.find("select").length).toBe(1)
+
+    expect(wrapper.find("Checkbox[data-test='setAsDefault']").length).toBe(1)
+    expect(wrapper.find("Clickable[data-test='deleteButton']").length).toBe(1)
+    expect(wrapper.find("Button[data-test='saveButton']").length).toBe(1)
+  })
+
+  it("renders EditModal without checkbox when address is default", () => {
+    const { wrapper } = getWrapper(
+      {},
```
damassi, line 117: this empty object first arg is super weird, lets make `getWrapper` take a single object and then just pluck what we need from that:

```tsx
const getWrapper = (props) => {
  // props.foo, props.bar 
}
```

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
```diff
@@ -0,0 +1,340 @@
...
+    expect(wrapper.find("input").length).toBe(7)
+    expect(wrapper.find("select").length).toBe(1)
+
+    expect(wrapper.find("Checkbox[data-test='setAsDefault']").length).toBe(1)
+    expect(wrapper.find("Clickable[data-test='deleteButton']").length).toBe(0)
+    expect(wrapper.find("Button[data-test='saveButton']").length).toBe(1)
+  })
+
+  it("clicking the delete button spawns a correct dialog", () => {
+    const { wrapper } = getWrapper({}, testAddressModalProps, mockRelayEnv)
```
damassi, line 155: ditto here -- `testAddressModalProps` can be another default arg, so it doesn't need to be passed every time:

```tsx
const getWrapper = ({
  addressModalProps = defaultAddressModalProps,
  relayEnvironment = mockRelayEnvironment,
  ...props 
}) => {
  ...
}
```

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
```diff
@@ -114,100 +106,131 @@ export const AddressModal: React.FC<Props> = ({
...
+    try {
+      if (modalAction.type === "createUserAddress") {
+        operation = () => {
+          return createSavedAddress({
+            variables: {
+              input: { attributes: addressInput },
+            },
+          }).then(result => {
+            return handleMutationPayload(result.createUserAddress)
+          })
```
damassi, line 170: No vanilla promises

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
```diff
@@ -114,100 +106,131 @@ export const AddressModal: React.FC<Props> = ({
...
+              },
+            },
+          }).then(result => handleMutationPayload(result.updateUserAddress))
+      }
+      const { data, errors } = await operation()
+      if (errors) {
+        handleErrors(errors, helpers)
+        return
+      }
+      const savedAddressID = data?.internalID
```
damassi, line 189: This is #superminor, but just something I frequently notice when reviewing PRs and something to be aware of for future (other devs) readability  -- there's rarely ever any space between statements. Like compare the below:

```tsx
const addressInput = addressWithFallbackValues(values)
let operation: () => Promise<ReturnType<typeof handleMutationPayload>>
try {
  ...
catch () { 
  ...
}
const data = await operation()
if (errors) {
 ... 
}
const savedAddressID = foo 
```

vs 

```tsx 
const addressInput = addressWithFallbackValues(values)
let operation: () => Promise<ReturnType<typeof handleMutationPayload>>

try {
  ...
catch () { 
  ...
}

const data = await operation()

if (errors) {
 ... 
}

const savedAddressID = foo 
```

In the latter, a dev reading yr code can _scan_, and notice when the scenery changes. New things are signified by significant white space. With everything constantly (or arbitrarily) squished together, folks reading code in the future are forced to squint. 

Its a non-blocking thing, but legibility is really important when working with non-trivial code on a larger team. (This also goes for PR descriptions where paragraph breaks are frequently omitted between key sections, making things feel needlessly sloppy.)

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+  step: ShippingRouteStep
+  helpers: {
+    fulfillmentDetails: FulfillmentHelpers & {
+      setFulfillmentFormHelpers: (helpers: FulfillmentHelpers) => void
+    }
+  }
+}
+
+export const ShippingContext = createContext<ShippingContextProps>({} as any)
```
damassi, line 39: I can understand keeping the context itself inside of the useShippingContext hook, but this is unconventional across the app. Contexts should declare themselves -- `Shipping2/ShippingContext.tsx`, and we can either add an additional hook inside of `Shipping2/Hooks/useShippingContext` (which returns the context), or just have the hook exported directly from `ShippingContext`. We do it both ways.

damassi: Also, no `as any` here. Its better to cast as `unknown` first, and then as the correct type:

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
```diff
@@ -0,0 +1,173 @@
...
+
+const useFulfillmentDetailsHelpers = (): ShippingContextProps["helpers"]["fulfillmentDetails"] => {
+  // Because there is a single button for both fulfillment details and
+  // shipping quote steps (and duplicated in the sidebar)
+  // we need to hack some formik values UP from the fulfillment details
+  // form.
+  // handleSubmit: Used to submit the form
+  // isValid: Used to disable the button
+  // values: Used to get the form values for un-saving the address if the user
+  //   unchecks it after saving it in the fulfillment details step.
```
damassi, line 81: I'm getting nitpicky here, but this is conventional code quality standards stuff, so please bear with me.  

Please use paragraph breaks more often for detailed comments, if you want folks to read them. Everything is all smooshed together. You're describing something, then on the very next line, with no break in between, there's some kind of psudo doc-block style argument descriptions. Hit enter 🙏 And ask yourself if those argument comments are necessary here, versus putting them right above where the definition actually is. There it will be seen, and more useful. 

```
// Used to submit the form
handleSubmit: () => {},
// Used to disable the button
isValid: false 
```

Also, in JavaScript, there's multi-line comments, which are very useful. All you have to do is type `/**` and hit enter and VSCode will auto expand. For multi-line comments, please use multiline comments.

Lastly, per our VSCode config, there is a right margin guide that should be visible, set to 80 character width. When writing code comments, write them out to that guide. Everything feels so arbitrary which makes legibility hard to read. Prettier automatically hits the 80 character margin and formats, but with comments its the devs responsibility. Best just to get into the habit.

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
```diff
@@ -0,0 +1,173 @@
...
+    parsedOrderData,
+    initialValues,
+    step,
+    helpers: {
+      fulfillmentDetails: useFulfillmentDetailsHelpers(),
+    },
+  }
+}
+
+const useFulfillmentDetailsHelpers = (): ShippingContextProps["helpers"]["fulfillmentDetails"] => {
```
damassi, line 73: Is it necessary for this to be in its own hook, or can this just be in the main context?

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
```diff
@@ -0,0 +1,173 @@
...
+  const [fulfillmentFormHelpers, setFulfillmentFormHelpers] = useState<
+    Pick<FormikProps<FulfillmentValues>, "handleSubmit" | "isValid" | "values">
+  >({
+    handleSubmit: () => {},
+    isValid: false,
+    values: {
+      attributes: {
+        saveAddress: false,
+      },
+    } as any,
```
damassi, line 91: No `as any`

damassi (approved):
This is real nice @erikdstock 🙏 

Really appreciate swapping out the tests as well as all of the mutations. 

Most of my comments are minor readability things (but important in terms of a long-term habit). But if we could update that `getWrapper` signature in the tests to just accept a single object argument then this is 👍 from me.

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
erikdstock: yeah I completely agree with this and am normally way more nitpicky about it 👍

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
erikdstock: Would it make sense to allow setting up a test wrapper without a query? Not blocking this PR of course.

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
erikdstock: Updated to redeclare the function locally, lmk if that's what you have in mind (once I push)

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
erikdstock: Thanks and yeah, I wondered about this, ultimately decided to stay simple for now since ShippingContext.Provider is *technically* a component, but in practice is mostly about the hook. Will move out.

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
erikdstock: good call

erikdstock:
@damassi addressed all your comments (I assumed the one about `getWrapper` meant to redefine it locally, not across the whole repo). Gonna manually squash on green, thanks for your feedback.

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
erikdstock: Went through the main files and tried to reformat comments that were still doing something, remove others that went on too long, left a few in place that include old code references and need to be addressed in upcoming work. I'll try switch to multi-line when a comment busts through the first line in the future

src/Apps/Order/Routes/Shipping2/AddressModal2.tsx
erikdstock: Addressed it in this place, will keep my eyes out for other big functions that could benefit from it in the future

src/Apps/Order/Routes/Shipping2/AddressModal2.jest.tsx
damassi: Its iffy :) Lets think about it outside of this PR. Rare case where this is needed

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
```diff
@@ -0,0 +1,154 @@
...
+): ShippingContextProps => {
+  const parsedOrderData = useParseOrderData(props)
+
+  /**
+   * Because there is a single button for both fulfillment details and
+   * shipping quote steps (and duplicated in the sidebar)
+   * we need to hack some formik values UP from the fulfillment details form.
+   *
+   * Currently we need to pass up:
+   */
```
damassi, line 38: 🙏

damassi (approved):
PR looks great @erikdstock, thanks for working through all of those changes 👍 Merge when ready!

damassi:
@erikdstock - i wonder if the failing build has to do with staging copy or something? Not sure why acceptance tests are failing around auctions 🤷

PR #13129, opened 2023-11-13
feat: Create alert logged-out experience

olerichter00:
Addresses [ONYX-506]

## Description

This PR implements the "Create Alert" logged-out experience for the new alert flow. 

[ONYX-506]: https://artsyproduct.atlassian.net/browse/ONYX-506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty.tsx
+++ b/src/Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty.tsx
@@ -26,6 +26,20 @@ const ArtistWorksForSaleEmpty: FC<ArtistWorksForSaleEmptyProps> = ({
   const { filters } = useArtworkFilterContext()
 
   const { alertComponent, showAlert } = useAlert({
+    authDialogOptions: {
+      options: {
+        title: "Sign up to create your alert",
+        afterAuthAction: {
+          action: "createAlert",
+          kind: "artist",
+          objectId: artist.internalID,
+        },
+      },
+      analytics: {
+        contextModule: ContextModule.artworkGrid,
+        intent: Intent.createAlert,
+      },
+    },
     initialCriteria: filters,
   })
   const newAlertModalEnabled = useFeatureFlag("onyx_artwork_alert_modal_v2")
--- a/src/Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty.tsx
+++ b/src/Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty.tsx
@@ -26,20 +26,6 @@ const ArtistWorksForSaleEmpty: FC<ArtistWorksForSaleEmptyProps> = ({
   const { filters } = useArtworkFilterContext()
 
   const { alertComponent, showAlert } = useAlert({
-    authDialogOptions: {
-      options: {
-        title: "Sign up to create your alert",
-        afterAuthAction: {
-          action: "createAlert",
-          kind: "artist",
-          objectId: artist.internalID,
-        },
-      },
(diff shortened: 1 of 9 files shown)


src/Apps/Artist/Routes/WorksForSale/Components/ArtistWorksForSaleEmpty.tsx
```diff
@@ -26,6 +26,20 @@ const ArtistWorksForSaleEmpty: FC<ArtistWorksForSaleEmptyProps> = ({
   const { filters } = useArtworkFilterContext()
 
   const { alertComponent, showAlert } = useAlert({
+    authDialogOptions: {
```
dblandin, line 29: **suggestion**: I'm wondering if we need to pass these option into `useAlert` at all. The hook might be able to figure out the `afterAuthAction` attributes on its own. I'm also not sure if we're getting much value out of explicitly setting the contextModule at this level — especially given that we're already tracking an event for clicking on "Create alert".

dblandin: Added a PR against this feature branch with some adjustments for consideration: https://github.com/artsy/force/pull/13133

olerichter00: That would be a great improvement and will make the usage of the hook much cleaner 🙏  

I will have a look later and try to integrate it into this PR.

We might need to adjust the logic for when the `alertComponent` is already wrapped with `ProgressiveOnboardingAlertCreateSimple` ([example](https://github.com/artsy/force/blob/e4c6686d623c61e0360c90115fee0b55b0cbf077/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx#L343)).

src/Utils/Hooks/useAuthIntent/index.tsx
```diff
@@ -16,6 +16,7 @@ export const AFTER_AUTH_ACTION_KEY = "afterSignUpAction"
 
 export type AfterAuthAction =
   | { action: "associateSubmission"; kind: "submission"; objectId: string }
+  | { action: "createAlert" }
   | { action: "createAlert"; kind: "artist"; objectId: string }
   | { action: "createAlert"; kind: "artworks"; objectId: string }
```
dblandin, line 21: I'm fairly certain that we no longer need these `kind` / `objectId` attributes. The page sets up the `initialCriteria` on load and all we need to do is dispatch the modal.

dblandin (dismissed):
LGTM 👍🏼

src/Utils/Hooks/useAuthIntent/index.tsx
olerichter00: It seems like we need the `kind` / `objectId` attributes. If they are not set on the artwork page, the modal doesn't open.

I've added this [commit](https://github.com/artsy/force/pull/13129/commits/a7f56cdc35430d2059ca0e020a219ca861152530) to the PR that makes the adjustment. But I'm happy to look into this together because I don't fully understand how we dispatch the modal.

dblandin: Hmmm, yeah - happy to take a look at this with you today 👍

olerichter00: Actually, it seems to work now without kind / objectId attributes and I cannot reproduce the issue. Maybe I did something else wrong. I'll merge the changes

PR #13135, opened 2023-11-13
feat: mark App Authenticator as preferred 2fa method

mc-jones:
The type of this PR is: Feat

This PR solves [PHIRE-443]

### Description

Add a `Sup` next to the "App Authenticator" option to indicate it is the preferred method for 2fa. 🔒 [Slack discussion](https://artsy.slack.com/archives/C03N12SR0RK/p1699632665246949?thread_ts=1699631939.121759&cid=C03N12SR0RK).

Screenshot 2023-11-14 at 9 59 50 AM

[PHIRE-443]: https://artsyproduct.atlassian.net/browse/PHIRE-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
+++ b/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
@@ -3,6 +3,7 @@ import {
   Button,
   Flex,
   ModalDialog,
+  Pill,
   Spacer,
   Text,
   useToasts,
@@ -167,7 +168,12 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
         flexDirection={["column", "row"]}
       >
         <Box flexBasis="50%">
-          <Text variant={["md", "lg"]}>App Authenticator</Text>
+          <Flex flexDirection={"row"} alignItems="center">
+            <Text variant={["md", "lg"]}>App Authenticator</Text>
+            <Pill mx={4} size="small" variant="gray">
+              Preferred
+            </Pill>
+          </Flex>
 
           {enabledSecondFactorLabel && (
             <Text variant={["md", "lg"]} color="black60">
--- a/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
+++ b/src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
@@ -3,8 +3,8 @@ import {
   Button,
   Flex,
   ModalDialog,
-  Pill,
   Spacer,
+  Sup,
   Text,
   useToasts,
 } from "@artsy/palette"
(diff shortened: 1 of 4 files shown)


src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
```diff
@@ -167,7 +168,12 @@ export const AppSecondFactor: React.FC<AppSecondFactorProps> = ({
         flexDirection={["column", "row"]}
       >
         <Box flexBasis="50%">
-          <Text variant={["md", "lg"]}>App Authenticator</Text>
+          <Flex flexDirection={"row"} alignItems="center">
+            <Text variant={["md", "lg"]}>App Authenticator</Text>
+            <Pill mx={4} size="small" variant="gray">
+              Preferred
+            </Pill>
```
dzucconi, line 175: I don't think we want to use a `Pill` for this info. What does design say? `Sup` with brand blue might be better?

mc-jones: >What does design say?

I dropped a message in 🔒 [slack](https://artsy.slack.com/archives/C02BGC5PW/p1699892493938409) before this PR and the response was essentially that we can make the call. 

> Sup with brand blue might be better?

This makes sense and was my first thought, but have only seen it used for counts, so went with this alternative. Happy to switch over to a `Sup` component if you think this is better.

Screenshot 2023-11-13 at 12 49 24 PM

src/Utils/isArtsyEmail.ts
```diff
@@ -1,3 +1,7 @@
-export const isArtsyEmail = (email: string) => {
+export const isArtsyEmail = (email: string | null | undefined): boolean => {
+  if (!email) {
+    return false
+  }
+
   return /@artsymail.com$/.test(email)
```
dzucconi, line 6:
```tsx
return email && /@artsymail.com$/.test(email)
```

src/Apps/Settings/Routes/EditSettings/Components/SettingsEditSettingsTwoFactor/TwoFactorAuthentication/Components/AppSecondFactor/index.tsx
dzucconi: 👍

dzucconi (dismissed):

src/Utils/isArtsyEmail.ts
mc-jones: Just want to clarify the intention here as this will return `falsy` values and so we could end up returning `null`, `underfined` or `""`. I would think we would want to always return a boolean.

Or do we just want it to be a one-liner? In which case I can switch to a ternary:

```JS
return email ? /@artsymail.com$/.test(email) : false
```

dzucconi: Oh, sorry, I just want it to not be something like: if (true) { return false } — which I find deeply weird to read.

```tsx
return !!email && /@artsymail.com$/.test(email)
```

PR #13152, opened 2023-11-16
fix(search): uses cover artworks for avatar images on artists

dzucconi:
Closes [DIA-192](https://artsyproduct.atlassian.net/browse/DIA-192)

| Before | After |
| --- | --- |
|  |  |

So this winds up being the minimal change we can make here. Swapping out the `imageUrl` field for just the artist type winds up being really painful in Metaphysics and of very limited utility. I'm just going to follow up with some deprecations and call this one. We will have to address Eigen separately.

The search could/should be simplified a lot by just switching to use fragment containers + matchConnection. I may also follow up with that.

[DIA-192]: https://artsyproduct.atlassian.net/browse/DIA-192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Search/SearchBarInput.tsx
+++ b/src/Components/Search/SearchBarInput.tsx
@@ -66,16 +66,16 @@ const SearchBarInput: FC<SearchBarInputProps> = ({
   const formattedOptions: SuggestionItemOptionProps[] = [
     ...options.map(option => {
       return {
-        text: option.displayLabel!,
-        value: option.displayLabel!,
+        text: option.displayLabel ?? "Unknown",
+        value: option.displayLabel ?? "unknown",
         subtitle:
           getLabel({
             displayType: option.displayType ?? "",
             typename: option.__typename,
           }) ?? "",
-        imageUrl: option.imageUrl!,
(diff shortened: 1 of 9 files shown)


src/Components/Search/SearchBarInput.tsx
```diff
@@ -66,16 +66,16 @@ const SearchBarInput: FC<SearchBarInputProps> = ({
   const formattedOptions: SuggestionItemOptionProps[] = [
```
dzucconi, line 66: This is one of the more obvious tells that something could be simpler + safer with fragment containers: some array `formattedFoo`

egdbear (approved):
I like the fallback for imageUrl.

Should we create follow-up Eigen ticket?

dzucconi:
@egdbear yeah — I thought there was one but I can't seem to locate.

egdbear:
[Done](https://github.com/artsy/force/pull/13152)

PR #13156, opened 2023-11-17
feat(ONYX-529): alert modal UI updates

tam-kis:
The type of this PR is: Fix

This PR solves [ONYX-529](https://artsyproduct.atlassian.net/browse/ONYX-529)

### Description

Update UI for new Alert Modal: limit number of pills, add sticky header and Set Filter button on the Filters modal.
Added animation but this is still being discussed and yet to be refined.

https://github.com/artsy/force/assets/17580625/c87d6256-72a4-49fa-ab3d-5c5c2f26656e

[ONYX-529]: https://artsyproduct.atlassian.net/browse/ONYX-529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Filters/QuickMultipleSelectAlertFilter.tsx
+++ b/src/Components/Alert/Components/Filters/QuickMultipleSelectAlertFilter.tsx
@@ -15,7 +15,6 @@ interface QuickMultipleSelectAlertFilterProps {
 export const QuickMultipleSelectAlertFilter: FC<QuickMultipleSelectAlertFilterProps> = ({
   criteriaKey,
   description,
-  expanded = false,
   label,
   options,
 }) => {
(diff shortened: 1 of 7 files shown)


src/Components/Alert/Components/Filters/QuickMultipleSelectAlertFilter.tsx
```diff
@@ -51,22 +49,20 @@ export const QuickMultipleSelectAlertFilter: FC<QuickMultipleSelectAlertFilterPr
       <Spacer y={2} />
 
       <GridColumns>
-        <ShowMore expanded={expanded}>
```
tam-kis, line 54: @olerichter00 I've removed the "ShowMore" as we're not using it yet and it was causing the spacing on the Medium filter when expanded.

src/Components/Alert/Components/Steps/Details.tsx
```diff
@@ -45,73 +47,89 @@ export const Details: FC = () => {
...
-                  <CriteriaPills />
-                </Flex>
-              </Box>
+          <Box
+            maxWidth={[null, 500]}
+            style={{
+              ...(isMounted
+                ? {
+                    opacity: 1,
+                    transform: "translateX(0)",
```
damassi, line 56: @tam-kis - lets not add animation here if we can avoid it. Having a screen that is snappy and fast is much better than adding transitions for the sake of transitions. If we're going to add UI that transitions between screens, this is something that should be raised up to the design-system working group and ideally be implemented by @dzucconi.

damassi: This is perfectly fine to merge without the animation -- get the functionality in, then we can work it in in a more appropriate fashion later, via #design-system and ideally a ticket in Diamond.

dzucconi: I totally don't have a problem with this transition actually :) — do I wish we had a consistent way of doing it, sure, but this is simple enough and helps communicate the UI change.

damassi: Cool, if yr ok with it then 👍 Just wanted to be sure that things are flowing through the proper channels. Will resolve!

dblandin (approved):

src/Components/Alert/Components/Filters/QuickMultipleSelectAlertFilter.tsx
olerichter00: Good idea to remove it 👍 In case we need it at some point, we can bring it back.

PR #13157, opened 2023-11-17
chore(search): Increase padding around pills

damassi:

### Description

Noticed the vertical padding was very crammed, this increases it from .5 to 1.

Before: 

https://github.com/artsy/force/assets/236943/132ca184-8a8d-4f22-9543-2808bdc63f1f

After:

Screenshot 2023-11-17 at 8 35 19 AM


dzucconi:
While although I do support that: this variant was added specifically for this use case so we'd really need current design input to change.

dzucconi:
Also notice the weird underlining on the superscript count. We wouldn't want to do that.

damassi:
Discussion here: https://artsy.slack.com/archives/C9XJKPY9W/p1700238665414979

For now, threw away the changes and just increased the padding to 1, so that everything isn't so crammed

PR #13158, opened 2023-11-20
feat: Alert filter price histogram

olerichter00:

This PR solves [ONYX-525]

### Description

This PR adds the Alert filter price histogram.

Screenshot 2023-11-20 at 14 03 44

[ONYX-525]: https://artsyproduct.atlassian.net/browse/ONYX-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Filters/Price.tsx
+++ b/src/Components/Alert/Components/Filters/Price.tsx
@@ -72,16 +72,15 @@ export const PriceQueryRenderer = () => {
       lazyLoad
       placeholder={<Price />}
       variables={{ artistID }}
+      // TODO: use rootlevel artworks connection instead to fetch aggregations for all artists
+      // https://github.com/artsy/force/pull/13158#discussion_r1399214348
       query={graphql`
         query PriceAggregationsQuery($artistID: String!) {
           artist(id: $artistID) {
             filterArtworksConnection(
               aggregations: [SIMPLE_PRICE_HISTOGRAM]
-              first: 1
+              first: 0
             ) {
-              counts {
-                total
-              }
               ...Price_aggregations
             }
           }
--- a/src/Components/Alert/Components/Filters/Price.tsx
+++ b/src/Components/Alert/Components/Filters/Price.tsx
@@ -5,9 +5,23 @@ import {
   CustomRange,
   DEFAULT_PRICE_RANGE,
 } from "Components/PriceRange/constants"
+import { Aggregations } from "Components/ArtworkFilter/ArtworkFilterContext"
+import { aggregationsToHistogram } from "Components/ArtworkFilter/ArtworkFilters/PriceRangeFilter"
+import { PriceAggregationsQuery } from "__generated__/PriceAggregationsQuery.graphql"
+import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
+import { createFragmentContainer, graphql } from "react-relay"
+import { Price_aggregations$data } from "__generated__/Price_aggregations.graphql"
 
-export const Price: React.FC = () => {
+interface PriceProps {
+  aggregations?: Price_aggregations$data | null
+}
+
+export const Price: React.FC<PriceProps> = ({ aggregations }) => {
   const { state, dispatch } = useAlertContext()
+
+  const bars = aggregationsToHistogram(
+    aggregations?.aggregations as Aggregations
+  )
 
   const handlePriceRangeUpdate = (updatedValues: CustomRange) => {
     dispatch({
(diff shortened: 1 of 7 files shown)


src/Components/Alert/Components/Filters/Price.tsx
```diff
@@ -18,11 +32,70 @@ export const Price: React.FC = () => {
...
+    <SystemQueryRenderer<PriceAggregationsQuery>
+      lazyLoad
+      placeholder={<Price />}
+      variables={{ artistID }}
+      query={graphql`
+        query PriceAggregationsQuery($artistID: String!) {
+          artist(id: $artistID) {
+            filterArtworksConnection(
+              aggregations: [SIMPLE_PRICE_HISTOGRAM]
+              first: 1
```
dblandin, line 80: **suggestion**: Since we don't need any artwork data, can we request `first: 0` here?

src/Components/Alert/Components/Filters/Price.tsx
```diff
@@ -18,11 +32,70 @@ export const Price: React.FC = () => {
...
+            ) {
+              counts {
+                total
```
dblandin, line 83: **question**: Are using this `total` value anywhere? I don't see it used outside of this query.

src/Components/Alert/Components/Filters/Price.tsx
```diff
@@ -18,11 +32,70 @@ export const Price: React.FC = () => {
...
+  if (!artistID) return <Price />
+
+  return (
+    <SystemQueryRenderer<PriceAggregationsQuery>
+      lazyLoad
+      placeholder={<Price />}
+      variables={{ artistID }}
+      query={graphql`
+        query PriceAggregationsQuery($artistID: String!) {
+          artist(id: $artistID) {
```
dblandin, line 77: **suggestion**: Instead of using `filterArtworksConnection` within the Artist type, which also means we need to fetch the artist separately, are we able to use the root-level artworksConnection query field instead? That would also allow us to easily thread multiple artistIDs into the query for the infrequent case when an initial alert criteria includes multiple artists, as is this case on some artwork pages:

```graphql
{
  artworksConnection(first: 0, artistIDs: $artistIDs, aggregations: [SIMPLE_PRICE_HISTOGRAM]) {
    ...Price_aggregations
  }
}
```

dblandin (commented):

src/Components/Alert/Components/Filters/Price.tsx
dblandin: I just tried to simulate this with multiple artist IDs and I get the following error from Gravity:

```json
{
  "error": "Price histogram aggregation requires an artist or artist series filter"
}
```

but I feel like we could easily adjust the condition [here](https://github.com/artsy/gravity/blob/15330a2d859281d2c71f9a22607e416657e5aa0e/app/api/v1/filter_endpoint.rb#L93) to also support queries with `artist_ids` supplied. In the meantime (until we discuss and action that Gravity change) we can pluck the first artist ID from the alert context but still use the root-level query field.

olerichter00: If you're ok with it, I'll add a TODO comment, merge this PR and then do the Gravity change.

dblandin: Is it possible to at least switch to the root-level query field but pass the first artist ID, instead of multiple, before merging? Then yeah, I think we can follow-up on that potential Gravity change to support a histogram backed by multiple artist ID arguments.

MounirDhahri (commented):
LGTM - only one minor comment about naming

src/Components/Alert/Components/Filters/Price.tsx
```diff
@@ -18,11 +32,70 @@ export const Price: React.FC = () => {
...
+        }
+      `}
+      render={({ error, props }) => {
+        if (error) {
+          console.error(error)
+        }
+
+        const aggregations = props?.artist?.filterArtworksConnection
+
+        return <PriceFragmentContainer aggregations={aggregations} />
```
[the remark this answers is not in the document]
olerichter00: Good catch! Fixed the naming 👍

MounirDhahri (approved):
LGTM - thanks

src/Components/Alert/Components/Filters/Price.tsx
olerichter00: Gotcha! I've changed the query to use the root artworks connection 👍

dblandin (approved):
This LGTM! One potential refactoring opportunity for later might be to wrap a QueryRenderer only around the histogram bars so the rest of the filter controls can snap into place right away. As is I'm guessing that the price filter controls are inaccessible until that data is queried and the bars render.

olerichter00:
> This LGTM! One potential refactoring opportunity for later might be to wrap a QueryRenderer only around the histogram bars so the rest of the filter controls can snap into place right away. As is I'm guessing that the price filter controls are inaccessible until that data is queried and the bars render.

Good idea! Currently, the price filter component gets rendered without the query renderer as a placeholder. So, the controls should be accessible before the histogram bar is rendered. I'll look into it, and your suggestion might be a bit cleaner.

PR #13159, opened 2023-11-20
feat(onyx-472): add size filter to alert filters

nickskalkin:

This PR solves [ONYX-472]

### Description

Add size filter to alert filters screen.

### Demo

- [Review app](https://alerts-size-filter.artsy.net/)

https://github.com/artsy/force/assets/3934579/48445303-a5ba-4ad8-8011-9cf8883167b5

[ONYX-472]: https://artsyproduct.atlassian.net/browse/ONYX-472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Filters/Size.tsx
+++ b/src/Components/Alert/Components/Filters/Size.tsx
@@ -1,0 +1,230 @@
+import { FC, useState } from "react"
+import {
+  Button,
+  Checkbox,
+  Clickable,
+  Flex,
+  Join,
+  Radio,
+  RadioGroup,
+  Spacer,
+  Text,
+} from "@artsy/palette"
+import { DEFAULT_METRIC, Metric } from "Utils/metrics"
+import { NumericInput } from "Components/NumericInput"
+import { useAlertContext } from "Components/Alert/Hooks/useAlertContext"
+import {
+  DEFAULT_RANGE,
+  getRangeValue,
+  isCustomRangeValue,
+} from "Utils/customRangeUtils"
+import {
+  CustomSize,
+  getCustomSizeRangeInInches,
+  getPredefinedSizesByMetric,
+  parseSizeRange,
+} from "Utils/customSizeUtils"
+import { SearchCriteriaAttributeKeys } from "Components/SavedSearchAlert/types"
+
+export const Size: FC = () => {
+  const { state, dispatch } = useAlertContext()
+  const width = state.criteria.width || DEFAULT_RANGE
+  const height = state.criteria.height || DEFAULT_RANGE
+  const sizes = state.criteria.sizes || []
+  const metric = state.metric || DEFAULT_METRIC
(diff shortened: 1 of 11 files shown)


src/Components/Alert/Components/Filters/Size.tsx
```diff
@@ -0,0 +1,231 @@
...
+          defaultValue={metric}
+          onSelect={handleSelectMetric}
+          flexDirection="row"
+          my={2}
+        >
+          <Radio value="cm" label="cm" flex={1} />
+          <Radio value="in" label="in" flex={1} />
+        </RadioGroup>
+
+        <Flex flexDirection="column">
```
olerichter00, line 148: Looking at the design, this should probably be a two-column design. You can probably use [QuickMultipleSelectAlertFilter](https://github.com/artsy/force/blob/db35f877ccb428046dd1da3df6d3759362be814b/src/Components/Alert/Components/Filters/QuickMultipleSelectAlertFilter.tsx#L14) here.

[Figma Design
](https://www.figma.com/file/lVbIxbnH2IUk1K1FoxyFHV/Filters-and-Alerts-redesign?type=design&node-id=144-32657&mode=design&t=wreNE8xNVYVrRNta-0)
Screenshot 2023-11-23 at 11 18 18

src/Components/Alert/Components/Filters/Size.tsx
```diff
@@ -0,0 +1,231 @@
...
+            </Flex>
+
+            <Button
+              mt={1}
+              variant="primaryGray"
+              onClick={handleSetSizesClick}
+              width="100%"
+            >
+              Set size
+            </Button>
```
olerichter00, line 225: Do we have a design for this part? The button looks quite big 🤔  


Could we simply remove the button and set the custom height and width directly when the user types in the input?

Screenshot 2023-11-23 at 11 24 11

src/Components/Alert/Components/Filters/Size.tsx
nickskalkin: I think this design is for app specifically, right? I've asked Barney about it back than and he said that the filters on web should be the same as on the live site.

I've pinged him in Slack to review your suggestions, thank you!

src/Components/Alert/Components/Filters/Size.tsx
nickskalkin: Same as above, pinged Barney to take a look!

nickskalkin: We've agreed to make this button smaller for desktop and keep it 100% for mobile @olerichter00:

image

dariakoko (approved):
Very clean, thank you!

olerichter00 (approved):
Looks great 🌟

PR #13168, opened 2023-11-21
fix(sticky): supports proper stacking and release

dzucconi:
Some work towards supporting sticky horizontal filters.

This took a moment to wrap my head around — we always want to register all Stickies on initialization, but we need to filter out any that may be "released" due to intercepting a bottom boundary, when getting the top offset.

I'm unable to merge sticky wrapped around the filters though because the drawer isn't actually implemented in a modal fashion (with portal).

### DIFF
--- a/src/Components/Sticky/StickyProvider.tsx
+++ b/src/Components/Sticky/StickyProvider.tsx
@@ -10,20 +10,23 @@ import {
 } from "react"
 import * as React from "react"
 
-type Sticky = {
+export type TSticky = {
   id: string
   height: number
+  status: "FIXED" | "ORIGINAL" | "RELEASED"
 }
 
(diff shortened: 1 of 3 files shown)


src/Components/Sticky/StickyProvider.tsx
```diff
@@ -10,42 +10,60 @@ import {
 } from "react"
 import * as React from "react"
 
-type Sticky = {
+export type TSticky = {
```
damassi, line 13: Whats the thinking with the T prefix (which we don't really use elsewhere)?

damassi: Ah, i see that there's another Sticky export for the component.

PR #13182, opened 2023-11-22
chore(eslint): Add rule guarding against relay-runtime `graphql` import

damassi:

### Description

With latest relay, the graphql import needs to come from `react-relay`, not `relay-runtime` for typescript to be satisfied. This also applies to a few other things (like `commitMutation`, etc). However, using those things directly only apply to legacy patterns given our current set of tools, so should be fine.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -109,6 +109,11 @@ module.exports = {
         name: "react-waypoint",
         message: "Please use `useIntersectionObserver`",
       },
+      {
+        name: "react-runtime",
+        importNames: ["graphql"],
+        message: "Please import `graphql` from `react-relay`.",
+      },
     ],
     "no-relative-import-paths/no-relative-import-paths": [
       "warn",


araujobarret:
I don't remember which bug was but @pvinis mentioned some weird bugs because some components were using `graphql` or other import from the `relay-runtime` instead of `react-relay`. In any case, is a good safeguard 💪

damassi:
Ohh interesting, i've never noticed this but def keep an eye 👍

MounirDhahri:
Nice - This can be useful for App as well

.eslintrc.js
```diff
@@ -109,6 +109,11 @@ module.exports = {
         name: "react-waypoint",
         message: "Please use `useIntersectionObserver`",
       },
+      {
+        name: "react-runtime",
```
MounirDhahri, line 113: I believe you mean **relay-runtime** here

PR #13191, opened 2023-11-28
fix: Fix for artist series pills not being displayed on the edit alerts page

nickskalkin:

### Description

Pills were not displayed on edit alert pages before.

### Demo

Here is a video which demonstrated the whole user flow working:

https://github.com/artsy/force/assets/3934579/8c9a1956-104b-4091-9bce-5556b975de35


anandaroop (approved):
Nice! 

Looks good to me, but we should probably rebuild this once the GH/Circle issue is sorted.

PR #13192, opened 2023-11-28
feat: Update notification preference defaults in alert modal

olerichter00:
Addresses [ONYX-552]

## Description

This PR updates the notification preference defaults in the alert modal.

**Email**
- Default: If user is subscribed to the right subscription group (Custom alerts), then enabled otherwise disabled.
- If toggle is enabled and user is not subscribed to the right subscription group, show warning and prompt to update email subscription status (currently used in edit flow).

**Push:** 
- Always disabled by default

## Screenshots

| If user is subscribed (Default)  | If user is not subscribed and checked box |
| --- | --- |
|Screenshot 2023-11-28 at 10 45 51 | Screenshot 2023-11-28 at 10 46 18|

[ONYX-552]: https://artsyproduct.atlassian.net/browse/ONYX-552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Alert/Components/Filters/NotificationPreferences.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { NotificationPreferences_viewer$data } from "__generated__/NotificationPreferences_viewer.graphql"
+import { NotificationPreferencesQuery } from "__generated__/NotificationPreferencesQuery.graphql"
+import { AlertFormikValues } from "Components/Alert/Components/Steps/Details"
+import { useFormikContext } from "formik"
+
+interface NotificationPreferencesProps {
+  viewer?: NotificationPreferences_viewer$data
+}
+
+export const NotificationPreferences: FC<NotificationPreferencesProps> = ({
```
olerichter00, line 24: Removed the component `[EmailPreferenceWarningMessage.tsx](https://github.com/artsy/force/pull/13192/files#diff-0f54fb26299f1257bcea26eff70538e9ff4d01d7bb6f07246300d8ab18f65c5a)` and created a new component with a query renderer for the notification settings.

src/Components/Alert/Components/Filters/NotificationPreferences.tsx
```diff
@@ -0,0 +1,132 @@
...
+      return (
+        preference.channel === "email" &&
+        preference.name === "custom_alerts" &&
+        preference.status === "SUBSCRIBED"
+      )
+    }
+  )
+
+  useEffect(() => {
+    setFieldValue("email", areCustomAlertsEmailNotificationsEnabled)
```
olerichter00, line 42: Updating the `email` value in an effect because I cannot set the default value from this component.

dblandin (commented):
Whoops! Never submitted this for review. I wonder if we should hold on this change until the edit support is merged in. I'm not sure how compatible this approach will be with the same components and provider used for an already persisted alert, in which case we don't need to modify any initial form values.

dblandin (approved):
Nice — thanks for updating to account for edit behavior! 👍🏼

PR #13203, opened 2023-11-29
fix: adjust sellers bio and photo

MounirDhahri:

### Description

This PR adjusts the sellers' bio and photo.


MounirDhahri:
cc @dariakoko @olerichter00 this comes as a request from Wendy

olerichter00 (approved):
👍

olerichter00:
Do we need to make the same change in Eigen (by updating the data in S3, if I remember correctly)?

PR #13214, opened 2023-11-30
feat: Reload artist page correctly when navigating from same artist page

olerichter00:
Addresses [ONYX-551]

## Description

Reload artist page correctly when navigating from same artist page.

From the Alert modal confirmation footer we are linking to the same artist page but with different filters (passed as query params in the URL) [here](https://github.com/artsy/force/blob/9f1078cdfebf737af3a0d2b173cd77ccf43f890b/src/Components/SavedSearchAlert/Components/ConfirmationStepFooter.tsx#L34).

The issue at the moment is that the filter context doesn't get remounted and therefore doesn't get updated (because the internal filter state doesn't get reinitialized) when navigating to the same artist page. By adding a key to the filter context component, we ensure that the component remounts whenever the URL changes. This way, the filters are set correctly on the page.

The artwork grid already gets updated correctly because of Relay 🙌 

[ONYX-551]: https://artsyproduct.atlassian.net/browse/ONYX-551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/Steps/Confirmation.tsx
+++ b/src/Components/Alert/Components/Steps/Confirmation.tsx
@@ -6,7 +6,7 @@ import { ConfirmationArtworksGridQueryRenderer } from "Components/SavedSearchAle
 import { CriteriaPills } from "Components/Alert/Components/CriteriaPills"
 
 export const Confirmation: FC = () => {
-  const { state } = useAlertContext()
+  const { dispatch, state } = useAlertContext()
 
   return (
     <Flex flexDirection="column" p={2}>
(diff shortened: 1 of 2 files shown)


src/Components/Alert/Components/Steps/Confirmation.tsx
```diff
@@ -25,7 +25,9 @@ export const Confirmation: FC = () => {
           </Join>
         </Flex>
         <ConfirmationArtworksGridQueryRenderer
-          onClose={() => false}
+          onClose={() => {
+            dispatch({ type: "RESET" })
+          }}
```
olerichter00, line 30: To ensure that the modal closes when clicking the "Show all matching artworks" button, which links to the artist page with adjusted filters as query parameters.

dblandin (approved):
LGTM 👍🏼

PR #13215, opened 2023-11-30
feat: allow routing to edit panel of alerts app

dblandin:

This PR solves [ONYX-548](https://artsyproduct.atlassian.net/browse/ONYX-548)

### Description

Allow explicit routing to `/settings/alerts/:id/edit`.

Supports routing to alerts not fetched within the list panel by fetching the alert data separately in a useEffect hook.

### Screen Recording

https://github.com/artsy/force/assets/123595/7c3f6b59-123d-4c59-a405-b27eb5b353e2

[ONYX-548]: https://artsyproduct.atlassian.net/browse/ONYX-548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
@@ -9,10 +9,12 @@ import {
 } from "@artsy/palette"
 import {
   createPaginationContainer,
+  Environment,
+  fetchQuery,
   graphql,
   RelayPaginationProp,
 } from "react-relay"
-import { useState } from "react"
+import { useEffect, useState } from "react"
 import { SavedSearchAlertsApp_me$data } from "__generated__/SavedSearchAlertsApp_me.graphql"
 import { Media } from "Utils/Responsive"
 import { EditAlertEntity } from "./types"
@@ -30,6 +32,9 @@ import { Sticky } from "Components/Sticky"
 import { SavedSearchAlertEditFormMobile } from "./Components/SavedSearchAlertEditFormMobile"
 import { useTracking } from "react-tracking"
 import { ActionType } from "@artsy/cohesion"
+import { useRouter } from "System/Router/useRouter"
+import { useSystemContext } from "System/SystemContext"
+import { SavedSearchAlertsApp_Alert_Query } from "__generated__/SavedSearchAlertsApp_Alert_Query.graphql"
 
 interface SavedSearchAlertsAppProps {
   me: SavedSearchAlertsApp_me$data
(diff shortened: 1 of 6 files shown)


olerichter00 (dismissed):
Looks great!

I added functionality to scroll to the correct item in the list (if it is loaded) in this [commit](https://github.com/artsy/force/pull/13215/commits/e10afb3439c5e216999e237f420c3a5ef55a65ef). Please have a look and feel free to remove the commit if we don't need it.

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -48,16 +54,20 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
...
   const [sort, setSort] = useState("CREATED_AT_DESC")
   const [loading, setLoading] = useState(false)
   const alerts = extractNodes(me.savedSearchesConnection)
   const isEditMode = editAlertEntity !== null
+  const { match } = useRouter()
+  const { jumpTo } = useJump()
 
   const closeEditForm = () => {
     setEditAlertEntity(null)
+    history.pushState({}, "", "/settings/alerts")
```
damassi, line 70: we should use our router unless we're trying to (silently) update the url bar and not perform any actions. 

```tsx
router.push('/settings/alerts')
```

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -126,11 +136,53 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
...
+              artistIDs
+              userAlertSettings {
+                name
+              }
+            }
+          }
+        }
+      `,
+      { searchCriteriaID: searchCriteriaID }
+    )?.subscribe?.({
```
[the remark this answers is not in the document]
dblandin: Yeah — I remember wrestling with this in another PR and found that the relay team [discourages](https://relay.dev/docs/api-reference/fetch-query/#behavior-with-topromise) the use of `toPromise()`, primarily because it's not possible to cancel the request as we're doing here by returning a function that calls `.unsubscribe()`.

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
damassi: This would actually be a nice framework addition since we do this in a number of places around force (and I had to add something similar to convos). 

Would you be willing to [update the `useRouter` hook](https://github.com/artsy/force/blob/main/src/System/Router/useRouter.tsx#L13) with two new functions -- `silentPush` and `silentReplace` which uses the code above?

damassi: (Merged https://github.com/artsy/force/pull/13217 -- tyyy )

damassi (approved):

PR #13217, opened 2023-11-30
chore: add silentPush and silentReplace to useRouter hook

dblandin:

### Description

We've occasionally needed to perform a silent URL update instead of performing a full navigation event via our routing layer. This commit adds two new functions, `silentPush` and `silentReplace`, to consolidate our pattern for doing so in one place.

Follow-up on this [review comment](https://github.com/artsy/force/pull/13215#discussion_r1410897022).


damassi (approved):
Thank you so much! Looks 👌

damassi:
Yep exactly what I was thinking. In the future if we can figure out how to achieve silent push / replace using the (actual) router, we can just update here, versus all over.

PR #13218, opened 2023-11-30
feat: add artist series horizontal filter

anandaroop:

This PR solves [ONYX-567]

### Description

Per discussion with @jpoczik — this adds the new Artist Series filter as one of the preferred quick filters in the sticky horizontal filters section above artwork grids.

https://github.com/artsy/force/assets/140521/c6558c2a-939a-467c-841c-7fe095e40b87

This new filter will appear if and only if:

- The feature flag is enabled
- Artist Series aggregations are present (which will only be true on artist pages currently)

### When feature flag is enabled

|Artist series are present|Artist series are absent|Non-artist page|
|---|---|---|
|presen|absen|nonen|

### When feature flag is disabled

|Artist series are present|Artist series are absent|Non-artist page|
|---|---|---|
|presdis|absdis|nondis|

[ONYX-567]: https://artsyproduct.atlassian.net/browse/ONYX-567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


anandaroop:
> Why does it require its own independent flag?

The flag is for the larger feature of artist series filtering + alerting. So it already governs a bunch of stuff,  but now this horizontal filter as well.

PR #13226, opened 2023-12-04
fix: Do not reset email preferences after submitting

olerichter00:
## Description

We should not reset the email preferences after submitting because the update in Gravity [runs in a background job](https://github.com/artsy/gravity/blob/35fd62b809c3e40fb09c7a948fd56aa9b8022a43/app/api/v1/notification_preferences_endpoint.rb#L30), and the preferences won't be updated immediately.

Even reading the return data from the mutation will return the previously set preferences. It takes a few seconds until the backend returns the updated preferences.

This PR doesn't fix the underlying issue but at least it fixes the form by not resetting the value.

https://github.com/artsy/force/assets/4691889/ec2454d6-c4b3-445b-8f0e-657b0a5043dd

### DIFF
--- a/src/Apps/Preferences/PreferencesApp.tsx
+++ b/src/Apps/Preferences/PreferencesApp.tsx
@@ -75,7 +75,6 @@ export const PreferencesApp: FC<PreferencesAppProps> = ({ viewer }) => {
         Email Preference Center
       </Text>
       <Formik<FormValuesForNotificationPreferences>
-        // @ts-ignore
         initialValues={{ ...NOTIFICATION_FIELDS, ...initialValues }}
         onSubmit={async (values, actions) => {
           try {
@@ -94,9 +93,7 @@ export const PreferencesApp: FC<PreferencesAppProps> = ({ viewer }) => {
               variables: { input: { authenticationToken, subscriptionGroups } },
             })
 
-            actions.resetForm({
-              touched: {},
-            })
+            actions.setTouched({})
 
             sendToast({
               variant: "success",


dzucconi:
I'm assuming this changed to be backgrounded at some point?

olerichter00:
> I'm assuming this changed to be backgrounded at some point?

I wonder how 🤔  The preferences are stored in Braze and get updated async via Sidekiq (context: https://github.com/artsy/gravity/pull/15092). 

Do you have an idea? I guess, the [mutation](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/notification_preferences.ts#L109) could return an optimistic response with the new preferences without waiting for the Sidekiq job.



https://github.com/artsy/gravity/blob/35fd62b809c3e40fb09c7a948fd56aa9b8022a43/app/api/v1/notification_preferences_endpoint.rb#L30

olerichter00:
@jonallured, you might have some context on why the email preferences update happens in a Sidekiq job ([PR](https://github.com/artsy/gravity/pull/15092)). Would it be possible to do the update synchronously in order to return the updated preferences?

If this is not possible, we may consider adding a message like "Updating the preferences may take a few seconds" in the success message to inform the user that these updates will not happen immediately. 

This would be good because we prompt users to update their email preferences in the "Create Alert" form (if alert notification emails are disabled) and add a link to this page ([Jira Ticket](https://artsyproduct.atlassian.net/browse/ONYX-552)).

Screenshot 2023-12-05 at 16 04 00

jonallured:
> Would it be possible to do the update synchronously in order to return the updated preferences?

Sadly no - the Braze api call happens async. This is a case where we might want to do something like an optimistic update but I don't really know what that means in terms of relay, etc.

olerichter00:
The solution proposed by this PR implements the optimistic response by simply not reading the response from the mutation. The state of the form remains unchanged and reflects the updated values.

We could allow Metaphysics to return an optimistic response as well, but it would not accurately represent the current state of the backend. Additionally, it would not provide any benefits because refreshing the page would reload the data, potentially showing the outdated preferences. So I would say it's best to not implement something like this.

PR #13230, opened 2023-12-04
fix: router import for partner offers

leamotta:

### Description

Addressing pending comments from https://github.com/artsy/force/pull/13204


damassi:
Thanks for the quick follow-up 🙏

PR #13231, opened 2023-12-04
Chore: [EMI-1523] artsy shipping with no saved addresses

erikdstock:
The type of this PR is: **chore**

This PR solves [EMI-1523], enabling all existing specs for a user who is selecting artsy shipping when they have no saved addresses.

There are a few implementation changes here for the better:
- initial values are no longer in order context, but calculated in the form itself using the parsed order data 
- order of operations when an address is saved, updated or deleted now matches legacy shipping route
- tests now update relay store accurately

One spec was added for a separate ticket - [EMI-1601].

[EMI-1523]: https://artsyproduct.atlassian.net/browse/EMI-1523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1601]: https://artsyproduct.atlassian.net/browse/EMI-1601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13232, opened 2023-12-04
fix: generate and fix types

erikdstock:
Follow-up to #13231 - #minor fix to re-generate some relay types affected by that change.


erikdstock:
#squashongreen

starsirius (approved):
Looks like a test failed with Jest timeout. Otherwise, LGTM!

PR #13234, opened 2023-12-05
fix: failed payment SCA (EMI-1214)

iskounen:
The type of this PR is: Fix

It fixes the SCA flow in the Fix Failed Payment flow by handling the post-SCA mutation to Exchange inline (rather than calling `onContinue` a second time).

This was needed because the second call to `onContinue` did not have access to the same state and ref values that the first call had, so SCA was never completing.

### DIFF
--- a/src/Apps/Order/Routes/NewPayment/index.tsx
+++ b/src/Apps/Order/Routes/NewPayment/index.tsx
@@ -145,9 +145,12 @@ export const NewPaymentRoute: FC<NewPaymentProps & StripeProps> = props => {
               orderId: order.internalID,
             },
           })
-        ).commerceFixFailedPayment?.orderOrError!
+        ).commerceFixFailedPayment?.orderOrError
 
-        if (orderOrError.error) {
+        if (!orderOrError) {
+          handleFixFailedPaymentError("invalid_response")
+          return
+        } else if (orderOrError.error) {
           handleFixFailedPaymentError(orderOrError.error.code)
           return
         }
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -163,7 +166,26 @@ export const NewPaymentRoute: FC<NewPaymentProps & StripeProps> = props => {
             })
             return
           } else {
-            onContinue()
```
iskounen, line 166: We used to called `onContinue` here because we wanted to call `fixFailedPayment` a second time, and the logic to handle that response is the same as the first time.

However, on the second call we were not able to fetch the selected credit card ID from state/refs (and I can't figure out why) so I am making that second `fixFailedPayment` call explicitly here.

This means that if the second `fixFailedPayment` call still `requires_action` that we will fail the operation (instead of showing the SCA modal a second time, and I think that's OK.

src/Apps/Order/Routes/NewPayment/index.tsx
```diff
@@ -289,19 +311,21 @@ export const NewPaymentRoute: FC<NewPaymentProps & StripeProps> = props => {
...
-                  countdownStart={order.lastOffer?.createdAt!}
-                  countdownEnd={order.stateExpiresAt!}
-                />
-              </Flex>
-              <Spacer y={[2, 4]} />
-            </>
-          )}
+          {order.mode === "OFFER" &&
+            order.lastOffer?.createdAt &&
+            order.stateExpiresAt && (
```
iskounen, line 316: This change was made to address a [no-non-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/v2.30.0/packages/eslint-plugin/docs/rules/no-non-null-assertion.md) tslint rule.

I'm not sure what the previous behavior was when either of these was `null`, but I think it makes sense to not display the countdown timer if either of them is `null`.

src/Apps/Order/Routes/NewPayment/index.tsx
iskounen: No. If 3DS authentication fails, then the user will not be able to try another credit card. Users effectively get one shot at fixing a failed payment (regardless of whether 3DS is required).

This is a bug that I can reproduce with failure test cards and is not limited to 3DS.

PR #13235, opened 2023-12-05
fix(passport): actually run specs; temporarily pend failing tests

dzucconi:
We weren't running any tests that were written in JavaScript — which, all the specs in the passport directory are.

Mostly passing but I had to pend the specs inside `passport-local-with-otp`, something to do with Chai. Where should I direct this ticket?


src/Components/Artwork/__tests__/GridItem.jest.js
MounirDhahri, line 1: thank you

PR #13236, opened 2023-12-05
fix(dia-333): prevent insertion of arbitrary error messages

dzucconi:
Closes [DIA-333](https://artsyproduct.atlassian.net/browse/DIA-333)

We were just rendering error toasts with the error message returned from the server. This is potentially a problem because you could send a login link with a custom (potentially malicious) error message. The utility of this would be limited though since you couldn't inject any code or links but 🤷 .

This switches to using error codes. We still have to use the query string to insert the email address but I don't see that as being a problem really?

Theoretically we could insert an error through Sharify but all this stuff being in JS makes small refactors annoying. But I can do that if we think this doesn't solve our problem.

[DIA-333]: https://artsyproduct.atlassian.net/browse/DIA-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Authentication/Components/AuthenticationInlineDialog.tsx
+++ b/src/Apps/Authentication/Components/AuthenticationInlineDialog.tsx
@@ -20,17 +20,23 @@ const AuthenticationInlineDialogContents: FC = () => {
     match: { location },
   } = useRouter()
 
-  // Errors might come back from 3rd party authentication
-  // as a string in an `error` query param, so display them if present.
+  // Errors might come back from 3rd party authentication so display them if present.
   useEffect(() => {
-    if (!location.query.error) return
+    if (!location.query.error_code) return
 
-    sendToast({
-      message: location.query.error,
-      variant: "error",
-      ttl: Infinity,
-    })
-  }, [location.query.error, sendToast])
+    const message = (
+      ERROR_CODES[location.query.error_code] || ERROR_CODES.UNKNOWN
+    )
+      // Resolve templates
+      .replace("{email}", location.query.email || "—")
+      .replace("{provider}", PROVIDERS[location.query.provider] || "—")
+
+    if (location.query.error) {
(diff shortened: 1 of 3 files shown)


src/Apps/Authentication/Components/AuthenticationInlineDialog.tsx
```diff
@@ -20,17 +20,23 @@ const AuthenticationInlineDialogContents: FC = () => {
...
-      message: location.query.error,
-      variant: "error",
-      ttl: Infinity,
-    })
-  }, [location.query.error, sendToast])
+    const message = (
+      ERROR_CODES[location.query.error_code] || ERROR_CODES.UNKNOWN
+    )
+      // Resolve templates
+      .replace("{email}", location.query.email || "—")
```
joeyAghion, line 31: This substitution essentially still allows arbitrary injection into the message. Would it be OK to just modify the messages to be more generic (like `this email address` instead of `the email address {email}`)?

dzucconi: The thing is, I've found that to be useful in the past — knowing what email address is actually associated, because I have a different one I use for Facebook than most other things, for instance.

I'll just port this to TS and pass the message through Sharify.

joeyAghion: Does that help the vulnerability though?

dzucconi: Yeah, just pass the whole message from the server. Can't alter it or inject anything that way.

dzucconi:
Closing in favor of https://github.com/artsy/force/pull/13251

PR #13246, opened 2023-12-07
feat: [EMI-1534] Reload shipping quotes on page refresh

erikdstock:

This PR solves [EMI-1534] by making our shipping route reload shipping quotes on load. It uses the same method to do this as the legacy route - re-saving the shipping address.

In order to do this I made a couple more changes:
- `submitFulfillmentDetails` is now defined in the `FulfillmentDetails` component where it belongs. We add some props to this component (like `processUserAddressUpdates`), and we can `submitFulfillmentDetails` directly with existing saved values or call its formik method (as before)
- More things added to helpers, including exchange error handling, tracking, and the new reducer actions.

<details><summary>Screencap</summary>

*Note: the spinner isn't working in this PR. That will immediately follow this one.*

</details> 

[EMI-1534]: https://artsyproduct.atlassian.net/browse/EMI-1534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Shipping2/ArtaErrorDialogMessage.tsx
```diff
@@ -0,0 +1,12 @@
+import { RouterLink } from "System/Router/RouterLink"
+
+export const ArtaErrorDialogMessage = () => (
+  <>
+    There was a problem getting shipping quotes. <br />
+    Please contact{" "}
```
erikdstock, line 6: Thinking I should move this into this file: https://github.com/artsy/force/blob/b7a3a68f306683a8e148b7da498be72370481733/src/Apps/Order/Utils/getErrorDialogCopy.tsx#L7

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+      setForceNewAddressFormMode(true)
+    }
+  }, [forceNewAddressFormMode, hasSavedAddresses])
+
+  const submitFulfillmentDetails = useCallback(
+    async ({
+      performUserAddressUpdates,
+      formValues,
+      formikHelpers,
+    }: {
```
erikdstock, line 93: Migrated from Shipping2/index.tsx

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+  // Once the user sees the address form, they should always see it.
+  const [forceNewAddressFormMode, setForceNewAddressFormMode] = useState(
+    !hasSavedAddresses
+  )
+
+  useEffect(() => {
+    if (!forceNewAddressFormMode && !hasSavedAddresses) {
+      setForceNewAddressFormMode(true)
+    }
+  }, [forceNewAddressFormMode, hasSavedAddresses])
```
erikdstock, line 86: Moved up from the form layout because we only perform address updates _after_ submit when we are in 'new address form' (not address modal) mode.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+      submitFulfillmentDetails({
+        performUserAddressUpdates: false,
+        formValues: {
+          attributes: existingFulfillmentDetails.fulfillmentDetails as ShipValues["attributes"],
+          fulfillmentType: existingFulfillmentDetails?.fulfillmentType,
+        },
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
erikdstock, line 201: The actual implementation on the ticket. Perform a one-time address re-save if the page loads with a special new 'automatically refresh shipping quotes' stage.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -99,12 +241,13 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
       onSubmit={handleSubmit}
       availableFulfillmentTypes={availableFulfillmentTypes}
+      forceNewAddressFormMode={forceNewAddressFormMode}
     />
   )
 }
 
 export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
-  FulfillmentDetails,
+  injectDialog(FulfillmentDetails),
```
erikdstock, line 250: 🔴 oops, I don't think this is needed any more. Dialog is moved into context helpers.

src/Apps/Order/Routes/Shipping2/Utils/shippingUtils.ts
```diff
@@ -86,14 +86,19 @@ export const getDefaultUserAddress = (
...
+    "name",
+    "phoneNumber",
+    "postalCode",
+    "region",
+  ]
+
+  const differences = fields.filter(field => !isEqual(a1[field], a2[field]))
+
+  return differences.length === 0
 }
```
erikdstock, line 104: I made this change to make debugging the differences easier than a big chain of conditionals.

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContextHelpers.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+  const orderTracking = useOrderTracking()
+  const isArtsyShipping = !!parsedOrderData.savedFulfillmentDetails
+    ?.isArtsyShipping
+
+  const handleExchangeError = useCallback(
+    (
+      error: { code: string; data: string | null | undefined },
+      logger: ReturnType<typeof createLogger>
+    ) => {
```
erikdstock, line 88: This was [handleSubmitError](https://github.com/artsy/force/pull/13246/files#diff-f6595a96ecf910d28cf63a298ec8347b3755f89334c02b4fa3b618f916ce10b4L152) in the index route. It is only used for exchange error codes (both shipping quotes and fulfillment details).

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -69,203 +60,114 @@ export interface ShippingProps {
...
+  const handleFulfillmentDetailsSaved = useCallback(
+    ({ requiresArtsyShipping }: { requiresArtsyShipping: boolean }) => {
+      if (requiresArtsyShipping) {
+        shippingContext.helpers.setStage("shipping_quotes")
+      } else {
+        advanceToPayment()
+      }
+    },
+    [advanceToPayment, shippingContext.helpers]
+  )
```
erikdstock, line 95: This is a new Prop for fulfillment details, along with `processUserAddressUpdates`. The diff in the rest of this file is big (much moved out) but the tests pass and the route works better than ever.

erikdstock (commented):
Comments and one or two things for me to revisit.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -17,18 +17,26 @@ import {
...
+import { CommerceSetShippingInput } from "__generated__/useSaveFulfillmentDetailsMutation.graphql"
+
+const logger = createLogger("Routes/Shipping2/FulfillmentDetails.tsx")
 
 export interface FulfillmentDetailsProps {
-  onSubmit: (
-    values: FulfillmentValues,
-    formikHelpers?: FormikHelpers<FulfillmentValues>
-  ) => void | Promise<any>
+  processUserAddressUpdates: (newValues: FulfillmentValues) => Promise<void>
```
damassi, line 28: is there a better name for this? "process" is a bit too ambiguous / unconventional

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
     setVerifyAddressNow(false)
   }
 
+  // Once the user sees the address form, they should always see it.
+  const [forceNewAddressFormMode, setForceNewAddressFormMode] = useState(
```
damassi, line 78: How about `isNewAddressFormMode, `setNewAddressFormMode`

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+    async ({
+      performUserAddressUpdates,
+      formValues,
+      formikHelpers,
+    }: {
+      performUserAddressUpdates: boolean
+      formValues: FulfillmentValues
+      formikHelpers?: FormikHelpers<FulfillmentValues>
+    }) => {
+      const { setSubmitting } = formikHelpers || {}
```
damassi, line 98: Format this a bit more readably, and no need to destructure :

```tsx
formikHelpers?.setSubmitting(true)
// line break
try {
  ... 
}
```

Also, should formikHelpers ever be optional? 

And we should always avoid this short hand `foo && bar()` single line execution. Always use if statements

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+      setSubmitting && setSubmitting(true)
+      try {
+        let fulfillmentMutationValues: CommerceSetShippingInput
+        let requiresArtsyShipping: boolean
```
damassi, line 102: line break

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -66,16 +74,150 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = props => {
...
+              : FulfillmentType.SHIP,
+            phoneNumber,
+            shipping: { ...addressValues, phoneNumber: "" },
+          }
+
+          if (addressVerifiedBy) {
+            fulfillmentMutationValues.addressVerifiedBy = addressVerifiedBy
+          }
+        } else {
+          requiresArtsyShipping = false
```
damassi, line 127: linebreak

src/Apps/Order/Routes/Shipping2/FulfillmentDetailsForm.tsx
```diff
@@ -54,6 +54,7 @@ export interface FulfillmentDetailsFormProps {
   onAddressVerificationComplete: () => void
   onSubmit: (values: FulfillmentValues, helpers: any) => void
   availableFulfillmentTypes: FulfillmentType[]
+  forceNewAddressFormMode: boolean
```
damassi, line 57: isNewAddressFormMode

src/Apps/Order/Routes/Shipping2/FulfillmentDetailsForm.tsx
```diff
@@ -68,6 +69,7 @@ export const FulfillmentDetailsForm = (props: FulfillmentDetailsFormProps) => {
         verifyAddressNow={props.verifyAddressNow}
         onAddressVerificationComplete={props.onAddressVerificationComplete}
         availableFulfillmentTypes={props.availableFulfillmentTypes}
+        forceNewAddressFormMode={props.forceNewAddressFormMode}
```
damassi, line 72: ditto

src/Apps/Order/Routes/Shipping2/FulfillmentDetailsForm.tsx
```diff
@@ -116,6 +119,13 @@ const FulfillmentDetailsFormLayout = (
     isValid,
   } = formikContext
 
+  const addressFormMode: AddressFormMode =
+    values.fulfillmentType === FulfillmentType.PICKUP
+      ? "pickup"
+      : props.forceNewAddressFormMode || savedAddresses.length === 0
+      ? "new_address"
+      : "saved_addresses"
```
damassi, line 127: avoid nested ternaries. Can do something like:

```tsx
const addressFormMode = (() => {
  switch (true) {
    case values.fulfillmentType === FulfillmentType.PICKUP: {
      return 'pickup'
    }
    case props.forceNewAddressFormMode || savedAddresses.length === 0: {
      return 'new_address'
    }
    default: {
      return 'saved_addresses'
    }
  }
})()

File: src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts [hunk 17]
```diff
@@ -1,12 +1,13 @@
...
 import { ShippingProps } from "Apps/Order/Routes/Shipping2"
 import { useParseOrderData } from "Apps/Order/Routes/Shipping2/Hooks/useParseOrderData"
 import {
   ShippingContext,
   ShippingContextProps,
+  ShippingStage,
+  State,
 } from "Apps/Order/Routes/Shipping2/Utils/ShippingContext"
-import { FulfillmentValues } from "Apps/Order/Routes/Shipping2/Utils/shippingUtils"
+import { useShippingContextHelpers } from "Apps/Order/Routes/Shipping2/Hooks/useShippingContextHelpers"
```
damassi, line 10: The `ShippingContext` is meant to provide all of the shared means to support itself, and is already global to the route. The logic in this file should be returned from the context itself, so that any helper could be retrieved via 

```tsx
const { doThing } = useShippingContext() 
```

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContextHelpers.tsx
```diff
@@ -0,0 +1,185 @@
...
+    // Used to disable the button
+    isValid: false,
+    // Used to get the form values for un-saving the address if the user
+    // unchecks it after saving it in the fulfillment details step.
+    values: ({
+      attributes: {
+        saveAddress: false,
+      },
+    } as unknown) as FulfillmentValues,
+  })
```
damassi, line 74: line break

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContextHelpers.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const orderTracking = useOrderTracking()
+  const isArtsyShipping = !!parsedOrderData.savedFulfillmentDetails
+    ?.isArtsyShipping
+
+  const handleExchangeError = useCallback(
+    (
+      error: { code: string; data: string | null | undefined },
+      logger: ReturnType<typeof createLogger>
+    ) => {
+      logger.error(error)
```
damassi, line 89: linebreak

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContextHelpers.tsx
```diff
@@ -0,0 +1,185 @@
...
+      const parsedData = error.data ? JSON.parse(error.data) : {}
```
damassi, line 90: linebreak

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext.ts
```diff
@@ -1,19 +1,21 @@
...
+export type ShippingStage =
+  | "fulfillment_details"
+  | "shipping_quotes"
+  | "refresh_shipping_quotes"
+
+export interface State {
+  newSavedAddressId: string | null
+  selectedShippingQuoteId: string | null
+  stage: ShippingStage
+}
```
damassi, line 14: linebreak

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext.ts
```diff
@@ -1,19 +1,21 @@
...
 export interface ShippingContextProps {
+  state: State
   parsedOrderData: ParsedOrderData
-  helpers: {
-    fulfillmentDetails: FulfillmentHelpers & {
-      setFulfillmentFormHelpers: (helpers: FulfillmentHelpers) => void
-    }
-  }
+  helpers: ShippingContextHelpers
```
damassi, line 18: no reason to separate context helpers from context, just keep it all internal to here

damassi (dismissed):
Just a few minor tidiness concerns, but as I don't know much about the behavior can't comment too much on it!

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: So the function could be called through a normal formik submit or directly through this refresh-shipping-quotes hack. I suppose I could submit the form directly, but then the refresh-shipping-quotes logic would have to go inside the more presentational FulfillmentDetailsForm component, and I thought it best to bypass address verification and form validation - if the values are already saved on the order, best to avoid letting the client try to mess with that.

Looking at the formik docs, I don't think I actually need access to the helpers (`setSubmitting`) at all, so I'll remove that and just have the function accept form.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: Not a hill I'm going to die on, but I prefer making explicit that we are **forcing** address form mode [here](https://github.com/artsy/force/pull/13246#discussion_r1421146403), since the default behavior would be to just show the saved addresses version once an address is saved. A more generic `is_` naming seems more confusing to me - why is it being set in a different file.

Actually I think this should be restructured to move the whole linked area's logic up into this component. The reason for putting it here was [this line](https://github.com/artsy/force/pull/13246/files#diff-71b8774db34fee1f35b7c4594e7226120f02c41e93065e5d7fb9551fd7543991R210) - new address form mode performs user address updates after saving details, while saved address modals do the updates first and update fulfillment details after. This inconsistency could obviously be improved too. But for now, I'm going to move this logic next to each other and it should read better.

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
damassi: If there are some utils (we don't use the term "helpers" in force, that's a rails idiom), its fine to group them into their own file (under "/utils") and import them into the context in order to funnel them to the rest of the app via the context, but I don't think that's necessary here. We can just keep everything co-located and refactor it all later.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
erikdstock: It was `handle` - i was probably overthinking it but since `handle` callbacks usually match `on` props, and this isn't an ON, this a callback prop that the component decides when to call to call, I tried to make it more generic. I'll just change it back.

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext.ts
erikdstock: not following, are you saying don't nest these functions?

src/Apps/Order/Routes/Shipping2/Hooks/useShippingContext.ts
erikdstock: Similar to my below comment, i'm open to changing this later but am going to merge this once tests pass.

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
damassi: makes sense

src/Apps/Order/Routes/Shipping2/FulfillmentDetails.tsx
```diff
@@ -211,7 +331,7 @@ export const FulfillmentDetailsFragmentContainer = createFragmentContainer(
 
 const getInitialValues = (
   me: FulfillmentDetailsForm_me$data,
-  orderData: ParsedOrderData
+  orderData: ShippingContextProps["parsedOrderData"]
```
damassi, line 334: 👍

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext/ShippingContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+    newSavedAddressId: null,
+    selectedShippingQuoteId: parsedOrderData.selectedShippingQuoteId ?? null,
+    stage: isArtsyShipping ? "refresh_shipping_quotes" : "fulfillment_details",
+  }
+
+  /*
+    *******
+    Helpers
+    *******
+  */
```
damassi, line 80: /*
 * Helpers
 */

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext/ShippingContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+  const [state, dispatch] = useReducer(shippingStateReducer, initialState)
```
damassi, line 82: line break

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext/ShippingContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Because there is a single button for both fulfillment details and
+   * shipping quote steps (and duplicated in the sidebar)
+   * we need to hack some formik values UP from the fulfillment details form.
+   *
+   * Currently we need to pass up:
+   */
+  const [fulfillmentFormHelpers, setFulfillmentFormHelpers] = useState<
+    Pick<FormikProps<FulfillmentValues>, "submitForm" | "isValid" | "values">
+  >({
```
damassi, line 92: can we move this into the reducer?

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext/ShippingContext.tsx
```diff
@@ -0,0 +1,236 @@
...
+      attributes: {
+        saveAddress: false,
+      },
+    } as unknown) as FulfillmentValues,
+  })
+
+  const fulfillmentDetailsHelpers = {
+    ...fulfillmentFormHelpers,
+    setFulfillmentFormHelpers,
+  }
```
damassi, line 109: reducer

damassi (approved):
PR is looking really good @erikdstock 👍 have dropped some follow-up stuff into slack around actions, but we can iterate. Also appreciate removing all of the callback / memo stuff for the time being; we can readd as necessary, but for now things are simpler and more deterministic.

src/Apps/Order/Routes/Shipping2/Utils/ShippingContext/ShippingContext.tsx
damassi: Or rather, just let the code speak for itself, it should be fine 👍

PR #13256, opened 2023-12-08
feat(EMI-1592): show the list price for a partner offer

narikazu:
The type of this PR is: Feat

This PR solves EMI-1592

### Description

When the order is a partner offer without a list price, the price label should be `List price` and the value for the price should be `Not publicly listed` now. 

##### Price range

Screenshot 2023-12-08 at 10 52 27

##### Price hidden

Screenshot 2023-12-08 at 15 24 44

### DIFF
--- a/src/Apps/Order/Components/ArtworkSummaryItem.tsx
+++ b/src/Apps/Order/Components/ArtworkSummaryItem.tsx
@@ -82,12 +82,17 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
             </Text>
           </>
         )}
-        {!isPrivateSale && artworkPrice && (
+        {!isPrivateSale && artworkPrice?.price && (
           <Text variant="sm">
             {`${priceLabel} ${appendCurrencySymbol(
               artworkPrice.price,
               currencyCode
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -81,14 +82,19 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
...
-        {!isPrivateSale && artworkPrice && (
+        {!isPrivateSale && artworkPrice?.price && (
           <Text variant="sm">
             {`${priceLabel} ${appendCurrencySymbol(
               artworkPrice.price,
               currencyCode
             )}`}
           </Text>
         )}
+        {isPartnerOffer && !artworkPrice?.price && (
```
leamotta, line 93: How about not checking for partner offer here? That way we can have a nicer copy also in the case of traditional make offer on a non public price artwork, example:
Screenshot 2023-12-08 at 15 37 19

```suggestion
        {!artworkPrice?.price && (
```

leamotta (approved):

PR #13274, opened 2023-12-13
fix(onyx-559): search pills - fixes wrong pill state that makes counter superscript invisible

nickskalkin:

This PR solves [ONYX-559]

### Description

Solves the issue when clicking on a pill makes a counter invisible:

| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/9e23dfbe-96ff-497c-8268-a863ee1cfd60" /> | <video src="https://github.com/artsy/force/assets/3934579/9cb5cf4e-c708-4ab1-bc6f-cb40bd62d29f" /> |

It seems like there is a "bug" in palette: when clicking on a pill that is an `active` state - such pill changes background to grey and text colour to blue (counter colour remains white though) . Maybe this shouldn't even be fixed, I think that `active` state shouldn't be used here at all. In my fix I've decided to use another state - `selected`, which doesn't have the counter bug.

I've also tried to address this [FIXME comment](https://github.com/artsy/force/blob/main/src/Apps/Search/Components/NavigationTabs.tsx#L98) made by @dzucconi.

[ONYX-559]: https://artsyproduct.atlassian.net/browse/ONYX-559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Search/Components/NavigationTabs.tsx
+++ b/src/Apps/Search/Components/NavigationTabs.tsx
@@ -1,9 +1,9 @@
 import * as React from "react"
-import { HorizontalOverflow, Pill } from "@artsy/palette"
+import { BoxProps, HorizontalOverflow, Pill } from "@artsy/palette"
 import { NavigationTabs_searchableConnection$data } from "__generated__/NavigationTabs_searchableConnection.graphql"
 import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
 import { createFragmentContainer, graphql } from "react-relay"
-import { RouterLink } from "System/Router/RouterLink"
+import { RouterLink, RouterLinkProps } from "System/Router/RouterLink"
 import { useIsRouteActive } from "System/Router/useRouter"
 import {
   ActionType,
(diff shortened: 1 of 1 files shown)


dzucconi:
We can fix Palette. This looks like it breaks normal clicking with cmd/shift.

nickskalkin:
moving back to draft - found a bug

dzucconi (changes requested):
I'd just close this. I can fix the underlying issue today.

src/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -37,42 +37,23 @@ const TAB_NAME_MAP = {
   gene: "Categories",
 }
 
-interface RouteTabProps extends BoxProps, RouterLinkProps {
```
dzucconi, line 40: This would break a number of things that rely on the Box interface.

src/Apps/Search/Components/NavigationTabs.tsx
```diff
@@ -88,79 +69,69 @@ export const NavigationTabs: React.FC<NavigationTabsProps> = ({
...
+      variant="search"
+      count={count}
+      as={RouterLink}
+      my={1}
+      mr={1}
+      selected={isSelected}
+      key={tabName}
+      onClick={event => {
+        event.preventDefault()
+        router.router.push(to)
```
dzucconi, line 84: Shouldn't ever do this as it would break cmd+click/shift+click.

nickskalkin:
@dzucconi there is a weird behaviour that I see on staging when I do cmd/shift click, but it doesn't exist when using `selected` state 🤷 . I think we should use `selected` in any case here, at least to align pills colors across other surfaces. We use `selected` color in the app and search dropdown.

dzucconi:
The selected state comes from the route simply being active.

nickskalkin:
I mean currently we use `active` state: https://github.com/artsy/force/pull/13274/files#diff-dbf1dea892b03179ba6855a49712655eab5ca3d4694c1acfe9f491b8fa0ecdc8L56, instead of `selected`: https://github.com/artsy/force/pull/13274/files#diff-dbf1dea892b03179ba6855a49712655eab5ca3d4694c1acfe9f491b8fa0ecdc8R80

dzucconi:
Yeah, but the other changes here you don't want to be doing. Just that one.

nickskalkin:
@dzucconi sure, I'll revert them and keep only this 'selected' thing. Just tried to quickly fix your FIXME comment without fully understanding what what broken

dzucconi:
So if you just roll back all those other changes and just pass selected if the route is active. We should be fine. The problem you're seeing is because of the way the states are stacking.

nickskalkin:
yeah, I simply got rid of this function that renders pills, extracted it to a separate component

dzucconi (approved):
OK, sorry the diff was hard to read. So there's already a bug in production that we should address (the route pushes). Would make sense to just do that in it's own PR. The `selected` state looks good though, thank you.

PR #13286, opened 2023-12-14
feat(alerts): allow artist series suggestion to be accepted

anandaroop:

Solves this QA item: https://www.notion.so/artsy/I-ve-found-a-case-where-one-suggested-filter-is-clicked-but-isn-t-added-a4f52b3dbe7c4f0ba13e65cb85856e31

### Description

Now that Artist Series suggestions have landed in Metaphysics we need to make sure Force can accept them.


nickskalkin (approved):
👍

PR #13289, opened 2023-12-14
feat: Always enable email notifications when creating alerts

olerichter00:
## Description

Always enable email notifications when creating alerts. Previously, it was disabled if the user had unsubscribed from Custom Alerts emails.

Screenshot 2023-12-14 at 17 03 53

Addresses [ONYX-607](https://artsyproduct.atlassian.net/browse/ONYX-607)

[ONYX-607]: https://artsyproduct.atlassian.net/browse/ONYX-607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/Components/NotificationPreferences.tsx
+++ b/src/Components/Alert/Components/NotificationPreferences.tsx
@@ -9,7 +9,7 @@ import {
 } from "@artsy/palette"
 import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
 import { RouterLink } from "System/Router/RouterLink"
-import { FC, useEffect } from "react"
+import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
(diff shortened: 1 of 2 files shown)


src/Components/Alert/Components/NotificationPreferences.tsx
```diff
@@ -39,14 +39,6 @@ export const NotificationPreferences: FC<NotificationPreferencesProps> = ({
     }
   )
 
-  useEffect(() => {
```
olerichter00, line 42: Removing this effect is enough because the default is 'true' and here it was set to false in case the user had unsubscribed from these emails.

dblandin (dismissed):
Awesome 💯

PR #13297, opened 2023-12-15
refactor(shipping2): Polish / organization pass 

damassi:
The type of this PR is: **Refactor**

### Description

It's cool seeing this come together. With the context in place all of the final pieces dropped right out of the hat 🎉

> ⚠️  Worth browsing through [the folder](https://github.com/artsy/force/tree/eb10e1f52fc7d4e0e2efe9a836ac3250052d89d2/src/Apps/Order/Routes/Shipping2) to get a look at the overall organizational changes, and in particular `index.tsx` and `ShippingContext.tsx` 

Since the Shipping2 refactor is getting close to completion, wanted to go through and do a quick polish pass. This was a two-PR process, but merged the second into this as a single review PR is better than two. [See here.](https://github.com/artsy/force/pull/13298)

This PR concerns itself with eliminating all props passing where possible (stuff from context doesn't need to be passed), and organizing shared code into hooks. Additionally it adds use of the `useFragment` hook, so that we can start eliminating FragmentContainers (can't use usePagination, etc hooks yet tho, but useFragment is fine). 

Additionally, this PR updates some of the context naming so that it matches typical react patterns, ie actions and state, and organizes stuff a bit more along the lines of strict co-location and hooks.

No logic has changed, only some light reorganization into shared hooks and some clarity updates.

PR #13301, opened 2023-12-18
fix SavedAddresses2 specs

erikdstock:
A few test fixes for #13297 - mostly updating spec setup for the new organization.


damassi (approved):
tysm! I was 😅

PR #13302, opened 2023-12-18
feat(alerts): source artist series suggestions from artwork

anandaroop:

This PR solves [ONYX-563]

### Description

Now that we suggest artist series criteria in the Create Alert UI, we want to make sure they are as relevant as possible.

This means that when we are creating an alert from an artwork, we want the artist series suggestion to be based on the artist series  which _that_ artwork belongs to (if any).

(This is made possible by https://github.com/artsy/metaphysics/pull/5471)

|Before|After|
|---|---|
|artist|artwork|

[ONYX-563]: https://artsyproduct.atlassian.net/browse/ONYX-563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -331,7 +331,10 @@ const WrappedArtworkApp: React.FC<Props> = props => {
         }}
         enabled={websocketEnabled}
       >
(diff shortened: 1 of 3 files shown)


nickskalkin (approved):
LGTM!

PR #13308, opened 2023-12-20
fix(images): increase quality of install shots

anandaroop:

### Description

In response to [this complaint](https://artsy.slack.com/archives/C03N12SR0RK/p1702895145151689) we learned that shows’ install shots are another case where our `quality: 50` setting bites us.

(Which makes sense since [white- or light-colored walls](https://observablehq.com/@anandaroop/artwork-color-casts) predominate in these photos.)

So this adds another targeted fix — very similar to https://github.com/artsy/force/pull/12913 — this time focusing on install shots.

The effect is subtle, but hopefully this gif (artifacts notwithstanding) makes the change clearer:

q50-q85-2

### DIFF
--- a/src/Apps/Show/Components/ShowInstallShots.tsx
+++ b/src/Apps/Show/Components/ShowInstallShots.tsx
@@ -147,6 +147,7 @@ export const ShowInstallShotsFragmentContainer = createFragmentContainer(
           internalID
           caption
           mobile: resized(
+            quality: 85
             width: 200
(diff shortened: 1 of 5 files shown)


dzucconi (approved):

dzucconi:
FWIW I've read in a few places recently that quality with webp doesn't really improve much over 80

anandaroop:
Gtk, I still think we should do the system-wide fix described in #12913, maybe we standardize on `80` then.

PR #13324, opened 2023-12-22
chore: fix remaining shipping2 tests, complete initial refactor

erikdstock:

It completes [EMI-1526] and relates to [EMI-1520]. After this PR all existing legacy tests will have been implemented, opening the door to more refactoring. All skipped tests have been unskipped (there will be failures while in **draft**), but TODOs remain in the test file.

### Summary of changes
- User address updates are now handled consistently for different flows, always happening _before_ the relevant mutation on the exchange order.
- The hook enabling user address updates has been updated to handle multiple flows.
- All integration tests in Shipping2.jest.tsx are working.
- Added tests at the component level, migrating some to use react testing library.
- Recently-pended tests have been left in the pending state (owing to timeout issues in CI), but they are passing in development. The ideal here will be to lean more on a mix of unit tests and integrity for testing in the long term.

The diff is quite large because of fixing these final specs revealed some bigger changes that were necessary.

<details><summary>old notes</summary>

### Functionality
Because the fulfillment details steps has two separate form layouts (saved addresses with a modal, or no saved addresses with a form on the page itself) there are multiple different flows to account for:
- With saved addresses we defer form validation to the saved addresses formik context (a single formik context could make this easier)
- With saved addresses, we perform user address updates first, then save fulfillment details. with new address it is the opposite
  - _if this were not necessary, it could be easier - ideally call the most restrictive mutation first_
- With saved addresses we do not perform address verification.
  - Someday we may want to add address verification when editing or creating a new address - on the old shipping form this was too complicated to implement.
- Saved addresses are persisted to the order when they are selected, without needing to click save and continue. Thus the user must click save and continue again to proceed [with no mutation].
- On load, SavedAddresses automatically selects an initial address value based on a) what is already saved on the order or b) the user's default address, triggering the save to the order. `a)` is new functionality designed to avoid a bug mentioned in [EMI-1520] and below and it does this by comparing address fields.
  - _Does the logic to match existing order fulfillment data with a user's saved addresses need to exist up in ShippingContext, or could it all happen in the SavedAddresses component?_

### Edge cases and possible bugs
So far in this draft PR I've covered most of these cases, but there may be more, especially with the edge cases noted in [EMI-1520].
- When the route loads, we want to re-save the fulfillment details on the order if it is an artsy shipping order (noted above). This causes shipping quotes to be re-fetched, but **does not un-save a previously-saved shipping quote**, which will no longer be visible and may have a different price or destination. **(Wont fix)**
- Our current logic doesn't respect whether an artwork _can_ ship to that automatically-selected address, relying on an error from the server to immediately tell the user to choose a different address. Likewise we do not disable address radio buttons that aren't valid for shipping. **(Partially fixed - we only automatically select from countries that can be shipped-to)**
- Once in SavedAddresses mode (ie the user has a saved address) there is no option to not save a new address **(Wont fix)**
- In this flow:
  - User applies shipping address to order, but doesn't save address for later use on their user
  - User adds a new, different saved address to their user elsewhere
  - User reloads shipping page, now in saved addresses mode
  - `=>` The originally-saved address may be clobbered by the auto-saving feature, since their original address won't be in the saved address list.

</details>

[EMI-1526]: https://artsyproduct.atlassian.net/browse/EMI-1526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1520]: https://artsyproduct.atlassian.net/browse/EMI-1520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1520]: https://artsyproduct.atlassian.net/browse/EMI-1520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -175,15 +182,19 @@ const FulfillmentDetailsFormLayout = (
   const tabbableIf = (activeForm: AddressFormMode): 0 | -1 =>
     addressFormMode === activeForm ? 0 : -1
 
-  const handleSelectSavedAddress = (address: ShippingAddressFormValues) => {
-    setValues({
+  const handleSelectSavedAddress = async (
+    address: ShippingAddressFormValues
+  ) => {
+    await setValues({
```
erikdstock, line 188: I had to upgrade the formik version to give this the [correct Promise return type](https://github.com/jaredpalmer/formik/pull/3871) (so that `submitForm()` will hopefully work as expected)

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -252,8 +263,8 @@ const FulfillmentDetailsFormLayout = (
             <SavedAddressesFragmentContainer
               active={addressFormMode === "saved_addresses"}
               me={props.me}
-              onSelect={a => {
-                handleSelectSavedAddress(a)
+              onSelect={address => {
+                handleSelectSavedAddress(address)
```
damassi, line 267: 🙏

src/Apps/Order/Routes/Shipping2/Components/SavedAddresses2.tsx
```diff
@@ -76,62 +79,55 @@ const SavedAddresses: React.FC<SavedAddressesProps> = props => {
...
+    setSelectedAddressID(address.internalID)
     setActiveModal({
       type: AddressModalActionType.EDIT_USER_ADDRESS,
       address: address,
     })
   }
 
   const refetchAndSelectAddress = async (addressID: string) => {
-    await refetchAddresses()
+    console.log("selecting address", addressID)
```
damassi, line 122: 🐟

src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -27,7 +27,7 @@ import { CollapseDetails } from "Apps/Order/Routes/Shipping2/Components/Collapse
 export type ShippingStage =
   | "fulfillment_details"
   | "shipping_quotes"
-  | "refresh_shipping_quotes"
+  | "advance_on_click"
```
damassi, line 30: #superminor - `a` comes before `s` and `f`

yarn.lock
```diff
@@ -11596,17 +11604,18 @@ formidable@^1.2.0:
   integrity sha512-Fs9VRguL0gqGHkXS5GQiMCr1VhZBxz0JnJs4JmMp/2jL18Fmbzvv7vOFRU+U8TBkHEE/CX1qDXzJplVULgsLeg==
 
 formik@^2.2.9:
-  version "2.2.9"
-  resolved "https://registry.yarnpkg.com/formik/-/formik-2.2.9.tgz#8594ba9c5e2e5cf1f42c5704128e119fc46232d0"
-  integrity sha512-LQLcISMmf1r5at4/gyJigGn0gOwFbeEAlji+N9InZF6LIMXnFNkO42sCI8Jt84YZggpD4cPWObAZaxpEFtSzNA==
+  version "2.4.5"
```
damassi, line 11607: minor version, so aok 👍

damassi (approved):
Once tests  are green, LGTM 👍 

(And congrats on wrapping this all up 🎉)

src/Apps/Order/Routes/Shipping2/Hooks/useSaveSelectedShippingQuote.tsx
```diff
@@ -36,12 +36,21 @@ export const useSaveSelectedShippingQuote = (
     try {
       shippingContext.actions.setIsPerformingOperation(true)
 
+      if (
+        shippingContext.state.fulfillmentDetailsCtx?.values.meta.mode ===
```
erikdstock, line 40: todo: can this be stored/pulled directly from shippingContext.state and not used in the form ctx?

src/Apps/Order/Routes/Shipping2/Components/SavedAddresses2.tsx
```diff
@@ -1,148 +1,177 @@
...
-    shippingContext.orderData.savedFulfillmentDetails?.selectedSavedAddressId
-  const [selectedAddressID, setSelectedAddressID] = useState<
-    string | undefined
-  >(
-    getBestAvailableAddress(
-      addressList,
-      selectedSavedAddressId,
-      shippingContext.orderData.availableShippingCountries
-    )?.internalID
+  const addressList = compact<SavedAddressType>(
```
erikdstock, line 81: Todo: Added addressList to shippingContext because it was used in hooks. Should it be used here too? I think the issue was that SavedAddresses has its own special fragment to pull this data from.

src/Apps/Order/Routes/Shipping2/Hooks/useSaveSelectedShippingQuote.tsx
```diff
@@ -36,12 +36,21 @@ export const useSaveSelectedShippingQuote = (
...
+        "new_address"
+      ) {
+        await handleNewUserAddressUpdates(
+          shippingContext.state.fulfillmentDetailsCtx.values
```
erikdstock, line 44: An effect of letting a user edit values in 2 steps at once - if they _change_ the address, the shipping quotes will disappear and they will be back on the fulfillment details step. But if they check/uncheck the box we just need to save/delete the address - hence this line, which for consistency now happens before saving the shipping quote.

Another way to do this could be to have the save/unsave mutation triggered from inside fulfillment details immediately upon clicking the box. Might be a good follow-up.

src/Apps/Order/Routes/Shipping2/Hooks/useHandleUserAddressUpdates.ts
```diff
@@ -1,108 +0,0 @@
...
-  addressWithFallbackValues,
-  matchAddressFields,
-} from "Apps/Order/Routes/Shipping2/Utils/shippingUtils"
-import createLogger from "Utils/logger"
-
-const logger = createLogger(
-  "Order/Routes/Shipping/Hooks/useHandleUserAddressUpdates.tsx"
-)
-
-export const useHandleUserAddressUpdates = () => {
```
erikdstock, line 17: moved to `useUserAddressUpdates`

src/Apps/Order/Routes/Shipping2/Hooks/useBackToFulfillmentDetails.tsx
```diff
@@ -1,64 +0,0 @@
...
-import { matchAddressFields } from "Apps/Order/Routes/Shipping2/Utils/shippingUtils"
-import { extractNodes } from "Utils/extractNodes"
-import { Shipping2_me$data } from "__generated__/Shipping2_me.graphql"
-import { useEffect } from "react"
-
-/**
- * Go back to fulfillment details stage if the user edits the address or
- * deletes a saved address.
- */
-export const useBackToFullfillmentDetails = (me: Shipping2_me$data) => {
```
erikdstock, line 12: Moved this logic into a wrapper around specific `handleChange` callbacks that should trigger the change. It's cleaner, less code and easier to trace.

src/Apps/Order/Routes/Shipping2/Components/SavedAddresses2.tsx
```diff
@@ -1,148 +1,177 @@
...
-    if (!selectedAddressPresent) {
-      setSelectedAddressID(
-        getBestAvailableAddress(
+  const addressSavedToOrderID = savedAddressOnOrder?.internalID
+
+  // TODO: Make sure this can't create an infinite loop if submitting fails
+  // Is it necessary beyond the initial render?
+  useEffect(() => {
+    // Automatically select (save) best available address ID if it isn't present
+    const automaticallySelectBestAddress = async () => {
```
erikdstock, line 105: To avoid endless submitting if something goes wrong (eg the saved address is not shippable for this artwork) i only put `[props.active]` in the dependency array.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -492,17 +555,16 @@ const FulfillmentDetailsFormLayout = (
               data-testid="FulfillmentDetailsForm_saveAddress"
               tabIndex={tabbableIf("new_address")}
               onSelect={selected => {
-                setFieldValue("attributes.saveAddress", selected)
+                setFieldValue("meta.saveAddress", selected)
```
erikdstock, line 558: Moved non-address values into a special `meta` object. This made a lot of other mutation-arg-passing cleaner.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -450,7 +511,9 @@ const FulfillmentDetailsFormLayout = (
                   autoCapitalize="characters"
                   autoCorrect="off"
                   value={values.attributes.postalCode}
-                  onChange={trackAutoCompleteEdits("postalCode", handleChange)}
+                  onChange={withBackToFulfillmentDetails(
+                    trackAutoCompleteEdits("postalCode", handleChange)
+                  )}
```
erikdstock, line 516: `useBackToFulfillmentDetails` hook behavior now wraps specific fields we want to track.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -132,60 +150,101 @@ const FulfillmentDetailsFormLayout = (
...
+      await formikContext.setValues({
+        ...formikContext.values,
+        fulfillmentType: FulfillmentType.SHIP,
+        attributes: addressWithFallbackValues(address),
+        meta: {
+          ...formikContext.values.meta,
+        },
+      })
+
+      await formikContext.submitForm()
```
erikdstock, line 171: After saving a user's address, the form receives it in this callback, sets all the values (because it is in saved address mode the inputs are not visible) and submits the form. FulfillmentDetails controlls submitting the values to exchange only _after_ SavedAddresses2 performs the user address updates.

src/Apps/Order/Routes/Shipping2/Components/AddressModal2.tsx
```diff
@@ -330,37 +429,45 @@ export const AddressModal: React.FC<AddressModalProps> = ({
...
+const validationSchema = Yup.object().shape({
+  attributes: Yup.object().shape(ADDRESS_VALIDATION_SHAPE),
+})
+
 export const GENERIC_FAIL_MESSAGE =
   "Sorry, there has been an issue saving your address. Please try again."
 
-const validationSchema = Yup.object().shape(ADDRESS_VALIDATION_SHAPE)
+const handleGravityErrors = (
+  errors: ReadonlyArray<{ message: string }>,
```
erikdstock, line 452: Gravity mutation helpers in `useUserAddressUpdates` now always return a type something like `{data: T} | {errors: [{message?: string}]} | null` (null if no action was performed). This lets us handle the errors a bit more consistently.

src/Apps/Order/Routes/Shipping2/Components/AddressModal2.tsx
```diff
@@ -11,287 +10,396 @@ import {
...
+            </Banner>
+          )}
 
+          <GridColumns mt={[1, 2]}>
+            <Column span={12}>
+              <Input
+                title="Full name"
+                placeholder="Full name"
+                id="name"
+                name="attributes.name"
```
erikdstock, line 243: I copied these fields in so i could give them the `attributes` key nesting. On reflection, I'm not sure if this was necessary or was part of an earlier attempt to get mutation-calling consistent.

src/Apps/Order/Routes/Shipping2/Components/AddressModal2.tsx
```diff
@@ -11,287 +10,396 @@ import {
...
-                </Banner>
-              )}
-
-              <AddressModalFields />
-
-              <Spacer y={2} />
+        <Form data-testid="AddressModal">
+          {formikContext.status && (
+            <Banner my={2} data-testid="form-banner-error" variant="error">
+              {formikContext.status}
```
erikdstock, line 233: using Formik's `status` for this now.

erikdstock (commented):
Comments on approach and major changes.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -103,23 +108,35 @@ const FulfillmentDetailsFormLayout = (
...
+   * Expose formik context to entire shipping route
+   * via `shippingContext.state.fulfillmentDetailsCtx`
+   */
   useEffect(() => {
-    /**
-     * Pass some key formik bits up to the shipping route
-     * TODO: This could be accomplished with useImperativeHandle(ref, formikContext)
-     */
-    shippingContext.actions.setFormHelpers(formikContext)
+    shippingContext.actions.setFulfillmentDetailsCtx(formikContext)
```
damassi, line 123: We should be more descriptive in the name of this function, just so we can see that its a formik context versus some other type of context:

```tsx
shippingContext.actions.setFulfillmentDetailsFormikContext(formikContext)
```

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
damassi: makes sense 👍

src/Apps/Order/Routes/Shipping2/Components/__tests__/SavedAddresses2.jest.tsx
```diff
@@ -0,0 +1,335 @@
...
+        addressLine2: "Basement",
+        city: "Madrid",
+        region: "NY",
+        postalCode: "28001",
+        country: "ES",
+        phoneNumber: "555-555-5555",
+      }
+      const addAddressButton = screen.getByText("Add a new address")
+      await userEvent.click(addAddressButton)
+      logTime("clicked button")
```
damassi, line 246: Might want to remove these loggers before merge

src/Apps/Order/Routes/Shipping2/Hooks/useHandleSaveFulfillmentDetails.ts
```diff
@@ -0,0 +1,129 @@
...
+
+type ExchangeOrGravityError =
+  | Error
+  | {
+      code: string
+      data: string | null | undefined
+      type: string
+    }
+
+export const useHandleSaveFulfillmentDetails = () => {
```
damassi, line 23: Maybe just `useSaveFulfillmentDetails`, since the `handleSaveFulfillmentDetails` return value is just below? No need to double up

src/System/useFeatureFlag.tsx
```diff
@@ -30,7 +30,7 @@ export function useFeatureFlag(featureName: string): boolean | null {
 
   if (flagEnabled === undefined) {
     warnInDevelopment(
-      "[Force] Warning: cannot find flagName in featureFlags: ",
+      `[Force] Warning: cannot find ${featureName} in featureFlags: `,
```
damassi, line 33: 👌

damassi (dismissed):
Beautiful @erikdstock 💯

src/Apps/Order/Routes/Shipping2/Hooks/useHandleSaveFulfillmentDetails.ts
erikdstock: yeah, why did I do this ... 🤔 ...
- `useSaveFulfillmentDetails` is the `useMutation` hook for just that operation
- This hook includes prepping the values for the different mutation inputs (pickup, ship, artsy shipping) as well as the error handling
I'll try to think about better naming.

src/Apps/Order/Routes/Shipping2/Hooks/useSaveSelectedShippingQuote.tsx
erikdstock: I'm going to leave this for now because it also includes the `pickup` mode, which isn't controlled from outside the component.

erikdstock: update: I went ahead and did it. cleaner and also caught an invisible bug.

PR #13326, opened 2023-12-29
refactor: do not throttle experiment_viewed tracking events

mc-jones:
The type of this PR is: refactor

This PR solves [PHIRE-533]

### Description
This PR refactors the `useFeatureFlag` hook to no longer throttle segment events when `trackFeatureView` is called.  

### Problem
Currently, we throttle the number of segment events to one event per user, per variant view (i.e., if a user sees an experiment multiple times, only the first view sends a segment event), using `localStorage` to track if an experiment was viewed or not. This implementation makes it difficult to analyze and accurately report on the outcome of our experiments and creates extra work for the data team. This was discussed in more detail in this [Slack thread](https://artsy.slack.com/archives/C0KEQD4B0/p1702548231405999) and the recommendation was made to change the implementation.

### Solution
With these changes, we will send a segment event each time a user views an experiment. In other words, if a user sees an experiment 10 times, we send 10 `experiment_viewed` events, one for each view. 

### TODO
- [x] Do some quick smoke testing locally, using the segment dashboard to verify. 

### Migration
Merge this PR when we no longer have an [experiment running ](https://unleash.artsy.net/features?sort=type)in `unleash`.

[PHIRE-533]: https://artsyproduct.atlassian.net/browse/PHIRE-533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/System/useFeatureFlag.tsx
+++ b/src/System/useFeatureFlag.tsx
@@ -4,6 +4,7 @@ import { Variant } from "unleash-client"
 import { getENV } from "Utils/getENV"
 import { useRouter } from "./Router/useRouter"
 import { pathToOwnerType } from "System/Analytics/AnalyticsContext"
+import { useRef } from "react"
 
 export type FeatureFlags = Record<string, FeatureFlagDetails>
 
@@ -78,6 +79,7 @@ export function useTrackFeatureVariant({
   payload,
 }: VariantTrackingProperties) {
   const router = useRouter()
+  const experimentViewTracked = useRef<boolean>(false)
 
   const trackFeatureVariant = () => {
     // HACK: Temporary hack while we refactor useAnalyticsContext
@@ -90,9 +92,7 @@ export function useTrackFeatureVariant({
     const pageSlug = pageParts[2]
     const pageType = pathToOwnerType(path)
 
-    const trackFeatureView = shouldTrack(experimentName, variantName)
-
-    if (trackFeatureView && variantName !== "disabled") {
+    if (experimentViewTracked.current !== true && variantName !== "disabled") {
       // HACK: We are using window.analytics.track over trackEvent from useTracking because
       // the trackEvent wasn't behaving as expected, it was never firing the event and
       // moving to using the solution below fixed the issue.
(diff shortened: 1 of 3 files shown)


src/System/useFeatureFlag.tsx
```diff
@@ -78,6 +79,7 @@ export function useTrackFeatureVariant({
   payload,
 }: VariantTrackingProperties) {
   const router = useRouter()
+  const experimentViewTracked = useRef<boolean>(false)
```
mc-jones, line 82: This guards against sending an event on every react re-render. The experiment will fire once per component mount.

src/System/useFeatureFlag.tsx
```diff
@@ -90,9 +92,7 @@ export function useTrackFeatureVariant({
     const pageSlug = pageParts[2]
     const pageType = pathToOwnerType(path)
 
-    const trackFeatureView = shouldTrack(experimentName, variantName)
-
-    if (trackFeatureView && variantName !== "disabled") {
+    if (experimentViewTracked.current !== true && variantName !== "disabled") {
```
dzucconi, line 95:
```suggestion
    if (!experimentViewTracked.current && variantName !== "disabled") {
```

damassi (approved):

cavvia (commented):
Thanks @mc-jones! I can't speak to the details of the implementation, but I just wanted to mention that we should audit any active experiments when pushing this out, since we have a monthly quota for Segment events. There seem to be a couple listed on the internal tools UI, one of which is app specific so it's unaffected.

mc-jones:
Looks good on the segment dashboard when tested locally. Marking off the todo.

mc-jones:
> but I just wanted to mention that we should audit any active experiments when pushing this out, since we have a monthly quota for Segment events

Noted. I was thinking we would want to push this out when no more experiments were running to avoid changing the tracking behavior mid-experiment. But let me know if I'm misunderstanding your point.

mc-jones:
I am realizing we should check the `eigen` implementation as I believe it sends `experiment_viewed` events as well, as I imagine we want the implementations to be consistent.

cavvia:
@mc-jones I'm not aware of any experiments which are currently running (I don't think either of the ones showing in internal tools are actually active). But I'll defer to @leodmz for the confirmation.

leodmz:
> I am realizing we should check the `eigen` implementation as I believe it sends `experiment_viewed` events as well, as I imagine we want the implementations to be consistent.

Onyx is currently planning on running an experiment in eigen. 
To my knowledge they're still defining what the test will look like but this concern of aligning eigen and force is relevant

cavvia:
@leodmz Any idea how many eigen screenviews we get per day? If they are firing on every homefeed load (e.g. a nfy recs experiment, which we have historically run pretty frequently) it would be good to get an estimate of how many extra segment events that will produce.

In general we only need one `experiment_viewed` event per session in which an experiment was viewed, but since sessions are defined downstream it's difficult for clients to adhere to that. 

We currently define an [idle timeout of 30 minutes](https://github.com/artsy/fulcrum/blob/main/dbt/models/sessions_web_folder/session_starts.sql#L52) in our session definition so one precaution would be for clients to send this event no more often than once every 30 minutes. That does add complexity to the client implementation though, so I would want to implement that only if we think we are going to come close to the 30M events per month limit when running two parallel tests (say one eigen and one force test).

leodmz:
> @leodmz Any idea how many eigen screenviews we get per day? If they are firing on every homefeed load (e.g. a nfy recs experiment, which we have historically run pretty frequently) it would be good to get an estimate of how many extra segment events that will produce.
> 
> In general we only need one `experiment_viewed` event per session in which an experiment was viewed, but since sessions are defined downstream it's difficult for clients to adhere to that.
> 
> We currently define an [idle timeout of 30 minutes](https://github.com/artsy/fulcrum/blob/main/dbt/models/sessions_web_folder/session_starts.sql#L52) in our session definition so one precaution would be for clients to send this event no more often than once every 30 minutes. That does add complexity to the client implementation though, so I would want to implement that only if we think we are going to come close to the 30M events per month limit when running two parallel tests (say one eigen and one force test).

[here ](https://artsy.looker.com/explore/analytics/session_pages_unified?qid=UAtsepzs3dKQmzss4Hl4Xf&toggle=dat,vis) - We have around 15k sessions on eigen per day or 20k home pageview. it wouldn't add up to more than a million events per month

cavvia:
@mc-jones @leodmz Is there anything blocking this? I'm not aware of any blockers on my end.

cavvia (approved):

mc-jones:
@cavvia - Nope, believe we have it all set now. Just a reminder that it will have a different event implementation from Eigen until an equivalent PR is made.

PR #13332, opened 2024-01-03
revert: revert the palette version to 37.2.2

narikazu:
The type of this PR is: **Revert**

### Description

Since upgrading the palette version to 37.2.3  was [merged](https://github.com/artsy/force/commit/b87756920b6fd9e8ca0d99da5cf94b18d7521c2a), [two tests](https://app.circleci.com/pipelines/github/artsy/force/56771/workflows/21833c23-7908-4f45-97e2-f54aaf6feadf/jobs/424662) have been failing. To unblock other works, this PR reverts the change:  https://github.com/artsy/force/pull/13318

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "@artsy/icons": "3.11.0",
     "@artsy/img": "1.0.3",
     "@artsy/multienv": "^1.2.0",
-    "@artsy/palette": "37.3.0",
+    "@artsy/palette": "37.2.2",
     "@artsy/palette-charts": "36.3.0",
     "@artsy/react-html-parser": "^3.0.2",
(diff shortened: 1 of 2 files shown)


dzucconi:
Eh, I don't see how this would cause a test failure. There's literally only the addition of a new text style.

dzucconi:
Ah, it's not that it's something in here. Probably the portal change.

dzucconi:
Actual bug with `DeepZoom` that I'm fixing. Now that Portals rely on state, there's a slightly strange set up in this component where there's a ref inside of the modal and an effect outside of that that is looking for it. So it doesn't initialize.

leamotta (approved):

dzucconi:
PR incoming

PR #13339, opened 2024-01-04
fix(EMI-1610): redirect after inventory insufficient

leamotta:

This PR solves [EMI-1610]

### Description

Routing after an error popup was broken (no redirection happened) due to the `route` prop being `nil` and trying to access `onTransition` in it.
This fixes it and changes where the user should land according to the new rules defined in the ticket. User won't see a confirmation to leave the page thanks to the work done here https://github.com/artsy/force/pull/7964/files

https://github.com/artsy/force/assets/7518671/010199c4-2a5b-46a7-8dbc-b68796400c39

[EMI-1610]: https://artsyproduct.atlassian.net/browse/EMI-1610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iskounen (approved):
I'm assuming that we were routing users to the artist page in hopes that they would find another artwork by the same artist that they would like to buy. (It may also have been in preparation of orders supporting more than 1 artwork.)

Thanks to alerts (and now saves) we now have better tools for match-making on the artwork page so I think this is a good change overall.

PR #13342, opened 2024-01-04
Fix: missing minutes and seconds in timer for partner offer

narikazu:

This PR solves [EMI-1628]

### Description

This PR adds the minutes and seconds to the timer in the checkout flow for a partner offer and fixes the progress bar by using the correct start date of the partner offer. 

Screenshot 2024-01-05 at 17 33 20

[EMI-1628]: https://artsyproduct.atlassian.net/browse/EMI-1628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Utils/Hooks/useCountdownTimer.ts
```diff
@@ -1,26 +1,50 @@
...
+    return `${hours}h ${minutes}m`
+  } else {
+    const timeDiff = differenceInSeconds(expiration, now)
+
+    const ONE_DAY = 86400 // 24 hours * 60 minutes * 60 seconds
+
+    const days = Math.floor(timeDiff / ONE_DAY)
+    const hours = Math.floor((timeDiff % ONE_DAY) / (60 * 60))
+    const minutes = Math.floor((timeDiff % (60 * 60)) / 60)
+    const seconds = Math.floor(timeDiff % 60)
```
iskounen, line 38: Instead of duplicating this logic with minutes vs seconds, why not always calculate it with seconds (and don't output the seconds if `includeSeconds` is false)?

src/Utils/Hooks/useCountdownTimer.ts
```diff
@@ -1,26 +1,50 @@
...
+
+    if (timeDiff <= 0) {
+      return "Expired"
+    }
+    if (timeDiff >= ONE_DAY) {
+      return `${days}d ${hours}h ${minutes}m ${seconds}s`
```
iskounen, line 44: Do we want to include the minutes and seconds if the time difference is >= ONE_DAY? If so, I think we can consolidate the logic in this method by having an `includeMinutes` flag that ensures minutes are always included.

leamotta (approved):
LGTM!

src/Utils/Hooks/useCountdownTimer.ts
narikazu: Thanks, @iskounen . That's a great idea. Updated the logic.

src/Utils/Hooks/useCountdownTimer.ts
narikazu: That's a good point. I'm going to ask Vivi if we want to show the minutes and seconds when the time difference is greater than ONE_DAY.

MrSltun (approved):
Looks good! Great job Kaz! 🌟

PR #13346, opened 2024-01-05
feat: New Notification route

olerichter00:
Addresses [ONYX-622]

## Description

This PR adds a route and app and query for the new Notification page. The implementation of the content of the page is not part of this PR.

I extracted most of the code from this prototype PR: https://github.com/artsy/force/pull/13165

[ONYX-622]: https://artsyproduct.atlassian.net/browse/ONYX-622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Notifications/NotificationApp.tsx
+++ b/src/Apps/Notifications/NotificationApp.tsx
@@ -1,0 +1,33 @@
+import { FullBleed } from "@artsy/palette"
+import { MetaTags } from "Components/MetaTags"
+import { createFragmentContainer, graphql } from "react-relay"
+import { NotificationApp_me$data } from "__generated__/NotificationApp_me.graphql"
+import { NotificationFragmentContainer } from "Components/Notifications/Notification"
+
+interface NotificationAppProps {
+  me: NotificationApp_me$data
+}
+
+const NotificationApp: React.FC<NotificationAppProps> = ({ me }) => {
+  return (
+    <>
+      <MetaTags title="Notification | Artsy" pathname="/notification" />
+
+      <FullBleed>
+        <NotificationFragmentContainer me={me} />
+      </FullBleed>
+    </>
+  )
+}
+
+export const NotificationAppFragmentContainer = createFragmentContainer(
+  NotificationApp,
+  {
+    me: graphql`
+      fragment NotificationApp_me on Me
+        @argumentDefinitions(notificationId: { type: "String!" }) {
+        ...Notification_me @arguments(notificationId: $notificationId)
+      }
+    `,
+  }
+)
--- a/src/Apps/Notifications/notificationsRoutes.ts
+++ b/src/Apps/Notifications/notificationsRoutes.ts
@@ -10,6 +10,14 @@ const NotificationsApp = loadable(
   }
 )
 
+const NotificationApp = loadable(
+  () =>
+    import(/* webpackChunkName: "notificationsBundle" */ "./NotificationApp"),
+  {
+    resolveComponent: component => component.NotificationAppFragmentContainer,
+  }
+)
+
 export const notificationsRoutes: AppRouteConfig[] = [
   {
     path: "/notifications",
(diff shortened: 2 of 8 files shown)


src/Components/Notifications/__tests__/Notification.jest.tsx
```diff
@@ -0,0 +1,38 @@
+import { screen } from "@testing-library/react"
+import { setupTestWrapperTL } from "DevTools/setupTestWrapper"
+import { graphql } from "react-relay"
+import { NotificationFragmentContainer } from "Components/Notifications/Notification"
+import { Notification_test_Query } from "__generated__/Notification_test_Query.graphql"
+
+jest.unmock("react-relay")
+
+const { renderWithRelay } = setupTestWrapperTL<Notification_test_Query>({
```
MounirDhahri, line 9: fwiw, I made some small improvements to `setupTestWrapperTL` to accept query renderers and therefore not need to define `Component` + `Query` and instead you can define the component using hooks and it should work fine as we do in Eigen.

src/Apps/Notifications/notificationsRoutes.ts
```diff
@@ -25,4 +33,18 @@ export const notificationsRoutes: AppRouteConfig[] = [
       }
     `,
   },
+  {
+    path: "/notifications/:id",
```
MounirDhahri, line 37: **nitpick:**
```suggestion
    path: "/notification/:id",
```

MounirDhahri (dismissed):
LGTM, just two minor comments

src/Apps/Notifications/notificationsRoutes.ts
olerichter00: Good catch! I guess using plural nouns for collections is best practice, but you're right. In Force & Eigen, we use singular nouns for these kinds of routes. ([example](https://github.com/artsy/eigen/blob/c872a882bb1a374e2b5db1d4a12e404de3209a99/src/app/routes.ts#L171)). I will update the URL.

MounirDhahri: I was also confused because we have unfortunately both patterns in some cases that's why I felt this should not be a blocker for us

src/Components/Notifications/__tests__/Notification.jest.tsx
olerichter00: Nice! At the moment the query is defined in [notificationRoutes.ts](https://github.com/artsy/force/blob/542aea7565f54ec1e130e67d80b6b9697d989592/src/Apps/Notifications/notificationsRoutes.ts#L21), so the component doesn't has a query defined. But this might change in the next PR with the split view implementation. I'll can then in this PR improve the test file to not define it's own query.

MounirDhahri: perfect - thanks for explaining @olerichter00. Please feel free to merge this

src/Apps/Notifications/NotificationApp.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface NotificationAppProps {
+  me: NotificationApp_me$data
+}
+
+const NotificationApp: React.FC<NotificationAppProps> = ({ me }) => {
+  return (
+    <>
+      <MetaTags title="Notification | Artsy" pathname="/notification" />
+
+      <FullBleed>
```
damassi, line 16: @olerichter00 - rather than wrap the component with FullBleed, you can define it via the `layout: "FullBleed"` prop directly on the route: https://github.com/artsy/force/blob/main/src/Apps/Components/Layouts/index.tsx#L23

olerichter00: Thanks! 🙏

PR #13355, opened 2024-01-09
feat: Notification page split view layout

olerichter00:
Addresses [ONYX-623]

## Description

This PR sets up the new Activity page split view, which displays the activity list on the left and the activity detail view on the right.

On mobile the page displays either only the list or only the detail view depending on the URL:

- `/notifications` -> List view
- /notification/$id` -> Detail view

When the user opens `/notifications` directly, the first activity/notification will be selected and displayed.

## Video & Screenshots

Screenshot 2024-01-09 at 17 56 57

https://github.com/artsy/force/assets/4691889/01dbac19-afa1-4a81-8feb-d539826a5e98

[ONYX-623]: https://artsyproduct.atlassian.net/browse/ONYX-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Notifications/NotificationsApp.tsx
```diff
@@ -1,27 +1,68 @@
...
+
+      <Media at="xs">
+        {isNotificationsPage ? (
+          <Notifications
+            mode="page"
+            unreadCounts={me.unreadNotificationsCount ?? 0}
+          />
+        ) : (
+          <NotificationQueryRenderer />
+        )}
```
olerichter00, line 63: Although this might not be the cleanest solution, it has some advantages:

I'm using the component `NotificationsApp` for both routes: `/notifications` and `/notification/$notificationId`. This allows for navigation from `/notifications` to `/notification/$notificationId` (by clicking on an element in the Activity list in the split view) without the Activity list reloading, as both routes render the same fragment container.

I'm interested in hearing your thoughts about the solution!

src/Components/Notifications/Notification.tsx
```diff
@@ -1,6 +1,21 @@
...
 import { createFragmentContainer, graphql } from "react-relay"
 import { Notification_me$data } from "__generated__/Notification_me.graphql"
+import { SystemQueryRenderer } from "System/Relay/SystemQueryRenderer"
+import { NotificationQuery } from "__generated__/NotificationQuery.graphql"
+import { useNotificationsContext } from "Components/Notifications/useNotificationsContext"
+
+export const SUPPORTED_NOTIFICATION_TYPES = [
+  "ARTWORK_ALERT",
+  "ARTWORK_PUBLISHED",
+]
```
olerichter00, line 18: Right now, we only have designs for these notification types. I'll clarify whether we want to show a preview for other "not-yet" supported types or simply link to the notification's `targetHref`.

src/Components/Notifications/Notification.tsx
```diff
@@ -38,3 +56,60 @@ export const NotificationFragmentContainer = createFragmentContainer(
...
+const QUERY = graphql`
+  query NotificationQuery($id: String!) {
+    me {
+      ...Notification_me @arguments(notificationId: $id)
+    }
+  }
+`
+
+export const NotificationQueryRenderer: React.FC = () => {
+  const { state } = useNotificationsContext()
```
olerichter00, line 69: I store the ID of the currently selected activity/notification in a context. This allows setting the first activity/notification in the list as active from `NotificationsList` for the route `/notifications` (which is a requirement).

src/Components/Notifications/NotificationsList.tsx
```diff
@@ -73,7 +85,7 @@ const NotificationsList: React.FC<NotificationsListProps> = ({
     }
 
     return (
-      <Box textAlign="center" mt={4}>
+      <Box textAlign="center" my={4}>
```
olerichter00, line 88: Because we want some padding below the "Load More" button.

src/Components/Notifications/NotificationsTabs.tsx
```diff
@@ -102,7 +102,9 @@ export const NofiticationsTabs: React.FC<NofiticationsTabsProps> = ({
         <MarkAllAsReadPanel unreadCounts={unreadCounts} />
       </Sticky>
 
-      {activeTab.current.child}
+      <Flex overflow="scroll" flexDirection="column">
+        {activeTab.current.child}
+      </Flex>
```
olerichter00, line 107: In order to enable scrolling in the split view.

src/Components/Notifications/useNotificationsContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    setCurrentNotificationId(match.params.notificationId)
+  }, [match.params.notificationId, setCurrentNotificationId])
+
+  return (
+    <NotificationsContext.Provider
+      value={{
+        state: { currentNotificationId },
+        setCurrentNotificationId,
+      }}
```
damassi, line 40: #minor, but if you assign the contents of `value` to a variable, it leads to more stable renders:

```tsx
const contextProps = {
  ...
}

<NotificationsContext.Provider value={contextProps}>.
```

damassi:
Nicely done @olerichter00

src/Components/Notifications/useNotificationsContext.tsx
dzucconi: @damassi do you have a reference for this? Intuitively, that doesn't make sense to me.

damassi: I'm looking for it! But i know what you mean. It seems like a `useMemo` would be required

damassi: Yeah, it looks like useMemo is how it is. But i could swear by experience that I stumbled upon this in the past, that for some reason the react render lifecycle would get funny on fully anonymous objects passed to contet, even tho they seem the same as passing an object. Anyways, nbd.

src/Apps/Notifications/NotificationsApp.tsx
```diff
@@ -1,27 +1,68 @@
...
 interface NotificationsAppProps {
   me: NotificationsApp_me$data
 }
 
-const NotificationsApp: React.FC<NotificationsAppProps> = props => {
-  const { me } = props
+const NotificationsApp: React.FC<NotificationsAppProps> = ({ me }) => {
+  const { match } = useRouter()
+
+  const isNotificationsPage = match.location.pathname === "/notifications"
```
anandaroop, line 22: **question:** Would a trailing slash in the URL (`/notifications/`) affect this, or would the router strip it out?

src/Components/Notifications/Notification.tsx
```diff
@@ -38,3 +56,60 @@ export const NotificationFragmentContainer = createFragmentContainer(
...
+
+  if (!state.currentNotificationId) {
+    return null
+  }
+
+  return (
+    <SystemQueryRenderer<NotificationQuery>
```
anandaroop, line 76: **question (non-blocking):** General question re: preferred practice in Force these days — `useLazyLoadQuery` vs `useClientQuery` vs `SystemQueryRenderer`. Is it just up to the developer, or are there reasons to prefer the QR here?

anandaroop (approved):
👍🏽

src/Apps/Notifications/NotificationsApp.tsx
```diff
@@ -1,27 +1,68 @@
...
-          mode="page"
-          unreadCounts={me.unreadNotificationsCount ?? 0}
-        />
-      </FullBleed>
-    </>
+      <Media greaterThan="xs">
+        <GridColumns>
+          <Column
+            span={3}
+            borderRight="1px solid #ddd"
```
nickskalkin, line 32: **super-minor-suggestion**: i would use one of pre-defined colours, in this case I think `black10`

nickskalkin, line 32:
```suggestion
            borderRight="1px solid"
            borderColor="black10"
```

nickskalkin (approved):
lgtm!

src/Apps/Notifications/NotificationsApp.tsx
dzucconi: oh yes _definitely_. you can specify borderColor as it's own prop

dzucconi: I'm going to do dark mode next week — and to do so _all_ hex values need to be banished from this codebase. I'm noticing this sort of thing in a few places. So to be clear: _never_ use a hex value. Always use a system value.

src/Apps/Notifications/NotificationsApp.tsx
olerichter00: Good catch! It does not. Replacing `===` with `startsWith` should solve the issue.

src/Components/Notifications/Notification.tsx
olerichter00: I changed the code to use `useLazyLoadQuery` instead of `QueryRenderer` 👍 
I'm not sure about `useClientQuery` and when to use it over `useLazyLoadQuery`. 🤔

src/Apps/Notifications/NotificationsApp.tsx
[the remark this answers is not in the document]
olerichter00: Sorry, forgot to commit 🙄 

https://github.com/artsy/force/pull/13373

PR #13366, opened 2024-01-10
feat(onyx-635): Support rendering of PartnerOfferCreated activity events

nickskalkin:

This PR solves [ONYX-635]

### Description

[Review app](https://partner-offer-notifications.artsy.net/) (I've added code snippets how to generate partner offer notifications [here](https://www.notion.so/artsy/Activity-Panel-cd0a9668895848a484b51bebb9723a8f?pvs=4#a9bf1a8aea894db7a46c762a0ae314eb))

Implementation differs a bit from design specs, because design specs are aimed for a new activity panel.

image

[ONYX-635]: https://artsyproduct.atlassian.net/browse/ONYX-635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Notifications/NotificationItem.tsx
```diff
@@ -64,21 +66,30 @@ const NotificationItem: React.FC<NotificationItemProps> = ({ item }) => {
     })
   }
 
+  // Temporary until new activity panel is ready
+  const message = () => {
+    if (item.notificationType === "PARTNER_OFFER_CREATED") {
+      return "Offer"
```
dblandin, line 72: **question**: Is this override necessary if the backend `message` is identical ([source](https://github.com/artsy/gravity/blob/4377ac31d091ff3d0b359b2fb81c517f14a66afe/app/models/activity_feed/partner_offer_created_activity.rb#L17))?

dblandin (approved):
This looks great to me!

src/Components/Notifications/NotificationItem.tsx
nickskalkin: Yeah, unfortunately we need it here. Metaphysics call `toLowerCase()` on messages ([code](https://github.com/artsy/metaphysics/blob/main/src/schema/v2/notifications/index.ts#L51)). I think we should remove it from there and change gravity to create activities with proper messages. I was just about to create a ticket for this.

dblandin: Gotcha. Thanks for the explanation. 👍🏼 

I suppose we could omit this message entirely depending on what we hear back from Barney regarding designs that fit into the existing structure. As is, it might feel a bit repetitive without much added value.

> Limited Time Offer · Today
> Saved work by Heather Boose Weiss
> ⏱️ Expires in 2 days

MounirDhahri (approved):
Nice!

dblandin (approved):
Awesome! Looks great 💯

olerichter00 (approved):
Great work! 🌟

PR #13368, opened 2024-01-10
fix: failing test after palette upgrade

iskounen:
We noticed that Force's test suite was failing after some dependency updates were auto-merged. We believe that the following updates could have changed something that caused these tests to fail:

- https://github.com/artsy/force/pull/13360
- https://github.com/artsy/force/pull/13363

The existing tests were failing because the "Offer" stepper button in the checkout flow was wrapped inside another button, like this:

```
<button>
  <div>
    <button>
      Offer
    </button>
  </div>
<button>
```

And the testing library was recognizing this as two buttons with the same text, so the tests were asserting the existence of 2 buttons. Whatever caused these tests to fail was a blessing in disguise because this allows the test to assert the existence of 1 button.

### DIFF
--- a/src/Apps/Order/Routes/Shipping2/__tests__/Shipping2.jest.tsx
+++ b/src/Apps/Order/Routes/Shipping2/__tests__/Shipping2.jest.tsx
@@ -345,7 +345,7 @@ describe("Shipping", () => {
           relayEnv
         )
 
-        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(2)
+        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(1)
       })
 
       it("renders fulfillment selection if artwork is available for pickup", async () => {
--- a/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
+++ b/src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
@@ -293,7 +293,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(2)
+        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(1)
       })
 
       it("renders fulfillment selection if artwork is available for pickup", async () => {


src/Apps/Order/Routes/Shipping2/__tests__/Shipping2.jest.tsx
```diff
@@ -345,7 +345,7 @@ describe("Shipping", () => {
           relayEnv
         )
 
-        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(2)
+        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(1)
```
MrSltun, line 348: Suggestion(non-blocking): maybe we could check for visibility instead?

```suggestion
        expect(screen.getByRole("button", { name: "Offer" })).toBeVisible()
```

MrSltun (approved):
Thank you Adam! 🙏  I just left a small suggestion but feel free to merge :)

src/Apps/Order/Routes/Shipping2/__tests__/Shipping2.jest.tsx
iskounen: The test failed when I asserted its visibility, so I took the shortcut of counting buttons instead. I agree that asserting its visibility is the better way to write this test.

src/Apps/Order/Routes/__tests__/Shipping.jest.tsx
```diff
@@ -293,7 +293,7 @@ describe("Shipping", () => {
           Me: () => meWithoutAddress,
         })
 
-        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(2)
+        expect(screen.getAllByRole("button", { name: "Offer" }).length).toBe(1)
```
erikdstock, line 296: [Commenting on the other identical fix to start a new thread] Before merging, do we want to understand _why_ this palette upgrade caused there to no longer be two elements with `role=button`?

From discussing with the web practice, I think we are ok with deploying the fix and letting the mystery be.

erikdstock (approved):
Good from here, had one question but got the 👍  from web practice.

PR #13373, opened 2024-01-12
chore: Use borderColor instead of hex value

olerichter00:
## Description

Use borderColor instead of hex value.


dzucconi (approved):
🙏 😄

PR #13378, opened 2024-01-16
Feat: Address autocomplete for auction registration and user settings

erikdstock:

This PR adds address autocomplete to the auction registration form and the settings > shipping address modal.
<details><summary>Screencap (UX is identical for both forms)</summary>

</details> 

TODOs if we wanted to merge one or both of these:
- the form layout is not logical for an autocompleted address field because the postal code comes first (see screencap).
- need to clarify analytics values - I did not immediately find a context module for auction registration.
- the auction registration address form is technically a billing address form. we may want to add an '...and also save to my shipping addresses' box. However this might impact auction ops.
- it might make sense to move the (now 3x duplicated) `trackAutocompleteEdits` helper somewhere. It makes sense to put it in the hook to me (for the future secondary address field search work, we would like to know if an address was already autocompleted) but that might require more thought. doesn't need to block this ticket.
- may impact integrity tests, but I hope not.
cc @dzucconi

### DIFF
--- a/src/Apps/Auction/Components/Form/AddressForm.tsx
+++ b/src/Apps/Auction/Components/Form/AddressForm.tsx
@@ -1,9 +1,41 @@
 import { Column, GridColumns, Input } from "@artsy/palette"
 import { CountrySelect } from "Components/CountrySelect"
 import { useFormContext } from "Apps/Auction/Hooks/useFormContext"
+import {
+  AddressAutocompleteInput,
+  useAddressAutocompleteTracking,
+} from "Components/Address/AddressAutocompleteInput"
+import { useState } from "react"
+import { useAnalyticsContext } from "System/Analytics/AnalyticsContext"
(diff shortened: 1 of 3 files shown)


rquartararo:
Testing this out now!

rquartararo:
When you say "if we wanted to merge one or both of these", is there another PR I'm not aware of?

erikdstock:
> When you say "if we wanted to merge one or both of these", is there another PR I'm not aware of?

oh, i just meant that it adds two new features at once.

It seems like there is more motivation behind auction registration, but that both could be relatively little work, but that also we should have a consistent layout with the postal code in a consistent place ideally. So my preference would be to do these one at a time after clarifying design or just making it consistent with the shipping route's input layout.

erikdstock:
Closing since this was for demonstration purposes.

PR #13379, opened 2024-01-17
feat: add link to artwork in checkout

leamotta:

### Description

While in checkout, a collector might be still thinking about the artwork’s particular details and wants to double check something by going to the main artwork page. Currently they would have to hit back on their browser and lose progress in the flow.
Therefore, it would be cool if we can allow collectors to open the artwork page in a new tab by clicking on the artwork details in checkout.

Furthermore, with partner offers releasing soon, a collector might land directly in checkout from a push notification or an email, further complicating this flow as the browser back button does not go to the artwork page.

https://github.com/artsy/force/assets/7518671/49be08d8-6d60-437c-8f9a-0d54553b65b5


MrSltun (commented):
I like the idea! 🦅 
But should we also implement it on the artwork title as well?

leamotta:
> I like the idea! 🦅 But should we also implement it on the artwork title as well?

Good point, I wasn't sure where to put the link, my first thought was to apply it to the whole artwork details box but went conservative in the end. 
Were you thinking the title would be underlined in that scenario to identify it as a link?
Let's get feedback from design on Slack!

MrSltun:
> Were you thinking the title would be underlined in that scenario to identify it as a link?

My approach would be without an underline to not distract the user from the flow, but it's there if they wanted to go to the artwork page

src/Apps/Order/Components/ArtworkSummaryItem.tsx
```diff
@@ -52,25 +53,35 @@ const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
...
 
-  const priceLabel = ((mode === "OFFER") || isPartnerOffer) ? "List price" : "Price"
+  const priceLabel = mode === "OFFER" || isPartnerOffer ? "List price" : "Price"
 
   return (
     <StackableBorderBox flexDirection="row" {...others}>
       <Box height="auto">
-        {imageURL && <Image src={imageURL} alt={title!} width="55px" mr={1} />}
+        {imageURL && (
+          <Link href={`/artwork/${artwork?.slug}`} target="_blank">
```
iskounen, line 62: For private offers, we may not have a public artwork page for the artwork that is being purchased. If the order `mode` is available, it may be worth checking before adding the href.

leamotta: Great point! I'll check

MrSltun (approved):
Looks great! 🌟 Thank you Leandro 🦅

PR #13380, opened 2024-01-17
Feat: Adds support for a timed (non-auction) sale

sweir27:
The type of this PR is: Build

### Description

The goal of this work is to support non-auction sales in our clients. Support for this feature has waned over time as we've rebuilt UIs and as the business has stopped running these. But we want to bring them back! The main task here is to go through and remove auction-specific references for these non-auction sales.

`Auction`s are actually a sub-type of these generic `Sale`s. Since the `Sale` itself (just a grouping of artworks with a timer) is a simple concept, I opted to create a new app instead of trying to retro-fit the existing `Auction` app. Since these are not auctions, I used our standard artwork grid which removes any mention of "lot".

The work included here was to:
- Add the new SaleApp, give it all the basics (header/description/artwork grid)
- Update the Artwork page to not reference "Auction" if the artwork is in a non-auction sale

There is still additional work that needs to be done for this to be complete, but that can be in a separate PR:
- Updating search results and autocomplete, which still reference "auction"
- Add back "Works by artists you follow" rail (if we want it) on the sale page
- Add an "Other works from this sale" section on the artwork page (if we want it)
- Similar updates in the App

Before & Afters
image
image
image


manifest-novo.json
```diff
@@ -0,0 +1,158 @@
+{
```
sweir27, line 1: This was added automatically – I'm assuming I should remove from the PR, but would like to know why it keeps being generated

dzucconi: @damassi? First I've ever seen this.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
```diff
@@ -132,6 +132,13 @@ export const ArtworkSidebar: React.FC<ArtworkSidebarProps> = ({
         <ArtworkSidebarCommercialButtonsFragmentContainer artwork={artwork} />
       )}
 
+      {!isInAuction && !hasEnded && sale && !sale.isAuction && (
```
dzucconi, line 135: This seems like it should be extracted into some named condition in Metaphysics?

sweir27: Oh yeah – this is messy. I'd like to confirm the logic first and then happy to extract it (or maybe it'll be simple enough to do here).

dzucconi (approved):
LGTM — I suspect @damassi will know about that manifest though.

manifest-novo.json
damassi: This is ancient! I think it might be from your old copy of force, before all this was cleared up (and therefore might be missing from gitignore? 

History is: we had two parallel webpack configs going temporarily while we figured out how to migrate away from old config for v2 work.

sweir27: 😅 Okay cool, that tracks!! Thanks @damassi 🙏

egdbear (dismissed):
🚀

damassi (dismissed):
Welcome back @sweir27 - we miss you in the code!!! 

Mind adding a quick exist-y smoke test before merge? (See [here](https://github.com/artsy/force/blob/main/cypress/e2e/artwork.cy.js) for an example.)

sweir27:
Alright @dzucconi I updated the sale timer logic and added a few tests – didn't move to MP because it ended up being pretty simple.

src/Apps/Sale/Components/SaleArtworks.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const aggregations = ["ARTIST", "MEDIUM", "TOTAL"]
+
+  if (user) {
+    aggregations.push("FOLLOWED_ARTISTS")
+  }
+
+  // Shared with saleRoutes
+  return {
+    aggregations,
+    first: 39,
```
dzucconi, line 111: 39?

dzucconi:
Feel free to merge!

src/Apps/Sale/Components/SaleArtworks.tsx
sweir27: Borrowed this from the auction artwork filter... https://github.com/artsy/force/pull/9667 but this is a case of copy pasta 😬.

I'll leave in for now but adding a follow-up since I'm curious what the context is.

PR #13392, opened 2024-01-18
feat(ONYX-627): activity list pills

tam-kis:
The type of this PR is: Feat

This PR solves [ONYX-627](https://artsyproduct.atlassian.net/browse/ONYX-627)

### Description

Refactor and replace existing NotificationsTabs with Pills.
Separate Notifications header and pills components.

Dropdown:

https://github.com/artsy/force/assets/17580625/3879ee8b-9b13-4978-846c-1bde392d9278

Activity Panel:

https://github.com/artsy/force/assets/17580625/653d1673-ffbc-403d-a367-100e65b2bb97

No alerts, empty state:

https://github.com/artsy/force/assets/17580625/eef6f0b8-f2f3-4c6c-97dc-9dd763bfdac9

[ONYX-627]: https://artsyproduct.atlassian.net/browse/ONYX-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Notifications/NotificationsEmptyStateByType.tsx
+++ b/src/Components/Notifications/NotificationsEmptyStateByType.tsx
@@ -22,6 +22,10 @@ const emptyStateByType: Record<
     message:
       "Create alerts on an artist or artwork page and get notifications here when there’s a match.",
   },
+  following: {
+    title: "",
+    message: "",
+  },
 }
 
 export const NotificationsEmptyStateByType: React.FC<NotificationsEmptyStateByTypeProps> = ({
--- a/src/Components/Notifications/NotificationsEmptyStateByType.tsx
+++ b/src/Components/Notifications/NotificationsEmptyStateByType.tsx
@@ -13,7 +13,8 @@ const emptyStateByType: Record<
   }
 > = {
   all: {
-    title: "Follow artists and galleries to stay up to date",
+    //TODO: add new copy for "all" notifications
+    title: "Stay up to date, follow artists and galleries",
     message:
       "Keep track of the art and events you love, and get recommendations based on who you follow.",
   },
@@ -23,8 +24,9 @@ const emptyStateByType: Record<
       "Create alerts on an artist or artwork page and get notifications here when there’s a match.",
   },
   following: {
-    title: "",
-    message: "",
+    title: "Follow artists and galleries to stay up to date",
+    message:
+      "Keep track of the art and events you love, and get recommendations based on who you follow.",
   },
 }
 
--- a/src/Components/Notifications/NotificationsEmptyStateByType.tsx
+++ b/src/Components/Notifications/NotificationsEmptyStateByType.tsx
@@ -14,7 +14,7 @@ const emptyStateByType: Record<
 > = {
   all: {
     //TODO: add new copy for "all" notifications
-    title: "Stay up to date, follow artists and galleries",
+    title: "Follow artists and galleries to stay up to date",
     message:
       "Keep track of the art and events you love, and get recommendations based on who you follow.",
   },
--- a/src/Components/Notifications/NotificationsHeader.tsx
+++ b/src/Components/Notifications/NotificationsHeader.tsx
@@ -1,0 +1,54 @@
+import {
+  Box,
+  Clickable,
+  DROP_SHADOW,
+  Flex,
+  Separator,
+  Spacer,
+  Text,
+} from "@artsy/palette"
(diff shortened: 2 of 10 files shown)


src/Components/Notifications/NotificationsHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export const NotificationsHeader: React.FC<NotificationsHeaderProps> = ({
+  mode,
+}) => {
+  return (
+    <Box
+      backgroundColor="white100"
+      width="100%"
+      p="20px 20px 0"
```
tam-kis, line 26: A more elegant way to apply padding with no bottom padding?

Screenshot 2024-01-18 at 18 37 41

src/Components/Notifications/NotificationsHeader.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <Flex
+        flexDirection="row"
+        justifyContent="space-between"
+        alignItems="center"
+      >
+        <Flex justifyContent="flex-start">
+          <Text variant="lg-display">Activity</Text>
+        </Flex>
+        <Flex justifyContent="flex-end" gap={4}>
+          <MoreIcon />
```
tam-kis, line 38: Just an icon/placeholder for now, this will be an icon that opens a dropdown.

src/Components/Notifications/NotificationsEmptyStateByType.tsx
```diff
@@ -22,6 +22,10 @@ const emptyStateByType: Record<
     message:
       "Create alerts on an artist or artwork page and get notifications here when there’s a match.",
   },
+  following: {
+    title: "",
+    message: "",
```
tam-kis, line 27: Will replace with real copy when we clarify it.

src/Components/Notifications/NotificationsHeader.tsx
nickskalkin: I would rather do something like:

```
px=2
pt=2
```

nickskalkin:
Minor remark - according to design, separator should be full-width, but has some margin:

Screenshot 2024-01-19 at 11 31 44

nickskalkin:
@tam-kis is it intentional that selected pill is blue? In Figma selected pill is black but everywhere else (search pill, for instance) they are blue, like in your implementation, so I think maybe this is ok to keep it light that.

nickskalkin (approved):
👍 looks great! left two small questions, but apart from that all good!

tam-kis:
> Minor remark - according to design, separator should be full-width, but has some margin:
> 
> Screenshot 2024-01-19 at 11 33 44

Updated, good catch!
Screenshot 2024-01-19 at 11 55 18

tam-kis:
> @tam-kis is it intentional that selected pill is blue? In Figma selected pill is black but everywhere else (search pill, for instance) they are blue, like in your implementation, so I think maybe this is ok to keep it light that.

We don't have a [variant](https://palette-storybook.artsy.net/?path=/story/components-pill--variants) for the black pills, throughout the platform we use blue pills so it should be first updated in Palette.

src/Components/Notifications/NotificationsPills.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { trackEvent } = useTracking()
+  const { setCurrentNotificationFilterType, state } = useNotificationsContext()
+
+  const handleClick = tabNumber => {
+    setCurrentNotificationFilterType(
+      NOTIFICATIONS_PILLS[tabNumber].name as NotificationType
+    )
+  }
+
+  const sendAnalytics = pill => {
```
dariakoko, line 22: **minor:** the attribute named `pill` is a little misleading as we already have a value of a different type with the same name. What do you think about renaming it? `pillName` for example?

dariakoko, line 22: **minor:** the attribute named `pill` is a little misleading as we already have a value of a different type with the same name. What do you think about renaming it? `pillName` for example?

dariakoko (approved):
Looks good!
I only left one minor comment

src/Components/Notifications/NotificationsEmptyStateByType.tsx
tam-kis: Updated copy for "Following", will update "All" copy async once there is a decision.

PR #13395, opened 2024-01-18
fix(errorboundary): fixes client-side routing-error error

dzucconi:
@joeyAghion [noticed](https://artsy.slack.com/archives/CP9P4KR35/p1705618588927709?thread_ts=1705595615.137369&cid=CP9P4KR35) our `ErrorBoundary` was... erroring.

It's due to found's `HttpError` class, not extending `Error`. So there's no message. While I'm in here I just made it so that our client-side routing errors will behave in the same manner as a server-side one, in terms of what they display.


joeyAghion:
🙇

PR #13396, opened 2024-01-19
fix: fix long press on search dropdown items

nickskalkin:

### Description

When I press (long click) on search results in the dropdown - redirect doesn't happen. This PR fixes the issue.

| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/a627fea6-4641-4563-a211-e3269af39e3e" /> | <video src="https://github.com/artsy/force/assets/3934579/e00cde18-3383-4aae-9aed-7046ccb3b9af" /> |

### DIFF
--- a/src/Components/Search/SearchBarInput.tsx
+++ b/src/Components/Search/SearchBarInput.tsx
@@ -177,6 +177,7 @@ const SearchBarInput: FC<SearchBarInputProps> = ({
     })
 
     resetValue()
+    redirect(option.href)
   }
 
   const handleFocus = () => {
@@ -221,6 +222,7 @@ const SearchBarInput: FC<SearchBarInputProps> = ({
       defaultValue={value}
       onChange={handleChange}
       onClear={resetValue}
+      onSelect={handleSelect}
       onSubmit={handleSubmit}
       onClick={handleFocus}
       header={


nickskalkin:
@dzucconi I know you've recommended relying on Link's native redirect behaviour, but for some reason it doesn't work in this case. My theory (might be complete nonsense though), is that when a user _presses_ on a search result item and then releases the mouse button (or trackpad) - Link is no longer in the DOM (because dropdown is closed / destroyed), thus redirect doesn't happen.

dariakoko:
Hey @nickskalkin 
I tested your changes, the long press issue is fixed. However, I cannot open the search result in a new tab any more. I believe this is confronting with this [resent PR](https://github.com/artsy/force/pull/13350). 
Do you think we could address both issues or should we compromise?

nickskalkin:
@dariakoko I'm not sure that I will be able to fix both quickly, but the issue with long press (which in reality is not even a **long** press, just a slightly longer click) definitely has higher priority. I would deploy this PR and then try to find a solution that satisfies both requirements, wdyt?

dariakoko:
@nickskalkin Sounds good, thank you 🙏 
Let me know if I can be helpful here

dzucconi:
Would it not make sense to fix the root cause?

joeyAghion:
:ceiling-cat: what's that root cause?

dzucconi:
@joeyAghion Presumably something to do with the `mousedown` handler in the `AutocompleteInput`? It'd be worth investigating though!

nickskalkin:
FYI: @dariakoko and I have pair programming session scheduled for tomorrow to try to find the root cause and fix it

nickskalkin:
@dzucconi question. Do you know why we have `onMouseDown` [here](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx#L385)? @dariakoko and I are wondering if this should be `onClick`. Seems like using `onClick` fixes the issue.

The problem is that when user presses on an option, in `handleMouseDown` function UI gets cleared (dropdown destroyed) after 100ms [delay](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/AutocompleteInput/AutocompleteInput.tsx#L141). If user keeps mouse pressed - dropdown UI is cleared before user releases the mouse button. But seems like router links (that are inside options) start handling redirect only after user releases mouse. But at this point of time there are no links in DOM, so redirects are not handled.

`onClick` will wait until user releases the mouse button, and only after that will clear UI.

Sorry it explanation is convoluted. Let's have a call if something is not clear

dzucconi:
@nickskalkin I think you're right. I don't recall the specific motivation, but it does look/feel like a mistake. Let's change that to click and go from there.

nickskalkin:
@dzucconi, oki, I'll open PR, thanks

PR #13405, opened 2024-01-24
fix(artist): Fix jump bug

damassi:

### Description

This fixes an issue that was [observed here](https://artsyproduct.atlassian.net/browse/ONYX-665), where scrolling after jump was blowing up scrolling on the page.

The issue has to do with useEffect dependencies. If we want something to execute once _and only once_, we shouldn't pass dependencies. Ignore that bloody react-hooks linter when it doesn't make sense 😄.


anandaroop:
Pushed up the [suggestion here](https://artsy.slack.com/archives/C05EQL4R5N0/p1706131688137069?thread_ts=1706034210.971619&cid=C05EQL4R5N0) for posterity, but closing this PR now.

PR #13329, opened 2024-01-03
chore(dia-325): retires test branch

dzucconi:
Closes [DIA-325](https://artsyproduct.atlassian.net/browse/DIA-325)

Retires the `diamond_revised-artwork-filters` branch and removes related code. Will archive flag in [Unleash ](https://unleash.artsy.net/projects/default/features/diamond_revised-artwork-filters)once deployed.

PR #13410, opened 2024-01-24
fix(artist): disable artist page jumpTo when coming from alert context

anandaroop:

This follows up [a few failed attempts](https://github.com/artsy/force/pull/13405) to solve the alerts-related artist page scroll bug [ticketed here](https://artsyproduct.atlassian.net/browse/ONYX-665).

At this point we are getting rid of the bug by actually _disabling_ the automatic `jumpTo` behavior completely. (It seems to interact poorly with the new filter UI which was released in https://github.com/artsy/force/pull/13329)

I'll leave the bug ticket open, in case anyone wants to take another swing at this. For context, the discussion took place in [this thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1706034210971619).


dzucconi (approved):
Yes

PR #11399, opened 2022-11-17
fix(CX-3129): Cleanup Inquiry flow

MrSltun:
The type of this PR is: **Fix**

This PR solves [CX-3129]

### Description
This PR updates the Inquiry flow by removing a few steps that collect information that are not being used in the Collector Resume

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CX-3129]: https://artsyproduct.atlassian.net/browse/CX-3129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13421, opened 2024-01-25
fix(EMI-1640): Update ask specialist workflow engine config

MrSltun:

This PR solves [EMI-1640]

### Description
This PR updates the Ask Specialist inquiry modal config to fix showing the Error status in the button.
~~We also updated the waiting period from `0.5s` to `1.5s` so users don't feel surprised when the modal suddenly disappears~~
We also added a toast when the message gets sent.

#### The motive behind the solution:
The fix here was to remove the `isLoggedOut` check from the `AskSpecialist` config for two reasons:
- We simplified the workflow a while ago in #11399. In the previous workflow, we used to have a confirmation modal at the end of the steps, so if the user was already logged in, it would skip the `isLoggedOut` check and move on to the confirmation modal.
- We don't have an Ask Specialist inquiry flow anymore, and if the user was logged in, the `isLoggedOut` check doesn't have a next step to go to, and it breaks

Due to these reasons, removing the check made sense

### Screenshots & Videos
| Before | After |
|---|---|
| <video src="https://github.com/artsy/force/assets/20655703/5f0f4ef8-3666-4b63-be5b-1ff3b9ad7c2a" /> | <video src="https://github.com/artsy/force/assets/20655703/78e66bfd-f801-491f-a726-6d558696d484" /> |

| New Toast |
|---|
| image |

[EMI-1640]: https://artsyproduct.atlassian.net/browse/EMI-1640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Inquiry/Views/InquirySpecialist.tsx
+++ b/src/Components/Inquiry/Views/InquirySpecialist.tsx
@@ -59,7 +59,7 @@ export const InquirySpecialist: React.FC = () => {
         contactGallery: false,
       })
       setMode("Success")
-      await wait(500)
+      await wait(1500)
       next()
     } catch (err) {
       logger.error(err)
--- a/src/Components/Inquiry/Views/InquirySpecialist.tsx
+++ b/src/Components/Inquiry/Views/InquirySpecialist.tsx
@@ -6,6 +6,7 @@ import {
   Spacer,
   Text,
   TextArea,
+  useToasts,
 } from "@artsy/palette"
 import * as React from "react"
 import { useSystemContext } from "System/useSystemContext"
@@ -23,6 +24,7 @@ type Mode = "Pending" | "Sending" | "Error" | "Success"
 
 export const InquirySpecialist: React.FC = () => {
   const { user } = useSystemContext()
+  const { sendToast } = useToasts()
 
   const { next, setInquiry, inquiry, artworkID } = useInquiryContext()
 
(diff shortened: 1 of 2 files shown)


src/Components/Inquiry/Views/InquirySpecialist.tsx
```diff
@@ -59,7 +59,7 @@ export const InquirySpecialist: React.FC = () => {
         contactGallery: false,
       })
       setMode("Success")
-      await wait(500)
+      await wait(1500)
```
dzucconi, line 62: What about just sending a toast?

dzucconi (approved):

src/Components/Inquiry/Views/InquirySpecialist.tsx
MrSltun: A toast is a great idea! I'll do that now :)

src/Components/Inquiry/config.ts
```diff
@@ -31,14 +31,7 @@ export const useEngine = ({ context, onDone }: UseEngine) => {
...
           askSpecialist: {
-            true: [
-              "Specialist",
-              {
-                isLoggedOut: {
-                  true: ["Account"],
-                },
-              },
-            ],
+            true: ["Specialist"],
```
iskounen, line 34: suggestion (non-blocking): I understand the problem from the PR description, but I don't understand why this solves it. Perhaps a bit of context about why this solves the problem in the description would be helpful.

iskounen (approved):

src/Components/Inquiry/config.ts
MrSltun: Great point! just updated it :)

PR #13431, opened 2024-01-26
fix: Open first notification correctly

olerichter00:
## Description

When navigating to `/notifications` the first notification in the list should open, but for some reason, it is not happening. This fix solves the issue. With this change also the URL is correctly replaced with `/notification/${firstNotificationID}`.

| Before | After |
| --- | --- |
|Screenshot 2024-01-26 at 16 15 42 |Screenshot 2024-01-26 at 16 15 53 |

### DIFF
--- a/src/Components/Notifications/NotificationsList.tsx
+++ b/src/Components/Notifications/NotificationsList.tsx
@@ -21,6 +21,7 @@ import { shouldDisplayNotification } from "./util"
 import { NotificationsListPlaceholder } from "./NotificationsListPlaceholder"
 import { useNotificationsContext } from "Components/Notifications/useNotificationsContext"
 import { NotificationListMode } from "Components/Notifications/NotificationsTabs"
+import { useRouter } from "System/Router/useRouter"
 
 interface NotificationsListQueryRendererProps {
   mode: NotificationListMode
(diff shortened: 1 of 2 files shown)


src/Components/Notifications/NotificationsList.tsx
```diff
@@ -65,8 +67,9 @@ const NotificationsList: React.FC<NotificationsListProps> = ({
       return
     }
 
-    setCurrentNotificationId(firstNotificationId)
-  })
+    router.replace(`/notification/${firstNotificationId}`)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
olerichter00, line 71: `NotificationContext` listens to URL changes and updates the notification ID automatically.

src/Components/Notifications/useNotificationsContext.tsx
```diff
@@ -38,6 +38,8 @@ export const NotificationsContextProvider: FC<NotificationsContextProviderProps>
   ] = useState<NotificationType>("all")
 
   useEffect(() => {
+    if (!match.params.notificationId) return
```
olerichter00, line 41: No need to set the ID if there is no URL in the query params.

tam-kis (commented):
LGTM, it should be enough just to rebase from main since Nikita fixed the failing test.

tam-kis (approved):

MounirDhahri (approved):
Thanks for the fix!

PR #13433, opened 2024-01-26
fix: Notification list scroll behaviour

olerichter00:
## Description

This PR fixes the scrolling behavior of the Activity list in the split view, which was broken after updating the design of the activity list (https://github.com/artsy/force/issues/13392).

Basically, it wasn't possible to scroll the list independently.

### DIFF
--- a/src/Components/Notifications/NotificationsWrapper.tsx
+++ b/src/Components/Notifications/NotificationsWrapper.tsx
@@ -47,6 +47,7 @@ export const NotificationsWrapper: React.FC<NotificationsWrapperProps> = ({
           </Sticky>
 
           <Box
+            // The notification list needs a maximum height to be independently scrollable.
             overflow="scroll"
             maxHeight={`calc(100vh - ${DROPDOWN_CONTENT_HEIGHT}px)`}
             pb={2}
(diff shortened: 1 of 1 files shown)


src/Components/Notifications/NotificationsWrapper.tsx
```diff
@@ -44,9 +44,13 @@ export const NotificationsWrapper: React.FC<NotificationsWrapperProps> = ({
             <Separator />
           </Sticky>
 
-          <Flex overflow="scroll" flexDirection="column">
+          <Box
+            overflow="scroll"
+            maxHeight={`calc(100vh - ${DROPDOWN_CONTENT_HEIGHT}px)`}
+            pb={2}
```
olerichter00, line 50: Adding this to have some bottom padding below the "Show more" button.

tam-kis (approved):

MounirDhahri (approved):
oh! missed this before

MounirDhahri:
To avoid the same regression, do you mind if you add a comment about it to make sure we don't break this again in the future

olerichter00:
> To avoid the same regression, do you mind if you add a comment about it to make sure we don't break this again in the future



Good idea! I've added a comment 👍

MounirDhahri (approved):
Thanks!

PR #13434, opened 2024-01-29
feat(onyx-630): article featured artist activity panel notification

nickskalkin:

This PR solves [ONYX-630]

### Description

Screenshot 2024-01-30 at 12 10 45

[ONYX-630]: https://artsyproduct.atlassian.net/browse/ONYX-630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Notifications/ArticleFeaturedArtistNotification.tsx
```diff
@@ -0,0 +1,174 @@
...
+  if (!article || !artists) {
+    return (
+      <Text variant="lg" m={4}>
+        Sorry, something went wrong.
+      </Text>
+    )
+  }
+
+  return (
+    <Box mx={4} my={4}>
```
tam-kis, line 47: I think this can be just `m={4}`

src/Components/Notifications/ArticleFeaturedArtistNotification.tsx
```diff
@@ -0,0 +1,174 @@
...
+      <Flex width="100%" justifyContent="space-between">
+        <Flex flex={1}>
+          <Text fontWeight="bold" variant="xl">
```
tam-kis, line 50: The in [figma](https://www.figma.com/file/PO3rjpDV29YOFR8kLI4InJ/Activity-Panel-%26-Alerts-Management?type=design&node-id=772-7261&mode=design&t=tc4NGSMnkuTRYYc3-0) headline should be just `variant="lg-display"` without `bold`

src/Components/Notifications/ArticleFeaturedArtistNotification.tsx
```diff
@@ -0,0 +1,174 @@
...
+          return (
+            <RouterLink
+              to={artist.href}
+              key={artist.internalID}
+              textDecoration="none"
+            >
+              <Text>{artist.name}</Text>
+            </RouterLink>
+          )
+        })}
```
dariakoko, line 80: **Question:** Have we discussed the case when we have more than one featured artist? I'm working on the same ticket but for Eigen - should I reuse your implementation?

nickskalkin: Yep, I've asked Barney if we should stick to this approach and he agreed @dariakoko

dariakoko: Perfect, thank you!

src/Components/Notifications/ArticleFeaturedArtistNotification.tsx
nickskalkin: Yeah, I've also spotted this different in other places ([here](https://github.com/artsy/force/blob/main/src/Components/Notifications/ArtworkPublishedNotification.tsx#L36) for example), but decided to follow what is already in the codebase, otherwise UI will look whimsical.

By the way the same goes for another comment from you above, I simply took [this code](https://github.com/artsy/force/blob/main/src/Components/Notifications/ArtworkPublishedNotification.tsx#L35) without looking too much into it. Is it ok if I do all these changes all-together in a separate PR?

nickskalkin: never mind, fixed both

tam-kis (approved):

PR #13441, opened 2024-01-30
fix: fixes unwanted redirect to a notification page from notifications list

nickskalkin:

### Description

Fixes unwanted redirect for mWeb, when user opens `/notifications` page and gets redirected to a top notification page right away. See before and after (notice that only mWeb was broken, I've attached desktop video just to demo that it works after changes)

Before:

| mWeb | Desktop |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/62abe90f-8df5-4688-baa8-b883901f0223" /> | <video src="https://github.com/artsy/force/assets/3934579/175c5369-6a73-4788-b6d9-6e95ba09dcd1" /> |

After:

| mWeb | Desktop |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/7e3326fe-4a88-458e-a132-48cb838aad72" /> | <video src="https://github.com/artsy/force/assets/3934579/e05fd6d1-2486-4d2c-af3c-0961e74107e8" /> |

### DIFF
--- a/src/Components/Notifications/NotificationsList.tsx
+++ b/src/Components/Notifications/NotificationsList.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Flex, Join, Separator, Text } from "@artsy/palette"
+import { Box, Button, Flex, Join, Separator, THEME, Text } from "@artsy/palette"
 import {
   createPaginationContainer,
   graphql,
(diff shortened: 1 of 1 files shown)


src/Components/Notifications/NotificationsList.tsx
```diff
@@ -22,6 +22,7 @@ import { NotificationsListPlaceholder } from "./NotificationsListPlaceholder"
 import { useNotificationsContext } from "Components/Notifications/useNotificationsContext"
 import { NotificationListMode } from "Components/Notifications/NotificationsTabs"
 import { useRouter } from "System/Router/useRouter"
+import { __internal__useMatchMedia } from "Utils/Hooks/useMatchMedia"
```
tam-kis, line 25: It's generally not advised to use [ __internal__useMatchMedia ](https://github.com/artsy/force/blob/cfd8aa9666d4dd3820b8fa962181b666c23ba1fc/src/Utils/Hooks/useMatchMedia.ts#L4-L6) , but in this case I don't think we can use [frensel](https://github.com/artsy/fresnel?tab=readme-ov-file#server-side-rendering-ssr-usage) so we can move on with this current approach

tam-kis (approved):

PR #13452, opened 2024-02-02
fix(consign): fixes inverted contrast; image placeholders

dzucconi:
Closes [DIA-438](https://artsyproduct.atlassian.net/browse/DIA-438)

OK so this _was_ dark mode related, but not in the way I thought.

These backgrounds were previously defined with: `background="black"` — `black` is of course a valid HTML color, not a themed value. I replaced it with `black100` but, surprisingly, `background` is not theme-aware. We typically use the prop `bg`, which is.

Additionally I fixed these images which were causing a content shift on page load. I do note that the assets we use are inadequately sized as well: we need much larger images to pull resizes from.

[DIA-438]: https://artsyproduct.atlassian.net/browse/DIA-438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/WaysWeSell.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/WaysWeSell.tsx
@@ -7,59 +7,57 @@ import {
   GridColumns,
   Column,
   FullBleed,
-  Flex,
+  Stack,
+  ResponsiveBox,
 } from "@artsy/palette"
 import { resized } from "Utils/resized"
 import { Media } from "Utils/Responsive"
 import { AppContainer } from "Apps/Components/AppContainer"
-import { HorizontalPadding } from "Apps/Components/HorizontalPadding"
-import { useWindowSize } from "Utils/Hooks/useWindowSize"
 
-export type StepsWithImageBlackDataType = {
-  src: string
-  srcSet: string
-  text?: string
-  title?: string
-  useBlackBackground?: boolean
-  imageHeight?: number
+interface ShelfItemProps {
+  image: {
+    src: string
+    width: number
+    height: number
+  }
+  text: string
+  title: string
 }
 
-const IMAGE_WIDTH = 600
-const BIG_IMAGE_HEIGHT = 575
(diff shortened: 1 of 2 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/WaysWeSell.tsx
```diff
@@ -68,113 +66,72 @@ const data: StepsWithImageBlackDataType[] = [
...
-  const { width } = useWindowSize()
-
-  const aspectRatio = BIG_IMAGE_HEIGHT / IMAGE_WIDTH
-
-  // calculating the width of the image contailer
-  // (screen width - HorizontalPadding - px of the parent - gridColumnGap) / 3
-  const collumnHeightMediumcreen =
-    ((width - 20 * 2 - 20 * 2 - 20 * 2) / 3) * aspectRatio
-  const collumnHeightBigScreen =
-    ((width - 40 * 2 - 40 * 2 - 40 * 2) / 3) * aspectRatio
```
dzucconi, line 111: Interesting in that, this was a lot of work to do something that is not so hard: https://www.notion.so/artsy/Handling-Images-On-The-Web-In-Force-5c6832bf4b3e431cb4830061ce057399

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/WaysWeSell.tsx
```diff
@@ -68,113 +66,72 @@ const data: StepsWithImageBlackDataType[] = [
...
+          height="100%"
+          lazyLoad
+          style={{ display: "block" }}
+        />
+      </ResponsiveBox>
+
+      <Stack
+        gap={[0.5, 1]}
+        // Magic numbers to bottom align images
+        minHeight={[100, 200, 140]}
```
dzucconi, line 110: This _should_ be structured so that we do this in 2 rows, eliminating the need for any manual adjustment. But this is fine for now.

PR #13458, opened 2024-02-03
feat(WIP): add blurhash to AuctionResults component

anandaroop:

### Description

Trying to flesh out use of blurhash on the artist page. More to come…


src/Apps/Artist/Routes/AuctionResults/ArtistAuctionResultItem.tsx
```diff
@@ -91,6 +92,7 @@ export const ArtistAuctionResultItem: React.FC<Props> = props => {
               <Image
                 src={image.src}
                 srcSet={image.srcSet}
+                placeHolderURL={blurhashDataURL ?? undefined}
```
anandaroop, line 95: At the moment, this does not seem to suffice.



I thought maybe the bg on L89 above, but removing that didn't help 🤔

mzikherman: Is this value accessible for auction results? I think with https://github.com/artsy/gravity/pull/17254 it's only available for `AdditionalImage`. Would be easy to set it more widely (like install shots), but if auction results are from diffusion - there might be something else to do to make it available.

anandaroop: Ahh, I did verify they were present in the Gemini's db, but yes, even that + the `GeminiImage` module over in Gravity  wouldn't necessarily be sufficient because of (a) the specific image model needing an update and (b) Diffusion lol 🤦🏽 

I'll close this for now.

mzikherman: Yea, I think adding the blurhash support to other artwork places (like artwork rails, `FlatGridItem`, etc.) is ready to go.

Adding blurhash support to other image types using `GeminiImage` in Gravity (like install shots), almost ready to go - needs small update to persist those in Gravity from the webhook.

For Diffusion....I think _only_ image url's are stored in Diffusion's image model within `Lot`,  _and_ there's no webhook setup either. So adding blurhash to auction results is probably the hardest/quirkiest one that could choose 😆

PR #13468, opened 2024-02-08
fix: wrong relay connection key used to refresh artwork lists

erikdstock:

This PR fixes a [bug reported in 🔒 slack](https://artsy.slack.com/archives/C02JHHHKP5K/p1707317247273669?thread_ts=1707317112.824079&cid=C02JHHHKP5K) where the artwork list list does not get updated after creating or deleting an artwork list. Looks like a tiny oversight after a migration on this page was completed.


MrSltun (approved):
Thank you Erik! 🌟

damassi:
Nice find ❇️

PR #13473, opened 2024-02-09
refactor(EMI-1691): rewrite Email PreferencesApp tests

TMcMeans:
The type of this PR is: **refactor**

@artsy/emerald-devs
This PR solves [EMI-1691]

### Description
The existing tests in the Email Preference Center lack coverage and fail to test user opt in/out of email actions. I removed some tests and updated the spec with new tests that correctly cover user behavior. 

[EMI-1691]: https://artsyproduct.atlassian.net/browse/EMI-1691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
+++ b/src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
@@ -48,22 +48,22 @@ describe("PreferencesApp", () => {
         submitMutation: submitMutationMock,
       })
 
-      const viewerMock = {
-        notificationPreferences: [
-          {
-            name: "recommendedByArtsy",
-            status: "SUBSCRIBED",
-          },
-          {
-            name: "artWorldInsights",
-            status: "SUBSCRIBED",
-          },
-          {
-            name: "productUpdates",
-            status: "UNSUBSCRIBED",
-          },
-        ],
-      }
+      // const viewerMock = {
+      //   notificationPreferences: [
+      //     {
+      //       name: "recommendedByArtsy",
+      //       status: "SUBSCRIBED",
+      //     },
+      //     {
+      //       name: "artWorldInsights",
+      //       status: "SUBSCRIBED",
+      //     },
+      //     {
+      //       name: "productUpdates",
+      //       status: "UNSUBSCRIBED",
+      //     },
+      //   ],
+      // }
 
       // render(<PreferencesApp viewer={viewerMock} />)
 
--- a/src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
+++ b/src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
@@ -65,7 +65,7 @@ describe("PreferencesApp", () => {
         ],
       }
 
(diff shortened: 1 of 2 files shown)


jonallured (dismissed):
Epic!! 🏆

src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
```diff
@@ -1,145 +1,261 @@
...
-    let checkboxes = screen.getAllByRole("checkbox")
-    let subscribeToAllCheckbox = checkboxes[0]
-    let unsubscribeFromAllCheckbox = checkboxes.pop()
-
-    fireEvent.click(subscribeToAllCheckbox)
-
-    expect(subscribeToAllCheckbox).toBeChecked()
+  it("saves preferences correctly on form submission", async () => {
+    const submitMutationMock = jest.fn(() => Promise.resolve())
+    ;(useEditNotificationPreferences as jest.Mock).mockReturnValue({
```
damassi, line 110: Can define this as a variable up above, below the first describe statement:
```tsx
const mockUseEditNotificationPreferences = useEditNotificationsPreferences as jest.Mock
```

src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
```diff
@@ -1,145 +1,261 @@
...
+        },
+      })
     })
+  })
 
-    fireEvent.click(subscribeToAllCheckbox)
+  describe("toasts", () => {
+    it("displays a success toast on successful form submission", async () => {
+      const submitMutationMock = jest.fn(() => Promise.resolve())
+      ;(useEditNotificationPreferences as jest.Mock).mockReturnValue({
```
damassi, line 142: ditto

src/Apps/Preferences/__tests__/PreferencesApp.jest.tsx
```diff
@@ -1,145 +1,261 @@
...
+      })
+
+      renderWithRelay({
+        Viewer: () => ({
+          notificationPreferences: [],
+        }),
+      })
+
+      fireEvent.click(screen.getByText("Save"))
+      const { sendToast } = require("@artsy/palette").useToasts()
```
damassi, line 153: Since we're already mocking this module, we can do something like this at the top of the file (in the describe block):

```
const mockUseToasts = useToasts as jest.Mock
```

And then in here, can mock the implementation:

```tsx
const spy = jest.fn()

mockUseToasts.mockReturnValue({
  sendToast: spy
})

...

expect(spy).toHaveBeenCalled(...)
```

And then we don't need the eslint disable rule at the top, as we'll be using that import.

TMcMeans:
> Looks great! You may need to run `yarn relay` and commit the generated file changes, as well as remove the unused imports to get the CI tests to pass.

ah okay so I should commit the generated graphql query file into this PR as well?

PR #13489, opened 2024-02-14
chore(alerts): clean up remaining v2 feature flags and dead code

anandaroop:

This PR solves [ONYX-711] and [ONYX-712] 

### Description

Cleans up the two remaining feature flags and dead code paths from the Alerts v2 epic.

More TK

[ONYX-711]: https://artsyproduct.atlassian.net/browse/ONYX-711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-712]: https://artsyproduct.atlassian.net/browse/ONYX-712?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Alert/Hooks/__tests__/AlertProvider.jest.tsx
```diff
@@ -95,11 +94,11 @@ describe("AlertProvider", () => {
 
     await flushPromiseQueue()
 
-    // expect(screen.getByText("Add Filters:")).toBeInTheDocument()
     expect(screen.getByText("Andy Warhol")).toBeInTheDocument()
+    expect(screen.getByText("Add Filters")).toBeInTheDocument()
 
     // transition to filters step
-    // screen.getByTestId("addFilters").click()
+    screen.getByTestId("moreFilters").click()
```
anandaroop, line 101: At this point I'd like to test the **More Filters ⟩** element, but I am struggling with the correct usage of `mockResolveOperation`

|Skeleton|Rendered|
|---|---|
|skeleton|rendered|

anandaroop: Fixed in [a8a9d21](https://github.com/artsy/force/pull/13489/commits/a8a9d21062647944441a86b5a99dbca73b123f3e) h/t @gkartalis

src/Components/Alert/Hooks/__tests__/AlertProvider.jest.tsx
```diff
@@ -88,26 +79,39 @@ describe("AlertProvider", () => {
...
     }
 
+    const mockedSuggestionResolver = {
+      PreviewSavedSearch: () => ({
+        suggestedFilters: [
+          { displayValue: "Photography" },
+          { displayValue: "Print" },
+        ],
+      }),
+    }
```
anandaroop, line 96: George helped me solve this by correcting the nesting here to mock [the operation in question](https://github.com/artsy/force/blob/bc9b7d2e1c3304bf4abfad34522fea0f7185e36d/src/Components/Alert/Components/Form/SuggestedFilters.tsx#L121-L134)

The above is correct, whereas the following was not:
```tsx
const mockedSuggestionResolver = {
      Viewer: () => ({.  // ❌
        previewSavedSearch: {
          suggestedFilters: [
            { displayValue: "Photography" },
            { displayValue: "Print" },
          ],
        }
      }),
    }
```

cc @damassi

src/Components/Alert/Components/Filters/__tests__/NotificationPreferences.jest.tsx
```diff
@@ -1,63 +1,85 @@
...
+  //             details: "",
+  //           }}
+  //           onSubmit={jest.fn()}
+  //         >
+  //           <NotificationPreferencesQueryRenderer mode={mode} />
+  //         </Formik>
+  //       </AlertProvider>
+  //     </MockBoot>
+  //   )
+  // }
```
anandaroop, line 55: We can delete this now right @MounirDhahri ?

src/DevTools/setupTestWrapper.tsx
```diff
@@ -126,9 +126,11 @@ export const setupTestWrapperTL = <T extends OperationType>({
...
         </MockBoot>
       ) : (
-        <RelayEnvironmentProvider environment={env}>
-          <Component {...componentProps} />
-        </RelayEnvironmentProvider>
+        <MockBoot relayEnvironment={env}>
+          <RelayEnvironmentProvider environment={env}>
+            <Component {...componentProps} />
+          </RelayEnvironmentProvider>
+        </MockBoot>
```
anandaroop, line 133: I get that `MockBoot` was added so that all standard wrappers are present (esp the Palette theme which was specifically affecting these tests)

I'm still hazy on why _both_ `MockBoot` and `RelayEnvironmentProvider` are needed. (The env passed into `MockBoot` eventually makes its way to `Boot` → `SystemContext` but that is not sufficient I take it?)

anandaroop (commented):
Awesome, I appreciate you digging in here @MounirDhahri  🙏🏽

src/DevTools/setupTestWrapper.tsx
MounirDhahri: Hi Roop, made a video on why I believe this makes sense here
https://www.loom.com/share/c19184b6446e4970826487e8413a8aa7?sid=64deccb3-9cb3-4ca0-a0f4-95eb05745fea

anandaroop: Excellent, thank you

damassi: This was really helpful in explaining things, thanks @MounirDhahri. 

Makes sense in why this hasn't come up before, due to the fact that we're just now starting to adopt relay hooks in Force. In all of our other (more modern) apps, this provider is there, and one of the first things setup when we lay out the framework of the app. 

Couple things:
- Can we move this `RelayEnvironmentProvider` into MockBoot? I suspect this will come up in the future, and having it available to all of our code would be good
- Can we also add this to `Boot`?

MounirDhahri: we actually spoke about this but we were unable to evaluate properly its impact. I think migrating this to the Mockboot would be the idea solution. 
Adding this to Boot as well sounds good as long as we make sure we use the same environment in our query renderers, I can PR that.

damassi: I think it'll be safe 👍  We pluck the env from the system context everywhere for query renderers, and if we share the same env with the relay provider it (should) be backwards compatible.

anandaroop: I'm going to rebase this on `main` and merge soon. I propose that we move any further test infra changes to a separate PR 🙏🏽

damassi: Roger that 👍

anandaroop:
Alright going for the merge

PR #13494, opened 2024-02-15
fix: Clean up notification pills query

olerichter00:
## Description

Clean up notification pills query because we always render "Alerts" and "Follows" filters


MounirDhahri (approved):
Bye

PR #13495, opened 2024-02-15
fix: Mark notification as read when opening notification page

olerichter00:
## Description

At the moment, we don't mark a notification as read when the user lands on the notification page directly (only when the user clicks the notification in the list). This PR fixes this by marking a notification as read when the notification page is opened. For notifications that we don't show a dedicated page for, the old behavior remains, and the notification is marked as read when the user clicks on the item in the list.


src/Components/Notifications/NotificationItem.tsx
```diff
@@ -43,6 +43,14 @@ const NotificationItem: FC<NotificationItemProps> = ({ item }) => {
     remainingArtworksCount > 0
 
   const markAsRead = async () => {
+    // If the notification is opened as a page, we don't need to mark the notification as read
```
nickskalkin, line 46: 👍

nickskalkin (approved):
👍

PR #13501, opened 2024-02-16
feat: orchestrate containers with hokuasi dev start

mc-jones:
The type of this PR is: Feat

This PR solves [PHIRE-584]

### Description

This PR modifies our existing `development.yml` and adds necessary configurations to build, connect to our core apis, and supports OAuth with other containerized services. 

### Follow-ups
- PRs will be made to downstream services that use OAuth for these changes to work. 

### Considerations
- The scope of this PR was to get `force` to (i) build, (ii) boot and successfully connect to other core services; and (iii) support `OAuth` for artsy clients. However, we still need to resolve a lot of data dependencies before a developer can build a feature in containerized `force`. We will be ticketing out that work.


.env.development
```diff
@@ -0,0 +1,7 @@
+API_URL=http://host.docker.internal:3000
```
mc-jones, line 1: Any service that needs to be reachable server and client side will need to use host.docker.internal:<PORT> in their env. This also requires developers to add 127.0.0.1 host.docker.internal to their /etc/hosts/ file.

Dockerfile.dev
```diff
@@ -0,0 +1,30 @@
...
+    cp lib/* /usr/local/lib && \
+    chmod 755 /usr/local/bin/watchman && \
+    chmod 2777 /usr/local/var/run/watchman && \
+    cd .. && \
+    rm -rf watchman.zip watchman-${WATCHMAN_VERSION}-linux
+
+COPY package.json yarn.lock /app/
+
+RUN --mount=type=cache,id=yarndev,target=/usr/local/share/.cache \
+    yarn install --frozen-lockfile --quiet
```
mc-jones, line 30: I don't copy the project files into the container since we [bind mount](https://docs.docker.com/storage/bind-mounts/) the project directory.

Dockerfile.dev
```diff
@@ -0,0 +1,30 @@
...
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+  bash \
+  build-essential \
+  git \
+  unzip \
+  wget 
+
+ENV WATCHMAN_VERSION=v2024.02.05.00
```
mc-jones, line 12: We may want to set the version dynamically and get the most recent release as watchman is released every week, now.

hokusai/development.yml
```diff
@@ -2,28 +2,21 @@
...
+      - 5100:5100
     volumes:
       - ../:/app
       # prevents bind mount from overwriting node_modules in docker image.
       - /app/node_modules
     depends_on:
       - force-redis
   force-redis:
-    image: redis:3.2-alpine
+    image: redis:5.0.6-alpine
```
mc-jones, line 20: match production.

hokusai/templates/docker-compose-service-dev.yml.j2
```diff
@@ -0,0 +1,4 @@
+    image: "hokusai_force:${BUILD_TAG:-latest}"
+    build:
+      context: ../
+      dockerfile: Dockerfile.dev
```
mc-jones, line 4: I've opted to create a new dockerfile specifically for dev rather than add more stages to the existing Dockerfile, which is already complex. I believe this to be a simpler and saner solution, but I'm happy to hear other opinions.

ovasdi (approved):
Nice job! 👏 

Spun up Force, MP and Gravity all containerized and was like 🤩 
You got the watchman dep sorted out and even `/collect` page loads!

PR #13503, opened 2024-02-19
chore: add relay environment to MockBoot

MounirDhahri:

This PR makes the following change:
- Add relay environment to Mockboot to make sure components using relay hooks, can find a valid relay environment.

**Bonus:** The error stack trace was inexistent and that made finding out why [some tests](https://github.com/artsy/force/pull/13489/files) broke not trivial. I made a small adjustment to our error boundary to log the full error stack trace when running tests. 

_[The video below shows what it is like after the changes and how it was before them]_

https://github.com/artsy/force/assets/11945712/d7b59a2e-f099-4a74-b55e-8ef29d03f4a6

This PR solves [PROJECT-XX]

### Description

### DIFF
--- a/src/System/Router/Boot.tsx
+++ b/src/System/Router/Boot.tsx
@@ -4,12 +4,15 @@ import {
   THEME,
   ToastsProvider,
 } from "@artsy/palette"
-import { SystemContextProvider } from "System/SystemContext"
+import {
+  SystemContextConsumer,
+  SystemContextProvider,
+} from "System/SystemContext"
 import { AppRouteConfig } from "System/Router/Route"
 import { useEffect, useState } from "react"
 import * as React from "react"
 import { HeadProvider } from "react-head"
-import { Environment } from "react-relay"
+import { Environment, RelayEnvironmentProvider } from "react-relay"
 // eslint-disable-next-line no-restricted-imports
 import { data as sd } from "sharify"
 // eslint-disable-next-line no-restricted-imports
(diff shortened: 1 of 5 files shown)


src/System/Router/Boot.tsx
MounirDhahri, line 1: I suggest hiding white spaces for easier reviewing

src/System/Router/Boot.tsx
```diff
@@ -102,34 +105,46 @@ export const Boot = track(undefined, {
...
-                          >
-                            <CookieConsentManager>
-                              <FocusVisible />
-                              <SiftContainer />
+            <SystemContextConsumer>
+              {contextValues => {
+                return (
+                  <RelayEnvironmentProvider
+                    environment={contextValues.relayEnvironment}
+                  >
```
MounirDhahri, line 113: _This needs some explanation._ 
Initially, I wanted to just inject `RelayEnvironmentProvider` within the `SystemContextProvider`. like this
```
<SystemContextProvider {...contextProps}>
    <RelayEnvironmentProvider environment={contextValues.relayEnvironment}>
    ...
    <RelayEnvironmentProvider />
<SystemContextProvider />
```

However this meant that `contextValues.relayEnvironment` is already defined although it can be `undefined`. The latter gets assigned inside `SystemContextProvider` and is only available when it's ready! This is why I needed to use the `Consumer` here. If you have a more elegant approach, please let me know

damassi: This works 👍

src/System/Router/ErrorBoundary.tsx
```diff
@@ -41,6 +41,11 @@ export class ErrorBoundary extends React.Component<Props, State> {
   }
 
   static getDerivedStateFromError(error: Error | HttpError) {
+    // This adds a useful stacktrace that can make debugging easier when writing tests
+    if (window.__TEST__) {
```
damassi, line 45: I (think) you can use jest's provided env variable! so something like `process.env.JEST`. Its there, but not entirely sure how to get at it. Maybe even `getENV('JEST')` could work?

damassi: Also we might need to check for SSR here, if needing to use window -- something like `typeof window !== undefined && window.TEST`, if the JEST approach doesn't work.

PR #13485, opened 2024-02-13
feat(ONYX-695): add offers filter

tam-kis:
The type of this PR is: Feat

This PR solves [ONYX-695](https://artsyproduct.atlassian.net/browse/ONYX-695)

### Description
This pull request does the following:

-  Adds a new "Offers" filter pill to the Activity panel
-  Implements logic to conditionally render the Activity panel filter pills based on whether the user has activities of this type.
-  Adjusts the minimum width of the Activity panel to display all filter pills in a single row.
-  Refactors the notification/activity list to address a pagination bug using Relay hooks (instead of QueryRenderer).
-  Uses a separate feature flag, [onyx_activity_panel_filter_pills](https://unleash.artsy.net/projects/default/features/onyx_new_notification_page), to enable independent release of the Activity pills (Staging: enabled; Production: disabled)

We are currently facing difficulties with the tests, so I have temporarily commented them out to ensure timely merging of this pull request for the "Partner Offers" release.

Screenshot 2024-02-14 at 11 00 11

Screenshot 2024-02-14 at 10 59 51

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-695]: https://artsyproduct.atlassian.net/browse/ONYX-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13504, opened 2024-02-19
fix: fixes broken pagination for activity panel

nickskalkin:

### Description

Pagination doesn't work - for some reason Force always sends the same `cursor`. I can't figure why this happens, but it was probably introduced in [this PR](https://github.com/artsy/force/pull/13485), when switching to `usePaginationFragment` hook.


olerichter00 (approved):
Thanks a lot for the fix 🙏

PR #13508, opened 2024-02-20
fix: wandering artist follows popover (DIA-471)

iskounen:
This PR changes:

1. The horizontal position of the artist follows popover when the artist has no:
    - Cover artwork
    - Biography blurb
    - Verified representatives
    - Insights
1. The placement of the arrow on the artist follows popover

(1) was needed because the popover was appearing to the far right of the follows button:

Screenshot 2024-02-20 at 9 00 55 AM

The solution was to fit the container that the popover is attached to to fit its content. However, (2) was needed because the popover was smooshed against the edge of the screen when there was no cover artwork:

Screenshot 2024-02-20 at 9 00 16 AM

Here is the final result for artists with/without a cover artwork:

Screenshot 2024-02-20 at 8 59 13 AM

Screenshot 2024-02-20 at 8 58 12 AM


dzucconi (approved):
Thanks!

egdbear (approved):
🚀

PR #13512, opened 2024-02-21
fix: Adjust notification page margin

olerichter00:
Addresses [ONYX-776]

## Description

Adjust notification page margin.

### DIFF
--- a/src/Components/Notifications/Notification.tsx
+++ b/src/Components/Notifications/Notification.tsx
@@ -1,4 +1,5 @@
 import {
+  Box,
   Flex,
   Skeleton,
   SkeletonBox,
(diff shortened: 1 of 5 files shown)


src/Components/Notifications/Notification.tsx
```diff
@@ -127,9 +128,11 @@ export const NotificationQueryRenderer: React.FC = props => {
   }
 
   return (
-    <Suspense fallback={<Placeholder />}>
-      <Notification notificationId={state.currentNotificationId} {...props} />
-    </Suspense>
+    <Box mx={[2, 4]} my={2}>
```
olerichter00, line 131: Moving the outer header to the `Notifications` component.

MounirDhahri (approved):
Makes sense. Thanks

PR #13531, opened 2024-02-26
feat(ONYX-727): top alert should be active when user visits settings alerts page

dariakoko:

This PR solves [ONYX-727]

### Description

Always show the Edit form of the last alert in the list (web only)

https://github.com/artsy/force/assets/36167539/1f4596ff-87ab-4fc4-a3b4-1b55c87d687a

[ONYX-727]: https://artsyproduct.atlassian.net/browse/ONYX-727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -48,19 +50,35 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
...
+    if (isMobile === null) return
+    if (!isMobile) {
+      setEditAlertEntity({
+        id: alerts[0].internalID,
+        name: alerts[0].settings.name ?? "",
+        artistIds: alerts[0]?.artistIDs as string[],
+      })
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isMobile])
```
MounirDhahri, line 81: I feel this logic is quite complicated and it could use some testing. the reason behind it is that it adds some side effect that is not trivial and that will make debugging hard in case of an error. What I suggest to move on with this, is speaking to Barney about a robust design solution here. Something like:

Screenshot 2024-02-27 at 17 04 05

MounirDhahri: @damassi I wonder how would you approach this? do you think this is worth the risk to build?

MounirDhahri (commented):
Thanks Daria for making this!

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -48,19 +50,35 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
...
-  ] = useState<EditAlertEntity | null>(null)
   const { relayEnvironment } = useSystemContext()
   const { sendToast } = useToasts()
   const { trackEvent } = useTracking()
+  const { silentPush } = useRouter()
   const [showDeleteModal, setShowDeleteModal] = useState(false)
   const [sort, setSort] = useState("ENABLED_AT_DESC")
   const [loading, setLoading] = useState(false)
   const alerts = extractNodes(me.alertsConnection)
+  const isMobile = __internal__useMatchMedia(THEME.mediaQueries.xs)
```
damassi, line 61: Unfortunately we can't use this, and that's why its marked as `__internal__`. Not SSR safe. Of course, if the code-path never hits ssr its fine, but that can't be promised in the future.

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
damassi: I couldn't say, unfortunately, just too far away from the feature. 

If we do want to go ahead with the current implementation, I think updating the code to something like:

```tsx
const isMobile = getENV("IS_MOBILE") 

useEffect(() => {
  if (isMobile) {
    // do thing
  }
}, [isMobile])
```

Is fine, but this won't execute responsively, and that's ok -- a user can't drag around their browser window. 

So yah, that's my take, just from a code perspective!

MounirDhahri: thanks, sorry for putting you on the spot 😸

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
MounirDhahri: thanks!

dariakoko: Thank you 🙏

src/Components/Notifications/NotificationsList.tsx
```diff
@@ -53,9 +53,7 @@ export const NotificationsList: React.FC<NotificationsListProps> = ({
 
   const { state } = useNotificationsContext()
 
-  const xs = __internal__useMatchMedia(THEME.mediaQueries.xs)
-  const sm = __internal__useMatchMedia(THEME.mediaQueries.sm)
-  const isMobile = xs || sm
+  const isMobile = getENV("IS_MOBILE")
```
dariakoko, line 57: based on @damassi [comment here](https://github.com/artsy/force/pull/13531#discussion_r1504536546) 
@nickskalkin could you, please, confirm this change is correct and we don't have isMobile for internal use only?

MounirDhahri (approved):
Awesome, thanks @dariakoko and sorry for the late reply 🙏

PR #13533, opened 2024-02-27
fix: Remove spacing below notification list item title

olerichter00:
## Description

Remove spacing below notification list item title.


MounirDhahri (approved):
Sorry missed this as well, I see it now thanks to Jasper

PR #13542, opened 2024-02-28
fix: Avoid scroll offset when switching notifications

olerichter00:
## Description

Avoid scroll offset when switching notifications.


nickskalkin (approved):
smart :)

PR #13532, opened 2024-02-27
fix: Rename Follows notification filter pill

olerichter00:
https://www.notion.so/artsy/Inconsistency-in-filter-pills-between-web-Following-and-app-Follows-5161c13888284a70b7b020a008de7ab5?pvs=4

## Description

Rename Follows notification filter pill.

PR #13546, opened 2024-02-28
chore: Rename "following" notification type to "follows"

olerichter00:
Context: https://github.com/artsy/force/pull/13532#discussion_r1506307872

## Description

Rename "following" notification type to "follows".


dariakoko (approved):
Thank youuuu

PR #13558, opened 2024-03-01
fix(nav-menu): Close artist and artworks dropdowns on nav item click

damassi:

### Description

Fixes an long-standing issue where top-nav clicks don't close the drop down, leading to some page confusion. Discussed with Jimi [in here](https://artsy.slack.com/archives/C9XJKPY9W/p1709241670001359). 

https://github.com/artsy/force/assets/236943/cdc8d5f4-aee4-432c-9ce1-65165f92fb47

### DIFF
--- a/src/Components/NavBar/NavBar.tsx
+++ b/src/Components/NavBar/NavBar.tsx
@@ -372,7 +372,7 @@ export const NavBar: React.FC = track(
                     />
                   )}
                 >
-                  {({ anchorRef, anchorProps, visible }) => (
+                  {({ anchorRef, anchorProps, visible, setVisible }) => (
                     <NavBarItemButton
                       ref={anchorRef as any}
                       px={0}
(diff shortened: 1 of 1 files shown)


src/Components/NavBar/NavBar.tsx
```diff
@@ -382,7 +382,14 @@ export const NavBar: React.FC = track(
...
                       <NavBarItemUnfocusableAnchor
                         href="/artists"
-                        onClick={handleClick}
+                        onClick={event => {
+                          handleClick(event)
+
+                          // Small timeout to avoid transition race condition
+                          setTimeout(() => {
+                            setVisible(false)
+                          }, 100)
```
damassi, line 391: I couldn't make this work without a delay -- thoughts @dzucconi?

dzucconi (approved):
👍

src/Components/NavBar/NavBar.tsx
dzucconi: Yeah, I mean we _could_ transition it using the web animations API and get a real callback. This is also fine. It's not exact in this case.

PR #13562, opened 2024-03-01
feat(private-artworks): Apply static comps design

damassi:

### Description

PR 2-3. This applies the design for the page statically. Next PR will actually query correct data and wire up completely. 

cc @artsy/amber-devs


src/Apps/Artwork/Components/PrivateArtwork/PrivateArtworkMetadata.tsx
```diff
@@ -0,0 +1,107 @@
...
+  artwork,
+}) => {
+  const data = useFragment(
+    graphql`
+      fragment PrivateArtworkMetadata_artwork on Artwork {
+        title
+      }
+    `,
+    artwork
+  )
```
JanaeHijaz, line 27: when you say that this page is static, you're saying that you're just setting up the bare bones layout, but this fragment will be updated to include the actual data that we need for each of these components, correct @damassi ?

damassi: Not a strict rule of course, but this really helps speed things along since one can focus on exactly the stage one is on, without mixing all the concerns all together. Instead of architecture, data, design / css, etc, each pr is one thing and one thing only.

damassi: Not a strict rule of course, but this really helps speed things along since one can focus on exactly the stage one is on, without mixing all the concerns all together. Instead of architecture, data, design / css, etc, each pr is one thing and one thing only.

JanaeHijaz (commented):
LGTM

PR #13565, opened 2024-03-04
fix: Update Alerts screen page to use split view

olerichter00:
Addresses [ONYX-823]

## Description

This PR updates the Alerts screen page layout to use a split view with independent scrollable views. We already use this layout for https://www.artsy.net/notifications and https://www.artsy.net/user/conversation. 

Using the split view instead of a sticky right panel resolves a couple of scrolling issues that we experienced (for example [ONYX-823] & https://github.com/artsy/force/pull/13599).

## Open Questions

- The split screen isn't usable with a browser window hight below 500px. Should we fix it?
- Should we make the whole row in the list on the left clickable?
- Should we adjust the ratio between list and detail view to match the 1:3 ratio from `/notifications`?

https://github.com/artsy/force/assets/4691889/ecde8919-2574-4484-bc5f-857b6bbaeb94

[ONYX-823]: https://artsyproduct.atlassian.net/browse/ONYX-823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
@@ -137,7 +137,6 @@ const SavedSearchAlertEditForm: React.FC<SavedSearchAlertEditFormProps> = ({
 
         const transitionToFilters = () => {
           goToFilters()
-          jumpTo("SavedSearchAlertEditForm")
         }
 
         return (
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
@@ -30,7 +30,6 @@ import { Modal } from "Components/Alert/Components/Modal/Modal"
 import { NotificationPreferencesQueryRenderer } from "Components/Alert/Components/NotificationPreferences"
 import { SugggestedFiltersQueryRenderer } from "Components/Alert/Components/Form/SuggestedFilters"
 import { Sticky } from "Components/Sticky"
-import { useJump } from "Utils/Hooks/useJump"
 
 interface SavedSearchAlertEditFormQueryRendererProps {
   editAlertEntity: EditAlertEntity
@@ -113,7 +112,6 @@ const SavedSearchAlertEditForm: React.FC<SavedSearchAlertEditFormProps> = ({
   onCompleted,
 }) => {
   const { state, goToFilters, dispatch, onComplete } = useAlertContext()
-  const { jumpTo } = useJump()
 
   const isMounted = useDidMount()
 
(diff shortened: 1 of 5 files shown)


src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
```diff
@@ -137,7 +135,6 @@ const SavedSearchAlertEditForm: React.FC<SavedSearchAlertEditFormProps> = ({
 
         const transitionToFilters = () => {
           goToFilters()
-          jumpTo("SavedSearchAlertEditForm")
```
olerichter00, line 140: The `jumpTo` isn't needed anymore because both containers scroll independently. I also reduced the height of the right panel's loading skeleton. This change causes the loading screen to force the container to scroll to the top.

src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -285,13 +289,15 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
       </Join>
 
       {relay.hasMore() && (
-        <Box textAlign="center" mt={4}>
-          <Button onClick={handleLoadMore} loading={loading}>
-            Show More
-          </Button>
-        </Box>
+        <>
+          <InfiniteScrollSentinel onNext={handleLoadMore} once={false} />
```
olerichter00, line 293: Replacing the "Show More" button with an infinite scroll 🥳

nickskalkin (dismissed):
looks awesome @olerichter00!

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -288,22 +284,25 @@ export const SavedSearchAlertsArtworksQueryRenderer: React.FC<SavedSearchAlertsA
...
-        </SkeletonText>
-        <SkeletonText variant="sm-display">
-          See our top picks for you:
-        </SkeletonText>
-        <Spacer y={[4, 2]} />
-      </Flex>
-      <ArtworkGridPlaceholder columnCount={2} />
-    </Join>
+    // Setting a max height to force scrolling to top when the content changes.
+    <>
```
damassi, line 288: Not necessary or blocking (maybe quiet and no animation is better) - but if you wrap this in the `<Skeleton>` it will [pulse as a loading](https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Skeleton/Skeleton.tsx#L66-L77) state

olerichter00: Thanks 🙏  This component is used and wrapped with `<Skeleton>` [here](https://github.com/artsy/force/pull/13565/files#diff-9acbdd2cc4ae3f70215edaa54eb40dbff5ba88c57080b2cd6e37fda3b21e1666R314). Looks like `SavedSearchAlertsArtworksPlaseholderContext` already 👍

dariakoko (approved):
Looks great, thank you!

I have just one minor thing, possibly for Barney.
I was experimenting with the same thing yesterday, and to me it seemed that the possibility to scroll the left side could be more obvious. The position of the elements seems very static, especially when the user has a lot of alerts, there is no indication that the component is scrollable (shadow on top or bottom of the list, for example). On the notifications page it feels different, probably because the left element is way smaller.

olerichter00:
@dariakoko , that's a very good observation! I will ask Barney what he thinks about making the list on the left more narrow. This way, it would look more like a list and match the design of `/notifications` and `/conversations`.

PR #13567, opened 2024-03-04
fix: avoids remounting

dzucconi:
Closes [DIA-456](https://artsyproduct.atlassian.net/browse/DIA-456)

Re: https://github.com/artsy/force/pull/13214 — this bug needs to be fixed rather than papered over by constantly remounting the filter whenever the route changes.

It's actually pretty concerning that one can't drive the filters with a simple route change. One would think that would be the primary way we'd change filter state...

[DIA-456]: https://artsyproduct.atlassian.net/browse/DIA-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
+++ b/src/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
@@ -67,8 +67,6 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
       ]}
       ZeroState={ZeroState}
       userPreferredMetric={userPreferences?.metric}
-      // Setting the key here to enforce a remount of the component when changing pages or query parameter filters within the query.
-      key={match.location.pathname + match.location.search}
     >
       <ArtworkFilterAlertContextProvider
         initialCriteria={{ artistIDs: [artist.internalID] }}


olerichter00 (commented):
Do you have an idea on how to update the filter state upon a simple route change? If I recall correctly, the issue is that the ArtworkFilterContextProvider [uses the props to initialize the filter](https://github.com/artsy/force/blob/10c84ac9d55e0ac2a7e4aaba7a858b96669284bb/src/Components/ArtworkFilter/ArtworkFilterContext.tsx#L320) store but then does not update the store when the props change.

dzucconi:
@olerichter00 I've seriously considered just rewriting the filters component to avoid having to fix anything in it 😆 — I'm not sure. If we can just update the internal state whenever the route changes, using an effect, that'd be great though there might be some complications with a recursive effect loop.

src/Apps/Artist/Routes/WorksForSale/Components/ArtistArtworkFilter.tsx
```diff
@@ -67,8 +67,6 @@ const ArtistArtworkFilter: React.FC<ArtistArtworkFilterProps> = props => {
       ]}
       ZeroState={ZeroState}
       userPreferredMetric={userPreferences?.metric}
-      // Setting the key here to enforce a remount of the component when changing pages or query parameter filters within the query.
-      key={match.location.pathname + match.location.search}
```
olerichter00, line 71: I'm not sure if I understand correctly why this causes the flashing images. Is it because forcing the `ArtworkFilterContextProvider` with the `key` prop also remounts it's children?

dzucconi: Yeah, key will forcibly re-render the component and anything under it. That means images will be re-loaded.

olerichter00: Thanks, then adding a key here is indeed not a good idea.

damassi: Ah @dzucconi - this was the same issue that fixed things before, I (think). 

Re URL state -- yep, driving by URL state, we've been talking about this for a while. It seems [straight forward](https://github.com/artsy/volt-v2/blob/main/src/pages/demand/hooks/useDemandFilters.tsx) 🤷

PR #13570, opened 2024-03-05
fix(AMBER-504): Fix Salesforce Chat Bubble for auctions

lidimayra:
[AMBER-504]

### Description
The current implementation is not working on the auction page because it needs different handling for the price thresold.

The chat bubble for artworks is based on the listed price. This works for BNMO artworks, but it doesn't work for Auction artworks. For Auction artworks we want to consider the starting bid value instead of listed price.

## Before
Screenshot 2024-03-05 at 10 13 22

## After
Screenshot 2024-03-05 at 10 12 59
Screenshot 2024-03-05 at 10 13 15

cc @artsy/amber-devs 

[AMBER-504]: https://artsyproduct.atlassian.net/browse/AMBER-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkChatBubble.tsx
+++ b/src/Apps/Artwork/Components/ArtworkChatBubble.tsx
@@ -17,6 +17,7 @@ const ArtworkChatBubble: FC<ArtworkChatBubbleProps> = ({ artwork }) => {
     isOfferable,
     listPrice,
     isInAuction,
+    saleArtwork,
   } = artwork
 
   const price = useMemo(() => {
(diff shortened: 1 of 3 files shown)


Mitch-Henson (dismissed):
From a functionality point of view it looks great!

Curious if we need to do anything from a design point of view so that it doesn't block functionality? In the screenshot, it overlaps the create alert button. Not sure if that was just a poorly timed screenshot or an actual problem we should look at in this ticket or another.

lidimayra:
ohhh that's a very very good call!! Thank you for raising! 🙌 

I just checked now with a different artwork and the same thing also happens:
Screenshot 2024-03-05 at 11 58 34

It doesn't happen for BNMO artworks, just for auctions. Probably because of the different sidebar layout.
I believe this is a blocker. I'll bring this back to the discussion with Allie!

damassi (dismissed):

PR #13599, opened 2024-03-13
Feat(ONYX-731): user can click view artworks link and see matches

dariakoko:

This PR solves [ONYX-731]

### Description

| web | mWeb |
| --- | --- |
| <video src="https://github.com/artsy/force/assets/36167539/9fadb394-c95e-42ce-ad3b-78bda429335d"/> | <video src="https://github.com/artsy/force/assets/36167539/08169479-fd38-4af8-b68d-2cb1eafa8b83" /> |

[ONYX-731]: https://artsyproduct.atlassian.net/browse/ONYX-731?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditFormPlaceholder.tsx
```diff
@@ -1,20 +1,49 @@
...
+        <AlertProvider isEditMode>
+          <Skeleton flex={1} p={4}>
+            <Modal onClose={onCloseClick} backgroundColor={"white100"}>
+              <Skeleton px={2}>
+                <SavedSearchAlertEditFormPlaceholderContext />
+              </Skeleton>
+            </Modal>
+          </Skeleton>
+        </AlertProvider>
+      </Media>
```
dariakoko, line 34: Placeholder for the Edit Alert view was broken, fixing it in this PR by adding `Modal` together with `AlertProvider` to the Placeholder

dariakoko:
There is one issue with this PR, I would appreciate your help 🙏 

When I press `Edit Alert` of `View Artworks` the url changes to `settings/alerts/alertID/edit` or `settings/alerts/alertID/artworks`. And when I press Alerts from the dropdown menu or the tab, I would expect 

1. to be redirected go to the `settings/alerts` route - it happens 🟢 
2. to be shown the edit form of the first alert on the list - can’t make it happen, the url stays the same (`settings/alerts`) and the page doesn't re-render 🔴 

I think we need to add one more useEffect or adjust the existing useEffects to include `match.location.pathname`. I can't figure out how to make it happen without breaking the functionality on mWeb when we don't need to open the edit form when the component is mounted

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
```diff
@@ -293,16 +293,20 @@ export const SavedSearchAlertEditFormQueryRenderer: React.FC<SavedSearchAlertEdi
...
       cacheConfig={{ force: true }}
       render={({ props, error }) => {
         if (error) {
           console.error(error)
           return null
         }
 
-        if (!props?.me?.alert || !props?.viewer) {
-          return <SavedSearchAlertEditFormPlaceholder />
+        if (!props || !props?.me?.alert || !props?.viewer) {
```
olerichter00, line 306: This should be equivalent: 
```suggestion
        if (!props?.me?.alert || !props?.viewer) {
```

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
```diff
@@ -115,7 +124,7 @@ export const SavedSearchAlertListItemFragmentContainer = createFragmentContainer
         href
         title: displayName(only: [artistIDs])
         subtitle: displayName(except: [artistIDs])
-        artworksConnection(first: 1) {
+        artworksConnection(first: 10) {
```
olerichter00, line 127: 👀

nickskalkin:
@dariakoko I've added `match.params` to the `useEffect` in the list component and it seems like it does the trick - could you confirm? Sorry for pushing to your branch!

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
nickskalkin, line 306: **question / non-blocking**: Is there a need for `!props` check? I see that we use only `props.me.alert` and `props.viewer` below

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -0,0 +1,317 @@
...
+                </Text>
+              </Text>
+            )}
+
+            <Spacer y={2} />
+
+            <ArtworkGrid
+              artworks={alert.artworksConnection}
+              columnCount={2}
+              emptyStateComponent={<ArtworksGridEmptyState />}
```
olerichter00, line 81: (non-blocking): What do you think about refactoring this to `count > 0 ? ... text and artwork grid ... : <ArtworksGridEmptyState />`? Like this, we would reduce the number of `count > 0` within the code and would not implicitly return <ArtworksGridEmptyState /> within `ArtworkGrid`.

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+      <Media lessThan="md">
+        <Modal onClose={onCloseClick}>
+          <Flex mx={2} mb={2} flexDirection="column">
+            <Join separator={<Spacer y={2} />}>
+              <Flex flexWrap="wrap" gap={1}>
+                <CriteriaPills editable={false} />
+              </Flex>
+
+              {count > 0 && (
```
olerichter00, line 127: Same here.

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditFormPlaceholder.tsx
```diff
@@ -1,20 +1,49 @@
...
-        Edit Alert
-      </Text>
+    <>
+      <Media greaterThanOrEqual="md">
+        <Skeleton flex={1} p={4}>
+          <SavedSearchAlertEditFormPlaceholderContext />
+        </Skeleton>
+      </Media>
+      <Media lessThan="md">
+        <AlertProvider isEditMode>
```
nickskalkin, line 25: **question**: is there a need for AlertProvider in the placeholder component?

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -0,0 +1,317 @@
...
+          me {
+            alert(id: $id) {
+              artworksConnection(first: 10) {
+                counts {
+                  total
+                }
+                ...ArtworkGrid_artworks
+              }
+              href
+              ...SavedSearchAlertsArtworks_alert @relay(mask: false)
```
olerichter00, line 228: Is `@relay(mask: false)` here on purpose? The documentation says it's [not recommended](https://relay.dev/docs/api-reference/graphql-and-directives/#relaymask-boolean) to use `mask: false`, so it might be good to avoid it if possible.

olerichter00 (approved):
Great work! 🌟 

I have noticed that clicking on "Edit Alert" below the artwork grid could benefit from a "scroll to top" feature. Currently, when clicking the button, the edit alert form appears correctly, but the page's scroll position is near the footer, making the form not visible.

Screenshot 2024-03-21 at 10 39 29

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -0,0 +1,317 @@
...
+      <Media greaterThanOrEqual="md">
+        <Skeleton p={4}>
+          <Text variant="lg" mb={2}>
+            View Artworks
+          </Text>
+          <SavedSearchAlertsArtworksPlaseholderContext />
+        </Skeleton>
+      </Media>
+      <Media lessThan="md">
+        <AlertProvider isAlertArtworksView>
```
nickskalkin, line 307: **question**: same question about AlertProvider as above

src/Components/Alert/Components/Modal/ModalHeader.tsx
```diff
@@ -18,5 +18,10 @@ export const ModalHeader: FC = () => {
       return <FiltersHeader />
     case "ALERT_CONFIRMATION":
       return <ConfirmationHeader />
+    case "ALERT_ARTWORKS":
+      return <Text variant="lg">View Artworks</Text>
+    default:
+      console.warn("no matching case is found in AlertContext")
```
nickskalkin, line 24: **question**: just wondering if such console warnings are useful in general and in this particular case, for debugging purposes maybe?

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
```diff
@@ -0,0 +1,317 @@
...
+}) => {
+  if (!alert || !alert.artworksConnection)
+    return <SavedSearchAlertsArtworksPlaseholder onCloseClick={onCloseClick} />
+
+  const count = alert.artworksConnection?.counts?.total
+  const href = alert.href
+
+  return (
+    <>
+      <Media greaterThanOrEqual="md">
```
nickskalkin, line 54: **nitpick / non-blocking**: there are a few duplicated parts inside (buttons, messages, pills), I think it's possible to extract them to a common place, but up to you

nickskalkin (approved):
great work @dariakoko 👍

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditForm.tsx
dariakoko: No, we don't need, will remove it

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertEditFormPlaceholder.tsx
dariakoko: I chose to use it to render the correct header
When working on alerts in the last iteration we added header component for mobile and I decided to reuse it to stay consistent with the UI and avoid copy and paste of code
I'm pretty sure the whole implementation can se simplified, but it's a task for another day :)

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertListItem.tsx
dariakoko: GraphQL was complaining
` Expected all fields on the same parent with the name or alias `artworksConnection` to have the same argument values after applying fragment arguments.`
Do you know if we could keep using `artworksConnection(first: 1)` in SavedSearchAlertListItem.tsx?

olerichter00: That makes sense. If there is another fragment spread in the query for the same `artworksConnection`, then  align the parameters need to be aligned.

If you would need to keep the parameter `first: `), you could rename the field. However, this would result in fetching the connection twice, and I don't think we should do it in this case.

```suggestion
        smallArtworkConnection: artworksConnection(first: 1) {
```

dariakoko: >you could rename the field. However, this would result in fetching the connection twice
+++

I would keep it as it is then. What do you think?

olerichter00: Yes, let's keep it as it is 👍

src/Apps/Settings/Routes/SavedSearchAlerts/Components/SavedSearchAlertsArtworks.tsx
dariakoko: Unfortunately yes, we need it. I tried to avoid using it, but when I `remove @relay(mask: false)` and adjust the query it breaks the AlertProvider
We use the same pattern in  SavedSearchAlertEditForm - tried in this file as well and couldn't make it

src/Components/Alert/Components/Modal/ModalHeader.tsx
dariakoko: It's only for debugging purposes
I added it because at some point the app was breaking while coding because the argument didn't match any of the values in switch-case and the default value is not specified

Do you think we could use something more practical

dariakoko:
Hey @nickskalkin, thank you! Answering your comment  
>I've added `match.params` to the useEffect in the list component and it seems like it does the trick - could you confirm? Sorry for pushing to your branch!

Unfortunately it doesn't look very stable
- It doesn't update the route for the first alert
- When I open Artoworks View and navigate to alerts it shows me the artworks view of the first alert
- And this change breaks the implementation on mWeb 😞 

I recorded a video to show you the bugs
https://github.com/artsy/force/assets/36167539/a93b0623-29d6-4d89-8d68-dfbef9fe1aac

nickskalkin:
@dariakoko could you check it again? I've made a change and tried to re-produce the bugs you've mentioned and it seems like they are solved now.

Is there any reason why you've removed Sticky component around edit form?

nickskalkin:
> I have noticed that clicking on "Edit Alert" below the artwork grid could benefit from a "scroll to top" feature. Currently, when clicking the button, the edit alert form appears correctly, but the page's scroll position is near the footer, making the form not visible.

agree with @olerichter00 here

nickskalkin:
jump issue demo: 

https://github.com/artsy/force/assets/3934579/cfe0366e-9f27-4381-8f88-214228e6d7fe

olerichter00:
> agree with @olerichter00 here. I don't know how to do it though without scrolling alerts list too. I've tried adding jumpTo and placing an anchor in the Edit form, but it still scrolls alerts list

It appears that this PR removes the [sticky wrapper around the alert form/artwork list](https://github.com/artsy/force/pull/13599/files#diff-7a834fcea36d0b618cb5a1be3672d07847cc8460b1fc22cb7b63d82446ec5113L303). Re-adding the Sticky component resolves the scroll issue. However, there appear to be other changes in this pull request that are causing additional issues with the stickiness.

We should add the Sticky wrapper again and resolve the other issues.

PR #13606, opened 2024-03-14
fix(partner): Fix scrollTo on partner/artist section

damassi:

### Description

Fixes issue noted [here](https://artsy.slack.com/archives/C05F8TNKGAV/p1710452735521909) where adding `jumpTo` to the dependencies array seems to "capture" the behavior and always scroll. Seems like `jumpTo` is rerendering in a weird way. 

Will file ticket on Diamond board, cc @dzucconi


anandaroop (approved):
Works for me, so long as we think the exhaustive-deps exception is safe 👍🏽

damassi:
Yeah, my take is that using `exhaustive-deps` linter makes code unsafe and undeterministic more often than not, as the case here proves yet again.

PR #13617, opened 2024-03-20
fix(ONYX-854): partner offer notification height - adding note v1

leamotta:

This PR solves [ONYX-854]

### Description

Because images can have extreme aspect ratios, the CTAs below the artwork image are sometimes not visible without scrolling. To avoid this case and make sure the CTAs and metadata are always visible when opening the page, we have decided to set a max_height for the images.

Also added the `note` field if present, as it would certainly require some space and wanted to consider it when calculating how much space to give to the image. It doesn't have a final design yet, Emerald will work on it as a follow up.

https://github.com/artsy/force/assets/7518671/e85f1e36-e163-4d55-8711-b47c46f31846

[ONYX-854]: https://artsyproduct.atlassian.net/browse/ONYX-854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


olerichter00 (approved):
Looks great! 🌟

PR #13622, opened 2024-03-21
feat: Implement app preferences + theme select

dzucconi:
Re: [DIA-415](https://artsyproduct.atlassian.net/browse/DIA-415)

The idea here is that we set preferences in a server side cookie, then parse and cast using a Yup schema and load it into sharify. A context provider pulls it out of sharify and into state for the client. You can then update both the client-side state and the cookie simultaneously. This provides a stable value for the server-side render + client-side mount.

The toggle for dark mode is located on the account settings page and hidden behind the feature flag `diamond_dark-mode`.

[DIA-415]: https://artsyproduct.atlassian.net/browse/DIA-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/AppPreferences/appPreferencesMiddleware.ts
+++ b/src/Apps/AppPreferences/appPreferencesMiddleware.ts
@@ -1,0 +1,15 @@
+import { getAppPreferences } from "Apps/AppPreferences/appPreferencesServerRoutes"
+import { APP_PREFERENCES_SHARIFY_KEY } from "Apps/AppPreferences/useAppPreferences"
+import { ArtsyRequest, ArtsyResponse } from "Server/middleware/artsyExpress"
+import { updateSharifyAndContext } from "Server/middleware/bootstrapSharifyAndContextLocalsMiddleware"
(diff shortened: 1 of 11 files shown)


src/Apps/AppPreferences/appPreferencesMiddleware.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const appPreferencesMiddleware = async (
+  req: ArtsyRequest,
+  res: ArtsyResponse,
+  next: NextFunction
+) => {
+  const preferences = getAppPreferences(req)
+  updateSharifyAndContext(res, APP_PREFERENCES_SHARIFY_KEY, preferences)
+  next()
+}
```
damassi, line 15: I'm happy that we're colocating all of this stuff 👍

damassi (approved):
Makes sense, and thanks for the additional tests 👍

PR #13626, opened 2024-03-21
feat: [EMI-1739] Show expired offer banner when expecting a partner offer id in query params

erikdstock:
This PR solves [EMI-1739]

### Description
This PR adds a query param to tell the artwork page that the user is expecting to see a specific partner offer id. We can link to this from anywhere with the url `/artwork/:slug?partner_offer_id={the-id}`.

It also moves the banner code from ArtworkApp into a single component with a few basic tests to get things started. This was motivated in part to make the code cleaner and in part because I was thinking about whether multiple banners could be triggered at once, and while that's technically possible - I don't think it would happen in practice due to the reliance on different artwork types and query params.

<details><summary>Example of unlikely scenario where we trigger multiple banners (out of scope)</summary>
<p>

image

</p>
</details> 

_This is behind the new `emerald_partner-offers-to-artwork-page` feature flag._

Existing banner behavior is retained.

<details><summary>Screenshots</summary>
<p>

### Banner appears with `?partner_offer_id=a not found id`
image

### Banner does not appear with a found id
image

### Banner respects legacy `expired_offer=bool` param
image

</p>
</details> 

[EMI-1739]: https://artsyproduct.atlassian.net/browse/EMI-1739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -96,12 +97,32 @@ export const ArtworkApp: React.FC<Props> = props => {
...
+
+  const expectedPartnerOfferNotFound =
+    partnerOfferVisibilityEnabled &&
+    expectedPartnerOfferID &&
+    (!partnerOffer || partnerOffer.internalID !== expectedPartnerOfferID)
+
+  const showExpiredOfferBanner =
+    expectedPartnerOfferNotFound || !!match?.location?.query?.expired_offer
+
+  // show unavailable artwork imperatively if the query param is present
```
damassi, line 125: Just a suggestion: 

We might as well just move all the banner-related code into an `<ArtworkBanners />` component. It's time, I think! 

And then spread in an `ArtworkBanners_viewer` fragment, by updating the existing fragment:

```
fragment ArtworkApp_viewer on Viewer @argumentDefinitions({ artworkID ...}) {
  ...ArtworkBanners_viewer @arguments(artworkID: $artworkID)
  
   artwork(id: $artworkID) {
     // all of the existing fragment code 
   } 
}

damassi (approved):
Looks great @erikdstock

PR #13628, opened 2024-03-24
fix: properly display privacy status after list creation

oxaudo:
This should complete https://artsyproduct.atlassian.net/browse/EMI-1761

Had several of hours of waiting in between the trains and tracked down this error. Super minor at the end but took a bit of figuring out how this page set up :)

Also noticed that 'lock icon' is right aligned in the card in designs. So adjusted that as well.


src/Apps/CollectorProfile/Routes/Saves/Components/Actions/Mutations/useCreateCollection.ts
```diff
@@ -50,6 +50,7 @@ export const useCreateCollection = () => {
               collection {
                 internalID
                 name
+                shareableWithPartners
```
starsirius, line 53: 👍

PR #13640, opened 2024-03-26
feat: enable pills and counts for the availability filter

anandaroop:

This PR solves [ONYX-702] and [ONYX-703]

### Description

This wraps up the Hackathon request that got stalled for a bit on the filter param logic (KS'd [here](https://www.notion.so/artsy/Boolean-filters-on-the-artwork-grid-0ccf4af13df54712b7c88b0480d377f8) and addressed in https://github.com/artsy/gravity/pull/17446)

Now that the for-sale toggle is filtering results correctly, we want to ensure the rest of the search UI works accordingly as well.

- A **pill** corresponding to the for-sale filter selection should be displayed, and allow for dismissal
- The **count** of active filters should account for the for-sale filter

Addressing the counts also turned up some gaps in the mWeb behavior, which should be addressed now.

Remains staging-only behind [an Unleash flag](https://unleash.artsy.net/projects/default/features/onyx_availability-filter) — but ready for final QA now I believe.

|Web|mWeb|
|---|---|
|||

[ONYX-702]: https://artsyproduct.atlassian.net/browse/ONYX-702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-703]: https://artsyproduct.atlassian.net/browse/ONYX-703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Alert/AlertProvider.tsx
+++ b/src/Components/Alert/AlertProvider.tsx
@@ -93,6 +93,12 @@ export const AlertProvider: FC<AlertProviderProps> = ({
     if (isEditMode || isAlertArtworksView) return
     const criteria = getAllowedSearchCriteria(initialCriteria ?? {})
 
+    // `forSale` is allowed as a filter criterion,
+    // but NOT as an alert criterion, so we remove it.
+    // (Alerts, by definition, stipulate forSale=true
+    // when they are created in Gravity.)
+    delete criteria.forSale
+
     dispatch({ type: "SET_CRITERIA", payload: criteria })
   }, [initialCriteria, isEditMode, isAlertArtworksView])
 
--- a/src/Components/ArtworkFilter/ArtworkFilters/__tests__/AvailabilityFilter.jest.tsx
+++ b/src/Components/ArtworkFilter/ArtworkFilters/__tests__/AvailabilityFilter.jest.tsx
@@ -4,16 +4,14 @@ import {
   createArtworkFilterTestRenderer,
   currentArtworkFilterContext,
 } from "Components/ArtworkFilter/ArtworkFilters/__tests__/Utils"
-import { ArtworkFilterContextProps } from "Components/ArtworkFilter/ArtworkFilterContext"
+import { useArtworkFilterContext } from "Components/ArtworkFilter/ArtworkFilterContext"
 import { AvailabilityFilter } from "Components/ArtworkFilter/ArtworkFilters/AvailabilityFilter"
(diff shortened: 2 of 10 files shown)


src/Components/Alert/AlertProvider.tsx
```diff
@@ -93,6 +93,12 @@ export const AlertProvider: FC<AlertProviderProps> = ({
     if (isEditMode || isAlertArtworksView) return
     const criteria = getAllowedSearchCriteria(initialCriteria ?? {})
 
+    // `forSale` is allowed as a filter criterion,
+    // but NOT as an alert criterion, so we remove it.
+    // (Alerts, by definition, stipulate forSale=true
+    // when they are created in Gravity.)
+    delete criteria.forSale
```
anandaroop, line 100: An important detail about this filter is that we _don't_ want to pass it into the initial alert criteria.

Alerts always specify **for sale: true** as a filter criteria, which [happens upstream in Gravity by default](https://github.com/artsy/gravity/blob/702f5b020082f9264d30f89088a9c8fc7650f2c2/app/models/search_criteria.rb#L23) when the SearchCriteria gets persisted.

src/Components/ArtworkFilter/ArtworkFilters/__tests__/AvailabilityFilter.jest.tsx
```diff
@@ -4,16 +4,14 @@ import {
...
 
-const artworkFilterContext: Partial<ArtworkFilterContextProps> = {
-  filters: {
-    forSale: undefined,
-  },
-}
+jest.mock("Utils/Hooks/useMatchMedia")
 
-const render = createArtworkFilterTestRenderer(artworkFilterContext)
+const render = createArtworkFilterTestRenderer()
```
anandaroop, line 14: This call to create the custom test renderer was simplified — a little less explicit but more like the actual implementation. This is so that the filter counts can be calculated correctly in the test.

mzikherman (approved):

nickskalkin (approved):
looks great 👍

tam-kis (approved):

olerichter00 (approved):

PR #13641, opened 2024-03-26
feat(EMI-1707): offer settings modal

rquartararo:
The type of this PR is: **Feature**

This PR solves [EMI-1707]

### Description

This adds the new "Offer Settings" modal that allows collections to bulk update the privacy settings for their saved artwork lists. 

@artsy/emerald-devs 

### Screen Recording 
https://github.com/artsy/force/assets/50849237/287cdab5-dd5f-44dc-ad74-c02c980bb3d2

### Mobile Web
Screenshot 2024-03-26 at 6 56 24 PM

### Progressive Onboarding Tooltip

Screenshot 2024-03-26 at 6 41 45 PM

### Edge cases long name and no artwork photo

Screenshot 2024-03-26 at 9 35 28 PM

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1707]: https://artsyproduct.atlassian.net/browse/EMI-1707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13643, opened 2024-03-27
feat(EMI-1773): update copy for saves page tooltips and modals

fladson:

This PR solves [EMI-1773]

### Description

- Update the "Create a new list" description copy
SCR-20240326-v9c

- Replace the lock icon for a hide icon
- Update the list header to remove the list privacy (Shared or Private) and move the icon to the title itself
- Add tooltips for both icons
SCR-20240326-v9x
SCR-20240326-v9n

- Replace the lock icon in the SelectArtworkListItem modal
SCR-20240327-oug

This will have merge conflicts with https://github.com/artsy/force/pull/13641, so we should merge that first.
@artsy/emerald-devs 

[EMI-1773]: https://artsyproduct.atlassian.net/browse/EMI-1773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListContent.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListContent.tsx
@@ -98,9 +98,9 @@ const ArtworkListContent: FC<ArtworkListContentProps> = ({ me, relay }) => {
       >
         <Join separator={<Spacer x={2} />}>
           <Flex alignItems="center">
-            <Text variant="lg-display">{artworkList.name}</Text>
+            <Text variant="lg-display">{artworkList?.name}</Text>
             {shareableWithPartnersEnabled &&
-              !artworkList.shareableWithPartners && (
+              !artworkList?.shareableWithPartners && (
                 <Tooltip
                   pointer
                   variant="defaultDark"
@@ -121,7 +121,7 @@ const ArtworkListContent: FC<ArtworkListContentProps> = ({ me, relay }) => {
                 </Tooltip>
               )}
           </Flex>
-          {!artworkList.default && (
+          {!artworkList?.default && artworkList && (
             <ArtworkListContextualMenu artworkList={artworkList} />
           )}
         </Join>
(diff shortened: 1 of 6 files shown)


src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListContent.tsx
```diff
@@ -55,11 +56,12 @@ function isContentOutOfView() {
 }
 
 const ArtworkListContent: FC<ArtworkListContentProps> = ({ me, relay }) => {
+  const { t } = useTranslation()
   const { match } = useRouter()
 
   const { jumpTo } = useJump()
   const { artworkListItemHasBeenTouched } = useArtworkListVisibilityContext()
-
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
fladson: Yeah, I had to add that because I haven't changed that line, so probably a later edit in this file triggered the linter.

With your proposed change, I am getting this.
image

src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListContent.tsx
```diff
@@ -55,14 +56,14 @@ function isContentOutOfView() {
...
 
 const ArtworkListContent: FC<ArtworkListContentProps> = ({ me, relay }) => {
+  const { t } = useTranslation()
   const { match } = useRouter()
 
   const { jumpTo } = useJump()
   const { artworkListItemHasBeenTouched } = useArtworkListVisibilityContext()
-
-  const artworkList = me.artworkList!
+  const artworkList = me.artworkList
```
oxaudo, line 64: I do support this change as I also thought that this enforcement is strange here when I looked at it first time, but with this change expected to see some more `?.` added down the file. Should we update all the usages for this `artworkList`?

oxaudo (changes requested):
This looks good to me. 

I think we are missing one more usage of this LockIcon that is replaced with HideIcon - `SelectArtworkListItem` that is used to when user selects which list to add saved artwork to. I assume the icon needs to be changed there as well?

fladson:
> SelectArtworkListItem

Good callout, it was not mentioned in the ticket but we should update it there as well, thanks for flagging!

src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListContent.tsx
fladson: no more lint errors, it would block the git push/commit.

oxaudo (approved):
👍

PR #13644, opened 2024-03-27
feat(collectorprofile): roughs in artists page

dzucconi:
Closes [DIA-535](https://artsyproduct.atlassian.net/browse/DIA-535)

[DIA-535]: https://artsyproduct.atlassian.net/browse/DIA-535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsList.tsx
+++ b/src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsList.tsx
@@ -1,0 +1,81 @@
+import { Stack, Text } from "@artsy/palette"
+import { FC } from "react"
+import { graphql } from "react-relay"
+import { useClientQuery } from "Utils/Hooks/useClientQuery"
+import { CollectorProfileArtistsListQuery } from "__generated__/CollectorProfileArtistsListQuery.graphql"
+import {
+  CollectorProfileArtistsListArtist,
+  CollectorProfileArtistsListArtistSkeleton,
+} from "Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsListArtist"
+import { compact } from "lodash"
+
+interface CollectorProfileArtistsListProps {}
+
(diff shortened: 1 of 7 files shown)


src/Apps/CollectorProfile/collectorProfileRoutes.tsx
```diff
@@ -131,6 +141,13 @@ export const collectorProfileRoutes: AppRouteConfig[] = [
       }
     `,
     children: [
+      {
+        path: "artists",
```
dzucconi, line 145: The route should be: `/collector-profile/artists`, not `/collector-profile/my-collection/artists`, to match the other routes.

src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsList.tsx
```diff
@@ -0,0 +1,81 @@
...
+      {loading ? (
+        new Array(10).fill(null).map((_, i) => {
+          return <CollectorProfileArtistsListArtistSkeleton key={i} />
+        })
+      ) : (
+        <>
+          {userInterestEdges.map(userInterestEdge => {
+            return (
+              <CollectorProfileArtistsListArtist
+                key={userInterestEdge.internalID}
```
araujobarret, line 58: nit: I think we have be more specific, if this list component reuses the internalID we'll have duplicated keys being used

araujobarret (dismissed):
I know sounds tedious to write unit test for a simple fragment like this, but that's the way we should work, with unit tests 😜

src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsList.tsx
araujobarret: nvm it's very uncommon for someone to use the `internalID` under any area of `CollectorProfileArtistsListArtist`. I was overthinking about the reuse of the same key (internalID).

PR #13647, opened 2024-03-27
feat: EMI-1736 Create partner-offer-order with purchase button if an active partner offer is present

erikdstock:
This PR completes [EMI-1736].
- update commitMutation usage to useMutation hook
- query partner offers and select [with feature flag]
- use timer to check when partner offer expires

 ### Description

The implementation is mostly straightforward. I re-used the existing mutation to create an offer from a partner offer (some error handling there could be redundant). I migrated the existing old `commitMutation`-style mutations to use our `useMutation()` hook. I did have trouble adding new behavior tests to the `systemContext.router.push()` functionality when the order mutation completes, so I ultimately dropped these assertions.

[EMI-1736]: https://artsyproduct.atlassian.net/browse/EMI-1736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercialButtons.jest.tsx
```diff
@@ -1,51 +1,88 @@
...
 describe("ArtworkSidebarCommercialButtons", () => {
   let user
+  let partnerOffers
 
-  let mockEnvironment = createMockEnvironment()
+  let mockEnvironment
+  const mockUseFeatureFlag = useFeatureFlag as jest.Mock
 
-  const { renderWithRelay } = setupTestWrapperTL<
+  const { renderWithRelay: renderRaw } = setupTestWrapperTL<
```
damassi, line 35: If `1:2` is _required_, rather than renaming the function to something random its helpful to amend it so that find / replace and other search operations remain useful:
```
const { renderWithRelay: baseRenderWithRelay } = setupTestWrapperTL
```

MrSltun (approved):
Looks good! 🌟

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -414,6 +414,7 @@ interface ArtworkResultProps extends RenderProps {
 const ArtworkResult: React.FC<ArtworkResultProps> = props => {
   const { artworkResult, ...rest } = props
   const { __typename } = artworkResult
+  console.log("ME!!!!!!!!", props.me)
```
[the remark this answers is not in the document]
oxaudo: 🙈

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -47,13 +53,34 @@ const SaleMessage: React.FC<SaleMessageProps> = ({ saleMessage }) => {
...
+
+  // Get the first not-ended partner offer, if available
+  const partnerOffer =
+    (partnerOfferVisibilityEnabled &&
+      me?.partnerOffersConnection &&
+      extractNodes(me.partnerOffersConnection)[0]) ||
+    null
+
+  // Fall back to a definitely past value because the timer hook doesn't like nulls
+  const partnerOfferTimer = useTimer(partnerOffer?.endAt || THE_PAST)
```
oxaudo, line 78: #notblocking: This usage of the timer is only applicable if offer is present. Ideally we should probably figure put how to not instantiate all this objecs if offer is not in a picture - that would probably avoid usage of this PAST hack. But I don't really see right away how to refactor it. Cool with trying this as a follow up.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -99,7 +131,51 @@ const ArtworkSidebarCommerialButtons: React.FC<ArtworkSidebarCommercialButtonsPr
     showInquiry({ enableCreateAlert: true })
   }
 
-  const handleCreateOrder = () => {
+  const handleCreatePartnerOfferOrder = async () => {
```
oxaudo, line 134: Might be more clear if this function expects activePartnerOffer and throws error if it's called without it. If I understand correctly currently it would just silently do nothing if called without offer. Seems like it can introduce some errors where it's called in a wrong place and would be tricky to debug as UI will just do nothing.

oxaudo (approved):
This looks good to me.

Have couple of notes about possible refactors here but none of this is blocking. Cool if we just merge this as is and work further in parallel.

src/Apps/Artwork/ArtworkApp.tsx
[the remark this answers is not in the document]
erikdstock: You don't like it? 🥺

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
erikdstock: Right, I had the same thought. I can't call the hook conditionally so this felt like the easiest route for now that wouldn't require handling a lot of weird/optional values in other usages of this timer.

src/Apps/Artwork/Components/ArtworkSidebar/__tests__/ArtworkSidebarCommercialButtons.jest.tsx
```diff
@@ -1,51 +1,88 @@
...
         }
+        me {
+          ...ArtworkSidebarCommercialButtons_me
+            @arguments(artworkID: "josef-albers-homage-to-the-square-85")
+        }
       }
     `,
   })
 
+  const renderWithRelay: typeof baseRenderWithRelay = (
```
araujobarret, line 61: Not sure if understand why we can't use the tooling we already have (`setupTestWrapperTL`)?
Moving away from it creates more tech debt IMO, especially when it comes to upgrades, I had a painful time upgrading Relay here because of multiple patterns for testing.

araujobarret, line 61: I wonder how much benefit this abstraction could bring. 
I think we already have a good tooling for testing we don't/should add anything on top, even if that means the tests being a bit more verbose, these abstractions have 2 big downsides imo:
- creates an extra layer of complexity, whoever comes here will need to read and understand this new abstraction, it moves things away from being straightforward
- creates some unknowns only discovered when you have an upgrade and some tests failing, so you need to come, spend time trying to understand what's going on 

So I'm very in favor of having a bit more of repetition during tests rather than more layers on top of tooling

erikdstock: OK, will remove.

PR #13649, opened 2024-03-28
feat: update footer links (DIA-123)

iskounen:
This PR reshuffles the footer links behind a feature flag. In addition to a new ordering, the following changes have been made:

1. "Terms of Use" becomes "Terms and Conditions"
2. "Conditions of Sale" becomes "Auction Supplement" and has a new URL
3. "ACA Seller Conditions" is removed

This PR also sneaks in a refactor of the footer link layout. Thanks to @araujobarret, we were able to remove the margin properties (which can cause layout problems when the parent component is updated in Palette) and space out each footer link using a [Flex gap](https://developer.mozilla.org/en-US/docs/Web/CSS/gap).

One side-effect of this change is that the footer links have a vertical gap on mobile web (see screenshots below). We decided that this was a good thing because it now gives users more space to actually tap on those links.

| Before | After |
|--------|------|
| Screenshot 2024-03-28 at 7 33 20 AM | Screenshot 2024-03-28 at 7 33 41 AM |
| localhost_4000_(iPhone SE) (1) | localhost_4000_(iPhone SE) |

### DIFF
--- a/src/Components/Footer/Footer.tsx
+++ b/src/Components/Footer/Footer.tsx
@@ -28,6 +28,7 @@ import TikTokIcon from "@artsy/icons/TikTokIcon"
 import SpotifyIcon from "@artsy/icons/SpotifyIcon"
 import { useSystemContext } from "System/useSystemContext"
 import ArtsyMarkIcon from "@artsy/icons/ArtsyMarkIcon"
+import { useFeatureFlag } from "System/useFeatureFlag"
 
 interface FooterProps extends BoxProps {}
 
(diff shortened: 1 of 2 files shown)


src/Components/Footer/Footer.tsx
```diff
@@ -276,6 +277,9 @@ export const Footer: React.FC<FooterProps> = props => {
 
 const PolicyLinks = () => {
   const { CCPARequestComponent, showCCPARequest } = useCCPARequest()
+  const newTermsAndConditionsEnabled = useFeatureFlag(
+    "diamond_new-terms-and-conditions"
+  )
```
dzucconi, line 282: Why does this simple change need a feature flag?

iskounen: There's a host of changes that are all being placed behind this feature flag so that we can update the terms and conditions across the ecosystem at once. In this case it's useful because the page that "Auction Supplement" links to hasn't been created yet.

dzucconi: 👍

araujobarret (approved):

PR #13658, opened 2024-04-01
feat: new disclaimer in signup dialog (DIA-533)

iskounen:
This PR updates the disclaimer at the bottom of the signup dialog behind a feature flag. In addition to the default signup dialog, there is a skeleton loader for that dialog, as well as a signup dialog in the inquiry flow that needs to be updated.

| Before | After |
|-------|------|
| before_auth_signup | after_auth_signup |
| before_inquiry_signup | after_inquiry_signup |

### DIFF
--- a/src/Components/AuthDialog/Components/AuthDialogSignUpPlaceholder.tsx
+++ b/src/Components/AuthDialog/Components/AuthDialogSignUpPlaceholder.tsx
@@ -7,9 +7,12 @@ import {
   SkeletonText,
   Spacer,
 } from "@artsy/palette"
+import { useFeatureFlag } from "System/useFeatureFlag"
 import { FC } from "react"
 
 export const AuthDialogSignUpPlaceholder: FC = () => {
+  const showNewDisclaimer = useFeatureFlag("diamond_new-terms-and-conditions")
+
   return (
     <Skeleton>
       <Join separator={<Spacer y={2} />}>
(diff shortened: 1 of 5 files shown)


araujobarret (approved):

src/Components/AuthDialog/Components/AuthDialogSignUpPlaceholder.tsx
```diff
@@ -42,10 +42,14 @@ export const AuthDialogSignUpPlaceholder: FC = () => {
...
-        <SkeletonText variant="xs" textAlign="center">
-          By clicking Sign Up or Continue with Apple, Google, or Facebook, you
-          agree to Artsy’s Terms of Use and Privacy Policy and to receiving
-          emails from Artsy.
+        <SkeletonText
+          variant="xs"
+          textAlign="center"
+          data-testid="skeleton-disclaimer"
+        >
+          By clicking Sign Up or Continue with Email, Apple, Google, or
```
egdbear, line 50: **nitpick:** Not sure if it was raised before to the writing team, but, should `Sign Up` be `"Sign Up"`?
I was looking at the GitHub Terms and Conditions, and they were using some of the quotation [here](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#l-cancellation-and-termination) for the "internal" features. (I definitely used it incorrectly in my review comments 😅 )

iskounen: I don't think that quotations are needed in this context, especially since we'd need to put "Continue with Email" and all its variations in quotations as well.

Regarding the GitHub Terms and Conditions, I see them putting words or phrases in quotations when they need to be defined, like [here](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#a-definitions). That is commonly done in legal documents to make them easier to read, but I don't think we need to do that here.

PR #13659, opened 2024-04-01
feat: EMI-1775 require login to view expected partner offer

erikdstock:
This PR solves [EMI-1775]

### Description

This PR adds a login check if the artwork page loads with a `partner_offer_id` in the query string. This param suggests that the user navigate to the page expecting to see a partner offer. The functionality of the param is currently just to show a banner if it's not found, but since this action is more _global_ to the artwork page I put it in ArtworkApp.

In addition the added code removes the param from the URL once the user is logged in. This prevents an edge case where the user log out, causing the login modal to appear immediately again.

_Added tests_ There weren't any tests for the ArtworkApp, and there isn't much here to test. I added tests for this effect hook.

[EMI-1775]: https://artsyproduct.atlassian.net/browse/EMI-1775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -88,12 +90,42 @@ const BelowTheFoldArtworkDetails: React.FC<BelowTheFoldArtworkDetailsProps> = ({
...
 
   const privateArtworksEnabled = useFeatureFlag(
     "amber_artwork_visibility_unlisted"
   )
 
+  useEffect(() => {
+    const expectingPartnerOffer = !!match?.location?.query?.partner_offer_id
+    const isLoggedIn = !!me
+
+    if (expectingPartnerOffer && typeof window !== "undefined") {
```
dzucconi, line 104: Shouldn't this check on window be unnecessary since this executes in an effect?

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -88,12 +90,42 @@ const BelowTheFoldArtworkDetails: React.FC<BelowTheFoldArtworkDetailsProps> = ({
...
+      if (!isLoggedIn) {
```
dzucconi, line 105: Short-circuiting `returns` would be a lot more clear than this nested conditional

src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -88,12 +90,42 @@ const BelowTheFoldArtworkDetails: React.FC<BelowTheFoldArtworkDetailsProps> = ({
...
   const { artwork, me, referrer, tracking, shouldTrackPageView } = props
-  const { match, silentPush } = useRouter()
+  const { match, silentPush, silentReplace } = useRouter()
+  const { showAuthDialog } = useAuthDialog()
 
   const privateArtworksEnabled = useFeatureFlag(
     "amber_artwork_visibility_unlisted"
   )
 
+  useEffect(() => {
```
dzucconi, line 100: Can you leave a comment on what exactly this is doing and why? It's really unclear just reading the code.

MrSltun (dismissed):
Looks good! 🌟
Maybe we should create a test for this case?

erikdstock:
> Looks good! 🌟 Maybe we should create a test for this case?

Added tests!

src/Apps/Artwork/__tests__/ArtworkApp.jest.tsx
```diff
@@ -0,0 +1,153 @@
...
+      },
+    },
+  }
+  useRouterMock.mockReturnValue(routerMock)
+})
+
+// Can only be done once per test because the property is read-only
+const setWindowLocation = (location: Partial<Location>) => {
+  Object.defineProperty(window, "location", { value: location })
+}
```
erikdstock, line 100: was wondering about this and found we have a way of doing it already [here](https://github.com/artsy/force/blob/e64ac8ce6881c2200aa56cc1d32c72d0cbbdfbb8/src/DevTools/mockLocation.ts#L4-L17)

PR #13660, opened 2024-04-01
feat[EMI-1748]: display offer details for user looking at artwork that they have an offer on.

oxaudo:
This adds display for the artwork offer data with suggestion on how to refactor the List price display once this ticket is merged:
https://github.com/artsy/metaphysics/pull/5637

Current look:
Screen Shot 2024-04-01 at 4 00 45 PM


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -354,15 +355,64 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
...
+            justifyContent="start"
+          >
+            <SaleMessage
+              saleMessage={activePartnerOffer.priceWithDiscount?.display}
+            />
+            <Spacer x={1} />
+            <Text variant="md" color="black60" style={{ whiteSpace: "nowrap" }}>
+              {/* TODO: Move this logic of display to Artwrok MP model */}
+              (List price: {artwork.price || "Not publicly listed"})
+            </Text>
```
erikdstock, line 388: :+1: Good to clarify this, I wonder if it is worth raising at web or platform practice

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -354,15 +355,64 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
...
+
+          <Spacer y={0.5} />
+
+          <Flex
+            flexDirection={["column", "row"]}
+            width="100%"
+            justifyContent="start"
+          >
+            <SaleMessage
+              saleMessage={activePartnerOffer.priceWithDiscount?.display}
```
erikdstock, line 382: Non-blocking: Should never happen, but it strikes me here that this could technically be not present.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -354,15 +355,64 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
     artwork.isAcquireable || activePartnerOffer
   )
 
+  const SaleMessageOrOfferDisplay: FC = () => {
+    if (activePartnerOffer) {
```
erikdstock, line 359: minor: Should this component be defined outside the main component body? Then artwork/activePartnerOffer could be props and the function wouldn't need to be redefined.

erikdstock, line 359: Judging from https://github.com/artsy/force/pull/13660/files#diff-7e67206208c67ae3f411657c84fa58ff49ccc4f44682f608a1ce8d56b46d4720R81-R82 I think `activePartnerOffer` will disappear when the timer runs out. In this case the offer from the gallery would disappear on a reload, but if the timer runs out while the user is on the page do we want the pricing to just suddenly change, timer disappear, etc? Not certain what the right answer is here.

If we don't want it to disappear completely, the variable we can use is just `partnerOffer.`

erikdstock (commented):
Looks good! A few minor comments, one about behavior that I'm not quite sure about.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
oxaudo: You mean `activePartnerOffer` ? This part is only rendered if it's present according to the check above. I guess I need to be checking `partnerOffer` as you noted above. And price should always be present there. This `?` is just for error cases so that page does not crash.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
oxaudo: Updated this part.

oxaudo: I just moved the actual display code to OfferDisplay component similar to how SaleMessage is done here. Theoretically we can make it to be stand alone thing just defining it's own fragment but for now did not seem necessary. 

Can refactor separately if we need to reuse it in any other places.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -52,6 +53,54 @@ const SaleMessage: React.FC<SaleMessageProps> = ({ saleMessage }) => {
...
 }
 
+interface OfferDisplayProps {
+  originalPrice: string | null | undefined
+  offerPrice: string | null | undefined
+  endAt?: string | null
+  isAvailable?: boolean | null
+}
+
+const OfferDisplay: React.FC<OfferDisplayProps> = ({
```
erikdstock, line 63: minor: For organization I'd move this component [down with the others below the main export](https://github.com/artsy/force/pull/13660/files#diff-7e67206208c67ae3f411657c84fa58ff49ccc4f44682f608a1ce8d56b46d4720R406).

erikdstock (dismissed):
One minor comment, LGTM!

PR #13669, opened 2024-04-02
feat(MyC, SwA): present disambiguating info when selecting an artist

anandaroop:
> [!IMPORTANT]  
> As a reminder, **any** time we present a list of artists from which a user will select, we should also display disambiguating info: nationality, birth date, death date.

This PR solves [ONYX-438]

### Description

(Counterpart to https://github.com/artsy/eigen/pull/10018)

This adds disambiguating information to the artist selection UI that is used in both My Collection and Sell with Artsy, for similar reasons as stated in the Eigen PR — to minimize the risk of misattributed works and duplicated artist data.

## My Collection

Before and after…

myc

---

## Sell with Artsy

Before and after…

swa

[ONYX-438]: https://artsyproduct.atlassian.net/browse/ONYX-438?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mzikherman (approved):
Nice!

dblandin (approved):
🔥

PR #13677, opened 2024-04-03
feat: redirect conditions of sale

iskounen:
This PR adds a permanent redirect from /conditions-of-sale to /supplemental-cos. This PR has a "Do not merge" label because the redirect destination has not yet been created. It is planned to be ready to launch by April 17, 2024.


dzucconi:
Why not just update the text when we're ready?

iskounen:
So I don't have to think about this anymore.

PR #13698, opened 2024-04-08
chore: AMBER-552 - private artworks checkout

Mitch-Henson:

This PR solves [AMBER-552]

### Description

This includes the `Exclusive Access` text which will be shown during checkout for an unlisted artwork. As part of this change set, I also updated some other places in Force which was previously manually checking the visibility level to instead take advantage of [this Metaphysics function](https://github.com/artsy/metaphysics/blob/d74fe3fabbfc2212b51a89e2108c6b2b4fd34e5b/src/schema/v2/artwork/index.ts#L783) which already takes care of this.

As part of this work, I had to update some existing code in commit 2e0d2807571c8fc4ed37ba7cedd258d0c631a36a as there we some complaints about non-null assertion issues.

Before / also for public works:
Screenshot 2024-04-08 at 12 14 30

After / only for private works:
Screenshot 2024-04-08 at 12 14 00

[AMBER-552]: https://artsyproduct.atlassian.net/browse/AMBER-552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lidimayra (approved):
> As part of this change set, I also updated some other places in Force which was previously manually checking the visibility level to instead take advantage of [this Metaphysics function](https://github.com/artsy/metaphysics/blob/d74fe3fabbfc2212b51a89e2108c6b2b4fd34e5b/src/schema/v2/artwork/index.ts#L783) which already takes care of this.

Tahnk you!!! 🙌 ❤️

PR #13702, opened 2024-04-08
feat(EMI-1800): update display to only have 2 active action buttons at a time

oxaudo:
I tried to refactor the logic how we determine what buttons to display in which state - hopefully that is more readable now.

Here is the current look. 
Screen Shot 2024-04-08 at 2 51 55 PM

The question I have here if we need to update partner section to display this 'contact gallery' button if purchase button bumped out the 'contact gallery' from main buttons?

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
@@ -312,16 +312,8 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
     setSelectedEditionSet(firstAvailableEcommerceEditionSet())
   }, [artwork.editionSets, firstAvailableEcommerceEditionSet])
 
-  const artworkEcommerceAvailable = !!(
-    artwork.isAcquireable || artwork.isOfferable
-  )
-  const shouldRenderButtons =
-    artworkEcommerceAvailable || !!artwork.isInquireable
-
   const isCreateAlertAvailable =
     artwork.isEligibleToCreateAlert && artwork.isSold
-  const isSecondaryContactGalleryButton =
-    artwork.isOfferable || isCreateAlertAvailable
 
   const AlertSwitch: FC = () => {
     if (!isCreateAlertAvailable) {
@@ -338,6 +330,19 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
   const buyNowOrPartnerOfferAvailable = !!(
     artwork.isAcquireable || activePartnerOffer
   )
+
+  const renderButtons = {}
+  if (buyNowOrPartnerOfferAvailable) {
+    renderButtons["buyNow"] = "primaryBlack"
+  }
+  if (artwork.isOfferable) {
+    renderButtons["makeOffer"] =
+      Object.keys(renderButtons).length == 0 ? "primaryBlack" : "secondaryBlack"
+  }
+  if (artwork.isInquireable && Object.keys(renderButtons).length < 2) {
+    renderButtons["contactGallery"] =
+      Object.keys(renderButtons).length == 0 ? "primaryBlack" : "secondaryBlack"
+  }
 
   const SaleMessageOrOfferDisplay: FC = () => {
     return partnerOffer ? (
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -380,13 +385,14 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
...
       )}
 
-      {shouldRenderButtons && <Spacer y={2} />}
+      {Object.keys(renderButtons).length > 0 && <Spacer y={2} />}
 
       <Flex flexDirection={["column", "column", "column", "column", "row"]}>
         <Join separator={<Spacer x={1} y={1} />}>
           <AlertSwitch />
-          {buyNowOrPartnerOfferAvailable && (
+          {renderButtons["buyNow"] && (
```
damassi, line 393: `renderButtons.buyNow`

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -400,13 +406,9 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
               {t("artworkPage.sidebar.commercialButtons.buyNow")}
             </Button>
           )}
-          {artwork.isOfferable && (
+          {renderButtons["makeOffer"] && (
```
damassi, line 409: `renderButtons.makeOffer`

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
oxaudo: To make it work this way seems like I need to be more strict with object typing. But maybe it's a good thing. What you think now @damassi ?

damassi: Yah we should type stuff! 

```
const foo: {
  buyNow: boolean
  makeOffer: boolean
} = {
 ... 
}

damassi (approved):

narikazu:
@oxaudo 

> The question I have here if we need to update partner section to display this 'contact gallery' button if purchase button bumped out the 'contact gallery' from main buttons?

Is there any case where the `Purchase` button bumps out the `Contact Gallery` from the main buttons?

oxaudo:
@narikazu - definitely. The case where the work was configured for MO only and Contact gallery. Now we are adding 'Purchase' but believe we are not supposed to have more than 2 buttons in this area - so Contact gallery is out. We can decide that we want to display contact gallery instead of Offer in this case - in fact it probably makes more sense. But that is not what design suggested.

Maybe something to bring up today in knowledge share.

oxaudo:
@narikazu - based on our conversation earlier I'm inclined to merge this as it gets us closer to what we want to see eventually. And we should create a separate card to add 'contact gallery' to partner details section on the page in case 'contact gallery' is not shown in the main section. The rules are still a bit confusing to me but the change is in a different component - so I think we can move it into separate story.

narikazu:
> @narikazu - based on our conversation earlier I'm inclined to merge this as it gets us closer to what we want to see eventually. And we should create a separate card to add 'contact gallery' to partner details section on the page in case 'contact gallery' is not shown in the main section. The rules are still a bit confusing to me but the change is in a different component - so I think we can move it into separate story.

@oxaudo 
Thank you for raising it in the KS. I've created [the ticket](https://artsyproduct.atlassian.net/browse/EMI-1812) to address it. Feel free to merge the PR!

oxaudo:
Ok @narikazu - hopefully last round here. Updated the logic for those button to align with latest thinking. Also tried to structure the specs to follow design from figma. Hope that is a bit more comprehensive in describing our expectations. 

Let me know if things look good to you plz!

narikazu (approved):
It looks good to me. Thank you for dealing with this complex senario!

To summarize the final decision, the `Purchase` and `Contact Gallery` buttons show up on the artwork page when the work has a partner offer and for MOOAEA, which is not eligible for MO.

PR #13719, opened 2024-04-11
fix: Correctly initializing SWA form contact info on back press

olerichter00:
Addresses [ONYX-821]

## Description

Correctly initializing the SWA form contact info on back press. This was not happening before at this step. It is likely that it was never set up because this is the first step of the form without loading an already existing submission.

Screenshot 2024-04-11 at 13 50 30

[ONYX-821]: https://artsyproduct.atlassian.net/browse/ONYX-821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/ContactInformation/ContactInformation.tsx
@@ -24,26 +24,9 @@ import {
   ContactInformationFormModel,
 } from "./Components/ContactInformationForm"
 import { TopContextBar } from "Components/TopContextBar"
+import { getContactInformationFormInitialValues } from "Apps/Consign/Routes/SubmissionFlow/Utils/formHelpers"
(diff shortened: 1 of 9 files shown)


nickskalkin (approved):
Oh, interesting! I actually tried to make other thing work - switch between tabs when clicking on them. Stepper component supports it, but here is a `disableNavigation` option enabled in [SubmissionStepper](https://github.com/artsy/force/blob/main/src/Apps/Consign/Components/SubmissionStepper.tsx#L38). When disabling this option it still doesn't work though, and to make it work we have put each step inside this stepper component. I didn't want to spend so much time on this, especially given that we are likely going to change the entire flow soon.

olerichter00:
@nickskalkin I agree with closing this ticket. Fixing the back button should already provide the functionality of going back in the form.

PR #13722, opened 2024-04-11
Chore: [EMI-1804] show unavailable banner for unavailable partner offer

erikdstock:

This PR solves [EMI-1804]

### Description

This PR adds a new condition for showing the unavailable banner when a user is expecting a partner offer.

[EMI-1804]: https://artsyproduct.atlassian.net/browse/EMI-1804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtworkPageBanner.tsx
```diff
@@ -35,26 +35,31 @@ export const ArtworkPageBanner: FC<ArtworkPageBannerProps> = props => {
...
+    !!match?.location?.query?.unavailable
+
   // show expired if query param says to expect one and it's not found
-  const expectedPartnerOfferNotFound =
-    partnerOfferVisibilityEnabled &&
-    expectedPartnerOfferID &&
-    (!partnerOffer || partnerOffer.internalID !== expectedPartnerOfferID)
+  // or imperatively if the expired_offer query param is present
+  const expectedPartnerOfferNotFound = !!(
+    !showUnavailableArtworkBanner &&
```
erikdstock, line 55: I added this to make the banners a little more mutually exclusive. Technically more than one can be shown but I don't think they stack. There could be some argument for banners triggered directly by a query param (ie `expired_offer=true` to take precedence over one that happens incidentally (`expectedPartnerOfferID && !artwork.isPurchasable`).

damassi: I wish there was a way to make this logic a bit more clear. There's double negations and nots and ors and parenthesis and nullish checks. Its too hard to understand at a glance, but if it is what it is, then all good.

damassi: I just worry that future devs will come upon this line and adding even the slightest change, might introduce a bug, or fail to understand a condition, etc. Its all good, but wanted to flag.

erikdstock: Ah missed your comment @damassi - yes I agree. I'd rather it be one long if/else that returns an enum | null for the banner type, or an array of them if we could want to show a stack of multiple banners (there's also email verification and IDK what else...). It's a little bit too complicated for what should be a simple component.

damassi: If you wouldn't mind giving it a another try to make clearer that would be appreciated, before merge 🙏

oxaudo: If we think that the scenarios are mutually exclusive - I usually find early returns helping readability quite a bit.

damassi: Agree. Early returns are the best things that one can do for code readability through time. Don't even need if / else. Just a bunch of concrete returns and / or exits.

erikdstock: Updated in https://github.com/artsy/force/pull/13722/commits/c8664a11957a162584817777b540c2437d1aa1fe - does this look ok?

damassi: Much better @erikdstock 🚢

damassi (approved):

PR #13717, opened 2024-04-11
fix: remove minHeight so that Partner Offers are fixed

Mitch-Henson:
The type of this PR is: **FIX**

This PR partially solves [AMBER-550]

### Description

The previous PR for 550: https://github.com/artsy/force/pull/13685 introduced a visual bug for Partner Offers. This PR solves that bug by removing the `minHeight` which will cause a currently unseen issue with partner artworks.

Reverting the offending PR would be the best choice, but the revert is not very clean due to the number of files changed so this is the next best option. We can fix forward on the private artworks rail issue.

Before (Rail looking good, but Partner Offers not looking good):
Screenshot 2024-04-11 at 11 05 49
Screenshot 2024-04-11 at 11 05 54

After (Rail not looking good, but Partner Offers fixed):
Screenshot 2024-04-11 at 10 57 37

Screenshot 2024-04-11 at 11 05 29

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AMBER-550]: https://artsyproduct.atlassian.net/browse/AMBER-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13727, opened 2024-04-11
feat(private-artworks): Add `<ExclusiveAccessBadge />` to carousel shelf, Artwork Notifications and Grid

damassi:

### Description

Follow-up to https://github.com/artsy/force/pull/13717 which moves the 'Exclusive Access' tag into a badge overlay on the artwork, versus in the artwork metadata details section. This solves a lot of issues with adding a 5th (rare) line item to the artwork brick. 

Screenshot 2024-04-11 at 2 27 30 PM

Screenshot 2024-04-12 at 10 53 42 AM

Screenshot 2024-04-12 at 11 15 56 AM

cc @artsy/amber-devs


Mitch-Henson (dismissed):
This looks great and solves the Artworks rail problem! I'm approving because we are in a better state than we were before and the rail looks beautiful again.

I'm holding off on merging, however, as it means that all other individual usages of `<Metadata>` outside of a rail will not have the "exclusive access" text and it will need to be manually added which I anticipate will cause issues in the future. Maybe there is another option where we can still update the Metadata / the lower level artwork component itself to ensure the most coverage with this text? Private artwork signage should ideally be visible throughout the site.

damassi:
@Mitch-Henson - Brought this up again trying to come up with a solution, and yah, the badge is the best way. At least we've now got a reusable component!

src/Components/Artwork/ShelfArtwork.tsx
```diff
@@ -76,7 +77,12 @@ const ShelfArtwork: React.FC<ShelfArtworkProps> = ({
         overflow="hidden"
         {...rest}
       >
-        <Box height={maxImageHeight} display="flex" alignItems="flex-end">
+        <Box
+          height={maxImageHeight}
+          display="flex"
+          alignItems="flex-end"
+          position="relative"
```
damassi, line 84: Also note that in order for this to be positioned correctly, the parent needs to be positioned relatively!

PR #13736, opened 2024-04-12
refactor(private-artworks): Update private artwork page design 

damassi:
Addresses [AMBER-619]

### Description

This updates the designs of the private artwork page to match our latest comps. 

https://github.com/artsy/force/assets/236943/39497468-11f0-4e59-8af4-1297620a0b7d

cc @artsy/amber-devs 

[AMBER-619]: https://artsyproduct.atlassian.net/browse/AMBER-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/PrivateArtwork/PrivateArtworkAboutArtist.tsx
```diff
@@ -37,48 +65,101 @@ export const PrivateArtworkAboutArtist: React.FC<PrivateArtworkAboutArtistProps>
...
+  const followsCount =
+    Number(formatFollowerCount(data.artist?.counts?.follows)) ?? 0
 
-  if (!data.displayArtistBio) {
-    return null
-  }
+  const biographyBlurb =
+    data.artist?.partnerBiographyBlurb?.text ??
+    data.artist?.biographyBlurb?.text ??
+    ""
```
damassi, line 74: Assuming private artworks will take greater advantage of the partner bio blurb in CMS, so prioritize that

lidimayra (approved):
Looks great!!!

src/Apps/Artwork/Components/PrivateArtwork/PrivateArtworkAboutArtist.tsx
Mitch-Henson: Unless I'm misunderstanding the code, I don't know if it is this simple? There are multiple choices here, firstly starting with whether the artist bio should be shown at all or not (`displayArtistBio`)
Screenshot 2024-04-17 at 14 58 50

Then there is a second choice which is either use the Artsy provided Artist bio, or show a partner provided Artist Bio (`biographyBlurb` vs `partnerBiographyBlurb`)
Screenshot 2024-04-17 at 14 55 48

Mitch-Henson (commented):
It looks great, though I did have a comment about biography and the levels of precedence and toggles.

damassi:
@Mitch-Henson - not too sure. We should experiment with some different configurations. 

Right now we can fairly safely assume that a private artwork will want to show an artist bio. We default to the partner supplied one first, then if that doesn't exist (or is turned off), we fallback to the artsy one (if it is turned on). If nothing exists, it is falsy, and the section remains invisible. 

Lets be sure to follow up in QA!

PR #13739, opened 2024-04-15
fix [EMI-1820]: do not add purchase button for inactive offers

oxaudo:
Believe it should be good to rely on offer 'isActive' as it seems to be updated properly when the works become unavailable.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
@@ -336,7 +336,7 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
     makeOffer?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
     contactGallery?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
   } = {}
-  if (artwork.isAcquireable || activePartnerOffer) {
+  if (artwork.isAcquireable || activePartnerOffer?.isAvailable) {
     renderButtons.buyNow = "primaryBlack"
   }
   if (artwork.isOfferable && !(activePartnerOffer && artwork.isInquireable)) {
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -336,7 +336,10 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
     makeOffer?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
     contactGallery?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
   } = {}
-  if (artwork.isAcquireable || activePartnerOffer) {
+  if (
+    artwork.isAcquireable ||
+    (activePartnerOffer && activePartnerOffer.isAvailable)
```
erikdstock, line 341: My only concern here is that our definitions align with how we test for showing the unavailable banner when an artwork page expects a partner offer: See [these lines from 13722](https://github.com/artsy/force/pull/13722/files#diff-8dfda62ff39e008f30d73abc8aaa0ee90fbfe4edfb36092f6c36bcb1bff3c745R66-R74) (uses `artwork.isPurchaseable`) and [this slack thread](https://artsy.slack.com/archives/C02JHHHKP5K/p1712782539452009) for some work I did there:
> isAcquirable is for buy now
isPurchasable is inventory + availability == for_sale + merchant account configured + price listed in supported commerce currency
isSold is availability == sold

☝️ These all refer to the artwork. Meanwhile `PartnerOffer#available` is `partnerOffer.artwork.purchasable` so it seems good (more names aside).

erikdstock (dismissed):

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
erikdstock: Could probably be shortened to `activePartnerOffer?.isAvailable`

oxaudo:
Merging this.

PR #13751, opened 2024-04-17
feat(collector-profile): implements "search"

dzucconi:
Closes [DIA-569](https://artsyproduct.atlassian.net/browse/DIA-569)

So this more or less works but obviously it's non ideal. When you start the search we just fetch the first 100 and filter from that, because we don't support search on this collection of items.

https://github.com/artsy/force/assets/112297/fb750200-0404-491c-a402-2d428c1f2bd3

[DIA-569]: https://artsyproduct.atlassian.net/browse/DIA-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsSearchResults.tsx
+++ b/src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsSearchResults.tsx
@@ -1,0 +1,90 @@
+import { FC, Suspense, useMemo } from "react"
+import { graphql, useLazyLoadQuery } from "react-relay"
+import { CollectorProfileArtistsSearchResultsQuery } from "__generated__/CollectorProfileArtistsSearchResultsQuery.graphql"
+import { compact } from "lodash"
+import { Message } from "@artsy/palette"
+import {
+  CollectorProfileArtistsListArtist,
+  CollectorProfileArtistsListArtistSkeleton,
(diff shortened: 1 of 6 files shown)


src/Apps/CollectorProfile/Components/CollectorProfileArtists/CollectorProfileArtistsSearchResults.tsx
```diff
@@ -0,0 +1,90 @@
...
+  )
+}
+
+interface CollectorProfileArtistsSearchResultsArtistsProps
+  extends CollectorProfileArtistsSearchResultsProps {}
+
+const CollectorProfileArtistsSearchResultsArtists: FC<CollectorProfileArtistsSearchResultsArtistsProps> = ({
+  term,
+}) => {
+  const { me } = useLazyLoadQuery<CollectorProfileArtistsSearchResultsQuery>(
```
damassi, line 31: Can we use yr client query hook so that we can avoid triggering suspense boundaries?

dzucconi: I tried that first. It lead to an impossible to debug problem where sometimes parts of the response were null.

damassi: ahh dang, all good 👍

PR #13769, opened 2024-04-19
fix: early exit from useEffect that checks for Private Artwork

ansor4:

### Description

@joeyAghion reported in [this slack thread](https://artsy.slack.com/archives/C05F8TNKGAV/p1713558385846229) that the error rate was jumping. Turns out this relay request will 500 when not on an artwork page. 

This PR adds logic to early exit from the `useEffect` that makes the relay request, only on non-artwork pages.


damassi (approved):
Looks good! ty for the fix @ansor4 👍

joeyAghion:
🙏

PR #13774, opened 2024-04-22
fix: Suppress duplicate edition set info on private artwork pages AMBER-629

jonallured:
This PR ensures that we do not double render edition set information for private artworks.

<details><summary>screenshots below!</summary>

Screenshot 2024-04-22 at 11 22 18 AM
Screenshot 2024-04-22 at 11 22 12 AM
Screenshot 2024-04-22 at 11 22 25 AM
Screenshot 2024-04-22 at 11 22 30 AM

</details> 

My approach was to start by extracting a boolean for whether the work is editioned and also extracting the edition set info into it's own component. These moves seemed helpful so that we could operate at the same level of abstraction - one component for editioned works, another for non-editioned works. Boolean flag to render one or the other.

Then the actual fix was to wrap that with a check for the flag to show price info. When false bail on the rendering whether editioned or not.

https://artsyproduct.atlassian.net/browse/AMBER-629

/cc @artsy/amber-devs

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
@@ -381,11 +381,13 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
     )
   }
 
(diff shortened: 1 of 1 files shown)


damassi (approved):
Thank you @jonallured 👍

PR #13778, opened 2024-04-22
chore(private-artworks): Update tracking calls a bit

damassi:

### Description

This is some minor convention-ish updates to https://github.com/artsy/force/pull/13773. Main changes are:
- make the tracking analytics payload more local in scope, vs "global" to the component. Move `payload` to where it is being called, since it doesn't need to be anywhere else (for sharing purposes)
- Add specific interface types where it makes sense, and is obvious


Mitch-Henson:
Thanks for the updates! 🙏

PR #13788, opened 2024-04-24
chore: AMBER-637 - NoIndex Artwork ID URL

Mitch-Henson:

This PR solves [AMBER-637]

### Description

All artworks are available to be directly browsed to. In this scenario, we want to ensure that there is a no-index flag as we don't want this page to affect our SEO. Only the nice, slugified URL should be indexed. This scenario can become more apparent if a previously private artwork URL is shared and then the artwork becomes public (and therefore indexable)

<details>
<summary>Before</summary>
Previously private artwork accessed via artwork ID URL does not have noindex

Screenshot 2024-04-24 at 17 12 34
</details>

<details>
<summary>After</summary>
Previously private artwork accessed via artwork ID URL has noindex

Screenshot 2024-04-24 at 17 07 49

Previously private artwork accessed via slugified URL does not have noindex

Screenshot 2024-04-24 at 17 08 26
</details>

[AMBER-637]: https://artsyproduct.atlassian.net/browse/AMBER-637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -247,7 +247,10 @@ export const ArtworkApp: React.FC<Props> = props => {
       <UseRecordArtworkView />
       <ArtworkPageBanner artwork={artwork} me={me} />
 
-      <ArtworkMetaFragmentContainer artwork={artwork} />
+      <ArtworkMetaFragmentContainer
+        artwork={artwork}
+        pathname={props.match.location.pathname}
```
damassi, line 252: We don't need to pass props! we can use our router hook. See below

Mitch-Henson:
Amazing! Thanks for modernising this. It is also really helpful to see each step broken out into its own commit. Will definitely be using this as a template should I run into more class based components 🙏

PR #13797, opened 2024-04-24
Fix: Adjust relay to query for partner provided biography and display if p…

jpotts244:
…resent

The type of this PR is: Fix

### Description

Fix for issue discovered [here](https://artsy.slack.com/archives/C05F8TNKGAV/p1713989447829839)
`biographyBlurb` was always returning the artsy provided bio as it has a constraint that the partner artist must be `featured` to show the partner provided bio. We need a little more flexibility for the private artworks flows

Associated MP changes: https://github.com/artsy/metaphysics/pull/5693

Bug (Note the same data is returned for `biographyBlurb(format: HTML, partnerBio: false/true)` 👍🏻 
Screenshot 2024-04-24 at 4 26 10 PM

---

Fix (one provides the partner bio and the other is the artsy bio):
Screenshot 2024-04-24 at 6 36 57 PM


src/Apps/Artwork/Components/PrivateArtwork/PrivateArtworkAboutArtist.tsx
```diff
@@ -48,13 +48,7 @@ export const PrivateArtworkAboutArtist: React.FC<PrivateArtworkAboutArtistProps>
...
           }
           name
-          partnerArtists {
-            biography
-          }
-          partnerBiographyBlurb: biographyBlurb(
-            format: HTML
-            partnerBio: true
-          ) {
+          partnerBiographyBlurb {
```
damassi, line 51: @jpotts244 - don't we need to update the code somewhere (vs just the graphql query?)

damassi: ahhhh I see 😄

damassi (approved):
Looks good! field name is the same as the aliased field, so 👌

src/Apps/Artwork/Components/PrivateArtwork/PrivateArtworkAboutArtist.tsx
jpotts244: Dont think so!! I think yours was the same naming, nice clean diff ✅

PR #13803, opened 2024-04-25
fix(artwork): only show partner bio if enabled

damassi:

### Description

This hides the private artwork partner bio unless `displayArtistBio` is true. 

cc @artsy/amber-devs


jpotts244 (approved):
Much better

Mitch-Henson (approved):

PR #13809, opened 2024-04-25
fix(private-artworks): Defer to artsy bio

damassi:

### Description

Only show artsy bio, and make sure to still show rest of artist section if bio is disabled.

cc @artsy/amber-devs


jpotts244 (approved):
Thanks!

PR #13810, opened 2024-04-25
feat: set up experiments directory for art advisory

anandaroop:
Two things:

- Sets up a folder structure similar to the one in artsy/quantum for adding numbered experiments (and moves the initial spike into experiment no. 1)

    

- Adds experiment no. 2 which demonstrates a few UI niceties to help makes this a little more presentable and chatgpt-esque for internal testing

### DIFF
--- a/src/Apps/ArtAdvisor/02-Markdown/App.tsx
+++ b/src/Apps/ArtAdvisor/02-Markdown/App.tsx
@@ -1,0 +1,152 @@
+import { FC, useEffect, useRef, useState } from "react"
+import { Text, Box, Input, Button, Flex } from "@artsy/palette"
+import { ChatCompletionStream } from "openai/lib/ChatCompletionStream"
+import Markdown from "marked-react"
+
+export const App: FC = () => {
+  const inputRef = useRef<HTMLInputElement>(null)
+  const conversationRef = useRef(null)
+  const scrollMeRef = useRef<HTMLDivElement>(null)
+  const [markdownText, setMarkdownText] = useState("")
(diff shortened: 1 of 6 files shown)


src/Apps/ArtAdvisor/02-Markdown/App.tsx
```diff
@@ -0,0 +1,152 @@
...
+      styleElement.sheet.cssRules.length
+    )
+    styleElement.sheet.insertRule(
+      "td, th { border: 1px solid gray; padding: 0.5em; }",
+      styleElement.sheet.cssRules.length
+    )
+  }
+
+  return styleElement
+}
```
damassi, line 148: So we don't get into any funny states later, whats this about?

anandaroop: Yeah I injected some styles for our very _un_-styled lists so that the markdown would look ok.

Then I use the return value here to discard these styles when the page unmounts (via useEffect), assuming this is the good-citizen thing to do given client side nav.

I feel like there’s a better way to inject a few ad hoc style rules, which we can pursue as an improvement here.

damassi: Could we do something like this styled-components setup? 

```tsx
const Container = styled(Box)`
  ul: ... 
`

const App = () => {
  return (
    <Container>
      ...
    </Container>
  )
}

anandaroop: Good call

PR #13812, opened 2024-04-26
feat(onyx-866): tooltip in My Collection

nickskalkin:

This PR solves [ONYX-866]

### Description

Add "Interested in Selling" tooltip which points to the first P1 artwork within my collection artworks grid:

image

I **really** don't like that for this task I have to make changes in `ArtworkGrid` and `GridItem` components - but I don't know how to avoid doing so without fully rewriting `MyCollectionArtworkGrid`. `ArtworkGrid` has logic to properly handle artworks layout, pagination logic, etc - this all needs to be copied to `MyCollectionArtworkGrid` if we want to not pollute generic `ArtworkGrid` component with MyCollection-related things.

### DIFF
--- a/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionArtworkGrid.tsx
+++ b/src/Apps/Settings/Routes/MyCollection/Components/MyCollectionArtworkGrid.tsx
@@ -1,22 +1,77 @@
+import { useDismissibleContext } from "@artsy/dismissible"
 import ArtworkGrid from "Components/ArtworkGrid/ArtworkGrid"
+import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
+import { Text } from "@artsy/palette"
+import { PROGRESSIVE_ONBOARDING } from "Components/ProgressiveOnboarding/progressiveOnboardingKeys"
+import { _FragmentRefs } from "relay-runtime"
 
-export const MyCollectionArtworkGrid = createFragmentContainer(ArtworkGrid, {
-  artworks: graphql`
-    fragment MyCollectionArtworkGrid_artworks on MyCollectionConnection {
-      edges {
-        node {
-          id
-          slug
-          href
-          internalID
-          image(includeAll: true) {
-            aspectRatio
+const POPOVER_KEY = PROGRESSIVE_ONBOARDING.startSelling
+
+interface MyCollectionArtworkGridProps {
+  artworks: _FragmentRefs<"MyCollectionArtworkGrid_artworks">
+  onLoadMore: () => void
+}
+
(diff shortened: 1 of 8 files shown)


src/Apps/Settings/Routes/MyCollection/Components/MyCollectionArtworkGrid.tsx
```diff
@@ -1,22 +1,77 @@
...
-          id
-          slug
-          href
-          internalID
-          image(includeAll: true) {
-            aspectRatio
+const POPOVER_KEY = PROGRESSIVE_ONBOARDING.startSelling
+
+interface MyCollectionArtworkGridProps {
+  artworks: _FragmentRefs<"MyCollectionArtworkGrid_artworks">
```
olerichter00, line 12: What is `_FragmentRefs`? I have never used it before. And would `MyCollectionArtworkGrid_artworks$data` work here as well? 👀

olerichter00 (approved):
Looks good 🌟

src/Apps/Settings/Routes/MyCollection/Components/MyCollectionArtworkGrid.tsx
damassi: And in any case, we should (always) be using `useFragment` here vs our old school `createFragmentContainer`

nickskalkin: @olerichter00  I've never used it before either, but without it get this typescript message:

image

olerichter00: Ah, it should work if you use [ArtworkGridContainer](https://github.com/artsy/force/blob/fdc195db92ebcad255819f5c611dac309de070e1/src/Components/ArtworkGrid/ArtworkGrid.tsx#L65) instead of `ArtworkGrid`. in line [25](https://github.com/artsy/force/pull/13812/files#diff-a57b784f4ea583c258b8362e8e383903c4b5decfd9da4f5793f9300ff31aceefR25) because `ArtworkGrid` doesn't like `MyCollectionConnection` so instead of wrapping the artwork grid fragment container, [MyCollectionArtworkGrid.tsx](https://github.com/artsy/force/pull/13812/files#diff-a57b784f4ea583c258b8362e8e383903c4b5decfd9da4f5793f9300ff31aceef) needs to wrap the artwork grid itself.

nickskalkin: oh, that works, thanks @olerichter00!

olerichter00 (approved):

PR #13816, opened 2024-04-26
feat(admin): Add admin/navigate-to-route route

damassi:

### Description

Inspired by a slack thread around needing to see request info for routes that we can only access via a direct link (and thus are SSR'd, and hard to get relay logging info), this adds a new admin app / route `/admin/navigate-to-route` which allows us to quickly jump anywhere on the site:

Screenshot 2024-04-26 at 10 43 37 AM

### DIFF
--- a/src/Apps/Admin/NavigateToRoute.tsx
+++ b/src/Apps/Admin/NavigateToRoute.tsx
@@ -1,0 +1,70 @@
+import {
+  Button,
+  Column,
+  Flex,
+  GridColumns,
+  Input,
+  Spacer,
+  Text,
+} from "@artsy/palette"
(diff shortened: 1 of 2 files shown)


Mitch-Henson:
@jonallured this comes off the back of [this slack conversation](https://artsy.slack.com/archives/C05F8TNKGAV/p1714141872668059). Essentially, we can't easily log or track server side requests made to MP when SSR is on. This page allows us to easily navigate to all routes, but essentially forcing it to be a client transition, meaning that we can track all external requests in the network tab.

Happy for there to be more documentation, I guess both on the page itself and maybe in the docs somewhere?

damassi:
Yep, what @Mitch-Henson said!

PR #13817, opened 2024-04-26
refactor(router): Upgrade found router libs to latest; fix client-side caching post artwork page view

damassi:

### Description

This upgrades our router-related libs to latest. In the found repo it looks like a major reorganization took place, with little code changes. But this impacted typescript types. 

We used to use patch package to patch some things here, but deleted that file and went the `found.d.ts` typescript override approach. 

Functionally, things _seem_ fine. 

This fixes our issue with broken page caching, in the sense that we could no longer force fetch a page (like artwork pages) on each visit, since cacheConfig was no longer respected after the relay upgrade.

Also fixes issue where client-side cache is destroyed after every artwork page visit. Now we create a new network instance to call the `useRecordArtworkView` mutation, vs using the global env and subjecting it to a cache clear.


src/Apps/Artwork/artworkRoutes.tsx
```diff
@@ -62,8 +62,13 @@ export const artworkRoutes: AppRouteConfig[] = [
         }
       }
     `,
-    cacheConfig: {
-      force: true,
+    getCacheConfig: ({ context }) => {
```
damassi, line 65: The ability to invoke this is new with the latest upgrade. We can do lots of things with this, and lots of route data comes back in the arguments

PR #13820, opened 2024-04-29
feat: collector profile/settings re-organization

dzucconi:
The following routes have changed:

Before | After
-- | --
settings/alerts | favorites/alerts
collector-profile/saves | favorites/saves
collector-profile/saves/:id | favorites/saves/:id
collector-profile/follows | favorites/follows
settings/alerts | favorites/alerts
settings/alerts/:id/edit | favorites/alerts/:id/edit
settings/alerts/:id/artworks | favorites/alerts/:id/artworks

This also launches the new`/collector-profile/artists` route on web.

------

Review app: https://collector-profile.artsy.net/


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c476829477d4e2bb0efa70d9a452c833b1170ab0
</p>

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/ArtworkAuctionCreateAlertHeader.tsx
```diff
@@ -144,7 +144,7 @@ const ArtworkAuctionCreateAlertHeader: FC<ArtworkAuctionCreateAlertHeaderProps>
                 Icon={BellStrokeIcon}
                 // @ts-ignore
                 as={RouterLink}
-                to={"/settings/alerts"}
+                to={"/favorites/alerts"}
```
MounirDhahri, line 147: this makes sense to me more than the `settings/alerts` - I would only suggest ensuring we align that with Eigen here and marketing to not break any deep links logic. cc @dblandin 

https://github.com/artsy/eigen/blob/9c5798f08c98d86739d7477006b9cdea2beb044e/src/app/routes.ts#L254

dzucconi: There would be redirects so it shouldn't be an issue.

src/Components/NavBar/Menus/NavBarUserMenuAvatar.tsx
```diff
@@ -0,0 +1,50 @@
+import { Avatar, Skeleton } from "@artsy/palette"
+import { FC } from "react"
+import { graphql, useLazyLoadQuery } from "react-relay"
+import { NavBarUserMenuAvatarQuery } from "__generated__/NavBarUserMenuAvatarQuery.graphql"
+
+interface NavBarUserMenuAvatarProps {}
+
+export const NavBarUserMenuAvatar: FC<NavBarUserMenuAvatarProps> = () => {
+  const { me } = useLazyLoadQuery<NavBarUserMenuAvatarQuery>(
```
damassi, line 9: Why use this over `useClientQuery`? (Maybe I already asked this elsewhere, forgive)

dzucconi: Caching doesn't work correctly with `useClientQuery`

src/System/Router/Boot.tsx
```diff
@@ -123,6 +125,19 @@ export const Boot = track(undefined, {
   )
 })
 
+const EnvironmentProvider: FC<{ environment: Environment }> = ({
+  children,
+  environment,
+}) => {
+  if (process.env.NODE_ENV === "test") return <>{children}</>
```
damassi, line 132: Why not provide the environment in tests? Is it because we wrap the setupTestWrapper util in an environment provider?

dzucconi: It _should_ already have been provided by `MockBoot` — I couldn't really make sense of it and it didn't seem to matter much since nothing really needed it. It's possible that as we move over to relay-hooks we'll have to address though.

src/Components/NavBar/Menus/NavBarUserMenuAvatar.tsx
damassi: oh that's good to know -- is that a hypothesis, or just going off of the fetchQuery docs?

src/System/Router/Boot.tsx
damassi: Yah we should avoid this if possible since we're rapidly moving to relay hooks! Mind dropping it into mock boot? It seems pretty harmless. I can imagine this will catch some unsuspecting dev later

dzucconi: I'll make a ticket to take a closer look; but that's the thing: `MockBoot` _is_ `Boot` — it should have just worked.

src/Components/NavBar/Menus/NavBarUserMenuAvatar.tsx
dzucconi: That, I don't know. I didn't notice anything in the docs. Seems like it might be a bug; but also in general I don't mind using `useLazyLoadQuery` so it seems like we should move over anyway.

src/System/Router/Boot.tsx
damassi: ahhh i see. Interesting, yah. Ty 👍

PR #13823, opened 2024-04-29
fix: EMI-1806 partner offer timer re-mounts over and over

erikdstock:
Fixes [EMI-1806]- or improves it at least.

### Description

_See linked ticket for bug_

This PR improves some visual bugs with the partner offer countdown timer by removing a component that is defined in the body of another component. As a result of doing this the `useTimer()` hook was repeatedly re-rendering, causing the internal hook to sync with our official server time to continually cancel itself out.

However, I think there may be something more to improve here. I wasn't able to eliminate 100% of the visual tics and there is still some flickering in the first instant the component loads. This might deserve more investigation.

[EMI-1806]: https://artsyproduct.atlassian.net/browse/EMI-1806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
@@ -357,47 +357,6 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
         : "primaryBlack"
   }
 
-  const SaleMessageOrOfferDisplay: FC = () => {
-    return (
-      <>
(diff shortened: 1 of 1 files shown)


fladson (approved):
LGTM

I would just create another ticket to address your observation.
>However, I think there may be something more to improve here. I wasn't able to eliminate 100% of the visual tics and there is still some flickering in the first instant the component loads. This might deserve more investigation.

erikdstock:
> I would just create another ticket to address your observation.

Good idea.

PR #13835, opened 2024-05-01
feat: add backend support for LLM demos

anandaroop:
This moves into Force the backend endpoints which send requests to the LLM and return the results to the clients. 

This allows us to avoid relying on a running Quantum server, and thus simplifies the deployment story (one app instead of two).

In this scenario Quantum can remain a place for one-off experiments that are focused more on LLM logic and don't need a UI, whereas any interactive prototypes would be built wholly within Force, following the patterns laid down here.

(Feature creeped with a few frontend commits for the purpose of the review app — happy to separate this into two PRs)


src/Apps/ArtAdvisor/01-Spike/server.ts
```diff
@@ -0,0 +1,15 @@
...
+  res.send("Hello root under experiment 1")
+})
+
+router.get("/a", (req, res) => {
+  res.send("Hello experiment 1a")
+})
+
+router.get("/b", (req, res) => {
+  res.send("Hello experiment 1b")
+})
```
mc-jones, line 15: Do we need to change the `fetch` request in the related `App.tsx` ? Maybe doesn't matter, as `02-Markdown` is just an improved version?

anandaroop: 01 still has the placeholder routing. 

I focused on 03 for the demo, and then was working backwards: wired up 02, but then I wasn't sure which Quantum backend 01 was paired with.

mc-jones: Got it. I think that is fine. IIRC it was originally for the `03-express-streaming` app, but it was changed to test some of the other approaches. Don't think we will need to go back to this, so probably fine to leave as is.

mc-jones (approved):
Looks good! Glad to see that utilizing the force server didn't seem too painful.

src/Apps/ArtAdvisor/01-Spike/server.ts
anandaroop: I timeboxed trying to pull that quantum endpoint into here, but it wasn't working. Since we agreed we were fine if demo 01 wasn't wired up, I marked it as such on the landing page.

anandaroop:
I took this out of draft and threw a `squash on green` on here. 

@mc-jones I also went ahead and removed the inline notes on the prompts — feel free to add them back in as file comments on this PR (either before or after this merges is fine)

src/Apps/ArtAdvisor/03-Instruction-Experiments/00-prompt-lib/collector-profile-one.ts
mc-jones, line 1: Observations on initial testing:

- It always asks me to confirm I want to build a user profile. 
- Does seem to do a fairly good job when getting some basic information from the user profile.See https://artsy.slack.com/archives/C06SSV1K10D/p1714418303277419
- Works when the user doesn't have a profile...though, would this be a use case we need to support?
- When there is no user profile, it asks for all the information at one time. While we may not ever ask for information from someone who is not a artsy user, we should be aware that a lack of information may result in a question dump. 
- I got much more specific questions when using the advisor without a user profile. 
- "...with a generous budget of up to $200,000", it should not include language like "generous". This does seem to be solved, when it was asked to change to the first person, though.
- Super awkward: "I collect art both for investment purposes and for aesthetic enjoyment, which reflects my appreciation for both the intrinsic and financial value of art."

src/Apps/ArtAdvisor/03-Instruction-Experiments/00-prompt-lib/collector-profile-two.ts
mc-jones, line 1: - Its trying to make assumptions about time periods of collecting that aren't provided by the user, we should watch out for these kinds of things.
- It will try to complete the steps without hitting all points (e.g., it didn't ask me about my budget)

damassi:
Thank you!

PR #13784, opened 2024-04-24
feat: Reorganize sell page mWeb

olerichter00:
Addresses [ONYX-868]

- [Figma Design](https://www.figma.com/file/VQRDUwdkTUslF62khMhoGZ/O2-Collector-Supply-%E2%80%93-Quick-Tasks?type=design&node-id=9-11422&mode=design&t=NX2linVmGMHANOmj-0)

## Description

This reorganizes `/sell` page (mainly mWeb) according to the new design.

- Copy updates (desktop & mWeb)
- Sticky footer with CTA and text (mWeb)
- section swap (mWeb)

| | |
| --- | --- |
| Screenshot 2024-04-25 at 11 25 54| Screenshot 2024-04-25 at 11 26 18|

[ONYX-868]: https://artsyproduct.atlassian.net/browse/ONYX-868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13836, opened 2024-05-02
chore: cleanup sell routes

dariakoko:

This PR solves [ONYX-947]

### Description
Followup on [the comment](https://github.com/artsy/force/pull/13784#discussion_r1579308469)

[ONYX-947]: https://artsyproduct.atlassian.net/browse/ONYX-947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d2ffe48d3bf719ab55c542d05f3548a333411d36
</p>

src/Apps/Consign/consignFromCollectorProfileMyCollectionRoutes.tsx
```diff
@@ -5,26 +5,6 @@ import { AppRouteConfig } from "System/Router/Route"
 import { getENV } from "Utils/getENV"
 import { getRedirect } from "./Routes/SubmissionFlow/Utils/redirects"
 
-const MarketingLandingApp = loadable(
```
olerichter00, line 8: I wonder if all the remaining routes within this file should be moved to `src/Apps/CollectorProfile/collectorProfileRoutes.tsx` because they all start with `/collector-profile/`... 🤔

olerichter00 (commented):
Thanks a lot for cleaning up these routes 🌟

src/Apps/Consign/consignFromCollectorProfileMyCollectionRoutes.tsx
dariakoko: Oh wow, we have a lot of routes 😅 I think yes, we can move them to collectorProfileRoutes. I'll do it and test

src/Apps/CollectorProfile/collectorProfileRoutes.tsx
```diff
@@ -292,4 +390,147 @@ export const collectorProfileRoutes: AppRouteConfig[] = [
       PriceEstimateConfirmation.preload()
     },
   },
+  {
+    path: "/collector-profile/my-collection/submission",
```
MounirDhahri, line 394: I am not a big fan of having a new submission route here - I understand we did that in the past but I think this has led to issues with data since they will have to take into account the new route and if we deep link into this route, it will break in app. Would it be possible to explore ways where we only have one sell route regardless from where you start your submission?

I understand this might not be a direct thing to pull here but I would suggest spiking on this given that we are going to be touching this surface in a few weeks time when we port the new submission flow to web. You can potentially ticket this and bring it to be discussed with the team over standup or refinement

src/routes.tsx
```diff
@@ -83,8 +81,6 @@ export const getAppRoutes = (): AppRouteConfig[] => {
     { routes: collectRoutes },
     { routes: collectorProfileRoutes },
     { routes: consignRoutes },
-    { routes: consignFromMyCollectionRoutes },
-    { routes: consignFromCollectorProfileMyCollectionRoutes },
```
MounirDhahri, line 87: nice!

src/routes.tsx
```diff
@@ -83,8 +81,6 @@ export const getAppRoutes = (): AppRouteConfig[] => {
     { routes: collectRoutes },
     { routes: collectorProfileRoutes },
```
MounirDhahri, line 82: I think if we also delete the consign routes from here - we would be at a great position

src/Apps/CollectorProfile/collectorProfileRoutes.tsx
MounirDhahri: You can as well do it in this PR since I see you already removed the 2 out the 4 routes that get you to consign. Only 1 left to remove

src/routes.tsx
dariakoko: Do you mean consignRoutes on line 83? It's actually /sell routes. I can rename it
Also, we still have /consign route that redirects to /sell -- I think we can already remove it. What do you think?

MounirDhahri: I actually meant to delete the sell routes from the collector profile and have all of them all within the sellRoutes/consignRoutes. This way there would be only one route for submissions

src/Apps/CollectorProfile/collectorProfileRoutes.tsx
dariakoko: That's a good point. We can definitely use /sell instead of /submission here, we just need to talk to the data team first. But I'm not sure if removing /collector-profile/my-collection/ is a good idea if this is what you mean. Having /collector-profile/my-collection/ maintains a cohesive experience for my collection (compare to add/edit artwork) and back button behaviour. 
Please, let me know if I misunderstood you

dariakoko: Actually having only /sell/submission/SUBMISSION-STEP/{artworkId} would make total sense.Is that what you mean?

olerichter00 (approved):

dariakoko:
@MounirDhahri hi! I'll go ahead and merge this one

PR #13837, opened 2024-05-02
feat(EMI-1524): shipping refactor scroll to errors

rquartararo:

This PR solves [EMI-1524]

@artsy/emerald-devs 

### Description

The shipping refactor adds Formik to the shipping form. This PR adds the utility `scrollToFormikErrors` to scroll to Formik field errors for a better mobile experience. 

https://github.com/artsy/force/assets/50849237/1b529207-5315-4f6d-acf6-79871aa502a6

[EMI-1524]: https://artsyproduct.atlassian.net/browse/EMI-1524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Utils/scrollToFormikErrors.tsx
```diff
@@ -0,0 +1,33 @@
...
+  if (Object.keys(errors).length === 0) return
+
+  const prefix = Object.keys(errors)[0]
+
+  const field = prefix && Object.keys(errors[prefix])[0]
+
+  return `${prefix}.${field}`
+}
+
+export const ScrollToFieldError = () => {
```
damassi, line 14: I imagine this is a typo but still wanna flag :) Can we update this so that it follows more standard function naming conventions?  (`CapsCase` is typically reserved for classes, components, interfaces, types and enums, not regular functions)

damassi: Also, might want to name file name after the main named export -- so `scrollToFieldError.ts` -- and then if more formik utils are added, we can rename the file to something more general. That way, things are 1:1.  

Otherwise, looks good! If this is used more frequently we might want to consider elevating it up to the main Utils folder (or something.)

damassi (dismissed):
Two minor requests, but otherwise looks great 👍

src/Apps/Order/Utils/scrollToFormikErrors.tsx
damassi: However you think is best! (Or not at all is totally fine -- just wanted to call it out, its super minor)

damassi (approved):

PR #13842, opened 2024-05-02
chore: disable social login

mc-jones:
The type of this PR is: Chore

### Description

This PR adds a note that social login [does not work ](https://artsy.slack.com/archives/C06SSV1K10D/p1714662670281049?thread_ts=1714662167.086639&cid=C06SSV1K10D)in this review app and disables the buttons. I figured it is better to be explicit and keep them in the UI but add a note, rather than just removing them. If we decide we want to support social login, we can dig into the root issue. 


anandaroop (approved):
Thanks!

PR #13843, opened 2024-05-03
fix(settings): immediately uploads avatar; polls for processing

dzucconi:
Closes [DIA-614](https://artsyproduct.atlassian.net/browse/DIA-614)

The motivation here is that the current avatar uploader has a confusing UX wherein when you add a file to upload, you also have to hit 'Save' below the fold in order for it to persist. If you leave, it does not save, but because it stores and uses a local copy of the image, it looks like it uploaded. (I did this and also @iskounen actually thought he had an avatar on staging that wasn't being used during Q/A — so it's definitely happening). This also leads to a ton of complexity when you try to use the avatar elsewhere on the site.

This reworks the avatar uploader to upload + save immediately, removes local image storage, and simply polls for the updated avatar. If image processing is slow then this can be annoying — but there ideally should be a priority for a thumbnail to come through. (In staging I'll randomly see this in the processing state for a few minutes.)

https://github.com/artsy/force/assets/112297/848221b0-9c94-4c72-bbd1-6f09148df060

[DIA-614]: https://artsyproduct.atlassian.net/browse/DIA-614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iskounen (approved):
This is a great improvement over the existing behavior. Thank you! 🎉

araujobarret (approved):

PR #13847, opened 2024-05-03
chore: require authentication for submission

MounirDhahri:

This PR solves [ONYX-907]

### Description

- Require authentication for submission

https://github.com/artsy/force/assets/11945712/c09805e1-258c-45a6-8f35-e7eaf77689ff

[ONYX-907]: https://artsyproduct.atlassian.net/browse/ONYX-907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/AuthDialog/AuthDialogContext.tsx
+++ b/src/Components/AuthDialog/AuthDialogContext.tsx
@@ -40,6 +40,7 @@ export type AuthDialogOptions = {
   title?: string | ((mode: AuthDialogMode) => string)
   onClose?: () => void
   onSuccess?: () => void
+  disableTapToClose?: boolean
 }
 
 export type AuthDialogAnalytics = {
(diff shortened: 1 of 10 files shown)


src/Components/AuthDialog/AuthDialogContext.tsx
```diff
@@ -40,6 +40,7 @@ export type AuthDialogOptions = {
   title?: string | ((mode: AuthDialogMode) => string)
   onClose?: () => void
   onSuccess?: () => void
+  disableTapToClose?: boolean
```
MounirDhahri, line 43: not sure if this is a fine approach

dzucconi: Why would we want to do this?

dzucconi: IMO, it's annoying and you'll be able to close via escape or the close button anyway. If we want a required auth step then redirect to the full page login/sign up.

MounirDhahri: True, didn't think about keyboard events - I will tell Barney about it. Thanks @dzucconi 🙏

MounirDhahri: Done now - added a redirect to the landing page if the user is not logged in

tam-kis (approved):

PR #13858, opened 2024-05-07
fix(ONYX-946): add MyBids on Auctions page

tam-kis:
The type of this PR is: Fix

This PR solves [ONYX-946](https://artsyproduct.atlassian.net/browse/ONYX-946)

### Description

Adds MyBids shelf "Your Auctions and Bids" to Auctions page.

Screenshot 2024-05-07 at 17 22 33

Without MyBids:

Screenshot 2024-05-07 at 19 52 43

[ONYX-946]: https://artsyproduct.atlassian.net/browse/ONYX-946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Auctions/AuctionsApp.tsx
+++ b/src/Apps/Auctions/AuctionsApp.tsx
@@ -9,6 +9,7 @@ import { RouteTabs, RouteTab } from "Components/RouteTabs"
 import { getENV } from "Utils/getENV"
 import { CuritorialRailsTabBarFragmentContainer } from "./Components/CuritorialRailsTabBar"
 import ChevronRightIcon from "@artsy/icons/ChevronRightIcon"
+import { MyBidsQueryRenderer } from "Apps/Auctions/Components/MyBids/MyBids"
 
 export interface AuctionsAppProps {
   viewer: AuctionsApp_viewer$data
@@ -20,10 +21,11 @@ const AuctionsApp: React.FC<AuctionsAppProps> = props => {
   // FIXME: Remove once new filter launches
   const enableNewAuctionsFilter = getENV("ENABLE_NEW_AUCTIONS_FILTER")
 
+  const hasMyBids = viewer.me?.myBids?.active?.length
+
   return (
     <>
       <AuctionsMeta />
-
       <GridColumns mt={4}>
         <Column span={6}>
           <Text variant="xl" as="h1">
@@ -57,6 +59,10 @@ const AuctionsApp: React.FC<AuctionsAppProps> = props => {
 
       <Spacer y={4} />
 
+      <MyBidsQueryRenderer />
+
+      {hasMyBids ? <Spacer y={12} /> : ""}
+
       <CuritorialRailsTabBarFragmentContainer viewer={viewer} />
 
       <Spacer y={12} />
@@ -89,6 +95,15 @@ export const AuctionsAppFragmentContainer = createFragmentContainer(
     viewer: graphql`
       fragment AuctionsApp_viewer on Viewer {
         ...CuritorialRailsTabBar_viewer
+        me {
+          myBids {
+            active {
+              sale {
+                slug
+              }
+            }
+          }
+        }
       }
     `,
   }
--- a/src/Apps/Auctions/AuctionsApp.tsx
+++ b/src/Apps/Auctions/AuctionsApp.tsx
@@ -21,8 +21,6 @@ const AuctionsApp: React.FC<AuctionsAppProps> = props => {
   // FIXME: Remove once new filter launches
   const enableNewAuctionsFilter = getENV("ENABLE_NEW_AUCTIONS_FILTER")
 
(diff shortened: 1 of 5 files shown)


src/Apps/Auctions/AuctionsApp.tsx
```diff
@@ -20,10 +21,11 @@ const AuctionsApp: React.FC<AuctionsAppProps> = props => {
   // FIXME: Remove once new filter launches
   const enableNewAuctionsFilter = getENV("ENABLE_NEW_AUCTIONS_FILTER")
 
+  const hasMyBids = viewer.me?.myBids?.active?.length
```
damassi, line 24: We should do: 

```tsx
const hasMyBids = !!viewer.me?.myBids?.active?.length
```
or 
```tsx
const hasMyBids = Boolean(viewer.me?.myBids?.active?.length)
```
or 
```tsx
const hasMyBids = viewer.me?.myBids?.active?.length > 0
```

The reason we need to coerce it into a boolean is because in JSX, something like 

```tsx
const foo = 0 

<div>{foo && "hi"}</div>
```
will actually emit 
```tsx
<div>0</div>
```
in the html. Numbers are totally valid outputs, while booleans will safely guard.

src/Apps/Auctions/AuctionsApp.tsx
```diff
@@ -57,6 +59,10 @@ const AuctionsApp: React.FC<AuctionsAppProps> = props => {
 
       <Spacer y={4} />
 
+      <MyBidsQueryRenderer />
+
+      {hasMyBids ? <Spacer y={12} /> : ""}
```
damassi, line 64:
```tsx
{hasMyBids && <Spacer y={12} />}
```

src/Apps/Auctions/AuctionsApp.tsx
tam-kis: Yes, this was really annoying, will update!

src/Apps/Auctions/AuctionsApp.tsx
```diff
@@ -89,6 +95,15 @@ export const AuctionsAppFragmentContainer = createFragmentContainer(
     viewer: graphql`
       fragment AuctionsApp_viewer on Viewer {
         ...CuritorialRailsTabBar_viewer
+        me {
```
damassi, line 98: Note that this is a very (very) slow query, so we should avoid adding this to the page load time just so we can emit a visual spacer. 

A better approach would be to update `MyBidsQueryRenderer` with the spacer. The query renderer loads on the client, and does not block rendering during the SSR pass.

tam-kis: Yes, thought about it but [here](https://github.com/artsy/force/blob/f13712879fce044acafcf45cb16ea132db120311/src/Apps/Home/HomeApp.tsx#L51) on the Homepage we're wrapping all of the Rails in `<Join separator={<Spacer y={[6, 12]} />}>` so in that case we would also have to rework that as well.

damassi: In that case, can do something like: 

```tsx
MyBidsQueryRenderer mt={12} />
```
and on the inside:

```tsx
interface MyBidsQueryRendererProps extends BoxProps {}

const MyBidsMyBidsQueryRenderer: React.FC<MyBidsQueryRendererProps> = ({
  mt = 0, // default arg for home 
  ...spaceProps
}) => {
  return (
    <Box mt={mt} {...spaceProps}>
      <SystemQueryRenderer ... />
    </Box>
  )
}

Comment (tam-kis) [reply src/Apps/Auctions/AuctionsApp.tsx:98]: Yes, now that you mentioned that it's very slow, would be good to avoid that query especially in this case. Will rework this!

File: src/Apps/Auctions/Components/MyBids/MyBids.tsx [hunk 4]
```diff
@@ -129,38 +130,43 @@ export const MyBidsFragmentContainer = createFragmentContainer(MyBids, {
...
 
   return (
-    <SystemQueryRenderer<MyBidsQuery>
-      lazyLoad
-      environment={relayEnvironment}
-      query={graphql`
-        query MyBidsQuery {
-          me {
-            ...MyBids_me
+    <Box mb={mb}>
```
damassi, line 144: Since we're extending `BoxProps` in the interface, might as well expand things after line 135 and keep things truthful with:
```tsx
mb = 0, 
...boxProps
```
And then fully spread that in here:
```tsx
<Box mb={mb} {...boxProps}>
```

damassi (dismissed):
One tiny req to keep interfaces and props all perfectly aligned, but once that's in LGTM! 👍

damassi (approved):
Thanks for the quick turn around @tam-kis 👌 All looks good now

PR #13864, opened 2024-05-08
fix: fix url for duplicate artist

pam-:

### Description

Changing URL for dupe artist


araujobarret (approved):
I thought the `-1` was the right, I tested in the other PR the URL and it worked, that's why I didn't flag anything

pam-:
@araujobarret same :lolsob: but better to be safe [per Jenna's post](https://artsy.slack.com/archives/C05EEBNEF71/p1715181326435359?thread_ts=1715179879.136499&cid=C05EEBNEF71)

PR #13867, opened 2024-05-08
chore: regenerate .secrets.baseline

ovasdi:

### Description

Considering the nature of this project I wanted to follow up https://github.com/artsy/force/pull/13855 with another PR and make the following changes:

1. Add a new function `regen()` into `detect-secrets.sh` script which regenerates the baseline. This function passes exclusions (already applied to the existing baseline) in order to fully regenerate the baseline. It turns out that this is necessary in order for the plugins released in the new version ([v1.5.0](https://github.com/Yelp/detect-secrets/releases/tag/v1.5.0)) to actually be added into the baseline.

    In addition, this serves as documentation so that it is abundantly clear which filters were used to generate the baseline and makes it easier to regenerate in the future, without having to rewrite the filters.

2. Rename the function `rebuild` to `rescan` which is more inline with what it [actually does](https://github.com/Yelp/detect-secrets/tree/52759b86b8e311a8ba0e8d464001b1bf353f54f8?tab=readme-ov-file#adding-new-secrets-to-baseline).

3. Finally, execute `yarn detect-secrets:regen` to include the new plugins in the baseline.


damassi:
@ovasdi - please feel free to self merge this kind of stuff, rather than assign! PRs like this hanging around open create roadbumps in the wider team's workflows, and are trivial.

joeyAghion:
For what it's worth I think code review is important even for changes like this. In addition to surfacing problems, it educates.

If a PR fixes an important workflow though, it's fine to nudge an expedited review.

PR #13871, opened 2024-05-09
feat: add 04-Bio-Follows-Alerts app

mc-jones:
This PR adds a new experimental app: `04-Bio-Follows-Alerts`.

This app expands on experiment `03` by allowing the assistant to:

1. Persist the collector profile blurb it generates. (Its currently saved as the `bio` field on the `CollectorProfile` model).
2. Follow artists on the user's behalf
3. Make alerts on the user's behalf

It also includes some minor chores (I can break these out into a separate PR, if we prefer, but they seemed small enough): 

- updating the `detect-secrets` baseline to `1.5.0`
- updating the spell checker's config to make spelling error more obvious, and remove the false positive for `openai`.


src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/App.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      if (inputRef.current) {
+        inputRef.current.value = ""
+        inputRef.current.focus()
+      }
+    } catch (e) {
+      sendToast({
+        variant: "error",
+        message:
+          "Something went wrong. Please try again. If you continue to see errors, message #product-ai-innovation-squad.",
```
mc-jones, line 66: We may not want to leave the second part of this, but felt nicer than just repeatedly erroring.

src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/lib/metaphysics.ts
```diff
@@ -0,0 +1,18 @@
+export async function metaphysics(args: {
+  query: string
+  variables: Record<string, unknown>
+  headers: Record<string, string>
+}) {
+  const { query, variables, headers } = args
+
+  const url =
+    `${process.env.METAPHYSICS_ENDPOINT}/v2` ||
```
mc-jones, line 9: I sometime like to run this with our dockerized environment, especially if I want to mess with the `api` or `persistence` layers. I don't think there is really a risk of this ever pointing to prod when we're testing, but let me know if I'm not thinking of something.

src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/lib/system-instructions.ts
mc-jones, line 1: Still some issues to resolve with this current prompt.

1. It doesn't offer to make follows or alerts on the user's behalf after generating / updating the bio. But it will do it when explicitly instructed. This could be better. 
2. Doesn't yet incorporate the user feedback provided around repeatedly asking for confirmation etc...
3. Needs to be more throughly tested around alerts behavior. We saw some issues around it not including a required `artistIDs` argument when calling `create_alert`

src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/lib/functions.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  const response = await metaphysics({ query, variables, headers })
+
+  console.log(
+    chalk.yellow("createAlert response: "),
+    JSON.stringify(response, null, 2)
+  )
+
+  const profileOrError = response.data.me || response.errors
```
mc-jones, line 53: Attempts to replicate the responseOrError pattern that _most_ our mutations respond with. Maybe a bit awkward, but kind of a work around to not having better validation or error handling at this stage. There _should not_ be anything coming back from a response that would problematic to feed back to the model.

damassi: I don't know how applicable it is here, but note that by using `fetchQuery` via relay, we get store caching around query/var reuse. This should speed things up where there are repeat hits to the same combination of factors.

src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/server.ts
```diff
@@ -0,0 +1,241 @@
...
+    function: {
+      name: "create_alert",
+      description: `Create an alert for an user on artsy.`,
+      parameters: {
+        type: "object",
+        properties: {
+          artistID: {
+            type: "string",
+            description:
+              "The ID of the artist to follow. It is the same as the artsy.net slug. This parameter is required.",
```
anandaroop, line 85: **suggestion (non-blocking):** I don't think this was a factor in the failure to create alerts, but maybe…

```suggestion
              "The ID of the artist for whom the alert is being set. It is the same as the artsy.net slug. This parameter is required.",
```

anandaroop (approved):
Some minor comments, but looks great 👍🏽

src/Apps/ArtAdvisor/ArtAdvisorRoutes.tsx
```diff
@@ -30,6 +30,14 @@ const ArtAdvisorApp03 = loadable(
   }
 )
 
+const ArtAdvisorApp04 = loadable(
+  () =>
+    import(/* webpackChunkName: "jobsBundle" */ "./04-Bio-Follows-Alerts/App"),
```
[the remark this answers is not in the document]
mc-jones: Yeah, good call. We definitely are. Can update the [chunk names](https://webpack.js.org/api/module-methods/#webpackchunkname) to `advisorBundle`. I'm not familiar with webpack chunking behavior enough to say what the consequence would be if we removed it, but its clearly the `force` [pattern](https://github.com/search?q=repo%3Aartsy%2Fforce%20webpackChunkName&type=code), so think it makes sense to keep, but with an appropriate name.

src/Apps/ArtAdvisor/04-Bio-Follows-Alerts/lib/metaphysics.ts
mc-jones: Yeah, good point. This should actually already be handled by the setup script and `.env.shared` so we should be able to delete.

src/Apps/ArtAdvisor/ArtAdvisorRoutes.tsx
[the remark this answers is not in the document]
mc-jones: Will push up in a different PR since we will want to change for all apps.

PR #13884, opened 2024-05-14
feat(artist): defers loading of artwork grid to client-side

dzucconi:
https://defer-grid.artsy.net/

Opening up a review app here so we can get a real comparison.

Before merging I'll want to address the fact that we have two competing breakpoint systems as this creates a problem between the placeholder and the actual filter wherein column counts might differ.

### DIFF
--- a/src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
+++ b/src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
@@ -1,10 +1,14 @@
-import React from "react"
+import React, { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { ArtistArtworkFilterRefetchContainer } from "./Components/ArtistArtworkFilter"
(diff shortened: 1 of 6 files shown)


src/Apps/Artist/Routes/WorksForSale/ArtistWorksForSaleRoute.tsx
```diff
@@ -74,3 +78,49 @@ export const ArtistWorksForSaleRouteFragmentContainer = createFragmentContainer(
     `,
   }
 )
+
+export const ArtistWorksForSaleRouteQueryRenderer: FC<{
```
damassi, line 82: Is this the sort of thing where for every lazily-loaded query renderer we need to create a wrapper component like this that defines the behavior? Or would it be possible to create a lazy-load grid that accepts a `query` and `variables prop where we pass in what we need for the grid render? As we'll be likely deferring things all over, it would be nice to make this a bit more reusable. 

(Can push this into a separate PR)

damassi:
@dzucconi - why close this?

dzucconi:
@damassi no plans for it. We can always revisit but the question of how to evaluate SEO impact remains an open one.

PR #13887, opened 2024-05-15
feat: change SWA tooltip copy and do not show tooltip for submitted artworks

nickskalkin:
### Description

- Change tooltip copy
- If artwork was already submitted - do not display tooltip on top of it (thanks @dariakoko for noticing this bug)

| Before | After |
|---|---|
| Screenshot 2024-05-15 at 09 06 48 | Screenshot 2024-05-15 at 09 06 57 |


olerichter00 (approved):
Thanks for the fix!

https://github.com/artsy/eigen/pull/10240

PR #13896, opened 2024-05-16
chore(EMI-1864): replace localization with blant conversion to comma separated numbers to match display

oxaudo:
I tried just removing this separator as the other option the ticket suggests but those big numbers where cents are not allowed felt strange to me. Also it's kinda nice when it's matching the display above.

Regex for the win!

### DIFF
--- a/src/Apps/Order/Components/OfferInput.tsx
+++ b/src/Apps/Order/Components/OfferInput.tsx
@@ -22,7 +22,7 @@ export const OfferInput: FC<OfferInputProps> = ({
 }) => {
   const formatValueForDisplay = (val: number | undefined) => {
     if (val !== undefined && val > 0) {
-      return val.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")
+      return val.toLocaleString("en-US")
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/OfferInput.tsx
```diff
@@ -22,7 +22,7 @@ export const OfferInput: FC<OfferInputProps> = ({
 }) => {
   const formatValueForDisplay = (val: number | undefined) => {
     if (val !== undefined && val > 0) {
-      return val.toLocaleString()
+      return val.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")
```
starsirius, line 25: I wonder if setting a fixed locale would just work. For example, `val.toLocaleString("en-US")`

oxaudo: I was hesitant to put specific "en-US" here but made a search based on your comment and looks like we use hardcoded "en-US" formatting already in other places around Pricing. So this works as well!

Thanks for the nudge.

PR #13908, opened 2024-05-22
fix(DIA-648): unifies breakpoints across Force

dzucconi:
Closes [DIA-648](https://artsyproduct.atlassian.net/browse/DIA-648)

Just going to spot check the few places we use this for anything other than `xs` — which doesn't actually change.

[DIA-648]: https://artsyproduct.atlassian.net/browse/DIA-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/System/Router/Boot.tsx
```diff
@@ -90,30 +81,25 @@ export const Boot = track(undefined, {
               <EnvironmentProvider environment={props.relayEnvironment}>
                 <ErrorBoundary>
                   <MediaContextProvider onlyMatch={onlyMatchMediaQueries}>
-                    <ResponsiveProvider
-                      mediaQueries={THEME.mediaQueries}
-                      initialMatchingMediaQueries={onlyMatchMediaQueries as any}
```
damassi, line 95: Wow, this is still around

PR #13927, opened 2024-05-28
feat: add initial scaffold for new Sell with Artsy flow

dblandin:

### Description

Provides an initial scaffold with some patterns in place (up for debate!) while trying to keep things as simple as possible.

- A collector enters into the persisted flow _after_ an artist is selected
- Each step can operate independently from other steps
- Some patterns to consolidate navigation / validation / mutation handling

### Screen Recording

https://github.com/artsy/force/assets/123595/07c78014-086a-4385-8319-c2149678e6df


src/Apps/Sell/SubmissionContext.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+export const SubmissionContext = createContext<SubmissionContextProps>({} as any)
+
+export const STEPS = [
+  "title",
+  "photos",
+  "details",
+  "purchase-history",
+  "dimensions"
+]
```
olerichter00, line 49: I wonder if we could derive the step directly from the `pathname`? This way, we wouldn't need to sync the step in the context and the pathname 🤔

src/Apps/Sell/Routes/NewRoute/SellArtistAutocomplete.tsx
```diff
@@ -0,0 +1,212 @@
...
+      >["node"]
+    >
+  | null
+  | undefined
+
+interface SellArtistAutocompleteOption extends AutocompleteInputOptionType {
+  option: SellAutocompleteArtist
+}
+
+export const SellArtistAutocomplete: React.FC<{
```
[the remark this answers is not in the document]
dblandin: As is, that component is not immediately compatible because it's dependent on the Formik, which makes is less portable / reusable. I think ideally we'd create an abstract component which we can wrap for either Formik usage or non-Formik usage.

Although I see some opportunity to reduce the surface size of this component right away by exposing some more elements from the ArtistAutoComplete component.

src/Apps/Sell/Components/StickyNavigation.tsx
```diff
@@ -0,0 +1,63 @@
...
+          >
+            {isDirty() ? "Continue" : "Skip" }
+          </Button>
+        )}
+      </Flex>
+    </StickyContainer>
+  )
+}
+
+const StickyContainer = styled(Flex)`
```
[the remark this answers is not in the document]
dblandin: Is that component used for the same purpose? Upon reading, it seemed like that was meant for scrolling behavior where you want a component to stick rather than float up with the rest of the scrolling content. I suppose this is more of a fixed placement, rather than sticky in that sense.

src/Apps/Sell/Components/SideNavigation.tsx
```diff
@@ -0,0 +1,23 @@
+import { Box, Join, Spacer } from "@artsy/palette"
+import { RouterLink } from "System/Router/RouterLink"
+import { useContext } from "react"
+import { useRouter } from "found"
+
+import { STEPS, SubmissionContext } from "Apps/Sell/SubmissionContext"
+
+export const SideNavigation = () => {
+  const { actions: { currentStep } } = useContext(SubmissionContext)
```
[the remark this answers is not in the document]
dblandin: Yep, totally 👍🏼

src/Apps/Sell/Routes/NewRoute/index.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from "react"
```
damassi, line 1: Can we name the folder `Routes/New` vs doubling up?

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
dzucconi: I have no idea what I'm looking at here but...

If you want something to be position fixed then you need to accommodate the rest of the UI; which the `Sticky` component does. We're going to be moving the entire global nav into this as well, which is the last hold out.

The manual height calc below is going to be very brittle.

src/Apps/Sell/Routes/NewRoute/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      mutation NewRouteCreateConsignmentSubmissionMutation($artistID: String!, $myCollectionArtworkID: String!) {
+        createConsignmentSubmission(input: { artistID: $artistID, myCollectionArtworkID: $myCollectionArtworkID }) {
+          clientMutationId
+          consignmentSubmission {
+            internalID
+          }
+        }
+      }
+    `,
+  })
```
damassi, line 40: We should move mutations in to a `Sell/Mutations` folder, following other patterns

src/Apps/Sell/DebugSubmissionContext.tsx
```diff
@@ -0,0 +1,26 @@
+import { Box } from "@artsy/palette"
+import { SubmissionContext } from "Apps/Sell/SubmissionContext"
+import { useContext } from "react"
+
+export const DebugSubmissionContext: React.FC = () => {
```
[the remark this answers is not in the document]
dblandin: Ahh, good point. Can simply call this `Debug.tsx` or `DevState.tsx` to avoid the naming conflict.

src/Apps/Sell/Routes/NewRoute/index.tsx
dblandin: Yep, definitely. Can de-duplicate the naming here.

damassi: I tend to always consider components used in sub routes to be placed in a `Foo/Routes/Route/Components` folder, treating each sub route as its own little sub-app. But this division isn't strict, especially if one is certain that the sub route will never scale out to any largish size. But that said, we should follow patterns as seen elsewhere (just in case), so yeah. 

TLDR: use yr best judgement.

src/Apps/Sell/SubmissionContext.tsx
```diff
@@ -0,0 +1,162 @@
...
+  submitSubmission: () => void
+  atFirstStep: () => boolean
+  atLastStep: () => boolean
+  currentStep: () => string
+  isSaving: () => boolean
+  isValid: () => boolean
+  onChange: (event: any) => void
+  updatePayload: (input: {}) => void
+  setInitialValues: (values: any) => void
+  isDirty: () => boolean
```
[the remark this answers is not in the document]
dblandin: re: formik — fair enough! Will circle back with some level of Formik integration.

> Yep, with a route for each step, we could (and probably should) split up the artwork model and use a different model for each step.

Yeah, that's really what I'm after regardless of Formik usage 👍🏼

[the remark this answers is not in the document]
damassi: Its cheap to spin up multiple formik contexts per route, and totally fine! Can't count the number of times we've had to unwind custom inputs in favor of formik through the years -- it always gets undone. And by using formik we might be able to use the ArtistAutoComplete component mentioned earlier.

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
dzucconi: You can achieve this layout without using anything `fixed` (or sticky for that matter) actually.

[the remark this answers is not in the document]
tam-kis: I was thinking more about steps like [Upload photos ](https://www.figma.com/design/E6rVm1tiSI4PW3EzibKKZw/Submission-%26-Pre-Consignment-Management?node-id=1448-6788&t=Esj7CSStlPd6lkW0-4) where we will definitely have a scroll if the user uploads more photos, could prevent future issues.

src/Apps/Sell/SubmissionContext.tsx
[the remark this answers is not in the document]
olerichter00: > I'm not completely sold on the upsides around consistency / reuse — our inputs are pretty robust even without Formik and I don't think copy/pasting around models/validation between web and app is particularly useful.

I'll try to explain a bit more why I think consistency between the two platforms and re-using code are good ideas.

For me, it's mainly about developer experience and velocity/efficency. In previous projects, I've realized how easy it is to develop a feature in one of these repositories and then just "convert" the code to the other platform (thanks to React/React Native, palette/palett-mobile, formik, routing by URLs, etc.). 

I see two main upsides here. The first is that it makes developing a feature for both platforms, web and app, way faster. The second is that fixing bugs or adding features on top usually involves doing it for the app and web. Keeping things consistent now means that whoever is working on the code doesn't have to understand two different code bases but just one and can use this knowledge to make the changes on both platforms.

Of course, converting the code involves more than just copying and pasting, and it's not always desirable because sometimes both platforms should behave differently, but, well, usually, it's pretty easy.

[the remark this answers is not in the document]
damassi: As someone who has been working in a very cross platform way for the past year, the copy/paste-ability between our systems has been astoundingly helpful, and almost like a miracle. So... lets continue doing what we're doing and avoid one-off, custom implementations of things 👍

dblandin:
Thanks for all the feedback so far! I'm going to rework this PR and will re-mark as ready for review soon.

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
tam-kis: The decision was to have a sticky footer and scroll the rest of the page when there is too much content - it will be addressed separately in another ticket.

src/Apps/Sell/Routes/NewRoute/SellArtistAutocomplete.tsx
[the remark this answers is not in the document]
dblandin: Now using the ArtistAutoComplete component from the Consign app 👍🏼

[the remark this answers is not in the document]
dblandin: Now using the ArtistAutoComplete component from the Consign app 👍🏼

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
dblandin: I used the `<Sticky>` component for the header but I didn't seem to be a good fit for the fixed footer. Open to suggestions or if there's another reference I should look at.

src/Apps/Sell/Components/SideNavigation.tsx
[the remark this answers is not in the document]
dblandin: Added this wrapped export 👍🏼

src/Apps/Sell/Routes/NewRoute/index.tsx
dblandin: All of the routes are single files now. 👍🏼

src/Apps/Sell/ArtworkFormContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  children: React.ReactNode
+  submissionID?: string
+}
+
+export const ArtworkFormContextProvider: React.FC<ArtworkFormContextProviderProps> = ({ children, submissionID }) => {
+  const { match, router } = useRouter()
+  const { relayEnvironment } = useSystemContext()
+  const [currentStep, setCurrentStep] = useState<string>(match.location.pathname.split("/").pop() ?? "")
+  const [atFirstStep, setAtFirstStep] = useState<boolean>(false)
+  const [atLastStep, setAtLastStep] = useState<boolean>(false)
```
dzucconi, line 44: Should just use a single state for all the steps that's a union.

Or we use `[useCursor](https://github.com/artsy/force/blob/dfb2f42b0a6c3a51d1310b457660202a3d694961/src/Components/HeroCarousel/HeroCarouselLarge.tsx#L42-L47)` for cycling through an index

src/Apps/Sell/DebugSubmissionContext.tsx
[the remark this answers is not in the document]
dblandin: Renamed this dev component to `DevDebug` for now. 👍🏼

src/Apps/Sell/sellRoutes.tsx
```diff
@@ -0,0 +1,170 @@
...
+        me {
+          ...SellApp_me
+        }
+      }
+    `,
+    children: [
+      {
+        path: "",
+        layout: "ContainerOnly",
+        Component: WelcomeRoute,
```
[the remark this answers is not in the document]
dblandin: All routes should be bundle split now. 👍🏼

src/Apps/Sell/ArtworkFormContext.tsx
dblandin: Ahh `useCursor` is very cool. Yeah, I'm aware that this initial state tracking is rough. I anticipate we'll improve upon this as we iterate.

Do you feel strongly about seeing a different state tracking implementation before merge?

dblandin:
@damassi, @olerichter00, @tam-kis This PR is ready for another review pass. 🔎

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
dzucconi: One could just write a normal flex layout — no need for `fixed`; just `overflow: auto` where appropriate?

src/Apps/Sell/ArtworkFormContext.tsx
dzucconi: I mean: the way this state is implemented can absolutely lead to invalid states and is just confusing. There's never a reason to have so many `useState` declarations like that.

src/Apps/Sell/Routes/TitleRoute.tsx
```diff
@@ -0,0 +1,74 @@
...
+        defaultValue={values.title} />
+    </>
+  )
+}
+
+interface TitleRouteProps {
+  submission: TitleRoute_submission$key
+}
+
+export const TitleRoute: React.FC<TitleRouteProps> = props => {
```
dblandin, line 44: One specific bit of feedback I'd like in this PR is on the structure/separation of these steps. I think it's valuable that each step be isolated/independent from others and each of the steps in this PR:
- Query for data scoped to the step
- Validate against form inputs scoped to the step
- Mutate data scoped to the step

This will make it easier to introduce new steps in the flow and eventually group steps into different phases, exposed depending on the state of the submission. For example we request and collect information in a Tier 1 group right away and plan to only request and collect information in a Tier 2 group of information gathering after the submission has been approved.

dzucconi: Yep — we do a lot of these stepped wizard things. You want completely isolated steps wherein they have no knowledge of how they might be arranged. Then a parent context that controls the arrangement, provides next/prev functions, and accumulates any state.

When they are linear that arrangement is just expressed as an array or a numeric index. When they might branch based on conditions we use the workflow engine. [Example](https://github.com/artsy/force/blob/main/src/Components/Onboarding/config.ts).

src/Apps/Sell/ArtworkFormContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { useSystemContext } from "System/SystemContext";
+
+interface Actions {
+  goToPreviousStep: () => void
+  goToNextStep: () => void
+  updateSubmission: (values: UpdateSubmissionMutationInput) => void
+}
+
+interface State {
+  atFirstStep: boolean
```
damassi, line 14: `isFirstStep`
`isLastStep`

src/Apps/Sell/ArtworkFormContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const [currentStep, setCurrentStep] = useState<string>(match.location.pathname.split("/").pop() ?? "")
+  const [atFirstStep, setAtFirstStep] = useState<boolean>(false)
+  const [atLastStep, setAtLastStep] = useState<boolean>(false)
+
+  useEffect(() => {
+    const step = match.location.pathname.split("/").pop() ?? ""
+    setCurrentStep(step)
+    setAtFirstStep(step === STEPS[0])
+    setAtLastStep(step === STEPS[STEPS.length - 1])
+  }, [match.location])
```
damassi, line 51: Is there something specific that we're looking for in `match.location`? That's a large object from found, so if it's a change in `pathname`, we should be specific about that.

Also, this would likely be fixed by @dzucconi's suggestion, but its a bit confusing that we're setting three separate steps like this, and on every change to match.location. What step are we at? Just worry about that. If something feels like it could be simpler, it likely can. Work from the ideal and go backwards 🙏

src/Apps/Sell/ArtworkFormContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  const navigateToStep = (step: string) => {
+    if (step) {
+      router.push(`/sell2/submissions/${submissionID}/${step}`)
+    }
+  }
+
+  const updateSubmission = (values: UpdateSubmissionMutationInput) => {
+    return createOrUpdateConsignSubmission(relayEnvironment, {
```
damassi, line 81: This is totally fine, but its also better to use our modern mutation forms. At this point in Force we should never have to pass a relay network anywhere except in very rare and specific cases.

src/Apps/Sell/ArtworkFormContext.tsx
```diff
@@ -0,0 +1,109 @@
...
+    goToNextStep,
+    updateSubmission,
+  }
+
+  const state = {
+    atFirstStep,
+    atLastStep,
+    currentStep,
+    submissionID
+  }
```
damassi, line 98: If we've got actions and state, why not use `useReducer` vs duplicating patterns? (Though this will likely go away with the cursor)

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { useArtworkFormContext } from "Apps/Sell/ArtworkFormContext"
+import { useFormikContext } from "formik"
+import { useRouter } from "found"
+import styled from "styled-components"
+import { themeGet } from "@styled-system/theme-get"
+
+export const BottomFormNavigation = () => {
+  const { router } = useRouter()
+  const { actions, state } = useArtworkFormContext()
+  const { isValid, isSubmitting, submitForm } = useFormikContext()
```
damassi, line 11: This might get a bit confusing later so wanted to call out now:

Rather than `useArtworkFormContext`, can we name it something else and more specific? The reason being, we typically reserve that sort of naming for wrapping formik context hooks, which typically need to be done in order to support proper typescript types. Eg, 

```tsx
export const useArtworkFormContext = () => {
  const formikContext = useFormikContext<ArtworkFormValues>()
  return formikContext
}
```

So having the somewhat-unrelated `useArtworkFormContext` _and_ the likely needed wrapped `useFormikContext` could lead to some confusion. 

Maybe rename to `useSellContext`?

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (state.currentStep === "thank-you") return null
+
+  const onContinue = async () => {
+    await submitForm()
```
damassi, line 16: Assuming this checks for `onError` (or something) higher up

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -0,0 +1,79 @@
...
+  left: 0;
+  position: fixed;
+  width: 100%;
+  border-top: 1px solid ${themeGet("colors.black10")};
+  align-items: center;
+  justify-content: center;
+  padding-bottom: env(safe-area-inset-bottom);
+  padding: env(safe-area-inset-top) env(safe-area-inset-right)
+    env(safe-area-inset-bottom) env(safe-area-inset-left);
+`
```
damassi, line 79: This might be fine, but every time actual CSS shows up in a PR it can be pretty safely assumed that it will degrade into a bug later on. 

Is this _strictly necessary_? Can we not place the footer by positioning it at the bottom of a flex box container and then adding a (prop) to the center element with `flex={1}`? We should try, and if there's other CSS anywhere we should factor that out as well and at the very least move what we can into props.

src/Apps/Sell/Components/StepsNavigation.tsx
```diff
@@ -0,0 +1,20 @@
+import { Box, Join, Spacer } from "@artsy/palette"
+import { RouterLink } from "System/Router/RouterLink"
+import { useRouter } from "found"
+
+import { STEPS, useArtworkFormContext } from "Apps/Sell/ArtworkFormContext"
+
+export const StepsNavigation = () => {
```
damassi, line 7: React.FC

src/Apps/Sell/Components/StepsNavigation.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const { state } = useArtworkFormContext()
+  const { match } = useRouter()
+
+  return (
+    <Box>
```
damassi, line 12: No need for outer wrapper

src/Apps/Sell/Routes/DetailsRoute.tsx
```diff
@@ -0,0 +1,109 @@
...
+  medium: Yup.string().trim(),
+})
+
+interface FormValues {
+  year: string
+  category: ConsignmentSubmissionCategoryAggregation
+  medium: string
+}
+
+const InnerForm: React.FC = props => {
```
damassi, line 41: Can we move the form stuff into its own component? `SellDetailsForm`

src/Apps/Sell/Routes/DetailsRoute.tsx
```diff
@@ -0,0 +1,109 @@
...
+    category:
+      (submission.category?.toUpperCase() as ConsignmentSubmissionCategoryAggregation) ??
+      "",
+    medium: submission.medium ?? ""
+  }
+
+  return (
+    <>
+      <Text mb={2} variant="xl">Artwork details</Text>
+      <Formik<FormValues>
```
damassi, line 95: SellDetailsForm can also contain this, since we're not passing anything into it really except for the locally-defined `initialValues`

src/Apps/Sell/Routes/DetailsRoute.tsx
```diff
@@ -0,0 +1,109 @@
...
+        initialValues={initialValues}
+        onSubmit={onSubmit}
+        validateOnMount
+        validationSchema={Schema}
+      >
+        <>
+          <InnerForm />
```
damassi, line 102: #non-blocking suggestion: Its fine, but I would use the renderProps form of Formik to pluck out any values needed for InnerForm and then just move all of the form related components inside of here, and remove InnerForm (here and elsewhere).

damassi (commented):
Looking really good @dblandin - thanks for updating with Formik, this will make things much friendlier in the future (as well as for devs who didn't write the intial implementation). 

A few (minor) comments, though we should investigate the state suggestion that @dzucconi proposed and / or refactor to use `useReducer`.

src/Apps/Sell/Components/StickyNavigation.tsx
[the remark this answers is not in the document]
tam-kis: Yes, we can go with something more simple if the Sticky didn't work, without height calculations and fixed position. I can look into it also in the other ticket once we have more content, if other comments are already addressed and ready to be approved.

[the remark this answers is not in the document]
dblandin: Just paired on this with @olerichter00 to provide a Flex layout which you'll find in the last commit 👍🏼

src/Apps/Sell/Components/BottomFormNavigation.tsx
dblandin: Just paired on this with @olerichter00 to provide a Flex layout which you'll find in the last commit 👍🏼

src/Apps/Sell/Components/BottomFormNavigation.tsx
damassi: Realizing now looking at the rest of the PR (with the individual routes) that there's multiple formik contexts for each route. So maybe this is fine (and can pass generic directly in to useFormikContext, until we need to make it more reusable). 

That said, it would be worthwhile renaming the useArtworkFormContext to something specific to avoid future confusion.

src/Apps/Sell/Components/BottomFormNavigation.tsx
damassi: sweet, thanks all

src/Apps/Sell/Routes/DetailsRoute.tsx
dblandin: Done 👍🏼

src/Apps/Sell/ArtworkFormContext.tsx
dblandin: Went with a `useCursor` based state approach in recent commits.

src/Apps/Sell/ArtworkFormContext.tsx
dblandin: Switched this over to use useMutation-backed hooks.

src/Apps/Sell/Components/BottomFormNavigation.tsx
dblandin: I renamed the context to avoid conflicting naming with the Formik/form setup.

> that there's multiple formik contexts for each route.

For any particular route there should only be one Formik context active, but I think I see what you mean about the typing. This shared component is only using the generic actions from the context but let me know if there's something else you'd suggest to tighten this up.

src/Apps/Sell/ArtworkFormContext.tsx
dblandin: I took a pass at converting this over to utilize useCursor. Let me know if anything else jumps out at you.

src/Apps/Sell/Routes/TitleRoute.tsx
dblandin: We might have some conditional needs for steps in the future so might reach for the workflow engine later.

src/Apps/Sell/ArtworkFormContext.tsx
dblandin: Removed this hook as part of the state management update.

dblandin:
@olerichter00 @tam-kis @damassi @dzucconi Should be ready for another look now. Thanks!

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -0,0 +1,60 @@
+import { Flex, Button } from "@artsy/palette"
+import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
+import { useRouter } from "System/Router/useRouter"
+import { useFormikContext } from "formik"
+
+export const BottomFormNavigation = () => {
+  const { isValid, isSubmitting, submitForm } = useFormikContext()
+  const { router } = useRouter()
+  const { actions, state: { isFirstStep, isLastStep, submissionID } } = useSellFlowContext()
```
damassi, line 9: 👍

src/Apps/Sell/Routes/PhotosRoute.tsx
```diff
@@ -0,0 +1,69 @@
...
+      imageUrls
+    }
+  }
+`
+
+const Schema = Yup.object().shape({
+  assets: Yup.array(),
+  // .min(1)
+  // .transform(fields => fields.filter(c => !c.errorMessage))
+  // .of(Yup.object().test("photos", value => value.assetId)),
```
damassi, line 23: (I imagine this is a future TODO)

damassi (approved):
Looks quite perfect now @dblandin through and through. (This will also scale nicely into the future.) Thanks so much for working through all of this feedback 👍 

:shipit:

src/Apps/Sell/Routes/PhotosRoute.tsx
dblandin: Yep, I didn't want to overload this PR with the photo uploading/associating logic but wanted to at least insert the step along with basic navigation+query support.

dblandin:
Going to go ahead and merge this in. Thanks all for the review!

PR #13933, opened 2024-05-29
feat(AMBER-700): Move Artwork Details to top of page/sidebar for private works

lilyfromseattle:

This PR solves [AMBER-700]

### Description

As discussed with Peter, we would like the artwork details to be at the top of the page as in the designs. I also limited read more truncation to 3 items instead of 4 and made some small styling adjustments.

This PR also adds "Frame not included" text for private works to artwork details, as requested by Peter.

__NOTE:__
We decided to render the COA info in two places because the icon doesn't give full details. Peter approved this

__Before:__
Screenshot 2024-05-29 at 1 35 58 PM

__After:__
Screenshot 2024-05-29 at 1 35 50 PM

Screenshot 2024-06-05 at 10 51 14 AM

[AMBER-700]: https://artsyproduct.atlassian.net/browse/AMBER-700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtworkSidebar/PrivateArtworkAdditionalInfo.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export interface PrivateArtworkAdditionalInfoProps extends FlexProps {
+  artwork: PrivateArtworkAdditionalInfo_artwork$key
+}
+
+export const PrivateArtworkAdditionalInfo: React.FC<PrivateArtworkAdditionalInfoProps> = ({
+  artwork,
+  ...flexProps
+}) => {
+  const data = useFragment(privateArtworkAdditionalInfoFragment, artwork)
```
damassi, line 36: 🙏

src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
```diff
@@ -42,7 +42,7 @@ export const AuthenticityCertificate: FC<AuthenticityCertificateProps> = ({
         <Text variant="xs" color="black100">
           Includes a{" "}
           <Clickable onClick={handleClick} textDecoration="underline">
-            Certificate of Authenticity
+            Certificate of Authenticity wooo
```
damassi, line 45: Lil leftover text ^

damassi (dismissed):
One little fix required, otherwise, LGTM 👍

src/Apps/Artwork/Components/ArtworkSidebar/PrivateArtworkAdditionalInfo.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const [openConditionModal, setOpenConditionModal] = useState(false)
+
+  const { trackEvent } = useTracking()
+  const {
+    contextPageOwnerId,
+    contextPageOwnerSlug,
+    contextPageOwnerType,
+  } = useAnalyticsContext()
+
+  const listItems = [
```
damassi, line 63: I'm not sure just how much of this can be reused from the other component, but if there's an opportunity to share the code that would be good (on a follow-up PR if you want -- if you think its helpful!):

Thinking something like:

```
export const getArtworkAdditionalInfo = (artwork) => {
  const {
    category,
    series,
    publisher,
    manufacturer,
    image_rights,
    internalID,
    canRequestLotConditionsReport,
    signatureInfo,
    conditionDescription,
    certificateOfAuthenticity,
    mediumType,
    isUnlisted,
  } = artwork
  
  if (isUnlisted) {
    // do unlisted things 
  }

  const listItems = [ ... ] 
  
  return listItems 
}
```


And then you can import it in both components. 

```tsx
// in PrivateArtworkkAdditionalInfo.tsx

const listItems = getArtworkAdditionalInfo(artwork)

...
```

And the code is now shared, with less maintenance required.

src/Apps/Artwork/Components/TrustSignals/AuthenticityCertificate.tsx
lilyfromseattle: whoooops lol

src/Apps/Artwork/Components/ArtworkSidebar/PrivateArtworkAdditionalInfo.tsx
lilyfromseattle: The truncation stuff fits nicely in an if but the list items are different for private works, would trying to push certain items to the list only for private works be kind of clunky?

damassi: Its a matter of balancing things. Are there more items for private artworks than the norm? If so than a new component suffices. But if there are more shared items than not, then pushing things in isn't clunky, it just makes things easier to maintain in the future

lilyfromseattle: There are less items for private works, but we also want to keep the order consistent, so it would be a matter of pushing things in at certain places which feels like a headache. Basically I've decided if Peter wants "framed" in both I'll share them but if not I'll probably keep as is to avoid weird array splicing logic.

damassi: Here's an example approach: https://github.com/artsy/force/pull/13961

PR #13092, opened 2023-11-01
feat: new artwork alert (create alert) modal

dblandin:
The type of this PR is: **Feat**

This PR relates to the following Jira tickets: [ONYX-386](https://artsyproduct.atlassian.net/browse/ONYX-386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ), [ONYX-385](https://artsyproduct.atlassian.net/browse/ONYX-385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ), [ONYX-389](https://artsyproduct.atlassian.net/browse/ONYX-389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ), [ONYX-387](https://artsyproduct.atlassian.net/browse/ONYX-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ), [ONYX-413](https://artsyproduct.atlassian.net/browse/ONYX-413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)

### Description

I made some adjustments, folding in the earlier work from https://github.com/artsy/force/pull/13088, including:
- Moved components to src/Components/ArtworkAlert
- Created a debug/dev app, accessible via /dev/alert, to more easily exercise the alert flow and inspect state as things change under the hood.
- Wired up some persistence when transitioning away from the Formik-backed details step
- Created some new React hooks to offer hopefully offer a better DX

An simple example of how the new component would be initialized:

```jsx
const { artworkAlertComponent, showArtworkAlert } = useArtworkAlert({ initialCriteria: { artworkIDs: ["andy-warhol"] })

return (
  <>
    <Button onClick={showArtworkAlert}>Create Alert</Button>
    {artworkAlertComponent}
  </>
)
```

[...]

PR #13936, opened 2024-05-30
fix(artwork): fixes re-rendering alert button

dzucconi:
Noticed this flashing alert button on the artwork page:

https://github.com/artsy/force/assets/112297/4d4f32bd-5da1-4059-a270-216ba6f7934a

------

This was primarily due to the alert button component being declared _inside_ of another component's body. Just _never do this_. I wish there was a linting rule for this sort of thing because it's happened a number of times.

Initially I thought it was due to the timer hook being used to just determine if a date was in the past, which is related, but in fact _any_ timer/polling on the page would cause this thing to re-render. I fixed that anyway because it was unnecessary.

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
+++ b/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
@@ -16,7 +16,7 @@ import {
   useTheme,
 } from "@artsy/palette"
 import * as DeprecatedSchema from "@artsy/cohesion/dist/DeprecatedSchema"
-import { FC, useCallback, useEffect, useState } from "react"
+import { useCallback, useEffect, useState } from "react"
 import { useInquiry } from "Components/Inquiry/useInquiry"
 import { ErrorWithMetadata } from "Utils/errors"
 import { logger } from "@sentry/utils"
@@ -351,18 +351,6 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
   const isCreateAlertAvailable =
     artwork.isEligibleToCreateAlert && artwork.isSold
 
-  const AlertSwitch: FC = () => {
-    if (!isCreateAlertAvailable) {
-      return null
-    }
-
-    return (
-      <ProgressiveOnboardingAlertCreateSimple>
-        <CreateAlertButton width="100%" size="large" />
-      </ProgressiveOnboardingAlertCreateSimple>
-    )
-  }
-
   const renderButtons: {
     buyNow?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
     makeOffer?: ResponsiveValue<"primaryBlack" | "secondaryBlack">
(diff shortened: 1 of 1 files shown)


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -77,12 +74,14 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
...
   const partnerIcon = artwork.partner?.profile?.icon?.url
 
-  const activePartnerOffer =
-    (!partnerOfferTimer.hasEnded && partnerOffer) || null
+  const isPartnerOfferPast = useMemo(() => {
+    if (!partnerOffer?.endAt) return false
+    return new Date(partnerOffer.endAt) < new Date()
+  }, [partnerOffer])
+
+  const activePartnerOffer = (!isPartnerOfferPast && partnerOffer) || null
```
dzucconi, line 84: Looking at this now: why would we fallback to `null`...

dzucconi: Oh wow — it's not a boolean that's why. So very confusing but I'll leave it alone.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -351,18 +350,6 @@ export const ArtworkSidebarCommercialButtons: React.FC<ArtworkSidebarCommercialB
   const isCreateAlertAvailable =
     artwork.isEligibleToCreateAlert && artwork.isSold
 
-  const AlertSwitch: FC = () => {
```
araujobarret, line 354: I wonder how that gets passed through review, it shouldn't

dzucconi: Probably just hard to notice with whitespace.

dzucconi: Oh! https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
erikdstock, line 84: I don't understand why we are removing the timer here. The timer is used because the partner offer is queried once but has an end time, at which time the button should no longer respect it when creating the order.

erikdstock (dismissed):
If we remove the `useTimer` hook we should add a check when the purchase button is created to fall back to the normal purchase behavior rather than using the partner offer price. That might be cleaner anyways. Otherwise LGTM.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
dzucconi: Oh, OK — You're right — I think I understand now. I think this just confused me trying to track down what was causing the re-render. Really it's just the nested component.

dzucconi: So I restored the timer code but it definitely is problematic: for instance it continues to check once per second even if the date is past.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
dblandin: I take [the blame](https://github.com/artsy/force/pull/13092/files#diff-022388e824526755e0276e0538a702780bf856cac90d5ca3c7d8e5dcc65850acR116-R146) for this one 🤦🏼 

Glad there's a linting rule to enable and will be sure to avoid this in the future 👍🏼

PR #13947, opened 2024-05-31
fix(shows): makes shows index useful again

dzucconi:
[Context here](https://artsy.slack.com/archives/C03N12SR0RK/p1715174220151949)

Some partners may abuse the notion of 'online shows' in order to put together loose collections of 'inventory', then manually alter the start dates so that they are constantly bumped to the top of the index.

Including 'online shows' here was initially a response to the pandemic and has outlived its usefulness.


anandaroop (approved):
🙏🏽 🙏🏽 🙏🏽 

(Minor point: online shows predate covid; but the pandemic did result in the much more full-featured Viewing Rooms feature.)

dzucconi:
Yeah, I think I'm misremembering that. Regardless!

PR #13952, opened 2024-06-03
feat(Conversation): shows a toast error when convo's cta fails

araujobarret:

### Description

This PR basically:
- makes extra safe to avoid CTAs enabled in a conversation when the artwork is unpublished(cannot fully reproduce it)
- shows a simple toast error when the conversation make offer/purchase fails


src/Apps/Conversations/components/ConversationCTA/ConversationMakeOfferButton.tsx
```diff
@@ -77,6 +78,7 @@ export const ConversationMakeOfferButton: React.FC<ConversationMakeOfferButtonPr
         )
       }
     } catch (error) {
+      sendToast({ message: "Error", variant: "error" })
```
damassi, line 81: The way we have it elsewhere in volt is like: `Error. Please try again.`

(Ditto below)

damassi (dismissed):
Thanks so much @araujobarret  👍

src/Apps/Conversations/components/ConversationCTA/ConversationMakeOfferButton.tsx
erikdstock: Not for this PR but i feel like with our structured mutation success/error types it should be so much easier to implement consistent error handling that we can add to with the general fallback message. Maybe there's already a good pattern somewhere.

erikdstock (dismissed):
Thanks for this!

src/Apps/Conversations/components/ConversationCTA/ConversationMakeOfferButton.tsx
dzucconi: Really? I feel like it's the opposite. Because now you have to query for the error. Previously, errors are just baked into everything.

araujobarret: done in https://github.com/artsy/force/pull/13952/commits/095bf0ccffb6896fd8102faf7bd962c9cf55c5e5

PR #13955, opened 2024-06-04
fix: Start Selling sticky hidden on mWeb

nickskalkin:

### Description

[Slack thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1717421505002459)

Should fix broken sticky with CTA on /sell page on mWeb.
I cannot reproduce the issue on my laptop - still looking for how to do it (going to create review app soon in any case). In the meantime I borrowed the code from the "cookie content" sticky. I think it should work.


olerichter00 (dismissed):
Thanks for the fix 🙏 

If we make the footer sticky with `position: fixed`, we can probably remove `maxHeight={mobileHeight} overflow="hidden"` from [here](https://github.com/artsy/force/blob/41c2c83708877b9a2eb94c050704314365399541/src/Apps/Consign/Routes/MarketingLanding/MarketingLandingApp.tsx#L55).

olerichter00 (approved):

PR #13965, opened 2024-06-06
feat: update text for the inquiry flow

egdbear:

This PR solves [DIA-402]

### Description
This PR updates the text copy as specified in the ticket.

Before: 
Screenshot 2024-06-06 at 11 40 56 AM

After: 

Screenshot 2024-06-06 at 11 41 04 AM

Screenshot 2024-06-06 at 12 05 31 PM

Screenshot 2024-06-06 at 12 19 16 PM

[DIA-402]: https://artsyproduct.atlassian.net/browse/DIA-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/EditProfile/SettingsEditProfileRoute.tsx
+++ b/src/Apps/Settings/Routes/EditProfile/SettingsEditProfileRoute.tsx
@@ -1,4 +1,5 @@
-import { Column, GridColumns, Join, Separator } from "@artsy/palette"
+import { Column, GridColumns, Join, Separator, Text } from "@artsy/palette"
+import InfoIcon from "@artsy/icons/InfoIcon"
 import { SettingsEditProfileFieldsFragmentContainer } from "Apps/Settings/Routes/EditProfile/Components/SettingsEditProfileFields"
 import { SettingsEditProfileRoute_me$data } from "__generated__/SettingsEditProfileRoute_me.graphql"
 import React from "react"
@@ -13,6 +14,13 @@ const SettingsEditProfileRoute: React.FC<SettingsEditProfileRouteProps> = ({
 }) => {
   return (
     <GridColumns>
+      <Column span={8} bg="black5" p={2} display="flex" gap={1}>
+        <InfoIcon flexShrink={0} />
+        <Text variant="sm-display">
+          The information you provide here will be shared when you contact a
+          gallery or make an offer.
+        </Text>
+      </Column>
       <Column span={8}>
         <Join separator={<Separator my={4} />}>
           <SettingsEditProfileFieldsFragmentContainer me={me} />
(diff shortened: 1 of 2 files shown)


araujobarret (dismissed):

src/Apps/Settings/Routes/EditProfile/SettingsEditProfileRoute.tsx
```diff
@@ -13,6 +22,17 @@ const SettingsEditProfileRoute: React.FC<SettingsEditProfileRouteProps> = ({
 }) => {
   return (
     <GridColumns>
+      <Column span={8}>
+        <Box p={2} mb={1} bg="black5">
+          <Flex>
+            <InfoIcon />
+            <Text variant="sm-display" marginLeft={"5px"}>
```
dzucconi, line 29: Use themed values.

src/Apps/Settings/Routes/EditProfile/SettingsEditProfileRoute.tsx
```diff
@@ -13,6 +22,17 @@ const SettingsEditProfileRoute: React.FC<SettingsEditProfileRouteProps> = ({
 }) => {
   return (
     <GridColumns>
+      <Column span={8}>
```
dzucconi, line 25: Both the box and the flex aren't needed. Set display flex on the column, remove the mb as well.

dzucconi: Notice also instance how that icon shrunk.

dzucconi, line 25:
```tsx
<Column span={8} bg="black5" p={2} display="flex" gap={1}>
        <InfoIcon flexShrink={0} />

        <Text variant="sm-display">
          The information you provide here will be shared when you contact a
          gallery or make an offer.
        </Text>
      </Column>
```

However! Wondering if this should actually be a [banner](https://palette-storybook.artsy.net/?path=/story/components-banner--default)?

egdbear: My understanding is that Banner components are typically used for status updates or confirmations (confirmed in the code). I'm not sure if we should use it here. Is this a design question, or can we proceed with the existing approach? Open to suggestions!

dzucconi: Yeah, that was also my question. I think typically we use the [Message](https://palette-storybook.artsy.net/?path=/story/components-message--default) for this sort of thing, though I do see the need for an icon slot.

dzucconi: Anyway, fine with existing approach (with the CSS edits I propose) for the time being.

egdbear: 👁️

egdbear: sounds good!

dzucconi (dismissed):

PR #13967, opened 2024-06-06
Chore(EMI-1671): remove name from pickup form

erikdstock:

This PR solves [EMI-1671]

### Description
This PR removes the name field from the pickup form in the refactored shipping route. It also fixes a (skipped) test that depends on filling in [jsdom's missing srcroll-into-view helper](https://stackoverflow.com/questions/53271193/typeerror-scrollintoview-is-not-a-function) on an HTML element.

[EMI-1671]: https://artsyproduct.atlassian.net/browse/EMI-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping2/Components/__tests__/FulfillmentDetailsForm.jest.tsx
+++ b/src/Apps/Order/Routes/Shipping2/Components/__tests__/FulfillmentDetailsForm.jest.tsx
@@ -150,19 +150,17 @@ describe("FulfillmentDetailsForm", () => {
       ).toBeVisible()
     })
 
-    it("has name and phone number fields", async () => {
+    it("has phone number field", async () => {
       renderTree(testProps)
 
       await userEvent.click(
         screen.getByRole("radio", { name: /Arrange for pickup/ })
       )
 
-      const fullNameField = await screen.findByPlaceholderText("Full name")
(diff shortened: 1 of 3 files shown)


src/Apps/Order/Routes/Shipping2/Components/__tests__/FulfillmentDetailsForm.jest.tsx
```diff
@@ -187,11 +185,9 @@ describe("FulfillmentDetailsForm", () => {
...
 
       await waitFor(() => {
-        ;["Full name is required", "Phone number is required"].forEach(
-          error => {
-            expect(screen.getByText(error)).toBeInTheDocument()
-          }
-        )
+        ;["Phone number is required"].forEach(error => {
+          expect(screen.getByText(error)).toBeInTheDocument()
+        })
```
MrSltun, line 190: **Suggestion**: maybe we could update it to one instead of an array?
```suggestion
        expect(screen.getByText("Phone number is required")).toBeInTheDocument()
```

MrSltun (dismissed):
Looks great! I just added one non-blocking suggestion

PR #13977, opened 2024-06-10
feat(fadeinbox): animation standardization fadein isMounted

tam-kis:
The type of this PR is: Feat

### Description
Add FadeInBox to standardize fade in animation for `isMounted` on Force

https://github.com/artsy/force/assets/17580625/3a721638-0fe4-47cf-be53-3119fa90ff7c

### DIFF
--- a/src/Components/FadeInBox.tsx
+++ b/src/Components/FadeInBox.tsx
@@ -2,10 +2,7 @@ import { Box, BoxProps, useDidMount } from "@artsy/palette"
 
 interface FadeInBoxProps extends BoxProps {}
 
-export const FadeInBox: React.FC<FadeInBoxProps> = ({
-  children,
-  ...boxProps
-}) => {
+export const FadeInBox: React.FC<FadeInBoxProps> = ({ children, ...rest }) => {
   const isMounted = useDidMount()
 
   return (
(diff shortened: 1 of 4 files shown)


src/Components/FadeInBox.tsx
```diff
@@ -0,0 +1,17 @@
...
+  opacity?: number
+}
+
+export const FadeInBox: React.FC<FadeInBoxProps> = ({ children }) => {
+  const isMounted = useDidMount()
+
+  return <FadeInWrapper opacity={isMounted ? 1 : 0}>{children}</FadeInWrapper>
+}
+
+const FadeInWrapper = styled(Box)<FadeInBoxProps>`
```
damassi, line 14: Rather than writing wrapper components + regular CSS, we should just use props if we can:

```tsx
<Box
  height="100%"
  opacity: isLoading ? 0 : 1,
  style={{
    transition: "opacity 250ms",
  }}
>
```

A good follow up here is to add the `transition` mixin here:

https://github.com/artsy/palette/blob/main/packages/palette/src/elements/Box/Box.tsx#L50

And then we don't need to jump into the `style` prop at all!

src/Components/FadeInBox.tsx
```diff
@@ -0,0 +1,22 @@
+import { Box, BoxProps, useDidMount } from "@artsy/palette"
+
+interface FadeInBoxProps extends BoxProps {}
+
+export const FadeInBox: React.FC<FadeInBoxProps> = ({
+  children,
+  ...boxProps
```
dzucconi, line 7:
```suggestion
  ...rest
```

would be more idiomatic

dzucconi (approved):

PR #13982, opened 2024-06-12
feat: add 05-near-object-rail experiment

mc-jones:
This PR adds and experiment to evaluate using a `nearObject` [search](https://weaviate.io/developers/weaviate/api/graphql/search-operators#nearobject) to populate and update an artwork rail based on user interactions. Specifically, "liking" or "disliking" an artwork. 

https://github.com/artsy/force/assets/38149304/9efb31ee-95e4-4fc7-ac59-15f7c7cbbe38

### Considerations

- This experiment has generated a few open questions that we would need to address before introducing this approach into a product. I believe they are best demonstrated through the code, so I have highlighted them in PR comments. 
- There are some UI rough edges that we would want to address, but my focus on this experiment was fleshing out the `nearObject` based rail concept.

### DIFF
--- a/src/Apps/ArtAdvisor/05-Near-Object-Rail/App.tsx
+++ b/src/Apps/ArtAdvisor/05-Near-Object-Rail/App.tsx
@@ -1,0 +1,144 @@
+import { FC, useEffect, useState } from "react"
+import {
+  Spinner,
+  Spacer,
+  Box,
+  Select,
+  Text,
+  useToasts,
+  GridColumns,
+  Column,
+} from "@artsy/palette"
+import { Rail } from "Components/Rail/Rail"
+import { Artwork } from "Apps/ArtAdvisor/05-Near-Object-Rail/components/Artwork"
+import _ from "lodash"
+
+export type ArtworkType = {
+  _additional: {
+    id: string
+  }
+  title: string
+  imageUrl: string | null
+}
+
+type UserType = {
+  _additional: {
+    id: string
+  }
+  name: string
+}
+
+type UsersListType = {
+  value: string
+  text: string
+}
+
+export const App: FC = () => {
+  const [artworks, setArtworks] = useState<ArtworkType[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const [usersList, setUsersList] = useState<UsersListType[]>([])
+  const [userId, setUserId] = useState<string>("")
+
+  const { sendToast } = useToasts()
+
+  const getNearObject = async (id: string) => {
+    if (!id) return
+
+    setIsLoading(true)
+
+    try {
+      const response = await fetch(`/api/advisor/5/near_object/${id}`)
+      const data = await response.json()
+      setArtworks(data)
+    } catch (error) {
+      console.error(error)
+      sendToast({
+        variant: "error",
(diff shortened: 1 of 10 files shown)


src/Apps/ArtAdvisor/05-Near-Object-Rail/App.tsx
```diff
@@ -0,0 +1,144 @@
...
+type UsersListType = {
+  value: string
+  text: string
+}
+
+export const App: FC = () => {
+  const [artworks, setArtworks] = useState<ArtworkType[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(false)
+  const [usersList, setUsersList] = useState<UsersListType[]>([])
+  const [userId, setUserId] = useState<string>("")
```
mc-jones, line 40: In this PR, the `userId` and `artworkId` are ALWAYS referring to their respective `weaviate` ids. I considered using `weaviateUserId` and `weaviateArtworkId`, but the longer name did not seem worth the tradeoff. I'm happy to clarify this more if we feel it would be an improvement.

src/Apps/ArtAdvisor/05-Near-Object-Rail/helpers/weaviate.tsx
mc-jones, line 1: I kept these helper methods fairly generalized (e.g, one `createReference` function instead of one `likedArtwork` function and one `dislikedArtwork` function).

This feels like a reasonable, but perhaps not quite right decision. They are more reusable, but the function signatures can be a bit awkward as a result; the `objectClass` and `targetClass` arguments in `getNearObjects`, for example. 

I don't have any issue with breaking these apart in our next iteration.

src/Apps/ArtAdvisor/05-Near-Object-Rail/server.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    let artworks
+
+    if (userLikes.length > 0) {
+      artworks = await getNearObjects({
+        objectId: userId,
+        objectClassName: "DiscoveryUsers",
+        targetClassName: "DiscoveryArtworks",
+        query: artworksQuery,
+        limit: 50,
```
mc-jones, line 79: The results of this search are calculated solely on the artworks "liked". We only track "disliked" artworks to filter them from the results. This is a trade off with the `nearObect` [search](https://github.com/artsy/force/pull/13982/files#r1636854765) vs the `nearText` search, which [allows us to](https://weaviate.io/developers/weaviate/api/graphql/search-operators#variables-2) `moveAwayFrom` and object.

src/Apps/ArtAdvisor/05-Near-Object-Rail/helpers/weaviate.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+    return await client.graphql
+      .get()
+      .withClassName(className)
+      .withFields(query)
+      .withWhere({
+        operator: "Equal",
+        path: ["id"],
+        valueText: objectId,
+      })
```
mc-jones, line 57: Uses [Metadata Filter](https://weaviate.io/developers/weaviate/search/filters#metadata-filter---by-object-id)

anandaroop (approved):
Nice

PR #13984, opened 2024-06-12
refactor(loading-bar): Polish top-bar loading states

damassi:

### Description

This brings over our more polished top loading bar from Volt. Makes things feel a bit more modern and I think faster, vs the more linear version that's currently on the site:

https://github.com/artsy/force/assets/236943/cda557ac-ffa0-4c73-b78d-6d0cdf79702e


dzucconi (approved):
👍

PR #13985, opened 2024-06-12
fix(nav-loader): tiny tweaks

damassi:

### Description

With SSR, we need to be sure to not show the loader on mount.


src/System/Router/RenderStatus.tsx
```diff
@@ -14,26 +14,37 @@ import { PageLoadingBar } from "System/Router/PageLoadingBar"
...
     </>
   )
 }
 
 export const RenderReady = ({ elements }: { elements: React.ReactNode }) => {
+  if (!isInitialized) {
+    setTimeout(() => {
+      isInitialized = true
+    }, 0)
+  }
```
damassi, line 37: ugly, but can't use hooks at this level of the stack.

PR #13989, opened 2024-06-13
feat: New submission flow details step

olerichter00:
Addresses [ONYX-1017]

## Description

This implements the **Details step** for the new submission flow.

| desktop | mWeb |
| --- | --- |
|Screenshot 2024-06-13 at 11 19 11 | Screenshot 2024-06-13 at 11 19 35|

[ONYX-1017]: https://artsyproduct.atlassian.net/browse/ONYX-1017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Sell/Routes/DetailsRoute.tsx
+++ b/src/Apps/Sell/Routes/DetailsRoute.tsx
@@ -82,7 +82,7 @@ export const DetailsRoute: React.FC<DetailsRouteProps> = props => {
 
           <Join separator={<Spacer y={4} />}>
             <GridColumns>
-              <Column span={4}>
+              <Column span={[6, 4]}>
                 <Input
                   onChange={handleChange}
                   name="year"
--- a/src/Apps/Sell/Routes/DetailsRoute.tsx
+++ b/src/Apps/Sell/Routes/DetailsRoute.tsx
@@ -1,14 +1,23 @@
-import * as React from "react"
-import * as Yup from "yup"
-import { DetailsRoute_submission$key } from "__generated__/DetailsRoute_submission.graphql"
-import { Input, Join, Select, Spacer, Text } from "@artsy/palette"
+import {
+  Column,
+  GridColumns,
+  Input,
+  Join,
+  Select,
+  Spacer,
+  Text,
+} from "@artsy/palette"
 import { acceptableCategoriesForSubmission } from "Apps/Consign/Routes/SubmissionFlow/Utils/acceptableCategoriesForSubmission"
-import { graphql, useFragment } from "react-relay"
-import { Formik } from "formik"
-import { ConsignmentSubmissionCategoryAggregation } from "__generated__/UpdateConsignSubmissionMutation.graphql"
 import { DevDebug } from "Apps/Sell/Components/DevDebug"
-import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
 import { SubmissionLayout } from "Apps/Sell/Components/SubmissionLayout"
+import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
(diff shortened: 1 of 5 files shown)


src/Apps/Sell/Routes/__tests__/ArtistRoute.jest.tsx
```diff
@@ -21,79 +19,74 @@ jest.mock("react-relay", () => ({
...
 })
 
 const { renderWithRelay } = setupTestWrapperTL({
   Component: (props: any) => {
-    return <ArtistRoute submission={props.submission} />
+    return (
+      <SubmissionRoute submission={props.submission}>
+        <ArtistRoute submission={props.submission} />
+      </SubmissionRoute>
+    )
```
olerichter00, line 38: Updating the Artist route test to wrap the step route with `SubmissionRoute`, which includes the context provider.

src/Apps/Sell/Routes/DetailsRoute.tsx
```diff
@@ -52,6 +61,12 @@ export const DetailsRoute: React.FC<DetailsRouteProps> = props => {
     medium: submission.medium ?? "",
   }
 
+  const yearInputRef = React.useRef<HTMLInputElement | null>(null)
+
+  useEffect(() => {
+    yearInputRef.current?.focus()
+  }, [])
```
olerichter00, line 68: Focusing the year input when the user enters the screen.

src/Apps/Sell/Routes/__tests__/DetailsRoute.jest.tsx
```diff
@@ -0,0 +1,162 @@
...
+              category: "PAINTING",
+              medium: '<mock-value-for-field-"medium">',
+            },
+          },
+        })
+      )
+    })
+  })
+
+  it("does not save the submission & navigates to the previous step when the back button is clicked", async () => {
```
olerichter00, line 115: This is something I'd like to discuss later. Maybe we should also save the submission state when the user clicks on "Back". 🤔

dariakoko (approved):

tam-kis (dismissed):

src/Apps/Sell/Routes/__tests__/DetailsRoute.jest.tsx
tam-kis: Yeah, I think it would make sense to save the state and not start over again.

dblandin: Yeah, I agree that it would make sense to also persist any unsaved changes before navigating back. In that case, we might want to reconsider the design treatment of the back control to also take a button form so we can use a button loading state during the mutation/transition.

src/Apps/Sell/Routes/DetailsRoute.tsx
dblandin: This seems like something we should make a broad UX decision about across steps. When navigating to a step, should we auto-focus on the first input? Would we want that behavior on both desktop and mobile web? I feel like this might result in a subpar experience on mWeb where the keyboard would automatically raise in response to an active input. This might limit our ability to take advantage to utilize the full screen to communicate any framing/guidance on the step.

dariakoko:
@olerichter00 just noticed -- the `Year` field should be half the screen, not full screen
image

src/Apps/Sell/Routes/__tests__/DetailsRoute.jest.tsx
olerichter00: https://artsyproduct.atlassian.net/browse/ONYX-1072

PR #14013, opened 2024-06-14
chore(storybooks): Remove CI deploys

damassi:

### Description

Storybooks is useful for local dev, but we don't need to deploy it (and also no reason to occupy CI runtime).


dzucconi (approved):
Thanks much

PR #14017, opened 2024-06-17
chore(page-loader): Improve it a bit more

damassi:

### Description

After updating the page loader, it really brought to light a lot of our slow loading pages, particularly around collections (which was kinda masked before). This improves the loader a bit by never really stopping -- there's always motion -- as well as animating out from where it was currently at. This was possible by leveraging the `match.elements == undefined` recent discovery which means the page is loading, and moving the bar into the appshell, vs being depending on the weird found router render states.

https://github.com/artsy/force/assets/236943/f9a27c53-7af1-4f3a-bf38-49808fca1178

### DIFF
--- a/src/System/Components/PageLoadingBar.tsx
+++ b/src/System/Components/PageLoadingBar.tsx
@@ -2,6 +2,8 @@ import styled, { keyframes, css } from "styled-components"
 import { useState, useEffect, useRef } from "react"
 import { Box } from "@artsy/palette"
 import { isServer } from "Server/isServer"
+import { useDidMount } from "Utils/Hooks/useDidMount"
+import { Z } from "Apps/Components/constants"
 
 interface PageLoadingBarProps {
   loadingState: "resting" | "loading" | "complete"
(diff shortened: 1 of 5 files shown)


damassi:
@dzucconi - curious to get your take. Prod feels better on pages that load fast like artwork / artist pages, but there's so many collections, and the current version makes those feel way slower, compared to this one, especially with the new nav bar item. So i'm inclined to go with this (in addition to the implementation being better).

src/System/Components/PageLoadingBar.tsx
```diff
@@ -25,26 +27,43 @@ export const PageLoadingBar: React.FC<PageLoadingBarProps> = ({
...
+        backgroundColor="brand"
+        height={2}
+        top="1px"
+        left={0}
+        width={isComplete ? barRef.current?.clientWidth : 0}
+        zIndex={100}
+        overflow="hidden"
+        position="fixed"
+        onAnimationEnd={() => {
+          if (isComplete) {
```
dzucconi, line 56: What happens here if it's not complete?

dzucconi (approved):
Oh yeah, didn't notice the old indicator just stopped. Definitely don't want that.

src/System/Components/PageLoadingBar.tsx
damassi: opacity starts at 0 to 1

damassi: hard to get into this state, but will continue looking at it

PR #14022, opened 2024-06-17
chore: Use logger everywhere in the new Sell flow

olerichter00:
Resolves [Slack Thread](https://artsy.slack.com/archives/C05EQL4R5N0/p1718379105202439)
## Description

This replaces `console.error` with `logger` everywhere in the new Sell flow.


damassi (approved):
thank you!

PR #14030, opened 2024-06-17
feat(ONYX-1018): add purchase history step

dariakoko:

This PR solves [ONYX-1018]

### Description

Add purchase history step
Also, updated Radio button to support boolean values: [PR1](https://github.com/artsy/palette/pull/1390), [PR2 (follow-up)](https://github.com/artsy/palette/pull/1391) 

| web | mWeb |
| --- | --- | 
| photo-size-2-5440399181257758808-y | photo-size-2-5440399181257758809-y |

[ONYX-1018]: https://artsyproduct.atlassian.net/browse/ONYX-1018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Components/SavedAddressItem.tsx
```diff
@@ -19,7 +19,7 @@ interface SavedAddressItemProps {
 
 export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
   props
-): React.ReactElement<RadioProps> => {
+): React.ReactElement<RadioProps<string>> => {
```
dariakoko, line 22: @erikdstock hi! Could you, please, confirm that I am using the right type here?
 The radio button component on Palette was changed and we need to specify the value type now

src/Apps/Order/Routes/Shipping2/Components/SavedAddressItem2.tsx
```diff
@@ -18,7 +18,7 @@ interface SavedAddressItemProps extends BoxProps {
 
 export const SavedAddressItem: React.FC<SavedAddressItemProps> = (
   props
-): React.ReactElement<RadioProps> => {
+): React.ReactElement<RadioProps<string>> => {
```
dariakoko, line 21: @erikdstock and here

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -189,9 +189,9 @@ export const PaymentContent: FC<Props> = props => {
  */
 const getAvailablePaymentMethods = (
   availablePaymentMethods: readonly CommercePaymentMethodEnum[]
-): ReactElement<RadioProps>[] => {
+): ReactElement<RadioProps<string>>[] => {
```
dariakoko, line 192: Here it's pretty straight forward :)

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
```diff
@@ -2,24 +2,49 @@ import * as React from "react"
...
+  "Purchased directly from gallery",
+  "Purchased directly from artist",
+  "Purchased at auction",
+  "Gift from the artist",
+  "Other",
+  "I don’t know",
+].map(provenance => ({
+  value: provenance,
+  text: provenance,
+}))
```
dariakoko, line 42: For the provenance list here I used the same values as we use on the app. [In Figma for Web the texts are slightly different](https://www.figma.com/design/E6rVm1tiSI4PW3EzibKKZw/Submission-%26-Pre-Consignment-Management?node-id=2036-30651&m=dev). Should I use those from Figma or should we stay consistent and keep the same wording across platforms? cc @olerichter00 @MounirDhahri

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
```diff
@@ -11,15 +20,30 @@ import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
...
+    signature
   }
 `
 
 const Schema = Yup.object().shape({
   provenance: Yup.string().trim(),
+  signature: Yup.boolean(),
 })
 
+export const PROVENANCE_LIST = [
```
olerichter00, line 32: I wonder if we could move it to a separate helper file to keep the component file a bit smaller. Something like `src/Apps/Sell/Utils/formUtils.ts`.

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
olerichter00: I think we should align the wording in the app and on the web. Maybe you can ask which version would be better and then update the values here or in the app to keep them consistent.

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
```diff
@@ -45,22 +70,52 @@ export const PurchaseHistoryRoute: React.FC<PurchaseHistoryRouteProps> = props =
...
+          <Join separator={<Spacer y={4} />}>
+            <Select
+              title="Purchase information"
+              name="provenance"
+              options={PROVENANCE_LIST}
+              selected={values.provenance}
+              onChange={handleChange}
+              data-testid="provenance-input"
+            />
+
```
olerichter00, line 88: 🙏

src/Apps/Sell/Routes/__tests__/PurchaseHistoryRoute.jest.tsx
```diff
@@ -0,0 +1,177 @@
...
+              provenance: "Other",
+              signature: false,
+            },
+          },
+        })
+      )
+    })
+  })
+
+  it("does not save the submission & navigates to the previous step when the back button is clicked", async () => {
```
olerichter00, line 136: This test we might need to update because now clicking on "Back" will save the submission as well. Rebasing should make this test fail.

olerichter00 (approved):
Looks good 🌟

src/Apps/Sell/Routes/__tests__/PurchaseHistoryRoute.jest.tsx
dariakoko: Oki, I will update it, thanks!

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
```diff
@@ -11,15 +20,30 @@ import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
 const FRAGMENT = graphql`
   fragment PurchaseHistoryRoute_submission on ConsignmentSubmission {
     provenance
+    signature
   }
 `
 
 const Schema = Yup.object().shape({
   provenance: Yup.string().trim(),
+  signature: Yup.boolean(),
```
MounirDhahri, line 29: shouldn't this be nullable since it's not required

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
```diff
@@ -11,15 +20,30 @@ import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
...
+  "I don’t know",
+].map(provenance => ({
+  value: provenance,
+  text: provenance,
+}))
+
 interface FormValues {
-  provenance
+  provenance: string
+  signature: boolean | undefined
```
MounirDhahri, line 46:
```suggestion
  signature: boolean | null
```

MounirDhahri (approved):

src/Apps/Sell/Routes/PurchaseHistoryRoute.tsx
olerichter00: Good idea to match the interface with the Metaphysics schema. The only problem could be that the input components might not accept `null` values.

olerichter00 (dismissed):

PR #14033, opened 2024-06-17
chore(DIA-718): add elements to auction sidebar 

pam-:

This PR solves [PROJECT-718]

### Description

This PR adds a new element to the auction sidebar and spaces the boxes more:
Screenshot 2024-06-17 at 4 55 30 PM

### DIFF
--- a/src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
+++ b/src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
@@ -10,7 +10,7 @@ interface AuctionInfoSidebarProps {
 
 const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
   return (
-    <Join separator={<Spacer y={1} />}>
+    <Join separator={<Spacer y={2} />}>
       <LiveAuctionToolTip show={!!sale.liveStartAt} />
 
       <Box>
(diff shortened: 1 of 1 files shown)


src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
```diff
@@ -39,6 +39,17 @@ const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
           <Text variant="sm">Supplemental Conditions of Sale</Text>
         </RouterLink>
       </Box>
+
+      <Box>
+        <Text variant="sm">Have a work you want to sell?</Text>
+
+        <Text>
+          Submit to our team {}
```
dzucconi, line 47:
```suggestion
          Submit to our team{' '}
```

is a bit less weird

src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
```diff
@@ -39,6 +39,17 @@ const AuctionInfoSidebar: React.FC<AuctionInfoSidebarProps> = ({ sale }) => {
           <Text variant="sm">Supplemental Conditions of Sale</Text>
         </RouterLink>
       </Box>
+
+      <Box>
```
dzucconi, line 43: Could also make this the parent `<Text variant="sm>`, then just use a `<br />` instead of 2 child elements. Does not really matter at all in this case. Sometimes we just have a lot of extra divs in grids.

dzucconi (approved):

src/Apps/Auction/Components/AuctionDetails/AuctionInfoSidebar.tsx
pam-: sure i can do that. izipizi!

PR #14041, opened 2024-06-20
feat: New Sell flow progress bar

olerichter00:
Addresses [ONYX-1023]

## Description

This implements the progress bar for the new Sell flow.

Screenshot 2024-06-20 at 13 05 51

Screenshot 2024-06-20 at 14 23 31

[ONYX-1023]: https://artsyproduct.atlassian.net/browse/ONYX-1023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Sell/Components/BottomFormNavigation.tsx
+++ b/src/Apps/Sell/Components/BottomFormNavigation.tsx
@@ -52,14 +52,14 @@ const BottomFormNextButton = () => {
   const { isValid, isSubmitting, submitForm } = useFormikContext()
   const {
     actions,
-    state: { isLastStep },
+    state: { isSubmitStep },
   } = useSellFlowContext()
 
   const onNext = async () => {
     try {
       await submitForm()
 
-      isLastStep ? actions.finishFlow() : actions.goToNextStep()
+      isSubmitStep ? actions.finishFlow() : actions.goToNextStep()
     } catch (error) {
       logger.error("Error submitting form", error)
     }
@@ -72,7 +72,7 @@ const BottomFormNextButton = () => {
       loading={isSubmitting}
       onClick={onNext}
     >
-      {isLastStep ? "Submit" : "Continue"}
+      {isSubmitStep ? "Submit" : "Continue"}
     </Button>
   )
 }
--- a/src/Apps/Sell/SellFlowContext.tsx
+++ b/src/Apps/Sell/SellFlowContext.tsx
@@ -101,10 +101,17 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
     push(`/sell2/submissions/${submissionID}/${STEPS[index]}`)
   }, [push, submissionID, index])
 
+  const getSubmissionState = () => {
+    return state.isSubmitStep ? "SUBMITTED" : "DRAFT"
+  }
+
   const createSubmission = (values: CreateSubmissionMutationInput) => {
     const response = submitCreateSubmissionMutation({
       variables: {
-        input: values,
+        input: {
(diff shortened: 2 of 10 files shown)


src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -52,14 +52,14 @@ const BottomFormNextButton = () => {
...
-    state: { isLastStep },
+    state: { isSubmitStep },
   } = useSellFlowContext()
 
   const onNext = async () => {
     try {
       await submitForm()
 
-      isLastStep ? actions.finishFlow() : actions.goToNextStep()
+      isSubmitStep ? actions.finishFlow() : actions.goToNextStep()
```
olerichter00, line 62: I've moved the "Thank You" page inside the steps (at the end) to render the progress bar correctly. This means that we need to define the step that submits the submission differently because it's not the last step anymore.

src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -97,10 +101,17 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
     push(`/sell2/submissions/${submissionID}/${STEPS[index]}`)
   }, [push, submissionID, index])
 
+  const getSubmissionState = () => {
+    return state.isSubmitStep ? "SUBMITTED" : "DRAFT"
+  }
```
olerichter00, line 106: Sorry for sneaking this update into this PR. We want to set the state of the submission depending on the step.

src/Apps/Sell/sellRoutes.tsx
```diff
@@ -284,6 +267,24 @@ export const sellRoutes: RouteProps[] = [
...
+              query sellRoutes_ThankYouRouteQuery($id: ID!) {
+                submission(id: $id) @principalField {
+                  ...ThankYouRoute_submission
+                }
+              }
+            `,
+            prepareVariables: ({ id }) => {
+              return { id }
+            },
+          },
```
olerichter00, line 287: Moving the thank-you route inside of the steps to render the progress bar for this step as well.

src/Apps/Sell/SellFlowContext.tsx
dblandin: It feels a bit excessive to pass a `state` value for every create/update mutation request. This might also result in submissions fliping from submitted > draft > submission if a user goes back via the URL to update submission info.

I wonder if we could instead update the `finishFlow` function to fire a second updateSubmission mutation with just `state: SUBMITTED`. This will likely become conditional in future iterations as we hope to enable collectors to update information in their submission without necessarily re-submitting.

olerichter00: Good catch! Probably not a good idea to set the state to "DRAFT" again when going back after submitting. I've updated the context so the state is only set when clicking "Submit" in the last step. This works because it looks like `step` defaults to "DRAFT" if it's not set explicitly.

olerichter00: https://github.com/artsy/force/pull/14041/commits/a76cdef72baff5d65e811fb0cf92ffd03b1f08bf

dariakoko (approved):
⭐

src/Apps/Sell/SellFlowContext.tsx
dblandin: Nice 👍🏼

PR #14043, opened 2024-06-20
feat(ONYX-1021): add thank you step

dariakoko:

This PR solves [ONYX-1021]

### Description

Add thank you step to the new sell flow
| web | mWeb |
| --- | --- |
| image | image | 

[ONYX-1021]: https://artsyproduct.atlassian.net/browse/ONYX-1021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Sell/Routes/ThankYouRoute.tsx
```diff
@@ -1,41 +1,57 @@
...
+
+              <Button
+                // @ts-ignore
+                as={RouterLink}
+                to="/collector-profile/my-collection"
+                variant="secondaryBlack"
+                width="100%"
+              >
+                View Artwork in My Collections
+              </Button>
```
dariakoko, line 50: This link doesn't work as expected -- instead of being navigated to My Collection when pressing the button, i see the url changing into `/collector-profile/my-collection` and then back to `/thank-you`. I think it might be because we moved this step in [this PR](https://github.com/artsy/force/pull/14041/files#diff-91221e8e1ff465fde08b9f4df582f30da99522b0efe825318ebb2594065d26f8R270). Do you know what has to be done for the link to work as expected?

src/Apps/Sell/Routes/ThankYouRoute.tsx
```diff
@@ -1,41 +1,57 @@
...
-      <Spacer y={2}/>
-      <Message variant="info" title="What happens next?">
-        <Text variant="xs">
-          If your artwork is accepted, we will send you a sales offer and guide
-          you in selecting the best selling option. Additional information may be
-          requested.
-        </Text>
-      </Message>
-    </SubmissionLayout>
+              <Message variant="success" title="What happens next?">
```
dariakoko, line 24: According to the design, we should use a smaller title text in the Message component. We cannot do this because the component doesn't have the support for custom styles

olerichter00: Let's use the default message design and add a comment in Figma to explain that the design deviates from Palette. Like this, we can keep the design consistent throughout Force.

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
```diff
@@ -0,0 +1,15 @@
+import { ThankYouRoute } from "Apps/Sell/Routes/ThankYouRoute"
+import { mount } from "enzyme"
```
olerichter00, line 2: Let's use `render` or `setupTestWrapperTL` from `DevTools/setupTestWrapper`.

```suggestion
import { render } from "DevTools/renderWithMockBoot"

import { setupTestWrapperTL } from "DevTools/setupTestWrapper"

```

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
```diff
@@ -0,0 +1,15 @@
...
+describe("ThankYouRoute", () => {
+  it("renders text and links", () => {
+    const wrapper = mount(<ThankYouRoute />)
+    const text = wrapper.text()
+
+    expect(text).toContain("Thank you for submitting your artwork")
+
+    expect(text).toContain("Submit Another Work")
+
+    expect(text).toContain("View Artwork in My Collections")
```
olerichter00, line 13:
```suggestion
    expect(screen.getByText("Submit Another Work").attributes["href"].value).toBe("/bar")

    expect(screen.getByText("View Artwork in My Collections").attributes["href"].value).toBe("/fooooo")
```

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
```diff
@@ -0,0 +1,15 @@
+import { ThankYouRoute } from "Apps/Sell/Routes/ThankYouRoute"
+import { mount } from "enzyme"
+
+describe("ThankYouRoute", () => {
+  it("renders text and links", () => {
+    const wrapper = mount(<ThankYouRoute />)
```
olerichter00, line 6: To make sure the component renders without any issues when wrapped with `SubmissionRoute`.

```suggestion
const { renderWithRelay } = setupTestWrapperTL({
  Component: (props: any) => {
    return (
      <SubmissionRoute submission={props.submission}>
        <ThankYouRoute />
      </SubmissionRoute>
    )
  },
  query: graphql`
    query ArtistRoute_Test_Query @raw_response_type {
      submission(id: "submission-id") {
        ...SubmissionRoute_submission
      }
    }
  `,
})
```

src/Apps/Sell/Routes/ThankYouRoute.tsx
```diff
@@ -1,41 +1,57 @@
...
-  children: React.ReactNode
-}
+export const ThankYouRoute: React.FC = () => {
+  return (
+    <FullBleed>
+      <AppContainer>
+        <SubmissionLayout hideNavigation>
+          <Join separator={<Spacer y={4} />}>
+            <Join separator={<Spacer y={2} />}>
+              <Text variant="lg-display">
```
olerichter00, line 14: [Figma](https://www.figma.com/design/E6rVm1tiSI4PW3EzibKKZw/Submission-%26-Pre-Consignment-Management?node-id=2036-30767&m=dev) 👀 

```suggestion
              <Text variant={["lg-display", "xl"]}>
```

olerichter00: We might need to update this for each step or even extract the title component.

olerichter00 (approved):
This looks great! Just a few comments and suggestions for the test.

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
dariakoko: I try to use it like this 
```
expect(screen.getByText("Submit Another Work")).toHaveAttribute(
      "href",
      "/sell2/submissions/new"
    )
```
And I get back this
```
    Expected the element to have attribute:
      href="/sell2/submissions/new"
    Received:
      null
```
I don't understand what's wrong here. Can you stop a mistake? 👀

dariakoko: When I use the syntaxis that you suggested it returns 
``` TypeError: Cannot read properties of undefined (reading 'value') ```

src/Apps/Sell/Routes/ThankYouRoute.tsx
dariakoko: Will create a separate ticket for it

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
olerichter00: Hmm, in [this](https://github.com/artsy/force/blob/f23c8def6659d9408216b1546833ce2dbc92a305/src/Apps/Sell/Routes/__tests__/IntroRoute.jest.tsx#L55) test file, it works. Maybe worth looking into what's different in the test setup (or how the link is constructed).

src/Apps/Sell/Routes/ThankYouRoute.tsx
olerichter00: Or just a PR, I don't think it's much work 🐈  Could even be done in this PR.

src/Apps/Sell/Routes/ThankYouRoute.tsx
olerichter00: Hmm, I really don't know 🤔 Maybe this is something for #dev-help

src/Apps/Sell/Routes/ThankYouRoute.tsx
olerichter00: To keep Jira clean and clear 👀

dariakoko: I would prefer to create a separate ticket for this because this pr has issues

src/Apps/Sell/Routes/__tests__/ThankYouRoute.jest.tsx
dariakoko: Daaamn I should have used getByTestId
Thanks!

src/Apps/Sell/Routes/ThankYouRoute.tsx
olerichter00: I've written it down and will update the title font later 👍

dariakoko: Thank youu, but no neeed, I've just opened a pr https://github.com/artsy/force/pull/14047

olerichter00 (approved):

PR #14046, opened 2024-06-20
feat: offer card registration only to those with saved payment methods (PHIRE-1013)

joeyAghion:
Makes ability to manage payment methods more strict based on [discussion here](https://artsy.slack.com/archives/CA8SANW3W/p1718713045925899).

https://artsyproduct.atlassian.net/browse/PHIRE-1013


starsirius (approved):
💪

PR #14047, opened 2024-06-21
fix: adjust title sizes on the submission screens

dariakoko:

This PR solves [PROJECT-XX]

### Description

Adjust title sizes on the submission screens

### DIFF
--- a/src/Apps/Sell/Components/SubmissionStepTitle.tsx
+++ b/src/Apps/Sell/Components/SubmissionStepTitle.tsx
@@ -1,0 +1,9 @@
+import { Text } from "@artsy/palette"
+
+export const SubmissionStepTitle: React.FC<{ title: string }> = ({ title }) => {
+  return (
+    <Text variant={["lg-display", "xl"]} mb={2}>
+      {title}
+    </Text>
+  )
+}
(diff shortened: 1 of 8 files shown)


src/Apps/Sell/Components/SubmissionStepTitle.tsx
```diff
@@ -0,0 +1,9 @@
+import { Text } from "@artsy/palette"
+
+export const SubmissionStepTitle: React.FC<{ title: string }> = ({ title }) => {
+  return (
+    <Text variant={["lg-display", "xl"]} mb={2}>
+      {title}
+    </Text>
+  )
+}
```
olerichter00, line 9: Both versions are okay, but I would prefer this one, which is a bit more standard React.


```suggestion
export const SubmissionStepTitle: React.FC = ({ children }) => {
  return (
    <Text variant={["lg-display", "xl"]} mb={2}>
      {children}
    </Text>
  )
}
```

olerichter00 (dismissed):
Thanks 🙏 

You were faster than me 🦥 

https://github.com/artsy/force/pull/14048

src/Apps/Sell/Components/SubmissionStepTitle.tsx
dariakoko: Agree, looks better 👍

PR #14049, opened 2024-06-21
feat: defer forced login / sign up to end of SWA submission flow

dblandin:

This PR addresses [ONYX-1089](https://artsyproduct.atlassian.net/browse/ONYX-1089)

### Description

Instead of forcing a user to login or sign up directly on the /sell landing page, we'd like to allow the user to start and progress through the submission and only force an authenticated state at the end, right before the submission graduates from draft > submitted. At that point in the flow we want to make sure we have an Artsy user account to link the submission to but we're okay with letting visitor / unauthenticated users start and experience the submission flow before the final step.

### Screen Recordings

| Title | Video |
|--------|--------|
| Current (Visitor) | <video src="https://github.com/artsy/force/assets/123595/dd0d1364-f50d-4c5c-a0a4-4833d9b07717" /> |
| New (Vistor) | <video src="https://github.com/artsy/force/assets/123595/c4642a74-c1f6-42e3-b004-7f9150ff7c92" /> | 
| New (User) | <video src="https://github.com/artsy/force/assets/123595/88250efd-4815-48ac-93cc-ec13e8b20c85" /> |

[ONYX-1089]: https://artsyproduct.atlassian.net/browse/ONYX-1089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
@@ -1,4 +1,3 @@
-import { ContextModule, Intent } from "@artsy/cohesion"
 import {
   Button,
   Column,
@@ -9,17 +8,13 @@ import {
   Text,
 } from "@artsy/palette"
 import { useMarketingLandingTracking } from "Apps/Consign/Routes/MarketingLanding/Utils/marketingLandingTracking"
-import { useAuthDialog } from "Components/AuthDialog"
 import { RouterLink } from "System/Components/RouterLink"
 import { useFeatureFlag } from "System/Hooks/useFeatureFlag"
-import { useSystemContext } from "System/Hooks/useSystemContext"
 import { Media } from "Utils/Responsive"
 import { resized } from "Utils/resized"
 
 export const HeaderSWA = () => {
   const getInTouchRoute = "/sell/inquiry"
-  const { isLoggedIn } = useSystemContext()
-  const { showAuthDialog } = useAuthDialog()
 
   const {
     trackStartSellingClick,
@@ -61,24 +56,6 @@ export const HeaderSWA = () => {
                   enableNewSubmissionFlow ? "sell2/intro" : "/sell/submission"
                 }
                 onClick={event => {
-                  if (!isLoggedIn) {
-                    event.preventDefault()
-
-                    showAuthDialog({
-                      mode: "Login",
-                      options: {
-                        title: () => {
-                          return "Log in to submit an artwork for sale"
-                        },
-                      },
-                      analytics: {
-                        contextModule: ContextModule.sellHeader,
-                        intent: Intent.login,
-                      },
-                    })
-
-                    return
-                  }
                   trackStartSellingClick("Header")
                 }}
                 mb={[1, 0]}
--- a/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
+++ b/src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
@@ -397,12 +397,17 @@ export const UploadPhotos: React.FC<UploadPhotosProps> = ({
                     event.preventDefault()
 
                     showAuthDialog({
-                      mode: "Login",
+                      mode: "SignUp",
                       options: {
                         title: mode =>
                           mode === "Login"
                             ? "Log in to submit for sale"
                             : "Sign up to submit for sale",
+                        afterAuthAction: {
+                          action: "associateSubmission",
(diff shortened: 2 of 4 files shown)


src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
```diff
@@ -384,6 +392,32 @@ export const UploadPhotos: React.FC<UploadPhotosProps> = ({
...
+                      mode: "Login",
+                      options: {
+                        title: mode =>
+                          mode === "Login"
+                            ? "Log in to submit for sale"
+                            : "Sign up to submit for sale",
+                          afterAuthAction: {
+                            action: "associateSubmission",
+                            kind: "submission",
+                            objectId: submission?.externalId!,
```
dblandin, line 409: Note that while this "associateSubmission" afterAuthAction is already defined, it doesn't seem to be used and currently assumes that we've already collected the visitor's email address in the flow. We recently removed the Contact Information step in the flow so we no longer collect the email address in the UI.

This change will need to be paired with a corresponding PR in Convection which currently defines logic to allow visitor submissions to be later associated with an Artsy account. I'll provide the link to that PR here when ready.

dblandin: PR for that Convection change [is here](https://github.com/artsy/convection/pull/1480).

MounirDhahri: thanks for the comment

MounirDhahri: can we also navigate the user to `Thank your submitting` step after the sign in/up and we associate the submission to their user.

MounirDhahri (dismissed):
LGTM

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
```diff
@@ -384,6 +392,32 @@ export const UploadPhotos: React.FC<UploadPhotosProps> = ({
                 disabled={isSubmitting || !isValid}
                 loading={isSubmitting || values.photos.some(c => c.loading)}
                 type="submit"
+                onClick={event => {
+                  if (!isLoggedIn) {
+                    event.preventDefault()
+
+                    showAuthDialog({
+                      mode: "Login",
```
MounirDhahri, line 400: Since we don't often log out users from their sessions, I wonder if it makes more sense here to present the `Sign Up` mode instead

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
```diff
@@ -61,24 +56,6 @@ export const HeaderSWA = () => {
                   enableNewSubmissionFlow ? "sell2/intro" : "/sell/submission"
                 }
                 onClick={event => {
-                  if (!isLoggedIn) {
```
olerichter00, line 64: Removing the auth check here will also affect the new Sell flow (`sell2/intro`). This should be fine because the new flow has an additional check on the Artist step.

Screenshot 2024-07-01 at 11 46 31

olerichter00 (approved):
Looks good 🌟

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
dblandin: Yeah, that's a good idea. We can default to the sign up modal instead. Existing users can easily switch to login from that. 👍🏼

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
dblandin: It's a bit tricky to add more after auth behavior here. We'd have to 1) associate the submission 2) submit the submission and 3) navigate to next step, all in the after auth callback.

As this will be soon be replaced by the redesigned flow, let's add any new behavior to that instead of adding to the behavior here.

src/Apps/Consign/Routes/SubmissionFlow/UploadPhotos/UploadPhotos.tsx
dblandin: Switched to `Signup` mode in last commit 👍🏼

PR #14044, opened 2024-06-20
feat: stub in a budget inference step

anandaroop:
The type of this PR is: **Feat**

This begins the step of budget inference from the user-provided budget free text.

Mostly this PR just lays down plumbing for:

- an SPA client, where the `Form` is now the first screen and the `Result` is the second screen
- backend server endpoints, starting with a stub for budget inferences

We can merge this in its current form, so as to unblock other results-related work; and I will iterate on the actual LLM logic in a separate PR.

https://github.com/artsy/force/assets/140521/4e5c27d5-93ed-462c-8886-8f603da1658b

PR #14050, opened 2024-06-21
feat: wire up marketing collections rail to 07-Curated-Discovery

mc-jones:
The type of this PR is: Feat

### Description

This PR wires up the marketing collections rail to the `07-Curated-Discovery` result page, taking input from the form page. Similar to #14044, this isn't the final representation of the collections rail, but stubs it out and allows us to keep iterating and avoid duplicative work.

I've punted on UI polish considerations.

---

https://github.com/artsy/force/assets/38149304/1452330a-d51e-4ab9-88f5-dde44bcb1751

### DIFF
--- a/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/MarketingCollectionsRail.tsx
+++ b/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/MarketingCollectionsRail.tsx
@@ -1,0 +1,70 @@
+import { Box, SkeletonBox } from "@artsy/palette"
+import { State } from "Apps/ArtAdvisor/07-Curated-Discovery/App"
+import { MarketingCollection } from "Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/MarketingCollection"
+import { Rail } from "Components/Rail/Rail"
+import { FC, useEffect, useState } from "react"
+
+interface MarketingCollectionsRailProps {
+  state: State
+}
+
+export interface DiscoveryMarketingCollections {
+  id: string
+  imageUrl: string
+  slug: string
+  title: string
+}
+
+export const MarketingCollectionsRail: FC<MarketingCollectionsRailProps> = ({
+  state,
+}) => {
+  const [marketingCollections, setMarketingCollections] = useState<
+    DiscoveryMarketingCollections[]
+  >([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const params = new URLSearchParams()
+    params.append("concept", state.goal) //TODO: improve use of goal
+    state.interests.forEach(concept => {
+      params.append("concepts", concept)
+    })
+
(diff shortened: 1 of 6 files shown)


src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/MarketingCollectionsRail.tsx
```diff
@@ -0,0 +1,70 @@
...
+  state,
+}) => {
+  const [marketingCollections, setMarketingCollections] = useState<
+    DiscoveryMarketingCollections[]
+  >([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const params = new URLSearchParams()
+    params.append("concepts", state.goal) //TODO: improve use of goal
```
mc-jones, line 28: In this first iteration, we're just adding the goal to the list of concepts to use in the `nearText` search. I will experiments with potentially better options, as this doesn't seem ideal (e.g. reranking based on `goal`, which will require a infra update to enable to module, unless we manually git a reranker).

src/Apps/ArtAdvisor/07-Curated-Discovery/server.ts
```diff
@@ -1,11 +1,31 @@
...
 
 const inferBudgetIntent = async (req: ArtsyRequest, res: ArtsyResponse) => {
   res.json({ amount: 42 })
 }
 
+const getMarketingCollections = async (
+  req: ArtsyRequest,
+  res: ArtsyResponse
+) => {
+  let { concepts, budget } = req.query
```
mc-jones, line 16: While we accept the budge query param, we're not doing anything with it, yet. We will need to wire this up fully once we know exactly what kind of input to expect.

src/Apps/ArtAdvisor/07-Curated-Discovery/weaviate-db.ts
mc-jones, line 1: Ultimately decided on only brining in what is necessary for this PR, right now, to make it a bit cleaner.

src/Apps/ArtAdvisor/07-Curated-Discovery/weaviate-db.ts
```diff
@@ -0,0 +1,80 @@
...
+   */
+  async getNearMarketingCollections({
+    concepts,
+    budget,
+    limit = 10,
+  }: {
+    /** List of concepts for semantic search */
+    concepts: string[]
+    /** Budget on which to filter collections */
+    budget?: number
```
mc-jones, line 38: We need to decide what strategy to use when filtering. We have the `priceGuidance` [field](https://github.com/artsy/gravity/blob/006bb5e97875f0b13a06592121302349f158c5b3/config/initializers/rails_admin.rb#L712) available to us in `weaviate`, so we can use that price in our filtering strategy.

anandaroop (approved):
Very nice.

PR #14054, opened 2024-06-24
feat(onyx-1016): allow photo upload for new sell flow

nickskalkin:

This PR solves [ONYX-1016]

### Description

- [Figma specs](https://www.figma.com/design/E6rVm1tiSI4PW3EzibKKZw/Submission-%26-Pre-Consignment-Management?node-id=2036-35642&t=NXDxkDKgsBYabWjU-0)

>[!NOTE]
>Cannot create a review app, asked for help in dev-help channel. That are a few things I need Barney's opinion on before merging, but I think this PR can be code-reviewed.

| Desktop | mWeb |
|---|---|
| Screenshot 2024-06-24 at 11 45 04 | Screenshot 2024-06-24 at 11 45 34 |

[ONYX-1016]: https://artsyproduct.atlassian.net/browse/ONYX-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Sell/Components/ImagePreviewsGrid.tsx
```diff
@@ -0,0 +1,24 @@
...
+    values: { photos },
+  } = useFormikContext<PhotosFormValues>()
+
+  if (!photos) return null
+
+  return (
+    <>
+      <Spacer y={2} />
+
+      <Flex gap={1} flexWrap="wrap">
```
olerichter00, line 17:
```suggestion
    <Flex gap={1} flexWrap="wrap" my={2}>
```

src/Apps/Sell/Components/UploadPhotosForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const handlePhotoUploadingProgress = useCallback(
+    (photo: Photo) => progress => {
+      photo.progress = progress
+      setFieldValue("photos", values.photos)
+    },
+    [values.photos, setFieldValue]
+  )
+
+  const uploadImage = useCallback(
```
olerichter00, line 34: What's the reason we are using `useCallback` here?

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -2,10 +2,17 @@ import { Box, Button, Flex } from "@artsy/palette"
 import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
 import createLogger from "Utils/logger"
 import { useFormikContext } from "formik"
+import { FC } from "react"
 
 const logger = createLogger("BottomFormNavigation.tsx")
 
-export const BottomFormNavigation = () => {
+interface BottomFormNavigationProps {
+  loading?: boolean
```
olerichter00, line 10: Can we move this loading state into the context (e.g. as `isAssetUploading`) or retrieve the state from Formik (e.g. `form.photos.some(photo => photo.loading)`)?

olerichter00 (commented):
Looks great! I would just try not to pass the new `loading` state as a prop and instead try to get it from Formik or from the Context to avoid prop drilling.

src/Apps/Sell/Components/BottomFormNavigation.tsx
olerichter00: And we might want to at least disable the back button in this case as well to avoid unpredictable behavior.

src/Apps/Sell/Components/UploadPhotosForm.tsx
nickskalkin: I guess I can move these functions to `useEffect` below and add their dependencies to the dependency array. I preferred to use `useCallback` instead (I guess it should be the same, no?) @olerichter00

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -102,8 +110,9 @@ const BottomFormNextButton = () => {
     <Button
       variant="primaryBlack"
       disabled={!isValid}
-      loading={isSubmitting}
+      loading={isSubmitting || loading}
```
olerichter00, line 113: I've just checked, and I think we can remove this additional loading state and derive the state directly from Formik. I think this would make things way simpler because we don't have another (kind of generic) state within the context.

The following code could either be here in the BottomFormNavigation or live in the context.

What do you think?
```
  const { isValid, submitForm, values } = useFormikContext()

  const isPhotoUploading = (values as PhotosFormValues)?.photos?.some(
    photo => photo.loading
  )
  
  ...
  loading={isSubmitting || isPhotoUploading}
  ...
  ```

File: src/Apps/Sell/Routes/PhotosRoute.tsx [hunk 5]
```diff
@@ -40,30 +49,55 @@ interface PhotosRouteProps {
...
+          (photo: Photo) => photo.loading
+        )
+
+        return (
+          <SubmissionLayout loading={isAnyPhotoLoading}>
+            <SubmissionStepTitle>
+              Upload photos of your artwork
+            </SubmissionStepTitle>
+
+            <Text mb={2} variant={["xs", "sm"]}>
```
olerichter00, line 79:
```suggestion
            <Text mb={2} variant={["xs", "sm"]} color="black50">
```

src/Components/PhotoUpload/Components/PhotoDropzone.tsx
```diff
@@ -124,22 +128,22 @@ export const PhotoDropzone: React.FC<PhotoDropzoneProps> = ({
           <Text variant="lg-display">Add photos here</Text>
         </Media>
 
-        <Text variant="sm-display" color="black60" mt={1}>
+        <Text variant={["xs", "sm-display"]} color="black60" mt={1}>
           Files supported: JPG, PNG
         </Text>
-        <Text variant="sm-display" color="black60" mt={1}>
+        <Text variant={["xs", "sm-display"]} color="black60" mt={1}>
```
olerichter00, line 134: It looks like there is too much space between these two lines compared to the Figma design. Maybe we put both in one Text and add a line break between them.

src/Apps/Sell/Components/BottomFormNavigation.tsx
nickskalkin: I am not strongly opinionated, but this was my logic: I didn't want to put a **Photo step**-related code in a generic component, that's why I pass `loading` property from Photo component itself. Imagine we have another component like that - we will need to add more and more checks to the `BottomFormNavigation` component. 

@olerichter00

olerichter00 (approved):

src/Apps/Sell/Components/BottomFormNavigation.tsx
olerichter00: Let's go with your solution and keep the state in the context 👍

dariakoko (dismissed):
👍

PR #14057, opened 2024-06-24
chore: Make "Start Selling" button ready for new sell flow

olerichter00:
Resolves [ONYX-1058](https://artsyproduct.atlassian.net/browse/ONYX-1058)

## Description

Make the "Start Selling" button ready for the new sales flow (behind a feature flag).

Screenshot 2024-06-24 at 13 17 55

[ONYX-1058]: https://artsyproduct.atlassian.net/browse/ONYX-1058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
@@ -11,6 +11,7 @@ import {
 import { useMarketingLandingTracking } from "Apps/Consign/Routes/MarketingLanding/Utils/marketingLandingTracking"
 import { useAuthDialog } from "Components/AuthDialog"
 import { RouterLink } from "System/Components/RouterLink"
+import { useFeatureFlag } from "System/Hooks/useFeatureFlag"
 import { useSystemContext } from "System/Hooks/useSystemContext"
 import { Media } from "Utils/Responsive"
 import { resized } from "Utils/resized"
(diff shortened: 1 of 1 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
```diff
@@ -54,7 +57,11 @@ export const HeaderSWA = () => {
                 as={RouterLink}
                 width="100%"
                 variant="primaryBlack"
-                to="/sell/submission"
+                to={
+                  enableNewSubmissionFlow
+                    ? "sell2/submissions/new"
```
dariakoko, line 62: are we planning to release the feature with the rout `sell2` and rename to `sell` after we remove the feature flag logic?

dariakoko (approved):

src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/HeaderSWA.tsx
MounirDhahri: good point - I think we can update the route when we plan on releasing instead of using a ff here

olerichter00: The reason why I added the feature flag is to get the flow ready for the QA sessions. Like this, we can test the complete flow, from clicking the "Start Selling" button until the end without any confusion.

PR #14058, opened 2024-06-24
fix(EMI-1924): check for offer availability to display banner

leamotta:
This PR solves [EMI-1924]

### Description

For displaying the expired offer banner, instead of not getting a partner offer in the backend response and assuming that the partner offer is expired, we can return the partner offer and show the banner if it's not available. This would cover the case when a user is not logged in.

Screenshot 2024-06-24 at 13 14 19

TODO:
[] Gravity PR to return unavailable partner offers
[] Check Eigen for similar logic


leamotta:
will try a different approach as this would imply changes in Eigen that'd need to be coordinated with backend

PR #14061, opened 2024-06-24
fix(EMI-1924): user logged in to show expired banner

leamotta:

This PR solves [EMI-1924]

### Description

An alternative to #14058 where we only check that the user is logged in to avoid backend + app changes that need to be coordinated

[EMI-1924]: https://artsyproduct.atlassian.net/browse/EMI-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/Components/ArtworkPageBanner.tsx
+++ b/src/Apps/Artwork/Components/ArtworkPageBanner.tsx
@@ -49,7 +49,10 @@ export const ArtworkPageBanner: FC<ArtworkPageBannerProps> = props => {
       return <ArtworkUnavailableBanner />
     }
 
-    if (!partnerOffer || partnerOffer.internalID !== expectedPartnerOfferID) {
+    if (
+      me &&
+      (!partnerOffer || partnerOffer.internalID !== expectedPartnerOfferID)
+    ) {
       return <ExpiredOfferBanner />
     }
   }


src/Apps/Artwork/Components/ArtworkPageBanner.tsx
```diff
@@ -49,7 +49,10 @@ export const ArtworkPageBanner: FC<ArtworkPageBannerProps> = props => {
       return <ArtworkUnavailableBanner />
     }
 
-    if (!partnerOffer || partnerOffer.internalID !== expectedPartnerOfferID) {
+    if (
+      me &&
```
oxaudo, line 53: That is fine I think but I was under impression that we want to ask collectors to log in in case they are trying to view artwork with offer link and happen to be not logged in. This will just skip rendering the banner but will not render request to log in, right?

But I guess if the offer is really expired it might be annoying to log in just to see the expired banner  :(

leamotta: It will render the log in, the logic is in a different file :)
Screenshot 2024-06-24 at 14 09 14

oxaudo: LGTM then! Looks good to me :)

erikdstock: Ah, I was confused about this. Good catch and easy fix.

PR #14064, opened 2024-06-24
feat: add 07 curated discovery articles rail

mc-jones:
This PR follows #14050 and uses it as a template to wire up the articles rail.

### Considerations

- As part of this work, I ingested just under `9,000` articles into a new `weaviate` collection called `DiscoveryArticles` using the `quantum` [scripts](https://github.com/artsy/quantum/tree/main/src/06-positron-rag). I also expanded the script to also return the `imageUrl` (PR forthcoming). 
- In the current implementation of the `DiscoveryArticles` collection, we include "news" articles. And in testing this PR I noticed they tend to be a lot of the top results. It seems like we will want to exclude news articles (an article from 2012 about that years gallery shows probably doesn't have a ton of discovery value). This  should be possible by updating the `articlesConnection` [MP query](https://github.com/artsy/quantum/blob/main/src/06-positron-rag/lib/extract/fetchArticles.ts#L15) in `quantum` with a `isEditorialFeed: true` [argument](https://github.com/artsy/metaphysics/blob/e366721dd750eec827d93575159c084190cc3070/src/schema/v2/articlesConnection.ts#L24-L28). But this can all be done via quantum and won't require any changes to this force PR. 

---
https://github.com/artsy/force/assets/38149304/524e38a1-c395-4c95-b8d6-45a9a4c7f513

### DIFF
--- a/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArticlesRail.tsx
+++ b/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArticlesRail.tsx
@@ -1,0 +1,61 @@
+import { Box, SkeletonBox } from "@artsy/palette"
+import { State } from "Apps/ArtAdvisor/07-Curated-Discovery/App"
+import { Article } from "Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Article"
+import { Rail } from "Components/Rail/Rail"
+import { FC, useEffect, useState } from "react"
+
+interface ArticlesRailProps {
+  state: State
+}
+
+export interface ArticleType {
+  id: string
+  imageUrl?: string
+  href: string
+  title: string
+  articleDescription: string
+}
+
+export const ArticlesRail: FC<ArticlesRailProps> = ({ state }) => {
+  const [articles, setArticles] = useState<ArticleType[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const params = new URLSearchParams()
+    params.append("limit", "100")
+    params.append("concepts", state.goal) //TODO: improve use of goal
+    state.interests.forEach(concept => {
+      params.append("concepts", concept)
+    })
+
+    const fetchArticles = async () => {
+      const response = await fetch(
+        `/api/advisor/7/articles?${params.toString()}`
+      )
+      const data = await response.json()
+      setArticles(data)
+      setIsLoading(false)
+    }
+
+    fetchArticles()
+  }, [state.interests, state.goal])
+
+  return (
+    <>
+      {articles.length ? (
+        <Box opacity={isLoading ? 0.2 : 1}>
+          <Rail
+            title="Artsy Editorial"
+            getItems={() => {
+              return articles.map((article: ArticleType) => {
+                return <Article key={article.id} article={article} />
+              })
+            }}
+          />
+        </Box>
+      ) : (
+        <SkeletonBox height={460} />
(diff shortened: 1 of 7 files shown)


src/Apps/ArtAdvisor/07-Curated-Discovery/weaviate-db.ts
```diff
@@ -60,6 +66,51 @@ export class WeaviateDB {
...
+      .withNearText({
+        concepts: conceptArray,
+      })
+      .withLimit(limit)
+      .withFields(
+        "internalID title articleDescription href imageUrl _additional { id distance }"
+      )
+      .do()
+
+    const result: ArticleType[] = response.data.Get[DEFAULT_ARTICLES_CLASS]
```
anandaroop, line 101: If we're setting the articles class in the config, then technically this would be more correct:

```suggestion
    const result: ArticleType[] = response.data.Get[this.articlesClass]
```

(Same idea on L65 above, my bad for not noticing in the earlier PR)

src/Apps/ArtAdvisor/07-Curated-Discovery/weaviate-db.ts
```diff
@@ -60,6 +66,51 @@ export class WeaviateDB {
...
+
+    /**
+     * Note: Because we return article sections, a single article can appear
+     * multiple times in the results. This removes duplicates based on the
+     * internalID. Order is preserved and the first occurrence is kept.
+     *
+     * TODO: handle better?
+     */
+    const dedupedResults = _.uniqBy(result, "internalID")
```
anandaroop, line 110: Good call — I wonder if this means we might want to overfetch (i.e. more than 10 sections), in order to reduce the chance of getting puny results because of the deduping.

src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArticlesRail.tsx
```diff
@@ -0,0 +1,61 @@
...
+  articleDescription: string
+}
+
+export const ArticlesRail: FC<ArticlesRailProps> = ({ state }) => {
+  const [articles, setArticles] = useState<ArticleType[]>([])
+  const [isLoading, setIsLoading] = useState<boolean>(true)
+
+  useEffect(() => {
+    const params = new URLSearchParams()
+    params.append("limit", "25")
```
anandaroop, line 25: Ah, I think I answered  my own question (just to confirm, that is the reason for the higher limit here yes?)

anandaroop (approved):
Looks great!

(May need a rebase since I just merged #14053)

src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArticlesRail.tsx
mc-jones: Yup! exactly.

src/Apps/ArtAdvisor/07-Curated-Discovery/weaviate-db.ts
mc-jones: Ah, nice catch. I will revert back to what we had before. This was just a quick hack while debugging.

mc-jones: That was a confusing response by me. But have changed both to your suggestion in [430d71b](https://github.com/artsy/force/pull/14064/commits/430d71b846575e3f1c458cb5b842f532f84e1866).

PR #14065, opened 2024-06-24
Fix: return if artist associated with saved alert cannot be found

jpotts244:
The type of this PR is: Fix

### Description
We had a report that /alerts page was returning 500 for some users

After pairing with @anandaroop, this is happening when an artist associated with some Saved alerts gets deleted
Thinking of ways to be a little more defensive on the client side so instead users can access their alerts instead of returning a 500 page exception

### Current experience
An artist for one of my saved alerts was deleted
Screenshot 2024-06-24 at 3 13 14 PM

After some console logging, the `node` itself is coming back as `null` if artist cannot be found, so we have an array of search criteria with null as the last value
Screenshot 2024-06-24 at 3 27 07 PM

### Experience After
Early Return if the node is `null`, that way users can at least view the page
Screenshot 2024-06-24 at 3 17 29 PM


anandaroop (approved):
Definitely an improvement 🙏🏽

PR #14069, opened 2024-06-25
feat(onyx-944): activity panel navigates directly to entity page whenonly one entity in notification row

nickskalkin:

This PR solves [ONYX-944]

### Description

When tapping on an Activity Panel notification containing one artwork (or show/etc), we want to redirect the user directly to the artwork page rather than the notifications app with the notification selected in the preview pane. The only exception I made is Partner offer notification, because it always has 1 object.

| mWeb | Desktop |
|---|---|
| <video src="https://github.com/artsy/force/assets/3934579/92854e7b-2994-434c-ad51-f75416149688" /> |  <video src="https://github.com/artsy/force/assets/3934579/a9ce17ad-bf03-4209-b8d7-7d8e99b17a11" /> |

[ONYX-944]: https://artsyproduct.atlassian.net/browse/ONYX-944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Notifications/NotificationItem.tsx
+++ b/src/Components/Notifications/NotificationItem.tsx
@@ -156,13 +156,13 @@ export const NotificationItemFragmentContainer = createFragmentContainer(
   }
 )
 
-interface NotificationItemLinkProps {
+interface NotificationItemWrapperProps {
   item: NotificationItem_item$data
   mode?: NotificationListMode
   children: React.ReactNode
 }
 
-const NotificationItemWrapper: FC<NotificationItemLinkProps> = ({
+const NotificationItemWrapper: FC<NotificationItemWrapperProps> = ({
   item,
   mode,
   children,
--- a/src/Components/Notifications/NotificationItem.tsx
+++ b/src/Components/Notifications/NotificationItem.tsx
@@ -1,4 +1,4 @@
-import { Flex, Image, Join, Spacer, Text } from "@artsy/palette"
+import { Flex, Image, Join, Spacer, THEME, Text } from "@artsy/palette"
 import { themeGet } from "@styled-system/theme-get"
 import {
   ExpiresInTimer,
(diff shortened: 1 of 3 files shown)


src/Components/Notifications/__tests__/NotificationItem.jest.tsx
```diff
@@ -142,9 +146,81 @@ describe("NotificationItem", () => {
...
+          expect(screen.getByRole("link")).toHaveAttribute(
+            "href",
+            "/target-href"
+          )
+        })
+      })
+    })
+
+    describe("supported notification types", () => {
+      it.each(SUPPORTED_NOTIFICATION_TYPES)(
```
olerichter00, line 200: 👏

src/Components/Notifications/NotificationItem.tsx
```diff
@@ -216,7 +220,23 @@ NotificationItemLink.defaultProps = {
  * Until we support all notification types in the new activity panel,
  * we only link to the notification detail page for the supported types.
  */
-const getNotificationUrl = (notification: NotificationItem_item$data) => {
+const getNotificationUrl = (
+  notification: NotificationItem_item$data,
+  mode: NotificationListMode = "page"
```
olerichter00, line 225: I wonder if we should change this parameter to a boolean (e.g. `alwaysOpenAsNotification`) to decouple this function from the concept of Notification mode.

olerichter00: This would also prevent us from "re-setting" the mode to "dropdown" on smaller screen sizes.

src/Components/Notifications/NotificationItem.tsx
```diff
@@ -17,18 +17,22 @@ import { createFragmentContainer, graphql } from "react-relay"
...
+  mode?: NotificationListMode
 }
 
 const UNREAD_INDICATOR_SIZE = 8
 
-const NotificationItem: FC<NotificationItemProps> = ({ item }) => {
+const NotificationItem: FC<NotificationItemProps> = ({ item, mode }) => {
   const { tracking } = useNotificationsTracking()
   const { relayEnvironment } = useSystemContext()
+  const isMobile = __internal__useMatchMedia(THEME.mediaQueries.xs)
```
olerichter00, line 35: Do you think there is a chance to use `Media` instead of `__internal__useMatchMedia`? If it gets to complicated, it should be ok to use `__internal__useMatchMedia`....

olerichter00 (dismissed):
Looks good! 🌟

src/Components/Notifications/NotificationItem.tsx
nickskalkin: @olerichter00 could you please check this commit? https://github.com/artsy/force/pull/14069/commits/66b369d84d6dbd38a1f6b22cdacbb1c638424787

src/Components/Notifications/NotificationItem.tsx
```diff
@@ -198,6 +156,87 @@ export const NotificationItemFragmentContainer = createFragmentContainer(
   }
 )
 
+interface NotificationItemLinkProps {
```
olerichter00, line 159:
```suggestion
interface NotificationItemWrapperProps {
```

src/Components/Notifications/NotificationItem.tsx
olerichter00: Looks good 👍

src/Components/Notifications/NotificationItem.tsx
nickskalkin: I tried to find a better name for this option, but I actually find mode (dropdown | page) more descriptive, because behaviour actually changes based on how the list of notifications is displayed.

olerichter00 (approved):

dariakoko (approved):

PR #14073, opened 2024-06-25
fix: Correctly link to Sell flow intro page

olerichter00:
## Description

Correctly link to the Sell Flow **intro** page when clicking "Start Selling" on the `/sell` page.


tam-kis (approved):
Yep, `sell2/intro` is the first step in the new flow. Thanks!

PR #14076, opened 2024-06-25
fix: New Sell flow design adjustments

olerichter00:
## Description

This PR includes several Sell flow design adjustments and small improvements to align with [Figma](https://www.figma.com/design/E6rVm1tiSI4PW3EzibKKZw/Submission-%26-Pre-Consignment-Management?node-id=2036-30168&t=NtEpZMeC7CU2fnML-0).

### DIFF
--- a/src/Apps/Sell/Components/BottomFormNavigation.tsx
+++ b/src/Apps/Sell/Components/BottomFormNavigation.tsx
@@ -1,5 +1,5 @@
 import { ContextModule, Intent } from "@artsy/cohesion"
-import { Box, Button, Flex, Text } from "@artsy/palette"
+import { Box, Button, Flex } from "@artsy/palette"
 import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
 import { useAuthDialog } from "Components/AuthDialog"
 import { useSystemContext } from "System/Hooks/useSystemContext"
@@ -51,7 +51,7 @@ const BottomFormBackButton = () => {
 
   return (
     <Button loading={isSubmitting} onClick={onBack} variant="tertiary">
-      <Text style={{ textDecoration: "underline" }}>Back</Text>
+      Back
     </Button>
   )
 }
--- a/src/Apps/Sell/Components/BottomFormNavigation.tsx
+++ b/src/Apps/Sell/Components/BottomFormNavigation.tsx
@@ -1,5 +1,5 @@
 import { ContextModule, Intent } from "@artsy/cohesion"
-import { Box, Button, Flex } from "@artsy/palette"
+import { Box, Button, Flex, Text } from "@artsy/palette"
 import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
 import { useAuthDialog } from "Components/AuthDialog"
 import { useSystemContext } from "System/Hooks/useSystemContext"
@@ -51,7 +51,7 @@ const BottomFormBackButton = () => {
 
   return (
     <Button loading={isSubmitting} onClick={onBack} variant="tertiary">
-      Back
+      <Text style={{ textDecoration: "underline" }}>Back</Text>
     </Button>
   )
 }
@@ -105,7 +105,7 @@ const BottomFormNextButton = () => {
       loading={isSubmitting}
       onClick={onNext}
     >
-      {isSubmitStep ? "Submit" : "Continue"}
+      {isSubmitStep ? "Submit Artwork" : "Continue"}
     </Button>
   )
 }
--- a/src/Apps/Sell/Components/SubmissionHeader.tsx
+++ b/src/Apps/Sell/Components/SubmissionHeader.tsx
@@ -12,7 +12,6 @@ export const SubmissionHeader: React.FC = () => {
   const context = useSellFlowContext()
   const isLastStep = context?.state?.isLastStep
   const submissionID = context?.state?.submissionID
-  const isSubmissionCompleted = isLastStep && submissionID
 
   return (
     <Sticky withoutHeaderOffset>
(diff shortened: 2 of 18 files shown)


src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -29,13 +28,14 @@ export const SubmissionHeader: React.FC = () => {
                   ]}
                   alignItems="center"
                   py={[1, 4]}
+                  mt={[0.5, 0]}
```
olerichter00, line 31: This is to add some extra padding to the top when on mobile.

src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -29,13 +28,14 @@ export const SubmissionHeader: React.FC = () => {
...
                 >
                   <Media greaterThan="xs">
                     <RouterLink to={"/sell"} display="block">
                       <ArtsyLogoIcon display="block" />
                     </RouterLink>
                   </Media>
-                  {submissionID ? (
+                  {submissionID && !isLastStep ? (
```
olerichter00, line 38: We don't want to show "Save & Exit" on the "/new" step and the last step.

src/Apps/Sell/Routes/SubmissionRoute.tsx
```diff
@@ -25,7 +26,10 @@ export const SubmissionRoute: React.FC<SubmissionRouteProps> = props => {
   return (
     <FullBleed>
       <AppContainer>
-        <SellFlowContextProvider submissionID={submission.externalId}>
+        <SellFlowContextProvider
+          submissionID={submission.externalId}
+          submissionState={submission.state}
```
olerichter00, line 31: We are saving the submission state in the context to prevent updating an already submitted submission because this would result in an error.

src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -75,9 +82,16 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
   } = useCreateSubmission()
   const { sendToast } = useToasts()
 
-  const stepFromURL = match.location.pathname.split("/").pop()
+  const isNewSubmission = !submissionID
+
+  const stepFromURL = isNewSubmission
+    ? INITIAL_STEP
+    : match.location.pathname.split("/").pop()
```
olerichter00, line 89: Correctly setting up the initial step when we are on the "/new" route.

src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -51,7 +51,7 @@ const BottomFormBackButton = () => {
 
   return (
     <Button loading={isSubmitting} onClick={onBack} variant="tertiary">
-      Back
+      <Text style={{ textDecoration: "underline" }}>Back</Text>
```
olerichter00, line 54: I'll ask if we really need the underline.

src/Apps/Sell/Routes/SubmissionRoute.tsx
dblandin: Instead of persisting the initial state here, can we better respond to the mutation error? What kind of message do we get back from the backend? Would that be revealing enough to expose an error via the same toast mechanism?

That said, we should have a broader discussion on possible submission states and which states allow updates to take place. For example, we're planning to allow updates after a submission is `approved` but are we considering it "locked" between `submitted` and `approved`? Might be useful to discuss and align on expectations.

dblandin (approved):
Looking good! Only element I might defer to another PR is the handling of submissions in a `submitted` state. I think that might need a bit more discussion.

src/Apps/Sell/Routes/SubmissionRoute.tsx
olerichter00: When the submission is in the state "SUBMITTED", updating the submission returns the error "Submission Not Found". So it's possible to load the submission but not to update it. 

What do you think about merging this as an interim solution until we have decided whether we allow updating submitted/approved submissions? This isn't ideal, but at least it returns a reasonable error message for now. Just in case a user somehow lands directly on a step for an already submitted submission.

Alternatively, we could render an error message in this case and not even allow viewing the step 🤔

dblandin: > Submission Not Found

Hmm! Sounds like some system (probably Convection) is trying to find the submission in a draft state right away and throws a 404 if not found.

I would expect that if the submission ID is valid but just in the wrong state for the attempted action (updating), that we could return a `422: Unprocessable Entity` error instead along with a more revealing error message. So we could make that improvement to make the client implementation a bit simpler.

I suppose if we don't support any editing on the submission in certain states — is there another surface we could redirect to? Maybe should the error message there? Thinking of the My Collection artwork page or even the My Collection landing surface...

olerichter00: You're right, the `updateConsignmentSubmission` resolver in Convection returns the incorrect error message ([GitHub](https://github.com/artsy/convection/blob/45f7d5984bc26ad90a9ed8ef92bb28f60facccfd/app/graphql/resolvers/update_submission_resolver.rb#L9)). Updating the error message in Convection would allow us to simplify this PR so that we don't have to check the submission's state in the client.

I'd say we go this route: correct the error message in Convection and display it if updating the submission fails.

olerichter00: https://github.com/artsy/convection/pull/1483

PR #14078, opened 2024-06-25
feat: add rudimentary 07 artworks rail

anandaroop:
Brings over a simple Artworks rail from 06 into 07.

Deliberately pared down and simplified to match the existing Collections rails so that we can incrementally, in future PRs:
- add the price filtering (once we've re-ingested [with better data](https://artsy.slack.com/archives/C06SSV1K10D/p1719323396490219))
- add liking/disliking back in
- do a design pass over all the rails

For now, pretty straighforward:

https://github.com/artsy/force/assets/140521/5976765e-03e4-4af3-8d01-6ac173dbcebc


mc-jones (approved):
Looks good. Think it just needs a rebase.

PR #14059, opened 2024-06-24
chore: New sell flow tracking

olerichter00:
Addresses [ONYX-1022]

## Description

This PR adds tracking for the new Sell flow.

**Events:**

```
action: "tappedNewSubmission",
context_module: "sell",
context_owner_type: "submitArtworkStepStart",

action: "tappedSubmissionBack",
context_module: "sell",
context_owner_type: null,
submission_id: "123",
submission_step: "title",

action: "tappedSubmissionSaveExit",
context_module: "sell",
context_owner_type: null,
submission_id: "123",
submission_step: "title",

action: "consignmentSubmitted",
context_module: "sell",
context_owner_type: "submitArtworkStepCompleteYourSubmission",
fieldsProvided: [],
submission_id: '<mock-value-for-field-"externalId">',

action: "tappedSubmitAnotherWork",
context_module: "sell",
context_owner_type: "submitArtworkStepCompleteYourSubmission",
submission_id: '<mock-value-for-field-"internalID">',

action: "tappedViewArtworkInMyCollection",
context_module: "sell",
context_owner_type: "submitArtworkStepCompleteYourSubmission",
submission_id: '<mock-value-for-field-"internalID">',
```

[ONYX-1022]: https://artsyproduct.atlassian.net/browse/ONYX-1022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14088, opened 2024-06-26
feat(ONYX-1020): add phone number step

dariakoko:

This PR solves [ONYX-1020]

### Description

Add phone number step
Use user's phone number for a new submission
Use saved submission phone number for the existing submission

https://github.com/artsy/force/assets/36167539/e1f7e24c-8c95-4f79-a234-9fe65d7a53ff

image

[ONYX-1020]: https://artsyproduct.atlassian.net/browse/ONYX-1020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Sell/Routes/PhoneNumberRoute.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const onSubmit = async (values: FormValues) => {
+    const phoneNumberInternational = `+${
+      COUNTRY_CODES[values.phoneNumberCountryCode.toLocaleUpperCase()]
+    } ${values.userPhone.trim()}`
+
+    const updatedValues = {
+      userPhone: phoneNumberInternational,
+    }
```
MounirDhahri, line 71: there is also a potential to improve this by moving the country code and the phone number to the mutation. Let's do it later though as a small refactor in app and web

src/Apps/Sell/Routes/PhoneNumberRoute.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const me = useFragment(FRAGMENT_ME, props.me)
+
+  const { actions } = useSellFlowContext()
+
+  const initialValues: FormValues = {
+    userPhone:
+      submission.userPhoneNumber?.display ?? me.phoneNumber?.display ?? "",
+    phoneNumberCountryCode:
+      submission.userPhoneNumber?.regionCode ??
+      me.phoneNumber?.regionCode ??
```
MounirDhahri, line 60: you are using the `regionCode` here instead of `countryCode` - if we start using the `countryCode`. I would be more in favor of using the country code

olerichter00 (dismissed):
Looks great! 🌟 

You might need to rebase and update the tests. I've updated the [tests for the "final" step](https://github.com/artsy/force/pull/14059/files#diff-fd8f9850f0483384c5f05491ad4ee1948c849f3b88570a8aa3952b1315072c61R104) when I added tracking in https://github.com/artsy/force/pull/14059

src/Apps/Sell/Routes/PhoneNumberRoute.tsx
```diff
@@ -0,0 +1,111 @@
...
+`
+
+const FRAGMENT_ME = graphql`
+  fragment PhoneNumberRoute_me on Me {
+    internalID
+    phoneNumber {
+      countryCode
+      regionCode
+      display(format: NATIONAL)
+      originalNumber
```
MounirDhahri, line 30: I don't see you using the `originalNumber` and the `countryCode`.

MounirDhahri (commented):
Nice one 👏

src/Apps/Sell/Routes/PhoneNumberRoute.tsx
dariakoko: we can't use the `countryCode` because the component accepts as initial value only `regionCode`

src/Apps/Sell/Routes/__tests__/PhoneNumberRoute.jest.tsx
```diff
@@ -0,0 +1,104 @@
...
+})
+
+describe("PhoneNumberRoute", () => {
+  it("renders the artist title step", () => {
+    renderWithRelay({})
+
+    expect(screen.getByText("Add phone number")).toBeInTheDocument()
+  })
+
+  it("saves the submission & navigates to the confirmation step when Submit button is clicked", async () => {
```
olerichter00, line 80: I've updated the tests for the last step (tracking + update mutation + update mutation to set state). You can probably just copy [these](https://github.com/artsy/force/blob/bd7f2dc884999407de8c00313ba84b726033c243/src/Apps/Sell/Routes/__tests__/DimensionsRoute.jest.tsx#L104) tests and adapt them for the "new" last step.

olerichter00: And the `dimensions" step tests would need an update too because it's now not the final step anymore.

src/Apps/Sell/Routes/PhoneNumberRoute.tsx
MounirDhahri: oki - let's maybe rename `phoneNumberCountryCode` to `phoneNumberRegionCode` then

MounirDhahri (dismissed):
Thanks for addressing my comments - I left a follow-up note but overall looks good to me

olerichter00 (approved):
🥳

src/Apps/Sell/Routes/__tests__/PhoneNumberRoute.jest.tsx
dariakoko: Added the changes 👌 
Some of the tracking still has to be implemented and tested, right?

PR #14094, opened 2024-06-27
chore(EMI-1940): update list of european countries that is used for EU shipping

oxaudo:
This simply updates the list of the countries that are considered EU shipping to be consistent with what we are promising partners.


erikdstock:
non-blocking: would it make sense to use [artsy/commerce_helpers](https://github.com/artsy/commerce_helpers?tab=readme-ov-file#as-a-node-package) for this? It looks like [the data is not all there](https://github.com/artsy/commerce_helpers/blob/main/package/data/data.json) (country name mappings for example) and personally I would prefer to add types to our exports, even if it is just json - I believe this is possible without too much work. Maybe a future (friday) improvement.

erikdstock (approved):
LGTM aside from my non-blocking comment

PR #13992, opened 2024-06-13
feat: add 06 neartext experiment

anandaroop:
Adds an experiment to source artwork discovery recommendations for a user based on:

- a list of plaintext "concepts" that describes the user
- a list of liked artworks
- a list of disliked artworks

The concepts act as a semantic search over the Weaviate `DiscoveryArtworks` collection (a few hundred artworks from our curated collections, as of this writing)

The dis/likes serve to refine the results based on explicit user signals.

https://github.com/artsy/force/assets/140521/e6aa165f-ee64-4ebc-92ed-4e6dcf390438

PR #14097, opened 2024-06-27
feat: add artwork like and dislike functionality

mc-jones:
This PR updates the `07-curated-discovery` app with functionality to allow a user to like or dislike artworks so that we may refine what is shown to them in the `ArtworksRail`. It largely ports over the implementation from #13992, with some exceptions, which are noted in the comments.

Follow-ups:

- UI shifting on refetch
- Handling different artwork sizes

https://github.com/artsy/force/assets/38149304/a3eba88b-4107-4369-b7bc-5dd9c7581e5e

### DIFF
--- a/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArtworksRail.tsx
+++ b/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArtworksRail.tsx
@@ -1,8 +1,9 @@
 import React, { FC, useEffect, useState } from "react"
-import { Box, SkeletonBox, Text } from "@artsy/palette"
+import { Box, SkeletonBox } from "@artsy/palette"
 import { Rail } from "Components/Rail/Rail"
 import { BudgetIntent, State } from "Apps/ArtAdvisor/07-Curated-Discovery/App"
 import { Artwork } from "Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Artwork"
+import { useSystemContext } from "System/Hooks/useSystemContext"
 
 interface ArtworksRailProps {
   state: State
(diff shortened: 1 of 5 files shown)


src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArtworksRail.tsx
```diff
@@ -39,48 +41,84 @@ export const ArtworksRail: FC<ArtworksRailProps> = props => {
   const { state } = props
   const [artworks, setArtworks] = useState<DiscoveryArtwork[]>([])
   const [isLoading, setIsLoading] = useState<boolean>(true)
+  const [excludeArtworkIds, setExcludeArtworkIds] = useState<string[]>([])
```
mc-jones, line 44: I decided to store these client side and send with the `fetchArtworks` request (L49) for a few reasons:

1. This prevents us from having needing the weaviate user client side, as we aren't reliant on the `likedArtworks` or `dislikedArtworks` fields. 
2. This allows us to filter out `artworks` the user has liked or disliked from the rail (note: a refresh would clear the filter list).
3. We need a state change in the `ArtworksRails` component  to trigger a refetch of the artworks.

src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/ArtworksRail.tsx
```diff
@@ -39,48 +41,84 @@ export const ArtworksRail: FC<ArtworksRailProps> = props => {
...
+                <ShelfArtworkPlaceholder
+                  key={i}
+                  index={i}
+                  hideSaleInfo={true}
+                  maxImageHeight={400}
+                />
+              )
+            })}
+          </Shelf>
+        </Skeleton>
```
mc-jones, line 104: Added this, which is the canonical rail loading screen. There is still a bit of ui shifting that needs to be addressed. But seemed reasonable to handle that better at the ui polish stage, if we keep this.

anandaroop (approved):
👍🏽

PR #14103, opened 2024-06-28
chore(AMBER-769): Do not hide the sale message on the grid

lidimayra:

This PR solves [AMBER-769]

### Description

We initially defined the sale message shouldn't be displayed in case of "Inquire about availability" (see https://github.com/artsy/force/pull/14002). This caused weirdness on the grid compared to other artworks. We're undoing this.

cc @artsy/amber-devs 

[AMBER-769]: https://artsyproduct.atlassian.net/browse/AMBER-769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Artwork/Details.tsx
+++ b/src/Components/Artwork/Details.tsx
@@ -163,13 +163,6 @@ const SaleMessage: React.FC<DetailsProps> = ({
     return <>{highestBid_display || openingBid_display || ""}</>
   }
 
-  if (sale_message?.toLowerCase() === "contact for price") {
-    return <>Price on request</>
-  }
-
-  if (sale_message?.toLowerCase() === "inquire about availability") {
-    return <>{NBSP}</>
-  }
   // NBSP is used to prevent un-aligned carousels
   return <>{sale_message ?? NBSP}</>
 }
--- a/src/Components/Artwork/__tests__/Details.jest.tsx
+++ b/src/Components/Artwork/__tests__/Details.jest.tsx
@@ -129,10 +129,10 @@ describe("Details", () => {
       expect(html).toContain("$2,400")
     })
 
-    it("shows Contact for price if sale_message equals the same", async () => {
+    it("shows sale_message", async () => {
       const data: any = {
         ...artworkInAuction,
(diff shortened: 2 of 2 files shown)


src/Components/Artwork/Details.tsx
```diff
@@ -163,13 +163,6 @@ const SaleMessage: React.FC<DetailsProps> = ({
     return <>{highestBid_display || openingBid_display || ""}</>
   }
 
-  if (sale_message?.toLowerCase() === "contact for price") {
-    return <>Price on request</>
-  }
```
lidimayra, line 168: The removal of this condition relates to the work done in:
- https://github.com/artsy/gravity/pull/17851
- https://github.com/artsy/metaphysics/pull/5816

We're handling this logic now in Gravity, so if `Price on request` is needed here, it will come like this from the query and the client doesn't have to worry about it.

src/Components/Artwork/Details.tsx
```diff
@@ -163,13 +163,6 @@ const SaleMessage: React.FC<DetailsProps> = ({
...
-
-  if (sale_message?.toLowerCase() === "inquire about availability") {
-    return <>{NBSP}</>
-  }
```
lidimayra, line 172: Removal of this condition will enable the sale message to be displayed here again

Screenshot 2024-06-28 at 13 26 26

lidimayra:
In theory, the work done in Gravity and MP should also enable us to remove the `Bidding closed` logic defined here:
https://github.com/artsy/force/blob/2e055b8d43fef65fa3464ee1acb465a094f10c1c/src/Components/Artwork/Details.tsx#L155-L157

Reasoning to keep it for now is that some inconsistency was noticed between Gravity and Force regarding what are the conditions to consider it as bidding closed.

I moved this discussion to the [dev-help channel](https://artsy.slack.com/archives/CP9P4KR35/p1719576095293039).

Mitch-Henson (approved):

PR #14114, opened 2024-06-28
feat: minor styling changes to 07 

anandaroop:
The type of this PR is: **Style**

A few opinionated style tweaks to get this looking closer to our design patterns.

- A fuller artwork tombstone, styled conventionally
- Like/dislike buttons positioned conventionally
- Conventionally styled RouterLinks throughout
- Skeletons adjusted to minimize layout shifts

tweaked

### DIFF
--- a/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Article.tsx
+++ b/src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Article.tsx
@@ -1,18 +1,27 @@
-import { Box, Image } from "@artsy/palette"
+import { Box, Image, Text } from "@artsy/palette"
 import { DiscoveryArticle } from "Apps/ArtAdvisor/07-Curated-Discovery/types"
+import { RouterLink } from "System/Components/RouterLink"
+import { resize } from "Utils/resizer"
 import { FC } from "react"
+
+const IMAGE_HEIGHT = 200
 
 interface ArticleRailsProps {
   article: DiscoveryArticle
 }
 
 export const Article: FC<ArticleRailsProps> = ({ article }) => {
+  const resizedImage = resize(
+    article.imageUrl || "https://www.artsy.net/images/missing_image.png",
+    { height: IMAGE_HEIGHT }
+  )
+
   return (
-    <Box>
-      <Image src={article.imageUrl} height={300} />
-      <a href={`${article.href}`} target="_blank">
-        {article.title}
-      </a>
-    </Box>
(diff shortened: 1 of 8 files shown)


src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Article.tsx
```diff
@@ -1,18 +1,27 @@
...
 
+const IMAGE_HEIGHT = 200
+
 interface ArticleRailsProps {
   article: DiscoveryArticle
 }
 
 export const Article: FC<ArticleRailsProps> = ({ article }) => {
+  const resizedImage = resize(
+    article.imageUrl || "https://www.artsy.net/images/missing_image.png",
```
mc-jones, line 15: Nice. Wasn't aware of the missing image link!

src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Article.tsx
```diff
@@ -1,18 +1,27 @@
...
+  )
+
   return (
-    <Box>
-      <Image src={article.imageUrl} height={300} />
-      <a href={`${article.href}`} target="_blank">
-        {article.title}
-      </a>
-    </Box>
+    <RouterLink to={article.href} textDecoration="none" overflow="hidden">
```
mc-jones, line 20: Maybe doesn't matter for a prototype, but this will reroute the current window instead of opening a new window, correct? This felt sub-optimal and was what led me to falling back to the vanilla anchor tag as I didn't see an obvious way to replicate that behavior with out router link.

mc-jones:
Awesome improvements 🙏

src/Apps/ArtAdvisor/07-Curated-Discovery/Components/Result/Artwork.tsx
```diff
@@ -74,22 +78,67 @@ export const Artwork: FC<ArtworkProps> = props => {
...
+        justifyContent={"space-between"}
+        gap={1}
+        pt={0.5}
+      >
+        <RouterLink
+          to={`/artwork/${artwork.slug}`}
+          target="_blank"
+          textDecoration={"none"}
+          overflow={"hidden"}
+        >
```
anandaroop, line 108: @mc-jones Good point about the new window, esp since we'd lose state by navigating away. I forgot that I did actually sort that out for the artwork cards, with the `target` attribute here on L105. I'll copy that over to the other two cards as well.

PR #14123, opened 2024-07-01
fix(ONYX-1037): change isMobile efinition

dariakoko:

This PR solves [ONYX-1037]

### Description

Show the first alert's details only when the screen size is greater than sm

https://github.com/artsy/force/assets/36167539/3b27dcb2-557c-4390-a5f9-d19e48ae4183

[ONYX-1037]: https://artsyproduct.atlassian.net/browse/ONYX-1037?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
@@ -9,6 +9,7 @@ import {
   Flex,
   FullBleed,
   Stack,
+  THEME,
 } from "@artsy/palette"
 import {
   createPaginationContainer,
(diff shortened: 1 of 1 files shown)


src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
```diff
@@ -62,7 +62,10 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
   const [sort, setSort] = useState("ENABLED_AT_DESC")
   const [loading, setLoading] = useState(false)
   const alerts = extractNodes(me.alertsConnection)
-  const isMobile = getENV("IS_MOBILE")
+
+  const xs = __internal__useMatchMedia(THEME.mediaQueries.xs)
```
nickskalkin, line 66: I know that it's not recommended to use it because of server-side rendering, but I do not see any implications of using it here. Maybe @dzucconi or @damassi can say something. In my experience using IS_MOBILE is more harmful

dzucconi: It's not server-side friendly (can create very unpredictable and hard to debug problems). You're using it in an effect, so that's fine since it only runs on the client.

PR #14124, opened 2024-07-01
fix: New Sell flow design adjustments and fixes

olerichter00:
## Description

New Sell flow design adjustments and fixes.

### DIFF
--- a/src/Apps/Sell/Components/SubmissionLayout.tsx
+++ b/src/Apps/Sell/Components/SubmissionLayout.tsx
@@ -4,6 +4,7 @@ import { StepsNavigation } from "Apps/Sell/Components/StepsNavigation"
 import { SubmissionHeader } from "Apps/Sell/Components/SubmissionHeader"
 import { SubmissionProgressBar } from "Apps/Sell/Components/SubmissionProgressBar"
 import { useSellFlowContext } from "Apps/Sell/SellFlowContext"
+import { FadeInBox } from "Components/FadeInBox"
 import { useEffect } from "react"
 
 const CONTENT_WIDTH = 640
@@ -44,9 +45,9 @@ export const SubmissionLayout: React.FC<SubmissionLayoutProps> = ({
             </Column>
           </GridColumns>
         ) : (
-          <Box maxWidth="100vw" width={CONTENT_WIDTH} p={2} pt={[2, 4]}>
+          <FadeInBox maxWidth="100vw" width={CONTENT_WIDTH} p={2} pt={[2, 4]}>
             {children}
-          </Box>
+          </FadeInBox>
         )}
       </Flex>
 
--- a/src/Apps/Sell/Routes/ArtistRoute.tsx
+++ b/src/Apps/Sell/Routes/ArtistRoute.tsx
@@ -190,16 +190,18 @@ export const ArtistRoute: React.FC<{
             title="Artist"
           />
 
-          <Text mt={2} variant="sm" color="black60">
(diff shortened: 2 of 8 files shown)


src/Apps/Sell/Components/SubmissionLayout.tsx
```diff
@@ -44,9 +45,9 @@ export const SubmissionLayout: React.FC<SubmissionLayoutProps> = ({
             </Column>
           </GridColumns>
         ) : (
-          <Box maxWidth="100vw" width={CONTENT_WIDTH} p={2} pt={[2, 4]}>
+          <FadeInBox maxWidth="100vw" width={CONTENT_WIDTH} p={2} pt={[2, 4]}>
```
olerichter00, line 48: This adds a fade transition between steps.

src/Apps/Sell/Routes/ArtistRoute.tsx
```diff
@@ -190,16 +190,18 @@ export const ArtistRoute: React.FC<{
...
-            Currently, artists can not sell their own work on Artsy.{" "}
-            <RouterLink
-              to="https://support.artsy.net/s/article/Im-an-artist-Can-I-submit-my-own-work-to-sell"
-              target="_blank"
-              rel="noopener noreferrer"
-            >
-              Learn more.
-            </RouterLink>
-          </Text>
+          {!formik.values.artistName && (
```
olerichter00, line 193: We don't want to display the text anymore when the user types.

src/Apps/Sell/Routes/DimensionsRoute.tsx
```diff
@@ -95,7 +95,8 @@ export const DimensionsRoute: React.FC<DimensionsRouteProps> = props => {
 
             <GridColumns>
               <Column span={[6, 6]}>
-                <Input
+                <LabeledInput
+                  label={values.dimensionsMetric}
```
olerichter00, line 99: Adding labels ("in" or "cm") to the input.

src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -185,7 +185,8 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
   }
 
   const state = {
-    isFirstStep: index === 0,
+    // Both the first and second steps are considered the first step because we don't want to allow going back to the artist step.
+    isFirstStep: index === 0 || index === 1,
```
olerichter00, line 189: We don't want to display the "Back" button on the Title step to avoid having a button to go back to change the artist.

I'm not so sure about the new definition of `isFirstStep`, but I thought it makes sense because depending on whether you're creating or editing a submission, the first step is either the "Artist" step or the "Title" step 🤔

nickskalkin: Just curious - why are we not allowing this? I can imaging user clicking on a wrong artist by mistake and going back to revert it

nickskalkin (approved):

dariakoko (approved):

src/Apps/Sell/SellFlowContext.tsx
olerichter00: Let's phrase it like this: we still allow it (by using the browser's back button) but decided not to provide the "Back" button in the form on the step after the "Artist" step.

olerichter00: Like this, we allow changing the artist and the back button keeps its normal behavior. But we don't provide a clear path to updating the artist of an already existing submission.

PR #14126, opened 2024-07-01
feat(onyx-1098): use MyC artwork images when creating new submission

nickskalkin:

This PR solves [ONYX-1098]

### Description

MyC artwork photos are automatically synced with submission when creating it from MyC artwork page:

https://github.com/artsy/force/assets/3934579/a8cb8e8d-23d7-414c-9d25-502cd2b2842b


olerichter00 (approved):
Thanks a lot 🙏 

The only issue I've noticed is that the images from the My Collection artwork get re-added when I remove all images and then reload the page. But I guess this is an edge case, and fixing it could be quite complicated.

PR #14127, opened 2024-07-01
fix: link behavior in 07

anandaroop:

Makes the article and collection cards behave the same as artwork cards — opens the link in a new window. This is non-standard, but acceptable for our prototype so that we mitigate the awkwardness of navigating away from the results page and losing the page's state.


mc-jones (approved):
🙏

PR #14129, opened 2024-07-01
feat: Add phone number validation to Sell flow

olerichter00:
Resolves https://www.notion.so/artsy/Phone-number-can-take-letters-without-any-invalidation-message-e358107930ac4df680b78869210b86d3?pvs=4

## Description

This adds phone number validation to the Sell flow. The validation only results in a warning, and it will still be possible to submit with an invalid number.

Screenshot 2024-07-03 at 11 53 24

### DIFF
--- a/src/Apps/Sell/SellFlowContext.tsx
+++ b/src/Apps/Sell/SellFlowContext.tsx
@@ -143,7 +143,7 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
       logger.error("Error creating submission.", err)
       sendToast({
         variant: "error",
-        message: err?.[0]?.message || "Something went wrong.",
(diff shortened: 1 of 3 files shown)


src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -143,7 +143,7 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
       logger.error("Error creating submission.", err)
       sendToast({
         variant: "error",
-        message: err?.[0]?.message || "Something went wrong.",
+        message: "Something went wrong." + ` ${err?.[0]?.message} || ""`,
```
olerichter00, line 146: Displaying the text and an error message in case something goes wrong to not only display something like "Unexpected item at ...."

MounirDhahri (approved):

PR #14136, opened 2024-07-02
fix: Point all Start Selling buttons to new Sell flow

olerichter00:
## Description

Point all Start Selling buttons to new Sell flow.


MounirDhahri (approved):
Nice - thanks

PR #14146, opened 2024-07-02
fix(EMI-1938): use correct local state when assigning default selected value on shipping quotes

oxaudo:
This gets the selection to parity to what is currently on staging. 

However it's still a bit confusing - currently this selection is kept UI only on client side until 'save and continue' is clicked + when the page reloads - we are requesting a new set of arta quotes - which makes curretnly selected quote invalid any more which in turn selects the first quote.  So if collector makes a shipping selection/clicks save/continue and then decides to click back button in a browser the selection gets out of sync with what is stored on the artwork. The side panel calculation is also not updated until 'save/continue' is clicked - so it's a bit of a confusing experience. But that is a different issue 🤷‍♀️


src/Apps/Order/Routes/Shipping2/index.tsx
```diff
@@ -72,9 +71,6 @@ const ShippingRouteLayout: FC<Omit<ShippingProps, "dialog">> = ({
 
   const isOffer = order.mode === "OFFER"
 
-  // Automatically selects first shipping quote when they change
-  useSelectFirstShippingQuote()
```
oxaudo, line 76: Removing this as the logic for selection the first available quote is now available in `ShippingQuotes2` component.

src/Apps/Order/Routes/Shipping2/Components/ShippingQuotes2.tsx
```diff
@@ -59,7 +59,7 @@ export const ShippingQuotes2: React.FC<ShippingQuotesProps> = ({
   return (
     <RadioGroup
       onSelect={handleShippingQuoteSelected}
-      defaultValue={shippingContext.orderData.selectedShippingQuoteID}
+      defaultValue={shippingContext.state.selectedShippingQuoteID}
```
oxaudo, line 62: Thanks @erikdstock for enabling this one line change here!

PR #14150, opened 2024-07-03
fix: Finish previous submission adjustments

olerichter00:
## Description

Finish previous submission adjustments.

### DIFF
--- a/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/PreviousSubmission.tsx
+++ b/src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/PreviousSubmission.tsx
@@ -48,7 +48,7 @@ const PreviousSubmission: React.FC<PreviousSubmissionProps> = ({
     router.push(`/sell2/submissions/${submissionID}/${currentStep}`)
   }
 
-  if (!submission) return null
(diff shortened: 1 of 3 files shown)


src/Apps/Consign/Routes/MarketingLanding/Components/LandingPage/PreviousSubmission.tsx
```diff
@@ -48,7 +48,7 @@ const PreviousSubmission: React.FC<PreviousSubmissionProps> = ({
     router.push(`/sell2/submissions/${submissionID}/${currentStep}`)
   }
 
-  if (!submission) return null
+  if (!submission || submission.state !== "DRAFT") return null
```
olerichter00, line 51: We don't want to display the previous submission if it's not in state DRAFT.

src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -14,6 +14,7 @@ import createLogger from "Utils/logger"
 import { useFormikContext } from "formik"
 import { useState } from "react"
 
+const HEADER_HEIGHT = 40
```
olerichter00, line 17: I'm setting the height, because now we use a mix of buttons and links which have different heights.

PR #14152, opened 2024-07-03
fix: Guard Sell flow step navigation

olerichter00:
## Description

Guard Sell flow step navigation.

### DIFF
--- a/src/Apps/Sell/SellFlowContext.tsx
+++ b/src/Apps/Sell/SellFlowContext.tsx
@@ -127,9 +127,11 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
   }
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -127,9 +127,11 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
   }
 
   useEffect(() => {
-    if (isNewSubmission) return
+    const newStep = STEPS[index]
 
-    push(`/sell2/submissions/${submissionID}/${STEPS[index]}`)
+    if (isNewSubmission || !newStep) return
```
olerichter00, line 132: Just to make sure we never end up in a step that does not exist.

dariakoko (approved):

PR #14108, opened 2024-06-28
feat: Finish previous submission experience

olerichter00:
Addresses [ONYX-1060]

## Description

This implements the "Finish previous submission" experience.

| | |
| --- | --- |
|Screenshot 2024-07-03 at 10 26 43 | Screenshot 2024-07-03 at 10 27 00|

[ONYX-1060]: https://artsyproduct.atlassian.net/browse/ONYX-1060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14153, opened 2024-07-03
fix: Add Yup validation to Previous submission localStorage data

olerichter00:

This PR solves https://github.com/artsy/force/pull/14108#discussion_r1663996003

### Description

This PR adds Yup validation to validate "previous submission" localStorage data.

### DIFF
--- a/src/Apps/Sell/Utils/previousSubmissionUtils.ts
+++ b/src/Apps/Sell/Utils/previousSubmissionUtils.ts
@@ -1,11 +1,13 @@
-import { SellFlowStep } from "Apps/Sell/SellFlowContext"
+import { STEPS, SellFlowStep } from "Apps/Sell/SellFlowContext"
 import createLogger from "Utils/logger"
 import { useEffect, useState } from "react"
+import * as Yup from "yup"
 
 const SUBMISSION_ID_KEY = "previousSubmissionID"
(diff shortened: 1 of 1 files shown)


src/Apps/Sell/Utils/previousSubmissionUtils.ts
```diff
@@ -18,20 +20,34 @@ export const usePreviousSubmission = () => {
   const [submissionID, setSubmissionID] = useState<string | null>(null)
   const [step, setStep] = useState<SellFlowStep | null>(null)
 
+  const stepSchema = Yup.string().oneOf([...STEPS])
+  const idSchema = Yup.string().required()
+
+  const isStepValid = (step: any): step is SellFlowStep | null =>
+    stepSchema.isValidSync(step)
+
+  const isIdValid = (id: any): id is string => idSchema.isValidSync(id)
```
dzucconi, line 29: You can move these out of the function body. Which should also avoid your dependency array ignore.

dzucconi: And you can use validateSync instead of isValidSync which should just return the correct type

dzucconi: (Rather than using it as a typeguard)

dzucconi (approved):

PR #14133, opened 2024-07-02
fix(ONYX-1097): adjust component structure

dariakoko:
The type of this PR is: **Fix**

This PR solves [ONYX-1097]

### Description

Adjust the Notification component structure to support displaying the component from the top and not from the previous scroll position

Before: https://artsy.slack.com/archives/C03N12SR0RK/p1719489084109179

After:

https://github.com/artsy/force/assets/36167539/a3a44651-a637-43e6-bd2c-793afa5256cd

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-1097]: https://artsyproduct.atlassian.net/browse/ONYX-1097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14156, opened 2024-07-03
feat: adds activity cards for complete profile prompts

dzucconi:
Closes [DIA-706](https://artsyproduct.atlassian.net/browse/DIA-706), [DIA-702](https://artsyproduct.atlassian.net/browse/DIA-702), [DIA-688](https://artsyproduct.atlassian.net/browse/DIA-688), [DIA-697](https://artsyproduct.atlassian.net/browse/DIA-697)

[DIA-706]: https://artsyproduct.atlassian.net/browse/DIA-706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-702]: https://artsyproduct.atlassian.net/browse/DIA-702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-688]: https://artsyproduct.atlassian.net/browse/DIA-688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DIA-697]: https://artsyproduct.atlassian.net/browse/DIA-697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dzucconi:
I need to make a ticket and take care of the dedicated page. Right now, personally, I think this is launchable without it (it just opens the dialog) — but...

src/Apps/Notifications/NotificationsApp.tsx
```diff
@@ -25,28 +26,33 @@ const NotificationsApp: React.FC<NotificationsAppProps> = ({ me }) => {
...
-  const isNotificationsPage = match.location.pathname.startsWith(
-    "/notifications"
-  )
+  const pathname = match.location.pathname
+  const isNotificationsPage = pathname.startsWith("/notifications")
+
+  // Scroll to top of content pane when route changes
+  useEffect(() => {
+    contentRef.current?.scrollTo(0, 0)
+  }, [pathname])
```
dzucconi, line 55: @dariakoko this is how you resolve https://github.com/artsy/force/pull/14133 with scrolling. The CSS solution in that previous PR created other problems I ran into. This is much simpler.

araujobarret: Q around the other PR, I don't think we should be fixing things from resizing to smaller breakpoints and then to larger ones, our UI is solid but that adds crazy more bug options to pop, what do you think @dzucconi  about that kind of fixes?

dzucconi: I just fix things as I run into them. The previous fix of this bug created problems with new notification layouts. There's an even better way to fix this for real and eliminate a ton of CSS but I'm not going to bother refactoring the whole thing.

araujobarret: I know and it makes sense, but my question was about your opinion of the other PR fix idea, if we should be hunting this kind of bugs on our UI

src/Components/Notifications/Notification.tsx
```diff
@@ -23,33 +26,28 @@ import createLogger from "Utils/logger"
...
 import { PartnerShowOpenedNotification } from "Components/Notifications/PartnerShowOpenedNotification"
-import { DESKTOP_HEIGHT } from "Apps/Notifications/notificationsutils"
+import { NotificationCollectorProfileUpdatePromptFragmentContainer } from "Components/Notifications/NotificationCollectorProfileUpdatePrompt"
 
 const logger = createLogger("NotificationItem")
 
+// FIXME: Should probably be a column span
 export const NOTIFICATION_MAX_WIDTH = 600
-export const SUPPORTED_NOTIFICATION_TYPES = [
+export const SUPPORTED_NOTIFICATION_TYPES: NotificationTypesEnum[] = [
```
araujobarret, line 35: Relay could refactor their compiler to generate literal types instead of enums 😞

src/Components/Notifications/Notification.tsx
```diff
@@ -66,44 +64,39 @@ const Notification: React.FC<NotificationProps> = ({ notificationId }) => {
...
-
-    try {
-      const response = await markNotificationAsRead(
-        relayEnvironment,
-        notification.id,
-        notification.internalID
-      )
-      const responseOrError = response.markNotificationAsRead?.responseOrError
-      const errorMessage = responseOrError?.mutationError?.message
+    const markAsRead = async () => {
```
araujobarret, line 76: shouldn't we use the hook you created here `useMarkNotificationAsRead`?

araujobarret (dismissed):
All good, just one minor comment 👍

src/Components/Notifications/Notification.tsx
dzucconi: Might as well

src/Components/Notifications/Notification.tsx
dzucconi: @mzikherman added the `item` type. There's a lot wrong with this. Like: why is there a redirect buried inside of a sub-sub-sub-component (took me like an hour to dig this out). And this isn't even a good way to handle this. 😭

src/Components/CompleteProfile/CompleteProfileCollectionDialog.tsx
```diff
@@ -8,10 +17,39 @@ interface CompleteProfileCollectionDialogProps {
...
+      action: ActionType.tappedMyCollectionAddArtworkArtist,
+      context_screen: OwnerType.myCollectionAddArtworkArtist, // FIXME: Should probably be ContextModule?
+      context_module: ContextModule.myCollectionAddArtworkAddArtist,
+      context_screen_owner_id: contextPageOwnerId,
+      context_screen_owner_slug: contextPageOwnerSlug,
+      platform: "web",
+    }
+
+    trackEvent(editedUserProfile)
+    trackEvent(tappedMyCollectionAddArtworkArtist)
```
damassi, line 45: Side non-blocking note: it would be great to collect all of these distributed tracking calls within their own hook. (not sure how many there are, but its always easier to manage if things are in one place.)

src/Components/Notifications/NotificationCollectorProfileUpdatePrompt.tsx
```diff
@@ -0,0 +1,120 @@
...
+      )}
+
+      {mode === "Open" && hasArtistsInCollection && (
+        <CompleteProfileInformationDialog onClose={handleClose} />
+      )}
+    </>
+  )
+}
+
+export const NotificationCollectorProfileUpdatePromptFragmentContainer = createFragmentContainer(
```
damassi, line 103: Can we use a `useFragment` hook here?

src/Components/Notifications/NotificationItemCollectorProfileUpdatePrompt.tsx
```diff
@@ -0,0 +1,144 @@
...
+      </Container>
+
+      {mode === "Open" && (
+        <CompleteProfileCollectionDialog onClose={handleClose} />
+      )}
+    </>
+  )
+}
+
+export const NotificationItemCollectorProfileUpdatePromptFragmentContainer = createFragmentContainer(
```
damassi, line 107: Ditto -- new ideally should use hooks

PR #14164, opened 2024-07-04
fix: add Select option to the Select component values

dariakoko:

This PR solves [PROJECT-XX]

### Description

Addresses this [Notion ticket](https://www.notion.so/artsy/Is-this-the-correct-default-state-of-the-drop-down-Or-should-it-have-Purchase-information-inside--80be530b056b4bf5a5111b8cb2e99aef?pvs=4)
Add Select option to the Select component values to display empty value correctly

image

### DIFF
--- a/src/Apps/Sell/Utils/formUtils.ts
+++ b/src/Apps/Sell/Utils/formUtils.ts
@@ -1,15 +1,12 @@
 export const PROVENANCE_LIST = [
-  { value: "", text: "Select" },
-  {
-    value: "Purchased directly from gallery",
-    text: "Purchased directly from gallery",
-  },
-  {
-    value: "Purchased directly from artist",
-    text: "Purchased directly from artist",
-  },
-  { value: "Purchased at auction", text: "Purchased at auction" },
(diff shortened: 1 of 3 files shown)


src/Apps/Sell/Utils/formUtils.ts
```diff
@@ -1,11 +1,15 @@
...
-  "Purchased directly from artist",
-  "Purchased at auction",
-  "Gift from the artist",
-  "Other",
-  "I don’t know",
-].map(provenance => ({
-  value: provenance,
-  text: provenance,
-}))
+  { value: "", text: "Select" },
```
olerichter00, line 2: To avoid the duplications, we could also use mal and do something like

text: value || "Select"

olerichter00 (commented):
Thanks 🙏

src/Apps/Sell/Utils/formUtils.ts
dariakoko: 🙈 you're right

PR #14165, opened 2024-07-04
fix: Display preview images in reverse order in Sell flow

olerichter00:
Solves https://www.notion.so/artsy/Photo-upload-given-that-on-small-screens-the-whole-grid-of-uploaded-photos-is-not-visible-does-it-828252b08d5049cbad9ae71b734a2ccb?pvs=4
## Description

Display preview images in reverse order in the Sell flow to ensure that the most recent photos are always visible. This is helpful when many photos are uploaded and not all preview images fit on the screen.

### DIFF
--- a/src/Apps/Sell/Components/ImagePreviewsGrid.tsx
+++ b/src/Apps/Sell/Components/ImagePreviewsGrid.tsx
@@ -15,7 +15,8 @@ export const ImagePreviewsGrid: React.FC = () => {
       <Spacer y={2} />
 
       <Flex gap={1} flexWrap="wrap">
-        {photos.map(photo => (
+        {/* Displaying the images in reverse order to ensure the most recent image is visible. */}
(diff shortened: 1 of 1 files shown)


src/Apps/Sell/Components/ImagePreviewsGrid.tsx
```diff
@@ -15,7 +15,8 @@ export const ImagePreviewsGrid: React.FC = () => {
       <Spacer y={2} />
 
       <Flex gap={1} flexWrap="wrap">
-        {photos.map(photo => (
+        {/* Displaying the images in reverse order to ensure the most recent image is visible. */}
+        {[...photos].reverse().map(photo => (
```
olerichter00, line 19: Creating a copy of the photos array to avoid mutating the original: `[...photos]`.

nickskalkin (approved):
👍

PR #14174, opened 2024-07-08
fix: Replace loading spinner on Sell flow image preview

olerichter00:
## Description

Replace loading spinner on Sell flow image preview.

### DIFF
--- a/src/Apps/Sell/Components/ImagePreviewItem.tsx
+++ b/src/Apps/Sell/Components/ImagePreviewItem.tsx
@@ -1,12 +1,5 @@
 import CloseFillIcon from "@artsy/icons/CloseFillIcon"
-import {
-  Box,
-  Clickable,
-  Flex,
-  Image,
-  ProgressBar,
-  Spinner,
-} from "@artsy/palette"
+import { Box, Clickable, Flex, Image, ProgressBar, Text } from "@artsy/palette"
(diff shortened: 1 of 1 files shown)


src/Apps/Sell/Components/ImagePreviewItem.tsx
```diff
@@ -119,7 +113,13 @@ export const ImagePreviewItem: React.FC<ImagePreviewItemProps> = ({
           />
         )}
 
-        {photo.loading && photo.progress && (
+        {!!isProcessing && !isLoading && (
```
olerichter00, line 116: When loading the image fails or Gemini hasn't processed the image yet, we show the "processing..." message instead of the spinner. 

The spinner has worked well, but can be confusing when the user reloads the page before Gemini has processed the images. This results in a spinner that never stops because there is no reloading logic.

nickskalkin (approved):

MounirDhahri (approved):
This is a nice addition. I think it makes sense for the time being.

When gemini finishes processing the image, does it show up later? or only after going back to the screen again

olerichter00:
> When gemini finishes processing the image, does it show up later? or only after going back to the screen again

When Gemini finishes processing the image, the user still needs to reload the screen before the image shows up. But I guess this is ok and when "processing..." shows up, it's quite clear what it means.

PR #14176, opened 2024-07-08
fix: Sell flow image preview onError logic

olerichter00:
## Description

Sell flow image preview onError logic.


src/Apps/Sell/Components/ImagePreviewItem.tsx
```diff
@@ -98,6 +98,8 @@ export const ImagePreviewItem: React.FC<ImagePreviewItemProps> = ({
                 objectFit: "contain",
               }}
               onError={() => {
+                if (!photoSrc.startsWith("data:")) return
```
olerichter00, line 101: We only want to show the "Processing..." message when the uploaded file itself cannot be displayed. When the image comes from Gemini and cannot be loaded, probably something else is wrong.

MounirDhahri (approved):

PR #14184, opened 2024-07-10
fix: switching from saved pickup back to shipping

erikdstock:

This PR solves [EMI-1957]

### Description

This PR fixes a bug where you cannot switch from pickup back to shipping if the order loads with a saved pickup fulfillment. The issue was that we were resetting the form to its initial values when changing from pickup to shipping, but those initial values were for a pickup fulfillment if that was what was saved to the order.

Added tests as todos, but existing tests passed.

[EMI-1957]: https://artsyproduct.atlassian.net/browse/EMI-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
+++ b/src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
@@ -300,8 +300,7 @@ const ME_FRAGMENT = graphql`
  */
 const getInitialValues = (
   meData: ShippingContextProps["meData"],
-  orderData: ShippingContextProps["orderData"],
-  forceNewAddressFormMode?: boolean
+  orderData: ShippingContextProps["orderData"]
 ): FulfillmentValues => {
   if (orderData.savedFulfillmentDetails) {
     return {
--- a/src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
+++ b/src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
@@ -82,29 +82,28 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = ({
   }, [hasSavedAddresses])
 
   /*
-   * Re-save fulfillment details on load if they are already saved
-   * and shipping quotes need refreshing for new address mode only
+   * Re-save fulfillment details on load if they are already saved &
+   * require artsy shipping (new address form mode only - saved addresses
+   * handle this separately)
    */
   useEffect(() => {
-    const existingFulfillmentDetails =
-      shippingContext.orderData.savedFulfillmentDetails
+    const { savedFulfillmentDetails } = shippingContext.orderData
 
     const isArtsyShippingSaved =
-      existingFulfillmentDetails?.fulfillmentType === FulfillmentType.SHIP &&
-      shippingContext.orderData.requiresArtsyShippingTo(
-        existingFulfillmentDetails.attributes.country
-      )
+      savedFulfillmentDetails?.fulfillmentType === FulfillmentType.SHIP &&
+      savedFulfillmentDetails.isArtsyShipping &&
+      shippingContext.meData.addressList.length === 0
 
     if (isArtsyShippingSaved) {
       const refreshShippingQuotes = async () => {
         // instead of handleSubmit, call the save fulfillment details function
         // directly
         const result = await handleSaveFulfillmentDetails({
-          attributes: existingFulfillmentDetails.attributes,
+          attributes: savedFulfillmentDetails.attributes,
           fulfillmentType: FulfillmentType.SHIP,
           meta: {
             // FIXME: Will clobber previous address verification (but we can't
-            // know what the previous status was unless we read from server)
+            // know what the previous status was until we can read from server)
             addressVerifiedBy: null,
           },
         })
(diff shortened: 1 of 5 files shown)


src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
```diff
@@ -83,7 +83,7 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = ({
 
   /*
    * Re-save fulfillment details on load if they are already saved
-   * and shipping quotes need refreshing for new address mode only
+   * and shipping quotes need refreshing for new address mode only ?? Is this correct?
```
erikdstock, line 86: I should probably stare at this until I can delete the new text here.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
```diff
@@ -319,40 +319,9 @@ const getInitialValues = (
...
 
-  // The default ship-to address should be the first one that
-  // can be shipped-to, preferring the default
-
-  const defaultUserAddress = getDefaultUserAddress(
+  return getInitialShippingValues(
     savedAddresses,
+    orderData.shipsFrom,
     orderData.availableShippingCountries
   )
```
erikdstock, line 326: Moved the 'calculate initial shipping address from the order' into its own function.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
oxaudo: Yea - the way I read it here it's not really checking for `new mode` below. It seems to always refresh artsy quotes and update fulfillment details  if `Ship` is selected and works is artsy shipped to a given address

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
```diff
@@ -190,47 +193,49 @@ const FulfillmentDetailsFormLayout = (
...
-        if (values.fulfillmentType === FulfillmentType.SHIP) {
-          // reset to current initialValues (based on calculation in FulfillmentDetails.tsx)
-          formikContext.resetForm()
-        }
+    shippingContext.actions.goBackToFulfillmentDetails()
+    await setTouched({})
+
+    if (newValue === FulfillmentType.PICKUP) {
+      const initialValuesForPickup: PickupValues = {
+        // ...values,
```
oxaudo, line 208: Did we mean to leave those `values` out here?

src/Apps/Order/Routes/Shipping2/Utils/shippingUtils.ts
```diff
@@ -135,11 +135,61 @@ export const getDefaultUserAddress = (
...
+      attributes: attributesFromDefaultAddress,
+      meta: {
+        saveAddress: false,
+        addressVerifiedBy: null,
+      },
+    }
+  }
+
+  // The user doesn't have a valid ship-to address, so we'll return empty values.
+  // TODO: This doesn't account for matching the saved address id
```
oxaudo, line 175: This TODO is confusing to me. But as long as you know what to do here? Maybe if it's important we could create an issue to describe what needed to be done here and link it here if this refactor is important.

Does this say that if we reload the page we are not setting it to the address that was previously saved on the server and instead to the first one that fits?

src/Apps/Order/Routes/Shipping2/__tests__/Shipping2.jest.tsx
```diff
@@ -319,6 +319,14 @@ describe.skip("Shipping", () => {
...
   })
 
+  describe("initial load with order data", () => {
+    it.todo("loads with saved shipping fulfillment and no saved addresses")
+    it.todo(
+      "loads with saved shipping fulfillment and matching saved addresses"
+    )
+    it.todo("loads with saved pickup fulfillment")
+  })
+
```
oxaudo, line 329: Maybe add a story for this too so that we don't forget?

oxaudo (approved):
This looks good to me.

Readability of what is happening here is still a bit tricky and maybe can be improved with some variable names changes as a first step but that is by no means blocking. Pairing with you have me reasonably convinced that this functions as we want. Or at lest better than it currently does :)

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetailsForm.tsx
erikdstock: Yes - we're completely re-setting these values to empty, so there is nothing to retain here.

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
```diff
@@ -83,7 +83,7 @@ export const FulfillmentDetails: FC<FulfillmentDetailsProps> = ({
 
   /*
    * Re-save fulfillment details on load if they are already saved
-   * and shipping quotes need refreshing for new address mode only
+   * and shipping quotes need refreshing for new address mode only ?? Is this correct?
    */
   useEffect(() => {
     const existingFulfillmentDetails =
```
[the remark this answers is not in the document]
erikdstock: Good point, I took the first suggestion (though the second might be more explicit, it would just be a larger rename)

src/Apps/Order/Routes/Shipping2/Components/FulfillmentDetails.tsx
erikdstock: I read through this and thought about how it works. Updated here, can share more details in slack.

src/Apps/Order/Routes/Shipping2/Utils/shippingUtils.ts
erikdstock: 🔪

PR #14182, opened 2024-07-09
feat: Post Approval Sell flow Steps

olerichter00:
The type of this PR is: **Feature**

* This PR solves [ONYX-1082](https://artsyproduct.atlassian.net/browse/ONYX-1082)
* Feature Flag: [onyx_post_approval_submission_flow](https://unleash.artsy.net/projects/default/features/onyx_post_approval_submission_flow)

> [!TIP]
> Sorry for the large PR! But actually, most of the code is just boilerplate for the new routes. The interesting stuff is [here](https://github.com/artsy/force/pull/14182/files#diff-959233afb6950a1614a14af87d938868ff446b983af8ad3e1ab74343decbcf6c).

### Description

This PR implements the extended sell flow for approved submissions. In case the submission is already approved, we extend the flow with additional steps.

The new steps are not completly implemented; I've just added some scaffolding and tests.

### Screenshots

Screenshot 2024-07-10 at 12 31 07

Screenshot 2024-07-10 at 12 30 21

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ONYX-1082]: https://artsyproduct.atlassian.net/browse/ONYX-1082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14188, opened 2024-07-10
chore: Use submission's `internalID` for Sell flow tracking

olerichter00:
Request from @daytavares  

## Description

Currently, we send the submission's `externalID` when tracking Sell flow events. Instead, we need to use the internal submission ID.

### DIFF
--- a/src/Apps/Sell/Routes/SubmissionRoute.tsx
+++ b/src/Apps/Sell/Routes/SubmissionRoute.tsx
@@ -28,10 +28,7 @@ export const SubmissionRoute: React.FC<SubmissionRouteProps> = props => {
       <AppContainer>
         <SellMeta />
 
-        <SellFlowContextProvider
-          submissionID={submission.externalId}
-          internalSubmissionID={submission.internalID}
-        >
+        <SellFlowContextProvider submission={submission}>
           {props.children}
         </SellFlowContextProvider>
       </AppContainer>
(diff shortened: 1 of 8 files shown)


MounirDhahri:
How is this new internal id different than the `submissionID` ? do we need to update something in app as well?

src/Apps/Sell/Routes/SubmissionRoute.tsx
```diff
@@ -28,7 +28,10 @@ export const SubmissionRoute: React.FC<SubmissionRouteProps> = props => {
       <AppContainer>
         <SellMeta />
 
-        <SellFlowContextProvider submissionID={submission.externalId}>
+        <SellFlowContextProvider
+          submissionID={submission.externalId}
+          internalSubmissionID={submission.internalID}
```
olerichter00, line 33: Instead of passing two props, we could only use one `submission` prop (thanks @dblandin https://github.com/artsy/force/pull/14182#discussion_r1672301184)

MounirDhahri: oki

olerichter00: updated it.

olerichter00:
> How is this new internal id different than the submissionID ? do we need to update something in app as well?

It's a secure ID that we use for the URL in Force (e.g. `71a7ba7d-2a82-499a-b942-5d215652e3dc`) but for tracking, we need the internal ID for joins. I don't think we have to do anything in Eigen.

MounirDhahri:
perfect - thank you 🙏

PR #14191, opened 2024-07-10
fix(partner): explicitly request articles sorted by pub date

anandaroop:

This PR solves the issue reported [here](https://artsy.slack.com/archives/C03N12SR0RK/p1720172832143009)

### Description

Articles displayed on the partner profile page are not sorted, which means they use the [upstream default](https://github.com/artsy/positron/blob/9ea829fefe6bb1118131c0fc3ff429e357507afe/src/api/apps/articles/model/retrieve.js#L189-L191) of sorting by _last updated_ date, instead of the _publication_ date.

The updated date often matches the published date — but not always, as the report shows. We should instead by explicitly requesting these articles in order of publish date.

Note that we already do this everywhere else we request articles — the current case must have just been an oversight 🙈 

## Before

bb

## After

aa


ansor4 (approved):
Thank you!!

PR #14195, opened 2024-07-11
fix: make manual prepending of US account for possibility of 'under' in price

oxaudo:
This should resolve [EMI-1843](https://artsyproduct.atlassian.net/browse/EMI-1843)

The solution here is pretty ugly but I think we need to maintain this custom function here as we can not use just streight 'saleMessage' for the artwork from the server. Sometimes this list price is not really the list price but rather 'last negotiated price' (there is a separate ticket to figure out proper display name here). But for now just modify this value client side to account for possible 'Under ' prefix there.

Before: 
Screen Shot 2024-07-11 at 2 22 58 PM

After:
Screen Shot 2024-07-11 at 2 22 36 PM

Also in the process I encountered a strange crash with works with very high value. Seems like in this case for whatever reason the server does not give us partner information(?!) - did not track the cause but made UI a bit more bullet proof if that is the case.

[EMI-1843]: https://artsyproduct.atlassian.net/browse/EMI-1843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Utils/currencyUtils.ts
+++ b/src/Apps/Order/Utils/currencyUtils.ts
@@ -5,7 +5,8 @@ export const appendCurrencySymbol = (
   const pricePresent: boolean =
     typeof price === "string" || typeof price == "number"
   if (currency === "USD" && pricePresent) {
-    return "US" + price
+    const parts = String(price).split("Under ")
+    return parts.length > 1 ? "Under US" + parts[1] : "US" + parts[0]
   }
   return price
 }
--- a/src/Apps/Order/Components/ArtworkSummaryItem.tsx
+++ b/src/Apps/Order/Components/ArtworkSummaryItem.tsx
@@ -21,18 +21,12 @@ export interface ArtworkSummaryItemProps extends Omit<FlexProps, "order"> {
 }
 
 const ArtworkSummaryItem: React.FC<ArtworkSummaryItemProps> = ({
-  order: {
-    lineItems,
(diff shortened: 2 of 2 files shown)


src/Apps/Order/Utils/currencyUtils.ts
```diff
@@ -5,7 +5,8 @@ export const appendCurrencySymbol = (
   const pricePresent: boolean =
     typeof price === "string" || typeof price == "number"
   if (currency === "USD" && pricePresent) {
-    return "US" + price
+    const parts = String(price).split("Under ")
+    return parts.length > 1 ? "Under US" + parts[1] : "US" + parts[0]
```
leamotta, line 9: I kinda wish we would get rid of the "US" prefix at checkout, like we did in the order confirmation emails, since there is only one $ currency possible at this point.

leamotta (approved):
Not the prettiest but I can't think of a better solution 😅

src/Apps/Order/Utils/currencyUtils.ts
oxaudo: I'm cool with that but it's not what the card is asking. I'm wondering how much it's adding to anything but it is what it is for now. Happy to remove it if you think nobody will notice. Then we can just display straight `price` here....

PR #14201, opened 2024-07-12
fix: handle disliked artworks correctly

mc-jones:
This PR patches #14194 to correctly handle disliked artworks. We are currently persisting all choices as "liked" so this will correct the bug and should improve the recommendation results.


anandaroop (approved):
🙏🏽

PR #14205, opened 2024-07-15
fix: Filter out null notifications in Activity panel        

olerichter00:

This PR solves https://artsy.slack.com/archives/C05EQL4R5N0/p1721034091096429

### Description

We should filter out `null` notifications in the Activity panel. This PR fixes the types of `shouldDisplayNotification` and adds a check that filters out `null` notifications to avoid the following error:

screenshot_2024-07-15_at_11 56 03_480


nickskalkin (approved):
thanks! Do you know if we have the same issue in Eigen?

PR #14212, opened 2024-07-17
feat(nav): removes auth button tier on mobile in favor of person icon

dzucconi:
Removes the top Login / Sign up buttons on mobile and replaces that with a person icon which links to the login page. Waiting on confirmation that that is indeed the page we want to link to.

So now:


damassi:
Mind adding a screenshot?

PR #14213, opened 2024-07-17
feat(ONYX-1114): add conditions step

dariakoko:

This PR solves [ONYX-1114]

### Description

Add conditions step

| Web | mWeb | Dialog Web | Dialog mWeb |
| --- | --- | --- | --- |
| image | image | image | image |

[ONYX-1114]: https://artsyproduct.atlassian.net/browse/ONYX-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Sell/Routes/__tests__/ConditionRoute.jest.tsx
```diff
@@ -85,70 +97,86 @@ describe("ConditionRoute", () => {
...
-            expect.objectContaining({
-              variables: {
-                input: {
-                  externalId: '<mock-value-for-field-"externalId">',
-                  state: "SUBMITTED",
-                },
-              },
-            })
-          )
+  /* describe("navigation", () => {
```
dariakoko, line 153: Navigation tests are failing, having troubles figuring out why

src/Apps/Sell/Routes/AdditionalRoutes/ConditionRoute.tsx
```diff
@@ -1,33 +1,48 @@
...
+import { acceptableConditionsForSubmission } from "Apps/Sell/Utils/acceptableConditionsForSubmission"
+import { ConditionInfoModal } from "Apps/Artwork/Components/ArtworkDetails/ConditionInfoModal"
 
 const FRAGMENT = graphql`
   fragment ConditionRoute_submission on ConsignmentSubmission {
-    title
     myCollectionArtwork {
+      artworkId: internalID
       condition {
+        description
```
MounirDhahri, line 30: I think the description value should be coming from `conditionDescription.details` instead https://github.com/artsy/eigen/blob/6db51a8103ac59d50983e437f2cb538b6362ab9d/data/schema.graphql#L2368
It seems to be the one used in prod and from cms https://github.com/artsy/eigen/blob/6db51a8103ac59d50983e437f2cb538b6362ab9d/data/schema.graphql#L2368

MounirDhahri (commented):
Nice - LGTM apart 2 small comments:
- I think we need to look into `conditionDescription` vs `condition.description` though. 
- Maybe we can look into fixing the broken tests over the KS?

src/Apps/Sell/Components/SubmissionStepTitle.tsx
```diff
@@ -2,7 +2,11 @@ import { Text } from "@artsy/palette"
 
 export const SubmissionStepTitle: React.FC = ({ children }) => {
   return (
-    <Text variant={["lg-display", "xl"]} mb={2}>
+    <Text
+      variant={["lg-display", "xl"]}
+      mb={2}
+      data-testid="submission-step-title"
```
olerichter00, line 8: 👏

src/Apps/Sell/Routes/__tests__/ConditionRoute.jest.tsx
olerichter00: You need to set the submission state to "APPROVED" because this is a step that is not reachable when the submission is in state "DRAFT" or "SUBMITTED" (Tier 2 step).

[You can have a look at this test file for another additional/tier 2 route:](https://github.com/artsy/force/blob/827a09088f3df25ad3a42df553e340aed56e5a91/src/Apps/Sell/Routes/__tests__/ShippingLocationRoute.jest.tsx#L194)

src/Apps/Sell/Utils/acceptableConditionsForSubmission.tsx
```diff
@@ -0,0 +1,58 @@
...
+    .reduce((accumulator, current) => {
+      if (current && current.trim()) {
+        accumulator.push(current.toUpperCase())
+      }
+      return accumulator
+    }, [] as string[])
+    .join("_") as AcceptableConditionValue
+}
+
+export const acceptableConditionsForSubmission = () => {
```
olerichter00, line 44: Hmm, what is this function doing and why can't we just use `conditionOptions` or a hardcoded array? This function looks very complicated and it does not have any input and would always return the same array anyway, right?

olerichter00 (commented):
Looks great! I just wonder if we really need this `acceptableConditionsForSubmission` function.

src/Apps/Sell/Utils/acceptableConditionsForSubmission.tsx
dariakoko: I thought since we have a `ArtworkConditionEnumType` defined, we could make sure when it changes, we will be forced to update the array we use for the `<Select />` component. If we hardcode the values, there is a chance we miss this change. 
I followed the implementation from `acceptableCategoriesForSubmission.ts`
If it sounds like too much, we totally can use `conditionOptios` or a hardcoded array

src/Apps/Sell/Routes/AdditionalRoutes/ConditionRoute.tsx
dariakoko: You're right, will address

src/Apps/Sell/Utils/acceptableConditionsForSubmission.tsx
olerichter00: I've checked it, and I'm not quite sure if it works. When I change a value in `conditionOptions`, I don't see a TypeScript error.

Could we alternatively just use a typed array in `ConditionRoute.tsx` and remove `acceptableConditionsForSubmission.tsx` and `const conditionOptions = useMemo(acceptableConditionsForSubmission, [])`? 

The following should give us a TypeScript error whenever the enum in the schema changes:

```typescript
// This type is here to make sure the condition options match the allowed values in the schema. We have to update the conditionOptions array whenever the ArtworkConditionEnumType changes.
type ConditionOption = {
  text: string
  value: Exclude<ArtworkConditionEnumType, "%future added value"> | ""
}

export const conditionOptions: ConditionOption[] = [
  { text: "Select Condition", value: "" },
  { text: "Excellent Condition", value: "EXCELLENT" },
  { text: "Fair Condition", value: "FAIR" },
  { text: "Good Condition", value: "GOOD" },
  { text: "Very Good Condition", value: "VERY_GOOD" },
]
```

src/Apps/Sell/Routes/__tests__/ConditionRoute.jest.tsx
dariakoko: Hey @olerichter00, Conditions is the final step of tier2, so, i think, the status should be RESUBMITTED or similar (what you discussed with Devon on slack)?
So, I should probably edit the SellFlowContext file like this when we introduce a new state:
```
if (submission?.state === "DRAFT") {
      await updateSubmission({
        state: "SUBMITTED",
      })
    } else if(submission?.state === "APPROVED") {
      await updateSubmission({
        state: "RESUBMITTED",
      })
    }
```

dariakoko: Maybe we could pair on the failing tests today? Would you have time?

src/Apps/Sell/Utils/acceptableConditionsForSubmission.tsx
```diff
@@ -0,0 +1,15 @@
+import { ArtworkConditionEnumType } from "__generated__/useUpdateArtworkMutation.graphql"
+
+// This type is here to make sure the condition options match the allowed values in the schema. We have to update the conditionOptions array whenever the ArtworkConditionEnumType changes.
+type ConditionOption = {
+  text: string
+  value: Exclude<ArtworkConditionEnumType, "%future added value"> | ""
+}
+
+export const conditionOptions: ConditionOption[] = [
```
olerichter00, line 9: What do you think about renaming this file to `conditionOptions.tsx` so it's easier to find? 🤔

olerichter00 (dismissed):
🌟 🌟 🌟

PR #14200, opened 2024-07-12
[WIP] feat(EMI-1997): artwork card collectors signals line

rquartararo:
The type of this PR is: **TYPE**

This PR solves [EMI-1997]

### Description

WIP PR for adding a new line to the artwork grid card details. We render a new (5th) line for a collector signal, but if there is no signal we render an empty line at the bottom to keep all artwork lines consistent. 

Screenshot 2024-07-12 at 11 46 08 AM

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMI-1997]: https://artsyproduct.atlassian.net/browse/EMI-1997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14218, opened 2024-07-18
feat: render consistent number of lines on artwork brick for collector signals

starsirius:
(This pulls the pure visual part out of https://github.com/artsy/force/pull/14200 and focus on [supporting consistent number of lines in artwork bricks, under a given context](https://artsy.slack.com/archives/C02JHHHKP5K/p1720458384696129).)

This PR solves [EMI-1997]

## Description

For the social proof and urgency signals project, some artworks might render additional lines for, e.g. an "increased demand" badge. We want to render consistent number of lines in artwork bricks under a given context, like a rail or a grid, to make sure, for example, artworks in a rail are properly aligned.

The basic idea here is to rely a `context`, so inside the artwork brick, it knows the exact lines that are possible to render under that context, and pads with empty lines when some are not applicable. For example, auction artworks [uses a similar context](https://github.com/artsy/force/blob/f09889ce22d12d566bfd82f8a15bafbbd21da187/src/Apps/Auction/Components/AuctionArtworkFilter.tsx#L33) to render the "lot" label.

Since we already know we'll need such context for at least artwork grids and rails. I extract a shared context for collector signal purposes.

<details><summary>Screenshot: Limited-time offer badge in an artwork grid</summary>
<p>

Screenshot 2024-07-17 at 11 37 51 PM

</p>
</details> 

<details><summary>Screenshot: Limited-time offer badge on a rail</summary>
<p>

Screenshot 2024-07-17 at 11 35 46 PM

</p>
</details> 

<details><summary>Screenshot: Rails without the context should remain the same height/number of lines</summary>
<p>

Screenshot 2024-07-17 at 11 53 26 PM

</p>
</details> 

cc @artsy/emerald-devs 

## Todos

- [ ] Placeholder height is out-of-sync

[EMI-1997]: https://artsyproduct.atlassian.net/browse/EMI-1997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListArtworksGrid.tsx
```diff
@@ -156,29 +157,31 @@ const ArtworkListArtworksGrid: FC<ArtworkListArtworksGridProps> = ({
...
-                contextPageOwnerId,
-                destinationPageOwnerId: artwork.internalID,
-                destinationPageOwnerSlug: artwork.slug,
-                position: artworkIndex,
-                sort: context?.filters?.sort,
-              })
-            )
-          }}
-        />
+        <CollectorSignalsContextProvider showCollectorSignalBadge>
```
starsirius, line 160: Wrap the artwork grid inside this context.

src/Apps/Home/Components/HomeEmergingPicksArtworksRail.tsx
```diff
@@ -33,48 +34,50 @@ export const HomeEmergingPicksArtworksRail: React.FC<HomeEmergingPicksArtworksRa
...
-                destination_page_owner_slug: artwork.slug,
-                type: "thumbnail",
-              }
-              trackEvent(trackingEvent)
-            }}
-          />
-        ))
-      }}
-    />
+    <CollectorSignalsContextProvider showCollectorSignalBadge>
```
starsirius, line 37: Wrap the rail inside the same context.

src/Components/Artwork/CollectorSignalsContext.tsx
```diff
@@ -0,0 +1,25 @@
+import { createContext, useContext } from "react"
+
+interface CollectorSignalsContextProps {
+  showCollectorSignalBadge?: boolean
+}
+
+const CollectorSignalsContext = createContext<CollectorSignalsContextProps>({
+  showCollectorSignalBadge: false,
```
starsirius, line 8: Specifically for this line, which can be one of the signals, like "Limited-time offer", "Increased demand", etc. I call it badge but open to naming suggestions.

src/Components/Artwork/CollectorSignalsContext.tsx
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+  return (
+    <CollectorSignalsContext.Provider value={rest}>
+      {children}
+    </CollectorSignalsContext.Provider>
+  )
+}
+
+export const useCollectorSignalsContext = () => {
+  const context = useContext(CollectorSignalsContext) ?? {}
```
starsirius, line 23: I still need to feature flag.

src/Components/Artwork/Details.tsx
```diff
@@ -203,12 +223,22 @@ export const Details: React.FC<DetailsProps> = ({
...
+
   const isP1Artist = rest?.artwork.artist?.targetSupply?.isP1
   const isHighDemand =
     Number((rest?.artwork.marketPriceInsights?.demandRank || 0) * 10) >= 9
 
   const showHighDemandInfo = !!isP1Artist && isHighDemand && showHighDemandIcon
 
+  // TODO: replace randomization with real signals and business logic
+  const hasCollectorSignalBadge: boolean = useMemo(
+    () => !!showCollectorSignalBadge && Math.random() < 0.2,
```
starsirius, line 236: Just for demo-ing the visual. We'll need to implement the desired data fetching and logic, in a separate PR.

src/Components/Artwork/CollectorSignalsContext.tsx
starsirius: Now I think about it, we probably need a flag for each applicable signals, like limited-time offer, increased demand, curator's pick, etc, to determine which signals to consider under a context (specific rail or grid). If that's aligned with folks' thought, I'll make the change.

Table

erikdstock: I've been thinking about [using the `ArtworkGridContext` for this](https://github.com/artsy/force/blob/10903dd8f595094e4560c2c453a978a4901b7d35/src/Components/ArtworkGrid/ArtworkGridContext.tsx#L7-L22) - Which I think should be the place we put these new context values rather than a new collector signals-specific context provider.

1 reason is that it already has a prop that does almost the same thing: `hideLotLabel`. It also has some other props that seem to be auction specific: `saveToDefaultList` and `isAuctionArtwork`. Taken together I think we should
- Use this context (which is not specific to artwork grids but [can wrap a collection like a rail as well](https://github.com/artsy/force/blob/246ba39f72ff05ccbeed171478e3f0afe3ce1631/src/Apps/Auction/Components/AuctionWorksByFollowedArtistsRail.tsx#L22-L42)) around all list views
- Make the existing auction-related props more intuitive
- **To your point**, allow some kind of prop to ignore specific signals in a given list.

damassi: Multiple contexts seem ok for this _assuming_ the feature will grow in complexity. Otherwise, for the time being, might be worth sharing this configuration and renaming the context. 

Another thought is: if this configuration option is going to be shared at a given collection level, surfacing this config from MP might be worthwhile, vs in code. And then that can be applied to Eigen.

damassi (dismissed):
This approach feels pretty vanilla and good 👍

starsirius:
OK I made a few more changes based on the feedback:

- [Reuse `ArtworkListArtworksGrid` for consistency](https://github.com/artsy/force/pull/14218/commits/a58016051c0e40ca43b57379689c99b1354f6420)
- [Rename flag to be collector signal-specific](https://github.com/artsy/force/pull/14218/commits/8fa06f20e5d85da8237fc4fb8d0e2ccf981d2429)
- [Guard collector signals line behind the feature flag](https://github.com/artsy/force/pull/14218/commits/d86720d6c4555d5bf9b61286489d5a8fcf86bd29)

If it looks good to you, I'm going to merge later today to unblock other work. 🙏

damassi (approved):
Looks great 👍

PR #14220, opened 2024-07-18
chore: update stripe

leamotta:

### Description

Needed to implement the credit card co-badged integration, as the version of `stripe-js` we use doesn't support the [onBehalfOf](https://docs.stripe.com/co-badged-cards-compliance?type=web-elements) property.
https://artsyproduct.atlassian.net/browse/EMI-1948

Updating the `react-stripe-js` version caused an unrelated test error with Bank Debit, locking the current version.


package.json
```diff
@@ -102,8 +102,8 @@
     "@sentry/integrations": "^6.7.2",
     "@sentry/node": "^6.7.1",
     "@sentry/tracing": "^6.7.1",
-    "@stripe/react-stripe-js": "^1.10.0",
-    "@stripe/stripe-js": "^1.34.0",
+    "@stripe/react-stripe-js": "1.10.0",
+    "@stripe/stripe-js": "^4.1.0",
```
starsirius, line 106: 💯 3 major versions bump!! I can't quickly tell from the release notes about breaking changes. Given it's a critical feature, I think it's worth QAing a bit on staging before releasing. I think Integrity covers the BNMO use cases pretty extensively, so we can probably manually test auction registration, card management in user settings, etc.

starsirius (approved):

PR #14221, opened 2024-07-18
feat(onyx-1112): sell frame step

nickskalkin:

This PR solves [ONYX-1112]

### Description

Adds Frame step:

image

[ONYX-1112]: https://artsyproduct.atlassian.net/browse/ONYX-1112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Sell/Mutations/useUpdateMyCollectionArtwork.tsx
+++ b/src/Apps/Sell/Mutations/useUpdateMyCollectionArtwork.tsx
@@ -1,0 +1,30 @@
+import { useMutation } from "Utils/Hooks/useMutation"
+import { graphql } from "react-relay"
+import { useUpdateMyCollectionArtworkMutation } from "__generated__/useUpdateMyCollectionArtworkMutation.graphql"
+
+export const useUpdateMyCollectionArtwork = () => {
+  return useMutation<useUpdateMyCollectionArtworkMutation>({
+    mutation: graphql`
+      mutation useUpdateMyCollectionArtworkMutation(
+        $input: MyCollectionUpdateArtworkInput!
+      ) {
+        myCollectionUpdateArtwork(input: $input) {
+          artworkOrError {
+            ... on MyCollectionArtworkMutationSuccess {
+              artwork {
+                title
+                internalID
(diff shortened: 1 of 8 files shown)


src/Apps/Sell/Mutations/useUpdateMyCollectionArtwork.tsx
```diff
@@ -0,0 +1,30 @@
...
+      mutation useUpdateMyCollectionArtworkMutation(
+        $input: MyCollectionUpdateArtworkInput!
+      ) {
+        myCollectionUpdateArtwork(input: $input) {
+          artworkOrError {
+            ... on MyCollectionArtworkMutationSuccess {
+              artwork {
+                title
+                internalID
+              }
```
dariakoko, line 17: What do you think about returning images here as well?

nickskalkin: Do we need images here?

dariakoko (approved):

src/Apps/Sell/Mutations/useUpdateMyCollectionArtwork.tsx
dariakoko: I was wondering, if the user added new images, went to see the updated artwork in my collection and the images are still processing, what would we show?

src/Apps/Sell/Routes/__tests__/FrameRoute.jest.tsx
```diff
@@ -35,10 +51,10 @@ beforeEach(() => {
       push: mockPush,
       replace: mockReplace,
     },
-    match: { location: { pathname: "/sell/submissions/submission-id/frame" } },
+    match: { location: { pathname: "/sell/submissions/externalId/frame" } },
```
dariakoko, line 54: Is externalId here the same as submission-id? Let's use submission-id and stay consistent

PR #14223, opened 2024-07-18
fix(loading-bar): Improve perf for safari

damassi:

### Description

Noticed that the loading bar was a bit shaky on safari, so moved the css animations to the GPU via `transform`.

### DIFF
--- a/src/System/Components/PageLoadingBar.tsx
+++ b/src/System/Components/PageLoadingBar.tsx
@@ -1,8 +1,6 @@
+import React, { useState, useEffect, useRef } from "react"
 import styled, { keyframes, css } from "styled-components"
-import { useState, useEffect, useRef } from "react"
 import { Box } from "@artsy/palette"
-import { isServer } from "Server/isServer"
-import { useDidMount } from "Utils/Hooks/useDidMount"
 import { Z } from "Apps/Components/constants"
 
 interface PageLoadingBarProps {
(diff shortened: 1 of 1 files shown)


src/System/Components/PageLoadingBar.tsx
```diff
@@ -44,18 +47,19 @@ export const PageLoadingBar: React.FC<PageLoadingBarProps> = ({
...
         isComplete={isComplete}
+        scaleX={scaleX}
         backgroundColor="brand"
         height={2}
         top="1px"
-        left={0}
-        width={isComplete ? barRef.current?.clientWidth : 0}
-        zIndex={100}
+        left="0px"
+        zIndex={1000}
```
dzucconi, line 55: https://github.com/artsy/force/blob/main/src/Apps/Components/constants.ts

src/System/Components/PageLoadingBar.tsx
```diff
@@ -44,18 +47,19 @@ export const PageLoadingBar: React.FC<PageLoadingBarProps> = ({
         ref={barRef as any}
         loading={loading}
         isComplete={isComplete}
+        scaleX={scaleX}
```
dzucconi, line 50: So there's only 2 values for this? 0 and 0.3? (Just want to be concerned with generating unnecessary styles)

src/System/Components/PageLoadingBar.tsx
```diff
@@ -68,31 +72,35 @@ const easeOutExpo = "cubic-bezier(0.19, 1, 0.22, 1)"
...
 `
 
-const LoadingBar = styled(Box)<{ loading: string; isComplete: boolean; width }>`
+const LoadingBar = styled(Box)<{
+  loading: string
+  isComplete: boolean
+  scaleX: number
+}>`
+  transform-origin: left;
+  transform: scaleX(${({ scaleX }) => scaleX});
```
dzucconi, line 103: I'd just set this direct with a `style` prop

src/System/Components/PageLoadingBar.tsx
damassi: this code is crazy, but all of it has a purpose. 0 -> .3 is the intial ease in, which then extends a bit for slow loads. Then... (next comment)

src/System/Components/PageLoadingBar.tsx
damassi: ...this is necessary to take the out transition from where its eased to, regardless of how much has already loaded.

src/System/Components/PageLoadingBar.tsx
damassi: Ah yah, actually this isn't needed at all and is irrelevant. I used Z.pageLoadBar up above 👍 Will remove

src/System/Components/PageLoadingBar.tsx
dzucconi: Oh sure, I'm saying you can set that value using style so that it's not generating a ton of classes.

src/System/Components/PageLoadingBar.tsx
damassi: https://github.com/artsy/force/pull/14224

src/System/Components/PageLoadingBar.tsx
dzucconi: But if it's just 2 values then it doesn't matter

PR #14251, opened 2024-07-29
feat(EMI-1956): display limited time offer signal

rquartararo:
The type of this PR is: **feature**

This PR solves [EMI-1956]

@artsy/emerald-devs 

### Description

This displays the "Limited-Time Offer" badge along with the countdown timer and price with discount to users who have an active partner offer on an artwork in these places:

- [x] Gallery detail page
- [x] Artist detail page
- [x] Recently Viewed Rail (all pages, including home feed and bottom of artwork detail page)
- [x] Home/Trending Now (rail)
- [x] Home/New works for you (rail)
- [x] Home/Curator’s Picks: Emerging Artists (rail)
- [x] Profile/Saves List (page)
- [x] Artwork detail page
- [x] Nav/Trending Now (page)
- [x] Nav/New this week (page)
- [x] Nav/Curator’s Picks: Emerging Artists (page)
- [x] Nav/Curator’s Picks: Blue Chip Artists (page)
- [x] Search Results/artworks (page)

Screenshot 2024-07-28 at 7 13 35 PM

Screenshot 2024-07-28 at 7 12 58 PM

[EMI-1956]: https://artsyproduct.atlassian.net/browse/EMI-1956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Artwork/Details.tsx
+++ b/src/Components/Artwork/Details.tsx
@@ -229,7 +229,7 @@ const BidInfo: React.FC<DetailsProps> = ({
 const ActivePartnerOfferTimer: React.FC<DetailsProps> = ({
   artwork: { collectorSignals },
 }) => {
-  const SEPARATOR = <>&nbsp;&nbsp;</>
+  const SEPARATOR = <>&nbsp;</>
   const { endAt } = collectorSignals?.partnerOffer ?? {}
   const { time } = useTimer(endAt ?? "")
   const { days, hours } = time


erikdstock:
Either as part of this PR or the next we can add tracking from artsy/cohesion#516 to these events. Some events on the artwork page are using the Deprecated Schema and will need to be updated to use the new event values: ClickedBid, ClickedContactGallery.

erikdstock (commented):
LGTM with a few comments and one bigger one, mostly on Details.tsx and its many sub-components. I wonder if this file can be simplified somehow. Maybe non-blocking, we can iterate on it.
- the `showActivePartnerOfferLine` prop is a [boolean](https://github.com/artsy/force/pull/14251/files?diff=unified&w=1#diff-9e4884fa02e562601553700c318d04eff9f3e5b962c609aa5b68bae462472850R35-R42) on `SaleInfoLine` and `SaleMessage`
- That prop is used to [conditionally render](https://github.com/artsy/force/pull/14251/files?diff=unified&w=1#diff-9e4884fa02e562601553700c318d04eff9f3e5b962c609aa5b68bae462472850R148) `ActivePartnerOfferTimer` and [the alternate pricing](https://github.com/artsy/force/pull/14251/files?diff=unified&w=1#diff-9e4884fa02e562601553700c318d04eff9f3e5b962c609aa5b68bae462472850R199-R200) in `SaleMessage`.
  - Each of these components then have to fetch the collector signals from the artwork and deal with a bunch of optional chaining again.
- Future concern: when we do tracking, we are doing to need to know at the artwork grid item click level what signals were visible on the card.

I'm thinking the last point is something we can answer later. For now what I would do is change that boolean prop to something like `partnerOfferSignal: PartnerOffer | null` and where we currently set the prop (previously randomized behind the feature flag) can just find the partner offer (which now definitely exists) or `?? null` and pass it down. Then ever subcomponent can check if it is present and proceed without needing to drill into the artwork again. How would this sound?

src/Components/Artwork/Details.tsx
```diff
@@ -150,16 +183,23 @@ const HighDemandInfo = () => {
...
 const NBSP = " "
 
-const SaleMessage: React.FC<DetailsProps> = ({
-  artwork: { sale, sale_message, sale_artwork },
-}) => {
+const SaleMessage: React.FC<SaleMessageProps> = props => {
+  const {
+    artwork: { sale, sale_message, sale_artwork, collectorSignals },
+    showActivePartnerOfferLine,
+  } = props
```
erikdstock, line 190: Out of scope of this PR, but something I would like to investigate: In order to query auction collector signals we have to query the sale/sale artwork. If we also request these fields on the artwork, does it add an unnecessary query, or do data loaders understand and skip redundant fetches?

PR #14254, opened 2024-07-30
feat(cache-proxy): Add cache invalidation on deploy / Prepare for release

damassi:

### Description

@mzikherman raised a scenario where the shape of the component changes, and due to the cached graphql response, yields an error. So this updates things to bust the redis cache on deploys. 

Also updates our logger to only log cache reads / writes during development so as to not clog up papertrail.


scripts/invalidate-server-cache.js
```diff
@@ -0,0 +1,14 @@
...
+
+require("dotenv/config")
+
+const redis = require("redis")
+
+const redisClient = redis.createClient({
+  url: process.env.OPENREDIS_URL,
+})
+
+redisClient.flushall(() => {
```
joeyAghion, line 11: I don't think we should flush the cache prior to releasing. There are still pods running with the old code at that point (and there will be for up to several more minutes). This causes them to slow down and increase upstream load right when we'd rather they complete their work and stop. It would be neater to just include a "digest" of the query in the cache key instead of the cache name. If you wanted to prevent cache from being shared across code versions _at all_, you'd include something that identifies the release in the cache key (like `ENV["DD_VERSION"]`). But that's aggressive... since this is a cache of Metaphysics data, there's no obvious reason a cached value from an older Force release should be invalidated.

damassi: Cool, I'll push up a commit that removes this!

damassi: I think we should feel it out in prod first.

damassi: https://github.com/artsy/force/pull/14259

PR #14265, opened 2024-07-30
feat: prompt for authentication for SWA > Add to MyC

dblandin:

This PR solves [ONYX-1145](https://artsyproduct.atlassian.net/browse/ONYX-1145)

### Description

- feat: prompt for authentication for SWA > Add to MyC

  After selecting an artist ineligible for submission, we suggest that the user
can add the artwork to My Collection instead. Since those surfaces require
authentication, trigger the auth modal in response to that button click.

- feat: redirect to login page if navigating to authenticated SWA route

  Applies some server-side redirects if we're navigating to /sell route known to
require authentication.

[ONYX-1145]: https://artsyproduct.atlassian.net/browse/ONYX-1145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Screen Recording

https://github.com/user-attachments/assets/0b93cc04-a472-4219-b2e9-4eece4b9e117

### DIFF
--- a/src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
+++ b/src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
@@ -1,10 +1,14 @@
+import { AuthIntent, ContextModule } from "@artsy/cohesion"
 import { Button, Flex, FullBleed, Spacer } from "@artsy/palette"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { ArtistNotEligiblText } from "Apps/Sell/Components/ArtistNotEligibleText"
 import { SubmissionLayout } from "Apps/Sell/Components/SubmissionLayout"
 import { SubmissionStepTitle } from "Apps/Sell/Components/SubmissionStepTitle"
+import { useAuthDialog } from "Components/AuthDialog"
 import { EntityHeaderArtistFragmentContainer } from "Components/EntityHeaders/EntityHeaderArtist"
 import { RouterLink } from "System/Components/RouterLink"
+import { useRouter } from "System/Hooks/useRouter"
+import { useSystemContext } from "System/Hooks/useSystemContext"
 import { ArtistNotEligibleRoute_artist$key } from "__generated__/ArtistNotEligibleRoute_artist.graphql"
 import * as React from "react"
 import { graphql, useFragment } from "react-relay"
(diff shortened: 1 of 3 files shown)


src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
```diff
@@ -20,6 +24,32 @@ interface ArtistNotEligibleRouteProps {
...
+    if (isLoggedIn) {
+      router.push("/collector-profile/my-collection/artworks/new")
+    } else {
+      showAuthDialog({
+        mode: "SignUp",
+        options: {
+          title: mode =>
+            mode === "Login"
+              ? "Log in to Add to My Collection"
+              : "Sign up to Add to My Collection",
```
olerichter00, line 41: That's the pattern we use in other auth dialogs ([examples](https://github.com/search?q=repo%3Aartsy%2Fforce+%22log+in+to%22&type=code)).
```suggestion
              ? "Log in to add to My Collection"
              : "Sign up to add to My Collection",
```

src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
```diff
@@ -42,10 +72,8 @@ export const ArtistNotEligibleRoute: React.FC<ArtistNotEligibleRouteProps> = pro
 
           <Flex flexDirection="column" justifyContent="space-between" pb={2}>
             <Button
-              // @ts-ignore
-              as={RouterLink}
-              to="/collector-profile/my-collection/artworks/new"
               data-testid="add-to-collection"
+              onClick={onAddToMyCollection}
```
olerichter00, line 76: Can we leave the `as={RouterLink}` in combination with the `onClick` so the button is a real link? Like this, it would be possible to do stuff like command+click to open the link in a new tab.

```suggestion
              // @ts-ignore
              as={RouterLink}
              to="/collector-profile/my-collection/artworks/new"
              onClick={onAddToMyCollection}
```

src/Apps/Sell/Routes/__tests__/ArtistNotEligibleRoute.jest.tsx
```diff
@@ -90,4 +90,38 @@ describe("ArtistNotEligibleRoute", () => {
...
+  describe("while logged out", () => {
+    const mockUseAuthDialog = useAuthDialog as jest.Mock
+
+    beforeEach(() => {
+      mockUseSystemContext.mockImplementation(() => {
+        return { isLoggedIn: false }
+      })
+    })
+
+    it("prompts for authentication", async () => {
```
olerichter00, line 103: 👏

olerichter00 (approved):
🌟 🌟 🌟

src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
dblandin: Good call – I'll just need to prevent the default behavior on click in the `onClick` handler. 👍🏼

src/Apps/Sell/Routes/ArtistNotEligibleRoute.tsx
```diff
@@ -20,6 +24,32 @@ interface ArtistNotEligibleRouteProps {
...
 export const ArtistNotEligibleRoute: React.FC<ArtistNotEligibleRouteProps> = props => {
   const artist = useFragment(FRAGMENT, props.artist)
+  const { showAuthDialog } = useAuthDialog()
+  const { router } = useRouter()
+  const { isLoggedIn } = useSystemContext()
+
+  const onAddToMyCollection = () => {
+    if (isLoggedIn) {
+      router.push("/collector-profile/my-collection/artworks/new")
+    } else {
```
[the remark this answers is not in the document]
dblandin: Yeah, I feel like early returns are useful when there's much more of the function definition to skip, but in this case it's clear that there are only two paths for the function flow.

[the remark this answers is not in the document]
olerichter00: I completely agree, it's pretty clear how it is. If there are just two paths, there is no need to use an early return.

Another argument, though, for early returns (even in this case) is to "highlight" the happy path, which is the only one that is not an early return. In this case, it would be `router.push("/collector-profile/my-collection/artworks/new")`.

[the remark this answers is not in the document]
dblandin: Switched to an early exit but also extracted the showAuthDialog behavior to keep things readable. 👍🏼

dblandin:
@olerichter00 Went ahead and reassigned to you for another review

olerichter00 (approved):

PR #14269, opened 2024-08-01
fix: Sell flow navigation logic

olerichter00:
## Description

This PR addresses a bug in the Sell flow navigation logic. The bug caused some redirects when accessing a step directly by pasting the URL in the browser.


src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -199,14 +199,7 @@ export const SellFlowContextProvider: React.FC<SellFlowContextProviderProps> = (
 
     router.push(`/sell/submissions/${submission?.externalId}/${newStep}`)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [
-    index,
-    isNewSubmission,
-    match.location.pathname,
```
olerichter00, line 205: We don't want to redirect when the URL changes.

MounirDhahri (approved):
Nice - I noticed that yesterday and thought it's because of me

PR #14270, opened 2024-08-01
fix: Sell flow Condition step link styling

olerichter00:

This PR solves https://www.notion.so/artsy/Condition-Definition-link-on-the-field-should-match-the-design-system-spacing-see-image-f641d103ac3b430cb48dbf2ed415b9cd?pvs=4

### Description

This PR fixes the Sell flow Condition step link styling.

| Before | After |
| --- | --- |
|Screenshot 2024-08-01 at 11 23 54 |Screenshot 2024-08-01 at 11 22 58| 

### DIFF
--- a/src/Apps/Sell/Components/BottomFormNavigation.tsx
+++ b/src/Apps/Sell/Components/BottomFormNavigation.tsx
@@ -83,7 +83,7 @@ const BottomFormNextButton = () => {
   const { value, clearValue } = useAuthIntent()
 
   const {
-    actions: { goToNextStep},
+    actions: { goToNextStep },
     state: { submission, isSubmitStep, nextStep, loading },
   } = useSellFlowContext()
 
@@ -129,7 +129,7 @@ const BottomFormNextButton = () => {
     try {
       await submitForm()
 
-      await goToNextStep()
+      await goToNextStepCallback()
     } catch (error) {
       logger.error("Error submitting form", error)
     } finally {
@@ -149,7 +149,11 @@ const BottomFormNextButton = () => {
     } finally {
       setIsSubmitting(false)
     }
-  }, [associateSubmissionMutation, submitAndNavigateToNextStepCallback, submission])
+  }, [
+    associateSubmissionMutation,
+    submitAndNavigateToNextStepCallback,
+    submission,
+  ])
 
   useEffect(() => {
     if (value?.action === "submitSubmission") {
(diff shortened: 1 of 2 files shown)


src/Apps/Sell/Components/BottomFormNavigation.tsx
```diff
@@ -149,7 +149,11 @@ const BottomFormNextButton = () => {
     } finally {
       setIsSubmitting(false)
     }
-  }, [associateSubmissionMutation, submitAndNavigateToNextStepCallback, submission])
+  }, [
+    associateSubmissionMutation,
+    submitAndNavigateToNextStepCallback,
+    submission,
+  ])
```
olerichter00, line 156: Just fixing some linting errors because the dependency array wasn't correct.

src/Apps/Sell/Routes/AdditionalRoutes/ConditionRoute.tsx
```diff
@@ -52,16 +53,24 @@ interface ConditionRouteProps {
...
-  if (!artwork) return null
+  if (!artwork) {
+    return (
+      <ErrorPage
+        code="Something went wrong."
+        message="The artwork could not be loaded. Please try again or contact support@artsy.net."
+        m={4}
+      />
+    )
+  }
```
olerichter00, line 73: Displaying a proper error page in case the My Collection artwork isn't present or cannot be loaded.

olerichter00: We probably need to ensure that the associated MyC artwork cannot be deleted when it has an associated submission. If the artwork no longer exists, it is not possible to access this step and update the condition .

dariakoko: If the artwork that has submission is removed from MyCollection, submission will still exist, right? What happens to the submission in this case? 
Probably an edge-case, asking mostly out of curiosity in case we talked about this before

dariakoko (approved):

src/Apps/Sell/Routes/AdditionalRoutes/ConditionRoute.tsx
olerichter00: That's a good question! I will investigate 🔍  We should probably not allow removing the MyC artwork when it has a submission that is not in Draft state.

olerichter00: https://artsyproduct.atlassian.net/browse/ONYX-1156

olerichter00: cc @MounirDhahri

PR #14274, opened 2024-08-01
feat: offer quick filters for featured keywords on fairs (WIP)

joeyAghion:
Companion to https://github.com/artsy/gravity/pull/17997 and depends on https://github.com/artsy/metaphysics/pull/5878 (draft). When a fair has featured keywords, offer quick-filter "pills" with those keywords. Clicking should populate the existing keyword filter field and refresh artwork results. Clicking again should clear the keyword filter and refresh results.

I'm not sure if these should _always_ appear on the fair's artworks page, or only in the base case of no active filters. If they always appear, there's a significant chance that _further_ filtering by these keywords leads to empty results.


joeyAghion:
@dzucconi I could use your input about whether I'm going about this incorrectly. These pills seem to operate correctly, _except_ they don't update the keyword field as expected in the "all filters" sidebar. The URL _is_ updated and the sidebar is even initialized correctly on a full navigation, but not upon clicking the pill.

Screenshot 2024-08-05 at 10 55 23 AM

dzucconi:
@joeyAghion I'd imagine that the component just doesn't update due to anything affecting it client side, and rather just pulls out the initial routing value. Probably just a bug.

Unfortunately I'm really not one to ask about anything to do with the filters.

damassi:
@dzucconi - but you do know how all the parts work together 😉 

@joeyAghion - something like this should work? https://github.com/artsy/force/compare/damassi/example?expand=1

It listens for changes to the `keyword` query param and updates the value of the search input. (Not tested, but something you could play with)

dzucconi:
@damassi i don't really. i try to retain no knowledge of filters in my brain. anyone on onyx would have more of an idea of what's going on here

joeyAghion:
[This worked](https://github.com/artsy/force/pull/14274/files#diff-7c4c9460cadc034a411d786d5dfe63a02b33e80156c875b7dbb7180a2b1c4b07R46). Thanks both for the input!

I think this is finally ready for review. As discussed, it's a no-op in production or any other environment, until some keywords are manually selected for a fair (as I have for 1 fair in staging).

src/Components/ArtworkFilter/ArtworkFilters/KeywordFilter.tsx
```diff
@@ -43,9 +43,7 @@ export const KeywordFilter: React.FC = () => {
   }, [])
 
   useEffect(() => {
-    if (keyword === undefined) {
-      setValue("")
-    }
+    setValue(keyword ?? "")
```
damassi, line 46: 👍

damassi (approved):
Very straightforward 👌

PR #14277, opened 2024-08-02
feat: expand auction artworks filter options

mc-jones:
The type of this PR is: Feat

### Description

This PR was inspired by this hackathon 14 project [suggestion](https://www.notion.so/artsy/Add-shoppable-curated-collections-like-gems-on-paper-or-small-and-special-and-better-filter-op-710f2df614484eae90fc57fecabcef9c?pvs=4) to include more filter options for auction artworks, which has significantly less than our other artwork surfaces. After spiking on this work, it seems like we can extend the auction artwork filter options with pre-existing filters (size, materials, time period and colors) with very few code changes. 

**In fact, there are so few code changes required, I'm wondering if we have limited the filter options intentionally? Any insight on this would be appreciated. If not, this seems like an easy win.**

### Current
| auctions | fairs |
| Screenshot 2024-08-02 at 2 28 43 PM |  |

### PR Demo

https://github.com/user-attachments/assets/2205c6a2-998b-4e17-aab1-78138945d83a


dzucconi (approved):
No insight, but seems like an obvious win to me!

joeyAghion (approved):

PR #14278, opened 2024-08-05
chore(cache): Update TTLs / docs

damassi:

### Description

In an effort to improve our cache hit rate, this bumps our server-side TTLs for a number of routes via a new `Apps/serverCacheTTLs.tsx` index file. Most of these choices I believe are uncontroversial but let me know if anything should be tweaked, or if I missed a route that should be bumped up more appropriately!

### DIFF
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -5,10 +5,20 @@ There are a few different layers of caching on Artsy.net, namely:
 - Caching at the Relay (GraphQL) level, which happens as users navigate from page to page on the client. This is typically configured via the route.
 - Caching at the server level, via Redis, while logged out.
 
-> ⚠️ If for whatever reason we need to disable the cache, unset the environment variables via hokusai: 
+> ⚠️ If for whatever reason we need to disable the cache, unset the environment variables via hokusai:
+>
 > ```
 > hokusai production env unset ENABLE_GRAPHQL_PROXY ENABLE_GRAPHQL_CACHE
 > ```
+
+#### Important Cache-related ENV vars:
+
+- `ENABLE_GRAPHQL_PROXY`: Enables the proxy in support of caching
+- `ENABLE_GRAPHQL_CACHE`: Enables the GraphQL cache
+- `GRAPHQL_CACHE_SIZE`: Total number of entries before eviction (on the client; server-cache sizes are memory dependent)
+- `GRAPHQL_CACHE_TTL`: Default expiration TTL (in milliseconds)
+
+Non-default route cache times are set in [`Apps/serverCacheTTLs.tsx`](src/Apps/serverCacheTTLs.tsx).
 
 ### General Cache Behavior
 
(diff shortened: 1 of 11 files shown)


src/Apps/serverCacheTTLs.tsx
```diff
@@ -0,0 +1,21 @@
...
+  artist: HOURS_24,
+  artists: HOURS_24,
+  artistSeries: HOURS_24,
+  auction: NO_CACHE,
+  categories: HOURS_24,
+  collect: HOURS_24,
+  collection: HOURS_24,
+  collections: HOURS_24,
+  gene: HOURS_24,
+  partners: HOURS_24,
```
starsirius, line 19: Some thoughts:

- Partners pay attentions to their listing pages (e.g. partners, partner) and I assume we'd get questions from partners if updates take max 1 day to be reflected on the site. _I think_ it's a relatively lower traffic route, so maybe it's safer to use a shorter/more reasonable TTL. Also, it's worth giving the galleries team a heads-up about the caching behavior so they can be prepared [when getting questions](https://artsy.slack.com/archives/C03N12SR0RK/p1722867890398409?thread_ts=1722861744.869819&cid=C03N12SR0RK).
- Not specific to this route, but for timely information (such as the urgency signals that Emerald is working on), updates would take max 1 day to be reflected. Not sure how significant the impact is to conversion, but I'll surface to our team to at least be aware of it.

docs/caching.md
```diff
@@ -5,11 +5,21 @@ There are a few different layers of caching on Artsy.net, namely:
...
 > ```
 > hokusai production env unset ENABLE_GRAPHQL_PROXY ENABLE_GRAPHQL_CACHE
 > ```
 
+#### Important Cache-related ENV vars:
+
+- `ENABLE_GRAPHQL_PROXY`: Enables the proxy in support of caching
+- `ENABLE_GRAPHQL_CACHE`: Enables the GraphQL cache
+- `GRAPHQL_CACHE_SIZE`: Total number of entries before eviction
+- `GRAPHQL_CACHE_SIZE`: Default expiration TTL
```
starsirius, line 19: It's not immediately clear to me about the unit, so maybe we can say it's in milliseconds.

src/Apps/serverCacheTTLs.tsx
damassi: Once the numbers stabilize we'll make a wider announcement! PDDE Core has already been informed and we as engineers should keep an eye on things. We're still getting a very low % cache hit rate.

damassi: But yes, that's a good point regarding partners being lower traffic, higher visibility (and not necessarily from collectors). I'll remove this entry so as to reduce the potential for confusion 👍

joeyAghion (approved):

sweir27:
Sorry for the late review – this looks reasonable to me and thank you for the clear config file! It's great to see it all in one place – makes it easier to reason about.

PR #14283, opened 2024-08-05
fix: Navigate to My Collection artwork page on the Sell flow "Save & Exit"

olerichter00:
Resolves https://www.notion.so/artsy/When-submitting-a-MyC-artwork-should-Save-Exit-navigate-back-to-the-artwork-Currently-it-opens--ebdf272100c94666ba75b1c943097327?pvs=4

## Description

Navigate to My Collection artwork page on the "Save & Exit" when associated My Collection artwork is present.

Screenshot 2024-08-05 at 12 58 45

### DIFF
--- a/src/Apps/Sell/Components/SubmissionHeader.tsx
+++ b/src/Apps/Sell/Components/SubmissionHeader.tsx
@@ -84,7 +84,11 @@ export const SubmissionHeader: React.FC = () => {
       // Save the submission and current step to local storage
       storePreviousSubmission(submission?.externalId as string, step)
 
-      routerPush("/sell")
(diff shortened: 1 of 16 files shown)


olerichter00:
@MounirDhahri do we need the same adjustment in the app?

dariakoko (approved):
⭐

dariakoko:
I don't think we need the same adjustment in the app, navigation there is built differently

MounirDhahri:
@dariakoko is right, we simply dismiss the screen on app - so it should be fine. We don't need to "Save" the submission to global store though in that case

PR #14304, opened 2024-08-07
feat: offer quick filters for featured keywords on partners

joeyAghion:
This PR adds optional featured keywords to the view of a partner's artworks (like https://github.com/artsy/force/pull/14274 did for fairs). Looks like:

Screenshot 2024-08-06 at 3 05 00 PM

I ran `yarn sync-schema` but am surprised by the unrelated changes that caused. I'm also curious about why there were so many relay changes but those at least seem related.


damassi (approved):
Looks good 👍 Note that `sync-schema` isn't what caused the file changes, it was the addition of `featuredKeywords` field to the `Works_partner` fragment, which is spread in a few different places. Relay's compiler then updated each generated file based on that.

PR #14317, opened 2024-08-08
fix(auth): Signin / Signup no longer 500 -> 4x

damassi:

### Description

Looking through our datadog logs noticed that a lot of 500ish errors come from `sign_in` / `sign_up`. Digging in, noticed that we're sending 500s for 400-ish errors. This fixes that. 

Let me know if we should have more fine-grained status codes here. 

Screenshot 2024-08-08 at 12 56 44 PM

### DIFF
--- a/src/Server/passport/lib/app/lifecycle.ts
+++ b/src/Server/passport/lib/app/lifecycle.ts
@@ -31,7 +31,35 @@ module.exports.onLocalLogin = function (
   passport.authenticate("local-with-otp")(req, res, function (err: any) {
     if (req.xhr) {
       if (err) {
-        return res.status(500).send({ success: false, error: err.message })
+        switch (true) {
+          case err.message?.includes("invalid email or password"): {
+            return res.status(403).send({ success: false, error: err.message })
+          }
+          case err.message?.includes(
+            "missing two-factor authentication code"
+          ): {
+            return res.status(403).send({ success: false, error: err.message })
+          }
+          case err.message?.includes(
+            "invalid two-factor authentication code"
+          ): {
+            return res.status(403).send({ success: false, error: err.message })
+          }
+          case err.message?.includes(
+            "account locked, try again in a few minutes"
+          ): {
+            return res.status(403).send({ success: false, error: err.message })
+          }
+          case err.message?.includes("Unexpected token"): {
+            return res.status(500).send({
+              success: false,
+              error: "An error occurred. Please try again.",
+            })
+          }
+          default: {
+            return res.status(400).send({ success: false, error: err.message })
+          }
+        }
       } else {
         return next()
       }
@@ -105,7 +133,7 @@ module.exports.onLocalSignup = function (
         } else if (err.message) {
           msg = err.message
         }
-        return res.status(500).send({ success: false, error: msg })
+        return res.status(400).send({ success: false, error: msg })
       } else if (err) {
         return next(new Error(err))
       } else {
--- a/src/Server/passport/lib/app/lifecycle.ts
+++ b/src/Server/passport/lib/app/lifecycle.ts
@@ -33,17 +33,17 @@ module.exports.onLocalLogin = function (
       if (err) {
         switch (true) {
           case err.message?.includes("invalid email or password"): {
-            return res.status(403).send({ success: false, error: err.message })
+            return res.status(401).send({ success: false, error: err.message })
           }
           case err.message?.includes(
             "missing two-factor authentication code"
           ): {
-            return res.status(403).send({ success: false, error: err.message })
+            return res.status(401).send({ success: false, error: err.message })
           }
           case err.message?.includes(
             "invalid two-factor authentication code"
           ): {
-            return res.status(403).send({ success: false, error: err.message })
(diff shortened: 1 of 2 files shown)


src/Server/passport/lib/app/lifecycle.ts
```diff
@@ -31,7 +31,38 @@ module.exports.onLocalLogin = function (
...
+            return res.status(403).send({ success: false, error: err.message })
+          }
+          case err.message?.includes("User Already Exists"): {
+            return res.status(409).send({ success: false, error: err.message })
+          }
+          case err.message?.includes("Unexpected token"): {
+            return res.status(500).send({
+              success: false,
+              error: "An error occurred. Please try again.",
+            })
```
damassi, line 60: Pulled all the relevant ones [from here](https://github.com/artsy/force/blob/damassi/fix/challenge-500/src/Components/AuthDialog/Utils/formatErrorMessage.ts)

src/Server/passport/lib/app/lifecycle.ts
```diff
@@ -31,7 +31,38 @@ module.exports.onLocalLogin = function (
...
+          }
+          default: {
+            return res.status(400).send({ success: false, error: err.message })
```
damassi, line 63: Default falls back to a 400 rather than 500

src/Server/passport/lib/app/lifecycle.ts
```diff
@@ -105,7 +136,7 @@ module.exports.onLocalSignup = function (
         } else if (err.message) {
           msg = err.message
         }
-        return res.status(500).send({ success: false, error: msg })
+        return res.status(400).send({ success: false, error: msg })
```
damassi, line 139: Ditto

src/Server/passport/lib/app/lifecycle.ts
narikazu: question (non-blocking): do you happen to know what caused `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` and `Unexpected token '<', "<html> <"... is not valid JSON`?

narikazu (dismissed):
LGTM
LGTM

src/Server/passport/lib/app/lifecycle.ts
```diff
@@ -31,7 +31,35 @@ module.exports.onLocalLogin = function (
   passport.authenticate("local-with-otp")(req, res, function (err: any) {
     if (req.xhr) {
       if (err) {
-        return res.status(500).send({ success: false, error: err.message })
+        switch (true) {
+          case err.message?.includes("invalid email or password"): {
+            return res.status(403).send({ success: false, error: err.message })
```
ovasdi, line 36: suggestion: After reviewing the [status codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses), I believe sending a `401` here would be more appropriate seeing as the client is (likely) unauthenticated at this point.

Same applies to L41 and L46 when responding to _missing_ or _invalid_ 2fa code.

src/Server/passport/lib/app/lifecycle.ts
damassi: I _think_ it relates to the error page from api.artsy.net? At times error page html can come down the wire instead of JSON responses.

ovasdi (dismissed):

src/Server/passport/lib/app/lifecycle.ts
damassi: Ah nice, updated 👍 

Also gonna add some quick test coverage.

src/Server/passport/test/app/lifecycle.jest.js
```diff
@@ -12,102 +11,154 @@ jest.mock("Server/passport/lib/options", () => ({
   APP_URL: "https://www.artsy.net",
   ARTSY_URL: "https://api.artsy.net",
 }))
+
 jest.mock("superagent")
 jest.mock("passport", () => {
-  const sinon = require("sinon")
   return {
-    authenticate: sinon.stub().returns((req, res, next) => next()),
+    authenticate: jest.fn().mockReturnValue((req, res, next) => next()),
```
damassi, line 18: sinon is old school, so replaced it with vanilla jest

joeyAghion:
I'm confused about this implementation, since the bug is actually upstream at the API. I.e., I think Gravity should probably be returning appropriate 4xx error codes in response to request errors like invalid credentials.

damassi:
@joeyAghion - lets chat [in slack](https://artsy.slack.com/archives/CA8SANW3W/p1723221128445689?thread_ts=1723138429.292979&cid=CA8SANW3W). This PR doesn't aim to refactor our implementation, but rather to fix the default 500 (bug) that's been in place for a long time and clogging our error rate.

If gravity is already returning correct status codes for these messages tho, it should be easy to work those in, but that should be in a follow-up PR.

joeyAghion (approved):
As discussed, I think Gravity _should_ still choose better status codes, and ideally clients like Force should switch on those (or the short error codes that accompany them). This works though.

damassi:
I will investigate and follow up 👍

PR #14321, opened 2024-08-09
feat(saves): Rebuilds saves; adds price filtering

dzucconi:
Closes [DIA-647](https://artsyproduct.atlassian.net/browse/DIA-647)

This adds price filtering to the saves page.

The saves collection isn't backed by filters but the frontend was using filters. Was _much_ simpler to just rip it out and redo it.

[DIA-647]: https://artsyproduct.atlassian.net/browse/DIA-647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Utils/Hooks/useDebounce.ts
```diff
@@ -22,6 +25,9 @@ interface UseDebouncedValue<T> {
   settings?: any
 }
 
+/**
+ * @deprecated Use `import { useDebounce } from "use-debounce"` instead
+ */
```
dzucconi, line 30: None of these debounces worked correctly; they would just delay the fetch and stack them. I think filters has a debounce internally that works that prevents this from being a problem.

package.json
```diff
@@ -202,6 +202,7 @@
     "unleash-client": "^3.12.0",
     "unstated": "^2.1.1",
     "use-cursor": "^1.2.3",
+    "use-debounce": "^10.0.2",
```
dzucconi, line 205: As reluctant as I am to add a dependency for something simple... [this actually isn't simple to do correctly](https://github.com/xnimorz/use-debounce/blob/master/src/useDebouncedCallback.ts) in React.

egdbear (dismissed):

src/Components/ClientSuspense.tsx
```diff
@@ -0,0 +1,24 @@
+import { useDidMount } from "@artsy/palette"
+import { FC, ReactNode, Suspense } from "react"
+
+interface ClientSuspenseProps {
+  fallback: NonNullable<ReactNode>
+  children: ReactNode
+}
+
+export const ClientSuspense: FC<ClientSuspenseProps> = ({
```
damassi, line 9: 👍 

So damon, what would be your guidance re `useClientQuery`? Lean into suspense? Or use whats best?

It still feels way nicer having a `loading` prop and not have to deal with vertical component boundaries. Curious about yr take

src/Components/PriceRange/Utils/parsePriceRange.ts
```diff
@@ -1,8 +1,57 @@
...
+  }
+
+  if (min === "*") {
+    return `$0–$${max}`
+  }
+
+  if (max === "*") {
+    return `$${min}+`
+  }
+  return `$${min}–$${max}`
```
damassi, line 56: Can we add some tests for these utils?

dzucconi: There's tests for the others. I just pulled this one out of a the filters context. Might as well.

src/Components/ClientSuspense.tsx
dzucconi: Yeah, I like it and find it pretty intuitive. We just can't use it with SSR given our current framework.

src/Components/PriceRange/Utils/parsePriceRange.ts
damassi: Yah seems easiest enough to just chatgpt it real quick 👍

src/Components/ClientSuspense.tsx
damassi: Cool 👍

PR #14331, opened 2024-08-09
feat: sales offer featured keywords for curated browsing (whether auctions or not)

joeyAghion:
This PR adds featured keyword options for sales, like https://github.com/artsy/force/pull/14274 did for fairs and https://github.com/artsy/force/pull/14304 did for partners. Sales were more complicated because there are 2 separate implementations (Auction and [non-auction] Sale), and because the organization of fragments made it difficult to pull `featuredKeywords` directly off of the `viewer`.

Please point out anything odd or non-idiomatic. I had to fiddle with type definitions (ahem `readonly string[] | null | undefined`) quite a bit to pass all the checks but I'm eager to learn a better way.

Co-authored by @ovasdi @damassi


src/Apps/Sale/Components/SaleArtworks.tsx
```diff
@@ -13,11 +13,13 @@ import { SaleArtworksFilter_viewer$data } from "__generated__/SaleArtworksFilter
 interface SaleArtworkFilterProps {
   relay: RelayRefetchProp
   viewer: SaleArtworksFilter_viewer$data
+  featuredKeywords?: readonly string[]
```
damassi, line 16: If we're ever defining a field in a relay query that we're then using in the UI (meaning, in some fashion its being passed through as props) then we try to avoid mirroring / duplicating the interface types, since there's already a typescript file that's been generated by relay, by virtue of us writing the query. This is all very confusing, but can explain via an example:

```tsx
// Avoid 
interface AppProps {
  sale: {
    name: string
  }
}

// Good

import { App_sale$data } from '__generated__/App_sale.graphql'

interface AppProps {
  sale: App_sale$data
}

const App: React.FC<AppProps> = ({ sale }) => {

}

export const AppFragmentContainer = createFragmentContainer(App, {
  sale: graphql`
    fragment App_sale on Sale {
      name 
    }
  `
})

So here, to avoid the strange `readonly` stuff, can just use the type that's generated and passed in directly:

```tsx
interface SaleArtworkFilterProps {
  ...
  featuredKeywords: SaleApp_sale$data['featuredKeywords']
}
```

Doing this also helps avoid having to do the `| undefined | null` dance.

All that said, in the end (since this is consumed way down in the ArtworkFilter, beyond a product area surface boundary, and is not a straightforward relay component) unless you want to provide a union type there it might be more straight forward to do things as you are. Just know that typically we try to avoid this and always lean on the generated typescript files.

damassi (approved):
Looks good 👍 Would say that typically we'd want to add a test at the `ArtworkFiltersQuick` level (since its a new feature) but i see there's no tests at all there 😄 so its all good. This is all relay driven, so it will break loudly should something change.

src/Apps/Sale/Components/SaleArtworks.tsx
joeyAghion: Exactly the kind of advice I need 🙏

PR #14332, opened 2024-08-12
fix: Hide back button on Sell flow title step to not allow editing artist

olerichter00:
Resolves https://www.notion.so/artsy/Web-We-should-not-allow-the-user-to-edit-artwork-s-Artist-and-title-13745a26be404bd5b9c0f5755dd31693?pvs=4

## Description

With this change the "Back" button on the Sell flow's title step is hidden. This prevents users from (accidentally) updating a submission's artist.

| Before | After |
| --- | --- |
|Screenshot 2024-08-12 at 10 44 14 |Screenshot 2024-08-12 at 10 44 29 |


src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkSWASubmissionStatus.tsx
```diff
@@ -227,9 +228,11 @@ const useGetButtonURL = (
 
   if (!submission) return null
 
-  const previousStep = submissionID === submission.externalID && step
+  // Do not link to the initial artist step.
+  const previousStep =
+    submissionID === submission.externalID && step !== INITIAL_STEP && step
```
olerichter00, line 233: To prevent linking to the Artist step in case it's the saved previous submission.
Screenshot 2024-08-12 at 11 40 19

nickskalkin (dismissed):

PR #14333, opened 2024-08-12
fix: Display previous submission on Sell page when in APPROVED state

olerichter00:

### Description

With this change, we also display the previous submission on the Sell page when it is in the `APPROVED` state. With this change, I consider `draft` and `approved` submissions as non-completed submissions (`PRE_SUBMITTED_STEPS`) that require additional information.

### DIFF
--- a/src/Apps/Sell/SellFlowContext.tsx
+++ b/src/Apps/Sell/SellFlowContext.tsx
@@ -27,6 +27,11 @@ const BASIC_FLOW_STATES: ConsignmentSubmissionStateAggregation[] = [
   "DRAFT",
   "SUBMITTED",
 ]
+// If a submission is in one of these states, additional information is required.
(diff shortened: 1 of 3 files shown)


src/Apps/Sell/SellFlowContext.tsx
```diff
@@ -27,6 +27,11 @@ const BASIC_FLOW_STATES: ConsignmentSubmissionStateAggregation[] = [
   "DRAFT",
   "SUBMITTED",
 ]
+// If a submission is in one of these states, additional information is required.
+export const PRE_SUBMITTED_STEPS: ConsignmentSubmissionStateAggregation[] = [
```
olerichter00, line 31: What do you think about the naming? I would be happy to hear any suggestions for a better or more precise name for these two steps.

nickskalkin (approved):

PR #14336, opened 2024-08-12
fix: ensure cached graphql responses are invalidated upon query changes

joeyAghion:

Currently, a graphql response can be cached by one release, and served by a later release. That's fine, and even intentional, as long as the query hasn't changed. If the query has changed, cached responses to the old query should be invalidated. This PR adds a `digest` property to the cache key that's computed over the query itself. When a query changes, the digest will change and requests will be cache misses until the updated query's responses are cached.

Those old responses will be evicted by redis in the usual course of things. This also allows review apps to employ slightly different queries with the same identifier. Their responses will be cached in isolation.

Example verbose logging output:
```
[1] [graphqlProxyMiddleware # writeCache] 
[1]  [cacheKey]: {"queryId":"SearchBarInputSuggestQuery","digest":"27031aefbeee826cc2590e069bfd2d77af967a81","variables":{"term":"","hasTerm":false,"entities":[]}} 

...

[1] [graphqlProxyMiddleware # get] Cache hit: 
[1]  [cacheKey]: {"queryId":"SearchBarInputSuggestQuery","digest":"27031aefbeee826cc2590e069bfd2d77af967a81","variables":{"term":"","hasTerm":false,"entities":[]}} 
```

### DIFF
--- a/src/Server/middleware/graphqlProxyMiddleware.ts
+++ b/src/Server/middleware/graphqlProxyMiddleware.ts
@@ -15,6 +15,7 @@ import {
 } from "System/Relay/middleware/cacheHeaderMiddleware"
 import { findRoutesByPath } from "System/Router/Utils/routeUtils"
 import { getENV } from "Utils/getENV"
+import { createHash } from "crypto"
 
 const isDevelopment = getENV("NODE_ENV") === "development"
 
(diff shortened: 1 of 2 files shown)


src/Server/middleware/graphqlProxyMiddleware.ts
```diff
@@ -195,15 +200,18 @@ export const writeCache = async (
...
 }
 
 const getCacheKey = (props: GetCacheKeyProps) => {
   const fallbackCacheKey = "" // Falsy value for redis
 
   try {
+    const digest = createHash("sha1").update(props.query).digest("hex")
     const cacheKey = JSON.stringify({
       queryId: props.queryId,
+      digest,
```
damassi, line 214: Very straight forward. Thank you @joeyAghion 🙏

damassi (approved):
Very straight forward. Thank you @joeyAghion 🙏

PR #14338, opened 2024-08-13
fix(EMI-2019): check if offer available

leamotta:

This PR solves [EMI-2019]

### Description

For displaying the expired offer banner, instead of not getting a partner offer in the backend response and assuming that the partner offer is expired, we can return the partner offer and show the banner if it's not available. This would cover the case when a user is logged in with a different user.
Similar to #14058 but with a feature flag to coordinate everything, we would need to use it in Gravity and Eigen as well

TODO:
[] Gravity PR to return unavailable partner offers
[] PR in Eigen for similar logic

[EMI-2019]: https://artsyproduct.atlassian.net/browse/EMI-2019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun (approved):
Looks good! 🌟

src/Apps/Artwork/Components/ArtworkPageBanner.tsx
```diff
@@ -49,7 +53,15 @@ export const ArtworkPageBanner: FC<ArtworkPageBannerProps> = props => {
       return <ArtworkUnavailableBanner />
     }
 
-    if (
+    if (allowExpiredPartnerOffers) {
```
araujobarret, line 56: [ceiling-cat] I think it's better to read if we avoid the nesting(follow-up):
```jsx
const hasPartnerOffer = partnerOffer?.internalID == expectedPartnerOfferID
  
if (allowExpiredPartnerOffers && hasPartnerOffer && !partnerOffer.isAvailable) {
  return <ExpiredOfferBanner />
}

if (!allowExpiredPartnerOffers && me && !hasPartnerOffer) {
  return <ExpiredOfferBanner />
}

return null
```

PR #14339, opened 2024-08-13
fix: allow all accounts to add cards (reverts #14046)

joeyAghion:
Reverts https://github.com/artsy/force/pull/14046, now that other safeguards have proven effective. This also fixes a typo in the original copy.

Follow-up to: https://artsyproduct.atlassian.net/browse/PHIRE-1013


brainbicycle (approved):
👍

PR #14341, opened 2024-08-14
feat: adds relative-ci agent

dzucconi:
Closes [DIA-793](https://artsyproduct.atlassian.net/browse/DIA-793)

We need to generate size diffs for bundle reports on pull requests. Testing out [relative-ci.com](https://relative-ci.com/) — free for open source.

[DIA-793]: https://artsyproduct.atlassian.net/browse/DIA-793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,6 +216,17 @@ jobs:
       - store_artifacts:
           path: cypress/screenshots
 
+  bundle-stats:
+    docker:
+      - image: 585031190124.dkr.ecr.us-east-1.amazonaws.com/force:$CIRCLE_SHA1-builder
+        aws_auth:
+          aws_access_key_id: $AWS_ACCESS_KEY_ID
+          aws_secret_access_key: $AWS_SECRET_ACCESS_KEY
+    working_directory: /app
+    steps:
+      - run: yarn bundle-report:client --json webpack-stats.json
+      - run: npx relative-ci-agent
+
 not_main_or_staging_or_release: &not_main_or_staging_or_release
   filters:
     branches:
@@ -330,6 +341,12 @@ workflows:
           requires:
             - electron-runner-image-push
 
+      - bundle-stats:
+          <<: *not_staging_or_release
+          context: hokusai
+          requires:
+            - builder-image-push
+
       # Staging
       - artsy-remote-docker/buildkit-build:
           <<: *only_main
@@ -349,6 +366,7 @@ workflows:
             - type-check
             - relay-check
             - acceptance
+            - bundle-stats
             - production-image-build
 
       - hokusai/deploy-staging:
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,6 +106,7 @@ COPY data ./data
 COPY patches ./patches
 COPY src ./src
 COPY webpack ./webpack
+COPY .git ./.git
 COPY .env.oss \
   .env.test \
   .eslintrc.js \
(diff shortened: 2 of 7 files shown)


egdbear (dismissed):
🚀

araujobarret (commented):
Can't we generate the artifacts to be stored within the job?
Might be nicer to take than the console

dzucconi:
@araujobarret I don't really follow what you're asking?

araujobarret:
image

Does this step generate something more visual to take a look(Artifacts in this case)?

dzucconi:
Still not sure what you're asking. I'm using an external service. It will indeed generate something visual.

araujobarret:
That was my question, since it's not working yet and there's no screenshot in the PR I wanted to see where something more visual would come in(didn't look at the tool so much to say, maybe I should)

dzucconi:
Yeah, lemme get this working first 😄

damassi:
Ahh nice output, and not in house 👌

araujobarret:
> Yeah, lemme get this working first 😄

a'ight 😂 

is there a way to prevent generated files from being tagged as duplicated?

araujobarret (dismissed):

damassi (approved):

.circleci/config.yml
```diff
@@ -330,6 +341,12 @@ workflows:
           requires:
             - electron-runner-image-push
 
+      - bundle-stats:
+          <<: *not_staging_or_release
+          context: hokusai
```
joeyAghion, line 346: What permissions are actually necessary for this step? This context has access to release to (and read from) our cluster which I imagine isn't necessary. If access to S3 is necessary, there's probably a more restrictive context we can lean on.

dzucconi: Right, this just needs to run the build to emit the stats then upload them using the agent + the environment variable that's set in Circle. I, however, don't understand the config or how to use a more restrictive context, in this case.

Dockerfile
```diff
@@ -118,6 +119,7 @@ COPY .env.oss \
   jest.config.js \
   package.json \
   relay.config.js \
+  relativeci.config.js \
```
starsirius, line 122: I thought we need to update [`.dockerignore`](https://github.com/artsy/force/blob/37dc7545141ff936a9ffbf85190fd04c3a8a5e2d/.dockerignore) accordingly, given we explicitly allow individual files. But looks like we actually don't! TIL!

.circleci/config.yml
joeyAghion: > upload them using the agent + the environment variable that's set in Circle

Which environment variable?

dzucconi: `RELATIVE_CI_KEY`

dzucconi: So, what context should I use though? I apparently need the AWS creds just to run the build.

joeyAghion: If you mean the prerequisite builder-image-push step by "the build," I think that already uses the `hokusai` context. Does this step also need AWS creds? If not, I'd create a new `relativeci` org context, add the key there, and specify it here.

dzucconi: Just did that. Apparently not:



In the interest of moving on I'm going to proceed with the Hokusai context for now.

PR #14351, opened 2024-08-16
refactor: optimize script evaluation

dzucconi:
Closes [DIA-788](https://artsyproduct.atlassian.net/browse/DIA-788)

So by opening the performance tab, enabling a CPU slow down and 'Start polling and reload page' — we can see exactly what pieces of code are being loaded for a given route. Previously the bulk of this load was artwork filters. By pulling apart the imports so that the functions the routes are using are isolated, we can remove all of that code from being loaded.

This is a screen capture taken after removing all that code:

https://github.com/user-attachments/assets/04567802-0ea2-431b-b071-c4bcf91ec37c

You'll see in this screen capture, towards the end, there's another big network of code loaded for onboarding, which is also obviously not needed most of the time. I'm going to make a new ticket for that and split that off as well.


src/Apps/ArtQuiz/artQuizRoutes.tsx
```diff
@@ -13,21 +13,28 @@ const artQuizServerSideRedirect = ({ res }: { res: ArtsyResponse }) => {
...
+)
 
-const ArtQuizResults = loadable(() => import("./Routes/ArtQuizResults"), {
-  resolveComponent: component => component.ArtQuizResultsFragmentContainer,
-})
+const ArtQuizResults = loadable(
+  () =>
+    import(/* webpackChunkName: "artQuizBundle" */ "./Routes/ArtQuizResults"),
+  { resolveComponent: component => component.ArtQuizResultsFragmentContainer }
+)
```
dzucconi, line 37: These routes all had separate bundles.

src/Apps/PriceDatabase/priceDatabaseRoutes.tsx
```diff
@@ -1,18 +1,15 @@
 import loadable from "@loadable/component"
 import { RouteProps } from "System/Router/Route"
-import { PriceDatabase } from "./PriceDatabase"
```
dzucconi, line 3: Wasn't deferred

araujobarret (approved):
Took a look and very nice job, so good to see moving things from here to there optimized so much!

So if I get it right the graph this route takes around 3s and before it was around 10s?

dzucconi:
No that is just the graph of the optimized load and the timings aren't so relevant. It's just being used to see what actually winds up in the load.

I just merged the Relative CI PR so we should see the bundles change somewhat, though I have no real way of knowing the difference in eval from that.

But we can compare against the existing analysis once this is deployed.

src/Apps/Artist/Routes/AuctionResults/Components/AuctionFilterMobileActionSheet.tsx
```diff
@@ -1,11 +1,9 @@
 import { FC, useEffect, useRef } from "react"
 import { isEqual, omit } from "lodash"
 import styled from "styled-components"
-import {
-  useAuctionResultsFilterContext,
-  initialAuctionResultsFilterState,
-} from "../AuctionResultsFilterContext"
+import { useAuctionResultsFilterContext } from "Apps/Artist/Routes/AuctionResults/AuctionResultsFilterContext"
 import { Box, Button, Clickable, Flex, ModalBase, Text } from "@artsy/palette"
+import { initialAuctionResultsFilterState } from "Apps/Artist/Routes/AuctionResults/initialAuctionResultsFilterState"
```
egdbear, line 6: 👏

egdbear (approved):
⚡

src/Apps/Auction/Components/getArtworkFilterInputArgs.ts
```diff
@@ -0,0 +1,13 @@
...
+  if (user) {
+    aggregations.push("FOLLOWED_ARTISTS")
+  }
+
+  // Shared with auctionRoutes
+  return {
+    aggregations,
+    first: 39,
+  }
+}
```
damassi, line 13: 👍

damassi (commented):
👍

PR #14382, opened 2024-08-23
feat(EMI-1961): implement auction signals

rquartararo:

This PR solves [EMI-1961]

### Description

View [Figma](https://www.figma.com/design/22RFPSkUu97g0ska3azFL2/Social-proof-%26-urgency-signals?node-id=1799-9320&m=dev) for signal display logic

Adds
- Lot watcher counts
- Updated bid counts
- `Bidding closed` signal
- `Bidding live now` signal
- Bidding closing time timer
- Register date signal

Removes 
- Grid item badge
- Lot close info
- Auction name from lot detail cards

### Lot watcher counts + Bidding closing time timer
Screenshot 2024-08-23 at 7 12 36 AM

### `Bidding closed` signal
Screenshot 2024-08-23 at 7 12 22 AM

### `Bidding live now` signal
Screenshot 2024-08-23 at 7 11 39 AM

### Register date signal
Screenshot 2024-08-23 at 7 12 49 AM

### All together on one rail
Screenshot 2024-08-22 at 10 59 06 PM

### Lot detail cards
Screenshot 2024-08-22 at 11 27 52 PM

### Aucton detail page
Screenshot 2024-08-22 at 11 28 45 PM

### Search

[EMI-1961]: https://artsyproduct.atlassian.net/browse/EMI-1961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Artwork/Details.tsx
+++ b/src/Components/Artwork/Details.tsx
@@ -444,8 +444,8 @@ export const Details: React.FC<DetailsProps> = ({
         </Flex>
       )}
 
-      <Flex justifyContent="space-between">
-        <Flex flexDirection="column" maxWidth="85%" alignItems="flex-start">
+      <Flex justifyContent="space-between" alignItems="flex-start">
+        <Flex flexDirection="column" maxWidth="85%">
           {showActivePartnerOfferLine && <CollectorSignalLine {...rest} />}
           {!hideArtistName && (
             <ArtistLine showSaveButton={showSaveButton} {...rest} />
--- a/src/Apps/Settings/Routes/Auctions/Components/__tests__/UserBidHistory.jest.tsx
+++ b/src/Apps/Settings/Routes/Auctions/Components/__tests__/UserBidHistory.jest.tsx
@@ -24,12 +24,12 @@ describe("UserBidHistory", () => {
         lotLabel: "1",
       }),
       Artwork: () => ({
-        sale_message: "$3,000",
+        sale_message: "Example Sale",
       }),
     })
 
     expect(screen.getByText("Lot 1")).toBeInTheDocument()
-    expect(screen.getByText("$3,000 (42 bids)")).toBeInTheDocument()
+    expect(screen.getByText("Example Sale")).toBeInTheDocument()
     expect(screen.getByText("Bid")).toBeInTheDocument()
   })
 


src/Components/Artwork/SaveButton/SaveButton.tsx
```diff
@@ -43,31 +54,39 @@ export const SaveButtonBase: React.FC<SaveButtonBaseProps> = ({
...
+            height={BTN_HEIGHT}
+          />
+        ) : (
+          <HeartStrokeIcon
+            title={title}
+            fill="black100"
+            width={BTN_WIDTH}
+            height={BTN_HEIGHT}
+          />
+        )}
```
erikdstock, line 87: non-blocking: do these need to be repeated or will clickable automatically wrap whichever icon is rendered?

src/Components/Artwork/SaveButton/SaveButton.tsx
```diff
@@ -17,18 +18,28 @@ export interface SaveButtonProps {
...
   const [isHovered, setIsHovered] = useState(false)
   const title = isSaved ? "Unsave" : "Save"
+  const { lotWatcherCount, lotClosesAt, liveBiddingStarted } =
+    artwork.collectorSignals?.auction || {}
+
+  const shouldDisplayLotCount =
+    lotWatcherCount != null &&
+    lotWatcherCount > 0 &&
+    !liveBiddingStarted &&
+    (!lotClosesAt || new Date(lotClosesAt) >= new Date())
```
erikdstock, line 42: Again here i think the count presence+non-zero could be simplified to `!!lotWatcherCount`

Bigger question I have here: Does this count update when the users saves/unsaves? As a user I would expect to see that change after i click the heart, but as a developer I would not expect to see that change reflected in the mutation result because it is processed in a job outside the mutation request.

3 ways I could see this working:
1. I could be wrong and it does just work.
2. We update the artwork store surgically when the save/unsave action happens
3. We just maintain some local state in this component including
  - the count on load (in this case i would avoid the null value altogether and set it like `const lotWatcherCount = artwork.collectorSignals.auction.lotWatcherCount || 0`)
  - whether the user had the artwork saved on load (so whether the count includes the user)
  - a state var for when the user saves/unsaves to +/-1

erikdstock (commented):
:+1: A few comments, mostly non-blocking. One question about the behavior of the lot watcher count on save which might not be blocking but would irritate me as a user.

src/Components/Artwork/SaveButton/SaveButton.tsx
```diff
@@ -17,18 +18,27 @@ export interface SaveButtonProps {
...
 }) => {
   const [isHovered, setIsHovered] = useState(false)
   const title = isSaved ? "Unsave" : "Save"
+  const { lotWatcherCount, lotClosesAt, liveBiddingStarted } =
+    artwork.collectorSignals?.auction || {}
+
+  const shouldDisplayLotCount =
+    !!lotWatcherCount &&
+    !liveBiddingStarted &&
+    (!lotClosesAt || new Date(lotClosesAt) >= new Date())
```
MrSltun, line 41: **Suggestion**: should we hide this behind a feature flag?

src/Components/Artwork/Details.tsx
```diff
@@ -166,7 +185,6 @@ const CollectorSignalLine: React.FC<DetailsProps> = ({
       color="blue100"
       backgroundColor="blue10"
       px={0.5}
-      marginBottom={0.5}
```
MrSltun, line 169: **Question**: was this change intentional? 👀

src/Components/Artwork/Details.tsx
```diff
@@ -175,14 +193,66 @@ const CollectorSignalLine: React.FC<DetailsProps> = ({
   )
 }
 
-const EmptyLine: React.FC = () => {
+const BidTimerLine: React.FC<DetailsProps> = ({
```
MrSltun, line 196: **Suggestion**: should we extract this one (and the others) into different components so it's easily readable (and have smaller files)?

MrSltun (commented):
Looks good! ✨  I just left a few comments  :)

PR #14383, opened 2024-08-23
chore: second part of defaulting to save address.

oxaudo:
This is a second case I found that should fix https://artsyproduct.atlassian.net/browse/EMI-1939

It's a bit roundabout way to fix it but I believe it's a proper condition to call this `useEffect` regardless of this bug even. But looks like it works here because when the last address deleted on the page the length of addresses change and the previous case that I fixed (default to save address when address is present on the order but there is no addresses in context - default to save address).

It's all a bit confusing but seems to work. Also not sure how to test any of this stuff cause the setup is pretty complicated. But maybe can tested with integration run after.

### DIFF
--- a/src/Apps/Order/Routes/Shipping2/Components/SavedAddresses2.tsx
+++ b/src/Apps/Order/Routes/Shipping2/Components/SavedAddresses2.tsx
@@ -77,7 +77,11 @@ export const SavedAddresses2: FC<SavedAddressesProps> = props => {
     }
     automaticallySelectBestAddress()
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [props.active])
+  }, [
+    props.active,
+    addressList.length,
+    shippingContext.state.selectedSavedAddressID,
+  ])
 
   /*
    * Select an address radio button and pass the address to the parent.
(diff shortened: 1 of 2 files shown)


erikdstock:
I still need to confirm this works but I have a fix for [EMI-1944] which I think might achieve the same result: https://github.com/artsy/force/compare/erik.emi1944-clear-address-on-delete

What do you think? This way the value would be set when the form resets.

oxaudo:
@erikdstock - I think your suggestion is legit as well but probably can be done in parallel.  Both of those changes are legit I think. The effect in `SavedAddresses` here depends on `savedAddresses.length` and `context` stuff and should be rerun if any of those change. I beieve we added those at some point with some bug fix but forgot to update run dependency.

erikdstock:
> @erikdstock - I think your suggestion is legit as well but probably can be done in parallel. Both of those changes are legit I think. The effect in `SavedAddresses` here depends on `savedAddresses.length` and `context` stuff and should be rerun if any of those change. I beieve we added those at some point with some bug fix but forgot to update run dependency.

My thinking is this:
- The change I made uses the result of `getInitialValues()` in FulfillmentDetails, outside of SavedAddresses. That function does return the default `saveAddress: true`
- If the length of savedAddresses becomes zero, responsibility for the address form is returned to the no-saved-address shipping form (`shippingFormMode === 'new_address'`)
- This responsibility makes sense higher up in the component tree, in the FulfillmentDetails component & at the same point that we change the mode.

Since I just did that at the end of the day yesterday I'd like to test whether it works first, but maybe there is some bug as you mentioned that this change is still worth it. Maybe we could talk through it?

oxaudo:
Happy to merge your change first and see what's still off if anything. Happy to chet as well but probably not now as my connection here is so crappy and electricity is on/off today.

erikdstock (approved):
I think you are right that this is a potential bug. If we only trigger this when it becomes active we might miss the case where a user deletes their selected address but still has another. I verified this in our code today - good catch!

PR #14385, opened 2024-08-23
fix(artwork-filter): Hide dropdown on sort click

damassi:

### Description

Fixes artwork filter issue where ever click to the sort requires two clicks to close the filter. Now close automatically on select:

https://github.com/user-attachments/assets/eaa49308-d4ff-468b-bba0-3cf867b4a65c


joeyAghion (approved):
Thank you!

egdbear (approved):

PR #14397, opened 2024-08-26
feat: update "last update prompt at" timestamp when user sees post-inquiry prompts

iskounen:
This PR updates the post-inquiry prompts to fire a mutation that will update the "last update prompt at" timestamp when the prompt is displayed. Updating this timestamp will prevent the user from being prompted to update their profile for 30 days if they don't complete the prompt.

### DIFF
--- a/src/Components/Inquiry/Views/InquiryArtistsInCollection.tsx
+++ b/src/Components/Inquiry/Views/InquiryArtistsInCollection.tsx
@@ -9,13 +9,19 @@ import {
 import { Box, Stack, Text } from "@artsy/palette"
 import { CollectorProfileArtistsAdd } from "Components/CollectorProfile/CollectorProfileArtistsAdd"
 import { useInquiryContext } from "Components/Inquiry/Hooks/useInquiryContext"
+import { useUpdateMyUserProfile } from "Components/Inquiry/Hooks/useUpdateMyUserProfile"
 import { FC } from "react"
+import { Environment } from "react-relay"
 import { useTracking } from "react-tracking"
 import { useAnalyticsContext } from "System/Hooks/useAnalyticsContext"
 import { useOnce } from "Utils/Hooks/useOnce"
 
 export const InquiryArtistsInCollection: FC = () => {
   const { next, relayEnvironment, context } = useInquiryContext()
+
+  const { submitUpdateMyUserProfile } = useUpdateMyUserProfile({
+    relayEnvironment: relayEnvironment.current as Environment,
+  })
 
   const {
     contextPageOwnerId,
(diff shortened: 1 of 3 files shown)


src/Components/Inquiry/Views/InquiryArtistsInCollection.tsx
```diff
@@ -9,22 +9,32 @@ import {
...
   } = useAnalyticsContext()
   const { trackEvent } = useTracking()
 
-  useOnce(() => {
+  useOnce(async () => {
+    await submitUpdateMyUserProfile({
+      promptedForUpdate: true,
+    })
+
     const payload: EditProfileModalViewed = {
```
iskounen, line 38: @leodmz we're currently firing a "EditProfileModalViewed" event when the user is shown the "add artists to collection" modal. Should we differentiate between the "complete profile" and "add artists..." prompt, or is it OK to use the same event for both impressions?

cc: @dzucconi

src/Components/Inquiry/Views/InquiryBasicInfo.tsx
```diff
@@ -21,27 +21,50 @@ import {
...
   const { sendToast } = useToasts()
-
   const { trackEvent } = useTracking()
+
   const { contextPageOwnerType } = useAnalyticsContext()
 
+  useOnce(async () => {
+    await submitUpdateMyUserProfile({
+      promptedForUpdate: true,
+    })
```
iskounen, line 54: @araujobarret just FYI that it was more straightforward to update this timestamp when the modal is displayed, rather than updating it for each action that can be performed on the modal.

dzucconi: You don't need to await this since the response doesn't matter

araujobarret: Question, as far as I know only when touching the button `I haven't started` should update it, not when closing the modal. Can you double-check with product please?

src/Components/Inquiry/Views/InquiryArtistsInCollection.tsx
```diff
@@ -25,6 +31,10 @@ export const InquiryArtistsInCollection: FC = () => {
   const { trackEvent } = useTracking()
 
   useOnce(() => {
+    submitUpdateMyUserProfile({
```
araujobarret, line 34: Question, as far as I know only when touching the button I haven't started should update it, not when closing the modal. Can you double-check with product please?

iskounen: Here is my understanding, but I can confirm:

- If the user adds an artist, we need to update it so that they aren't prompted to complete their profile within 30 days.
- If the user clicks "I haven't started...", we need to update it so they aren't prompted to add an artist again within 30 days.
- If the user dismisses the modal, we need to update it so they aren't prompted to add an artist again within 30 days.

src/Components/Inquiry/Views/InquiryArtistsInCollection.tsx
leodmz: My idea was to use the same event for both impressions to save me the trouble of having to create a separate event and later stitch them together to get all impressions

araujobarret (approved):

PR #14398, opened 2024-08-27
fix: removes unstated usage in normal app

dzucconi:
We currently use `unstated` in 2 places:

* A artist follow suggestion popover that hasn't been in use in at least a year
* Showing an error message in a dialog during the new checkout flow, where it should rather use a toast

We can get rid of both and drop the lib entirely. I've just deleted the popover here for now which should at least eliminate it from the bundle most of the time. If @erikdstock could follow up and delete it from the checkout flow, that'd be great.


damassi:
yesss, ty 👍

PR #14409, opened 2024-08-28
fix: Use MP's `isListed` field to check whether submission is listed

olerichter00:
Resolves https://www.notion.so/artsy/Web-listed-artwork-status-displayed-incorrectly-74aa11acc1c54278877f7d2425b9c7f8?pvs=4

## Description

We want to use Metaphysic's `isListed` field (https://github.com/artsy/metaphysics/pull/5857) to check whether the submission is listed or not. Also this PR removes the action label on the grid view when the artwork is listed.


src/Apps/MyCollection/Routes/MyCollectionArtwork/Components/MyCollectionArtworkSWASubmissionStatus.tsx
```diff
@@ -53,7 +53,7 @@ export const MyCollectionArtworkSWASubmissionStatus: React.FC<Props> = props =>
   // Setting the submission state as the URL query parameter to allow testing the complete Sell flow with different submission states with integrity.
   const submissionState = testSubmissionState ?? submission.state
 
-  const isListed = extractNodes(artwork.listedArtworksConnection).length > 0
+  const isListed = artwork?.isListed
```
olerichter00, line 56: No need to check the connection, we can just use `isListed`.

dariakoko (approved):
Thanks!

PR #14422, opened 2024-08-30
fix: ensure proxy *can* be disabled gracefully (DIA-818)

joeyAghion:
We noticed while experimenting with edge-caching that already-loaded Force sessions begin erroring if the proxy is disabled server-side. That's because the endpoint targeted by existing clients starts to respond with 404s. Instead, let's continue to support the proxy route even if the proxy is disabled. Freshly loaded pages will target un-proxied Metaphysics, while existing pages should continue to operate.


src/Server/config.ts
```diff
@@ -56,7 +56,7 @@ export const ENABLE_NEW_AUCTIONS_FILTER: any = false
 export const ENABLE_QUERY_BATCHING: any = false
 export const ENABLE_SAVED_SEARCH: any = false
 export const ENABLE_GRAPHQL_CACHE: any = false
-export const ENABLE_GRAPHQL_PROXY: any = true
+export const ENABLE_GRAPHQL_PROXY: any = false
```
joeyAghion, line 59: This allows [the documented command for disabling the cache](https://github.com/artsy/force/blob/main/docs/caching.md#caching) to work correctly. Deployed environments all have explicit values set. I've updated our `.env.shared` with `ENABLE_GRAPHQL_PROXY=true` so that development environments continue to proxy.

PR #14423, opened 2024-09-03
feat(EMI-2046): auction signals improvements

rquartararo:
The type of this PR is: **feat/fix**

This PR solves [EMI-2046]

### Description

This PR makes some improvements and fixes to the auction signals from the QA:

- Makes lot watcher count update in real-time
- Exports `BidInfoLine` into own component 
- Move `Details` related files into it’s own folder 
- Changes the “Bell” icon on the auction details to the “Heart” icon
- Only shows "Extended" text for lot detail cards
- Show lot watcher count when live auction is live

### Lot watcher count

https://github.com/user-attachments/assets/88f321e9-65f3-4150-8388-16c9ba046a88

https://github.com/user-attachments/assets/65b1d46d-9c76-4808-8001-d765d21b8ca0

[EMI-2046]: https://artsyproduct.atlassian.net/browse/EMI-2046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/Artwork/Details/BidTimerLine.tsx
+++ b/src/Components/Artwork/Details/BidTimerLine.tsx
@@ -3,15 +3,17 @@ import { useArtworkGridContext } from "Components/ArtworkGrid/ArtworkGridContext
 import { useTimer } from "Utils/Hooks/useTimer"
 import { Text } from "@artsy/palette"
 import { DateTime } from "luxon"
-import { createFragmentContainer, graphql } from "react-relay"
-import { BidTimerLine_artwork$data } from "__generated__/BidTimerLine_artwork.graphql"
+import { graphql, useFragment } from "react-relay"
+import { BidTimerLine_artwork$key } from "__generated__/BidTimerLine_artwork.graphql"
 
 interface BidTimerLineProps {
-  artwork: BidTimerLine_artwork$data
+  artwork: BidTimerLine_artwork$key
 }
 
 export const BidTimerLine: React.FC<BidTimerLineProps> = ({ artwork }) => {
-  const { collectorSignals } = artwork
+  const data = useFragment(bidTimerLineFragment, artwork)
+
+  const { collectorSignals } = data
   const { lotClosesAt, registrationEndsAt, onlineBiddingExtended } =
     collectorSignals?.auction ?? {}
   const { time } = useTimer(lotClosesAt ?? "")
@@ -82,19 +84,14 @@ export const BidTimerLine: React.FC<BidTimerLineProps> = ({ artwork }) => {
   )
 }
 
-export const BidTimerLineFragmentContainer = createFragmentContainer(
-  BidTimerLine,
-  {
-    artwork: graphql`
-      fragment BidTimerLine_artwork on Artwork {
-        collectorSignals {
-          auction {
-            lotClosesAt
-            registrationEndsAt
-            onlineBiddingExtended
-          }
-        }
+const bidTimerLineFragment = graphql`
+  fragment BidTimerLine_artwork on Artwork {
+    collectorSignals {
+      auction {
+        lotClosesAt
+        registrationEndsAt
+        onlineBiddingExtended
       }
-    `,
+    }
   }
-)
+`
--- a/src/Components/Artwork/Details/Details.tsx
+++ b/src/Components/Artwork/Details/Details.tsx
@@ -19,7 +19,7 @@ import { getSignalLabel } from "Utils/getSignalLabel"
 import { getSaleOrLotTimerInfo } from "Utils/getSaleOrLotTimerInfo"
 import { useAuctionWebsocket } from "Utils/Hooks/useAuctionWebsocket"
 import { useState } from "react"
-import { BidTimerLineFragmentContainer } from "./BidTimerLine"
+import { BidTimerLine } from "./BidTimerLine"
 
 export interface DetailsProps {
   artwork: Details_artwork$data
@@ -457,7 +457,7 @@ export const Details: React.FC<DetailsProps> = ({
         />
       )}
 
-      <BidTimerLineFragmentContainer artwork={rest.artwork} />
+      <BidTimerLine artwork={rest.artwork} />
 
       {padForActivePartnerOfferLine && <EmptyLine />}
     </Box>
--- a/src/Components/Artwork/Details/__tests__/BidTimerLine.jest.tsx
+++ b/src/Components/Artwork/Details/__tests__/BidTimerLine.jest.tsx
@@ -1,6 +1,6 @@
 import React from "react"
 import { screen } from "@testing-library/react"
-import { BidTimerLineFragmentContainer } from "Components/Artwork/Details/BidTimerLine"
+import { BidTimerLine } from "Components/Artwork/Details/BidTimerLine"
 import { useArtworkGridContext } from "Components/ArtworkGrid/ArtworkGridContext"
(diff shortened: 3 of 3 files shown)


src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
```diff
@@ -34,12 +34,28 @@ export const ArtworkActionsSaveButton: FC<ArtworkActionsSaveButtonProps> = ({
...
 
   const { handleSave: saveToDefaultCollection } = useSaveArtwork({
     isSaved: !!artwork.isSaved,
-    artwork,
+    artwork: {
+      internalID: artwork.internalID,
+      id: artwork.id,
+      slug: artwork.slug,
+      collectorSignals: {
+        auction: {
```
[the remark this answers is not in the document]
erikdstock: From the [hook](https://github.com/artsy/force/blob/d336d682aa0226a29b4d67d72b36b0bb5af85ba1/src/Components/Artwork/SaveButton/useSaveArtwork.ts#L36-L57) it looks like the artwork arg is mostly used for its ID parameter to [mutation it calls'](https://github.com/artsy/force/blob/d336d682aa0226a29b4d67d72b36b0bb5af85ba1/src/Components/Artwork/SaveButton/SaveArtworkMutation.ts#L8-L41) optimistic response. In addition I think that optimistic updates are [rolled back after getting the actual response.](https://relay.dev/docs/tutorial/mutations-updates/#improving-the-ux-with-an-optimistic-updater)  I'm not sure what here is necessary at all, but it's our older mutation style so I may be missing something.

src/Apps/Artwork/Components/ArtworkImageBrowser/ArtworkActionsSaveButton.tsx
```diff
@@ -98,6 +114,11 @@ export const ArtworkActionsSaveButtonFragmentContainer = createFragmentContainer
           isAuction
           isClosed
         }
+        collectorSignals {
+          auction {
+            lotWatcherCount
+          }
+        }
         ...ArtworkActionsWatchLotButton_artwork
```
erikdstock, line 122: This has me wondering if the number should just be part of the `ArtworkActionsWatchLotButton` component so the count can be  included in the update automatically - might make relay updates easier too.

src/Components/Artwork/SaveButton/useSaveArtwork.ts
```diff
@@ -38,6 +43,12 @@ export const useSaveArtwork = ({
                 id: artwork.id ?? "",
                 slug: artwork.slug ?? "",
                 isSaved: !isSaved,
+                collectorSignals: {
+                  auction: {
+                    lotWatcherCount:
+                      artwork.collectorSignals?.auction?.lotWatcherCount ?? 0,
+                  },
+                },
```
erikdstock, line 51: Ah, now I see. github linked me to these files from outside the PR branch. makes more sense now. Would `collectorSignals: artwork.collectorSignals ?? null` accomplish the same thing?

src/Components/Artwork/SaveButton/useSaveArtworkToLists.ts
```diff
@@ -87,6 +92,12 @@ export const useSaveArtworkToLists = (options: SaveArtworkToListsOptions) => {
             id: artwork.id,
             slug: artwork.slug,
             isSaved: !artwork.isSavedToDefaultList,
+            collectorSignals: {
+              auction: {
+                lotWatcherCount:
+                  artwork.collectorSignals?.auction?.lotWatcherCount ?? 0,
+              },
+            },
```
erikdstock, line 100: The repetition of this seems like a smell to me ...

erikdstock (commented):
A few comments, nothing that feels very certain. If we are using the `optimisticResponse` argument to relay I'm surprised that the counts are updating - no explicit increment in what we are providing and I wouldn't expect them to be in the metaphysics response if gravity hasn't reprocessed the count when the save artwork request completes.

Nothing blocking from me because I don't completely understand it but I would be interested to learn more or pair.

src/Apps/Settings/Routes/Auctions/Components/SettingsAuctionsLotStanding.tsx
```diff
@@ -3,7 +3,7 @@ import { Button, Column, Image, Flex, Text, Box } from "@artsy/palette"
 import React from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { RouterLink } from "System/Components/RouterLink"
-import { DetailsFragmentContainer } from "Components/Artwork/Details"
+import { DetailsFragmentContainer } from "Components/Artwork/Details/Details"
```
damassi, line 6: 👌

src/Components/Artwork/Details/BidTimerLine.tsx
```diff
@@ -0,0 +1,100 @@
...
+      lineHeight="22px"
+      color={textColor}
+      alignSelf="flex-start"
+    >
+      {renderLotCloseTime} left to bid
+    </Text>
+  )
+}
+
+export const BidTimerLineFragmentContainer = createFragmentContainer(
```
damassi, line 85: Since this is a new file, we should use modern relay patterns (meaning hooks -- `useFragment`)

src/Components/Artwork/Details/Details.tsx
```diff
@@ -207,7 +207,8 @@ const CollectorSignalLine: React.FC<DetailsProps> = ({
 
   return (
     <Text
-      variant="xs"
+      variant="sm-display"
+      lineHeight="22px"
```
damassi, line 211: I know its mentioned that things might revert up above, but: when we start messing with typography like this (especially on large surfaces like the artwork brick) we break the lego-block uniformity of the design system, which has its own baked in line-heights right inside each typography variant. Adjustments like this should only be a last resort.

src/Components/Artwork/Details/__tests__/BidTimerLine.jest.tsx
```diff
@@ -0,0 +1,144 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+    mockTimer("0", "5", "30")
+    mockArtworkGridContext(false)
+  })
+
+  const { renderWithRelay } = setupTestWrapperTL<BidTimerLineTestQuery>({
+    Component: props => {
+      if (props.artwork) {
+        return <BidTimerLineFragmentContainer {...(props as any)} />
```
damassi, line 38: Using relay hooks, we can do `<BidTimer ... />`

src/Components/Artwork/SaveButton/useSaveArtworkToLists.ts
```diff
@@ -15,6 +15,11 @@ type Artwork = {
   isInAuction: boolean
   isSavedToDefaultList: boolean
   isSavedToCustomLists: boolean
+  collectorSignals: {
+    auction: {
+      lotWatcherCount: number
+    } | null
+  } | null
```
damassi, line 22: #non-blocking: this `type Artwork` looks like it should be pulling from an auto-generated relay file, vs being hand rolled

damassi (dismissed):
Just a couple minor comments, but nothing blocking 👍

MrSltun (dismissed):
Looks good 🌟

PR #14428, opened 2024-09-04
chore(EMI-2043): auction signal tracking

rquartararo:

This PR solves [EMI-2043]

### Description

Adds tracking for auction signals lot watcher count and # of bids. 

[EMI-2043]: https://artsyproduct.atlassian.net/browse/EMI-2043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Home/Components/__tests__/HomeEmergingPicksArtworksRail.jest.tsx
+++ b/src/Apps/Home/Components/__tests__/HomeEmergingPicksArtworksRail.jest.tsx
@@ -40,7 +40,7 @@ describe("HomeEmergingPicksArtworksRail", () => {
   it("tracks artwork click", () => {
     renderWithRelay({
       Viewer: () => artworksConnection,
-      CollectorSignals: () => ({ partnerOffer: null }),
+      CollectorSignals: () => ({ partnerOffer: null, auction: null }),
     })
 
     fireEvent.click(screen.getAllByRole("link")[2])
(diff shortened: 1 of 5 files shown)


src/Apps/Home/Components/HomeEmergingPicksArtworksRail.tsx
```diff
@@ -72,6 +72,11 @@ export const HomeEmergingPicksArtworksRail: React.FC<HomeEmergingPicksArtworksRa
                 signal_label: artwork.collectorSignals
                   ? getSignalLabel(artwork.collectorSignals)
                   : "",
+                signal_bid_count:
+                  artwork.collectorSignals?.auction?.bidCount ?? undefined,
```
damassi, line 76: #minor / non-blocking nit / fyi -- `undefined` isn't something that's typically assigned, as undefined is undefined, by definition. Null however can be a value, and in this case is what you'd expect to see here and elsewhere if `bidCount` is falsy.

MrSltun (dismissed):
Looks good to me 🌟

src/Apps/Home/Components/HomeEmergingPicksArtworksRail.tsx
damassi: Ah, in that case, no problem at all and don't worry about it 👍 This was more of a general coding idiom thing, but i'd lean on whatever is over in cohesion. Didn't think about that side.

PR #14432, opened 2024-09-04
fix: load fewer lots into rails

joeyAghion:
I noticed these frequent queries for `size=99` hitting the API and thought they must be excessive, as no UIs appeared to ever list that many items. Tracing them, though, it seems like a few rails _could_ display up to `99` results. It's unlikely, but even if you e.g. follow Andy Warhol and visit a Warhol-only sale with hundreds of lots, I don't think it's necessary or practical to list that many in the top-most rail. Similarly for the other 2 cases (buy-now works associated with a sale, or current auctions while viewing a sale preview), though I think those are even less likely.


dzucconi:
Concur that it's totally not practical to have this many things in a rail

PR #14437, opened 2024-09-05
fix: add missing source/userAgent attributes for non-MyC submissions

dblandin:

Jira: [ONYX-1132](https://artsyproduct.atlassian.net/browse/ONYX-1132) 🔒


dariakoko (approved):
Thank you!

PR #14440, opened 2024-09-05
feat: match shipping and handling display on artwork to app.

oxaudo:
This follows the changes requested by this issue: https://artsyproduct.atlassian.net/browse/EMI-2024 - matches display to what app currently displays for this section.

BEFORE the changes:
Screen Shot 2024-09-05 at 5 57 58 PM

AFTER the changes:
Screen Shot 2024-09-05 at 6 00 41 PM

Out of scope for this ticket but I do wonder if we display proper info here. Like the case illustrated above I'm not sure what we are conveying by saying 'VAT included in price' and 'Taxes may apply at checkout'? Is not VAT the same as tax for EU locations? I also tried to view 'view more' link and it asks me for a separate login? Is that expected?

Another question here:  looks like we have a separate rendering for 'unlisted works' here. I've updated this section to also use text/data returned from MP. Hope that is what we want. I also notice that we only seem to track clicking 'learn more' link for unlisted works but not for listed works. Is that what we want @daytavares ? Or was it just overlooked?


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -11,71 +11,88 @@ export interface ShippingInformationProps {
...
-            rel="noopener noreferrer"
-            onClick={() => {
-              const payload: ClickedOnLearnMore = {
-                action: ActionType.clickedOnLearnMore,
-                context_module: "Sidebar",
-                subject: "Learn more",
-                type: "Link",
-                flow: "Shipping",
-              }
+        {!!taxInfo && (
```
oxaudo, line 34: I assumed that we also only want to display this line if tax info is present - similar to how we do this for listed works. Is that not the case?

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -11,71 +11,88 @@ export interface ShippingInformationProps {
...
-              trackEvent(payload)
-            }}
-          >
-            {t`artworkPage.sidebar.shippingAndTaxes.taxInformationLearnMore`}
-          </RouterLink>
-        </Text>
+                trackEvent(payload)
+              }}
+            >
+              {taxInfo.moreInfo.displayText}
```
oxaudo, line 54: Same here - use data from the server instead of hardcoding the string here.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -11,71 +11,88 @@ export interface ShippingInformationProps {
...
+      {!!priceIncludesTaxDisplay && (
+        <Text variant="xs" color="black60">
+          {priceIncludesTaxDisplay}
+        </Text>
+      )}
+
+      {!!taxInfo && (
+        <Text variant="xs" color="black60">
+          {taxInfo.displayText}{" "}
+          <RouterLink
```
oxaudo, line 85: Notice that the link clicked does not include tracking event. Only the unlisted link does. Not sure if that is intentional.

oxaudo (commented):
Here are the questions I mentioned in description:

daytavares:
checking :)

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
lidimayra: Ohh I don't feel confident on weighing on this one. I remember Peter spent quite some time discussing this with legal team before making a call about the rules for displaying shipping and tax copies. It might be worth to double check these with Jini.

daytavares:
hi @oxaudo ideally we should track for all artworks, but it's low importance / prio.

oxaudo:
@lidimayra - I propose we merge this and do a follow up based on the outcome of this story: https://artsyproduct.atlassian.net/browse/EMI-2074. Private artworks get very little usage anyways and we are not really displaying any wrong info here just seem to be omitting some info even if it's present.

oxaudo:
Cool. Merging here but I do think we should double check what the intention is for Private works. So thanks for raising that @lidimayra! We'll follow up on this one as a separate ticket.

PR #14442, opened 2024-09-06
chore: bump @babel/plugin-transform-runtime to 7.23.2

mc-jones:
The type of this PR is: chore

### Description
This PR bumps `@babel/plugin-transform-runtime` to `7.23.2` to address security vulnerability `CVE-2023-45133`. [Dependabot](https://github.com/artsy/force/security/dependabot/161).

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #161, opened 2016-09-15
Hide gallery follower counts on GPPs


garrengotthardt:
As we move follower counts into analytics emails we should hide them from GPPs (both 2.0 style GPPs and 1.0) for Subscribing Galleries.

We'll also be hiding institution and institutional seller counts but **not until early October** when a separate issue will follow

garrengotthardt:
![](https://github.trello.services/images/mini-trello-icon.png) [Hide Gallery Followers on GPPs (both 2.0 and 1.0 style pages) 🎣](https://trello.com/c/bON8Jrx0/578-hide-gallery-followers-on-gpps-both-2-0-and-1-0-style-pages)

PR #14444, opened 2024-09-06
chore: bump babel traverse to 7.25.6

mc-jones:
The type of this PR is: chore

### Description
This PR follows #14442, which aimed to resolve [dependabot alert #161](https://github.com/artsy/force/security/dependabot/161). However, in my quest for a quick win EOD yesterday, I didn't trace the dependency tree closely enough and addressed the incorrect issue. 

Therefore, this PR:

1. Reverts #14442 so that we can continue keep the `@babel/plugin-transform-runtime` within a close version of the other `@babel` versions. It also didn't resolve anything.
2. Updates `babel-plugin-inline-react-svg`, and `require-control` dev dependencies, which allows us to use the patched version of `@babel/traverse`.


narikazu (approved):
Nice one!
LGTM

PR #14454, opened 2024-09-10
feat(performance): Add usePrefetchRoute hook and add prefetch on RouterLink hover

damassi:

This PR solves [DIA-832]

### Description

This adds support for route prefetching on hover, via a new `usePrefetchRoute` hook which we can use this like so:

```tsx
const MyApp = () => {
  const { prefetch } = usePrefetchRoute({ path: '/foo/bar' })
  
  return (
    <>
      <Box onMouseOver={() => prefetch() }>Hover over me to prefetch</Box>
      <Box onMouseOver={() => prefetch('/bar') }>Hover over me to prefetch</Box>
    </>
}
```

Right now this is wired into our `RouterLink` component but we could extend its use any which way. 

Internally we're using Relay's `fetchQuery` fn, with `store-or-network` so that if something has already been fetched, it returns the result immediately and skips subsequent fetches. 

Still need to wire up A/B test but wanted to open this up so folks can pull it down and play with it, as well as branch off of this PR to wire up a review app that isn't segmented by unleash. 

**Demo**: 

https://github.com/user-attachments/assets/1792dea0-3591-4a8f-9c07-dfba2baf2981

cc @artsy/diamond-devs 

[DIA-832]: https://artsyproduct.atlassian.net/browse/DIA-832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/System/Components/RouterLink.tsx
```diff
@@ -27,24 +28,45 @@ export type RouterLinkProps = Omit<
...
 
-    const { target } = rest
     // If displaying the linked URL in the same browsing context, e.g. browser tab.
-    const isSameBrowsingContext = !target || target === "_self"
+    const isSameBrowsingContext = !rest.target || rest.target === "_self"
     const isRouterAware = isSupportedInRouter && isSameBrowsingContext
 
+    const handleMouseOver = () => {
+      prefetch()
+    }
```
damassi, line 49: Prefetch on hover ☝️

src/System/Contexts/SystemContext.tsx
```diff
@@ -1,99 +1,71 @@
...
 export type UserPreferences = {
   metric: Metric
 }
 
-/**
- * FIXME: Use a proper state management library. Ran into problems with useReducer
- * leading to an infinite loop.
- */
-export type SystemContextState = Partial<{
+export interface SystemContextState {
```
damassi, line 14: Just a bunch of cleanup in this file

src/System/Hooks/usePrefetchRoute.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface UsePrefetchRouteProps {
+  path: string
+}
+
+export interface RouteMatch {
+  route: RouteProps
+  match: MatchResult<any>
+}
+
+export const usePrefetchRoute = (
```
damassi, line 18: Main addition

src/System/Hooks/usePrefetchRoute.tsx
```diff
@@ -0,0 +1,82 @@
...
+  props: UsePrefetchRouteProps
+): { prefetch: () => any } => {
+  const { relayEnvironment } = useSystemContext()
+
+  const { match } = useRouter()
+
+  const prefetch = () => {
+    const routes = findRoutesByPath({ path: props.path })
+
+    routes.forEach(foundRoute => {
```
damassi, line 28: This is somewhat eager just in the sense that in our router we sometimes have multiple matchable routes for a given path. 

A good example of this [is here](https://github.com/artsy/force/blob/damassi/feat/prefetch-on-hover/src/Apps/Artist/artistRoutes.tsx#L132), in the `artistRoutes`. We have the parent route (`/artist/:artistID`) which contains `children` routes, of which one child route path (the default subroute) is the works for sale route, keyed by `path: ""`. So this automatically matches when navigating to `/artist/foo`. There's a few places in our router like this, so lets be safe and fetch all related stuff. This might lead to less matches, however.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3e0e761bb64a7d290620c9c263f75a615bb9ada0
</p>

src/System/Hooks/usePrefetchRoute.tsx
```diff
@@ -0,0 +1,79 @@
...
+  props: UsePrefetchRouteProps
+): { prefetch: () => Array<Subscription | null> } => {
+  const { relayEnvironment } = useSystemContext()
+
+  const { match } = useRouter()
+
+  const prefetch = () => {
+    const routes = findRoutesByPath({ path: props.path })
+
+    const querySubscriptions = routes.map(foundRoute => {
```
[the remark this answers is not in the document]
damassi: We return a query subscription for each route, so in the future we'll be able to cancel fetches that scroll out of the viewport (during the next phase of this spike).

src/System/Relay/createRelaySSREnvironment.ts
```diff
@@ -30,7 +30,8 @@ const isServer = typeof window === "undefined"
 const isDevelopment = getENV("NODE_ENV") === "development"
 
 // Only log on the client during development
-const loggingEnabled = isDevelopment && !isServer
+const loggingEnabled =
+  !isServer && (isDevelopment || window.artsy.isLoggerEnabled)
```
damassi, line 34: This is so we can enable debuggable logging when deployed to staging / prod

src/System/Hooks/usePrefetchRoute.tsx
```diff
@@ -0,0 +1,75 @@
...
+      })()
+
+      const isPrefetchable = !!(query && variables)
+
+      if (!isPrefetchable) {
+        return null
+      }
+
+      const subscription = fetchQuery(relayEnvironment, query, variables, {
+        fetchPolicy: "store-or-network",
```
olerichter00, line 49: So here we're using store-**or**-network instead of store-**and**-network because we don't want to refetch an already prefetched route, right?

src/Components/NavBar/NavBar.tsx
```diff
@@ -92,6 +95,15 @@ export const NavBar: React.FC = track(
     }
   }, [isMobile])
 
+  // Prefetch most clicked links.
+  // See: https://artsy.slack.com/archives/C05EEBNEF71/p1726074422149369?thread_ts=1726073316.088559&cid=C05EEBNEF71
+  useEffect(() => {
+    const PREFETCH_NAVBAR_LINKS = ["/artists", "/collect"]
+
+    PREFETCH_NAVBAR_LINKS.forEach(prefetch)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
araujobarret, line 104: 🧹 ? 😬

araujobarret (dismissed):
Nice feature, really looks interesting!

src/Components/NavBar/NavBar.tsx
damassi: Heh, I am an anti exhaustive-deps person. I have seen more subtle bugs by following the guidance than without. (Deps are something one should be extremely deliberate with.)

araujobarret: Usually it's a sign that we forgot something like memoize a function signature before using it in a useEffect, dealt with some bugs in the past because ppl forgot to put the deps there
I know @dzucconi shares some thoughts on it too

dzucconi: If u wanna just do something once on mount u can use: https://github.com/artsy/force/blob/main/src/Utils/Hooks/useOnce.ts

damassi: > forgot something like memoize a function signature before using it in a useEffect

I'm just into minimalism in this regard. I don't think wiring up basic things with `useCallback` and so on is necessary. It just adds complexity overhead. That's not to say those things don't have a purpose, but I'd rather skip the linter than satisfy it by adding another layer of (unnecessary) abstraction. 

BUT I like that useOnce 👍 Didn't know about it. Will use.

araujobarret: I feel it, the code becomes verbose and the gain is none! I hope they come soon with react Forget that could change this way

damassi:
Things are behind a feature flag; merging

PR #14479, opened 2024-09-16
feat(prefetch): Expand surface areas with useInternalHref

damassi:

This PR solves [DIA-850]

### Description

This expands coverage of prefetching on home by checking to see if any urls have the full domain, and then stripping it to return an internal href (which is prefetchable). 

cc @artsy/diamond-devs


damassi:
Lemme know if you have any requests here @dzucconi - can follow up.

dzucconi (commented):
Great idea

PR #13426, opened 2024-01-26
chore(alerts): clean up `onyx_artwork_alert_modal_v2` flag and dead code

anandaroop:
The type of this PR is: **Chore**

This PR solves [ONYX-680]

Review app: https://onyx-680.artsy.net

### Description

Now that v2 alerts have been out in the wild for a while, we are starting to clean up feature flags and dead code.

This PR starts that process with the `onyx_artwork_alert_modal_v2` feature flag.

Tests are passing; but for extra confidence that this does not introduce any regressions feel free to try out https://onyx-680.artsy.net

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ONYX-680]: https://artsyproduct.atlassian.net/browse/ONYX-680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

(h/t @tam-kis for pairing on the test fixes!)

PR #14530, opened 2024-09-20
chore: remove overfetching in ArtworkSidebarCreateAlert

dblandin:

### Description

Follow-up to the cleanup [here](https://github.com/artsy/force/pull/13426). This removes some GraphQL querying that was necessary in the previous implementation.

Since revising the alert flow, most of the querying for this button is no longer necessary.


anandaroop (approved):
🙏🏽 Nice! 

Curious how this was caught, just happened to notice it, or something else?

dblandin:
@anandaroop I was looking into [this data issue](https://artsy.slack.com/archives/C05EQL4R5N0/p1726758056459239?thread_ts=1726758056.459239&cid=C05EQL4R5N0) and just stumbled upon it. Not very scientific I'm afraid!

PR #14532, opened 2024-09-20
feat(performance): Enable prefetching on enter for logged in

damassi:

### Description

We've let prefetch-on-enter run for a bit for logged out and things are looking good! So lets enable it for logged-in users as well. 

cc @artsy/diamond-devs


damassi:
closing this because we merged the review app!

PR #14534, opened 2024-09-23
chore: removes feature flags

dzucconi:
I noticed this looking at the load flame graph. Importing `getFeatureFlag` would also bring with it all of `SystemContext`. I pulled that function out into its own file to prevent that, but might as well remove these flags anyway.

### DIFF
--- a/src/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
+++ b/src/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
@@ -1,5 +1,4 @@
 import { getInitialFilterState } from "Components/ArtworkFilter/Utils/getInitialFilterState"
-import { getFeatureFlag } from "System/Hooks/useFeatureFlag"
 
 export function getWorksForSaleRouteVariables({ artistID }, { location }) {
   // FIXME: The initial render includes `location` in props, but subsequent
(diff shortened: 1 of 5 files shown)


src/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts
```diff
@@ -22,12 +21,10 @@ export function getWorksForSaleRouteVariables({ artistID }, { location }) {
...
   ]
 
-  const includeBlurHash = !!getFeatureFlag("diamond_blurhash-on-artist-pages")
-
   return {
     input: filterParams,
     aggregations,
     artistID,
-    includeBlurHash,
+    includeBlurHash: false,
```
dzucconi, line 28: Rather than remove all of the query arguments scattered around. Dunno if there's a preferable way to do this ATM.

src/Apps/Artwork/artworkRoutes.tsx
```diff
@@ -33,16 +31,8 @@ export const artworkRoutes: RouteProps[] = [
...
 
-      const enableCustomErrorPage = getFeatureFlag(
-        "onyx_custom_artwork_error_page"
-      )
-
       if (requestErrorStatusCode) {
-        if (enableCustomErrorPage) {
-          updateContext("statusCode", requestErrorStatusCode)
-        } else {
-          throw new HttpError(requestErrorStatusCode)
```
araujobarret, line 44: don't you need to update where we handle these errors to not to anymore?

araujobarret (approved):

src/Apps/Artwork/artworkRoutes.tsx
dzucconi: I did — it just does what the flag was enabling now, which is apparently putting the error code into async local storage.

araujobarret:
CI looks fine:
image

PR #14535, opened 2024-09-23
chore: updates browserlist

dzucconi:
Not expecting much in the way of changes since it looks like this was actually updated 2 months ago. Though the way this was resolving in the lockfile was a little weird? Anyway we were getting a warning on boot to update this, so.


dzucconi:
No change!

PR #14553, opened 2024-09-24
refactor(ci): Simplify CI pipeline 

damassi:

Addresses [DIA-894] 

### Description

Force's CI somehow ended up vastly more complex than makes sense for a simple node.js express app that serves as a pass-through to MP. It had multi-stage builds, a wild Dockerfile, a custom image setup for cypress, very little parallelization (and thus lots and lots of waiting), and... confusion everywhere. We used to have fast runs before this previous refactor pass (that I'm now refactoring away) took place. So lets get cycle times back to acceptable. 

Along the way:
- Removed reliance on artsy/yarn orb in favor of circleci/node. AFAIK there are very few apps that are still using that, so we should consider archiving that repo since CircleCI's version is better than hand-rolled code. 

**Before:**

Screenshot 2024-09-23 at 10 05 07 PM

Screenshot 2024-09-23 at 9 51 06 PM

**After:**

Screenshot 2024-09-23 at 10 04 20 PM

<br>

Screenshot 2024-09-23 at 9 50 40 PM

[DIA-894]: https://artsyproduct.atlassian.net/browse/DIA-894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.circleci/config.yml
```diff
@@ -2,22 +2,135 @@ version: 2.1
...
+            JEST_JUNIT_ADD_FILE_ATTRIBUTE: "true"
+      - store_artifacts:
+          path: /app/reports/junit/
+      - store_test_results:
+          path: /app/reports/junit/
+
+  smoke-tests:
+    executor: cypress/default
+    working_directory: ~/app
+    parallelism: 6
```
damassi, line 80: Unfortunately we couldn't use the slick new cypress orb API because their parallel features require a subscription. So had to go the manual way in order to lean on CircleCI's parallelism. But things are still vastly simpler setup-wise than before.

.circleci/config.yml
```diff
@@ -119,263 +239,70 @@ jobs:
...
-                command: echo 'export BUILD_TARGET="builder";' >> $BASH_ENV
+      # Danger
+      - node/run:
+          <<: *not_main_or_staging_or_release
+          context:
+            - danger-github-oss
+          name: check-pr
+          pkg-manager: yarn
+          yarn-run: danger:ci
+          version: 18.15.0
```
damassi, line 256: Note `node/run` as an official replacement for `artsy/yarn`, running node scripts, and caching. Whenever we encounter `artsy/yarn` we should remove!

damassi:
Not sure whats up with these phantom jobs 🤷 there's no longer any mention of these in the Circle config file: 

Screenshot 2024-09-23 at 10 08 30 PM

Dockerfile
```diff
@@ -1,185 +1,27 @@
...
-#              |                  |                                       |                  |
-#              |  electron-runner |                                       |  production      |
-#              |                  |                                       |                  |
-#              +------------------+                                       +------------------+
-
-# ---------------------------------------------------------
-# Base build dependencies
-# ---------------------------------------------------------
-FROM node:18.15.0-alpine3.17 as builder-base
+FROM node:18.15-alpine as builder-base
```
starsirius, line 1: I think one goal of the multistage builds are smaller image size (e.g. for production). I'm not sure if it's still a requirement, but it could be good information to compare the size before and after.

starsirius: And I agree it was kind of complicated. 😅

.circleci/config.yml
```diff
@@ -119,263 +239,70 @@ jobs:
...
-            - run:
-                command: echo 'export BUILD_TARGET="electron-runner";' >> $BASH_ENV
+          name: type-check
+          pkg-manager: yarn
+          yarn-run: type-check
+          version: 18.15.0
 
-      # Test steps
+      # Jest Tests
       - test:
```
starsirius, line 275: It seems test results (and artifacts) failed to upload so we'll lose test insights. The paths may need to update after migrating to use CircleCI images.

Screenshot 2024-09-24 at 9 05 23 AM

Dockerfile
artsyjian: > I think one goal of the multistage builds are smaller image size (e.g. for production). I'm not sure if it's still a requirement, but it could be good information to compare the size before and after.

Image size still important. If we can keep it [< 1GB](https://artsy.slack.com/archives/CA8SANW3W/p1726765549835789?thread_ts=1726761304.356259&cid=CA8SANW3W), we are good.

artsyjian (approved):
nice!

.circleci/config.yml
```diff
@@ -2,22 +2,135 @@ version: 2.1
...
+          name: Run Jest Tests
+          command: |
+            yarn test --reporters=default --reporters=jest-junit --shard=$(expr $CIRCLE_NODE_INDEX + 1)/$CIRCLE_NODE_TOTAL
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: reports
+            JEST_JUNIT_ADD_FILE_ATTRIBUTE: "true"
+      - store_artifacts:
+          path: /app/reports/junit/
+      - store_test_results:
+          path: /app/reports/junit/
```
starsirius, line 75: It seems test results (and artifacts) failed to upload so we'll lose test insights. The paths may need to update after migrating to use CircleCI images.

Screenshot 2024-09-24 at 9 05 23 AM

damassi: Will follow up with another PR 👍

Dockerfile
damassi: Last we checked in slack the prod image was 400mb? We should double check this tho next later this week since I did adjust the docker file.

damassi:
Fixed in https://github.com/artsy/force/pull/14559

PR #14555, opened 2024-09-24
chore: removes jsdom middleware

dzucconi:
Just something random I noticed while trying to get a flame graph up of the server render. This is one of the only JS files so it stood out. We haven't needed `jsdom` on the server since we redid the article pages.


araujobarret (dismissed):
🧹

egdbear (dismissed):

PR #14575, opened 2024-09-25
fix: fix 500 when creating artworks list (saves)

nickskalkin:

### Description

When user creates a saves list (from `/favorites/saves`) - page fails with 500 error. Honestly idk what happens exactly, but the error suggest to wrap the section in Suspense. For that I had to extract the whole saves lists rail into a query renderer component.

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/c1b5b56b-0d2e-43bb-9dc8-8b4042d79a90" /> | <video src="https://github.com/user-attachments/assets/166aa1dd-b451-48ff-92fb-52c23380d102" /> |

### DIFF
--- a/src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListItemsList.tsx
+++ b/src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListItemsList.tsx
@@ -1,0 +1,112 @@
+import { Shelf, SkeletonBox } from "@artsy/palette"
+import { ArtworkListItemFragmentContainer } from "Apps/CollectorProfile/Routes/Saves/Components/ArtworkListItem"
+import { FC, useRef } from "react"
+import { graphql, useLazyLoadQuery } from "react-relay"
+import { useRouter } from "System/Hooks/useRouter"
+import { extractNodes } from "Utils/extractNodes"
+import { ArtworkListItemsListQuery } from "__generated__/ArtworkListItemsListQuery.graphql"
+import { times } from "lodash"
+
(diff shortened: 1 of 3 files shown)


src/Apps/CollectorProfile/Routes/Saves/Components/ArtworkListItemsList.tsx
```diff
@@ -0,0 +1,112 @@
...
+    match: { params },
+  } = useRouter()
+
+  const initialArtworkListId = useRef(params.id)
+
+  if (!me) {
+    return null
+  }
+
+  const savedArtworksArtworkList = me.savedArtworksArtworkList
```
nickskalkin, line 23: All this logic was straight extracted from [CollectorProvileSavesRoute](https://github.com/artsy/force/blob/b7c4dfe237bf7948428cf62df7fd7ec6f6ce90a5/src/Apps/CollectorProfile/Routes/Saves/CollectorProfileSavesRoute.tsx#L25) component - I don't want to refactor anything in this. PR

nickskalkin:
I **think** that https://github.com/artsy/force/blob/b7c4dfe237bf7948428cf62df7fd7ec6f6ce90a5/src/Apps/CollectorProfile/Routes/Saves/CollectorProfileSavesRoute.tsx#L25 can be refactored and optimized - I think it fetches data it doesn't need, but I don't want to refactor it in this PR

dblandin (approved):
LGTM. Thanks for tracking this down 👍🏼

PR #14579, opened 2024-09-25
feat(EMI-2113): add payment failed status and link to order history

rquartararo:

This PR solves [EMI-2113]

### Description

This allows the order history page to display "Payment failed" and links to add a new payment to the order history page when `displayState` is `PAYMENT_FAILED`. 

Screenshot 2024-09-26 at 12 31 20 PM

https://github.com/user-attachments/assets/68aeb9d3-41da-478a-a31e-b0f4e858b664

[EMI-2113]: https://artsyproduct.atlassian.net/browse/EMI-2113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
+++ b/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
@@ -1,5 +1,6 @@
 import {
   Box,
+  Button,
   Column,
   EntityHeader,
   Flex,
@@ -77,7 +78,7 @@ const getPaymentMethodText = (
 }
 
 const getOrderLink = order => {
-  const isOrderActive = order.state !== "CANCELED" && order.state !== "REFUNDED"
+  const isOrderActive = !["CANCELED", "REFUNDED"].includes(order.state)
   const isOrderPaymentFailed = order.displayState === "PAYMENT_FAILED"
 
   if (isOrderPaymentFailed) {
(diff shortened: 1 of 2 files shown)


oxaudo (dismissed):
This looks good to me! The only thing according to this: https://www.figma.com/design/UyNOc2vz3bf9k6Cw2itqwd/Home-Feed-(App)?node-id=1097-52417&t=qD3JG8pm5vNdsIwc-4 seems like we also want a big black button in the content of this row in case the payment is failing. Can add here or do a separate Pr. Either way is cool with me.

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -195,13 +220,7 @@ const SettingsPurchasesRow: FC<SettingsPurchasesRowProps> = ({ order }) => {
...
 
           <Text variant="sm-display" color="black60">
-            {isOrderActive ? (
-              <RouterLink inline to={`/orders/${order.internalID}/status`}>
-                {order.code}
-              </RouterLink>
-            ) : (
-              order.code
-            )}
+            {getOrderLink(order)}
```
erikdstock, line 223: Rather than a mix of function calls and object indexing to get color, label, text and link, I think I would get these values in the body of the FC - maybe using one single function at the top of the component body to resolve all of these values, or maybe just doing it inline.

erikdstock (commented):
LGTM - one minor/non-blocking comment.

oxaudo (approved):
Looks good to me!

PR #14581, opened 2024-09-26
feat(dia-835): setup sentry performance dashboard

dzucconi:
Closes [DIA-835](https://artsyproduct.atlassian.net/browse/DIA-835)

This upgrades Sentry to the latest version and optionally enables tracing by setting `SENTRY_TRACING_ENABLED` to `true`.


dzucconi:
Hm — realizing that the way our routes are set up is that it's thinking these are Express routes — so there's just a single mega route:

src/Server/config.ts
```diff
@@ -107,6 +107,7 @@ export const SEGMENT_WRITE_KEY_SERVER: any = null
 export const SEGMENT_WRITE_KEY: any = null
 export const SENTRY_PRIVATE_DSN: any = null
 export const SENTRY_PUBLIC_DSN: any = null
+export const SENTRY_TRACING_ENABLED: any = false
```
egdbear, line 110: nitpick: boolean?

egdbear (dismissed):

src/server.ts
```diff
@@ -1,3 +1,5 @@
+import "instrument"
```
araujobarret, line 1: do we want to import it all the time or only when `SENTRY_PRIVATE_DSN` is present?

araujobarret (approved):

src/Server/config.ts
dzucconi: It's coming from an environment variable so you can set whatever. Something does apparently coerce though.

src/server.ts
dzucconi: All the time. This is how it was before but it was set up incorrectly.

damassi (approved):

dzucconi:
@damassi do you want to take a look at this?

damassi (approved):
@dzucconi - do you need to set the hokusai ENV var?

damassi:
I'll peek at this later in the week 👍

PR #14585, opened 2024-09-26
feat(router-link): Add enablePrefetch flag to link

damassi:

### Description

Adds the ability to toggle off link prefetching via `prefetchEnabled` prop on RouterLink, and disables for admin items. Also guards against erronious queries to the user menu. Not sure if that fixes anything related to [this reported issue](https://artsy.slack.com/archives/C03N12SR0RK/p1727365743754609) as I cannot reproduce, but lets just be safe.

### DIFF
--- a/src/Components/NavBar/NavBarLoggedInActions.tsx
+++ b/src/Components/NavBar/NavBarLoggedInActions.tsx
@@ -24,6 +24,10 @@ import { ProgressiveOnboardingSaveFind } from "Components/ProgressiveOnboarding/
 import { ProgressiveOnboardingAlertFind } from "Components/ProgressiveOnboarding/ProgressiveOnboardingAlertFind"
 import { extractNodes } from "Utils/extractNodes"
 import { getENV } from "Utils/getENV"
+import createLogger from "Utils/logger"
+import { FallbackErrorBoundary } from "System/Components/FallbackErrorBoundary"
+
+const logger = createLogger("NavBarLoggedInActions.tsx")
 
 /** Displays action icons for logged in users such as inbox, profile, and notifications */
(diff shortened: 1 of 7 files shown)


src/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -196,7 +199,16 @@ export const NavBarLoggedInActionsQueryRenderer: React.FC<{}> = () => {
           }
         }
       `}
-      render={({ props }) => {
+      render={({ error, props }) => {
+        if (error) {
+          logger.error(error)
```
damassi, line 204: We should see this come through sentry now since we're using our logger

mzikherman (approved):
Another potential API is to maintain a list of routes that aren't pre-fetched, and then just don't prefetch those. That might live as part of route configuration, vs. more ad-hoc sprinkled around on the actual `RouterLink` calling components.

damassi:
Ah, good idea 👍 I'll add that here real quick.

src/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -196,9 +199,27 @@ export const NavBarLoggedInActionsQueryRenderer: React.FC<{}> = () => {
...
-      render={({ props }) => {
+      placeholder={PLACEHOLDER}
+      render={({ error, props }) => {
+        if (error) {
+          logger.error(error)
+          return PLACEHOLDER
+        }
+
+        if (!props || !props.me) {
+          return PLACEHOLDER
```
damassi, line 210: Needed to add a placeholder to guard against query errors.

PR #14588, opened 2024-09-26
feat: separate quick filter menus from featured keywords

joeyAghion:
Chatting with Ryan today, we puzzled over how to visually distinguish featured keywords, which operate almost as toggles, from the quick filter menus that expand. He's refining a few alternatives, but in the meantime we thought they should _at least_ be separated somehow. In this PR, I borrow the separator used on artists' artwork grids:

Screenshot 2024-09-26 at 5 02 51 PM

Looks like:

Screenshot 2024-09-26 at 4 59 13 PM

Q: should this separator and its spacing be elevated to Palette?

### DIFF
--- a/src/Components/ArtworkFilter/ArtworkFiltersQuick/FeaturedKeywordsFilterQuick.tsx
+++ b/src/Components/ArtworkFilter/ArtworkFiltersQuick/FeaturedKeywordsFilterQuick.tsx
@@ -1,6 +1,6 @@
 import { FC } from "react"
 import { useArtworkFilterContext } from "Components/ArtworkFilter/ArtworkFilterContext"
-import { Pill } from "@artsy/palette"
+import { Box, Pill, Spacer } from "@artsy/palette"
 
 export interface FeaturedKeywordsFilterQuickProps {
   featuredKeywords: readonly string[] | null | undefined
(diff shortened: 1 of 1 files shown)


src/Components/ArtworkFilter/ArtworkFiltersQuick/FeaturedKeywordsFilterQuick.tsx
```diff
@@ -16,6 +16,10 @@ export const FeaturedKeywordsFilterQuick: FC<FeaturedKeywordsFilterQuickProps> =
   if (!props.featuredKeywords || !props.featuredKeywords.length) return null
   return (
     <>
+      <Spacer y={2} />
+      <Box width="1px" bg="black30" />
+      <Spacer y={2} />
```
dzucconi, line 21: It'd be better to insert this in the parent component: https://github.com/artsy/force/blob/bd89cb774fc63423f06146bbddb9e03194184927/src/Components/ArtworkFilter/ArtworkFiltersQuick.tsx#L44

Rather than having space hanging off this one.

dzucconi:
and no need to put anything in Palette

src/Components/ArtworkFilter/ArtworkFiltersQuick/FeaturedKeywordsFilterQuick.tsx
joeyAghion: Even if that requires the parent to do that conditionally, using the same logic as above?

dzucconi: Oh... I didn't see that. Yeah, that's annoying. This is fine!

dzucconi: Technically you could get away with out the empty space hanging off, but it doesn't matter if this thing isn't going to be used anywhere else.

PR #14601, opened 2024-10-01
fix(EMI-2123): dont redirect for displayState != submitted on payment failed route

erikdstock:

This PR solves [EMI-2123]

### Description
We have a redirect rule which is incompatible with the new PAYMENT_FAILED display state. I don't think it is needed for this route but wouldn't mind a second set of eyes. Might need tests.

[EMI-2123]: https://artsyproduct.atlassian.net/browse/EMI-2123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/redirects.tsx
+++ b/src/Apps/Order/redirects.tsx
@@ -118,12 +118,16 @@ const goToStatusIfOrderIsNotSubmitted = goToStatusIf(
 )
 
 const goToStatusIfNotLastTransactionFailed = goToStatusIf(
-  order => !order.lastTransactionFailed,
-  "Order's lastTransactionFailed must be true"
+  order => !(order.lastTransactionFailed && order.state === "SUBMITTED"),
+  "Order's lastTransactionFailed must be true + state submitted"
 )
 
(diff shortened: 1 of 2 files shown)


src/Apps/Order/redirects.tsx
```diff
@@ -205,7 +205,6 @@ export const redirects: RedirectRecord<OrderQuery> = {
       path: "payment/new",
       rules: [
         goToStatusIfBuyNowCreditCardOrder,
-        goToStatusIfOrderIsNotSubmitted,
```
oxaudo, line 208: Thanks for finding it! 

You think the idea here is that we only allow to update payment for the order if payment is failing? Wonder if we should update `goToStatusIfNotLastTransactionFailed` to look at displayState now? This way we are moving a step closer to using one consolidated `state` for those checks?

oxaudo (dismissed):

PR #14606, opened 2024-10-02
chore(EMI-2132): remove secondary button from payment failed dialog

oxaudo:
We noticed that in the 'update payment' flow if the new payment is failing as well we are displaying really confusing CTA. The easy solution was to at least remove extra CTA button that does not seem to add anything.

Before:
Screenshot 2024-10-01 at 11 25 11 AM

After:
Screenshot 2024-10-02 at 1 10 25 PM

### DIFF
--- a/src/Apps/Order/Routes/Accept/index.tsx
+++ b/src/Apps/Order/Routes/Accept/index.tsx
@@ -194,7 +194,6 @@ export const Accept: FC<AcceptProps & StripeProps> = props => {
   }) => {
     const { confirmed } = await dialog.showConfirmDialog({
       ...props,
-      cancelButtonText: "OK",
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Accept/index.tsx
```diff
@@ -194,7 +194,6 @@ export const Accept: FC<AcceptProps & StripeProps> = props => {
   }) => {
     const { confirmed } = await dialog.showConfirmDialog({
       ...props,
-      cancelButtonText: "OK",
```
starsirius, line 197: OK! 👌

starsirius (approved):
💯 Good improvement!

PR #14607, opened 2024-10-03
feat(sentry): Add route-level span tracing

damassi:

### Description

This updates our sentry client performance config to tap into route-level transition spans. Will likely have to tune this, but this starts giving us more accurate stats (I think!). 

Matches by route id for proper grouping in sentry:
- Eg `/artist/:slug` 

See docs here: https://docs.sentry.io/platforms/javascript/tracing/instrumentation/automatic-instrumentation/

Screenshot 2024-10-02 at 5 00 16 PM


dzucconi:
Nice!

PR #14609, opened 2024-10-03
chore: review app defer tag

dzucconi:
Making the minimal change here because I still don't understand what's going on. Wanting to see if I can replicate what I'm seeing locally in a more production-like environment.


dzucconi:
Will revisit the defer.

PR #14613, opened 2024-10-04
fix: do not show # of bids after lot ends

starsirius:

This PR solves [EMI-2022]

### Description

See [Slack](https://artsy.slack.com/archives/C02JHHHKP5K/p1727989357076919?thread_ts=1727370726.687139&cid=C02JHHHKP5K) for context.

We show `(0 bids)` on lots with bids after the lot ends. It's because we only hide the # of bids when _both_ the lot count from auction signals and bidder position count from sale artwork are 0 or not available. We shouldn't be checking both of them at the same time, and the fix is to only check the desired one based on the feature flag.

Screenshot 2024-10-03 at 5 01 37 PM

### DIFF
--- a/src/Components/Artwork/Details/__tests__/Details.jest.tsx
+++ b/src/Components/Artwork/Details/__tests__/Details.jest.tsx
@@ -636,6 +636,28 @@ describe("Details", () => {
 
       expect(html).toContain("Increased Interest")
     })
+
+    it("does not show the number of bids when there are bids on the sale artwork but no auction signals", async () => {
+      const data: any = {
+        ...artworkInAuction,
+        collectorSignals: {
+          ...artworkInAuction?.collectorSignals,
+          auction: null,
+        },
+        sale_artwork: {
+          ...artworkInAuction?.sale_artwork,
+          counts: {
+            ...artworkInAuction?.sale_artwork?.counts,
+            bidder_positions: 2,
+          },
+        },
+      }
+
+      const wrapper = await getWrapper(data)
+      const html = wrapper.html()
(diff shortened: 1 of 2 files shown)


src/Components/Artwork/Details/__tests__/Details.jest.tsx
```diff
@@ -636,6 +636,28 @@ describe("Details", () => {
...
+        ...artworkInAuction,
+        collectorSignals: {
+          ...artworkInAuction?.collectorSignals,
+          auction: null,
+        },
+        sale_artwork: {
+          ...artworkInAuction?.sale_artwork,
+          counts: {
+            ...artworkInAuction?.sale_artwork?.counts,
+            bidder_positions: 2,
```
starsirius, line 651: There are bids on the lot.

src/Components/Artwork/Details/__tests__/Details.jest.tsx
```diff
@@ -636,6 +636,28 @@ describe("Details", () => {
 
       expect(html).toContain("Increased Interest")
     })
+
+    it("does not show the number of bids when there are bids on the sale artwork but no auction signals", async () => {
+      const data: any = {
+        ...artworkInAuction,
+        collectorSignals: {
+          ...artworkInAuction?.collectorSignals,
+          auction: null,
```
starsirius, line 645: This reproduces the state where an artwork has no auction signal (e.g. after the lot ends), [but...](https://github.com/artsy/force/pull/14613/files#r1786984886)

mzikherman (approved):

PR #14615, opened 2024-10-04
feat: improve lcp on artwork page

dzucconi:
Cherry picking what should be a straightforward change from https://github.com/artsy/force/pull/14609 to see if I can deploy a review app. Though this should be mergeable without.

https://web.dev/articles/fetch-priority


dzucconi:
Seems to actually move the needle a bit on the LCP.

PR #14622, opened 2024-10-07
fix: Adds handling in case of `me` being null

jpotts244:
The type of this PR is: Fix

### Description

Noticed this bug on staging and production
If I am viewing my saved alerts and then go to log out from the menu bar, I receive a `Internal Server error 500`
Though, I am technically signed out, seeing the 500 seems to be a bit of a dead end as a user

Also to note, this only happens on desktop sized screens! Mobile correctly shows the sign in page

This is happening because I'm not logged out and `me` == null when trying to load in the alerts page

Example of the error in Sentry and Force:
- https://artsynet.sentry.io/issues/5900188608/?project=159064&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=1

Screenshot 2024-10-07 at 10 31 26 AM

---

I've adjusted the code to handle the case when `me` is `null` (aka logged out user), so they can see something like this and not a 500 exception

Screenshot 2024-10-07 at 10 44 51 AM

---

⚠️ Pinging @artsy/onyx-devs for feedback if this is not the desired page experience!

### DIFF
--- a/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
+++ b/src/Apps/Settings/Routes/SavedSearchAlerts/SavedSearchAlertsApp.tsx
@@ -61,7 +61,7 @@ export const SavedSearchAlertsApp: React.FC<SavedSearchAlertsAppProps> = ({
   const [showDeleteModal, setShowDeleteModal] = useState(false)
   const [sort, setSort] = useState("ENABLED_AT_DESC")
   const [loading, setLoading] = useState(false)
-  const alerts = extractNodes(me.alertsConnection)
+  const alerts = extractNodes(me?.alertsConnection)
 
   const xs = __internal__useMatchMedia(THEME.mediaQueries.xs)
   const sm = __internal__useMatchMedia(THEME.mediaQueries.sm)


Mitch-Henson (approved):

olerichter00 (approved):
Thanks 🌟

dblandin:
This seems fine as a quick fix to mitigate the 500 error but ideally this should mimic the behavior of some of our other pages that assume an authenticated user such as:

- https://www.artsy.net/user/conversations
- https://www.artsy.net/favorites/saves
- https://www.artsy.net/settings/purchases

Viewing any of these pages while unauthenticated triggers an authentication prompt and upon successful authentication, navigates you back to your intended destination.

We do not seem to consistently apply that approach to many pages, including `favorites/alerts/:id/edit`, but unless the page has an intentionally designed empty state — I think we should. Otherwise, the logged out experience feels broken.

Here's an example of two pages that properly apply this auth prompt and one that doesn't:


https://github.com/user-attachments/assets/0dd39877-0b87-4b5b-9dd3-e0e5e77ba68f

damassi:
@jpotts244 - _should_ be a 1 liner... https://github.com/artsy/force/blob/main/src/Apps/Settings/settingsRoutes.tsx#L172

jpotts244:
Hero ^ thanks for the breadcrumbs, looking into it now 🍞

PR #14670, opened 2024-10-21
fix(feature): rendering of subheading

anandaroop:

This PR solves [PBRW-102]

### Description

Fixes a display issue related to the `Feature#subheadline` attribute

That field is meant to support markdown and does so correctly in the `FeatureHeaderDefault` component, but not in `FeatureHeaderFull`, which is what we fix in this PR.

This never surfaced before because the current Vanguard feature appears to be the first one with `layout: full` **and** a subheading with markdown formatting.

| Before | After |
|--------|--------|
| ht | md | 

(Example formatting^ in staging only)

[PBRW-102]: https://artsyproduct.atlassian.net/browse/PBRW-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
+++ b/src/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
@@ -39,12 +39,11 @@ export const FeatureHeaderFull: React.FC<FeatureHeaderFullProps> = ({
             </Text>
 
             {subheadline && (
-              <Text
+              <HTML
                 variant={["md", "lg-display"]}
                 color="rgba(255, 255, 255, 0.8)"
-              >
-                {subheadline}
-              </Text>
+                html={subheadline}
+              />
             )}
           </FullBleedHeaderOverlay>
         </FullBleedHeader>


src/Apps/Feature/Components/FeatureHeader/FeatureHeaderFull.tsx
```diff
@@ -39,12 +39,11 @@ export const FeatureHeaderFull: React.FC<FeatureHeaderFullProps> = ({
...
-              <Text
+              <HTML
                 variant={["md", "lg-display"]}
                 color="rgba(255, 255, 255, 0.8)"
-              >
-                {subheadline}
-              </Text>
+                html={subheadline}
+              />
             )}
```
anandaroop, line 47: Simply brings `FeatureHeaderFull` in line with `FeatureHeaderDefault`, while preserving the text color:

https://github.com/artsy/force/blob/9c67452663d06f548714d68155be0f275b3b20ce/src/Apps/Feature/Components/FeatureHeader/FeatureHeaderDefault.tsx#L29-L35

damassi (approved):

PR #14682, opened 2024-10-22
fix: Allow logged out users to exit Sell flow by clicking on Artsy logo

olerichter00:
Addresses [ONYX-1266]

## Description

With this change, we will allow logged-out users to exit the Sell flow by clicking on the Artsy logo in the header. 

The current problem is that clicking on "Save & Exit" opens the "Sign up" modal for logged-out users and does not allow to (simply) exit the Sell flow.

Clicking on the Artsy logo in the header will now open a dialog to exit the flow without signing up and navigate the user to the "/sell" page.

Screenshot 2024-10-22 at 15 37 51

[ONYX-1266]: https://artsyproduct.atlassian.net/browse/ONYX-1266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Sell/Components/SubmissionHeader.tsx
+++ b/src/Apps/Sell/Components/SubmissionHeader.tsx
@@ -1,7 +1,16 @@
 import { AuthIntent, ContextModule } from "@artsy/cohesion"
 import ArtsyLogoIcon from "@artsy/icons/ArtsyLogoIcon"
+import ArtsyMarkIcon from "@artsy/icons/ArtsyMarkIcon"
 import CloseIcon from "@artsy/icons/CloseIcon"
-import { Button, Flex, FullBleed } from "@artsy/palette"
+import {
+  Box,
+  Button,
+  Clickable,
+  Flex,
+  FullBleed,
+  ModalDialog,
+  Text,
+} from "@artsy/palette"
 import { AppContainer } from "Apps/Components/AppContainer"
 import { HorizontalPadding } from "Apps/Components/HorizontalPadding"
 import { useSubmissionTracking } from "Apps/Sell/Hooks/useSubmissionTracking"
@@ -41,6 +50,8 @@ export const SubmissionHeader: React.FC = () => {
   const exitURL = submission?.myCollectionArtworkID
     ? `/my-collection/artwork/${submission?.myCollectionArtworkID}`
     : "/sell"
+
+  const isFirstOrLastStep = !submission?.externalId || isLastStep
 
   const handleSaveAndExit = async () => {
     if (!submission?.externalId) return
@@ -127,7 +138,9 @@ export const SubmissionHeader: React.FC = () => {
                 <Flex
                   flexDirection="row"
                   justifyContent={[
-                    submission?.externalId ? "end" : "space-between",
+                    submission?.externalId && isLoggedIn
+                      ? "end"
+                      : "space-between",
                     "space-between",
                   ]}
                   alignItems="center"
@@ -135,9 +148,7 @@ export const SubmissionHeader: React.FC = () => {
                   pt={[1, 0]}
                   height={HEADER_HEIGHT}
                 >
-                  <Media greaterThan="xs">
-                    <ArtsyLogoIcon display="block" />
-                  </Media>
+                  <HeaderArtsyLogo withExitConfirmation={!isFirstOrLastStep} />
 
                   {submission?.externalId && !isLastStep ? (
                     <>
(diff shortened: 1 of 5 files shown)


src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -127,17 +138,17 @@ export const SubmissionHeader: React.FC = () => {
                 <Flex
                   flexDirection="row"
                   justifyContent={[
-                    submission?.externalId ? "end" : "space-between",
+                    submission?.externalId && isLoggedIn
```
olerichter00, line 141: This change is needed because we will display the logo on mobile for logged out users.

dblandin (commented):
I think would be worth looking into this [WarnBeforeUnload](https://github.com/artsy/volt/blob/255af6805e16c97f8717ddf2c650955a92bbb458/app/javascript/legacy/routes/artworks/edit/ArtworkForm/components/InnerForm/InnerForm.tsx#L161-L163) component in Volt. It's a nice solution because you can drop it within the component tree without needing to nest the behavior within specific navigational components, like the Artsy logo.

The condition to trigger the warn check could be:
- Persisted submission is present (ID in URL)
- Submission is in DRAFT state OR Formik state is "dirty" (unsaved changes)

Seems useful to have this check for either logged out or logged in users, consistently.

I think Artsy logo, if linked, should always link to the root route. Seems odd to have it link to /sell. And if we can port over a version of the WarnBeforeUnload component, then we can keep that link/logo very simple and uncomplicated by state checks.

olerichter00:
@dblandin, we use the `window.addEventListener("beforeunload", this.warnIfDirty)` pattern from Volt's [WarnBeforeUnload](https://github.com/artsy/volt/blob/255af6805e16c97f8717ddf2c650955a92bbb458/app/javascript/legacy/routes/artworks/edit/ArtworkForm/components/InnerForm/InnerForm.tsx#L161-L163) as well in Force ([here](https://github.com/artsy/force/blob/main/src/Apps/Order/OrderApp.tsx#L55)). But unfortunately it's not working with our router. `router.push` doesn't emit the event (probably because the navigation is client-side); only closing the window or reloading the page seems to be guarded with this pattern. 

So I guess, it could be a good idea to add it as an addition, but I couldn't make it work with the clickable Artsy logo in the SWA flow.

src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -199,3 +210,89 @@ export const SubmissionHeader: React.FC = () => {
...
+  )
+
+  const handleClick = () => {
+    withExitConfirmation
+      ? setShowExitConfirmationModal(true)
+      : router.push("/sell")
+  }
+
+  // When logged in, it's possible to easily exit the flow by clicking on "Save & Exit"
+  // and the Artsy logo should not be clickable.
```
dariakoko, line 231: I think, we could consider making the Artsy logo clickable in any case for a more cohesive user experience. For a logged in user after clicking on the Artsy logo we could save the draft submission and inform the user about it by showing a banner or a dialog. A draft submission would be shown on the Sell page (as per current implementation)

src/Apps/Sell/Components/SubmissionHeader.tsx
```diff
@@ -199,3 +210,89 @@ export const SubmissionHeader: React.FC = () => {
...
+    </>
+  )
+}
+
+interface ExitConfirmationModalProps {
+  onClose: () => void
+  onLeave: () => void
+}
+
+export const ExitConfirmationModal: React.FC<ExitConfirmationModalProps> = ({
```
dariakoko, line 270: We already have an `ConfirmationModalBack` component that has the same styling and copies. It is used in MyCollection Artwork Form.
What do you think about reusing it for this flow? It might require some adjustments

olerichter00: That's a very good idea, thanks! The copy is almost the same as well.

src/Apps/Sell/Components/SubmissionHeader.tsx
olerichter00: This could also be an option. I decided against it to provide as few ways as possible to leave the flow because we want users to finish the form. I think this is the reason why we removed the link from the logo in the first place. And there is still the "Exit" or "Save & Exit" button. 

I'm open to do this change but think we should discuss it first with the team if we consider it.

src/Apps/Sell/Components/SubmissionHeader.tsx
olerichter00: I've updated the code to use the already existing component from My Collection 👍

PR #14685, opened 2024-10-22
fix: add safeguards for rendering the insights

egdbear:

This PR solves [PBRW-107]

### Description

This PR adds safeguards for insights, addressing the issue described in the ticket and discussed in this [Slack thread](https://artsy.slack.com/archives/C05EEBNEF71/p1729601303168049).

[PBRW-107]: https://artsyproduct.atlassian.net/browse/PBRW-107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlight.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlight.tsx
@@ -11,6 +11,7 @@ interface ArtistCareerHighlightProps {
 export const ArtistCareerHighlight: FC<ArtistCareerHighlightProps> = ({
   insight,
 }) => {
+  if (!insight?.description && !insight?.entities?.length) return null
   return (
     <Expandable label={insight.label} pb={1}>
       <Description
--- a/src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
+++ b/src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
@@ -22,6 +22,10 @@ interface ArtistCareerHighlightsProps {
 const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
   artist,
 }) => {
+  if (!artist || !artist.insights?.length) {
+    return null
+  }
+
   const insights = artist.insights.slice(ARTIST_HEADER_NUMBER_OF_INSIGHTS)
   const numOfColumns = insights.length > 4 ? 2 : 1
   const mid = Math.ceil(insights.length / 2)
@@ -29,6 +33,10 @@ const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
     numOfColumns === 2
       ? [insights.slice(0, mid), insights.slice(mid, insights.length)]
       : [insights]
+
+  const hasEntitiesOrDescription = artist.insights.some(insight => {
+    return insight.entities.length > 0 || insight.description
+  })
 
   return (
     <>
@@ -55,7 +63,7 @@ const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
         </Media>
       )}
 
-      {insights.length > 0 && (
+      {insights.length > 0 && hasEntitiesOrDescription && (
         <Media greaterThan="xs">
           <Box display="flex" gap={4} flexDirection="column">
             <RailHeader
@@ -96,6 +104,8 @@ export const ArtistCareerHighlightsFragmentContainer = createFragmentContainer(
         insights {
           ...ArtistCareerHighlight_insight
           kind
+          entities
+          description(format: HTML)
         }
       }
     `,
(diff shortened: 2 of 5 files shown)


src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
```diff
@@ -55,7 +59,7 @@ const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
         </Media>
       )}
 
-      {insights.length > 0 && (
+      {insights.length > 0 && hasEntitiesOrDescription && (
```
egdbear, line 62: I wanted to add the specs for this here, but whenever I try to mock the Relay store and artist, I consistently end up with mocked values for the fields. It would be great if this could be handled more effectively.

src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
```diff
@@ -30,6 +30,10 @@ const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
       ? [insights.slice(0, mid), insights.slice(mid, insights.length)]
       : [insights]
 
+  const hasEntitiesOrDescription = artist.insights.some(insight => {
```
damassi, line 33: We'll need to make this a bit more safe!

in the top of the render method, we'll want to look and see if there's no artist or insights. This doesn't exit early on null data:

```
if (!artist) {
  return nulll
}
if (!artist.insights.length) {
  return null
} 
```

Ditto over on ArtistInsight.tsx.

src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlight.tsx
```diff
@@ -11,6 +11,7 @@ interface ArtistCareerHighlightProps {
 export const ArtistCareerHighlight: FC<ArtistCareerHighlightProps> = ({
   insight,
 }) => {
+  if (!insight?.description || !insight?.entities.length) return null
```
damassi, line 14: entities might be null and that's likely the bug, so `!insights?.entities?.length`

src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
```diff
@@ -22,6 +22,10 @@ interface ArtistCareerHighlightsProps {
 const ArtistCareerHighlights: FC<ArtistCareerHighlightsProps> = ({
   artist,
 }) => {
+  if (!artist || !artist.insights.length) {
```
damassi, line 25: !artist?.insights?.length

damassi: if there's an issue with the data fidelity on `insights` we'd get a runtime error trying to guard for the error :)

egdbear: yes!!

src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlight.tsx
egdbear: yes!! 🦺

damassi (approved):
Thank you 👍

src/Apps/Artist/Routes/Overview/Components/ArtistCareerHighlights.tsx
```diff
@@ -96,6 +104,8 @@ export const ArtistCareerHighlightsFragmentContainer = createFragmentContainer(
         insights {
           ...ArtistCareerHighlight_insight
           kind
+          entities
```
mzikherman, line 107: What does adding these do? I think the issue is that, `insights` is totally `null` initially - then it gets 'partially' created b/c a refetch happens - and since `insights` is now not `null`, the component tries to render something, which then crashes because `entities` or other parts of that component's fragment are missing.

So I think the fix is to be ultra-defensive in the component, assume the fragment might be 'missing', so even though `entities` is not-null, need to check for its presence, and so forth.

I _think_ you're doing that above - not sure this is needed?

egdbear: This comes from fragment. If I don't have them I'll have the ts error.

image

mzikherman: Yea, I'm saying I don't think you need this check. The defensive check, as you've already done, should be in the child/individual component, not this list one. Let this list one have `insights` that is partial.

It's a bit of _over_ defensiveness that also invalidates the CDN-cached query that's making this issue apparent by modifying the query - so the potential for a 'false positive' is there - https://artsy.slack.com/archives/C05EEBNEF71/p1729610632196429?thread_ts=1729601303.168049&cid=C05EEBNEF71

egdbear: I think we want to keep this check here (doch 😅 ) because if we encounter a glitched insight (such as one with relay references), the child component won't render due to the safeguard. However, in this component, we would still display the title on this [line](https://github.com/artsy/force/pull/14685/files#diff-5eaef4e9cc9c1a503eb96404aba57af63430f665267b50214066bca229acefb6R70).

PR #14686, opened 2024-10-22
fix: pass through data-test* properties

joeyAghion:
Fixes integration tests, which currently fail in multiple suites expecting `data-testid` attributes. A search indicates that `data-test` is another attribute specified throughout the codebase, and probably gets removed before generating the DOM.


araujobarret (approved):

damassi (approved):
Thanks @joeyAghion

PR #14688, opened 2024-10-22
ci(slack): fix slack notifications

damassi:

### Description

Noticed that slack deploy failure alerts stopped working with the latest CI refactor. Not sure why! So hard code the value in here and see if it works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,11 +122,6 @@ jobs:
 
   validate_production_schema:
     executor: node-executor
-    parameters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -339,7 +334,7 @@ workflows:
                 - staging
 
       - validate_production_schema:
-          <<: *only_release
+          <<: *not_staging_or_release
```
damassi, line 337: temporary, will revert this in the next commit if its working properly

PR #14709, opened 2024-10-23
fix: use backend field to determine if an order is already with payment method set

starsirius:

This PR solves [EMI-2085]

### Description

When the payment method (e.g. credit card) of an order is later deleted, visiting the order page would lead to infinite redirection. The [redirect rule](https://github.com/artsy/force/blob/994c500db249b75c7b0f83958b0588962661ba03/src/Apps/Order/redirects.tsx#L63) fetches the payment method to determine if payment is already set. In a deleted payment method case, it returns null and the rule would incorrectly treat it as payment not yet set.

We don't actually need to fetch the payment method object again in the UI. Instead, it's better to [move the logic to the backend](https://github.com/artsy/exchange/pull/2270), and we can simply respect the boolean field returned.

This happens more often in staging, because payment method data were wiped out each week.

<details><summary>Before screen recording</summary>
<p>

infinite-redirection-before

</p>
</details> 

<details><summary>After screen recording</summary>
<p>

infinite-redirection-after

</p>
</details> 

[EMI-2085]: https://artsyproduct.atlassian.net/browse/EMI-2085?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Order/Utils/orderUtils.ts
```diff
@@ -10,27 +10,12 @@ const initialPaymentMethods: CommercePaymentMethodEnum[] = [
   "WIRE_TRANSFER",
 ]
 
-export const isPaymentSet = (
```
starsirius, line 13: Retire the entire utility to use the field returned from Exchange. This refactor is in [a63a36b](https://github.com/artsy/force/pull/14709/commits/a63a36b863f0386722729983ae8c570679cb3209).

PR #14743, opened 2024-10-28
chore(performance): Attempt to improve LCP by disabling lazyload on Artwork/Artist

damassi:

### Description

Removes the bot bits on here: https://github.com/artsy/force/pull/14708 and sets lazyload explicitly to `false` to improve LCP.


src/Apps/Artwork/Components/ArtworkLightboxPlaceholder.tsx
```diff
@@ -44,6 +44,7 @@ export const ArtworkLightboxPlaceholder: React.FC<ArtworkLightboxPlaceholderProp
           height="100%"
           alt="Loading"
           lazyLoad={lazyLoad}
+          style={{ filter: "blur(10px)" }}
```
damassi, line 47: @joeyAghion - check out the comment in this file. I think its safe to keep this since we're now disabling lazyload here. It looks a lot better with it on.

joeyAghion (approved):

PR #14744, opened 2024-10-28
chore(EMI-2094): Dont show the end times banner if the sale is ended

erikdstock:

This PR solves [EMI-2094]

### Description

This makes us not show the 'lots may close later due to cascading end times' banner if the sale has ended.

I chose to test for presence of `sale.endedAt` directly rather than one of the other options [here](https://github.com/artsy/metaphysics/blob/c6d866aba033fbddb3ef803e0a46e1dfdbdb5636/src/schema/v2/sale/index.ts#L372-L399) (`sale.isOpen`, `isClosed`, `isLotsClosing` etc). 

[EMI-2094]: https://artsyproduct.atlassian.net/browse/EMI-2094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/CascadingEndTimesBanner.tsx
```diff
@@ -13,7 +13,9 @@ const CascadingEndTimesBanner: React.FC<CascadingEndTimesBannerProps> = ({
 }) => {
   const helpArticleLink = getENV("CASCADING_AUCTION_HELP_ARTICLE_LINK")
 
-  if (!sale.cascadingEndTimeIntervalMinutes) {
+  const saleIsClosed = !!sale.endedAt
```
damassi, line 16: If there's a boolean `isClosed` field, why opt for coercing a date field `endedAt` and then aliasing the coerced value to local variable name representing something we can just return from the backend? (1:2)

damassi: If we can't trust `isClosed` (which is used all over many of our codebases if one does a quick search) then there are larger problems! Its minor, but still something to take to heart as a pattern to avoid.

PR #14745, opened 2024-10-28
chore: remove deep link dep

brainbicycle:

### Description

Removes deep link dependency now that we are serving this file via cloudflare, see here: https://github.com/artsy/artsy-eigen-web-association/pull/58


damassi (approved):

MrSltun (approved):
🧹🧹🧹

PR #14756, opened 2024-10-30
refactor(tests): Migrate legacy enzyme tests to .jest.enzyme extension

damassi:

### Description

PR 1 of 2: Since Enzyme is no longer supported in React 18, we need to segment our legacy tests from our primary test runner. 

To do this, we 
- Search for all tests that contain `from "enzyme"` and rename the file to `test.jest.enzyme.ts`
- Add new jest test config (`jest.legacy.config`) that matches this new extension 
- Add a new package.json script for devs to run `yarn jest-enzyme` against old files 
- Update CI config with a new job `test-legacy`

One surprising positive I discovered here is that only 1/6th of our tests use Enzyme! That's quite amazing tbh (and good work team ❇️). 

The next PR will create a new aliased lib import similar to [how we do it in volt](https://github.com/artsy/volt/blob/main/package.json#L130) that specifically targets React 17 and then updates our config to execute legacy tests against that.


package.json
```diff
@@ -36,7 +36,7 @@
     "docker:clean": "docker rmi force force:builder force:electron",
     "jest": "node_modules/.bin/jest --config jest.config.js",
     "jest:debug": "node --inspect node_modules/.bin/jest --runInBand",
-    "jest:terse": "TERSE_RTL_OUTPUT=true jest",
+    "jest-enzyme": "node_modules/.bin/jest --config jest.legacy.config.js",
```
anandaroop, line 39: **question:** Intentionally dropping the `terse` package script?

anandaroop (approved):
LGTM, just one q

Also I think that Peril warning is a false alarm based on [this recent PR](https://github.com/artsy/force/pull/14745), yeah?

package.json
damassi: Yah, that was a silly "me problem" 🤦 And don't know what i was thinking there

damassi: Ie: 99% of the time its helpful. I ran into the 1% and caused a lil fuss and having this extra script just complicates things.

damassi:
Ah! re that dep update -- lemme rebase.

PR #14759, opened 2024-10-30
chore(tests): Split react versions for enzyme tests

damassi:

Addresses [DIA-951]

### Description

This adds the setup pointers for using an explicit `react-17` version for enzyme tests. Don't have React 18 in the app yet, but this finishes the prework on the test setup side. 

[DIA-951]: https://artsyproduct.atlassian.net/browse/DIA-951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -166,8 +163,10 @@
     "qrcode.react": "1.0.0",
     "qs": "6.5.3",
     "react": "17.0.2",
+    "react-17": "npm:react@17.0.2",
     "react-dfp": "^0.21.0",
     "react-dom": "17.0.2",
+    "react-dom-17": "npm:react-dom@17.0.2",
```
anandaroop, line 169: **question (non-blocking):** What's the deal with the `npm:` syntax? That's how you declare an alias to a second version of the same lib?

jest.legacy.config.js
```diff
@@ -7,7 +7,9 @@ module.exports = {
   moduleFileExtensions: ["js", "json", "jsx", "ts", "tsx"],
   moduleNameMapper: {
     "^luxon$": "<rootDir>/node_modules/luxon",
-    "^react$": "<rootDir>/node_modules/react",
+    "^react$": "react-17",
+    "^react-dom$": "react-dom-17",
+    "^react-dom/test-utils$": "react-dom-17/test-utils",
```
anandaroop, line 12: **question (non-blocking):** And here we use the alias name only, not the full `node_modules` path?

anandaroop (approved):
👍🏽 

Just some noob question re the funky syntax…

package.json
damassi: Yup! Its very useful, assuming you can remap the internals via some tool like jest

jest.legacy.config.js
damassi: Yup, when jest runs its `transform` over the code (via `babel-jest`) it will preprocess the code and swap out refs before execution.

src/Components/Address/__tests__/useAddressAutocomplete.jest.ts
```diff
@@ -96,7 +96,7 @@ describe("useAddressAutocomplete", () => {
           })
           const { result } = setupHook({ country: "US" }, hookOptions)
 
-          await act(() => {
```
damassi, line 99: not sure why these are suddenly failing, but in any case, this sort of invocation on act is incorrect! So better is better

PR #14774, opened 2024-11-03
feat(web-vitals): Better DX around web-vitals debugging

damassi:

### Description

Been getting very annoyed having to use lighthouse and other 3rd party tools for debugging web vitals stuff, and it turns out google released a package called [web-vitals](https://github.com/GoogleChrome/web-vitals) which we can use to tap into all of those underlying stats. So this wires that up and makes things much easier to debug! 

Can enable via ENV variable:
- `ENABLE_WEB_VITALS_LOGGING` 

Examples:

Screenshot 2024-11-02 at 9 00 21 PM

For areas that we need to dig into, it emits helpful info (like offending DOM nodes) which you can hover right over and it will select on the page:

Screenshot 2024-11-02 at 9 05 39 PM

cc @artsy/diamond-devs


damassi:
(Going to merge this and turn on on staging 🚶)

PR #14782, opened 2024-11-04
fix(preconnect): Remove unnecessary crossorigin attributes

damassi:

### Description

Been looking at this nicer (realtime) page analysis tool ([debugbear](https://www.debugbear.com/test/website-speed/ZCsVHyUb/overview?info=errors&metric=hints.errors)) and noticed that it returned a `preconnect` crossorigin error related to our cdn. Since this is public (as with our other cdn urls) I went ahead and removed that flag to address this error:

Screenshot 2024-11-03 at 4 11 29 PM

Not sure if i'm missing anything here but seems fine. 

Also flagged our prefetch hint as not being high priority (not an actual image, but...?), so added the attribute to fix that too.


damassi:
Actually, digging around further, advice here is ambiguous.

PR #14787, opened 2024-11-04
feat(performance): Add PreloadWebpackPlugin for preloading initial JS chunks and simplify html pipeline

damassi:

This PR solves [DIA-969]

### Description

This adds support for preload hints around initial JS chunks that are shipped down to the page via https://github.com/vuejs/preload-webpack-plugin. 

Idea first came from reading [this blogpost](https://mmazzarolo.com/blog/2024-08-13-async-chunk-preloading-on-load) and dug around a bit and realized that webpack can do this all for us. 

Example output:

Screenshot 2024-11-04 at 12 33 35 PM

Also refactored / simplified our html setup by not manually defining JS chunk output in our html template, and instead let HTMLWebpackPlugin do the work it was originally designed to do 🤷 

cc @artsy/diamond-devs 

[DIA-969]: https://artsyproduct.atlassian.net/browse/DIA-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webpack/sharedPlugins.js
+++ b/webpack/sharedPlugins.js
@@ -1,6 +1,7 @@
 // @ts-check
 
 import { RetryChunkLoadPlugin } from "webpack-retry-chunk-load-plugin"
+import PreloadWebpackPlugin from "@vue/preload-webpack-plugin"
 import NodePolyfillPlugin from "node-polyfill-webpack-plugin"
 import webpack from "webpack"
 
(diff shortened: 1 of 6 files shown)


webpack/sharedPlugins.js
```diff
@@ -26,4 +27,10 @@ export const sharedPlugins = () => [
       return "cache-bust=" + Date.now();
     }`,
   }),
+
+  new PreloadWebpackPlugin({
+    rel: "preload",
+    as: "script",
+    include: "initial",
```
damassi, line 34: Initial = just the scripts needed to render whatever initial landing page the user first requests. 

Looking at the output, its a lot! Really need to take a deep dive into our JS layer.

damassi:
Gonna merge to get this to staging. Happy to follow up with any change requests!

PR #14793, opened 2024-11-04
feat(performance): Add PreloadWebpackPlugin for preloading initial JS chunks -- Review App 

damassi:
### Description

Opening a review app for preloading JS chunks: https://github.com/artsy/force/pull/14787 (PR merged and reverted).

The issue is we were double embedding our script tags for some reason. Removed the second call, and all good.

### DIFF
--- a/webpack/envs/clientProductionConfig.js
+++ b/webpack/envs/clientProductionConfig.js
@@ -65,11 +65,11 @@ const clientProductionConfig = () => {
         template: path.resolve(basePath, "src/html.ejs"),
         inject: true,
         scriptLoading: "defer",
-        minify: {
-          collapseWhitespace: true,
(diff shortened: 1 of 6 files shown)


webpack/envs/clientProductionConfig.js
```diff
@@ -62,13 +62,14 @@ const clientProductionConfig = () => {
...
-          removeComments: true,
-        },
         template: path.resolve(basePath, "src/html.ejs"),
+        inject: true,
+        scriptLoading: "defer",
+        // minify: {
+        //   collapseWhitespace: true,
+        //   conservativeCollapse: true,
+        //   removeComments: true,
+        // },
```
damassi, line 72: Oops - fixed in https://github.com/artsy/force/pull/14795

PR #14794, opened 2024-11-04
feat: add Link headers corresponding to preconnect and preload instructions (DIA-967)

joeyAghion:
This PR adds `Link` headers corresponding to the `preconnect` and `preload` declarations already made [at the top of our HTML template](https://github.com/artsy/force/blob/7fc5ed5cb75c865cc0b146072e26258e2772d10c/src/html.ejs#L9-L17).

Once these are part of Force's response, we should be able to toggle on [Cloudflare's intelligent "early hints" handling](https://developers.cloudflare.com/cache/advanced-configuration/early-hints/). Supposedly, `103` responses will automatically be streamed to compatible clients before the server's response is available.

[This thread](https://webmasters.stackexchange.com/questions/89466/when-should-i-use-the-crossorigin-attribute-on-a-preconnect-link) was helpful in understanding the appropriate `crossorigin` settings.

(Very open to feedback about how to better reuse the preconnect/preload logic from elsewhere.)

Once this is ready to merge, I'd suggest we first release on its own, then enable early hints once we're satisfied that there's no adverse impact.

https://artsyproduct.atlassian.net/browse/DIA-967

### DIFF
--- a/src/Server/middleware/linkHeadersMiddleware.ts
+++ b/src/Server/middleware/linkHeadersMiddleware.ts
@@ -22,7 +22,7 @@ export function linkHeadersMiddleware(
       `<${WEBFONT_URL}/ll-unica77_regular.woff2>; rel=preload; as=font; crossorigin`,
       `<${WEBFONT_URL}/ll-unica77_medium.woff2>; rel=preload; as=font; crossorigin`,
       `<${WEBFONT_URL}/ll-unica77_italic.woff2>; rel=preload; as=font; crossorigin`,
-    ] as any)
+    ])
   }
   next()
 }
--- a/src/Server/middleware/linkHeadersMiddleware.ts
+++ b/src/Server/middleware/linkHeadersMiddleware.ts
@@ -1,0 +1,28 @@
+import type { NextFunction } from "express"
+import type { ArtsyRequest, ArtsyResponse } from "./artsyExpress"
+
+import { CDN_URL, GEMINI_CLOUDFRONT_URL, WEBFONT_URL } from "Server/config"
+
+/**
+ * Link headers allow 103: Early Hints to be sent to the client (by Cloudflare).
+ * These should correspond to preconnect and preload items from html.ejs.
+ * See: https://developers.cloudflare.com/cache/advanced-configuration/early-hints/
+ */
+export function linkHeadersMiddleware(
+  req: ArtsyRequest,
+  res: ArtsyResponse,
+  next: NextFunction
+) {
+  if (!res.headersSent) {
+    res.header("Link", [
+      `<${CDN_URL}>; rel=preconnect; crossorigin`,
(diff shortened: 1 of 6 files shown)


src/Server/middleware/linkHeadersMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!res.headersSent) {
+    res.header("Link", [
+      `<${CDN_URL}>; rel=preconnect; crossorigin`,
+      `<${GEMINI_CLOUDFRONT_URL}>; rel=preconnect; crossorigin`,
+      `<${WEBFONT_URL}>; rel=preconnect; crossorigin`,
+      `<${WEBFONT_URL}/all-webfonts.css>; rel=preload; as=style`,
+      `<${WEBFONT_URL}/ll-unica77_regular.woff2>; rel=preload; as=font; crossorigin`,
+      `<${WEBFONT_URL}/ll-unica77_medium.woff2>; rel=preload; as=font; crossorigin`,
+      `<${WEBFONT_URL}/ll-unica77_italic.woff2>; rel=preload; as=font; crossorigin`,
+    ] as any)
```
joeyAghion, line 25: The `header` or `set` methods _do_ in fact accept an array. I attempted to update the types:

```bash
yarn add -D @types/express@4.17.21
```

...but that encountered many more errors of the form:

```
The inferred type of 'adminServerRoutes' cannot be named without a reference to 'express/node_modules/@types/express-serve-static-core'. This is likely not portable. A type annotation is necessary.
```

damassi: Lemme try locally real quick; this could likely be a quick fix (though any is totally fine here). One moment...

damassi: Pushed two small commits!

src/Server/middleware/linkHeadersMiddleware.ts
```diff
@@ -0,0 +1,28 @@
...
+      `<${CDN_URL}>; rel=preconnect; crossorigin`,
+      `<${GEMINI_CLOUDFRONT_URL}>; rel=preconnect; crossorigin`,
+      `<${WEBFONT_URL}>; rel=preconnect; crossorigin`,
+      `<${WEBFONT_URL}/all-webfonts.css>; rel=preload; as=style`,
+      `<${WEBFONT_URL}/ll-unica77_regular.woff2>; rel=preload; as=font; crossorigin`,
+      `<${WEBFONT_URL}/ll-unica77_medium.woff2>; rel=preload; as=font; crossorigin`,
+      `<${WEBFONT_URL}/ll-unica77_italic.woff2>; rel=preload; as=font; crossorigin`,
+    ])
+  }
+  next()
```
dzucconi, line 27: Very interesting. Had no idea this existed

dzucconi (approved):
Very cool

PR #14796, opened 2024-11-05
feat(early-hints): Add support for emitting webpack .js early hint headers and passing via linkHeadersMiddleware

damassi:

This PR solves [DIA-972]

### Description

Building on work completed in: 
- https://github.com/artsy/force/pull/14794 

This adds a new Webpack plugin for generating a list of "initial" webpack chunks for `early-hints` (`initial` being, those assets that are required on every page to run the site). On the express side we read the list and then expand the headers that are being passed down via the new `linkHeadersMiddleware`. With this now in place we can take further advantage of the new `<link rel="preload" as="script">` tags that we're now writing to our html template on build. 

**Screenshots:**

1. Can now see that we've got a high fetch priority on our main scripts: 

Screenshot 2024-11-05 at 8 51 59 AM

2. And here are the script link headers as passed over from the server:

Screenshot 2024-11-05 at 9 09 52 AM

cc @artsy/diamond-devs 

[DIA-972]: https://artsyproduct.atlassian.net/browse/DIA-972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/html.ejs
+++ b/src/html.ejs
@@ -36,6 +36,8 @@
     <%- htmlWebpackPlugin.tags.headTags.map((originalTag) => { const tag = { ...originalTag, attributes: { ...originalTag.attributes } }; tag.attributes['src'] = "<" + "%= cdnUrl %" + ">" + originalTag.attributes['src']; return tag; }) %>
   <%% } %>
 
+  <%%- content.linkPreloadTags.join("") %>
+
   <%%- content.head %>
   <%%- content.style %>
   <%%- content.data %>
--- a/package.json
+++ b/package.json
@@ -275,7 +275,6 @@
     "@types/yup": "^0.29.13",
     "@typescript-eslint/eslint-plugin": "2.30.0",
     "@typescript-eslint/parser": "4.18.0",
-    "@vue/preload-webpack-plugin": "^2.0.0",
     "@wojtekmaj/enzyme-adapter-react-17": "0.6.6",
     "babel-jest": "^29.5.0",
     "babel-loader": "8.2.3",
--- a/src/Server/__tests__/getWebpackEarlyHints.jest.ts
+++ b/src/Server/__tests__/getWebpackEarlyHints.jest.ts
@@ -1,0 +1,54 @@
+import fs from "fs"
+import path from "path"
+import { getWebpackEalyHints } from "Server/getWebpackEarlyHints"
+
+jest.mock("fs")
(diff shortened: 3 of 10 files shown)


src/html.ejs
```diff
@@ -36,6 +36,8 @@
     <%- htmlWebpackPlugin.tags.headTags.map((originalTag) => { const tag = { ...originalTag, attributes: { ...originalTag.attributes } }; tag.attributes['src'] = "<" + "%= cdnUrl %" + ">" + originalTag.attributes['src']; return tag; }) %>
   <%% } %>
 
+  <%%- content.linkPreloadTags.join("") %>
```
damassi, line 39: Since https://github.com/artsy/force/pull/14793 didn't allow for us to insert our CDN url, we ended up having to do this manually, which is fine!

damassi:
Not sure whats going on with smoke tests -- everything looks good locally. Investigating 👀

src/html.ejs
joeyAghion: When I try this locally, I see a slew of new `<link rel="preload"` entries outputted here, similar to but different from the `<script defer` tags outputted just above. I'm not totally sure how this should work, but do we need both? They're basically opposites.

Locally I'm not sure if my CDN_URL is set up correctly but I see 2 separate requests for e.g. `runtime.*.js`, one of which is a CORS error.
Screenshot 2024-11-05 at 2 38 26 PM

joeyAghion (commented):
I've tried several webpagetest runs against production, and from what I can tell most of these js bundles aren't critical to the Largest Contentful Paint timeline. The `runtime.*.js` one usually appears in the waterfall, but others don't. My fear is that preloading many many (47 in my local tests of artist pages) js files will actually delay loading critical assets like the main image.

src/html.ejs
damassi: Script and link tags are different things. The link tags are hints to preload.

You can't use CDN_URL's locally during dev, it will error out. It'll work on staging.

damassi:
Chatted in [slack](https://artsy.slack.com/archives/C05EEBNEF71/p1730826661725229) about concerns and have agreed to launch this and monitor as an **experiment**. Waiting on a calibre run to complete before sending to prod, however. 

Merging to get to staging!

PR #14800, opened 2024-11-05
fix: Update the term on the search page correctly

olerichter00:
Resolves [ONYX-1384](https://artsyproduct.atlassian.net/browse/ONYX-1384)

## Description

Entering a new search term on the search page does not always update the text on the page. This PR fixes the issue.

Screenshot 2024-11-05 at 11 42 12

**Steps to reproduce:**

Type "richter" in search input, press enter
Type "warhol" in search input, press enter
Type "richter" in search input, press enter

Outcome before this fix:

`... results for "warhol"` is still displayed

Expected outcome:

`... results for "richter"` should be displayed

https://github.com/user-attachments/assets/d1bb780c-1b1a-4c4a-92d3-4dcc363e8025

[ONYX-1384]: https://artsyproduct.atlassian.net/browse/ONYX-1384?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Search/SearchApp.tsx
+++ b/src/Apps/Search/SearchApp.tsx
@@ -1,4 +1,3 @@
-import React from "react"
 import {
   Box,
   FullBleed,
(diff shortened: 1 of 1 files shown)


src/Apps/Search/SearchApp.tsx
```diff
@@ -50,7 +49,7 @@ export const SearchApp: React.FC<SearchAppProps> = ({ viewer, children }) => {
   const { query } = location
   const { aggregations } = searchConnection ?? {}
 
-  const term = usePrevious(query.term ?? "")
```
olerichter00, line 53: `usePrevious` was introduced in https://github.com/artsy/force/pull/12640/files#diff-ee3b179a5ab057d764ecdac38ef50ce5692a5d1ea9ec9de5096fa8199870908aR55 and causes the missing update of the search term.

anandaroop: Interestingly the behavior added in #12640 (updating the input when navigating through the dropdown options) does not seem to exist anymore, either before or after this change as far as I can tell.

Wondering about that – but this is a clear improvement over the currently glitchy behavior, so 👍🏽 from me

anandaroop (approved):

dzucconi (approved):

olerichter00:
> Interestingly the behavior added in https://github.com/artsy/force/pull/12640 (updating the input when navigating through the dropdown options) does not seem to exist anymore, either before or after this change as far as I can tell.

@anandaroop, good catch! I've opened a thread in [Slack](https://artsy.slack.com/archives/C05EQL4R5N0/p1731501184044299) to discuss how to fix or bring back the behavior.

PR #14803, opened 2024-11-05
fix: Require logged in user for  all`/favorites/alerts/` sub route

olerichter00:
Addresses [ONYX-1248]

## Description

Require logged-in user for `/favorites/alerts/$id/edit` route instead of redirecting to `/`.

Screenshot 2024-11-05 at 16 41 31

[ONYX-1248]: https://artsyproduct.atlassian.net/browse/ONYX-1248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Favorites/favoritesRoutes.tsx
+++ b/src/Apps/Favorites/favoritesRoutes.tsx
@@ -87,11 +87,6 @@ export const favoritesRoutes: RouteProps[] = [
         onClientSideRender: () => {
           Alerts.preload()
         },
-        onServerSideRender: ({ req, res }) => {
-          if (!req.user) {
-            res.redirect("/")
-          }
-        },
         query: graphql`
           query favoritesRoutesAlertsAppEditQuery {
             me {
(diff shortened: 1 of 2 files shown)


src/Apps/Favorites/favoritesRoutes.tsx
```diff
@@ -87,11 +87,6 @@ export const favoritesRoutes: RouteProps[] = [
         onClientSideRender: () => {
           Alerts.preload()
         },
-        onServerSideRender: ({ req, res }) => {
-          if (!req.user) {
-            res.redirect("/")
-          }
-        },
```
olerichter00, line 94: There is no need to redirect anymore because we show the "Log in" dialog.

dblandin (approved):
Awesome 👍🏼

PR #14816, opened 2024-11-07
fix: Correctly unset availability filter

olerichter00:
Addresses [ONYX-699]

## Description

Currently, unchecking the "Only works for sale" filter on the artwork grid does not remove the filter; instead, it filters for artworks that are not for sale.

With this change, the filter will be **removed** correctly when unchecked.

----

In https://github.com/artsy/force/pull/14817, I'm making sure the filter does not appear in the URL as `for_sale=`.

Screenshot 2024-11-07 at 15 50 24

[ONYX-699]: https://artsyproduct.atlassian.net/browse/ONYX-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/ArtworkFilter/ArtworkFilters/AvailabilityFilter.tsx
+++ b/src/Components/ArtworkFilter/ArtworkFilters/AvailabilityFilter.tsx
@@ -11,8 +11,10 @@ interface AvailabilityFilterProps {
   expanded?: boolean
 }
 
-export const AvailabilityFilter: React.FC<AvailabilityFilterProps> = ({ expanded }) => {
-  const { setFilter } = useArtworkFilterContext()
+export const AvailabilityFilter: React.FC<AvailabilityFilterProps> = ({
+  expanded,
+}) => {
+  const { unsetFilter, setFilter } = useArtworkFilterContext()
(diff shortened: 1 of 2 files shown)


anandaroop (approved):
Nice 🎉 

One correction though…

> Currently, unchecking the "Only works for sale" filter on the artwork grid does not remove the filter; instead, it filters for artworks that are not for sale.

I don't think that's quite right.

After some back and forth earlier this year we ended up with a Gravity change that produced the desired filtering behavior: https://github.com/artsy/gravity/pull/17446

So the filtering _behavior_ should already be good now, but the URL state is what needed some cleanup. 

Looks like this PR addresses that, but we won't actually see the desired URL behavior until #14817, yes?

olerichter00:
> Looks like this PR addresses that, but we won't actually see the desired URL behavior until https://github.com/artsy/force/pull/14817, yes?

Thanks for the correction! You're right; this PR is pretty much only a preparation for https://github.com/artsy/force/pull/14817, which will bring the desired update.

PR #14821, opened 2024-11-07
fix: Disable prefetching artworks on the Notification page

olerichter00:
## Description

Somehow, prefetching the `artworkRoutes_ArtworkQuery` query seems to cause this error on the Notification page. It looks like prefetching the query causes the removal of the data in the Relay store (maybe by overriding or [garbage collecting](https://relay.dev/docs/guided-tour/accessing-data-without-react/retaining-queries/) the data 🤔).

**Internal error after a few milliseconds:**

Screenshot 2024-11-07 at 18 51 35

**Artwork is visible and disappears after a few milliseconds with this error:**

Screenshot 2024-11-07 at 18 55 23

### DIFF
--- a/src/Components/Notifications/NotificationArtwork.tsx
+++ b/src/Components/Notifications/NotificationArtwork.tsx
@@ -1,14 +1,14 @@
+import { AuthContextModule } from "@artsy/cohesion"
+import { Box, Button, Image } from "@artsy/palette"
+import { NotificationArtwork_artwork$key } from "__generated__/NotificationArtwork_artwork.graphql"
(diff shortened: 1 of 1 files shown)


damassi (approved):
Thank you!

damassi:
@olerichter00 - did you find that this actually fixes things?

olerichter00:
I thought it had fixed both issues from the description, but now I'm seeing the same error (just not caused by `usePrefetchRoute` again)...

olerichter00:
@damassi, this PR does not disable all prefetching that is happening on the notification page. 

I've opened a follow-up PR that disables prefetching on some other links on the page: https://github.com/artsy/force/pull/14822

PR #14822, opened 2024-11-07
fix: Disable more prefetching on the Notification page

olerichter00:
## Description

Disable more prefetching on the Notification page.


damassi (approved):
Get itttt

PR #14827, opened 2024-11-07
[Do Not Merge] => feat(early-hints): Add support for image early hint link tags

damassi:

### Description

Add support for image early hints (in addition to js, and other preconnects).

Additionally, refactors things slightly to remove the middleware and instead move the header assignment into renderServerApp, where its near our HTML generation. 
 
Screenshot 2024-11-07 at 2 25 22 PM

cc @artsy/diamond-devs


src/System/Router/renderServerApp.tsx
```diff
@@ -80,7 +84,51 @@ export const renderServerApp = ({
...
+  imageLinkHeaders,
+  scriptLinkHeaders,
+  res,
+}: SetEarlyHintsResponseHeadersProps) => {
+  try {
+    if (!res.headersSent) {
+      res.header("Link", [
+        `<${CDN_URL}>; rel=preconnect; crossorigin`,
+        `<${GEMINI_CLOUDFRONT_URL}>; rel=preconnect; crossorigin`,
+        `<${WEBFONT_URL}>; rel=preconnect; crossorigin`,
```
joeyAghion, line 122: Thought: Now that we're preloading js assets and fonts and a main image, I wonder if we even need these preconnects (that likely refer to the same domains). We can try removing them later.

src/html.ejs
```diff
@@ -36,7 +36,8 @@
     <%- htmlWebpackPlugin.tags.headTags.map((originalTag) => { const tag = { ...originalTag, attributes: { ...originalTag.attributes } }; tag.attributes['src'] = "<" + "%= cdnUrl %" + ">" + originalTag.attributes['src']; return tag; }) %>
   <%% } %>
 
-  <%%- content.linkPreloadTags.join("") %>
+  <%%- content.imageLinkPreloadTags.join("") %>
```
joeyAghion, line 39: Won't this duplicate the `<link>` elements rendered within the body ([e.g.](https://github.com/artsy/force/blob/7ec3ee23f5b3c11582d92bf9ea5494525db18ee9/src/Apps/Artist/Components/ArtistHeader/ArtistHeaderImage.tsx#L36))?

damassi: Ah yep, i was thinking we can remove this stuff now since its now all automated.

src/System/Router/renderServerApp.tsx
```diff
@@ -44,15 +44,19 @@ export const renderServerApp = ({
 
   const { WEBFONT_URL } = sharify.data
```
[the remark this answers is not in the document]
damassi: @dzucconi has a semi-mysterious comment up there so i didn't want to mess with any of those env imports

src/html.ejs
damassi: fixed

joeyAghion:
If I understand correctly, this prescribes that components simply render `fetchpriority` attributes on ``s, and that will _automatically_ render `<link>` elements in the header and add `Link: ...` headers on the response.
* If so, there are a few other places that render `<link>` elements explicitly. Should we drop those and make sure the ``s have `fetchpriority` instead?
* The regular expression seems to match any `fetchpriority` value, but I think it should only match `fetchpriority="high"`. [Our code sometimes](https://github.com/search?q=repo%3Aartsy%2Fforce%20fetchpriority&type=code) outputs `fetchpriority="auto"`

damassi:
Bailing on this. Here's a bit of a papertrail from [Slack](https://artsy.slack.com/archives/C05EEBNEF71/p1731339117647579):

The issue is that with this in place, I’m noticing that we’re now overfetching. Meaning, right now, if there’s say the artist image, with fetchPriority="high" , by adding the hint and preload tag automatically, we end up loading a few versions of it if checking the network inspector. This applies to all of the different pages.

I thought it might have something to do with the URLs coming back and formatted differently (encoding or something?) but when I started messing with that, I noticed that it would kick things down to low priority, and couldn’t track down high.

I also had a version that extracted and formated imagessrcset, but that would break things and for some reason wouldn’t early hint.

So in short, i’m running into edge-cases that I can’t figure out and so i’m gonna bail on it for now and let our current results simmer a bit.

If you’ve got some time maybe give it a shot? Perhaps i’m missing something obvious.

QA:
- tweak regex
-load page, filter network panel by image, make sure to show the priority column.
- notice the number of identical images loading
- check headers / image urls
- and so on.

damassi:
Closing for now. We can reopen if we feel the need to optimize further, but so far `fetchPriority="high"` _seems_ to be working good.

PR #14842, opened 2024-11-12
chore: avoid loading extensive @artsy libraries until necessary

joeyAghion:
Based on Chrome's coverage tooling, @dzucconi and I identified that the `artsy-framework` bundle was a little too liberal in the modules it included. Because that bundle is `preload`ed, that meant that _all_ of the referenced code was being loaded on all pages, and with a high priority.

* `@artsy/to-title-case` appears unused, so we removed it.
* `@artsy/icons` is designed for icons to be imported individually, and [occasionally](https://github.com/artsy/icons/pull/54) gains a bloated image, so is excluded.
* Modules like `@artsy/xapp` and `/multienv` should only be required server-side.
* Other libs like `@artsy/commerce_helpers`, `/dismissible`, `/express-reloadable`, `/palette-charts`, and `react-html-parser` should only be needed in a few places, so load them there.
* Q: What about `@artsy/cohesion`? It's needed for almost every route, but it also changes frequently so invalidates caching of these other libraries (which are more stable). Maybe we should add it back into this list.

Co-authored by: @dzucconi

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,6 @@
     "@artsy/palette": "^39.0.4",
     "@artsy/palette-charts": "^38.0.4",
     "@artsy/react-html-parser": "^3.0.2",
-    "@artsy/to-title-case": "^1.1.0",
     "@artsy/xapp": "2.0.0",
     "@babel/runtime": "7.13.10",
     "@loadable/component": "5.15.2",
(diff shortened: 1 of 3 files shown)


dzucconi (approved):
Nice

[](https://capture.static.damonzucconi.com/Screen-Shot-2024-11-12-16-18-00.47-2LGRRh97fxYXJmYVHLREl8HeQ0nrMRmwcS7ckNvBGe2pdRqd2seju3hFDzpNbg3nBBUgohGgISyFvCPyQrv7Qib8olYTs2g71hLs.png)

damassi (approved):
Yess, thanks for diving into webpack!

PR #14846, opened 2024-11-14
feat(EMI-2159): Auction registration address autocomplete

erikdstock:

This PR solves [EMI-2159] and relates to [EMI-2196]

### Description

This PR adds autocomplete for US addresses on auction registration and makes minor changes to the registration form layout to follow our latest specs for address forms.

It also simplifies the autocomplete input with the tradeoff of eliminating tracking for when a user edits an input on an autocompleted address.

[EMI-2159]: https://artsyproduct.atlassian.net/browse/EMI-2159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMI-2196]: https://artsyproduct.atlassian.net/browse/EMI-2196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Auction/Components/Form/AddressForm.tsx
+++ b/src/Apps/Auction/Components/Form/AddressForm.tsx
@@ -1,9 +1,42 @@
 import { Column, GridColumns, Input } from "@artsy/palette"
 import { useFormContext } from "Apps/Auction/Hooks/useFormContext"
 import { CountrySelect } from "Components/CountrySelect"
+import {
+  AddressAutocompleteInput,
+  useAddressAutocompleteTracking,
+} from "Components/Address/AddressAutocompleteInput"
+import { useState } from "react"
+import { useAnalyticsContext } from "System/Hooks/useAnalyticsContext"
+// import { ContextModule } from "@artsy/cohesion"
 
 export const AddressForm = () => {
-  const { handleChange, handleBlur, errors, values, touched } = useFormContext()
+  const {
+    handleChange,
+    handleBlur,
+    errors,
+    values,
+    touched,
+    setValues,
+    setFieldValue,
+  } = useFormContext()
+  const { contextPageOwnerId, contextPageOwnerType } = useAnalyticsContext()
+
+  const autocompleteTrackingValues = {
+    contextModule: "auctionRegistration" as any, //ContextModule.auctionRegistration,
+    // contextModule: ContextModule.auctionRegistration,
+    contextOwnerType: contextPageOwnerType,
+    contextPageOwnerId: contextPageOwnerId || "",
+  }
+
+  const trackAutoCompleteEdits = (fieldName: string, handleChange) => (
+    ...args
+  ) => {
+    if (hasAutocompletedAddress) {
+      autocompleteTracking.editedAutocompletedAddress(fieldName)
+      setHasAutocompletedAddress(false)
+    }
+    handleChange(...args)
+  }
 
   return (
     <GridColumns>
@@ -22,7 +55,7 @@ export const AddressForm = () => {
         />
       </Column>
 
-      <Column span={6}>
+      <Column span={12}>
         <CountrySelect
           name="address.country"
           title="Country"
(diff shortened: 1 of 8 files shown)


src/Apps/Auction/Components/Form/AddressForm.tsx
```diff
@@ -1,9 +1,42 @@
...
+
+  const trackAutoCompleteEdits = (fieldName: string, handleChange) => (
+    ...args
+  ) => {
+    if (hasAutocompletedAddress) {
+      autocompleteTracking.editedAutocompletedAddress(fieldName)
+      setHasAutocompletedAddress(false)
+    }
+    handleChange(...args)
+  }
```
damassi, line 39: Couple things with this:
- if we have a tracking hook that co-locates all tracking, then we shouldn't be defining local functions that also handle tracking; we should figure out how to make it ideal, such as adding a second callback argument to `editedAutocompletedAddress` and then invoking that
- Or (and better) renaming the local function: `handleAutocompleteChange` and then invoking the tracking function from that
- Also, no need to pass in `handleChange` as an argument; its already in scope up above
- Also, we don't need to pass

src/Apps/Auction/Components/Form/AddressForm.tsx
```diff
@@ -80,7 +127,7 @@ export const AddressForm = () => {
         />
       </Column>
 
-      <Column span={6}>
+      <Column span={12}>
```
damassi, line 130: Defaults to 12

src/Apps/Auction/Components/Form/AddressForm.tsx
erikdstock: yeah, mostly agree here. I'll see if i can improve it a little more.

erikdstock: OK, came back to this and reminded myself how we landed on this structure. The [AddressAutocomplete input exports both its Component and a separate Hook](https://github.com/artsy/force/blob/50611a5316b3b493aff40d96b4e479e180384470/src/Components/Address/AddressAutocompleteInput.tsx#L74) for tracking interactions related to the autocomplete functionality. Normally we would the tracking to exist inside the component, but in this case the job of the autocomplete is to act on the other form inputs - [you can see that down below](https://github.com/artsy/force/pull/14846/files#diff-4517e067c32e835135b2ffc91b55b7d5a0cd01ae8aebe3b1ca2033e7e2c92a38R86-R106). Even so, that change handling could just go in as a prop along with the tracking values like context module up above.

The problem is that we also track when a user _edits_ an autocompleted address and which input that was, which creates a chicken/egg problem between the input and the tracking. We solved it in the first autocomplete implementation (or maybe first iteration on it) by hoisting that one bit of state and tracking functionality up to the same scope as all those inputs in the form of this hook. I remember that another possibility I tried at the time was having the hook return the component itself, like `const { addressAutocompleteTracking, AddressAutocomplete } = useAddressAutocomplete(trackingValues)` so they would still come from a single place, but ultimately this is where we landed. Another option might be a ref or other stateful value.

Finally I wonder if we still need this level of tracking at all. All that said, this is the first time we are bringing the input to a second address form so let me think about best options to move forward.

erikdstock:
After discussing, we'll probably split this into two separate PRs - one for the layout and another for the autocomplete input. I'm going to take a look at what we have so far here and see if it is worth using that or starting from scratch.

damassi (dismissed):

src/Apps/Auction/Components/Form/AddressForm.tsx
erikdstock: When I removed this the layout broke so I think the 12 is actually required. Replaced in [4fa5922](https://github.com/artsy/force/pull/14846/commits/4fa5922cd24b93d4bf9b7ffb3b92404b65e2e941).

PR #14858, opened 2024-11-18
chore(perf): avoid loading associated artists unnecessarily

joeyAghion:
This **chore** attempts to eliminate 1 or a few unnecessary API requests from one of the most common queries: for the artwork page. In the context of an artwork, the associated `artist`, `artists`, or `partner` can often be resolved from the embedded data already available in the artwork's JSON.

This will probably only yield a tiny improvement, but artworks are aggressively prefetched so it may be meaningful or at least give us more headroom.

A similar optimization could be done for artworks' `partner` association, except for [this dependency on its `isInquireable` property](https://github.com/artsy/force/blob/847542408143b63df4ae151acc9f7faee56348b8/src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarPartnerInfo.tsx#L56). That [isn't returned in partners' "short" JSON](https://github.com/artsy/gravity/blob/9796903d89ba263726400c4df0237979a9e46b5c/app/models/domain/partner.rb#L751), but maybe it _should_ be since it only depends on readily available partner fields. That's left as a separate exercise.


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -433,7 +433,7 @@ const ArtworkAppFragmentContainer = createFragmentContainer(
           extendedBiddingIntervalMinutes
         }
         saleMessage
-        artists {
+        artists(shallow: true) {
```
damassi, line 436: I know this is already done on the MP side, and in general i'm always super opposed to 1:2 representations, but it'd be great to avoid the `shallow` `all` `short` abstractions on the MP side, and use something a little more literal

src/Apps/Artwork/Components/ArtworkAuctionCreateAlertHeader/ArtworkAuctionCreateAlertHeader.tsx
```diff
@@ -172,7 +172,7 @@ export const ArtworkAuctionCreateAlertHeaderFragmentContainer = createFragmentCo
         isInAuction
         artistNames
         internalID
-        artists {
+        artists(shallow: true) {
```
damassi, line 175: ...especially now that this is about to spread, and future devs won't be able to understand what this means, given the fact that most current devs don't (actually) understand what this abstraction means

damassi: #nonblocking

src/Apps/Artwork/ArtworkApp.tsx
joeyAghion: What would you recommend? It seems like this should simply be documented on the `shallow` arguments within Metaphysics. Currently that looks like:

```
"Use whatever is in the original response instead of making a request"
```

damassi: The comment says it all, though it'd be nice if we could summarize that comment in a way that [makes sense](https://www.merriam-webster.com/dictionary/shallow) for a variable name. GraphQL is our idealized UI for our backends. I don't know if any of us would chose this API if we could start again.

joeyAghion: Note that `shallow` _usually_ corresponds with the "short" JSON from the upstream API, but not always when `reference_properties` [are specified](https://github.com/mongoid/mongoid-cached-json?tab=readme-ov-file#reference-properties).

damassi: It is what it is, and the above further emphasizes my point 🤷 Just a thought to consider GraphQL as UI, and noting a potential opportunity, but don't hold up anything for this comment 👍

damassi (dismissed):

damassi (approved):

PR #14862, opened 2024-11-19
fix: stabilize transitions on top nav

dzucconi:
https://github.com/user-attachments/assets/529525f4-9d12-4ddb-b504-3446cd1f4201

The palette updates required two key changes: implementing the ability to disable transitions on dropdowns, and synchronizing the visibility prop changes with pointer events. Without this synchronization, focus isolation would incorrectly activate during transition-out states.

For the nav implementation, we maintain an entry counter for the row, limiting transitions to the first entry only, with the counter resetting on nav exit.


dzucconi:
Dedicated to @damassi

dzucconi:
Woops. Pushed to wrong branch. Will reopen.

PR #14419, opened 2024-08-30
chore: Move Inquiry routes to Sell app

olerichter00:
Addresses [ONYX-1074]

## Description

Move Inquiry routes to Sell app.

[ONYX-1074]: https://artsyproduct.atlassian.net/browse/ONYX-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #11578, opened 2022-12-16
feat(CX-2939): consignment inquiry flow web

lordkiz:
The type of this PR is: **Feat**

This PR solves [CX-2939]

### Description

- Implement consignment inquiry flow on web.

https://user-images.githubusercontent.com/18648835/208103298-3fb09fd7-6f6d-439e-a806-4f285a6daef5.mov

[GRO-434]: https://artsyproduct.atlassian.net/browse/GRO-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CX-2939]: https://artsyproduct.atlassian.net/browse/CX-2939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14864, opened 2024-11-19
fix(consignment inquiry): correct routing for email form

anandaroop:

This PR solves [ONYX-1401]

### Description

- The consignment inquiry flow was added in https://github.com/artsy/force/pull/11578

- Then those routes were re-organized in https://github.com/artsy/force/pull/14419

This resulted in some extraneous nesting that broke the inquiry email flow:

387841704-d0acb203-47cd-4d45-afb9-3ba23f27da0a

[ONYX-1401]: https://artsyproduct.atlassian.net/browse/ONYX-1401?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Sell/Routes/ConsignmentInquiry/ConsignmentInquiryConfirmation.tsx
+++ b/src/Apps/Sell/Routes/ConsignmentInquiry/ConsignmentInquiryConfirmation.tsx
@@ -13,7 +13,7 @@ export const ConsignmentInquiryConfirmation: React.FC = () => {
       <Spacer y={4} />
 
       <Button
-        width={["100%", "20%"]}
+        width={["100%", "auto"]}
         data-test-id="back-to-sell-with-artsy-button"
         size="large"
         variant="primaryBlack"
--- a/src/Apps/Sell/sellRoutes.tsx
+++ b/src/Apps/Sell/sellRoutes.tsx
@@ -579,8 +579,7 @@ export const sellRoutes: RouteProps[] = [
             },
           },
           {
-            path: "inquiry/:recipientEmail?",
-            getComponent: () => ConsignmentInquiryContainer,
+            path: ":recipientEmail?",
             children: [
               {
                 path: "",


src/Apps/Sell/sellRoutes.tsx
```diff
@@ -579,8 +579,7 @@ export const sellRoutes: RouteProps[] = [
             },
           },
           {
-            path: "inquiry/:recipientEmail?",
-            getComponent: () => ConsignmentInquiryContainer,
+            path: ":recipientEmail?",
```
anandaroop, line 582: This fixes the route nesting, as well as the visual nesting (the double logo you can see in the screen cap above)

| Before | After |
|--------|--------|
| Screenshot 2024-11-19 at 4 34 29 PM | Screenshot 2024-11-19 at 4 34 12 PM |

src/Apps/Sell/Routes/ConsignmentInquiry/ConsignmentInquiryConfirmation.tsx
```diff
@@ -13,7 +13,7 @@ export const ConsignmentInquiryConfirmation: React.FC = () => {
       <Spacer y={4} />
 
       <Button
-        width={["100%", "20%"]}
+        width={["100%", "auto"]}
```
anandaroop, line 16: Threw this in for free:

| Before | After |
|--------|--------|
| Screenshot 2024-11-19 at 4 14 38 PM | Screenshot 2024-11-19 at 4 15 51 PM |

anandaroop:
There seems to be an open question about [the recipients for these emails](https://artsy.slack.com/archives/C07PRTJSD6G/p1732051418462999?thread_ts=1732029154.377049&cid=C07PRTJSD6G). 

I think that's going to be a Pulse or Convection concern, but we can verify before merging this.

anandaroop:
Confirmed that the specialist email is included in the inquiry payload.

So the recipient fix (if any) will have to be addressed upstream. 

This PR should be good as-is.

olerichter00 (approved):
Thanks for the fix! 🙏

PR #14873, opened 2024-11-20
fix: Retain useClientQuery queries

olerichter00:
Resolves [Slack Bug Thread](https://artsy.slack.com/archives/C07PRTJSD6G/p1727723121828589)
## Description

To prevent runtime issues from queries being garbage collected by Relay, I think it would be a good idea to [retain](https://relay.dev/docs/guided-tour/accessing-data-without-react/retaining-queries/) all queries fetched with `useClientQuery`.

This fix should resolve the [garbage collection issue](https://artsy.slack.com/archives/C07PRTJSD6G/p1727723121828589). 

I am also open to removing `useClientQuery` entirely in favor of `useLazyLoadQuery`. While `useClientQuery` has a great interface (no `Suspense` 🥳 ), it may be better to only use Relay's standard ways for data fetching (`QueryRenderer`, `fetchQuery` only for callbacks, and `useLazyLoadQueries`).

### DIFF
--- a/src/Utils/Hooks/useClientQuery.ts
+++ b/src/Utils/Hooks/useClientQuery.ts
@@ -1,6 +1,11 @@
 import { useUpdateEffect } from "@artsy/palette"
 import { useEffect, useRef, useState } from "react"
-import { Environment, fetchQuery, GraphQLTaggedNode } from "react-relay"
+import {
+  Disposable,
+  Environment,
+  fetchQuery,
+  GraphQLTaggedNode,
+} from "react-relay"
 import {
   CacheConfig,
   createOperationDescriptor,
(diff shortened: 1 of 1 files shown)


damassi:
Checking in the app, we're only using this hook in a few places, and I think adding retain is safe. 

Screenshot 2024-11-20 at 7 46 29 AM

One thing I would suggest we do tho is _return_ the dispose method so that we can call it on unmount!

src/Utils/Hooks/useClientQuery.ts
```diff
@@ -50,6 +52,14 @@ export const useClientQuery = <T extends OperationType>({
...
       setData(res)
       setLoading(false)
+
+      const operation = createOperationDescriptor(
+        getRequest(query),
+        variables ?? {}
+      )
+
+      // Retain the operation to prevent it from being garbage collected. Garbage collection can compromise type safety (e.g. non-nullable values being `null`), potentially leading to runtime errors.
+      relayEnvironment.retain(operation)
```
damassi, line 62: lets return the disposable instance in the return value of the hook:

```tsx
const disposable = relayEnvironment.retain(operation)
...
return {
  ... 
  dispose: disposable.dispose 
}
```

olerichter00: Good idea 👍

PR #14874, opened 2024-11-20
fix(EMI-2179): artwork summary item image

rquartararo:

This PR solves [EMI-2179]

### Description

This fixes the issue where the `ArtworkSummaryItem` image url was returning `null`. The problem was related to the aliasing. Not sure if this is the best solution as we are not reziging the image to `55px` but to `185px` to align with `ItemReview`. 

Screenshot 2024-11-20 at 7 57 30 AM

[EMI-2179]: https://artsyproduct.atlassian.net/browse/EMI-2179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MrSltun (dismissed):
Looks good! 🌟

fladson:
Yeah, I was reluctant to submit this solution since this alias was introduced a long time ago and things were working just fine. I would just ping some Metaphysics people to see if they have any context on this.
Non-blocking though, the UI seems to handle 185 fine so it's better than having nothing.

PR #14875, opened 2024-11-20
feat: removes cookie consent banner for googlebot

dzucconi:
This change is straightforward - it disables the cookie consent banner specifically for Googlebot requests. We've observed the consent banner appearing in web.measure snapshots during Lighthouse testing, so this should impact LCP. Should result in improved page performance in certain scenarios. It also just makes sense.

@joeyAghion I've assigned you, feel free to merge; as I'm uncertain how this plays with the other changes we're doing.


joeyAghion (approved):
I believe some network requests to Segment are required for populating the consent modal. Will a `false` value for `isDisplayable` result in skipping those?

PR #14885, opened 2024-11-21
chore(perf): avoid unnecessary partner loads (DIA-989)

joeyAghion:
Like https://github.com/artsy/force/pull/14858 and https://github.com/artsy/force/pull/14877, but for artworks' associated `partner`s. Depends on https://github.com/artsy/gravity/pull/18353.

https://artsyproduct.atlassian.net/browse/DIA-989


damassi (approved):
✌️

joeyAghion:
I've now reversed this PR's direction. Practically _all_ `Artwork.partner` queries are for properties available in the "shallow" JSON, potentially saving an upstream API request. The only exception was [this query](https://github.com/artsy/force/blob/50eb09335c9d0401d582b1e5de4390edcf2012e4/src/Apps/Artwork/Components/TrustSignals/VerifiedSeller.tsx#L39) in a component that... is unused.

Rather than propagate `shallow: true` everywhere, I'm following @damassi's suggestion to make that the default (within Metaphysics' `Artwork` type). I've already confirmed that Eigen's queries can also all be satisfied from shallow data, without code changes. I'll make that Metaphysics update separately. This PR now just fixes a lint-ing offense and removes the unused component.

PR #14888, opened 2024-11-21
fix(artwork): Fix deepzoom

damassi:

### Description

Noticed that deepzoom was no longer working, but passing the ref into the dependency array fixed it. Some change between React 17 and 18? Not too sure. 

(My internet is slow, just heads up): 

https://github.com/user-attachments/assets/fed6732b-ecde-487f-9c7c-d6911b79ae34


dzucconi (approved):
Weird!

PR #14893, opened 2024-11-22
fix(artwork-brick): Fix artwork hover zoom performance

damassi:

### Description

Perf on this has become janky over the past few months. This fixes that: 

https://github.com/user-attachments/assets/6449cbcf-8ac6-47fb-9062-cb48ab037984

cc @artsy/diamond-devs

### DIFF
--- a/src/Components/MagnifyImage.tsx
+++ b/src/Components/MagnifyImage.tsx
@@ -3,13 +3,6 @@ import { useRef, useState } from "react"
 import styled from "styled-components"
 import { isTouch } from "Utils/device"
 
-type Position = number | null
-
-interface Positions {
-  x: Position
-  y: Position
-}
-
 interface MagnifyImageProps extends ImageProps {
   scale?: number
   scaleDuration?: number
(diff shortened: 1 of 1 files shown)


src/Components/MagnifyImage.tsx
```diff
@@ -63,9 +55,8 @@ export const MagnifyImage: React.FC<React.PropsWithChildren<MagnifyImageProps>>
     const x = e.clientX - containerRect.left
     const y = e.clientY - containerRect.top
 
-    setPositions({
-      x,
-      y,
+    requestAnimationFrame(() => {
```
damassi, line 58: ☝️ the fix

egdbear (approved):

src/Components/MagnifyImage.tsx
starsirius: Nice! So I guess without `requestAnimationFrame`, it'd be rendering faster than it needs to be and that caused the possible janky animation.

src/Components/MagnifyImage.tsx
```diff
@@ -78,16 +69,16 @@ export const MagnifyImage: React.FC<React.PropsWithChildren<MagnifyImageProps>>
...
-      ref={containerRef as any}
+      ref={containerRef}
     >
       <Image
         {...rest}
         title={undefined} // don't display a tooltip when hovering over the image
         style={{
           ...rest.style,
-          transformOrigin: getTransformOrigin(positions),
-          transition: `transform ${scaleDuration}s ease, transform-origin 100ms ease, opacity 0.25s`,
```
starsirius, line 89: TIL `transform-origin` is invalid here!

dzucconi:
Future optimizations could include disabling this behavior during scroll and eliminating state management entirely in favor of direct DOM manipulation.

PR #14894, opened 2024-11-22
fix: ensures download banner is shown for all users

dzucconi:
The implementation issue stems from enabling page caching, which prevents page pre-fetching in numerous scenarios. Previously, route change completion was only being detected on routes where the loading bar appeared at the top of the screen. In this case, route completion status isn't relevant - we simply need to detect and display based on the number of routes, which is what this implementation now does.

There are two other call sites using this hook that require investigation to determine if additional refactoring opportunities exist.

### DIFF
--- a/src/Components/AppDownloadFooter.tsx
+++ b/src/Components/AppDownloadFooter.tsx
@@ -6,7 +6,7 @@ import {
   Text,
   useDidMount,
 } from "@artsy/palette"
-import { FC, useRef } from "react"
+import { FC, useEffect, useRef } from "react"
 import ArtsyMarkIcon from "@artsy/icons/ArtsyMarkIcon"
 import { useDeviceDetection } from "Utils/Hooks/useDeviceDetection"
 import { Z } from "Apps/Components/constants"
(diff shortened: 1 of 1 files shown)


src/Components/AppDownloadFooter.tsx
```diff
@@ -37,20 +39,36 @@ export const AppDownloadFooter: FC<AppDownloadFooterProps> = () => {
...
+    // Increment route changes counter when pathname changes
+    if (match.location.pathname) {
+      routeChanges.current++
+    }
+
+    // Show banner after second route change
+    if (routeChanges.current === 2) {
+      showBanner()
+    }
+  }, [match.location.pathname])
```
damassi, line 71: This approach makes sense but would it work to detect elements (meaning, if the page is loading)? This is the way we can always see if things are fetching / showing top bar or not:

```
const { match } = useRouter() 

if (!match.elements) {
  // is fetching 
}

if (match.elements) {
  // page is done loading
}
```

src/Components/AppDownloadFooter.tsx
```diff
@@ -37,20 +39,36 @@ export const AppDownloadFooter: FC<AppDownloadFooterProps> = () => {
...
+      isVisible.current = true
+    }
+
+    // Increment route changes counter when pathname changes
+    if (match.location.pathname) {
+      routeChanges.current++
+    }
+
+    // Show banner after second route change
+    if (routeChanges.current === 2) {
```
damassi, line 68: if the `match.elements` approach doesnt work, could we roll this up into a reusable hook? Or is the `=== 2` bit specific to this page?

dzucconi: It indeed no longer works and this is specific to this component

src/Components/AppDownloadFooter.tsx
dzucconi: That's what was happening here previously that doesn't work for cache pages. It will never fire, because it will never be loading.

damassi: makes sense. I wonder how we can generalize this a bit more, but seems a bit fussy...

damassi (approved):

PR #14897, opened 2024-11-22
feat: load limited-time offer signals on the client side

starsirius:

This PR solves [EMI-2161]

### Description

In the [previous attempt](https://github.com/artsy/force/pull/14807), we tried a few design options to display the limited-time offer signals on the client-side. We then realized the signal has to be considered together with other badge signals (i.e. "increased interest" and "curators' pick") because we only want to display at most one. So we [decided](https://artsy.slack.com/archives/C02JHHHKP5K/p1731340451105599) to switch back to the original signals design and let client-side loaded signals, if any, dynamically replace server-side rendered content. We now think it's acceptable to have content shift for this particular data which is relatively lower volume.

- This is an example that the client-side loaded limited-time offer signal replaces an increased interest badge in-place. We can see the offered price also replaces the listed price shortly after page loads, as well as the timer and save state pops in.
    <details><summary>Recording</summary>
    <p>
    
    client-side-collector-signals-with-primary-label
    
    </p>
    </details> 

- This is an example when without other badges, the limited-time offer badge is _inserted_ which causes some content shift.
    <details><summary>Recording</summary>
    <p>
    
    client-side-collector-signals-no-primary-label
    
    </p>
    </details> 

cc @artsy/emerald-devs 

[EMI-2161]: https://artsyproduct.atlassian.net/browse/EMI-2161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/Artwork/Details/__tests__/SaleMessage.jest.tsx
```diff
@@ -0,0 +1,235 @@
...
+      return { ref: createRef() }
+    })
+
+    mockEnvironment = createMockEnvironment()
+    mockUseSystemContext.mockImplementation(() => {
+      return {
+        relayEnvironment: mockEnvironment,
+      }
+    })
+  })
```
damassi, line 42: #non-blocking (but recommended): Rather than mocking out all of this stuff, can simply test the fragment via our relay testing tools and skip all of the qury renderer wiring (which historically we typically don't test):
```tsx
  Component: SaleMessageFragmentContainer
  query: ...
})

it('works', () => {
  renderWithRelay(someFixture)
  
  expect(screen.getByText("foo")).toBeOnScreen() 
})
```

src/Components/Artwork/Details/__tests__/SaleMessage.jest.tsx
```diff
@@ -0,0 +1,235 @@
...
+      const renderSaleMessage = artwork => {
+        render(
+          <SaleMessageQueryRenderer
+            artwork={artwork}
+            id={"opaque-internal-id"}
+          />
+        )
+
+        act(() => {
+          mockEnvironment.mock.resolveMostRecentOperation(operation =>
```
damassi, line 55: This allows us to skip manually mocking both the test environment and the relay operation; all of those things happen inside of our test tools automatically.

src/Components/Artwork/Details/SaleMessage.tsx
```diff
@@ -0,0 +1,132 @@
...
+      <Text lineHeight="22px">{offeredPrice}</Text>
+      &nbsp;
+      <ActivePartnerOfferTimer artwork={artwork} />
+    </Flex>
+  ) : (
+    publicSaleMessage
+  )
+}
+
+const SaleMessageFragmentContainer = createFragmentContainer(SaleMessage, {
```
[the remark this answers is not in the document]
starsirius: Good to know! I was confused about the best practices doc that [recommends relay containers](https://github.com/artsy/force/blob/d33e0a69b5e6a39805e4bbc0925682677e7c93ca/docs/best_practices.md#prefer-relay-containers-higher-order-components-over-relay-hooks).

src/Components/Artwork/Details/__tests__/SaleMessage.jest.tsx
starsirius: Yeah I intentionally wanted to test the query renderers (vs. the fragment containers) to cover the 2-step rendering, e.g. server renders signal A and after fetching data on the clientside, replaces signal A with signal B.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -92,6 +91,8 @@ export const ArtworkSidebarCommercialButtons: React.FC<React.PropsWithChildren<A
   const createOrder = useCreateOrderMutation()
   const createOfferOrder = useCreateOfferOrderMutation()
 
+  const { signals } = useArtworkGridContext()
```
starsirius, line 94: Would this populate the data when the `PrimaryLabelLine` component is not rendered with this `ArtworkSidebarCommercialButtons` component?

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -124,9 +125,7 @@ export const ArtworkSidebarCommercialButtons: React.FC<React.PropsWithChildren<A
       context_owner_type: OwnerType.artwork,
       context_owner_slug: artwork.slug,
       context_owner_id: artwork.internalID,
-      signal_label: artwork.collectorSignals
-        ? getSignalLabel(artwork.collectorSignals)
-        : "",
+      signal_label: getSignalLabel(signals?.[artwork.internalID] ?? []),
```
starsirius, line 128: Would this populate the data when the `PrimaryLabelLine` component is not rendered with this `ArtworkSidebarCommercialButtons` component? I think we might still need a way to use the backend sourced signals. But I could be missing something here.

starsirius:
OK @rquartararo as we [agreed on](https://artsy.slack.com/archives/C02JHHHKP5K/p1735576517653679), the [last commit](https://github.com/artsy/force/pull/14897/commits/46fc278851ed0cd5b91dcfce95d18c4e72043266) put the clientside behavior behind a feature flag. I tested it locally (see below) and it seems to behave correctly. If this looks good to you, I'm going to merge this PR and we can create a separate PR for your tracking changes. And then we'll QA as a group next week! 🎉 

Note that I didn't guard everything behind the feature flag since I think some visual improvement (such as [fixing height](https://github.com/artsy/force/pull/14897/commits/a2a2b373eaf2b8c0c51bbadb379645f19c8f9d08)) is worth adding in.

- When the feature flag is disabled:
    <details><summary>Screenshot (staging on the left, this PR on the right)</summary>
    <p>
    
    Screenshot 2025-01-02 at 3 31 29 PM
    
    </p>
    </details> 

- When the feature flag is enabled:
    <details><summary>Recording (staging on the left, this PR on the right)</summary>
    <p>
    
    clientside-signals
    
    </p>
    </details>

src/Components/Artwork/Details/Details.tsx
```diff
@@ -355,12 +401,16 @@ export const Details: React.FC<React.PropsWithChildren<DetailsProps>> = ({
   }
 
   return (
-    <Box>
+    <Box height={CONTAINER_HEIGHT + "px"}>
```
damassi, line 404: fwiw, shouldn't need to provide "px" (but also fine)

src/Components/Artwork/Details/SaleMessage.tsx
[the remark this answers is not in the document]
damassi: Ah, that's quite updated. I'll update that.

src/Components/Artwork/Details/__tests__/PrimaryLabelLine.jest.tsx
```diff
@@ -0,0 +1,209 @@
...
+    })
+    mockUseArtworkGridContext.mockImplementation(() => {
+      return {}
+    })
+  })
+
+  describe("before fetching partner offer", () => {
+    it("shows increased interest when it's passed in as the label", () => {
+      render(
+        <PrimaryLabelLineQueryRenderer
```
damassi, line 51: This is non-blocking, but suggested: We don't need to test query renderers,  since they typically only render fragments. So just test the fragment, using our standard relay testing  tools (`setupTestRendererTL`). That allows us to avoid having to wire up networking manually. Everything can happen inside of the test helper.

src/Components/Artwork/Details/__tests__/PrimaryLabelLine.jest.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  describe("after fetching null partner offer", () => {
+    const renderPrimaryLabelLine = label => {
+      const { container } = render(
+        <PrimaryLabelLineQueryRenderer id="opaque-internal-id" label={label} />
+      )
+
+      act(() => {
+        mockEnvironment.mock.resolveMostRecentOperation(operation =>
+          MockPayloadGenerator.generate(operation, {
```
damassi, line 110: For example, this stuff can just be provided as props to the fragment component in the test helper setup.

src/Components/Artwork/Details/__tests__/SaleMessage.jest.tsx
damassi: Ah, just saw this comment! ignore the other 👍

src/Components/Artwork/Details/__tests__/PrimaryLabelLine.jest.tsx
damassi: ah, just saw my other comment. Ignore 👍

src/Components/Artwork/Details/Details.tsx
starsirius: Good to know! I thought it didn't work without it but I could be wrong.

damassi (approved):
Lots of detail in this PR! Thanks so much for taking it on. 

My only Q is: do we need the feature flag, or can we just push it to a quick review-app and test things that way?

src/Components/Artwork/Details/__tests__/PrimaryLabelLine.jest.tsx
starsirius: Yeah you had the similar suggestion above. I understand it's usually not needed but I intentionally wanted to cover the [2-step rendering](https://github.com/artsy/force/pull/14897#discussion_r1876793670) due to this new server-renders-first-and-clientside-replaces-later behavior, and I thought testing the query renderer is a direct way for that purpose.

starsirius:
> My only Q is: do we need the feature flag, or can we just push it to a quick review-app and test things that way?

Yeah, my original plan was exactly that, and then the change grows to require more tracking changes (which will be included in a separate PR) and needs to be coordinated with the [upstream changes](https://github.com/artsy/metaphysics/pull/6220). Plus the wide surface areas this change will touch, so I want to roll it out in a more manageable/less risky way.

PR #14898, opened 2024-11-22
feat(EMI-2196): Invoice address form layout

erikdstock:

_The changes are relatively small when viewed commit-by-commit._

This PR relates to [EMI-2196]. It fixes the layout in our invoice payment address form and promotes the nearly identical auction + invoice address form fields to a single component. It also ports some tests from enzyme and fixes some type errors.

### Description

While doing #14846 I audited our address forms and made a document tracking where they diverged from our latest designs. I also found that
- Recent palette changes caused some inputs to be aligned wrong on several forms that didn't use this design
- The invoice and auction registration forms were almost 100% identical and wrong.

Because I had just fixed this on the auction form, I wondered if this might be our chance to consolidate address form fields to a single component. I think what I have here will do the trick and can eventually be brought to other address forms tracked in that audit. (Checkout shipping, payment & user settings)

- [ ] Needs integration with our invoice payment form. Only thing I need to solve here is the absence of the phone input. I think I can make this optional via a prop.
- [ ] Need to decide the shape of fields in the formik context. I have some thoughts here and will come back to it in an upgrade

[EMI-2196]: https://artsyproduct.atlassian.net/browse/EMI-2196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Auction/Hooks/useCreateTokenAndSubmit.ts
```diff
@@ -80,11 +80,16 @@ export const useCreateTokenAndSubmit = ({
 
     helpers.setSubmitting(true)
 
+    if (!values.address) {
+      helpers.setStatus("SUBMISSION_FAILED")
+      return
+    }
```
erikdstock, line 86: This caused a bunch of test failures [fixed [here](https://github.com/artsy/force/pull/14898/files#diff-fb8e302bbf531d00bb5320d725e45048f714ccda3f47e775506aedff347c0488R43)] even though I don't think it would be an issue in real life due to form validations. I only did it to satisfy some ts complaints so we could remove it.

erikdstock:
Hmm, I see some more validation schemas that I might need to reconcile:
- [Auction app validation schemas](https://github.com/artsy/force/blob/9fb0f70c7abda2db06a79d80d4a77dea68c53334/src/Apps/Auction/Components/Form/Utils/validationSchemas.ts#L12)
- [AuctionBidRoute](https://github.com/artsy/force/blob/53e03269f0f823ad82e1948078904203b6c257bc/src/Apps/Auction/Routes/Bid/AuctionBidRoute.tsx#L275-L279) uses some other schemas from the previous file

OTOH maybe this is ok as long as the validations work - but it would be nice to keep it all consistent and in one place.

erikdstock (commented):
PR tour.

src/Components/Address/AddressFormFields.tsx
```diff
@@ -1,11 +1,34 @@
...
+  address: Address
+  phoneNumber?: string
+}
+
+interface Props {
+  withPhoneNumber?: boolean
+}
+
+export const addressFormFieldsValidator = (args: {
+  withPhoneNumber: boolean
```
damassi, line 23: Should this be typed with `Props` right up above?

damassi (dismissed):
PR looks great @erikdstock 💯 

Only feedback is: The PR title seems a bit misleading considering everything that's going on in here. If we can perhaps summarize it there (for search) that'd be ideal, but otherwise 🍏

src/Components/Address/AddressFormFields.tsx
erikdstock: Done.

PR #14902, opened 2024-11-23
feat(react-18): Add modern streaming implementation 

damassi:

### Description

This one is quite exciting, because I wasn't sure if it was possible with styled-components, due to the core team waiting on the react team to release some new APIs for interleaving CSS within streams using the latest `renderToPipeableStream`. But to our luck, someone in the community just released an [example app](https://github.com/styled-components/styled-components/issues/3658#issuecomment-2480721193) solving this problem last week and I tested it out here and things seem good ❇️ 

**ENV Var**
- [ ] `hokusai staging env set ENABLE_SSR_STREAMING=true`
- [ ] `hokusai production env set ENABLE_SSR_STREAMING=true`

To test in dev, set ^ in `.env` file. 

For an overview on the powerful React 18 streaming features, see https://www.patterns.dev/react/react-selective-hydration/. 

One of the coolest examples of how this works is note how in the below, things are still streaming and JS still downloading, but features start becoming active even without all of the JS being ready. React has prioritized interactions and internally the system figures it out as stuff continues to load in the background. As seen in the network panel, many scripts are still pending. This (should) lead to much faster TTI stats, amongst other things. 

https://github.com/user-attachments/assets/72ad07d4-4f86-4344-b6e7-f3bece4843ba

There's still some more work to do to _optimize_ for first render (ie, hoist the page shell up above our router so that renders immediately before any network requests get sent out, but with React 18 + `renderToPipeableStream` in place we can start implementing that.

cc @artsy/diamond-devs


src/Components/FlashBanner/index.tsx
```diff
@@ -10,7 +10,7 @@ import { useRouter } from "System/Hooks/useRouter"
 import { FlashBanner_me$data } from "__generated__/FlashBanner_me.graphql"
 
 interface FlashBannerProps {
-  me?: FlashBanner_me$data
+  me?: FlashBanner_me$data | null
```
damassi, line 13: (Just fixing this old annoying console warning)

src/Components/NavBar/NavBarLoggedInActions.tsx
```diff
@@ -159,6 +159,7 @@ export const NavBarLoggedInActionsQueryRenderer: React.FC<React.PropsWithChildre
     <FallbackErrorBoundary FallbackComponent={Placeholder}>
       <SystemQueryRenderer<NavBarLoggedInActionsQuery>
         environment={relayEnvironment}
+        placeholder={user ? <Placeholder /> : undefined}
```
damassi, line 162: Fixed bug where nav flickers when checking whether logged in (after you've already logged in)

PR #14905, opened 2024-11-25
chore(perf): avoid loading artworks' associated artists unnecessarily

joeyAghion:
This is a 2nd attempt at avoiding API requests for artists on every artwork page query, like https://github.com/artsy/force/pull/14858 (which needed the follow-up fix in https://github.com/artsy/force/pull/14900). Now that https://github.com/artsy/gravity/pull/18377 is in place, artists' `follow_count` should always be available.

While I was in there, I removed queries for `counts.artists`, `counts.forSaleArtwork`, and a 2nd `name` which didn't seem to be used.

If this approach is acceptable, we should just wait until the Gravity change is released and its cache is cleared.


mzikherman (approved):
Sweet! Yup - once Gravity is deployed to prod with a `Rails.cache.clear` this should be good to go IMO.

joeyAghion:
Those JSON changes are released and Gravity's cache cleared so I'll merge this. 👀

PR #14906, opened 2024-11-25
feat(react-18): Add streaming via react-streaming

damassi:
The type of this PR is: **Feat**

### Description

**ENV Var**
- [ ] `hokusai staging env set ENABLE_SSR_STREAMING=true`
- [ ] `hokusai production env set ENABLE_SSR_STREAMING=true`

This updates https://github.com/artsy/force/pull/14902 with a more robust and less hand-rolled version of React's streaming API via [`react-streaming`](https://github.com/brillout/react-streaming). 

See [this commit](https://github.com/artsy/force/pull/14906/commits/57193d7ba4e5efe2189e1b5cdc0da3b376ebc2dc) for a diff of the changes.

When doing some manual testing, I noticed that a few things were less than perfect around error handling. Notably, when runtime exceptions were thrown (only in certain areas!) the error wasn't propagated up in a way that would trigger the error boundary and the 500 screen, and that felt a bit suspicious. 

With that in mind, I wanted to return to [`react-streaming`](https://github.com/brillout/react-streaming) and investigate. When wired up, everything behaved as expected giving me confidence that using this version is the way to go. 

cc @artsy/diamond-devs

PR #14911, opened 2024-11-26
feat(react-18): Add streaming via react-streaming

damassi:
Reverts artsy/force#14909

Reopening ^ in order to fix the linux import path issue.  See original PR for details: https://github.com/artsy/force/pull/14906

This fixes the react-streaming import path issue. For some reason our app isn't wanting to resolve the correct cjs import as defined in the library's [package.json file](https://github.com/brillout/react-streaming/blob/main/package.json#L16-L43). So for now, we're patch packaging it in order to force commonjs module resolution. 

See https://github.com/brillout/react-streaming/issues/46 for more information.


patches/react-streaming+0.3.43.patch
```diff
@@ -0,0 +1,20 @@
...
+ Object.defineProperty(exports, "disable", { enumerable: true, get: function () { return renderToStream_1.disable; } });
++
++// Node.js Stream
++const renderToNodeStream_1 = require("./renderToStream/renderToNodeStream");
++const renderToStream_1 = require("./renderToStream");
++(0, renderToStream_1.renderToNodeStream_set)(renderToNodeStream_1.renderToNodeStream);
++// Web Stream
++const renderToWebStream_1 = require("./renderToStream/renderToWebStream");
++const renderToStream_2 = require("./renderToStream");
++(0, renderToStream_2.renderToWebStream_set)(renderToWebStream_1.renderToWebStream);
```
damassi, line 20: Unfortunately we had to patch-package the lib in order to fix the issue 😢, but its worth it for what this library provides 😅  

Lines 13-20 were added to the main `react-streaming/server` import path. This patch-fixes the issue defined [here](https://github.com/brillout/react-streaming/issues/46) where `globalObject.renderToNodeStream` is null, due to things being unable to correctly resolve the file-path for a node.js environment.

damassi:
This looks good locally, so going to merge @joeyAghion

PR #14916, opened 2024-11-27
fix(artwork): adds loading indicator to zoom view

dzucconi:
I can't actually tell if something changed since upgrading to React 18 or I'm just noticing it now but the black screen while the deep zoom view loads seemed excessive. I've implemented a loading indicator to provide better user feedback during this delay, ensuring users see something on screen rather than empty space.

https://github.com/user-attachments/assets/00122232-66ce-4a56-8d91-4303f435b290


joeyAghion (approved):

damassi:
Thank you, I actually just noticed this too. Did we ever have a loader there? I don't think so?

PR #14918, opened 2024-11-27
feat(prefetching): Prefetch JS alongside network responses

damassi:

### Description

Very nice UX improvement in one line of code, which eliminates any blank screen behavior when transitioning between routes in our app (if a user hasn't visited the route before, and thus hasn't async downloaded the js for said route)! 

Adds a new route callback `onPrefetchJS` to request JS ahead of time (when the `usePrefetchRoute` hook is invoked), and renamed most `onClientSideRender` hooks as we typically preload bundle split JS there. 

https://github.com/user-attachments/assets/b4a507a3-5f03-4c01-9857-e68789a7d687

cc @artsy/diamond-devs

### DIFF
--- a/src/Apps/Components/AppShell.tsx
+++ b/src/Apps/Components/AppShell.tsx
@@ -1,6 +1,5 @@
 import { useNetworkOfflineMonitor } from "Utils/Hooks/useNetworkOfflineMonitor"
 import { Match } from "found"
-import { isFunction } from "lodash"
 import { useEffect } from "react"
 import * as React from "react"
 import createLogger from "Utils/logger"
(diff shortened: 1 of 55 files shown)


src/Apps/Components/AppShell.tsx
```diff
@@ -40,6 +42,10 @@ export const AppShell: React.FC<React.PropsWithChildren<AppShellProps>> = props
         logger.error(error)
       }
     }
+
+    // Preload current route's JS bundle
+    routeConfig?.onPreloadJS?.()
```
damassi, line 47: Invoke the new function in the app shell (for when users first land on a specific page via url)

joeyAghion (commented):
Heard, chef 👨‍🍳

PR #14921, opened 2024-11-27
Revert "feat(prefetching): Prefetch JS alongside network responses"

damassi:
Integrity is being flakey, so going to roll this back until after holidays.


damassi:
per slack: https://artsy.slack.com/archives/C02JHHHKP5K/p1732750569287779?thread_ts=1732750465.252429&cid=C02JHHHKP5K

PR #14933, opened 2024-12-02
chore(footer): eliminate app download footer query

dzucconi:
Introduced in https://github.com/artsy/force/pull/13755

This appears to be an unnecessary complexity that can be removed. If we need to maintain this functionality, we can implement it by linking to private artworks with a query string parameter that toggles the footer display or page layout. Reaching into the query itself would create difficulties in propagating those changes into the layout.


joeyAghion (dismissed):
> Reaching into the query itself would create difficulties in propagating those changes into the layout.

Could you explain this more? Do we render the layout before we know the URL?

dzucconi:
@joeyAghion The artwork's private/public status cannot be determined from the URL alone as far as I can tell?

If there was some indication in the actual URL, then we can just use that to suppress this.

joeyAghion:
There's [this `utm_content` param](https://github.com/artsy/gravity/blob/3976e6b1ca6bab80450a6ce8a3fe6a51b3219bc4/app/models/domain/artwork.rb#L1811) on the generated "share" URL but I'm not sure if that covers everything, like previews by the partner. Also not sure if that would be too gross to depend on.

dzucconi:
@joeyAghion I think that's completely reasonable actually! Will add

jpotts244 (approved):
🙏🏻

PR #14937, opened 2024-12-02
feat(artwork actions): add shortcut to image inspector tool

anandaroop:

### Description

Follow-up to https://github.com/artsy/forque/pull/1256

This adds an admin-only shortcut from the Artwork page directly to the **new image inspector tool** in Forque:

Screenshot 2024-12-02 at 1 08 19 PM

It replaces the previous **Genome** shortcut, which we can confidently say is no longer used. (While batch genome updating of artworks still happens occasionally via Rosalind, individual genoming via Helix has been essentially non-existent for years now.)

## TODO 

- [x] Add TOOLS_URL to Force's .env.shared 
- [x] Add TOOLS_URL to staging and production env
- [x] Remove GENOME_URL from Force's .env.shared 
- [ ] Remove GENOME_URL from staging and production env


anandaroop:
Going for it…

PR #14944, opened 2024-12-04
perf(EMI-1685): re-enable slow test

erikdstock:
The type of this PR is: **Perf**

This PR relates to [EMI-1685]
cc @artsy/emerald-devs 

### Description
**Note: after enabling the test it immediately failed due to a timeout. I therefore increased the timeout to make the test pass. Open to feedback on the approach, and we could check out the intermediate commit if we want to try more improvements.**

This PR makes refinements to some of our test helpers and re-enables a test that has been skipped. The test has been ported from enzyme and hopefully performance will be better now.

The intermediate commit 330b3539251c69f9090124ac4d99a233127845ce shows two different implementations of the once-problematic test for comparison. In short the improvements in this PR don't make a meaningful difference in runtime. Repeated runs in that commit show roughly the same timing, with the unoptimized initial code seeming to win slightly more often (difference usually ~100ms, but sometimes the newer implementation wins). The advantage is that we spend less time waiting for arbitrary `flushPromiseQueue()`s to resolve. One possible explanation for this difference is that the calls to `act()` around address form filling helpers may be adding a few ms. Otherwise the time spent waiting for the `flushPromiseQueue()` after filling in the form seems to mostly move into the form-filling time with `act()`.

On my M1 mac this re-enabled test usually took about 5100 ms. However depending on the load of the system it could sometime climb up past 8000ms. This sensitivity has been worse in CI, which is why we disabled the spec to begin with.

Longer but still pretty naive analysis of the different optimizations i have tried [in slack here](https://artsy.slack.com/archives/C2TQ4PT8R/p1733252086261459?thread_ts=1733245233.848769&cid=C2TQ4PT8R) 🔒. I noticed the biggest improvement (~1600ms) by stubbing out the palette Input in tests with a very basic html input/label, but this wouldn't be good for testing purposes.

Related changes
- In [this PR](https://github.com/artsy/force/commit/78d40839a9dd9699b2e717dbaefbdf4c84bf2369#diff-33851a69a0b6c5b35f9e2f3b152c1792af02bbb718ac49e2036f555a7c74b273R218) I changed the test file name, and from looking at the diff on this page I can't see where the original skipped test was added. The
- related test was disabled in https://github.com/artsy/force/pull/13435

[EMI-1685]: https://artsyproduct.atlassian.net/browse/EMI-1685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Shipping/Components/__tests__/SavedAddresses.jest.tsx
+++ b/src/Apps/Order/Routes/Shipping/Components/__tests__/SavedAddresses.jest.tsx
@@ -25,8 +25,8 @@ jest.mock("Utils/Hooks/useMatchMedia", () => ({
   __internal__useMatchMedia: () => ({}),
 }))
 
-// TODO: Can we get away with less?
-jest.setTimeout(10000)
+// Long-running tests when we `fillAddressFormFields()`
+jest.setTimeout(15000)
 
 let testProps: SavedAddressesProps
 let mockShippingContext: ShippingContextProps
--- a/src/Apps/Order/Routes/Shipping/Components/__tests__/SavedAddresses.jest.tsx
+++ b/src/Apps/Order/Routes/Shipping/Components/__tests__/SavedAddresses.jest.tsx
@@ -28,20 +28,6 @@ jest.mock("Utils/Hooks/useMatchMedia", () => ({
 // Long-running tests when we `fillAddressFormFields()`
 jest.setTimeout(15000)
 
-jest.mock("@artsy/palette", () => {
-  const originalModule = jest.requireActual("@artsy/palette")
-  return {
-    ...originalModule,
-    Input: ({ title, id, ...props }) => (
-      <>
-        {/* eslint-disable-next-line jsx-a11y/label-has-for */}
-        <label htmlFor={id}>{title}</label>
-        <input id={id} {...props} />
-      </>
-    ),
-  }
-})
-
 let testProps: SavedAddressesProps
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Shipping/Components/__tests__/SavedAddresses.jest.tsx
```diff
@@ -240,26 +240,29 @@ describe("Saved Addresses", () => {
...
         data: { ...validAddress },
       })
 
-      await userEvent.click(screen.getByText("Save"))
+      await waitFor(
+        () => {
+          userEvent.click(screen.getByText("Save"))
+        },
+        // Bottleneck waiting for form updates
+        { timeout: 4000 }
```
damassi, line 258: Might as well bump this more vs erroring out PR?

erikdstock: It was actually the overall test timeout limit of 10s that failed. I just increased it to see what we are working with here.

erikdstock:
😠 the results are in and if merged [this will become the slowest test](https://app.circleci.com/insights/github/artsy/force/workflows/default/tests?branch=erik.emi-1685-test-speed) at 12s. We could go back to the drawing board and try to improve the speed, give up and just disable the test, or change it to be about editing an address instead so there is less to fill in.

PR #14945, opened 2024-12-05
feat(compiler): Add RSBuild and remove Webpack, optimize JS bundles, speed up CI  

damassi:

This PR solves [DIA-1027]

Review App: https://rsbuild.artsy.net/

### Description

This is a major performance refactor for how we build, develop and deploy to CI in force, by replacing Webpack and Babel with [RSBuild](https://rsbuild.dev/), a vite-like application bundler built on top of Rust. 

There’s a few reasons for this:
- Speeding up deploys to prod. RSPack is built on Rust, so production builds clock in ~10s locally (~1s for dev); should be faster on CI. Vs a few minutes before
- Battle tested bundle splitting strategies, from ByteDance. Right now in force we’re using a very [low level chunkSplitting](https://github.com/artsy/force/blob/main/webpack/bundleSplitting.js) strategy that nobody understands. Lets lean on RSPack’s, which [simplifies everything](https://rsbuild.dev/guide/optimization/split-chunk)
- Smaller bundle sizes. Without any configuration whatsoever, have noticed ~4mb decrease in our total JS size
- Simpler config. Rather than needing to configure [babel](https://github.com/artsy/force/blob/main/babel.config.js) and [Webpack](https://github.com/artsy/force/tree/main/webpack) and understand both tools, the entirety of our configuration now lives in in rsbuild.config.ts, which is [much less that our team](https://github.com/artsy/force/blob/b461859c628ff2b8d813c36d19a1d84a1ebdfb10/rsbuild.config.ts) needs to understand!
- We get to remove 25 npm dependencies
- We also speed up our tests — Jest now runs on Rust, so in addition to faster deploys, we get faster CI cycles.
- Instant DX during development.

Mostly, the ultimate aim of this work comes down to simplifying our stack all around, and RSBuild has excellent straight-forward docs that is a whole universe away from Webpack, which required specialized knowledge to read and understand.

**Some Stats (before/after)**

- Cold boots during dev: ~1.5m -> ~4s, and we no longer suffer extreme server warmup delays of up to 30s-1m when opening `http://localhost:4000` in the browser after `yarn start` 
- Hot reloading during dev: 1s -> ~17ms for client-side code changes 
- Building production JS: ~1.5m -> 10s 

Because so many of our CI operations require bundling of assets, this migration impacts things up and down: 

- CI: Building Docker Images: 6m -> 1m 
- CI: Tests: 5m -> 1.5m 
- CI: Smoke Tests: ~9m -> ~3m 
- **Overall CI time on PRs: reduced to ~3.5m** 

Snapshot of this branch's CI run: 

Screenshot 2024-12-07 at 2 52 57 PM

Remarkably, TypeScript type-checking is now the bottleneck! Seems impossible but its true 😍 

[DIA-1027]: https://artsyproduct.atlassian.net/browse/DIA-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Artwork/Components/ArtistInfo.tsx
```diff
@@ -28,77 +27,71 @@ interface ArtistInfoProps {
...
-  @track({
-    action_type: DeprecatedSchema.ActionType.Click,
-    flow: DeprecatedSchema.Flow.ArtworkAboutTheArtist,
-    subject: DeprecatedSchema.Subject.ReadMore,
-    type: DeprecatedSchema.Type.Button,
-  })
-  trackArtistBioReadMoreClick() {
-    // Tracking
-  }
+const ArtistInfo: React.FC<ArtistInfoProps> = ({ artist }) => {
```
damassi, line 30: The following app changes were migrations to React.FC, to remove our use of `@decorators` in the app. Decorators subtly break every form of tooling that i've tried historically (and they didn't work well with rspack), so finally just ripped them out. About 5 components. 

We used them for tracking, so converted things into our normal canonical tracking patterns.

damassi:
@araujobarret - It comes down to the Rust toolchain being much more mature now. Just had to wait a couple of years for RSPack to come along.

rsbuild.config.ts
```diff
@@ -0,0 +1,124 @@
...
+  server: {
+    port: Number(process.env.PORT) || 3000,
+  },
+
+  performance: {
+    // bundleAnalyze: {
+    //   // TODO: Verify this works: yarn bundle-stats, along with relative-ci-agent
+    //   generateStatsFile: false,
+    // },
+  },
```
anandaroop, line 93: **question (non-blocking):** Deliberately keeping this omitted for now?

anandaroop (approved):
Exciting!

rsbuild.config.ts
damassi: This will be reenabled when I fix the relative-ci-agent job (the one that posts our bundle size stats to PRs). I'll add a fixme above the code comment 👍

joeyAghion (commented):
I'm most curious about how this will affect performance via JS bundle size and separation, especially the initial bundle that populates "early hints." Locally, I see just a few tiny files in `Link` header responses.

damassi:
Link to lack thread discussing ^ https://artsy.slack.com/archives/C2TQ4PT8R/p1733427321760689?thread_ts=1733419497.340759&cid=C2TQ4PT8R

src/Components/Address/__tests__/AddressAutocompleteInput.jest.tsx
```diff
@@ -298,7 +298,7 @@ describe("AddressAutocompleteInput", () => {
 
     // See TestImplementation for implementation details
     describe("tracking", () => {
-      it("tracks when autocomplete results are received", async () => {
+      it.skip("tracks when autocomplete results are received", async () => {
```
damassi, line 301: @erikdstock - this was timing out again

src/Server/passport/lib/app/__tests__/analytics.jest.ts
```diff
@@ -1,5 +1,3 @@
-import sinon from "sinon"
```
damassi, line 1: Cleaning up deps since everything we need comes from Jest

rsbuild.config.ts
```diff
@@ -0,0 +1,125 @@
...
+    writeToDisk: true,
+  },
+  performance: {
+    // FIXME: Uncomment and play around with different bundle split strategies,
+    // along with stats below:
+    // chunkSplit: {
+    //   strategy: "split-by-size",
+    //   minSize: 30000,
+    //   maxSize: 50000,
+    // },
```
MounirDhahri, line 85: Good idea on keeping this commented. I think it might make more sense to test this separately once everything else is stable and in prod to evaluate it properly.

MounirDhahri (approved):
Wow! this is impressive 👏

Apart from the major performance improvements, we get to simplify our infra a lot by not needing to configure babel and most importantly Webpack which is pretty tricky to get right and scary to evaluate the impact of changes to it. I am really excited about this jump- it's a good time given how mature RSBuild has become. 

Btw, I am not sure if you're familiar with [RSDoctor](https://rsdoctor.dev/guide/start/intro) or not, but it might be a nice addition (to this or a future PR) to get a "full" diagnose and analysis of your bundle.

How do you think about QA'ng this before rollout? using a review app? code freeze and monitor dashboards?

mc-jones:
Some delayed props. 👏 👏 👏

Dockerfile
```diff
@@ -47,21 +47,16 @@ COPY --chown=deploy:deploy --from=builder-base /opt/node_modules.prod ./node_mod
...
-COPY --chown=deploy:deploy --from=builder-base /app/public ./public
-COPY --chown=deploy:deploy --from=builder-base /app/src ./src
-
-# Server assets
-COPY --chown=deploy:deploy --from=builder-base /app/server.dist.js .
-COPY --chown=deploy:deploy --from=builder-base /app/server.dist.js.map .
 
 ENTRYPOINT ["/usr/bin/dumb-init", "./scripts/load_secrets_and_run.sh"]
 
+ENV NODE_PATH=/app/src
```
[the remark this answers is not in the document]
damassi: If you look below this line, we're using rust JIT compiler to execute the start script against the server (for the entrypoint only - its fast); we need this NODE_PATH variable to support absolute import paths.

PR #14490, opened 2024-09-17
fix(nav): lazy loads notifications dropdown

dzucconi:
Closes [DIA-858](https://artsyproduct.atlassian.net/browse/DIA-858)

Notifications require a pretty large network of code that isn't needed on page load and can be deferred until the notifications bell is clicked.

[DIA-858]: https://artsyproduct.atlassian.net/browse/DIA-858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #14947, opened 2024-12-05
fix: Revert lazy loading notifications dropdown

olerichter00:
- [Related Slack Thread](https://artsy.slack.com/archives/C07PRTJSD6G/p1727723121828589)
## Description

This reverts lazy loading notifications dropdown (https://github.com/artsy/force/pull/14490) to find out if this causes the bug described in the Slack thread. I cannot reproduce the bug locally and it only happens in Chrome.


MounirDhahri (approved):

olerichter00:
I'll merge this now to test if this is related to the error and will bring back the lazy loading in case it is not related.

PR #14953, opened 2024-12-05
Feat: update order history tab to reflect counteroffer received state

oxaudo:
This PR solves https://artsyproduct.atlassian.net/browse/EMI-2204

### Description

Opted in for combining displayState and buyerAction on the client here to cover 'counteroffer received' state.

Looks like this now:

Screenshot 2024-12-05 at 3 58 46 PM

### DIFF
--- a/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
+++ b/src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
@@ -16,7 +16,11 @@ import { DateTime } from "luxon"
 import { FC } from "react"
 import { createFragmentContainer, graphql } from "react-relay"
 import { RouterLink } from "System/Components/RouterLink"
-import { SettingsPurchasesRow_order$data } from "__generated__/SettingsPurchasesRow_order.graphql"
+import {
+  SettingsPurchasesRow_order$data,
+  CommerceOrderDisplayStateEnum,
+  CommerceBuyerOfferActionEnum,
+} from "__generated__/SettingsPurchasesRow_order.graphql"
 import { LocaleOptions } from "luxon"
 import { extractNodes } from "Utils/extractNodes"
 import { appendCurrencySymbol } from "Apps/Order/Utils/currencyUtils"
@@ -39,6 +43,7 @@ const ORDER_LABELS = {
   CANCELED: "Canceled",
   FULFILLED: "Delivered",
   IN_TRANSIT: "In transit",
+  OFFER_RECEIVED: "Counteroffer received",
   PROCESSING: "Processing",
   REFUNDED: "Refunded",
   SUBMITTED: "Pending",
@@ -51,6 +56,7 @@ const ORDER_ICONS = {
   CANCELED: <CloseStrokeIcon fill="red100" />,
   FULFILLED: <CheckmarkFillIcon fill="green100" />,
   IN_TRANSIT: <PendingIcon fill="black60" />,
+  OFFER_RECEIVED: <PendingIcon fill="blue100" />,
   PROCESSING: <PendingIcon fill="black60" />,
   REFUNDED: <CloseStrokeIcon fill="red100" />,
   SUBMITTED: <PendingIcon fill="black60" />,
@@ -63,6 +69,7 @@ const ORDER_COLORS = {
   CANCELED: "red100",
   FULFILLED: "green100",
   IN_TRANSIT: "black60",
+  OFFER_RECEIVED: "blue100",
   PROCESSING: "black60",
   REFUNDED: "red100",
   SUBMITTED: "black60",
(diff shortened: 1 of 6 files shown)


src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -63,13 +69,18 @@ const ORDER_COLORS = {
...
   PROCESSING: "black60",
   REFUNDED: "red100",
   SUBMITTED: "black60",
   PROCESSING_APPROVAL: "black60",
   PAYMENT_FAILED: "red100",
 } as const
 
+type CommerceBuyerOrderOfferStateEnum =
+  | CommerceBuyerOfferActionEnum
+  | CommerceOrderDisplayStateEnum
```
erikdstock, line 82: This seems like it will become a source of confusion if we don't resolve the tension soon, but :+1: for now

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -85,14 +96,61 @@ const getPaymentMethodText = (
...
+          variant="primaryBlack"
+          size="large"
+          width="50%"
+          minWidth={["240px", "200px"]}
+          my={[1, 0]}
+        >
+          Update Payment Method
+        </Button>
+      )
+    case "OFFER_RECEIVED":
```
erikdstock, line 119: This function is called `getOrderActionButton()` but it is basically a component, right? What if we had a single map up above in [our consts](https://github.com/artsy/force/pull/14953/files#diff-a958048d8ace47200344fc15c5ba10bc9029a8250be78f2c2497a402111e3f75R41), something like this:
```tsx
const ORDER_STUFF: Record<CommerceBuyerOrderOfferStateEnum, {
  label: string,
  icon: JSX.Element
  ActionButton?: (props: { order: any }) => JSX.Element
} = { 
  APPROVED: { 
//...

// later
{ ActionButton && <ActionButton order={order} />
```

This way logic could live in one place for the most part.

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -85,14 +96,61 @@ const getPaymentMethodText = (
...
+>> = ({ order }) => {
   const [lineItem] = extractNodes(order?.lineItems)
   const { artwork, artworkVersion, fulfillments } = lineItem
-  const { requestedFulfillment } = order
+  const { requestedFulfillment, buyerAction } = order
+  let displayState: CommerceBuyerOrderOfferStateEnum = order.displayState
+
+  if (displayState == "SUBMITTED" && buyerAction == "OFFER_RECEIVED") {
+    displayState = "OFFER_RECEIVED"
+  }
```
erikdstock, line 153: Just going back to my previous comment - I hope we can come to a permanent solution soon, maybe as part of checkout flow improvements. in the mean time i think i would name this variable something different from `displayState` since that is a specific value on the order from the server. `buyerState` seems right to me...

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -240,20 +298,7 @@ const SettingsPurchasesRow: FC<React.PropsWithChildren<SettingsPurchasesRowProps
...
-                variant="primaryBlack"
-                size="large"
-                width="50%"
-                minWidth={["240px", "200px"]}
-                my={[1, 0]}
-              >
-                Update Payment Method
-              </Button>
-            )}
+            {getOrderActionButton(displayState, order.internalID)}
```
erikdstock, line 301: Re: long comment above, this is where a component rendering would look better than a `getSomethingMightBeNull()` function call.

erikdstock (commented):
You Love To See It 👊

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
```diff
@@ -85,14 +96,61 @@ const getPaymentMethodText = (
...
+  orderId: string
+) => {
+  switch (displayState) {
+    case "PAYMENT_FAILED":
+      return (
+        <Button
+          // @ts-ignore
+          as={RouterLink}
+          to={`/orders/${orderId}/payment/new`}
+          variant="primaryBlack"
```
erikdstock, line 110: Is there a rule about limiting the number of primary buttons that appear on a page?

src/Apps/Settings/Routes/Purchases/Components/SettingsPurchasesRow.tsx
oxaudo: Good point. Will update. Thanks for suggestion. I think it did not make sense to begin with as we have only one special case for button. And I did not think of it as a change of pattern. But you are right - and it looks like we age going in a direction of having this button for more states. Totally legin change request.

oxaudo: Ok. Circling back here and realizing that not sure if it's worth doing the bigger refactor you've proposed before we move this 'buyerDisplayState' or whatever we are going to call it to the server. In current state we need to massage this state and buyerAction data before display so went with the function suggestion here. Hopefully names are a bit more descriptive. 

Also ended up moving the function inline as realized the button to update payment was missing the tracking hook.

Also this brings a question - should we track a click to 'Respond to Counteroffer' button. Will check with analytics and do a follow up if so.

erikdstock (approved):
LGTM

PR #14968, opened 2024-12-11
feat(dx): Open browser on yarn start during dev

damassi:

### Description

Now that it doesn't take a painful minute to actually serve the app in a browser window, during development automatically open the browser when the server is done booting.

cc @artsy/diamond-devs


joeyAghion:
I think this is a step too far. Starting a server and starting a client are 2 different operations that devs may understandable want to do independently. It could even hide bugs, by often starting development in a fresh tab on the home page.

PR #14974, opened 2024-12-11
chore(perf): include a high-priority image in Link response headers, if available

joeyAghion:
(Work-in-progress for feedback only.)

This is heavily based on https://github.com/artsy/force/pull/14827, and a few naïve assumptions. It attempts to add a `Link: ` header on responses indicating that browsers should preload an image from the rendered HTML.

* It only covers _the first_ image with `fetchpriority="high"`.
* It uses only the `src` attribute of that `img` element, so doesn't benefit from `srcSet` and may even cause a 2x browser to preload a 1x image unnecessarily.


damassi:
This was something that I noticed in the first version, where we'd be overfetching images. Check out this artwork page:
- https://staging.artsy.net/artwork/aboudia-nouchi-graffiti-bastik-14

PR (three images): 

Screenshot 2024-12-11 at 4 06 57 PM

Staging (two images): 

Screenshot 2024-12-11 at 4 07 25 PM

Might be worth removing the `<link>` tag from that page and then trying to hint it using the automated process in here (but that said, we might _need_ to also set the `imageSrcSet` for this to be valuable for our demographic (who are typically on 2x devices).

joeyAghion:
This PR currently results in `Link:` headers as follows:
* for artwork pages, the `quality=1` loading placeholder image, which I think is correct (if sneaky)
* for artist pages, the lower-resolution of the 2 available main images, which seems reasonable (even if we _could_ sense whether the client is 1x- or 2x-capable, Cloudflare would cache and return that hint to other clients)

I will try to verify if this _actually_ improves LCP locally or in a review app.

Questions:
* In both cases, it only returns a single image. Was your example referring to the prior PR?
* Alternatively, what if we searched in the body for `<link rel="preload"...` instead of `<img fetchpriority="high"...` as this does now? It might be more explicit.

damassi:
Discussed this [on slack](https://artsy.slack.com/archives/C05EEBNEF71/p1734023771763799). But yah, makes sense! We should just prioritize the 2x images getting into the header tags and then 🚢

joeyAghion:
Latest commit avoids any sort of HTML-parsing and just extracts the images for preloading from the already available head tags. That way, any components can render `<link>` tags via `react-head`, and they'll automatically appear in response headers. I think this is much cleaner than parsing the body for `` tags.

With this code, artwork pages respond with 2 images for preloading (the `quality=1` placeholder and the main image that replaces that), while artist pages respond with the single header image, as expected.

Next question, which can be considered in this PR or a separate one: those page currently populate `link` tags' `href`s with the "1x" image variant, similar to the `img` tags' `src` attribute. If we believe the "2x" variant will be more commonly needed by browsers, maybe we should default to the "2x" version in those `href`/`src` attributes. (That way, these response headers will automatically preload _that_.)

damassi (dismissed):
Love this elegant solution 👍 

Lets hint the 2x image and monitor for now, but otherwise merge when ready!

PR #6509, opened 2020-10-21
feat: redirect old artist series collections to true artist series

anandaroop:
https://artsyproduct.atlassian.net/browse/FX-2360

This is one step of [the migration happening currently](https://artsy.slack.com/archives/C9SATFLUU/p1603306225234000) to schlep all the old kaws-based artist series Collections ⇒ over to gravity ArtistSeries.

Once this is merged and deployed, any requests to the legacy `/collection/:slug` pages on artsy.net will be redirected to `/artist-series/:slug`

PR #14975, opened 2024-12-12
refactor(collections): modify redirects to be purely server-side

anandaroop:

This PR solves [ONYX-1445]

### Description

See [thread](https://artsy.slack.com/archives/C05EEBNEF71/p1733943061015619?thread_ts=1733940262.518139&cid=C05EEBNEF71) for context.

At [multiple](https://github.com/artsy/force/pull/6177) [various](https://github.com/artsy/force/pull/6509) [points](https://github.com/artsy/force/pull/12687) we've migrated artist series-esque `MarketingCollections` to "true" `ArtistSeries`.

We therefore added some redirect mappings for those collections. 

But that mapping file itself was unbeknownst to us adding >100k to the bundle size.

This refactors the redirection strategy to happen purely server-side at the Express layer rather than in the React SSR layer, in order to avoid this bundle bloat!

[ONYX-1445]: https://artsyproduct.atlassian.net/browse/ONYX-1445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -17,6 +17,7 @@ import { setupServerRouter } from "System/Router/serverRouter"
 import { getRoutes } from "System/Router/Utils/routeUtils"
 import { initializeMiddleware } from "middleware"
 import { errorHandlerMiddleware } from "Server/middleware/errorHandler"
+import { redirectCollectionToArtistSeries } from "Apps/Collect/Server/redirectCollectionToArtistSeries"
 
 const app = express()
 
@@ -24,6 +25,12 @@ const app = express()
 initializeMiddleware(app)
 
 const { routes, routePaths } = getRoutes()
+
+// Pre-emptive redirects
+//
+// These selectively match routes that would otherwise be handled by
+// the route matchers below
+app.get("/collection/:slug", redirectCollectionToArtistSeries)
 
 // React app routes
 app.get(
--- a/cypress/e2e/collection.cy.js
+++ b/cypress/e2e/collection.cy.js
@@ -38,3 +38,19 @@ describe("/collection/:id (a collection hub)", () => {
     artworkGridRenders()
   })
 })
+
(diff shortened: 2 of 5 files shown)


src/server.ts
```diff
@@ -25,6 +26,12 @@ initializeMiddleware(app)
 
 const { routes, routePaths } = getRoutes()
 
+// Pre-emptive redirects
+//
+// These selectively match routes that would otherwise be handled by
+// the route matchers below
+app.get("/collection/:slug", redirectCollectionToArtistSeries)
```
dzucconi, line 33: https://github.com/artsy/force/blob/main/src/Apps/Redirects/redirectsServerRoutes.tsx is maybe a more appropriate place

anandaroop: Yep that was my first approach.

But I found that because that middleware is [loaded after](https://github.com/artsy/force/blob/c127c7b32c77ffd4557dd90267cceba2724b3a2d/src/server.ts#L68) the [regular route matchers](https://github.com/artsy/force/blob/c127c7b32c77ffd4557dd90267cceba2724b3a2d/src/server.ts#L35-L37) that I was therefore not able to match the `/collection/foo` paths that I needed to intercept (which were already getting matched by [the Collection app](https://github.com/artsy/force/blob/c127c7b32c77ffd4557dd90267cceba2724b3a2d/src/Apps/Collect/collectRoutes.tsx#L63))

That was why I hoisted this app and added the comment. Lmk if you think there is still a better way to do this.

dzucconi: Oh! I see. Nice catch

damassi (approved):
Very nice ❤️

PR #14989, opened 2024-12-13
fix(router): Fix query param issue

damassi:

### Description

Un-applies https://github.com/artsy/force/pull/14984 and then fixes the underlying issue, which originated in https://github.com/artsy/force/pull/14962. We weren't passing query params across the route in the way that was required.

(Though one would assume passing a string with query params would be sufficient, as a pathname. But no -- pathname is pathname, and query is query, and search is search. When expanding the location (`to`) it expects a full location descriptor which encodes all the props.) 

https://github.com/user-attachments/assets/0cde56a8-3101-4025-b797-45ccefb11810

cc @artsy/diamond-devs

### DIFF
--- a/src/System/Components/RouterLink.tsx
+++ b/src/System/Components/RouterLink.tsx
@@ -89,12 +89,17 @@ export const RouterLink: React.FC<React.PropsWithChildren<
       }
     }
 
+    const location = router?.createLocation?.((to as string) ?? "") ?? {
+      pathname: to,
+      query: {},
+    }
+
     if (isRouterAware) {
       return (
         <RouterAwareLink
           inline={inline}
           to={{
-            pathname: to ?? "",
+            ...location,
             state: { isPrefetched },
           }}
           onMouseOver={handleMouseOver}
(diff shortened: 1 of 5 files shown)


src/System/Components/RouterLink.tsx
```diff
@@ -89,12 +89,14 @@ export const RouterLink: React.FC<React.PropsWithChildren<
       }
     }
 
+    const location = router.createLocation((to as string) ?? "")
```
damassi, line 92: this was the fix 👍

src/System/Components/RouterLink.tsx
```diff
@@ -89,12 +89,17 @@ export const RouterLink: React.FC<React.PropsWithChildren<
       }
     }
 
+    const location = router?.createLocation?.((to as string) ?? "") ?? {
+      pathname: to,
+      query: {},
+    }
```
damassi, line 95: Not the prettiest thing but need a fallback for our millions of tests that intersect this component

PR #14996, opened 2024-12-16
feat: Add Create Alert ability to artist series page

olerichter00:
Addresses [ONYX-646]

## Description

Add Create Alert ability to artist series page.

### DIFF
--- a/src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
+++ b/src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
@@ -1,20 +1,24 @@
+import { OwnerType } from "@artsy/cohesion"
 import { ArtistSeriesArtworksFilter_artistSeries$data } from "__generated__/ArtistSeriesArtworksFilter_artistSeries.graphql"
+import { ArtistSeriesArtworksFilterQuery } from "__generated__/ArtistSeriesArtworksFilterQuery.graphql"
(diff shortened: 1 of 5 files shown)


src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
```diff
@@ -33,6 +39,25 @@ const ArtistSeriesArtworksFilter: React.FC<React.PropsWithChildren<ArtistSeriesA
   // we still want to render the rest of the page.
   if (!hasFilter) return null
 
+  const artist = artistSeries.artists[0]
```
anandaroop, line 42: **thought (non-blocking):** I think this is reasonable for the vastly more common case where a series has _only one_ artist. 

But I do want to point out the edge case of >1 artist such as https://www.artsy.net/artist-series/christo-jeanne-claude-wrapped-reichstag. In that case, the alert will be created only with Christo and not Jeanne-Claude.

anandaroop (approved):
Nice, thanks! 🎉

src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
olerichter00: You're right, I forgot to map through all artists when setting the search criteria. I've created a PR with a fix: https://github.com/artsy/force/pull/14998

PR #14998, opened 2024-12-16
fix: Allow creating alerts with multiple artists from Artist series page

olerichter00:
Addresses [ONYX-646]

## Description

Allow creating alerts with multiple artists from Artist series page.

[ONYX-646]: https://artsyproduct.atlassian.net/browse/ONYX-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
+++ b/src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
@@ -78,7 +78,7 @@ const ArtistSeriesArtworksFilter: React.FC<React.PropsWithChildren<
     >
       <ArtworkFilterAlertContextProvider
         initialCriteria={{
-          artistIDs: [artist.internalID],
+          artistIDs: artistSeries.artists.map(artist => artist.internalID),
           artistSeriesIDs: [artistSeries.internalID],
         }}
       >


src/Apps/ArtistSeries/Components/ArtistSeriesArtworksFilter.tsx
```diff
@@ -78,7 +78,7 @@ const ArtistSeriesArtworksFilter: React.FC<React.PropsWithChildren<
     >
       <ArtworkFilterAlertContextProvider
         initialCriteria={{
-          artistIDs: [artist.internalID],
+          artistIDs: artistSeries.artists.map(artist => artist.internalID),
```
anandaroop, line 81: I suppose this is debatable.

Recall that our alert matching [logic](https://github.com/artsy/gravity/blob/eedfd50e4d02052faea45eaeb55a97f432101fd4/app/models/search_criteria.rb#L165-L222) basically works as follows:

- within each facet, criteria are OR'd together
- between facets, criteria are AND'd together 

For example an alert with the following settings…

**Artists**: Christo; Jeanne-Claude
**Mediums**: Photography; Sculpture 

…translates into the following matching logic…

```
(artist:christo OR artist:jeanne-claude) 
AND 
(medium:photography OR medium:sculpture)
```

Which means that the change here most likely does not have the intended effect and would create an overly broad alert, matching the first artist OR the second artist.

So we have two imperfect choices here — overly narrow (Christo only) or overly broad (Christo OR Jeanne-Claude).

I'm actually not convinced that overly broad is better, but this is maybe more of a product question.

olerichter00: Puh, thanks for the reminder about this detail... In the end, it probably doesn't make much of a difference unless the artist series criteria is removed from the alert.

How does the logic work for artist series? Is it possible to have a series by more than one artist and have artworks within the series that are only associated with one of the artists? In this case, the broader criteria (`(artist:christo OR artist:jeanne-claude)`) would make sense, with the risk that the user removes the artist series later in the flow and creates a very broad criteria.

anandaroop: (Also worth noting that only 5 / 2303 series (0.2%) even have more than one artist, so it's not a hugely impactful decision)

olerichter00: If it's a very rare case, I'd go with adding all artists because it's more intuitive to see all artists in the Create Alert form. But I would also be ok with only adding one artist if this is the preferred pattern when creating alerts in general.

Bildschirmfoto 2024-12-16 um 17 01 58

anandaroop: I think you're right that it's more intuitive to see the two names rather than have one of them disappear.

And in an edge case such as this, the series criteria _itself_ will constrain the results in the desired way (unless as you point it they remove the series at some point).

Let's go with this 👍🏽

PR #15008, opened 2024-12-17
chore(perf): prefer preloading 2x quality images

joeyAghion:
After https://github.com/artsy/force/pull/14974, "main" images are included in "103 early hints" responses and `Link` headers. However, those are the 1x resolution images and we believe most clients end up rendering the 2x images specified in `srcset` attributes. This draft PR starts to return the 2x URL from the `resized` and `cropped` functions, so they can populate the `link.href` property and `Link` headers.

_Still pending:_ I don't fully understand what's going on with artwork images [here](https://github.com/artsy/force/blob/main/src/Apps/Artwork/Components/ArtworkLightbox.tsx): there are local images, resized images, and fallbacks. I'm not sure how to update that to specify 2x images in `<link rel="preload">` tags. If it makes things easier, we only need to render these tags when there's a potential performance benefit to preloading.


joeyAghion:
@matt-z if you have a chance to pair or poke at this independently, the remaining challenge is to adjust [the artwork page's image preloading](https://github.com/artsy/force/blob/0dd8e661679651b2649dd29558ebad1dff629548/src/Apps/Artwork/Components/ArtworkLightbox.tsx#L68-L74) to prefer a 2x source.

damassi (approved):

joeyAghion:
We realized the current hints were inconsistent across mobile and desktop, so Cloudflare’s early hints might also be wrong (depending on the recent requests and edge-cache state).

Since we have to choose one or the other for early hints, we opted to always set `link.href` to the mobile 2x image source. This might cause an unnecessary request form desktop visitors, but we think it's the best practical way to optimize mobile visits.

PR #15012, opened 2024-12-18
fix(dia-1031): fixes double firing pageview on artwork pages on web

dzucconi:
This just removes the extraneous tracking, which I assume was added for some legitimate reason at some point but no longer does anything other than track additionally. This covers all of the scenarios that @mzikherman laid out:

* client-side navigation from a non-artwork page to an artwork page
* client-side navigation from one artwork page to another
* initial SSR navigation to an artwork page

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .env
 .env.*
 !.env.development
+.swc
 .yalc
 *.csv
 *.dat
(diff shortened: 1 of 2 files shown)


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -221,13 +218,6 @@ export const ArtworkApp: React.FC<React.PropsWithChildren<Props>> = props => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
dzucconi, line 218: Tracking ☝️

.gitignore
```diff
@@ -6,6 +6,7 @@
 .env
 .env.*
 !.env.development
+.swc
```
joeyAghion, line 9: 🙇

joeyAghion (approved):

PR #15024, opened 2024-12-19
chore: Update pallete-charts to latest

damassi:

This should fix the bundle duplicates issue.


damassi:
Note how it dropped back down again! subtle things we gotta keep an eye on. It would be the most fabulous thing if other folks on the team took an interest.

PR #15028, opened 2024-12-19
chore: removes palette from debug routes

dzucconi:
Not really going to move the needle much but this is a reasonable change. Should only load route code when the route is requested.

The only other place I noticed this was this strange thing where we redirect to a 404 page but have a 200 status code? https://github.com/artsy/force/blob/f52da8f0d093eed7dde6d700fd95226eb942925f/src/Apps/Order/orderRoutes.tsx#L315-L325 — will look at that separately.

### DIFF
--- a/src/Apps/Debug/debugRoutes.tsx
+++ b/src/Apps/Debug/debugRoutes.tsx
@@ -1,22 +1,31 @@
-import { Clickable, Text } from "@artsy/palette"
 import loadable from "@loadable/component"
-import { RouterLink } from "System/Components/RouterLink"
 import type { RouteProps } from "System/Router/Route"
 import { HttpError } from "found"
-import { useState } from "react"
 
-const DebugApp = loadable(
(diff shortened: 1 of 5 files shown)


src/Apps/Debug/debugRoutes.tsx
```diff
@@ -26,13 +35,11 @@ export const debugRoutes: RouteProps[] = [
     children: [
       {
         path: "baseline",
-        // layout: "LogoOnly",
-        Component: DebugApp,
+        Component: DebugBaselineRoute,
```
damassi, line 38: @dzucconi - part of the issue is that this isn't using `getComponent`, which invokes a promise and executes only when matching the route. Right now this loads the bundle statically, at boot.  Ditto elsewhere.

damassi: https://github.com/artsy/force/blob/main/src/Apps/Artwork/artworkRoutes.tsx#L18

PR #15034, opened 2024-12-20
fix(artwork): correctly tracks on artwork => artwork nav

dzucconi:
Fixed tracking for artwork-to-artwork navigation alongside the other criteria @mzikherman noted. Issue was resolved by adding track to the dependency array and separating it into its own effect.


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -217,9 +217,12 @@ export const ArtworkApp: React.FC<React.PropsWithChildren<Props>> = props => {
       // this.props.router.replace(this.props.match.location.pathname)
       silentPush(props.match.location.pathname)
     }
-    track()
   }, [])
 
+  useEffect(() => {
+    track()
+  }, [track])
+
```
damassi, line 225: 🤯 no way

damassi (approved):

PR #15039, opened 2024-12-20
feat(start): Add --open cli argument

damassi:

### Description

After some trial, this removes the default open browser on start behavior in favor of a simple cli argument:
- `yarn start --open` -- Falls back to default localhost:PORT url 
- `yarn start --open=http://foo.bar` -- pass in a specific url

cc @artsy/diamond-devs

### DIFF
--- a/package.json
+++ b/package.json
@@ -185,7 +185,7 @@
     "use-deep-compare-effect": "^1.2.0",
     "uuid": "3.3.2",
     "web-vitals": "^4.2.4",
-    "yargs": "13",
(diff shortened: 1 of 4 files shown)


anandaroop (dismissed):
Nice

erikdstock (approved):
Seems good to me. Non-blocking: Are these args self-documenting or will i need to know about `--open`?

damassi:
Great question @erikdstock. I wanted to keep it simple for now, but I was actually about to update the README so will include it there 👍

PR #15041, opened 2024-12-20
docs(setup): Update setup readme docs and setup.sh script

damassi:
The type of this PR is: **Docs**

### Description

Updated our README and setup script with the latest.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -13,6 +13,15 @@ else
   echo "Skipping language dependencies installation (asdf not found)"
 fi
 
+if command -v code >/dev/null; then
+  echo "Installing Artsy VSCode extensions..."
+  code --install-extension Artsy.artsy-studio-extension-pack
+  # TODO: Move this into artsy-studio-extension-pack
+  code --install-extension biomejs.biome
+else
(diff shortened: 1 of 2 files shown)


scripts/setup.sh
```diff
@@ -13,6 +13,15 @@ else
   echo "Skipping language dependencies installation (asdf not found)"
 fi
 
+if command -v code >/dev/null; then
+  echo "Installing Artsy VSCode extensions..."
+  code --install-extension Artsy.artsy-studio-extension-pack
+  # TODO: Move this into artsy-studio-extension-pack
```
damassi, line 19: I tried, but I can't figure out for the life of me how to authenticate properly, have tried and tried. I know @MounirDhahri was there recently, so looking for any help 🙏

anandaroop (approved):
👍🏽

scripts/setup.sh
MounirDhahri: Sorry for the late reply here. I opened a PR here https://github.com/artsy/vscode-artsy/pull/13

MounirDhahri: Authentication here was pretty weird that's why I specified the email last time

PR #15046, opened 2024-12-24
chore:  open action prompt in a new window instead of the modal for order actions in conversations

oxaudo:
This PR solves [https://artsyproduct.atlassian.net/browse/EMI-2218]

This is just removing the modal and opening the appropriate order in the new tab regardless of the state of the order. We currently have this modal opened for multiple states (counteroffer received in different variations, order approved, order payment failing, etc). So I applied this 'open in the new window' across the board. 

Need to discuss with Cami and Vivi what is the appropriate flow we want at the end but for now confirmed with Cami that it's good enough of a change. For now don't want to merge this immediately as we want to have a bit of time to test it before putting it out (mainly make sure there is no major objections to this flow).

So if all looks well - we should merge it when more people from the team is around and can play with it.

### DIFF
--- a/src/Apps/Conversations/components/ConversationCTA/ConversationReviewOfferCTA.tsx
+++ b/src/Apps/Conversations/components/ConversationCTA/ConversationReviewOfferCTA.tsx
@@ -1,4 +1,8 @@
-import { ActionType, OwnerType, type TappedViewOffer } from "@artsy/cohesion"
+import {
+  ActionType,
+  OwnerType,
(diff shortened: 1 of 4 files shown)


MrSltun (approved):

oxaudo:
@artsy/emerald-devs - I'm removing 'do not merge' from here and maybe this can be merged and tested on Jan 2nd in the new year! I'll be out for couple of days there but don't anticipate any issues here.

PR #15053, opened 2025-01-06
fix: Guard against unnecessary timer rerenders on /artwork/id for closed auction artworks

damassi:

### Description

This fixes an issue where we were unnecessarily rerendering the artwork page even if an auction artwork has closed by:
- Checking if the artwork is still open auction-wise
- adjusting the function signature of `useTimer` and  `useCurrentTime` by adding an additional `enabled` prop
- adding an `enabled` prop to the AuctionWebsocket context and checking if we should enable it

Additionally:
- fixes `yarn scan` by downgrading the lib to a working version ([something broke in the latest](https://github.com/aidenybai/react-scan/issues/162), which made it so that we couldn't see rerenders. Fixed it `0.0.39`
- bumped rsbuild because there's some memory improvements. Triggered an OOM issue after all of the saves trying to debug this. 

cc @artsy/diamond-devs


src/Apps/Artwork/ArtworkApp.tsx
```diff
@@ -346,8 +354,10 @@ const WrappedArtworkApp: React.FC<React.PropsWithChildren<Props>> = props => {
   // Check to see if referrer comes from link interception.
   // @see interceptLinks.ts
   const referrer = state && state.previousHref
+  const isLotClosed = lotIsClosed(sale, saleArtwork)
 
-  const websocketEnabled = !!sale?.extendedBiddingIntervalMinutes
+  const websocketEnabled =
+    !!sale?.extendedBiddingIntervalMinutes && isLotClosed === false
```
joeyAghion, line 360: Haven't totally grokked this, but does this imply that a sale without extended bidding will always skip polling?

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -71,8 +72,13 @@ export const ArtworkSidebarCommercialButtons: React.FC<
       extractNodes(me.partnerOffersConnection)[0]) ||
     null
 
+  const isLotClosed = lotIsClosed(artwork.sale, artwork.saleArtwork)
+
   // Fall back to a definitely past value because the timer hook doesn't like nulls
-  const partnerOfferTimer = useTimer(partnerOffer?.endAt || THE_PAST)
+  const partnerOfferTimer = useTimer({
+    endDate: partnerOffer?.endAt || THE_PAST,
+    enabled: !isLotClosed,
```
joeyAghion, line 80: Shouldn't this be based on the offer rather than lot status?

joeyAghion (commented):
Like you I was surprised this was so tricky. Left a few questions but am not familiar enough with this logic to guide.

src/Apps/Artwork/ArtworkApp.tsx
damassi: This should probably be ||

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
damassi: Just set up so that if the lot is closed, we don't poll. I'm not too sure beyond that.

src/Utils/Hooks/useCurrentTime.tsx
```diff
@@ -20,11 +20,13 @@ import { useEffect, useRef, useState } from "react"
 interface UseCurrentTimeProps {
   interval?: number
   syncWithServer?: boolean
+  enabled?: boolean
 }
 
 export const useCurrentTime = ({
-  interval = 1000,
+  interval = 5000,
```
dzucconi, line 27: This is used to display  so at the very least, that usage of it needs the 1 second interval

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
dzucconi: It seems to me that we can just have use timer disable itself once it reaches the end date. And then we wouldn't need to explicitly enable or disable in these cases.

damassi:
Gonna close this because in the end it had no impact on INP. Thanks for the detailed comments everyone; we can return to this if / when we want to optimize.

PR #15055, opened 2025-01-07
chore(EMI-2213): client side signals tracking

rquartararo:

This PR solves [EMI-2213]

### Description

This updates the signals tracking to work with client-side loading. 

My solution was to use the grid context to get accurate data for client-side signal tracking.

**Review with most recent changes:**

https://client-side-signals-with-tracking.artsy.net/

[EMI-2213]: https://artsyproduct.atlassian.net/browse/EMI-2213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Components/ArtworkGrid/ArtworkGridContext.tsx
```diff
@@ -15,20 +15,41 @@ interface ArtworkGridContextProps {
   saveOnlyToDefaultList?: boolean
 
   hideSignals?: boolean
+
+  signals?: { [id: string]: string[] }
+
+  updateSignals?: (id: string, newSignals: string[]) => void
```
starsirius, line 21: To summarize the approach, the context maintains a `signals` data structure that keeps track of all the artworks and their individual collector signals (which can be updated client side in the `PrimaryLabelLine` component). In the grid/rail item `onClick` callback, we'll refer to this `signals` map to fire tracking calls with the correct signals (which may or may not be overwritten on the client side).

src/Components/Artwork/Details/PrimaryLabelLine.tsx
```diff
@@ -1,23 +1,35 @@
...
+  useEffect(() => {
+    if (updateSignals && artwork) {
+      const signals: string[] = []
+      if (partnerOffer) {
+        signals.push("PARTNER_OFFER")
+      }
+      if (label) {
+        signals.push(label)
+      }
+      updateSignals(artwork.internalID, signals)
```
starsirius, line 29: This is the only place we update signals (for tracking purposes). I think it makes sense because this component represents the display of the collector signals to users. We might have to keep in mind if in the future the visual or the exact component implementation changes, we'll update the tracking accordingly.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarCommercialButtons.tsx
```diff
@@ -697,7 +691,11 @@ const ARTWORK_FRAGMENT = graphql`
       }
     }
     collectorSignals {
-      primaryLabel(ignore: [PARTNER_OFFER])
+      partnerOffer {
+        endAt
+      }
+      increasedInterest
+      curatorsPick
```
starsirius, line 698: Now on the MP schema side, we separate out the fields requiring auth (e.g. `partnerOffer`) from fields without requiring auth (e.g. `primaryLabel`), we'll need to request them both.

src/Apps/Collect/Routes/Collect/index.tsx
```diff
@@ -84,76 +85,80 @@ export const CollectApp: React.FC<React.PropsWithChildren<CollectAppProps>> = ({
...
-                      counts @include(if: $shouldFetchCounts) {
-                        followedArtists
-                      }
-                      aggregations {
-                        slice
-                        counts {
-                          value
-                          name
-                          count
+          <ArtworkGridContextProvider>
```
starsirius, line 88: I'm not sure if there is a difference wrapping the context provider here vs. one level up from `<ArtworkFilter>` below in the render function, but, intuitively, staying closer to the consumer of it seems more direct.

src/Apps/Home/HomeApp.tsx
```diff
@@ -39,9 +39,8 @@ export const HomeApp: React.FC<React.PropsWithChildren<HomeAppProps>> = ({
       <HomeHeroUnitsFragmentContainer heroUnits={heroUnitsConnection} />
 
       <Spacer y={[4, 6]} />
-
-      <Join separator={<Spacer y={[6, 12]} />}>
-        <ArtworkGridContextProvider>
+      <ArtworkGridContextProvider>
+        <Join separator={<Spacer y={[6, 12]} />}>
```
starsirius, line 43: 👍

starsirius (approved):
LGTM! Thanks for organizing the QA and working through all the back and forth patiently!! 👏

PR #15063, opened 2025-01-08
fix: include initial input/sort when prefetching artist artworks

anandaroop:

This PR solves [ONYX-1475]

### Description

When prefetching an artists's artwork grid content, we are correctly requesting the prepared variables, but  we are not passing the prepared `input` into the fragments as needed. Thus the `{ input : { sort: … } }` argument is missing, which explains the recent reports of unsorted artworks in the grid.

The nature of this bug meant that it was apparent only when client-side navigating to a prefetched artist page (e.g. from an artwork page by clicking on the artist's name).  A direct navigation or hard-refresh would not have repro'd the bug. 

[ONYX-1475]: https://artsyproduct.atlassian.net/browse/ONYX-1475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artist/artistRoutes.tsx
+++ b/src/Apps/Artist/artistRoutes.tsx
@@ -144,12 +144,14 @@ export const artistRoutes: RouteProps[] = [
             $artistID: String!
             $isPrefetched: Boolean!
             $aggregations: [ArtworkAggregation]
+            $input: FilterArtworksInput
           ) {
             artist(id: $artistID) @principalField {
               ...ArtistWorksForSaleRoute_artist
                 @arguments(
                   isPrefetched: $isPrefetched
                   aggregations: $aggregations
+                  input: $input
                 )
             }
           }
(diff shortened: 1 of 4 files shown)


src/Apps/Artist/artistRoutes.tsx
```diff
@@ -144,12 +144,14 @@ export const artistRoutes: RouteProps[] = [
             $artistID: String!
             $isPrefetched: Boolean!
             $aggregations: [ArtworkAggregation]
+            $input: FilterArtworksInput
```
anandaroop, line 147: The `prepareVariables` config above on L141 ensures that we [have a proper default](https://github.com/artsy/force/blob/3f74e47f4a92d3fe5f57fdfdd907b5d74e980cf3/src/Apps/Artist/Routes/WorksForSale/Utils/getWorksForSaleRouteVariables.ts#L9) for the `input` argument, but we were not passing it down as needed.

joeyAghion (approved):

damassi:
Thanks @anandaroop 👍

PR #15064, opened 2025-01-08
chore: tag review app CNAME records accordingly, to make automatic clean-up safe (later)

joeyAghion:
[Cloudflare docs](https://developers.cloudflare.com/api/resources/dns/subresources/records/methods/create/) indicate that an array of tags (`:`-delimited strings) can be set when creating DNS records. Let's add a tag on any records created automatically by this script to support review apps. That should make it possible to clean up these records later on without too much risk of breaking production.

Caveat: I haven't run this script but errors should surface upon the next review app creation.


starsirius:
🙏

PR #15070, opened 2025-01-13
fix(xss): removes usage of buggy url sanitization

dzucconi:
This change switches to the more secure URL sanitization approach that is used for redirects during authentication. The function uses Node's URL parsing, only permitting internal paths or allowed hosts. While this means external URLs will no longer function, we don't have a use case for supporting external links.

~~Upon review, I realize this implementation is currently only server-side. The `url` import's bundle size impact is unclear, let's open this PR and see the diff. If significant I'll switch to using Braintree's [sanitize-url](https://github.com/braintree/sanitize-url/blob/main/src/index.ts) library instead.~~ Not an issue.


cavvia:
Thanks @dzucconi!

PR #15080, opened 2025-01-14
fix(EMI-2036): use last submitted offer for new payment route

leamotta:

This PR solves [EMI-2036]

### Description

Use last submitted offer in new payment route to avoid having the buyer see his old offer when fixing the payment. Checked also other uses of the component.

Before:
Screenshot 2025-01-14 at 14 25 00
After:
Screenshot 2025-01-14 at 14 24 37


MrSltun (approved):

starsirius:
🙏

PR #15085, opened 2025-01-15
fix: add additional line to the artwork metadata placeholder

oxaudo:
This PR solves [EMI-2256] (https://artsyproduct.atlassian.net/browse/EMI-2256)

### Description

This seems to be too easy of a fix but seems to work for me on all the screens that I was checking. The thing I'm not sure about if we need to make this line conditional. Like is there rails that would never include this 5th line?

Also remember @starsirius that you said that you've hardcoded the hight in some rail and wonder if we should clean that up and see if things are still working.

[EMI-2256]: https://artsyproduct.atlassian.net/browse/EMI-2256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/ArtworkGrid/ArtworkGrid.tsx
+++ b/src/Components/ArtworkGrid/ArtworkGrid.tsx
@@ -85,8 +85,11 @@ export class ArtworkGridContainer extends React.Component<
 
   private get _columnCount(): number[] {
     const columnCount = this.props.columnCount
-    // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-    return typeof columnCount === "number" ? [columnCount] : columnCount
+
+    if (typeof columnCount === "number") {
+      return [columnCount]
+    }
+    return Array.isArray(columnCount) ? columnCount : []
   }
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


src/Components/ArtworkGrid/ArtworkGrid.tsx
```diff
@@ -85,8 +85,11 @@ export class ArtworkGridContainer extends React.Component<
 
   private get _columnCount(): number[] {
     const columnCount = this.props.columnCount
-    // @ts-expect-error PLEASE_FIX_ME_STRICT_NULL_CHECK_MIGRATION
-    return typeof columnCount === "number" ? [columnCount] : columnCount
+
+    if (typeof columnCount === "number") {
+      return [columnCount]
+    }
+    return Array.isArray(columnCount) ? columnCount : []
```
starsirius, line 92: Are the changes here auto generated?

starsirius (approved):
LGTM! 🎉

src/Components/ArtworkGrid/ArtworkGrid.tsx
oxaudo: I was trying to remove the easy todo's that I've noticed and that is what I ended up with.

PR #15099, opened 2025-01-21
fix(dia-1058): avoid extraneous app download ctas on homepage

dzucconi:
This implementation disables the top download banner specifically on the homepage, which adjusts the global nav height accordingly. The footer banner is also disabled on the homepage only. The system treats homepage visits as standard route changes—the banner will hide when arriving on the homepage if previously displayed, then reappear when navigating away via route change.

### DIFF
--- a/src/Components/AppDownloadFooter.tsx
+++ b/src/Components/AppDownloadFooter.tsx
@@ -16,7 +16,7 @@ import { useRouter } from "System/Hooks/useRouter"
 import { useSystemContext } from "System/Hooks/useSystemContext"
 import { useDeviceDetection } from "Utils/Hooks/useDeviceDetection"
 import Cookies from "cookies-js"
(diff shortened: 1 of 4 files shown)


src/Components/AppDownloadFooter.tsx
```diff
@@ -149,3 +123,28 @@ const AppDownloadFooterPanel = styled(Box)`
...
+}) => {
+  const ref = useRef<HTMLDivElement | null>(null)
+
+  useEffect(() => {
+    if (!ref.current || !("animate" in ref.current)) return
+
+    ref.current.animate(
+      [{ transform: "translateY(100%)" }, { transform: "translateY(0)" }],
+      { duration: 250, easing: "ease-out", fill: "forwards" },
+    )
```
damassi, line 142: Nice!

PR #15106, opened 2025-01-22
feat(hackathon): ships free signal 

erikdstock:
The type of this PR is: **feat** and it is part of Hackathon 15. Paired with @lilyfromseattle.

### Description
This PR adds a collector signal for free shipping which takes user location into account. It appears just below Limited-time offer in priority.

It also has a small hook that distills shipping and pickup availability from different options which could be useful in other contexts.


src/Components/Artwork/Details/PrimaryLabelLine.tsx
```diff
@@ -1,27 +1,44 @@
...
   label: string | null | undefined
   artwork?: PrimaryLabelLine_artwork$data
+  me?: PrimaryLabelLine_me$data
+}
+
+const doesShipFree = (
+  artwork: PrimaryLabelLine_artwork$data,
+  me?: PrimaryLabelLine_me$data,
+): boolean => {
+  const result = useFulfillmentOptionsForArtwork(artwork, me)
```
damassi, line 21: This doesn't seem right as this manner of using a hook in a function breaks the rules of hooks

src/Components/Artwork/Details/PrimaryLabelLine.tsx
```diff
@@ -1,27 +1,44 @@
...
+  const { freeShippingToUserCountry, freeGlobalShipping } = result || {}
+  return !!(freeShippingToUserCountry || freeGlobalShipping)
```
damassi, line 23: Can this logic be cleaned up a bit? there's fallbacks and boolean casts and then fallbacks on the option plucked from a fallback null object (or result)

src/Components/Artwork/Details/PrimaryLabelLine.tsx
```diff
@@ -1,27 +1,44 @@
...
 }
 
 export const PrimaryLabelLine: React.FC<
   React.PropsWithChildren<PrimaryLabelLineProps>
-> = ({ label, artwork }) => {
+> = ({ label, artwork, me }) => {
   const { hideSignals, updateSignals } = useArtworkGridContext()
   const partnerOffer = artwork?.collectorSignals?.partnerOffer
+  const shipsFree = artwork && doesShipFree(artwork, me)
```
damassi, line 31: we should check for artwork existence inside the function, which is already doing a lot of work. Return a false value from there if artwork doesn't exist.

src/Utils/Hooks/useFulfillmentOptions.ts
```diff
@@ -0,0 +1,156 @@
...
+  const artworkLocation: FulfillmentOptions["artworkLocation"] = {
+    countryCode: countryCode,
+  }
+
+  const domesticShipping: FulfillmentOptions["domesticShipping"] =
+    artsyShippingDomestic
+      ? "ARTSY_SHIPPING"
+      : domesticShippingFee
+        ? domesticShippingFee
+        : null
```
damassi, line 112: We should always avoid double ternaries as they're hard to read for those who didn't write the original code

src/Utils/Hooks/useFulfillmentOptions.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+  pickupAvailable: boolean
+  onlyShipsDomestically: boolean
+  domesticShipping: ShippingUnion
+  internationalShipping: ShippingUnion
+  freeGlobalShipping: boolean
+  freeShippingToUserCountry?: boolean
+}
+
+export const useFulfillmentOptionsForArtwork = (
```
damassi, line 56: @erikdstock - We've gone over this a few times in the past, but to rehash that discussion, we should avoid using hooks to fetch relay data in this way unless strictly necessary. Its not the worst thing in the world, but its a uncommon idiom and tricky due to data masking at the props / component level

src/Components/Artwork/Details/PrimaryLabelLine.tsx
erikdstock: oh good point. I was just trying to get it working since if implemented we would probably move at least some of this logic upstream to MP/Gravity

src/Utils/Hooks/useFulfillmentOptions.ts
erikdstock: If we really implemented this feature I think we would move the data loading up into a MP type or even gravity fields. I did it here specifically for hackathon POC purposes and to have the option of quick collaboration with the work @oxaudo is doing to simplify some incorrect shipping messaging - I wanted a helper to encapsulate that logic in one place, and for temporary code this seemed better and more readable than writting a mirror of a graphql type.

If it becomes clear that we want to actually implement something like this or if I'm able to finish another cleanup PR i have in review I'll come back to this but regardless I'll nothing like this will merge.

src/Utils/processFulfillmentOptions.ts
```diff
@@ -0,0 +1,154 @@
...
+  readonly internationalShippingFee:
+    | {
+        readonly minor: any
+      }
+    | null
+    | undefined
+  readonly isPurchasable: boolean | null | undefined
+  readonly onlyShipsDomestically: boolean | null | undefined
+  readonly pickupAvailable: boolean | null | undefined
+}
```
erikdstock, line 69: responded to feedback by removing this which was previously a hook with a fragment. I just copied the generated type in but if we actually merged this (I think we shouldn't, this logic should go up to MP/Gravity) we might want to restrict the types in some other way to make sure the user queries the full data necessary (since everything is nullable)

src/Components/Artwork/Details/PrimaryLabelLine.tsx
```diff
@@ -14,21 +14,15 @@ interface PrimaryLabelLineProps {
...
 export const PrimaryLabelLine: React.FC<
   React.PropsWithChildren<PrimaryLabelLineProps>
 > = ({ label, artwork, me }) => {
   const { hideSignals, updateSignals } = useArtworkGridContext()
   const partnerOffer = artwork?.collectorSignals?.partnerOffer
-  const shipsFree = artwork && doesShipFree(artwork, me)
+
+  const { freeShippingToUserCountry, freeGlobalShipping } =
+    processArtworkFulfillmentOptions(artwork, me) || {}
+  const shipsFree = freeShippingToUserCountry || freeGlobalShipping
```
damassi, line 25: Not trying to be finnicky here with the feedback, but just to note that the only reason this complexity is necessary is due to destructuring. It can be simplified (we don't always have to destructure):

```tsx
const fulfillmentOptions = processArtworkFulfillmentOptions(artwork, me)

const shipsFree = fulfillmentOptions?.freeShippingToUserCountry || fulfilmentOptions.freeGlobalShipping 
```

src/Utils/processFulfillmentOptions.ts
damassi: I know its hackathon and hacks are fine, but using the hook is probably better than copying out types from a generated fragment. (That feedback was just encouragement to try to obtain that relay data via the component fragment, vs a hook that's spread.)

PR #15110, opened 2025-01-23
feat(hackathon): shareable artwork lists

dzucconi:
This implements a basic UI for sharing and publishing collections. The share button auto-publishes the collection. The new  route is accessible to logged-out users, making collections shareable. Still need to add pagination to make it fully functional, but wanted to get an initial version deployed.

The "Share" button is enabled behind the `diamond_shareable-artwork-lists` flag. The user nested route is unprotected.

## TODO

- [x] Pagination for artwork grid
- [ ] Shareable "My Collection"
- [ ] Ability to unpublish a collection

https://github.com/user-attachments/assets/17273124-662d-451f-b3b3-8ae0e6429326


dzucconi:
OK, added pagination too.

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e227baa75148b92a73a25e06c54f36ea25c0dfbe
</p>

src/Apps/CollectorProfile/Routes/Saves/Components/SavesArtworksShareDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+      sendToast({
+        message: "Failed to share collection",
+        variant: "error",
+      })
+    }
+  })
+
+  if (!user) return null
+
+  const url = `${getENV("APP_URL")}/user/${user.id}/collection/${collectionId}`
```
mzikherman, line 89: Could be cool to use a slug instead of id here (after the mutation completes), maybe

dzucconi: Yeah, will do

src/Apps/CollectorProfile/Routes/Saves/Components/SavesArtworksShareDialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+    setMode("Updating")
+
+    try {
+      await submitMutation({
+        variables: {
+          input: {
+            id: collectionId,
+            name: collectionName, // TODO: Shouldn't be required
+            private: false,
+            shareableWithPartners: true,
```
anandaroop, line 72: Over on Onyx we noticed that sharing a list also [caused it to be shared with partners](https://artsy.slack.com/archives/C05EQL4R5N0/p1741717600477789), which makes sense given the mutation input here.

We feel that is surprising and possibly undesirable behavior since it opens the collector to offers from partners — something that should instead be controlled by the existing toggle:



Gonna PR a change to remove L72 (which I think should leave the value of `shareableWithPartners` untouched, right?)

Holler if you have any concerns.

dzucconi: Ah, makes sense. Thanks!

PR #15126, opened 2025-01-28
feat(EMI-2164): Artsy shipping estimate widget - initial implementation

erikdstock:

This PR solves [EMI-2164]

### Description

This PR is the initial implementation of the shipping estimate widget to be shown in the shipping & taxes section of an artwork page. It does not contain widget styling and will need some more changes to work fully.

_Note on context provider_
Due to specifics of how we load the browser-only shipping estimate module, it is not available at the moment the page renders. This makes the loading async, a problem that snowballs because the widget is hidden in a Collapsing element. My initial solution was to create a whole context to load the module outside of the Collapse, but there is still an async-ness to loading the widget link itself, so it might be worth removing this extra complexity for the two separate layers of loading.

A/C
- [x] Attempt to load the widget on artworks with global (domestic/international) artsy shipping enabled
- [x] Support artworks with no edition sets
- [ ] There are still `console.log()s` in the code - will remove as I finish some of the investigations below
- [ ] If required artwork metadata is missing, render a normal fallback shipping availability message _(partially complete, want a little more of a guard)_
- [ ] Tests

Unknowns & followup work
- [ ] Depends on artsy/metaphysics#6367
- [ ] artwork weight may be required
- [ ] artwork location extraction is hacky
- [ ] Support artworks with a single edition set, but not multiple (may be added to a future ticket)


src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,366 @@
...
+    }
+    priceCurrency
+    priceListed {
+      major
+    }
+    heightCm
+    widthCm
+    # shippingWeight # may need to be added?
+  }
+`
```
erikdstock, line 208: Problems with our schema:
- `category` is deprecated but used to map to values required for estimate
- weight not accessible
- location not accessible, so using `shippingOrigin`
- `priceListed` does not include `major` (see https://github.com/artsy/metaphysics/pull/6367)

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -1,28 +1,30 @@
...
     shippingOrigin,
     shippingInfo,
     pickupAvailable,
     taxInfo,
-  },
-}) => {
+  } = artwork
   const { trackEvent } = useTracking()
+  const globalArtsyShipping =
+    !!artwork.artsyShippingDomestic || !!artwork.artsyShippingInternational
```
oxaudo, line 27: Don't think this is an accurate condition that we want here. Believe it has to be `!!artwork.artsyShippingDomestic && (!!artwork.artsyShippingInternational || (!artwork.artsyShippingInternationa && !artwork.international_shipping_fee))`. So either arta shipping both domestic and international or arta domestic and international not configured.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -1,28 +1,30 @@
...
 import { createFragmentContainer, graphql } from "react-relay"
-
-import { ActionType, type ClickedOnLearnMore } from "@artsy/cohesion"
-import { RouterLink } from "System/Components/RouterLink"
 import { useTracking } from "react-tracking"
 
 export interface ShippingInformationProps {
   artwork: ArtworkSidebarShippingInformation_artwork$data
 }
+const CALCULATED_IN_CHECKOUT_MESSAGE = "Shipping: Calculated in checkout"
```
oxaudo, line 12: This is not the case any more after this pr: https://github.com/artsy/metaphysics/pull/6383

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -72,7 +80,7 @@ const ArtworkSidebarShippingInformation: React.FC<
         </Text>
       )}
 
-      {!!shippingInfo && (
+      {!!shippingInfo && !globalArtsyShipping && (
```
oxaudo, line 83: should we account for the unleash feature flag here so we do not even try to render estimates when feature is not enabled?

oxaudo (commented):
Those are immediate comments but maybe worth pairing on it tomorrow AM.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
```diff
@@ -101,142 +102,144 @@ export const ArtworkSidebar: React.FC<
...
 
   return (
-    <Flex flexDirection="column" data-test={ContextModule.artworkSidebar}>
-      {shoudlDisplayLotLabel && (
-        <Text variant="sm" color="black100" mb={0.5}>
-          Lot {lotLabel}
-        </Text>
-      )}
-      <ArtworkSidebarArtistsFragmentContainer artwork={artwork} />
+    <ArtsyShippingEstimateProvider>
```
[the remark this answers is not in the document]
erikdstock: I'll remove the provider, agree it's not needed here

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
erikdstock: Let's follow up on this in person, i believe you have the expertise and it's a small change if so

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
erikdstock: Removing for now bc designs don't need it

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,292 @@
...
+    }
+  }
+  return null
+}
+
+type ArtworkValue = Pick<ArtaObject, "value" | "value_currency">
+const artworkValue = (artwork: ShippableArtwork): ArtworkValue | null => {
+  if (!!artwork.heightCm && !!artwork.widthCm) {
+    return {
+      value: artwork.priceListed?.major || 500, // TODO: get a real value
```
starsirius, line 239: I think `price_listed` is the backend price that is [not exposed publicly](https://github.com/artsy/gravity/blob/f99e2848373c795f1b06d6eacbba219ca4855729/app/models/domain/artwork.rb#L1379) so I guess users will not be able to see this. Also, do we aim to support artworks displayed as price range, etc? (sorry if I missed the product requirements)

oxaudo: Good point @starsirius! I believe there was not a discussion about excluding range works. And I forgot about this whole 'price being private' for MO sometimes. Think taking midpoint makes sense here but def something to discuss.

starsirius: Perhaps [this is the right field](https://github.com/artsy/metaphysics/blob/4d5fab5395d127f5d4de0470b62a80e219455ee3/src/schema/v2/artwork/index.ts#L1274) for public facing pricing. Confusing indeed!

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
```diff
@@ -200,7 +210,16 @@ export const ArtworkSidebar: React.FC<
 
       {!isUnlisted && !isSold && artworkEcommerceAvailable && (
         <>
-          <SidebarExpandable label="Shipping and taxes">
+          <SidebarExpandable
+            label={
+              <Flex flexDirection="column" justifyContent="flex-start">
+                <Text>Shipping and taxes</Text>
+                {globalArtsyShipping && artsyShippingEstimateEnabled && (
```
oxaudo, line 217: Like that all the guarding is scoped to this one place! Thanks for the update.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
```diff
@@ -200,7 +210,16 @@ export const ArtworkSidebar: React.FC<
...
+                  <ArtsyShippingEstimate artwork={artwork} />
+                )}
+              </Flex>
+            }
+          >
             <ArtworkSidebarShippingInformationFragmentContainer
```
oxaudo, line 223: As discussed in person - we are not hiding the actual shippingInfo line in expanded section of this view. We can definitely do it - just don't think it's adding anything to the test and believe keeping code less intermingled with the estimate test here is a good goal. Might still be worth confirming out assumption with Cami and Vivi.

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebarShippingInformation.tsx
```diff
@@ -122,6 +121,8 @@ export const ArtworkSidebarShippingInformationFragmentContainer =
         isUnlisted
         priceIncludesTaxDisplay
         shippingOrigin
+        artsyShippingDomestic
```
oxaudo, line 124: Maybe worth cleaning this up as those are not needed in this component now that we've moved the check into the parent component.

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,348 @@
...
+      setState({ loaded: true, widget: null })
+      return
+    }
+
+    if (state.loaded || !Arta) {
+      return
+    }
+
+    const loadEstimate = async () => {
+      console.log("***", "oading estimate", estimateInput)
```
oxaudo, line 84: 🔪

oxaudo (approved):
I feel good now if we merge this and continue iterating. Comments are mostly to remember where some clean up needed.

oxaudo:
Spec fails are legit though and should be fixed before merging.

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+const Loader = () => <Spacer y={2} />
+
+const LinkButton = styled(Link)`
+  color: ${props => props.theme.colors.black60};
+  &:hover {
+    text-decoration: underline;
+    color: ${props => props.theme.colors.black100};
+  }
+`
```
damassi, line 39: We should move all of this secondary code below the main component; its typically just noise and rarely updates

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,348 @@
...
+const WIDGET_TITLE = "Estimate Shipping Cost"
+
+type Arta = typeof import("@artaio/arta-browser").default
+
+type EstimateWidgetState = {
+  loaded: boolean
+  widget: ArtaEstimate | null
+}
+
+interface Props {
```
damassi, line 27: Idiomatically, `ArtsyShippingEstimateProps`

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,348 @@
...
+    }
+    loadEstimate()
+  }, [Arta, estimateInput, state.loaded, state])
+
+  if (!state.loaded) {
+    return <Loader />
+  }
+
+  if (!state.widget) {
+    return null
```
damassi, line 116: Maybe also return loader here?

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  const estimateWidget = state.widget
+
+  return estimateWidget.isReady ? (
```
damassi, line 121: we should avoid this ternary form of return. 

```tsx
if (!estimateWidget.isReady) {
  return <Loader />
}

return (
  ...
)
```

damassi (approved):
The current refactor / simplification is much improved 👍

src/Components/ArtsyShippingEstimate.tsx
erikdstock: If it is already loaded and there is no widget, then something failed and/or the artwork isn't able to do a shipping estimate. So in this case we fall back to nothing. 

... though technically the loader is now just a spacer instead of skeleton text so we could keep that to avoid a content jump.

package.json
```diff
@@ -189,6 +189,7 @@
     "yup": "0.32.6"
   },
   "devDependencies": {
+    "@artaio/arta-browser": "^2.16.1",
```
erikdstock, line 192: Added here for the types - the lib is downloaded by script.

PR #15128, opened 2025-01-29
Revert "feat(performance): Add INP optimization strategy for mobile web"

damassi:
Reverts artsy/force#15096.

Turns out this INP hack/fix didn't actually improve things in google search console, so lets revert and avoid weird hacks that nobody will understand a few years down the line 😢


nickskalkin (approved):
sad

PR #15129, opened 2025-01-29
fix(EMI-2244): fix price null for express element

rquartararo:

This PR solves [EMI-2244]

### Description
Don't return the express element if `buyerTotalCents` is null and moves spacing so it doesn't look weird if it's not rendered. 

[EMI-2244]: https://artsyproduct.atlassian.net/browse/EMI-2244?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/ExpressCheckoutPrototype/ExpressCheckout.tsx
+++ b/src/Apps/Order/Routes/Payment/ExpressCheckoutPrototype/ExpressCheckout.tsx
@@ -33,6 +33,7 @@ export const ExpressCheckout = () => {
       <Text variant="lg-display">Express checkout</Text>
       <Spacer y={1} />
       <ExpressCheckoutElement onConfirm={onConfirm} />
+      <Spacer y={4} />
     </>
   )
 }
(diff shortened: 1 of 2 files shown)


src/Apps/Order/Routes/Payment/ExpressCheckoutPrototype/ExpressCheckoutProvider.tsx
```diff
@@ -6,6 +6,10 @@ import { getENV } from "Utils/getENV"
 export const ExpressCheckoutProvider = ({ order }) => {
   const { buyerTotalCents } = order
 
+  if (!buyerTotalCents) {
+    return null
+  }
```
starsirius, line 11: Looks like [this](https://github.com/artsy/force/blob/db4d317d99e9fb6166f9a57345632d67a8204bc1/src/Apps/Order/Components/TransactionDetailsSummaryItem.tsx#L223-L232) might be the current way of displaying buyer total for both BN and MO order at the payment step.

PR #15131, opened 2025-01-30
chore(EMI-2277): add tests for shipping estimate widget

erikdstock:
Resolves [EMI-2277]

### Description
This PR adds some tests for the shipping widget. Due to the way we load it from a script tag the tests are a little strange; maybe there is a better way. We have the library in our dev dependencies for example, but i'm just mocking out the entire lib with `jest.fn()`s.

We can add more tests or things for the specific complex estimate params creation as we go - wanted to get a solid base to work off.

It also corrects two minor bugs not visible outside staging.

cc @artsy/emerald-devs 

[EMI-2277]: https://artsyproduct.atlassian.net/browse/EMI-2277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/ArtsyShippingEstimate.tsx
+++ b/src/Components/ArtsyShippingEstimate.tsx
@@ -63,12 +63,11 @@ export const ArtsyShippingEstimate = ({
   }, [state.widget])
 
   useEffect(() => {
-    if (!estimateInput || !ARTA_API_KEY) {
-      setState({ loaded: true, widget: null })
+    if (state.loaded || !Arta) {
       return
     }
-
-    if (state.loaded || !Arta) {
+    if (!estimateInput || !ARTA_API_KEY) {
+      setState({ loaded: true, widget: null })
       return
     }
 
@@ -104,7 +103,7 @@ export const ArtsyShippingEstimate = ({
   }
 
   if (!state.widget) {
-    return <Spacer y={2} />
+    return <Spacer data-testid="loaded-no-widget" y={2} />
   }
 
   const estimateWidget = state.widget
@@ -116,7 +115,10 @@ export const ArtsyShippingEstimate = ({
   return (
     <LinkButton
       tabIndex={0}
-      onClick={() => estimateWidget.open()}
+      onClick={e => {
+        e.preventDefault()
+        estimateWidget.open()
+      }}
       onKeyDown={e => {
         if (e.key === "Enter") {
           e.preventDefault()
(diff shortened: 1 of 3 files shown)


src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -116,7 +115,10 @@ export const ArtsyShippingEstimate = ({
   return (
     <LinkButton
       tabIndex={0}
-      onClick={() => estimateWidget.open()}
+      onClick={e => {
+        e.preventDefault()
```
oxaudo, line 119: 👍

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -116,7 +115,10 @@ export const ArtsyShippingEstimate = ({
...
+        estimateWidget.open()
```
oxaudo, line 120: Playing with it on staging now I wonder if this should work as a toggle as well - similar to how the whole section works as a toggle for shipping section. The whole toggle/untoggle thing seems to be slightly off to me now - I personally think that the whole section should start expanded and this link to estimate should go straight into the shipping section next to shipping description. But I think this works for now. We can def see if people even pay attention to it and then try to iterate from there.

src/Components/__tests__/ArtsyShippingEstimate.jest.tsx
```diff
@@ -0,0 +1,154 @@
...
+      query ArtsyShippingEstimate_Test_Query @relay_test_operation {
+        artwork(id: "example") {
+          ...ArtsyShippingEstimate_artwork
+        }
+      }
+    `,
+  })
+
+describe("ArtsyShippingEstimate", () => {
+  describe("with an artwork that cannot be estimated", () => {
```
oxaudo, line 78: Thanks for adding spec!

src/Components/__tests__/ArtsyShippingEstimate.jest.tsx
```diff
@@ -0,0 +1,154 @@
...
+    const mockArtworkData = {
```
oxaudo, line 79: #suggestion - maybe worth to be a bit more descriptive here. I would name it with some reference to why this can not be estimated. Like 'artworkMissingFullLocationData'? That is why it can not be estimated, right? 
Is that the case that Arta does not return estimate only based on country?

oxaudo (approved):
Very minor style suggestion. Feel free to merge as is.

src/Components/ArtsyShippingEstimate.tsx
erikdstock: 👍  I generally agree with making these tweaks after we get to test it a little bit

PR #15145, opened 2025-02-04
fix(EMI-2242): order stepper counteroffer flow

rquartararo:

This PR solves [EMI-2242]

### Description

This fixes the order stepper so it can handle the counteroffer flow, allowing a user to return to "Respond" once at the "Review" step. 

I did some refactoring to clean things up, however since we're redesigning the checkout and the stepper will soon be phased out, I believe this approach will be sufficient for now.

https://github.com/user-attachments/assets/21da4cb0-78cd-489f-9a10-87419d86d84a

[EMI-2242]: https://artsyproduct.atlassian.net/browse/EMI-2242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Components/OrderStepper.tsx
+++ b/src/Apps/Order/Components/OrderStepper.tsx
@@ -26,6 +26,43 @@ export interface OrderStepperProps {
   steps: string[]
 }
 
+export interface StepperComponentProps {
+  steps: string[]
+  currentStep: string
+  completedOrderSteps: string[]
+  onStepClick: (step: string) => void
+}
+
+const StepperComponent: FC<StepperComponentProps> = ({
+  steps,
+  currentStep,
+  completedOrderSteps,
+  onStepClick,
+}) => {
+  const stepIndex = steps.indexOf(currentStep)
+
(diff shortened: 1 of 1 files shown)


src/Apps/Order/Components/OrderStepper.tsx
```diff
@@ -62,66 +66,57 @@ export const OrderStepper: FC<React.PropsWithChildren<OrderStepperProps>> = ({
...
+              </Clickable>
+            ) : (
+              step
+            )
+          }
+          key={step}
+        />
+      ))}
+    </Stepper>
+  )
```
erikdstock, line 112: Can this be defined as a component outside of the body instead of a `render_()` function that returns JSX like a component?

erikdstock (commented):
LGTM. one question about a function which returns JSX and could be treated as a component (or memoized so it doesn't need to be re-declared on each render).

PR #15151, opened 2025-02-05
chore(EMI-2276): Tracking for shipping estimate widget

erikdstock:
cc @artsy/emerald-devs 

This PR solves [EMI-2276]
depends on https://github.com/artsy/cohesion/pull/555

### Description
This PR implements tracking for the shipping estimate test. We track both opening the widget and viewing an estimate, including the min/max values, origin/destination strings and currency. Because the widget doesn't have an api for the estimate event, the view event is triggered using a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) which watches the dom for changes.

Tests: I don't think this is worth testing due to the amount of mocking that would be required to manipulate the dom. I think we should manually test instead.

[EMI-2276]: https://artsyproduct.atlassian.net/browse/EMI-2276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Components/ArtsyShippingEstimate.tsx
+++ b/src/Components/ArtsyShippingEstimate.tsx
@@ -5,11 +5,11 @@ import type {
   SupportedCurrency,
 } from "@artaio/arta-browser/dist/MetadataTypes"
 import type ArtaEstimate from "@artaio/arta-browser/dist/estimate"
-import { Link, Spacer, Text } from "@artsy/palette"
+import { Link, Spacer, Text, usePrevious } from "@artsy/palette"
 import { useLoadScript } from "Utils/Hooks/useLoadScript"
 import { getENV } from "Utils/getENV"
 import type { ArtsyShippingEstimate_artwork$key } from "__generated__/ArtsyShippingEstimate_artwork.graphql"
-import { useEffect, useState } from "react"
+import { useCallback, useEffect, useRef, useState } from "react"
 import { graphql, useFragment } from "react-relay"
 import styled from "styled-components"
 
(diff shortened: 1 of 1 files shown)


src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -170,6 +176,104 @@ export const ArtsyShippingEstimate = ({
...
+
+interface UseWidgetObserverProps {
+  onViewEstimatedPrice: (price: PriceEstimate) => void
+}
+
+const useWidgetObserver = ({
+  onViewEstimatedPrice,
+}: UseWidgetObserverProps) => {
+  const widgetObserver = useRef<MutationObserver | null>(null)
+  const [visiblePrice, setVisiblePrice] = useState<PriceEstimate | null>(null)
```
erikdstock, line 194: This function got so big (mostly about setting up the DOM mutation observer) that i pulled it into its own hook below the component.

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -170,6 +176,104 @@ export const ArtsyShippingEstimate = ({
...
+
+    return { minPrice, maxPrice, currency, destination }
+  }, [])
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: We only want to run this when the price changes
+  useEffect(() => {
+    if (visiblePrice?.minPrice && visiblePrice?.maxPrice) {
+      onViewEstimatedPrice(visiblePrice)
+    }
+  }, [visiblePrice])
```
erikdstock, line 240: This is our tracking callback passed to the hook as an arg.

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -170,6 +176,104 @@ export const ArtsyShippingEstimate = ({
...
+
+const useWidgetObserver = ({
+  onViewEstimatedPrice,
+}: UseWidgetObserverProps) => {
+  const widgetObserver = useRef<MutationObserver | null>(null)
+  const [visiblePrice, setVisiblePrice] = useState<PriceEstimate | null>(null)
+
+  const extractEstimateFromDom = useCallback((): PriceEstimate | null => {
+    const priceAmountEl = document.getElementsByClassName(
+      "artajs__modal__quotes__price__amount",
```
oxaudo, line 198: This is so hacky! Relying on the dom selectors here. But it's a test - I'm ok with it :)

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -170,6 +176,104 @@ export const ArtsyShippingEstimate = ({
...
+    )[0]
+    const amountTextContent = priceAmountEl?.textContent
+    // e.g. "$1,000"
+    const [minPrice, maxPrice] =
+      amountTextContent
+        ?.replace(/[^0-9,.-]+/g, "")
+        ?.split("-")
+        .map(n => Number.parseFloat(n)) || []
```
oxaudo, line 206: 🙈

oxaudo (approved):
Feel free to merge @erikdstock.

Believe that is ok for the test - we should def reconsider this whole approach if we end up doing Estimates in the long run....

Like arta def should be able to let us access raw data for the estimate instead of us trying to parse it from DOM objects.

Ok for now I think though...

src/Components/ArtsyShippingEstimate.tsx
erikdstock: yeah...

PR #15162, opened 2025-02-06
fix: Disable Artsy logo link for Eigen

MrSltun:

This PR solves []

### Description
This PR updates the Artsy Logo `to` prop when the client is Eigen to disable navigating to the homepage in the Order App


gkartalis (commented):
beautiful

gkartalis (approved):

PR #15163, opened 2025-02-06
feat: adding more condition to arta estimate widget display

oxaudo:
This is not pretty as there is some trickiness with analyzing shipping weight. I'm not even convinced it's necessary but for now doing something hacky for this experiment. We do not have a way to get shippingWeight for editions so this part is skipped for now.

So think this is good enough for our experiment.

Will work on passing proper data to Arta next.


src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
```diff
@@ -90,15 +94,30 @@ export const ArtworkSidebar: React.FC<
...
+  const isOneEdition = isEdition && editionSets && editionSets.length === 1
+  let isWeightArtaEstimatable = true
+  // TODO: for now ignoring weight for works with one edition set.
+  // Need to either support in MP or roll data to the artwork level.
+  // TODO: have standard shippingWeightKg in MP?
+  if (!isEdition && shippingWeight && shippingWeightMetric) {
+    isWeightArtaEstimatable =
+      shippingWeightMetric === "kg"
+        ? shippingWeight <= 68
+        : shippingWeight <= 150
```
erikdstock, line 114: We talked about this outside of github - I suggested putting this instead in the function that creates the Shipping estimate request body (which may already return null based on the artwork shape and short-circuit the estimate) would be closer to where this kind of validation needs to happen. @oxaudo pointed out that at that point we are loading an entire extra library via script and I agreed it's fine to leave here for now.

minor: Could we update the comment to describe what this check is doing and make tickets or raise the TODOs in slack to make sure we come back to it

erikdstock (commented):
L G T M, one minor comment 👍

src/Apps/Artwork/Components/ArtworkSidebar/ArtworkSidebar.tsx
oxaudo: I'll add a placeholder card to 'make a decision and either clean or ticket the work once the future of Arta estimate experiment is decided". Does that sound right?

PR #15165, opened 2025-02-07
fix: expand scrollview on signed press

MounirDhahri:

### Description

**Context:** https://www.notion.so/artsy/mWeb-last-filter-group-signed-remains-hidden-when-expanded-18ccab0764a080c3afa6cb099e8b3f1c?pvs=4

This PR fixes the issue of the hidden last item on the filters.
https://github.com/user-attachments/assets/4447184e-43e5-415d-9114-3a6fe8646485

### DIFF
--- a/src/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
+++ b/src/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
@@ -4,8 +4,8 @@ import type React from "react"
 import { useRef } from "react"
 
 export const FilterExpandable: React.FC<
-  React.PropsWithChildren<ExpandableProps>
-> = ({ expanded, ...rest }) => {
+  React.PropsWithChildren<ExpandableProps & { ignoreBounds?: boolean }>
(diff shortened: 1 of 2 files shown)


src/Components/ArtworkFilter/ArtworkFilters/FilterExpandable.tsx
```diff
@@ -19,7 +19,7 @@ export const FilterExpandable: React.FC<
       const anchor = ref.current
       if (!anchor) return
       const { top } = anchor.getBoundingClientRect()
-      if (top < window.innerHeight) return
+      if (top < window.innerHeight && !ignoreBounds) return
```
MounirDhahri, line 22: The sticky button breaks the calculation and it should not be take in to account here. We didn't notice this earlier because previous sections were big so it didn't matter as much. I am open to suggestions for a better fix here

MounirDhahri: good idea

PR #15166, opened 2025-02-07
chore: align framed and signed copy with app

MounirDhahri:

This PR solves https://www.notion.so/artsy/Position-name-and-copy-of-framed-filter-differs-between-app-and-web-18ccab0764a0809b8440f23dad62f25f?pvs=4

### Description
align copy with app


olerichter00 (approved):
👍

PR #15175, opened 2025-02-10
chore: remove sell with artsy routes

MounirDhahri:

This PR solves [ONYX-1537]

### Description

This PR removes Sell with Artsy routes.

[ONYX-1537]: https://artsyproduct.atlassian.net/browse/ONYX-1537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MounirDhahri:
Tests are working fine but the review-app build is breaking for some reason.

anandaroop:
Not sure what's up with the review app, but I'll try pushing these changes to the earlier review app that I spun up last week.

anandaroop:
Looks like that push from this branch to my branch [succeeded](https://app.circleci.com/pipelines/github/artsy/force/67926/workflows/32291e05-c6e1-4732-8156-d0d4c2402bf4). 

That means we can use https://swa.artsy.net to review this.

(No idea why your branch didn't work 🤷🏽)

src/Apps/Sitemaps/sitemapsServerApp.tsx
```diff
@@ -83,7 +83,6 @@ sitemapsServerApp
       { loc: `${APP_URL}/categories` },
       { loc: `${APP_URL}/collect`, priority: 1 },
       { loc: `${APP_URL}/conditions-of-sale` },
-      { loc: `${APP_URL}/consign` },
```
anandaroop, line 86: Good call! I'm not sure we've updated the sitemap for other removed pages.

**suggestion:** Maybe do the same for L73 and L96.

damassi:
Yah, forgot to follow up but a fresh test worked over here: https://test-working.artsy.net/

src/Apps/Sell/Components/ArtistAutocomplete.tsx
anandaroop, line 1: **question:** Thoughts on hoisting this up to `/src/Components/`? 

It seems like a general-purpose component; and would be one less reason to keep the `Sell` folder hanging around.

src/Apps/Artist/artistRoutes.tsx
```diff
@@ -250,38 +239,6 @@ export const artistRoutes: RouteProps[] = [
           }
         `,
       },
-      {
-        path: "consign",
```
anandaroop, line 254: This should be accompanied by a 301 redirect — and I see that it _is_ …

```sh
# consign route redirects correctly...
❯ curl -s -I https://swa.artsy.net/artist/kaws/consign | grep -i -E "(30\d|location)"
HTTP/2 301 
location: /artist/laws

# ...as does any subpath apparently 😄
❯ curl -s -I https://swa.artsy.net/artist/kaws/lololjkjk | grep -i -E "(30\d|location)"
HTTP/2 301 
location: /artist/kaws

```

**question (non-blocking):** I'm not sure where in the stack this originates — do you know @damassi?

anandaroop (dismissed):
Thanks for the cleanup 🙏🏽

src/Apps/Artist/artistRoutes.tsx
damassi: I believe it comes from here: https://github.com/artsy/force/blob/review-app-test-working/src/Apps/Sitemaps/sitemapsServerApp.tsx/#L86

damassi: Not too sure!

olerichter00 (dismissed):
🧹 🧹 🧹

src/Apps/Artist/artistRoutes.tsx
MounirDhahri: Maybe that's not required because found matches in a depth first manner. https://4catalyzer.github.io/found/docs/configuration/route-config#path

PR #15181, opened 2025-02-10
fix: Disable logo clicks only in Order App

MrSltun:

This PR solves []

### Description

This PR is a follow-up on #15162 

This PR refactors the disabling of the navigating to the homepage to be only happening in the Order App


damassi (approved):
Thanks for the quick follow-up!

PR #15185, opened 2025-02-10
chore(EMI-2278): Include best available request params in estimate setup

erikdstock:

paired with @oxaudo 

Resolves [EMI-2278].

### Description

This PR adds more parameters to our estimates widget configuration to make the estimates as realistic as possible. 
#minor

### DIFF
--- a/src/Components/ArtsyShippingEstimate.tsx
+++ b/src/Components/ArtsyShippingEstimate.tsx
@@ -1,4 +1,4 @@
-import type { EstimateBody } from "@artaio/arta-browser"
+import type { EstimateBody, PartialEstimateConfig } from "@artaio/arta-browser"
 import type {
   ArtaLocation,
   ArtaObject,
@@ -324,7 +324,7 @@ const ARTWORK_FRAGMENT = graphql`
   }
 `
 
-const widgetConfig = {
+const widgetConfig: PartialEstimateConfig = {
   style: {
     color: {
       border: "none",
@@ -553,7 +553,7 @@ const artaLocation = (artwork: ShippableArtwork): ArtaLocation | null => {
     : null
 }
 
-const estimateRequestBodyForArtwork = (
+export const estimateRequestBodyForArtwork = (
   artwork: ShippableArtwork,
 ): EstimateBody | null => {
   try {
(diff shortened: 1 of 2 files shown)


src/Components/__tests__/ArtsyShippingEstimate.jest.tsx
```diff
@@ -44,24 +47,24 @@ jest.mock("Utils/Hooks/useLoadScript", () => {
 })
 
 const validArtworkData = {
-  depthCm: null,
-  diameterCm: null,
-  framedHeight: null,
-  framedWidth: null,
-  framedDepth: null,
-  framedDiameter: null,
+  depthCm: undefined,
```
damassi, line 50: Is this due to the API regarding null as a value, vs undefined being stripped somehow?

src/Components/__tests__/ArtsyShippingEstimate.jest.tsx
```diff
@@ -44,24 +47,24 @@ jest.mock("Utils/Hooks/useLoadScript", () => {
...
+  framedDiameter: undefined,
   framedMetric: "cm",
   heightCm: 100,
   isFramed: false,
   listPrice: {
     major: 1000,
   },
   mediumType: {
-    name: "Painting",
+    name: "Painting" as const,
```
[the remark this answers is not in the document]
erikdstock: The type is a string union in the 3rd party lib code and this seems to cause it to be recognized as a member of the union instead of just a `string`.

src/Components/ArtsyShippingEstimate.tsx
```diff
@@ -405,32 +403,32 @@ const UNFRAMED_CATEGORY_MAP = {
...
-  framedDepth?: number
-  framedDiameter?: number
-  framedHeight?: number
-  framedMetric?: string
-  framedWidth?: number
-  heightCm?: number
-  isFramed: boolean
-  listPrice: {
-    major?: number
+  depthCm?: number | null
```
erikdstock, line 406: Made the mirrored type here a better mirror. In retrospect we could probably have used a relay type for this prop if we don't need to support edition sets too. That might be some cleanup worth taking on later.

PR #15195, opened 2025-02-11
chore: set CREDIT_CARD as default payment method when available

oxaudo:
Results in this initial state:

Screenshot 2025-02-11 at 4 33 43 PM


src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -217,15 +217,14 @@ export const PaymentContent: FC<React.PropsWithChildren<Props>> = props => {
 const getAvailablePaymentMethods = (
   availablePaymentMethods: readonly CommercePaymentMethodEnum[],
 ): ReactElement<RadioProps<string>>[] => {
-  let paymentMethod: CommercePaymentMethodEnum = "CREDIT_CARD"
```
oxaudo, line 220: Not sure why we needed this and I was getting warning about non ideal assignments below.

src/Apps/Order/Routes/Payment/PaymentContent.tsx
```diff
@@ -238,32 +237,12 @@ const getAvailablePaymentMethods = (
...
-          </>
-        }
-      />,
-    )
-  }
-
-  // when available, unshift ACH since it's the first option we want to offer for US artworks
   if (availablePaymentMethods.includes("US_BANK_ACCOUNT")) {
-    paymentMethods.unshift(
+    paymentMethods.push(
```
oxaudo, line 241: The logic of pushing/unshifting was confusing and seemed unnecessary. So I just ended up assembling array in order. Nothing seems to break and it seems cleaner to me.

src/Apps/Order/Utils/__tests__/orderUtils.jest.ts
```diff
@@ -21,32 +21,32 @@ describe("order utils", () => {
       })
     })
 
-    describe("order payment is invalid", () => {
-      it("returns US_BANK_ACCOUNT if available payment methods includes it", () => {
+    describe("when currently selected order payment payment is invalid (paymentSet is false)", () => {
```
oxaudo, line 24: Took me a while to remember what is this `paymentSet` is about and how it correlates with `paymentMethod`. So tried to update descriptions to make it a bit clearer.

src/Apps/Order/Utils/orderUtils.ts
```diff
@@ -3,10 +3,10 @@ import type {
   Payment_order$data,
 } from "__generated__/Payment_order.graphql"
 
-const initialPaymentMethods: CommercePaymentMethodEnum[] = [
+const orderedPaymentMethods: CommercePaymentMethodEnum[] = [
```
oxaudo, line 6: Again. This this rename makes the intent of this var a bit more explicit.

src/Apps/Order/Routes/Payment/PaymentContent.tsx
erikdstock: Agree this logic is confusing. Another smell to me is defining a function that returns JSX called `getAnything()` - this is a functional component, it's just breaking the conventions of react. Is there any reason this couldn't be a normal react fc that takes a normal props interface and renders the elements in order with whatever conditionals are necessary?

```tsx
const PaymentMethodRadios = ({availablePaymentMethods}: PaymentMethodRadiosProps): JSX.Element => (
  <>
    {availablePaymentMethods.includes("CREDIT_CARD") && <BorderedRadio
        data-test-id="credit-card"
        key="CREDIT_CARD"
        value={paymentMethod}
        value="CREDIT_CARD"
        label={
          <>
            <UnknownCardIcon fill="black100" />
            <Text variant="sm-display" ml={0.5}>
              Credit card
            </Text>
          </>
        }
      />}
    {availablePaymentMethods.includes("US_BANK_ACCOUNT") && <BorderedRadio
        data-test-id="us-bank-account"
        key="US_BANK_ACCOUNT"
        value={(paymentMethod = "US_BANK_ACCOUNT")}
        value="US_BANK_ACCOUNT"
        label={
          <>
            <InstitutionIcon fill="green100" />
            <Text variant="sm-display" ml={0.5}>
              ACH bank transfer
            </Text>
          </>
        }
      >
        <Text ml="24px" variant="xs" color="black60">
          US bank account only
        </Text>
      </BorderedRadio>
    }
    /* ETC */
  </>
)

erikdstock (commented):
One big further improvement I think we can make, otherwise looks good & uncontroversial to me.

starsirius (approved):
🎉  LGTM! Eventually this will be handled with no code on Stripe!

PR #15205, opened 2025-02-13
fix(artwork): Attempt to fix mysterious flashing details

damassi:

This PR solves [PBRW-479]

### Description

First reported in the [bugs channel](https://artsy.slack.com/archives/C07PRTJSD6G/p1739468756482839), was noticed that on certain artworks there's a strange flashing behavior. 

Its hard to repo, but for now lets disable lazy loading (when scrolling into view) on that component. Its already lazy loaded via a QueryRenderer, so in a sense it just makes the page feel more stable (as an improvement). 

cc @artsy/amber-devs 

[PBRW-479]: https://artsyproduct.atlassian.net/browse/PBRW-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Artwork/ArtworkApp.tsx
+++ b/src/Apps/Artwork/ArtworkApp.tsx
@@ -69,25 +69,29 @@ interface BelowTheFoldArtworkDetailsProps {
 
 const BelowTheFoldArtworkDetails: React.FC<
   React.PropsWithChildren<BelowTheFoldArtworkDetailsProps>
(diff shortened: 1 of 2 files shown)


dzucconi (approved):

dzucconi:
The thing I'd be curious about is what field is causing this to be different.

PR #15231, opened 2025-02-19
fix(DIA-1098): Remove App Download CTA from /art-fairs 

JanaeHijaz:

This PR solves [DIA-1098]

### Description

I'm removing the FairsPhonePromo CTA and component since we don't want to continue using this. 

Before:
Screenshot 2025-02-19 at 2 23 19 PM

After:
Screenshot 2025-02-19 at 2 23 26 PM

[DIA-1098]: https://artsyproduct.atlassian.net/browse/DIA-1098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Apps/Fairs/Routes/FairsIndex.tsx
```diff
@@ -260,16 +259,6 @@ export const FairsIndex: React.FC<React.PropsWithChildren<FairsIndexProps>> = ({
...
-
-                {
-                  // Avoids orphaned CTA
-                  currentFairs[currentFairs.length - 1].bannerSize !==
-                    "x-large" && (
-                    <Column span={6}>
-                      <FairsPhonePromo pt={2} />
-                    </Column>
-                  )
-                }
```
JanaeHijaz, line 272: Question about this @damassi : I noticed that this page was displaying additional fairs in production that I don't see locally, so this <FairsPhonePromo /> component was not actually rendering for me when I removed it (the last fair.bannerSize === 'x-large' in my case locally), so this did not appear. But I'm not sure why.

damassi: I know we don't copy _all_ production data to staging every where -- so maybe its that?

damassi (approved):

PR #15232, opened 2025-02-19
feat: update myBids screens to respect sale isLiveOpen attribute

oxaudo:
Address EMI-2234

This solution is not ideal but as we currently don't have a good way to pull updated current bid info from live auction - add 'live bidding' instead of displaying bid status in relevant screens similar to how we display it on Artwork screen:
Screenshot 2025-02-19 at 4 32 26 PM

|Before|After|
|---|---|
|My Bids screen||
|Screenshot 2025-02-19 at 4 33 54 PM|Screenshot 2025-02-19 at 4 35 26 PM
|Auction screen| |
|Screenshot 2025-02-19 at 4 32 00 PM|Screenshot 2025-02-19 at 4 31 29 PM
|Auctions screen| |
|Screenshot 2025-02-19 at 5 00 17 PM|Screenshot 2025-02-19 at 5 00 46 PM

There is probably some specs failing and I need to figure out what data is available for myBids screen once the auction is closed and bring back either 'Won' or 'Outbid' info there.

### DIFF
--- a/src/Apps/Auction/Components/AuctionActiveBids.tsx
+++ b/src/Apps/Auction/Components/AuctionActiveBids.tsx
@@ -226,12 +226,8 @@ const BidButton: React.FC<
   const { router } = useRouter()
   const { tracking } = useAuctionTracking()
 
-  if (!lotStanding) return null
-
-  const { saleArtwork } = lotStanding
-  if (!saleArtwork) return null
-
-  const { sale } = saleArtwork
+  const saleArtwork = lotStanding?.saleArtwork
+  const sale = saleArtwork?.sale
   if (!sale) return null
 
   if (!!saleArtwork.endedAt) {
--- a/src/Apps/Auction/Components/AuctionActiveBids.tsx
+++ b/src/Apps/Auction/Components/AuctionActiveBids.tsx
@@ -19,6 +19,7 @@ import { RouterLink } from "System/Components/RouterLink"
 import { useRouter } from "System/Hooks/useRouter"
 import { usePoll } from "Utils/Hooks/usePoll"
 import { Media } from "Utils/Responsive"
+import { getENV } from "Utils/getENV"
 import type { AuctionActiveBids_me$data } from "__generated__/AuctionActiveBids_me.graphql"
 import {
   type RelayRefetchProp,
@@ -61,7 +62,7 @@ const AuctionActiveBids: React.FC<
       <Media at="xs">
         {me?.lotStandings?.map((lotStanding, index) => {
           return (
-            <>
+            <Box key={index}>
               <Box my={4}>
                 <Join separator={<Spacer y={0.5} />}>
                   <AuctionLotInfoFragmentContainer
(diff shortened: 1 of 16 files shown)


src/Apps/Auction/Components/AuctionActiveBids.tsx
```diff
@@ -218,11 +226,15 @@ const BidButton: React.FC<
...
+  if (!lotStanding) return null
+
+  const { saleArtwork } = lotStanding
+  if (!saleArtwork) return null
 
-  if (!!lotStanding.saleArtwork?.endedAt) {
+  const { sale } = saleArtwork
+  if (!sale) return null
+
+  if (!!saleArtwork.endedAt) {
```
erikdstock, line 237: I think this line needs to go after `if (!!sale.isLiveOpen)` - a live open sale will not be ended.

src/Apps/Auction/Components/AuctionActiveBids.tsx
```diff
@@ -218,11 +226,15 @@ const BidButton: React.FC<
...
-    return null
-  }
+  if (!lotStanding) return null
+
+  const { saleArtwork } = lotStanding
+  if (!saleArtwork) return null
 
-  if (!!lotStanding.saleArtwork?.endedAt) {
+  const { sale } = saleArtwork
+  if (!sale) return null
```
erikdstock, line 235: Could this be simplified to 
```ts
if (!lotStanding?.saleArtwork?.sale) { return null }`

File: src/Apps/Auction/Components/AuctionActiveBids.tsx [hunk 3]
```diff
@@ -237,10 +249,24 @@ const BidButton: React.FC<
...
+      </Button>
+    )
+  }
+
   // By default, we redirect to /artwork/id on successful bid, but since we're
   // in the ActiveBids component, redirect to /auction/id (i.e., close modal)
-  const redirectTo = `/auction/${lotStanding.saleArtwork?.saleID}`
-  const href = `/auction/${lotStanding.saleArtwork?.saleID}/bid/${lotStanding?.saleArtwork?.slug}?redirectTo=${redirectTo}`
+  const redirectTo = `/auction/${saleArtwork.saleID}`
+  const href = `/auction/${saleArtwork.saleID}/bid/${saleArtwork.slug}?redirectTo=${redirectTo}`
```
erikdstock, line 269: non-blocking: might be nice to move all these url constructors including the getENV one above into simple descriptively-named functions outside the component.

erikdstock (commented):
A few minor comments, nothing really blocking.

src/Apps/Auction/Components/AuctionActiveBids.tsx
oxaudo: This whole thing can use some major refactors. I did not change anything about this except of storing saleArtwork into a variable. If we end up working more in this area I think we should do bunch of consolidatinf between here and settings/auctions and MyBids on the auction page. For now not messing with any of this.

PR #15236, opened 2025-02-20
feat(EMI-2294): move and style express checkout element

rquartararo:

This PR solves [EMI-2294]

### Description

This PR:
- Moves the express checkout element to the shipping route page
- Does not display header of spacers unless their are available payment methods
- Styles to button to match designs 

### Express checkout with Apple Pay

https://github.com/user-attachments/assets/ca9a0977-eb73-41a1-aacf-a6880169bfb7

### Express checkout with no available payment methods 

https://github.com/user-attachments/assets/a02a1b96-16e1-422f-8a6c-8426b00c39bc

[EMI-2294]: https://artsyproduct.atlassian.net/browse/EMI-2294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Apps/Order/Routes/Payment/ExpressCheckoutPrototype/ExpressCheckout.tsx
+++ b/src/Apps/Order/Routes/Payment/ExpressCheckoutPrototype/ExpressCheckout.tsx
@@ -4,11 +4,14 @@ import {
   useElements,
   useStripe,
 } from "@stripe/react-stripe-js"
+import type { StripeExpressCheckoutElementOptions } from "@stripe/stripe-js"
+import { useState } from "react"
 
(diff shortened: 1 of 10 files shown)


damassi (commented):
Curious - whats up with the submit button loading state automatically being triggered on route transition? It feels a little thrash-y; can we fix that or is it necessary? If so, perhaps that action can be transparent to the user and instead of `loading` its simply `disabled`.

erikdstock (dismissed):

PR #15250, opened 2025-02-25
fix(notifications): adds missing geometry

dzucconi:
So this should be the only real stumbling block in the way of https://github.com/artsy/palette/pull/1423

And this seems like something that we should actually correct, of course. I'm tempted to also make width and height required params in the type signature on image, but I'll probably do that separately.

### DIFF
--- a/src/Components/Notifications/NotificationItem.tsx
+++ b/src/Components/Notifications/NotificationItem.tsx
@@ -75,7 +75,8 @@ const NotificationItem: FC<React.PropsWithChildren<NotificationItemProps>> = ({
                       key={index}
                       srcSet={image.resized.srcSet}
                       alt=""
-                      height={58}
+                      height={image.resized.height}
+                      width={image.resized.width}
                       lazyLoad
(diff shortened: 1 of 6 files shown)


damassi:
> width and height required params in the type signature on image

@dzucconi - this really isn't too practical at scale, though it makes sense in an ideal world. Can you elaborate a bit? This is sure to lock out numerous apps from updates to palette.

dzucconi:
how would it lock anyone out? they could be 100% for instance. you never want intrinsic geometry. anyway it would be a whole separate thing with the goal of inproving dx

dzucconi:
@damassi to be clear this image was missing the width attribute

damassi:
I understand we would never want intrinsic geometry --ideally-- but its common that one just passes width, or just passes height, or just passes maxWidth, or maxHeight, and all the variations in between. 


This component is used on a vast scale and so flexibility is best so as to not create a lot of unexpected work. And in general, flexibility here is good.

PR #15257, opened 2025-02-26
chore: clean up some unused props from fragments

oxaudo:
This just removes some unused props and the only thing that is added - show bid info for closed auctions if they were not live auction.


starsirius (approved):
👍

oxaudo:
@starsirius - one more addition here. After looking at some data figuring that it's better not to display any bidding info for live auction actively in Bidding:

Screenshot 2025-02-27 at 4 18 42 PM

Screenshot 2025-02-27 at 4 18 57 PM

PR #15262, opened 2025-02-27
feat(EMI-2307): create CheckoutApp

fladson:

This PR solves [EMI-2307]

### Description

Barebones app rendered under a feature flag to support the development of the new checkout flow.

SCR-20250227-oyry

@artsy/emerald-devs 

[EMI-2307]: https://artsyproduct.atlassian.net/browse/EMI-2307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fladson:
Not sure if we need the specs at this point.

damassi (dismissed):
Nice @fladson, welcome 👍 

And no need for specs right now! With react, and functional programming in general -- and especially with greenfield apps -- its often vastly more efficient to add specs towards the end, once yr sure that absolutely nothing will change. Otherwise you start to lock yourself into a rigid structure too early, when there's little risk. (Just my suggestion, though I know others have different opinions.)

The same goes for mobile responsive pages. If you start to wire that in before the desktop version is done, one ends up just creating unnecessary complexity, and rigidity, hampering iteration. Given our flexible tools, responsive work can also come at the end and it will only serve  to further.

src/Apps/Checkout/checkoutRoutes.tsx
```diff
@@ -0,0 +1,35 @@
...
+  () => import(/* webpackChunkName: "checkoutBundle" */ "./CheckoutApp"),
+  {
+    resolveComponent: component => component.CheckoutApp,
+  },
+)
+
+const CheckoutOverviewRoute = loadable(
+  () =>
+    import(
+      /* webpackChunkName: "exampleBundle" */ "./Routes/Welcome/CheckoutOverviewRoute"
```
starsirius, line 14: I think we can simplify and get rid of the nested `Welcome` structure for the routes.

starsirius (commented):
One suggestion on the nested routes. Otherwise, LGTM! Start of the a new era! 🚀

src/Apps/Checkout/checkoutRoutes.tsx
damassi: Also note the `exampleBundle` -- thats the name of the JS file that will be auto-generated at this route, so we'll want to name it right

fladson: True, my bad for not noticing it, thanks for catching it!

ArtsyOpenSource:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 88fb34c9d0be31921276c5073f5f9a59e403127e
</p>

